PR #7, opened 2017-05-17
Register prometheus metrics based on name

checketts:
I made the default constructor use the default Registry so the endpoint would display the same metrics.

Also the keys for the maps needed to change since the Prometheus metric registry only uses names and the metric object's hashcodes don't include the labels so we needed to use that as the hashkey for the `meterMap`

### DIFF
--- a/src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
+++ b/src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
@@ -36,13 +36,13 @@ import static org.springframework.metrics.instrument.internal.MeterId.id;
 public class PrometheusMeterRegistry extends AbstractMeterRegistry {
     private final CollectorRegistry registry;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
```diff
@@ -36,13 +36,13 @@
 public class PrometheusMeterRegistry extends AbstractMeterRegistry {
     private final CollectorRegistry registry;
 
-    private final Map<MeterId, Collector> collectorMap = new HashMap<>();
+    private final Map<String, Collector> collectorMap = new HashMap<>();
```
checketts: Should this be a `ConcurrentHashMap` to ensure thread safety?

jkschneider: Done in e280fc9864f4a4ed3ca0bf39e3200ae8b88fec4e

PR #26, opened 2017-05-28
Return FilterReply.NEUTRAL instead of ACCEPT.

mlausch1963:
Returning NEUTRAL let the other filters decide what to do. Otherwise
logging configuration is not obeyed and everything is logged.


jkschneider:
Thanks @mlausch1963!

Issue #22, opened 2017-05-24
Instrument `ThreadPoolTaskExecutor` and `ThreadPoolTaskScheduler`


jkschneider:
In the case of `ThreadPoolTaskExecutor`, can monitor by adding a `TaskDecorator`.

`ThreadPoolTaskScheduler` will have to be wrapped at instantiation time.

jkschneider:
Continuation of #15

PR #29, opened 2017-05-30
Threadpool exec binder

checketts:
Solves #22.  Until now we haven't used Binders this way. It feels like a good approach. What do you think @jkschneider ?

### DIFF
--- a/src/main/java/org/springframework/metrics/instrument/scheduling/ExecutorServiceMetrics.java
+++ b/src/main/java/org/springframework/metrics/instrument/scheduling/ExecutorServiceMetrics.java
@@ -1,0 +1,31 @@
+package org.springframework.metrics.instrument.scheduling;
+
+import org.springframework.metrics.instrument.MeterRegistry;
+import org.springframework.metrics.instrument.binder.MeterBinder;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadPoolExecutor;
+
+public class ExecutorServiceMetrics implements MeterBinder {
+
+  private final String name;
+  private final ExecutorService executorService;
+
+  public ExecutorServiceMetrics(String name, ExecutorService executorService) {
+    this.name = name;
+    this.executorService = executorService;
+  }
+
+  @Override
+  public void bindTo(MeterRegistry registry) {
+    if(executorService instanceof ThreadPoolExecutor) {
+      ThreadPoolExecutor tp = (ThreadPoolExecutor) executorService;
+      registry.gauge(name+"_scheduled_total", tp, ThreadPoolExecutor::getTaskCount); //Really should be a counter
+      registry.gauge(name+"_completed_total", tp, ThreadPoolExecutor::getCompletedTaskCount); //Really should be a counter
+      registry.gauge(name+"_active_count", tp, ThreadPoolExecutor::getActiveCount);
+      registry.gauge(name+"_queue_size", tp, (i) -> i.getQueue().size());
+      registry.gauge(name+"_pool_size", tp, ThreadPoolExecutor::getPoolSize);
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


src/test/java/org/springframework/metrics/instrument/scheduling/ExecutorServiceMetricsTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assertions.*;
+
+class ExecutorServiceMetricsTest {
+
+  @DisplayName("tp task executor can be instrumented after being initialized")
+  @ParameterizedTest
+  @ArgumentsSource(MeterRegistriesProvider.class)
+  void taskExecutor(MeterRegistry registry) {
```
jkschneider: What does this test prove differently than the other test in this class?

src/main/java/org/springframework/metrics/instrument/scheduling/ExecutorServiceMetrics.java
```diff
@@ -0,0 +1,31 @@
...
+  public ExecutorServiceMetrics(String name, ExecutorService executorService) {
+    this.name = name;
+    this.executorService = executorService;
+  }
+
+  @Override
+  public void bindTo(MeterRegistry registry) {
+    if(executorService instanceof ThreadPoolExecutor) {
+      ThreadPoolExecutor tp = (ThreadPoolExecutor) executorService;
+      registry.gauge(name+"_scheduled_total", tp, ThreadPoolExecutor::getTaskCount); //Really should be a counter
```
jkschneider: Can't be a counter because it does not have monotonicity?

src/main/java/org/springframework/metrics/instrument/scheduling/ExecutorServiceMetrics.java
```diff
@@ -0,0 +1,31 @@
...
+      registry.gauge(name+"_completed_total", tp, ThreadPoolExecutor::getCompletedTaskCount); //Really should be a counter
```
jkschneider: Any reason not to make it a counter like you suggest?

src/test/java/org/springframework/metrics/instrument/scheduling/ExecutorServiceMetricsTest.java
checketts: Just validating the scheduler not supporting the decorator. I wouldn't mind deleting it.

 Also, would it make sense to add the task decorator upstream?

src/main/java/org/springframework/metrics/instrument/scheduling/ExecutorServiceMetrics.java
checketts: Counters don't have the `ToDoubleFunction<T> f`, like gauges. If you agree I can add that ability to the Counter signature.

jkschneider: Wow, yeah. That's interesting. This would likely result in a custom collector for Prometheus (not that this is a show stopper). How would we guarantee monotonicity from `f` across sampling intervals?

checketts: In this case we can assume the guarantee since it is documented to always go up in the ThreadPoolExecutor.

The root problem may be that we need a way to create custom collectors against the spring-metrics api instead of bloating the `Registry` API? 

Would that be a problem for Spectator metrics?

jkschneider: It would not be a problem for Spectator. The semantics are very similar to Prometheus, as you can see here: https://github.com/spring-projects/spring-metrics/blob/master/src/main/java/org/springframework/metrics/instrument/spectator/SpectatorToDoubleGauge.java#L37

jkschneider:
This work was cherry picked into 6221a6947077f7efffd515908d0b3325f9f75e8f and then substantially altered in 97e3e9aa479d638be89426b609b2670fa2c1404c to support a wider range of pooled executors.

PR #33, opened 2017-05-31
Add MeterReporter that allows registering customer groupings of metrics

checketts:
Initial pass at `MeterReporter` implementation. I wanted some initial feedback before proceeding too far.

The test at the end demonstrates the usage pattern best.

### DIFF
--- a/src/main/java/org/springframework/metrics/instrument/Meter.java
+++ b/src/main/java/org/springframework/metrics/instrument/Meter.java
@@ -20,4 +20,12 @@ package org.springframework.metrics.instrument;
  */
 public interface Meter {
     String getName();
+
+    enum Type {
+        COUNTER,
+        GAUGE,
+        DISTRIBUTION_SUMMARY,
+        LONG_TASK_TIMER,
+        TIMER
+    }
 }
--- a/src/main/java/org/springframework/metrics/instrument/MeterRegistry.java
+++ b/src/main/java/org/springframework/metrics/instrument/MeterRegistry.java
@@ -501,4 +501,6 @@ public interface MeterRegistry {
     default ExecutorService monitor(String name, ExecutorService executor) {
         return monitor(name, emptyList(), executor);
     }
+
+    MeterRegistry monitor(MeterReporter meterReporter);
 }
--- a/src/main/java/org/springframework/metrics/instrument/MeterReporter.java
+++ b/src/main/java/org/springframework/metrics/instrument/MeterReporter.java
@@ -1,0 +1,15 @@
+package org.springframework.metrics.instrument;
+
+import org.springframework.metrics.instrument.binder.MeterBinder;
+
+import java.util.List;
+
+abstract public class MeterReporter implements MeterBinder{
+
+  @Override
+  public void bindTo(MeterRegistry registry) {
+    registry.monitor(this);
+  }
+
+  abstract public List<MeterSamples> report();
+}
--- a/src/main/java/org/springframework/metrics/instrument/MeterReporter.java
+++ b/src/main/java/org/springframework/metrics/instrument/MeterReporter.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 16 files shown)


jkschneider:
## MeterReporter

I've thought for a while that there may come a time when this would need to be added to `Meter` itself:

```java
  /**
   * Get the set of measurements for this meter.
   */
  Iterable<Measurement> measure();
```

This is the Spectator equivalent of Prometheus':

```java
public abstract List<MetricFamilySamples> collect()
```

I find the `MetricFamilySamples` structure from Prometheus unnecessarily complex, but also would favor dropping the `timestamp` field in the version of `Measurement` found in Spectator.

There is some utility in adding this even to our existing `Meter` implementations, if nothing else as a quick debugging tool for folks new to monitoring to internalize that `Meter`s may generate multiple time series in the monitoring backend (and just what those time series are for a particular `Meter`).

So suppose we add a method to `MeterRegistry` similar to what you have added:

```java
MeterRegistry monitor(Meter meter)
```

The implementation of this method in `PrometheusMeterRegistry` and `SpectatorMeterRegistry` will map to a custom `Collector` and a custom `com.netflix.spectator.api.Meter`, respectively. This extra level of mapping is an extra layer of garbage created on each sampling, but such is the cost of abstraction I suppose.

For our existing supported `Meter`s like `Timer`, `Gauge`, etc. `measure()` never gets called by the underlying registry and we continue to allow the underlying registry to use its native mechanism to extract the time series. No additional garbage per sampling.

All of this would eliminate the need for a new `MeterReporter` interface.

One last detail concerning `Meter.Type`: I'm not sure what Prometheus does if you create a custom collector that doesn't fit one of the defined types. I'm pretty sure in Spectator land, type is just a tag convention.

## MeterBinder

I'm still a little conflicted about this interface, and your PR from yesterday around `ThreadPoolExecutor`s makes me more conflicted. I think you are right that having a concept like binders eases the bloat pressure on `MeterRegistry` itself, and maybe we've gone too far already in bloating the API by adding `monitor(Cache)`, `monitor(DataSource)`, etc.

I originally imagined `MeterBinder` to be used when:

1. Multiple meters are going to be registered
2. Something about the system is being observed, rather than a particular instance of something.

I don't have a good reason for (2).

But faced with these two options, I prefer the former:

1. More methods on `MeterRegistry`:
```java
@Bean
ExecutorService pool(MeterRegistry registry) {
   return registry.monitor("beep_pool", Executors.newSingleThreadExecutor());
}
```

2. Binder approach:
```java
@Bean
ExecutorService pool(MeterRegistry registry) {
   ExecutorService pool = Executors.newSingleThreadExecutor();
   registry.bind(new ExecutorMetrics("beep_pool", pool));
   return pool;
}
```

So I'm open to your thoughts on this.

checketts:
> I find the MetricFamilySamples structure from Prometheus unnecessarily complex

Agreed.

> but such is the cost of abstraction I suppose

I think there is a way to avoid the abstraction. We could skip the prometheus registry and just generate our own output. We already are duplicating the tracking metrics in the `PrometheusMeterRegistry`. The only downside there is if the user has their own instance of a `PrometheusRegistry` they wanted to use directly, but I think we could account for that.

Also the good news is that duplication only happens when `Collect` is called. So it is relatively short lived, but I think minimizing memory would be nice.

> So suppose we add a method to MeterRegistry

I think that will work great. It also helps avoid the need for how I used `Binder`.  You mentioned preferring a monitor method per type, I think that works for primary uses. However we'll hit some hard snags when wanting to monitor Guava, Caffeine, and Ehcache caches since each would be a different type and thus a different monitor method. So I think `monitor(Meter meter)` will definitely be needed.


# Outcome:

* We scrap `MeterReporter` and avoid my current attempts at using `Binder`
* Add `Iterable<Measurement> measure()` to `Meter` (without the timestamp)

Would you like me to do that work in this PR? Or did you want to tackle the `Iterable<Measurement> measure()` part?

jkschneider:
Feel free to take a stab at as much as you have time for! I'll be occupied for a little while longer figuring out the histogram.

jkschneider:
> We could skip the prometheus registry and just generate our own output.

Interesting idea. After all, a prometheus scrape expects a text/plain response and we know how to do String concatenation ;)

checketts:
> Iterable<Measurement> measure();

I'm going down this path. However how should I treat `Meter.Type`? I kind of favor making it optional on `Measurement` that way a service could report it, but it wouldn't have to.

jkschneider:
> However how should I treat Meter.Type?

I've wondered the same thing about Prometheus custom collectors. Do they always have a type associated with them?

checketts:
They return the `MetricsFamilySamples` and those families are the metric type.

jkschneider:
@checketts Could you commit whatever you've got? I've ran into a need for the heisen-counter as well and would be happy to finish this off to unblock.

checketts:
Pushed branch https://github.com/spring-projects/spring-metrics/tree/add-measures

It is highly incomplete and doesn't compile

PR #34, opened 2017-06-01
Fix closing code brackets for Prometheus documentation

Starefossen:
This PR adds closing code brackets to a code example under the Prometheus section in the documentation which broke asciidoc rendering.


jkschneider:
Good catch @Starefossen. Thanks!

PR #36, opened 2017-06-07
Add custom meter registration

jkschneider:
As a side effect:

* Every `Meter` now implements `Iterable<Measurement> measure()`
* Every `Meter` now implements `Iterable<Tag> getTag()`.
* `MeterRegistry` now has a `MeterRegistry register(Meter meter)` method.


jkschneider:
Replaces #33.

src/main/java/org/springframework/metrics/instrument/DistributionSummary.java
```diff
@@ -63,10 +64,14 @@ default Builder tags(Iterable<Tag> tags) {
             return this;
         }
 
+        default Builder tags(String... tags) {
+            return tags(Tag.tags(tags));
```
checketts: This line makes me smile

src/main/java/org/springframework/metrics/instrument/Measurement.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.metrics.instrument;
+
+import java.util.*;
```
checketts: Any style guide against star imports?

src/main/java/org/springframework/metrics/instrument/Measurement.java
```diff
@@ -0,0 +1,99 @@
...
+ * A measurement sampled from a meter.
+ *
+ * @author Clint Checketts
+ * @author Jon Schneider
+ */
+public final class Measurement {
+
+    private final String name;
+    private final SortedSet<Tag> tags = new TreeSet<>(Comparator.comparing(Tag::getKey));
+    private final double    value;
```
checketts: spacing

checketts: Why do we keep name and tags separate? Why not reuse MeterId? As a bonus the Measurement can just reuse the reference to the parent Meter's MeterId

src/main/java/org/springframework/metrics/instrument/prometheus/CustomCollectorChild.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.metrics.instrument.Measurement;
+import org.springframework.metrics.instrument.Tag;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+public interface CustomCollectorChild {
+    Stream<Collector.MetricFamilySamples.Sample> collect();
```
checketts: The Prometheus interface is a `List`, was the stream intentional?

src/main/java/org/springframework/metrics/instrument/Measurement.java
jkschneider: I'm trying to keep `MeterId` internal now for a few reasons:

1. I think it winds up creating a less friendly API for the end user. Spectator does this and
`registry.gauge(registry.createId(name, tags), ...).` just doesn't seem as nice as
`registry.gauge(name, tags, ...)`.

2. There are certain contexts where name and tags occur together and the tags need to be sorted (like here), and some where they do not.

3. I can't think of a good name for it. In the context of a measurement, it's really a different type of id - a name that is potentially prefixed by the containing meter's name, and a set of tags that may be a superset of the meter's tags.

jkschneider: Of course, I've agonized over the name/tag combination a few dozen times already. There is likely room for improvement!

src/main/java/org/springframework/metrics/instrument/prometheus/CustomCollectorChild.java
jkschneider: The set of samples from all the children wind up getting flattened in the final `MetricFamilySamples`

src/main/java/org/springframework/metrics/instrument/prometheus/CustomPrometheusLongTaskTimer.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        public int activeTasks() {
+            return tasks.size();
+        }
+    }
+
+    @Override
+    public List<MetricFamilySamples> collect() {
+        return Collections.singletonList(new MetricFamilySamples(name, Type.UNTYPED, " ", children.stream()
```
checketts: `UNTYPED` is getting lossy. One metric is a gauge and one is a counter. Shouldn't we try to surface that? Even if it is just to document that one always goes up and the other doesn't?

src/main/java/org/springframework/metrics/instrument/Measurement.java
jkschneider: I'm using the IntelliJ default.

src/main/java/org/springframework/metrics/instrument/prometheus/CustomPrometheusLongTaskTimer.java
jkschneider: Absolutely, there is room for improvement here. I started creating a `Measurement.Type` and got a little lost in the weeds of how spectator is typing its time series.

jkschneider: Noted in #38.

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
```diff
@@ -132,12 +136,38 @@ public double get() {
...
+                            }
+                            return new MetricFamilySamples.Sample(m.getName(), tagKeys, tagValues, m.getValue());
+                        })
+                        .collect(toList());
+
+                return Collections.singletonList(new MetricFamilySamples(meter.getName(), Type.UNTYPED, " ", samples));
+            }
+        };
+        registry.register(collector);
+        collectorMap.put(meter.getName(), collector);
```
checketts: Should we check the collectorMap before registering?

src/test/java/org/springframework/metrics/instrument/MetersTest.java
```diff
@@ -0,0 +1,65 @@
...
+ * @author Jon Schneider
+ */
+class MetersTest {
+    @Test
+    void customMeter() {
+        NavigableSet<String> letters = new TreeSet<>();
+        letters.addAll(Arrays.asList("a", "b", "c", "d", "e"));
+
+        Meter customMeter = Meters.build("letters_after")
+                .tags("letter", "b")
```
jkschneider: @checketts Is it apparent enough that these tags get added to every measurement?

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
```diff
@@ -132,12 +136,38 @@ public double get() {
...
+            public List<MetricFamilySamples> collect() {
+                List<MetricFamilySamples.Sample> samples = stream(meter.measure().spliterator(), false)
+                        .map(m -> {
+                            List<String> tagKeys = new ArrayList<>(m.getTags().size());
+                            List<String> tagValues = new ArrayList<>(m.getTags().size());
+                            for (Tag tag : m.getTags()) {
+                                tagKeys.add(tag.getKey());
+                                tagValues.add(tag.getValue());
+                            }
+                            return new MetricFamilySamples.Sample(m.getName(), tagKeys, tagValues, m.getValue());
```
checketts: I think we should include a `groupingBy` on the metric names to generate the MetricFamilySamples for each differing sub measure.

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
jkschneider: I thought about this a bit too. Question is: what would we do if there was such a name already in the list? If the answer is blow up, `registry.register` is going to do that for us. Otherwise we are left with some sort of log and squash approach.

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusTimer.java
```diff
@@ -16,31 +16,32 @@
...
 public class PrometheusTimer extends AbstractTimer {
     private CustomPrometheusSummary.Child summary;
-    private Quantiles quantiles;
 
-    public PrometheusTimer(String name, CustomPrometheusSummary.Child summary, Clock clock, Quantiles quantiles) {
-        super(name, clock);
+    PrometheusTimer(MeterId id, CustomPrometheusSummary.Child summary, Clock clock) {
+        super(id, clock);
         this.summary = summary;
-        this.quantiles = quantiles;
```
checketts: Were quantiles moved up into AbstractTimer?

src/main/java/org/springframework/metrics/instrument/simple/SimpleGauge.java
```diff
@@ -38,6 +42,16 @@ public double value() {
...
+    }
+
+    @Override
+    public Iterable<Tag> getTags() {
+        return id.getTags();
+    }
+
+    @Override
+    public Iterable<Measurement> measure() {
+        return Collections.singletonList(id.withTags(Tag.of("type", "GAUGE")).measurement(value()));
```
checketts: It might make sense to have the Gauge type tags as constants

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
jkschneider: Interesting idea. In the case of a `Histogram`, Prometheus reports two different names under the same `MetricFamilySamples`. In fact, I couldn't find a single Prometheus collector that returned more than one `MetricFamilySamples`.

jkschneider: i.e. Do two or more `MetricFamilySamples` imply multiple meters?

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusTimer.java
jkschneider: Rather moved down into the custom summary. This was a holdover from an earlier attempt, and I goofed by not removing it here. Results in the value being observed to the quantiles multiple times per.

src/main/java/org/springframework/metrics/instrument/spectator/SpectatorMeterRegistry.java
```diff
@@ -144,9 +149,38 @@ public LongTaskTimer longTaskTimer(String name, Iterable<Tag> tags) {
     }
 
     @Override
+    public MeterRegistry register(Meter meter) {
+        com.netflix.spectator.api.Clock spectatorClock = new com.netflix.spectator.api.Clock() {
```
checketts: Should this just be an instance variable?

src/main/java/org/springframework/metrics/instrument/simple/SimpleGauge.java
jkschneider: #38, look here!

src/main/java/org/springframework/metrics/instrument/spectator/SpectatorMeterRegistry.java
jkschneider: Good point. 05f62d19fcda0ab2d86865d82f23ac5b4042e49d

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
checketts: > Do two or more MetricFamilySamples imply multiple meters?

No take the [guava cache collector](https://github.com/prometheus/client_java/blob/master/simpleclient_guava/src/main/java/io/prometheus/client/guava/cache/CacheMetricsCollector.java#L91) as an example

src/main/java/org/springframework/metrics/instrument/DistributionSummary.java
jkschneider: It's a really embarrassing line. Such is the obsession with flexibility.

jkschneider: I thought about letting IntelliJ go hog wild with a static import. But then it would have been `tags(tags(tags))`.

src/test/java/org/springframework/metrics/instrument/MetersTest.java
```diff
@@ -0,0 +1,65 @@
...
+                    after.retainAll(Arrays.asList("a", "e", "i", "o", "u", "y"));
+                    Measurement vowels = new Measurement(name,
+                            tags("statistic", "vowels"),
+                            after.size());
+
+                    return Arrays.asList(total, vowels);
+                });
+
+        assertThat(customMeter.measure())
+                .allSatisfy(m -> assertThat(m.getTags()).contains(Tag.of("letter", "b")))
```
checketts: > Is it apparent enough that these tags get added to every measurement?

You could make it a little more apparent by spelling it out in the assertion:

```
.allSatisfy(m -> assertThat(m.getTags()).as("letter tag is automatically added to all measurements").contains(Tag.of("letter", "b")))
```

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
jkschneider: Is that really more of a `MeterBinder` in our language -- i.e. something registering multiple counters?

src/test/java/org/springframework/metrics/instrument/MetersTest.java
jkschneider: Sure, but from the perspective of a custom meter in your application code?

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
checketts: > Is that really more of a MeterBinder in our language -- i.e. something registering multiple counters?

I like that actually. I'm used to Prometheus where both types are collectors, but I think using binders for integrations makes good sense and keeps integrators from having to dig into internals too much.

src/main/java/org/springframework/metrics/instrument/prometheus/CustomPrometheusLongTaskTimer.java
checketts: I'm getting the feeling that types are largely for documentation purposes.
https://prometheus.io/docs/instrumenting/writing_exporters/#types

I think `UNTYPED` is a fine default.

At a meetup I recall them mentioning how 'someday they would like to do something with `#HELP`, I suspect `#TYPE` is similar. And only counters are really interesting because then the user knows they can count on the value going up and using `rate` or `increase` makes sense.

src/main/java/org/springframework/metrics/instrument/prometheus/PrometheusMeterRegistry.java
jkschneider: This did make me think though about similar situations, and #39.

src/main/java/org/springframework/metrics/instrument/prometheus/CustomPrometheusLongTaskTimer.java
jkschneider: I have a note to talk to Brian Harrington about the effect type has on Atlas.

src/test/java/org/springframework/metrics/instrument/MetersTest.java
checketts: I believe so.  I'll revisit the sample app with the latest updates and try to get a feel for how the api has evolved. The guava and caffeine caches in Prometheus use a mechanism like that to set the cache name for all the samples it outputs: https://github.com/prometheus/client_java/blob/master/simpleclient_guava/src/main/java/io/prometheus/client/guava/cache/CacheMetricsCollector.java#L34

checketts:
Great changes! Thanks

PR #47, opened 2017-07-18
Fix `MetricsSchedulingAspect` bean condition

Camelion:
`MetricsSchedulingAspect` bean no needs in RestTemplate class, look like mistype


jkschneider:
Thanks @Camelion

PR #50, opened 2017-07-22
Simple change for base interface in AbstractStepRegistry constructor

Camelion:
and restore imports for AnnotationUtils class in micrometer-spring module


jkschneider:
Ah, great catch.

PR #51, opened 2017-07-23
Quantiles and possibly histogram buckets observing 

Camelion:
for spectator and simple meter registries

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Observer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Observer.java
@@ -13,15 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micrometer.core.instrument.stats.hist;
 
-import io.micrometer.core.instrument.Observer;
-
-import java.util.Collection;
+package io.micrometer.core.instrument;
 
 /**
- * @author Jon Schneider
+ * @author Dmitry Poluyanov
+ * @since 23.07.17
  */
-public interface Histogram<T> extends Observer {
-    Collection<Bucket<T>> getBuckets();
+public interface Observer {
+    /**
+     * Add a sample
+     * @param value
+     */
+    void observe(double value);
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/ObserverHolder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/ObserverHolder.java
@@ -13,15 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micrometer.core.instrument.stats.hist;
 
-import io.micrometer.core.instrument.Observer;
-
-import java.util.Collection;
+package io.micrometer.core.instrument;
 
 /**
- * @author Jon Schneider
+ * @author Dmitry Poluyanov
+ * @since 23.07.17
  */
-public interface Histogram<T> extends Observer {
-    Collection<Bucket<T>> getBuckets();
+public interface ObserverHolder {
+
+    void observe(double amount);
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/datadog/DatadogRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/datadog/DatadogRegistry.java
@@ -38,7 +38,7 @@ import static java.util.stream.Collectors.joining;
  *
  * @author Jon Schneider
  */
-final class DatadogRegistry extends AbstractStepRegistry {
+public final class DatadogRegistry extends AbstractStepRegistry {
 
     private final URL metricsEndpoint;
     private final String hostTag;
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractDistributionSummary.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractDistributionSummary.java
@@ -13,15 +13,25 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micrometer.core.instrument.stats.hist;
 
-import io.micrometer.core.instrument.Observer;
+package io.micrometer.core.instrument;
 
-import java.util.Collection;
+import io.micrometer.core.instrument.util.MeterId;
 
 /**
- * @author Jon Schneider
+ * @author Dmitry Poluyanov
+ * @since 23.07.17
  */
-public interface Histogram<T> extends Observer {
-    Collection<Bucket<T>> getBuckets();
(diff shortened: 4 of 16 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/ObserverHolder.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package io.micrometer.core.instrument;
+
+/**
+ * @author Dmitry Poluyanov
+ * @since 23.07.17
+ */
+public interface ObserverHolder {
```
adriancole: eyes are crossed.. I can't see any difference between this and Observer. The combination of this + adding Abstracts seems a bit more types than needed at least today. Libraries with a lot of types to say the same thing can put off certain users. Is this pulling its cognitive weight?

micrometer-core/src/main/java/io/micrometer/core/instrument/datadog/DatadogRegistry.java
```diff
@@ -38,7 +38,7 @@
  *
  * @author Jon Schneider
  */
-final class DatadogRegistry extends AbstractStepRegistry {
+public final class DatadogRegistry extends AbstractStepRegistry {
```
jkschneider: The original intent here was to hide Spectator as an implementation detail. Might it be used directly?

micrometer-core/src/main/java/io/micrometer/core/instrument/Observer.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package io.micrometer.core.instrument;
+
+/**
+ * @author Dmitry Poluyanov
+ * @since 23.07.17
+ */
+public interface Observer {
```
jkschneider: Is the `Observer` intended as an abstraction for a variety of summary statistics commonly found on timers and distribution summaries?

If not, read no further...

If so, note that we currently have two things you could call a summary stat: client side quantiles and histograms. 

Client-side quantiles are errorless (because of course Micrometer sees all the samples), but are not aggregable in general. For this reason, I believe they are of little value in systems of sufficient scale. By sufficient scale, I mean when the number of instances (or containers) reporting a metric is large enough that grouping the metric by instance results in an impractical number of lines to plot on a graph and still be able to visualize an appreciable difference between instances.

Histograms really are a discretization of the continuous input domain whose chief purpose is to compute an aggregable quantile approximation through server-side math in the monitoring system. Their weakness lies in the essentially unbounded potential of their error bound, depending on how well (or poorly) the buckets are chosen. And the bucket choosing problem is unfortunately non-trivial.

If there was a way for us to infer a good bucket selection that would yield an acceptable error bound, I would be very tempted to strike client-side quantile support entirely (or at least move it out of timers and distribution summaries as a toy that you can record separately). I just need a little more time to research this. I believe Netflix has some recent research on a fixed bucket set that empirically works well for a wide variety of timers.

There are other summary stats I expect we'll hear requests for at some point, e.g. specifically client-side standard deviation. Many of them assume a normal distribution, which experience has shown is unfortunately rare.

All this is a super long winded way of saying: we may not ultimately need an abstraction if the only summary stat is a histogram?

micrometer-core/src/main/java/io/micrometer/core/instrument/datadog/DatadogRegistry.java
Camelion: Otherwise it can't be used in [DatadogMetricsConfiguration](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-spring/src/main/java/org/springframework/boot/metrics/export/datadog/DatadogMetricsConfiguration.java#L39) but i change it in another way, and want update my PR later

micrometer-core/src/main/java/io/micrometer/core/instrument/Observer.java
Camelion: My bad, completely forgot about aggregation quantiles through all instances. Yes, `Observer` originally was planned as abstraction for different summaries meter without affecting current micrometer `Quantile` and `Histogram` design.

And later i want to experiment with storing large amount of measurements-per-step (possible off-heap) in one metric (e.g. not-aggregated timings for every HTTP request) for sends it to backend and use storage-side quantiles calculation. 
Think that i can do it with my own `Meter` and `Quantile` implementation and there are no need in common `Observer` abstraction. I will update this PR after my researches.

jkschneider: Could you elaborate a bit more on what you intend to send to the backend for storage-side quantiles calculation? Is there an approach fundamentally different from the histogram?

Camelion: No, thats approach similarly to histogram, but in histogram we always need choose between size of buckets, that would be collected and accuracy of every value (mean size of bucket and count of buckets). 

I want to create timer implementation that will collect every record pair(timestamp/value), and send it to storage with given measure step or by reaching some batch size, and later use this metrics with one of [quantile*/median*](https://clickhouse.yandex/docs/en/agg_functions/index.html#quantile-level-x) CH function.
Raw measurements more useful for different aggregations, but i need more performance tests for graphic calculation queries and insertion queries. Now i have tests only with ~500 records per second on small number of timers (around 400000 sent measurements per minute) from small amount of application. 
Different approaches can be used in different places and it will be useful to be able to choose between them.

jkschneider: FYI, Netflix did a study with t-digests with the same goal and couldn't get it to scale. Not sure which approach you are looking at.

jkschneider: I'm totally supportive of any experiments in this space. Being able to deliver a tight error bound on aggregable quantiles would be a huge win. Let me know how I can help.

brharrington: FWIW, [this thread](https://groups.google.com/forum/#!topic/netflix-atlas/iUtiMEGLZJo) has some details on what we are currently using at Netflix and the error compared to t-digests. The current histogram approach we use in spectator has worked fairly well for us over the last 2 years. The main draw back and where we are looking to improve it is to reduce the number of buckets needed, and thus are storage and query costs, to get acceptable accuracy.

jkschneider: Thanks @brharrington. I think you alluded to this before, and the explanation is fantastic. I see us doing away with both histograms and client side quantiles and adopting this approach then dropping down support for it on prometheus, et al.

PR #60, opened 2017-07-27
Pass injected parameter to Bean constructor

shakuzen:
Method injection is used for the `CollectorRegistry` dependency, but it is not passed to the constructor for `PrometheusMeterRegistry` as mentioned in the following comment: https://github.com/micrometer-metrics/micrometer/commit/45ebbb3144a0ec7864fba4503425a51900fe71dd#commitcomment-23304373


checketts:
Thanks!

Issue #69, opened 2017-08-01
Add static holder for injecting MeterRegistry


checketts:
Similar to how libraries can code against slf4j by using the LoggerFactory to get a logger, we need a similar way for libraries to get the MeterRegistry.

Instead of using classpath introspection, we can leverage the constructor usage. I'll send a PR shortly.

jkschneider:
```java
final Counter thingCounter = MeterFactory.counter("thing", "region", "test");
```

This could generate a no-op shim (or uses a global SimpleRegistry). 

In a Spring context, a `@BeanPostProcessor` could be used to replace the shim with a counter generated from the registry/registries wired in your `ApplicationContext`.

Frameworks that don't have an equivalent to `@BeanPostProcessor` could affect static state of the factory to change how the "shim" is generated?

jkschneider:
If we implement something like this, #37 probably goes away. You can do more in code than you can in an annotation for sure.

checketts:
Yes, that is exactly how I think it can work. I really like the `@BeanPostProcessor` idea too, that means we wouldn't recommend `static` fields and instead toward normal field variables.

I actually prefer that anyways since the static nature has caused me trouble in the past, and not due to shared state, but rather just the fact that I want a gauge to reference an instance field.

danveloper:
I can't speak for _all_ frameworks that do not have the `@BeanPostProcessor` equivalent, but in Ratpack, the `Execution` (which can-be/is a DI-backed registry) can be got at in a static context, provided the code path is executing on a Ratpack-managed thread.

brharrington:
In spectator we do this using [Spectator.globalRegistry](http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/Spectator.html#globalRegistry--). It returns a composite registry type so the DI provided registry would just get added to that composite if it is desirable to pick up meters from the global registry. This has worked fairly well for us though in rare cases there have been concerns about missing some activity before the DI initialization is complete and the registry is added to the global composite.

Not sure if it is a concern here, but in spectator we also make sure to avoid any use of logging in the composite registry so it is safe to use `Spectator.globalRegistry()` in log appender implementations assuming the logging configuration is setup prior to and independent of DI, e.g., using standard log4j properties configuration.

PR #71, opened 2017-08-03
Add initial implementation of static MeterFactory

checketts:
Here is the initial instrumentation. I had to comment out the Gauge one due to the fact that the registry can't return Gauges and instead only returns values.

I'm fine iterating in this PR, so it is by no means complete. I'll add some more samples with the various metric types next.

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/MeterRegistryConfigurationSupport.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/MeterRegistryConfigurationSupport.java
@@ -15,6 +15,7 @@
  */
 package io.micrometer.spring;
 
+import io.micrometer.core.api.MeterFactory;
 import io.micrometer.core.instrument.MeterRegistry;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Configuration;
@@ -41,6 +42,8 @@ public class MeterRegistryConfigurationSupport {
 
     @PostConstruct
     void bindAll() {
+        MeterFactory.Config.setRegistry(registry);
+
         // Important that this happens before binders are applied, as it
         // may involve adding common tags that should apply to metrics registered
         // in those binders.
--- a/micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/PrometheusSample.java
+++ b/micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/PrometheusSample.java
@@ -15,16 +15,38 @@
  */
 package io.micrometer.spring.samples;
 
+import io.micrometer.core.api.MeterFactory;
+import io.micrometer.core.api.MeterProvider;
+import io.micrometer.core.instrument.Counter;
 import io.micrometer.spring.export.prometheus.EnablePrometheusMetrics;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
 
 @SpringBootApplication(scanBasePackages = "io.micrometer.spring.samples.components")
 @EnablePrometheusMetrics
 @EnableScheduling
+@RestController
 public class PrometheusSample {
+    private static final Counter wayTooEarly = MeterFactory.counter("way_early_greetings");
+    private final Counter tooEarly = MeterFactory.counter("early_greetings");
+
+    @Autowired
+    MeterProvider meterProvider;
+
     public static void main(String[] args) {
         SpringApplication.run(PrometheusSample.class, args);
+    }
+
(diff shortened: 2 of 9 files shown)


jkschneider:
Good start. I think the main thing to address is how we deal with this static state?

micrometer-spring-legacy/src/main/java/io/micrometer/spring/MeterRegistryConfigurationSupport.java
```diff
@@ -41,6 +42,8 @@
 
     @PostConstruct
     void bindAll() {
+        MeterFactory.Config.setRegistry(registry);
```
checketts: Here is how it gets wired in. However, I suspect it won't be as simple as I had hoped. I fear if a class had a static field or instance variables, then they would have already called the factory and retrieved their meters from the SimpleMeterRegistry that had been in place beforehand.

I'll play with it to determine if that is true. If so, I can make the default MeterRegistry account for it, but it will add complexity.

micrometer-core/src/main/java/io/micrometer/core/instrument/prometheus/PrometheusMeterRegistry.java
```diff
@@ -179,7 +179,7 @@ public double get() {
     }
 
     @Override
-    public MeterRegistry register(Meter meter) {
+    public Meter meter(Meter meter) {
```
checketts: I renamed the `register(Meter meter)` method to match the other methods like `counter` and `timer` that return the Meter. So that affected all the registries.

micrometer-core/src/main/java/io/micrometer/core/api/MeterProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package io.micrometer.core.api;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Timer;
+
+public interface MeterProvider {
```
checketts: Here is the simplified MeterRegistry interface

micrometer-spring-legacy/src/main/java/io/micrometer/spring/MeterRegistryConfigurationSupport.java
checketts: I also couldn't find the equivalent `@PostConstruct` in the `spring` project.

micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/PrometheusSample.java
```diff
@@ -15,16 +15,38 @@
...
     }
-}
\ No newline at end of file
+
+    @RequestMapping("/hello")
+    public String hello() {
+        wayTooEarly.increment();
+        tooEarly.increment();
+        MeterFactory.counter("greetings").increment();
+        meterProvider.counter("injected_greetings").increment();
```
checketts: OK my fears had proven true. The static and instance counters (`wayTooEarly` and `tooEarly`) had been created before the true implementation was set in the `@PostConstruct`. I solved it by creating the `SwappableMeterRegistry` that iterates through its counters and points them to counters from the new registry when it is set.

If this approach seems good to you @jkschneider, I'll finish the remaining implementation and JavaDocs.

Also I intend to remove these from the `PrometheusSample` before merging in in an effort to keep them simple. However I did want to include some sort of sample to discuss around,

micrometer-spring-legacy/src/main/java/io/micrometer/spring/MeterRegistryConfigurationSupport.java
jkschneider: This still breaks test parallelism I think?

checketts: Yes and no. It won't break any existing tests unless the test is trying to validate a metric that is being set via the static `MeterFactory` approach. 

Tests that are relying on parallelism/isolation should be using a `MeterProvider` that is injected and guaranteed to match the current application context.

adriancole:
Meta comment: while it certainly is routine to have static metrics registries, it probably helps to add rationale about this. Ex when you absolutely need static, when injected is a better choice, etc (with examples). We have a chance to direct new users.

checketts:
@adriancole Agreed. We had a similar discussion in slack. I'll be updating it to make this available while making it clear that the injectable approach is the best option

jkschneider:
Many of the ideas here were incorporated into work on #69 and #81. Here is [the doc](https://docs.google.com/document/d/1nYclbtIA9KgTPnJoGVsXjofgWNGwk-cmaAIit_SDZ8w/edit) outlining the proposal to deal with library instrumentation. Summarizing:

* `MeterRegistry.globalRegistry` holds a static reference to a `CompositeRegistry`. You can add or remove registries from the composite at any time.
* Field-based meters are generated using one of the `Meters.lazyXXX` methods.

Issue #73, opened 2017-08-07
io.micrometer.core.instrument.Meters should not depend on the guava cache class


tokuhirom:
When I want to use the Meters class, I need to import google guava. I don't want to use the Google guava at some time because it doesn't care the backward compatibility well.
Other targets for the Meters class are JDK's core classes. I suggest splitting the guava support and other parts like ExecutorService and Execturor.

(Note: I heard Spring 5 will support the caffeine library as a primary cache library instead of guava cache)

checketts:
Guava is not intended to be a required dependency. Can you please include details on why it is being required or what error you see when it is not being required?

jkschneider:
Hopefully 6527f3a shows how the absence of optional dependencies like guava on the classpath does not preclude the use of other methods on `Meters`. Please let us know if you run into a specific issue that we haven't covered.

Also, thanks for the note on caffeine. Tracking micrometer support for caffeine in #76.

tokuhirom:
@checketts @jkschneider 

Please lok https://github.com/tokuhirom/micrometer-issue-73/tree/master .

I got this error with ↑ repo.

```
micrometer-issue-73/src/main/java/com/example/demo/MicrometerIssue73Application.java:26: error: cannot access Cache
            return Meters.monitor(registry,
                         ^
  class file for com.google.common.cache.Cache not found
1 error
```

Because `Meters.monitor()` method uses method overloading. A method uses method loading loads all classes required by every methods' arguments.

jkschneider:
Compilation failure -- that's what I didn't expect. Thanks for the detail @tokuhirom.

PR #78, opened 2017-08-08
Initial Caffeine cache metric support

checketts:
Note that I changed some of the approach from the Guava monitoring

* Cache name isn't a prefix, but rather a dimension now, this allows grouping and aggregating among the caches
* Adds Mockito as a test dependency (just checking if that is OK)
* Also neglected the `Meters.cache` helper method. Wanted to discuss tradeoffs of just using the constructor directly.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/CaffeineCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/CaffeineCacheMetrics.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2017 Pivotal Software, Inc.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


ben-manes:
Can you please special case for AsycLoadingCache?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/CaffeineCacheMetrics.java
```diff
@@ -0,0 +1,90 @@
...
+    List<Tag> tags = Tags.zip("cache", name);
+    registry.gauge("caffeine_cache_estimated_size", tags, cache, Cache::estimatedSize);
+
+    registry.counter("caffeine_cache_requests", Tags.zip("cache", name, "result", "miss"), cache, c -> c.stats().missCount());
+    registry.counter("caffeine_cache_requests", Tags.zip("cache", name, "result", "hit"), cache, c -> c.stats().hitCount());
+    registry.counter("caffeine_cache_requests_total", tags, cache, c -> c.stats().requestCount());
+    registry.counter("caffeine_cache_evictions_total", tags, cache, c -> c.stats().evictionCount());
+    registry.gauge("caffeine_cache_eviction_weight", tags, cache, c -> c.stats().evictionWeight());
+
+    if (cache instanceof LoadingCache) {
```
checketts: @ben-manes  Are you saying this line should be `cache instanceof LoadingCache || cache instanceof AsyncLoadingCache`  ?

ben-manes: No, I think you just need another constructor that sets the field to `cache.synchronous()` and the bindings should work.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/CaffeineCacheMetrics.java
```diff
@@ -0,0 +1,90 @@
...
+  *     caffeine_cache_load_failure_total{cache="mycache"} 2.0
+  *     caffeine_cache_loads_total{cache="mycache"} 7.0
+  *     caffeine_cache_load_duration_seconds_count{cache="mycache"} 7.0
+  *     caffeine_cache_load_duration_seconds_sum{cache="mycache"} 0.0034
+  * </pre>
+  *
+  * @author Clint Checketts
+ */
+public class CaffeineCacheMetrics implements MeterBinder {
+  private final String name;
```
jkschneider: Can you do IntelliJ default formatting (4 spaces)?

jkschneider:
> Also neglected the Meters.cache helper method. Wanted to discuss tradeoffs of just using the constructor directly.

In light of #78 I'd like to hear your thoughts on this. We may just have to drop the whole `Meters` class.

checketts:
@jkschneider I've reformatted to 4 spaces, and included an `.editorconfig` file to ensure that format going forward

Regarding #73 I think we should drop the `Meters` class. I like it for discoverability, but I think adding those to the docs site is probably a good enough solution.

jkschneider:
> we should drop the Meters class

I agree. We can add static methods to the `XXXMetrics` classes to do the binding to a registry and returning the instrumented thing like `Meters` did. Like you said, little loss of discoverability, but in light of the compilation issue, doesn't seem like there is a way forward with `Meters`.

checketts:
One last option is to use static inner classes. I don't think we should do it, but wanted to point out a potential option.

I think it would still do `Meters.Guava.cache()`

I don't think is really adds that much so I would recommend we get rid of `Meters`.

trustin:
I'm curious if it is worth providing `TimeUnit`-aware counter which is similar to `Timer`. `CacheStats.totalLoadTime()` returns cumulative nanoseconds but I believe Prometheus users will prefer the conversion to seconds.

trustin:
Do you think it's a good idea to add this sort of meta-information getters to `MeterRegistry`? e.g. `MeterRegistry.prefersConversionToBaseUnit()` If so, I can send a PR.

checketts:
> I'm curious if it is worth providing TimeUnit-aware counter which is similar to Timer

I think you are correct that it isn't a Timer aspect but a presentation/registry aspect. I think moving Registry config into its own object is needed instead of putting too many more on the Registry itself. (Working on PR for this very thing)

ben-manes:
Yeah, I'd expect this to be part of the reporter if a time unit wasn't supported by the data sink. Caffeine's stats are monotonically increasing to allow the presentation to do time series and other calculations. The conversion and other logic should be pushed into the presentation layer above the data sink if possible.

jkschneider:
I'm going to merge this as is for now and iterate on it a bit. Recently, I added a specialized counter for tracking monotonically increasing functions that, depending on the backend, either adds increments the counter by the change since the last interval or simply replaces a cumulative value. I'll do something similar for timers that allows you to track two monotonically increasing functions (duration and count) and that will also account for base unit normalization.

checketts:
@jkschneider What is the next step in regards to the `Meters` class? Want me to clean it up so the Guava binders match the Caffeine ones? Also any tips regarding that docs site I'm happy to help document this there.

ben-manes:
oh, Caffeine added `eviction_weight` metric in addition to Guava's `eviction_total`. That was by request from Druid when monitoring their cache churn (root issue was a bug in FJP causing it to drop tasks).

jkschneider:
@checketts I just deleted `Meters` locally, will push that in a bit.

checketts:
> `oh, Caffeine added eviction_weight metric in`
@ben-manes  Are you saying we should add that too?

ben-manes:
I think so

Issue #80, opened 2017-08-08
Binders should register metric names that follow different conventions


jkschneider:
Example from Slack on Aug 8:

> It seems like Prometheus convention tells us to 1)  include unit in the name and 2) append '_total' for cumulative values while Atlas does not.

cc / @trustin

trustin:
Thanks for creating an issue for this. I see different libraries have different conventions. To list a few:

- Dropwizard
  - Dot separator
  - Camel case
  - No unit suffix
  - No total suffix for cumulative value
  - No tags
  - No help
  - e.g. `caffeineCache.hits`
- Prometheus
  - Underscore separator
  - Snake case
  - Unit suffix
  - Total suffix for cumulative value
  - Tags
  - Help
  - e.g. `caffeine_cache_hits_total`
- Atlas
  - Dot separator
  - Camel case
  - No unit suffix
  - No total suffix for cumulative value
  - Tags
  - No help
  - e.g. `caffeineCache.hits`

Prometheus is interesting in that:

- It uses the same separator (underscore) for components and words, resulting in ambiguity. e.g. There's no way to tell the difference between `caffeineCache.hits` and `caffeine.cacheHits`.
- It requires to append unit suffix.
- It requires to append `_total` for cumulative meters
- It allows specifying 'help' string for a meter, which is a human-readable description of the meter.

trustin:
One option to solve this is to let a user provide enough information so that micrometer can generate the names that meet the convention:

- Specify the unit (bytes/nanoseconds, ..), cumulativeness and help string of a meter.
- Define a universal naming convention which can be translated into the naming convention of provider. e.g. `"caffeineCache.hits"` rather than `"caffine_cache_hits"` ?

trustin:
Here's my simple name generator:

https://github.com/trustin/armeria/blob/stats_micrometer/core/src/main/java/com/linecorp/armeria/common/metric/MeterRegistryUtil.java#L50

Example:

https://github.com/trustin/armeria/blob/stats_micrometer/core/src/main/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupport.java#L76

and test case:

https://github.com/trustin/armeria/blob/stats_micrometer/core/src/test/java/com/linecorp/armeria/internal/metric/CaffeineMetricSupportTest.java#L46

PR #89, opened 2017-08-10
POC of fluent interface for meter creation

checketts:
This solves the intermixing of overloaded methods in `MeterRegistry`. I haven't done that in this PR yet, since I wanted to get feedback first.

This change introduces `MeterIdBuilder`.

Example usage:

Create a meter with a name:
`registry.meter("my_counter").counter()`

Create a meter with a set of tags:
`registry.meter("my_tagged_counter").tag("level", "error").counter()`

Create a meter with lots of tags:
`registry.meter("counter_with_tags").tag("first","tag").tag("second","tag").counter()`
`registry.meter("counter_with_tags").tags(Tags.zip("first","tag","second","tag")).counter()`

It also lays the groundwork to address #80 since the `MeterIdBuilder` could be the location to process the name that is built as in: `.meter("tag", "that", "isformated").tag...`

It also means that we could remove the String var arg methods from `MeterRegistry`.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractMeterRegistry.java
@@ -24,12 +24,17 @@ import java.util.function.ToDoubleFunction;
 
 import static java.util.stream.StreamSupport.stream;
 
-public abstract class AbstractMeterRegistry implements MeterRegistry {
+public abstract class AbstractMeterRegistry implements MeterRegistry, MeterRegistryConfigurator {
     protected final Clock clock;
     protected final List<Tag> commonTags = new ArrayList<>();
 
     protected AbstractMeterRegistry(Clock clock) {
         this.clock = clock;
+    }
+
+    @Override
+    public MeterIdBuilder meter(String name) {
+        return new MeterIdBuilder(name, this);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


jkschneider:
There is already a similar concept for timers and distribution summaries, see `MeterRegistry.timerBuilder` and `MeterRegistry.summaryBuilder`

jkschneider:
I'm OK with adding more fluent builders, but I think it would be good to preserve the varargs. Spectator has a `Registry.createId`, which is somewhat similar to this. I just don't like having it as the only option, because it feels too verbose for the common case of a name and a fixed set of tags.

Personally I don't feel the same pressure to limit the number of overloads right now, but if we must, once a fluent API is in place for all the meter types, I'd ditch the overloads with `Iterable<Tag>` first.

checketts:
> I don't feel the same pressure to limit the number of overloads in MeterRegistry

The main reason I've brought it up right now is that once 1.0 is released (next month?!?) it seems the API will be 'frozen' and we won't be able to clean up the user experience.

I really want MicroMeter to be as straight forward as slf4j is. Most people will be fine with counters, gauges, and timers. Digging into distributions, histograms, and more can be a kind of secondary level. (SImilar to how slf4j you just have log levels, but advanced usage uses parameter markers or MDC)

In the Prometheus client repo it seems like weekly someone reports a problem and the dev's response of 'you are using the API wrong' seems like they are avoiding improving the API to ensure user success  (not saying that micrometer has the same problem)

checketts:
Will rebase against master...

brharrington:
> I'm OK with adding more fluent builders, but I think it would be good to preserve the varargs. Spectator has a Registry.createId (see this), which is somewhat similar to this.

FWIW the varargs reception has been mixed in Spectator. The overloads we added were due to calls for convenience and many users find the string varargs versions to be quite convenient. However, you need to call those methods with pairs or strings (key/value pairs), but there isn't a way to force that so unfortunately it can result in a runtime issue. The `withTag` calls, while more verbose, are also safer.

PR #93, opened 2017-08-15
Add library examples

checketts:
Library examples for 

https://docs.google.com/document/d/1nYclbtIA9KgTPnJoGVsXjofgWNGwk-cmaAIit_SDZ8w/edit?usp=sharing

### DIFF
--- a/micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/library/ProposalLibrary.java
+++ b/micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/library/ProposalLibrary.java
@@ -1,0 +1,17 @@
+package io.micrometer.spring.samples.library;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Meters;
+
+public class ProposalLibrary {
+
+    //@Inject no available on classpath
+    MeterRegistry registry = MeterRegistry.globalRegistry;
+    Counter workCounter = Meters.lazyCounter(() -> registry.counter("work_counter"));
+
+    public void doWork() {
+            workCounter.increment();
+    }
+
+}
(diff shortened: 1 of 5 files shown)


micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/library/ProposalLibrary.java
```diff
@@ -0,0 +1,17 @@
+package io.micrometer.spring.samples.library;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Meters;
+
+public class ProposalLibrary {
+
+    //@Inject no available on classpath
```
jkschneider: `javax.inject` is a single package jar with only one version ever released. If you are a library owner, you would add it to the compile classpath?

checketts: Sorry the comment was for why I hadn't included it in the code here, since `javax.inject` wasn't in the project right now.

As a library developer I wouldn't want to include it, unless it works as an optional dep so i can leave it as an annotation even if the downstream user isn't going to use it

micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/library/ProposalLibrary.java
```diff
@@ -0,0 +1,18 @@
...
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Meters;
+
+import javax.inject.Inject;
+
+public class ProposalLibrary {
+
+    @Inject MeterRegistry registry = MeterRegistry.globalRegistry;
```
rhart: There is a potential con here that users of my library need to know which of my components to bind with their DI framework.  Or, I have to provide a guice module or equivalent for each DI framework I intend to support.

jkschneider: Good point.

jkschneider:
Changed the recommendation to the static global composite based on these examples and comments in the doc. Thanks @rhart and @checketts!

Issue #106, opened 2017-08-31
JVM related metrics should be unambiguous


mweirauch:
I just stumped upon that the thread and class metrics are not prefixed with "jvm" as the others (memory, buffer, gc).

When somebody comes up and scrapes `procfs` for some process level metrics he would need to come up with a distinct name for these. (`jvm.threads.*`  would align nicely with `process.threads`)
Also the threads queried via `ThreadMetrics` are just the JVM managed threads. The process from the operating system view is actually running more threads. (`cat /proc/<pid>/status | grep Threads`)

I wanted to raise this "concern" and ask for comments before the names are nailed with a 1.0 release.
I can come up with a small PR if decided.

jkschneider:
Sounds like a good suggestion to me. Perhaps we should also rename `ThreadMetrics` to `JvmThreadMetrics` to parallel `JvmGcMetrics`, etc?

mweirauch:
Yes, for the class rename.
Regarding the `jvm.threads`, how about making them dimensional as they are all gauges?
```
# HELP jvm_threads  
# TYPE jvm_threads gauge
jvm_threads{id="daemon",} 28.0
jvm_threads{id="live",} 30.0
jvm_threads{id="peak",} 42.0
```
And for the `jvm.classes` just the prefix:
```
# HELP jvm_classes_loaded  
# TYPE jvm_classes_loaded gauge
jvm_classes_loaded 6848.0
# HELP jvm_classes_unloaded_total  
# TYPE jvm_classes_unloaded_total counter
jvm_classes_unloaded_total 2.0
```
From a grouping aspect I could imagine them to be dimensional as well (`jvm_classes{id=[current,loaded,unloaded]`}, but that would make the total loaded (currently not tracked) and total unloaded gauges.

checketts:
I recommend against the dimensions for the threads if they aren't agreggatable. The risk there is that users do `sum(jvm_threads) by (instance)` to see how many threads are being used. and you are multicounting a daemon thread (since it would have existed in all the ids.)

Essentially dimensions should represent parts of a whole.

mweirauch:
Ah, thanks for that argument! Gotcha.

mweirauch:
Can be closed as #107 is merged.

PR #107, opened 2017-09-01
prefix JVM thread and class metrics with "jvm" (#106)

mweirauch:


jkschneider:
Thanks again @mweirauch!

PR #108, opened 2017-09-03
add area tag (heap,nonheap) to JVM memory pools

mweirauch:
This way you can easily aggregate heap and non-heap memory usage without
having to know pool names assigned by the different garbage collector
implementations. (Especially useful when graphing different JVMs running
different GC implementations in a common "JVM" dashboard.)

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/JvmMemoryMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/JvmMemoryMetrics.java
@@ -22,8 +22,7 @@ import io.micrometer.core.instrument.Tags;
 import java.lang.management.BufferPoolMXBean;
 import java.lang.management.ManagementFactory;
 import java.lang.management.MemoryPoolMXBean;
-
-import static java.util.Collections.singletonList;
+import java.lang.management.MemoryType;
 
 /**
  * Record metrics that report utilization of various memory and buffer pools.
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/JvmMemoryMetrics.java
```diff
@@ -52,7 +51,8 @@ public void bindTo(MeterRegistry registry) {
         }
 
         for (MemoryPoolMXBean memoryPoolBean : ManagementFactory.getPlatformMXBeans(MemoryPoolMXBean.class)) {
-            Iterable<Tag> tags = Tags.zip("id", memoryPoolBean.getName());
+            String type = MemoryType.HEAP.equals(memoryPoolBean.getType()) ? "heap" : "non-heap";
```
checketts: Good change. What do you think about matching prometheus' treatment of the hyphen: https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/MemoryPoolsExports.java#L56

As in `nonheap` versus `non-heap`. I 'm OK deviating, just want to make sure it is considered.

mweirauch: Yeah, I sympathised with that. Just wen't for the JMX API naming. But from a perception and cannonical naming perspective, `area` is indeed better. Also `nonheap` is fine.
Will update the PR.

Generally, micrometers identifying tag name is `id` on purpose for a unified "name-the-thing" convention or just because it was chosen deliberately? (`jvm_memory_used{id="Par Eden", area="heap"}` vs `jvm_memory_used{pool="Par Eden", area="heap"}`) As I see it, all binders go with "id". Just asking.

mweirauch:
Updated (--force'd) the PR.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/JvmMemoryMetrics.java
jkschneider: @mweirauch There hasn't been a great amount of thought about it honestly -- trying to follow the already established names present in Spectator/Prometheus/elsewhere, depending on where we are drawing inspiration from.

jkschneider:
Thanks again!

Issue #66, opened 2017-07-31
Add AWS CloudWatch metrics support


mrcasablr:
AWS CloudWatch can take custom metrics and is supported by their Java SDK - http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/examples-cloudwatch-publish-custom-metrics.html

jkschneider:
https://github.com/spring-cloud/spring-cloud-aws/blob/2.0.x/spring-cloud-aws-actuator/src/main/java/org/springframework/cloud/aws/actuate/metrics/CloudWatchMetricWriter.java

dkublik:
Created PR: https://github.com/micrometer-metrics/micrometer/pull/250

jkschneider:
Added Boot binding in 3af71a2d2ba033baa8147e98d6ae400d4724d141.

PR #110, opened 2017-09-07
Add UptimeMetrics

mweirauch:
Wen't for a gauge instead of counter as rate calculations are "unlikely"
on an uptime counter. Recommended by @jkschneider.

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -36,6 +36,8 @@ dependencies {
     testCompile 'org.junit.jupiter:junit-jupiter-params:5.0.0-M5'
     testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.0.0-M5'
 
+	testCompile "org.mockito:mockito-core:2.+"
+
     testCompile 'javax.inject:javax.inject:1'
     testCompile 'org.springframework:spring-context:4.3.9.RELEASE'
     testCompile 'com.google.inject:guice:4.1.0'
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/UptimeMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/UptimeMetrics.java
@@ -1,0 +1,48 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder;
+
+import static java.util.Objects.requireNonNull;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+
+import io.micrometer.core.instrument.MeterRegistry;
+
+/**
+ * Uptime metrics.
+ *
+ * @author Michael Weirauch
+ */
+public class UptimeMetrics implements MeterBinder {
+
+    private final RuntimeMXBean runtimeMXBean;
+
+    public UptimeMetrics() {
+        this(ManagementFactory.getRuntimeMXBean());
+    }
+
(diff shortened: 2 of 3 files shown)


micrometer-core/build.gradle
```diff
@@ -36,6 +36,8 @@ dependencies {
     testCompile 'org.junit.jupiter:junit-jupiter-params:5.0.0-M5'
     testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.0.0-M5'
 
+	testCompile "org.mockito:mockito-core:2.+"
```
checketts: I would prefer not using dynamic versions here

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/UptimeMetrics.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    // VisibleForTesting
+    UptimeMetrics(RuntimeMXBean runtimeMXBean) {
+        this.runtimeMXBean = requireNonNull(runtimeMXBean);
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        registry.gauge("uptime", runtimeMXBean, (x) -> Long.valueOf(x.getUptime()).doubleValue());
```
checketts: I'm not opposed to this metric, but do want to point out the Prometheus approach is to report the process start time (see https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java#L72)

The logic being that the process start time won't change and the query can calculate the uptime based on that. This is also the typical approach for all timestamps with Prometheus, since you want the metric to report when something happened and the query can determine the 'time since' logic.

mweirauch: I was tempted to add "starttime" additionally in the beginning. The thing is only "starttime" is crucial to those math-agnostic metric-stores like CloudWatch.
Is adding both (uptime,starttime) in two MeterBinders an option, so people can choose what they wan't/need?

Regarding resolution/units: Should one drop precision just because Promtheus' time() returns unix-time (seconds) instead of millis? (micrometer != prometheus)

jkschneider: > Should one drop precision just because Promtheus' time() ...

The gauge value will be reported as a floating-point value, so no precision will be lost. I'm working on support for scaling time-based gauges now for this PR.

jkschneider: Maybe I shouldn't say _no_ precision is lost, because at a sufficient scale, the cumulative error of floating point math will add up. Apparently this is something Prometheus devs are willing to accept, and I'm happy to accommodate for as long as it doesn't cause that community any heartburn.

jkschneider: > Is adding both (uptime,starttime) in two MeterBinders an option

My vote would be to just add them both to the same binder. It's only one additional metric which can be ignored if not needed. Your statement that starttime is crucial to CloudWatch has me a bit alarmed: I _really_ need to look at CloudWatch before the 1.0 release (#66).

jkschneider:
Yikes, this PR gave me a lot to think about. There were some pretty extensive changes in 7a0bf86d47de0eb1dad1c2988408268244202360 that I think improves things overall:

1. There is now a `registry.more().timerGauge(..)` for this purpose. A few other binders have been changed to use it as well. This type scales the sampled value to the monitoring system's base time unit.
2. There is a new `registry.createId(..)` method. Every meter type can now be constructed with the result, e.g. `registry.counter(Meter.Id)`. I've thought about doing this for a while, and finally caved. The benefit is that we can support description, base units, etc on the more esoteric types without bloating the API too much. This is beneficial to types like function tracking counters, timer gauges, and the soon to be interval counters and function tracking timers.
3. All the meter builders have moved to their base types. What used to be `registry.timerBuilder(..).create()` is now `Timer.builder(..).register(registry)`.
4. Descriptions have been added to all binder provided metrics
5. The Prometheus registry now has a configuration option to flip descriptions on and off in case you get interested in saving a little transfer overhead on each scrape. I imagine some may want to enable descriptions on a per-profile basis -- keeping them on to do development to aid in building new queries and then turning them off in prod.

checketts:
> `Timer.builder(..).register(registry)`

That looks familiar 😄 

> There is a new registry.createId(..) method.

Cool! I'm eager to try it out. Thanks for all the great work on this.

jkschneider:
> That looks familiar 😄

Yep, I thought that too. Not a bad thing at all!

PR #112, opened 2017-09-10
Add missing dot

eddumelendez:
Property `influx.uri` is specified in the docs. Previously, property
`influxuri` was used. This commit adds the missing dot.


jkschneider:
Good catch, thanks!

PR #113, opened 2017-09-10
Add missing metadata to enable specific instrumentation

eddumelendez:
Currently, the following property `metrics.<provider>.enabled` validate
if the instrumentation should apply. This commit adds the missing
metadata which can be inspected by the IDE.

### DIFF
--- a/micrometer-spring-legacy/build.gradle
+++ b/micrometer-spring-legacy/build.gradle
@@ -69,3 +69,7 @@ dependencies {
     samplesCompile 'org.springframework.integration:spring-integration-ws'
     samplesCompile 'org.springframework.integration:spring-integration-xml'
 }
+
+license {
+    exclude "**/META-INF/*.json"
+}
--- a/micrometer-spring-legacy/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/micrometer-spring-legacy/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -1,0 +1,46 @@
+{
+  "properties": [
+    {
+      "name": "metrics.atlas.enabled",
+      "type": "java.lang.Boolean",
+      "description": "Enable metrics for Atlas.",
+      "defaultValue": true
+    },
+    {
+      "name": "metrics.datadog.enabled",
+      "type": "java.lang.Boolean",
+      "description": "Enable metrics for DataDog.",
+      "defaultValue": true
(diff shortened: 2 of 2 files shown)


jkschneider:
This metadata is generated automatically by `org.springframework.boot:spring-boot-configuration-processor` via an annotation processor during the build. However, the descriptions are missing, and it appears that the annotation processor adds descriptions based on javadoc comments it finds on fields. Currently we don't have any fields on many of our `@ConfigurationProperties` classes, so this could use improvement.

eddumelendez:
> This metadata is generated automatically by org.springframework.boot:spring-boot-configuration-processor via an annotation processor during the build.

True, but in this cases at least for `enabled` property is not necessary to have a field. This is very common at `spring-boot` project.

`CassandraRepositoriesAutoConfiguration` has a `@ConditionalOnProperty` for `spring.data.cassandra.repositories.enabled`, which is not a field in any class.

https://github.com/spring-projects/spring-boot/blob/46470b51969ef3bdac87c9112a978afb0e0b6999/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/cassandra/CassandraRepositoriesAutoConfiguration.java#L41

At `additional-spring-configuration-metadata.json`, `spring.data.cassandra.repositories.enabled` is added to provide the hints in `application.properties`

https://github.com/spring-projects/spring-boot/blob/46470b51969ef3bdac87c9112a978afb0e0b6999/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L77-L82

I think we are in the same case here.

jkschneider:
Ah, thanks for clarifying. Our case is a little different, because in additional to enablement, there are several other properties that vary by monitoring system. We have a `@ConfigurationProperties` for each monitoring system, e.g. `AtlasConfigurationProperties`.

eddumelendez:
My intention in this PR is just add for `enabled`. Regarding to the other properties sure it needs fields in order to be processed by  `org.springframework.boot:spring-boot-configuration-processor`.

@jkschneider should that work be covered in this PR? or can be handle in another?

jkschneider:
I think it would be best to address `enabled` and all the other properties simultaneously, since addressing the other properties will obviate the need for a manual `additional-spring-configuration-metadata.json`.

jkschneider:
Related to #10292

Issue #91, opened 2017-08-12
HikariCP binder


tokuhirom:
I want to get metrics data from HikariCP using micrometers.

jkschneider:
For implementation, see the hook for adding a Dropwizard registry [here](https://github.com/brettwooldridge/HikariCP/wiki/Dropwizard-Metrics)

shakuzen:
Note: HikariCP has already added some initial support for Micrometer in HikariCP 2.7.0, as can be seen in [this package](https://github.com/brettwooldridge/HikariCP/tree/dev/src/main/java/com/zaxxer/hikari/metrics/micrometer). Does this cover what this issue was looking for?

jkschneider:
Updated API in https://github.com/brettwooldridge/HikariCP/pull/1002.

checketts:
The Hikari PR is merged can we close this @jkschneider ?

jkschneider:
I'm pretty sure we broke it with rc3 because it had references to quantiles. We should eliminate quantiles config in hikari altogether and allow users to configure them after-the-fact with the new meter filter props support.

jkschneider:
Hikari is on rc6 as of right now. After 1.0, we can circle back with a PR that brings Hikari up to 1.0 GA and removes client-side percentiles.

snicoll:
@jkschneider I can see you dropped the 1.0.0 milestone on this one. Does that mean we shouldn't provide auto-config for this? See the PR @shakuzen submitted to Spring Boot.

shakuzen:
From Slack:

> Tommy Ludwig [12:08 PM]
@jon It would be nice to have the HikariCP binder supported and configured in Spring Boot 2 GA. Is that the plan?

> jon [12:08 PM]
i believe it is?
i had a note on issue #91 because we should do a follow-up PR on hikari to remove its specification of client-side percentiles (edited)
ouch, you’re right. it’s not in boot 2 right now
when i do the 1.0 GA release tomorrow i’ll do a quick PR to Hikari and see if we can get a release cut before Boot 2 GA next week

I just tested that the current HikariCP metrics tracker works with the latest snapshots of Micrometer. As far as I understand, @jkschneider just wants to follow up after Micrometer GA is released to bump the version HikariCP depends on and to remove the currently declared client-side percentiles (since now percentiles can be configured via metrics filters in a user's application rather than needing to be predefined in a library like Hikari).

[...]

PR #131, opened 2017-09-19
Fixed missing active.connections metric for HikariCP

gavlyukovskiy:
While binding data source metrics HikariCP may not be initialized, if `DataSource#getConnection()` is not called before binding. 
In this case `HikariDataSourcePoolMetadata` returns null because `HikariDataSource#pool` is null and gauge is not even created.

Also I don't think storing pool metadata in static list is not a good way to prevent it being garbage collected, so I stored `dataSource` inside gauge and changed function to use it's `poolMetadata`, so that this metric depends on `dataSource` liveness.

As a workarround `dataSource.getConnection().close()` may be used before binding metrics.

### DIFF
--- a/micrometer-spring-legacy/build.gradle
+++ b/micrometer-spring-legacy/build.gradle
@@ -53,6 +53,7 @@ dependencies {
     // for testing DataSource metrics
     testCompile 'org.hsqldb:hsqldb'
     testCompile 'org.springframework.boot:spring-boot-starter-jdbc'
+    testCompile 'com.zaxxer:HikariCP'
 
     // for use in mock servers
     testCompile 'org.apache.tomcat.embed:tomcat-embed-core:8.5.15'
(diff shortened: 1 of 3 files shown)


jkschneider:
Cool, there is an open issue on HikariCP support at #91. Is there anything else about HikariCP that you think we should instrument?

gavlyukovskiy:
I don't think you can instrument anything else without proxying `DataSource`, 
I planned to add some with micrometer into [my library](https://github.com/gavlyukovskiy/spring-boot-data-source-decorator) it uses some proxy provider (such as [p6spy](https://github.com/p6spy/p6spy), [datasource-proxy](https://github.com/ttddyy/datasource-proxy) or [flexy-pool](https://github.com/vladmihalcea/flexy-pool)), With first two I can provide additional metrics like time to acquire, query time, connection live time, etc. Flexy Pool by default exposes some metrics to Dropwiazrd/Codahale.

jkschneider:
> With first two I can provide additional metrics like time to acquire, query time, connection live time, etc.
@gavlyukovskiy Really look forward to seeing this. Drop a note in Slack if you need any help.

PR #133, opened 2017-09-20
Fix executor correctly instrumented

kojilin:
Apply this change will make executor's execute is called.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/ExecutorServiceMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/ExecutorServiceMetricsTest.java
@@ -16,12 +16,14 @@
 package io.micrometer.core.instrument.binder;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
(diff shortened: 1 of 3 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/ExecutorServiceMetricsTest.java
```diff
@@ -38,6 +40,28 @@ void before() {
...
+            r.run();
+            lock.countDown();
+        };
+        Executor executor = ExecutorServiceMetrics.monitor(registry, exec, "exec");
+        executor.execute(() -> System.out.println("hello"));
+        lock.await();
+        assertThat(registry.find("exec").timer()).map(Timer::count).hasValue(1L);
+    }
+
+    @DisplayName("thread pool executor can be instrumented even treat as Executor")
```
checketts: What does `even treat` mean?

kojilin: ah, assign to Executor, so overloading with Executor's one

jkschneider:
Good catch! Merged to master.

PR #137, opened 2017-09-21
Document the requirement for the query parameter to return a single row and first column is the metric

checketts:


jkschneider:
Thanks for the diligence on this!

checketts:
Been feeling the itch since I hadn't contributed any code recently... (hope to rectify this soon)

PR #141, opened 2017-09-30
Add support for file descriptor metrics

### LINKED ISSUE #130: Add support for file descriptor metrics
Retrieved from `OperatingSystemMXBean` respectively `UnixOperatingSystemMXBean`.

Naming proposal:
* `process.fds.open`
* `process.fds.max`

Inspiration:
* https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java#L75
* https://github.com/mweirauch/metrics-jvm-extras/blob/master/src/main/java/io/github/mweirauch/metrics/jvm/extras/FileDescriptorCountGauge.java

mweirauch:
This PR only targets the HotSpot JVM. If time permits I am going to look at OpenJ9 later.

Fixes #130.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetrics.java
@@ -1,0 +1,90 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.system;
+
+import static java.util.Collections.emptyList;
+import static java.util.Objects.requireNonNull;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Objects;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+/**
+ * File descriptor metrics.
+ *
+ * @author Michael Weirauch
+ */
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetrics.java
```diff
@@ -0,0 +1,90 @@
...
+        this.tags = tags;
+
+        this.openFdsMethod = detectMethod("getOpenFileDescriptorCount");
+        this.maxFdsMethod = detectMethod("getMaxFileDescriptorCount");
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        registry.gauge(
+                registry.createId("process.fds.open", tags, "The open file descriptor count"), this,
```
jkschneider: It is possible if the binder will get garbage collected (e.g. instantiated in the body of a `@PostConstruct`), which would cause the gauges to drop to 0. Any way we can determine if the `OperatingSystemMXBean` ever gets collected, and if not, base the gauge on that instead?

mweirauch: I see and just tried: "NaN". The thing is, I wanted to safe on the method reflection being performed every time the Meter is queried. Or would you dare to call this premature optimization?
I can drop that for the moment.

But nevertheless, is there some solution you could envision?
A static holder for the methods perhaps? (This would render the class usable only once, but it is unlikely you'd instantiate it twice with different `OperatingSystemMXBean`s anyway.)

jkschneider: Looking at the implementation of `ManagementFactory`, the MBean is held in a static field anyway, so I'd just write `osBean` rather than `this`.

mweirauch: Yap, the thing is the public API (`java.lang.management.OperatingSytemMXBean`) does not expose the queried methods. It's the specific JVM implementation (e.g. HotSpot) providing them via the `com.sun.management.UnixOperatingSystemMXBean`. Other JVM implementations may or may not provide them, or name them differently.
That's why I reflect the methods in question once and keep them in order to not reflect on them on each call to value(). But, as that is complicating it a bit, I am going to rework it the easy way for now - reflecting on each value fetch of the gauge.

mweirauch: Ok, I should have tried before commenting... If it is ok, that the Binder is not GCed then the most recent update should do.

jkschneider: I'm not too concerned with reflecting on each value fetch of the gauge, as that only happens once per sampling interval (default 1m on most implementations).

jkschneider:
Thanks for this feature @mweirauch. Really awesome stuff.

Issue #70, opened 2017-08-01
Add ability to mute metrics that are being provided that are not needed and reducing cardinality


checketts:
Spectator has a concept of [PlaceHolders](https://github.com/Netflix/spectator/blob/master/docs/ext/placeholders.md) for reducing cardinality. 

We need a similar ability but have it contained in the MeterRegistry so that any metric being recorded can be influenced

jkschneider:
Partially related to #17

checketts:
# Design ideas

## Spring Config

`OFF` Ability to disable metric completely. This will ensure the metric is never registered/exposed. (Out of scope: removing metrics if config was live updated)
```
spring.metrics.http_requests: OFF
```

`MINIMAL` Disables all dimensions/tags (perhaps alias as `MERGE_DIMENSIONS` for clarity?)
```
spring.metrics.http_requests: MINIMAL
```

`FULL` Default value for metrics that aren't set. No filtering/will occur
```
spring.metrics.http_requests: FULL
```

I also want to include a `FILTER(some filtering expression)` option eventually. I think it is out of scope for this for now. This is similar to `slf4j`'s Janino integration. 

In logging, they are hierarchical by definition. So you can enable debugging for all `org.springframework` very easily.  I believe however we solve #80 could include some sort of hierarchy. But that will also be out of scope for this issue.


## Usage proposal:

Expose boolean check for if a meter should be used. This is more needed if the user wants to avoid doing work in calculating a metric. (Overloading #89 additions)
```
registry.meter("meter_name").isEnabled()

\\Alternative
registry.counter("meter_name").isEnabled()
```

Other usage won't require code changes by users. However adding filters will be allowed (and how the initial 'meter levels' will be implemented):
```
registry.addFilter(new MeterLevelFilter());
```

`MeterFilter` interface:
```
Filter.Result filter(String meterName, Iterable<Tag> tags);
```

I'll likely update this based on how it feels when I actually code it up. Any design feedback welcome.

jkschneider:
I looked at this a bit today, and was considering an API very similar to Logback's TurboFilter:

```
Counter c = registry.counter("my.counter");
c.filter(counterState -> MeterFilter.Reply.NEUTRAL); // <1>
c.isEnabled(); // <2>
```
1. Reply with `ACCEPT`, `NEUTRAL`, or `DECLINE`
[...]

Issue #17, opened 2017-05-23
Tiered metrics support


jkschneider:
Support tiering metrics by priority to allow for different push intervals (the relatively fewer critical metrics are published at a shorter interval than less critical metrics).

jkschneider:
In a moment of exceptional clarity, @nebhale tied this conceptually to reactive backpressure. Some metrics backends may be capable of exerting backpressure on the exporter (e.g. via a property, response headers).

jkschneider:
Spectator has a [Placeholders](http://netflix.github.io/spectator/en/latest/ext/placeholders/) concept that basically splits metrics into either detailed or not. Then something like a property can flip off detailed dimension publishing.

PR #179, opened 2017-10-27
This adds a FilterRegistry that that you can wrap another registry with.

checketts:
This addresses #70 and #17

I would love feedback on the combine syntax. 

I didn't add in the Spring config properties yet. But here is the idea. In one of my projects we are pushing to 3 registries, and so will have 3 different config sections.
```
spring.micrometer.metrics:
  defaultCombinedBucketName: other
  filter:
    test.filter: exclude
  combine:
    test.combine: tagKey|val1,val3;tagKey2
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/filter/FilterProperties.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/filter/FilterProperties.java
@@ -1,0 +1,116 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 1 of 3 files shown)


checketts:
I based this off the hystrix branch so I could use the `firstValue` for test assertions. I can rebase against master once I extract that helper.

micrometer-core/src/main/java/io/micrometer/core/instrument/filter/FilterProperties.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public Map<String, String> getFilter() {
+        return filter;
+    }
+
+    public Map<String, String> getCombine() {
+        return combine;
+    }
+
+    public String findMostSpecificRule(String name, Map<String,String> map, String defaultVal) {
```
checketts: Current implementation only finds the most specific rule. I can add in wildcard support and multiple rules in a subsequent PR.

Issue #156, opened 2017-10-11
Make `HierarchicalNameMapper` customizable for Etsy-flavored Statsd


jkschneider:
Also we may want to provide a generator for a `HierarchicalNameMapper` that turns a defined set of tags into a prefix in the generated metric name. In this way, we accomplish name prefixing on Etsy-flavored StatsD while still maintaining the prefix components as tags for maximum portability to other monitoring systems.

cc / @seonmila

dheld-expedia:
Yes, please!!!

PR #185, opened 2017-11-02
Added `HierarchicalNameMapper` as a dependency of `StatsdMeterRegistry`

dheld-expedia:
* Defaults to `HierarchicalNameMapper.DEFAULT` if none is provided

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,3 +48,5 @@ subprojects {
 task wrapper(type: Wrapper) {
     gradleVersion = '4.0.1'
 }
+
+defaultTasks 'build'
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdLineBuilder.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdLineBuilder.java
@@ -20,7 +20,6 @@ import io.micrometer.core.instrument.NamingConvention;
 import io.micrometer.core.instrument.Statistic;
 import io.micrometer.core.instrument.util.HierarchicalNameMapper;
 
-import java.beans.Introspector;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.text.NumberFormat;
@@ -28,15 +27,15 @@ import java.util.Locale;
 import java.util.Objects;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static io.micrometer.statsd.internal.MemoizingSupplier.memoize;
 import static java.beans.Introspector.decapitalize;
-import static java.util.stream.Stream.*;
+import static java.util.stream.Stream.of;
 
 class StatsdLineBuilder {
     private final Meter.Id id;
     private final StatsdFlavor flavor;
+    private final HierarchicalNameMapper nameMapper;
     private final NamingConvention convention;
 
     private final Supplier<String> datadogTagString;
@@ -61,9 +60,10 @@ class StatsdLineBuilder {
         return numberFormatter;
     });
 
-    StatsdLineBuilder(Meter.Id id, StatsdFlavor flavor, NamingConvention convention) {
+    StatsdLineBuilder(Meter.Id id, StatsdFlavor flavor, HierarchicalNameMapper nameMapper, NamingConvention convention) {
         this.id = id;
         this.flavor = flavor;
+        this.nameMapper = nameMapper;
         this.convention = convention;
 
         // service:payroll,region:us-west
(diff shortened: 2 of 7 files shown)


implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -15,8 +15,19 @@
  */
 package io.micrometer.statsd;
 
-import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.Clock;
```
dheld-expedia: My IntelliJ settings do this automatically, but if you prefer the wildcards, I will undo this.

dheld-expedia:
I believe this is the minimum change possible to allow custom configuration of a HierarchicalNameMapper for the StatsD back-end.

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
spencergibb: We generally prefer non `*` imports.

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdLineBuilder.java
```diff
@@ -61,9 +60,10 @@
         return numberFormatter;
     });
 
-    StatsdLineBuilder(Meter.Id id, StatsdFlavor flavor, NamingConvention convention) {
+    StatsdLineBuilder(Meter.Id id, StatsdFlavor flavor, HierarchicalNameMapper nameMapper, NamingConvention convention) {
```
spencergibb: Maybe overload the constructor with `HierarchicalNameMapper.DEFAULT` as the default.

dheld-expedia: Yeah...I thought that `build` was the default task for the project, so I didn't actually run the tests before pushing. :/

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
jkschneider: Ah, not in Micrometer @spencergibb. It's the IntelliJ default, and they are everywhere ;)

dheld-expedia:
I believe this implements https://github.com/micrometer-metrics/micrometer/issues/156, but I'm not sure what the intended solution there was.

build.gradle
```diff
@@ -48,3 +48,5 @@ subprojects {
 task wrapper(type: Wrapper) {
     gradleVersion = '4.0.1'
 }
+
+defaultTasks 'build'
```
checketts: Was this change intentionally checked in?

dheld-expedia: Yup.  It will make Gradle do the thing which most devs will probably expect.  Do you think it's bad?

checketts: I wasn't familiar with that feature.

I think `'clean', 'build'` might be a better default

dheld-expedia:
That is true if the value is in [0.0, 1.0].  However, the value I get is -1.0, because the `java.lang.management` version of the bean does not implement this call on Windows.

http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6336608

To prove it, I ran this little test, which printed -1.0 for as long as I was willing to wait:
```
        final OperatingSystemMXBean operatingSystemBean = ManagementFactory.getOperatingSystemMXBean();
        while (operatingSystemBean.getSystemLoadAverage() <= 0) {
            System.out.println("Waiting for load average: " + operatingSystemBean.getSystemLoadAverage());
            sleep(1000);
        }
```

dheld-expedia:
If you are willing to provide a Sun JVM-specific solution for Windows, there is this option: https://docs.oracle.com/javase/7/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getSystemCpuLoad().  I guess you could try to load the class and only use it if it's present.

build.gradle
dheld-expedia: Well, in a regular dev cycle, you will want to call `build` many times, but `clean` only rarely.  I'd rather not pay the cost of `clean` every time, and having to specify `build` every time defeats the purpose of making it the default!

checketts: Sounds good

jkschneider:
@dheld-expedia Wow, thanks for the research. I had no idea.

PR #186, opened 2017-11-02
Adjust metric name for "system load average"

### LINKED ISSUE #182: Metric "cpu.load.average" unfortunate naming
Everybody mistakes this as the CPU load average, when it basically is the system load average. Rightly so the metric is coming from the better named `OperatingSystemMXBean::getSystemLoadAverage`.

There is a nice post by Brendan Gregg detailing some archeological digging on when that metric became something like "system load average": http://www.brendangregg.com/blog/2017-08-08/linux-load-averages.html
The post mentions, that for other OSes the value might really mean only "cpu load average", though.

Is the naming on purpose or hasn't any deeper thought put on it?
Also, should it be made clear it's the "1m" load average? (at least on my Linux based system)

mweirauch:
Fixes #182.


jkschneider:
Thanks!

PR #188, opened 2017-11-03
Replace HandlerInterceptor with Filter

jkschneider:
* Allows us to time requests that are blocked with a 401 by spring-security (or redirected to a login page with a 302) #77.
* Short-circuit the URI tag for certain types of response codes (302 -> "REDIRECTION", 404 -> "NOT_FOUND") to prevent dimensional explosion on these type of requests.
* Don't prematurely time async requests.


micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/servlet/MetricsFilter.java
```diff
@@ -0,0 +1,84 @@
...
+        HandlerExecutionChain handler;
+        try {
+            MatchableHandlerMapping matchableHandlerMapping = mappingIntrospector.getMatchableHandlerMapping(request);
+            handler = matchableHandlerMapping.getHandler(request);
+        } catch (Exception e) {
+            logger.debug("Unable to time request", e);
+            return;
+        }
+
+        if (handler != null) {
```
checketts: Should we add `filterChain.doFilter(request, response);` to the else so the filter will let non-handler traffic through?

jkschneider: Good catch, done with dad918a

micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/servlet/MetricsFilter.java
```diff
@@ -80,5 +80,8 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
                 }
             }
         }
+        else {
+            filterChain.doFilter(request, response);
```
checketts: Do we need to worry about the `if` on line `68` and add a similar `else`?

PR #191, opened 2017-11-03
Added support for `ThreadPoolTaskExecutor` metrics

dheld-expedia:
* Upgraded spring-legacy to JUnit 5 for consistency

### DIFF
--- a/micrometer-spring-legacy/build.gradle
+++ b/micrometer-spring-legacy/build.gradle
@@ -48,9 +48,12 @@ dependencies {
 
     testCompile 'org.springframework.boot:spring-boot-starter-web'
     testCompile 'org.springframework.boot:spring-boot-actuator'
-
-    testCompile 'junit:junit:4.+'
     testCompile 'org.springframework.boot:spring-boot-starter-test'
+
+    // JUnit 5
+    testCompile 'org.junit.jupiter:junit-jupiter-api:5.0.0'
+    testCompile 'org.junit.jupiter:junit-jupiter-params:5.0.0'
+    testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.0.0'
 
     // for testing DataSource metrics
     testCompile 'org.hsqldb:hsqldb'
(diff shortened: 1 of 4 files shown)


micrometer-spring-legacy/src/main/java/io/micrometer/spring/async/TimedThreadPoolTaskExecutor.java
```diff
@@ -0,0 +1,81 @@
...
+        this.name = name;
+        this.tags = tags;
+        this.timer = registry.timer(name, tags);
+    }
+
+    @Override
+    public void initialize()
+    {
+        super.initialize();
+        new ThreadPoolTaskExecutorMetrics(this, name, tags).bindTo(registry);
```
dheld-expedia: You can't call `getThreadPoolExecutor()` until it's been initialized, and I need to do that to build the monitors.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/async/TimedThreadPoolTaskExecutor.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
+
+/**
+ * A {@link ThreadPoolTaskExecutor} which is timed
+ *
+ * @author David Held
+ */
+public class TimedThreadPoolTaskExecutor extends ThreadPoolTaskExecutor {
```
dheld-expedia: This doesn't support every kind of `AsyncTaskExecutor`, but that would be weaker anyway (because there is no guarantee that top-level metrics like "completed" are available).  Most folks are likely using a `ThreadPoolTaskExecutor`, and that maps most closely to the other thread pools supported by Micrometer.

micrometer-spring-legacy/build.gradle
```diff
@@ -48,10 +48,13 @@ dependencies {
 
     testCompile 'org.springframework.boot:spring-boot-starter-web'
     testCompile 'org.springframework.boot:spring-boot-actuator'
-
-    testCompile 'junit:junit:4.+'
     testCompile 'org.springframework.boot:spring-boot-starter-test'
 
+    // JUnit 5
```
jkschneider: Many of the test facilities in Spring Boot 1.5.x assume JUnit 4, which is why all the tests in this module are in JUnit 4 (e.g. `@RunWith(SpringRunner.class)` doesn't have a JUnit 5 corollary in Boot 1.5.x.

jkschneider:
Merged with 1175216817036377f25e140a146a8e815448d7e3

mweirauch:
Might wan't to add to rc.4 milestone?

PR #192, opened 2017-11-03
Add initial MeterFilterConfigProperties

checketts:
Here is an example config that we could check in and others could extend if they want. I left off the weird combining syntax since that likely won't be as generally useful

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/MeterFilterConfigPropertiesTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/MeterFilterConfigPropertiesTest.java
@@ -1,0 +1,57 @@
+package io.micrometer.core.instrument;
+
+
+import io.micrometer.core.instrument.histogram.HistogramConfig;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+public class MeterFilterConfigPropertiesTest {
+
+    private MeterFilterConfigProperties sut;
+
+    @BeforeEach
+    public void setup(){
+        sut = new MeterFilterConfigProperties();
+        sut.getFilter().put("ALL.enabled","false");
+        sut.getFilter().put("ALL.percentiles","0.3,0.4");
+        sut.getFilter().put("ALL.percentileHistogram","false");
+        sut.getFilter().put("ALL.maximumExpectedValue","PT7S");
+        sut.getFilter().put("ALL.minimumExpectedValue","10000000");
+        sut.getFilter().put("ALL.histogramBufferLength","100");
+        sut.getFilter().put("ALL.durationExpiry","PT60S");
+        sut.getFilter().put("ALL.sla","PT0.1S,PT1S");
+        sut.getFilter().put("jvm.enabled","true");
+        sut.getFilter().put("jvm.my.daemon.enabled","false");
+        sut.getFilter().put("jvm.memory.used","enabled");
+        sut.getFilter().put("jvm.combineTags","tag1|a,b;tag2");
+        sut.getFilter().put("jvm.some.timer.percentiles","0.95,0.99");
+    }
+
+    @Test
+    public void accept() {
+        assertThat(sut.accept(newMeter("my.counter"))).isEqualTo(MeterFilterReply.DENY);
+        assertThat(sut.accept(newMeter("jvm.thing"))).isEqualTo(MeterFilterReply.ACCEPT);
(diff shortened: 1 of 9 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/MeterFilterConfigPropertiesTest.java
```diff
@@ -0,0 +1,57 @@
...
+        sut.getFilter().put("ALL.percentiles","0.3,0.4");
+        sut.getFilter().put("ALL.percentileHistogram","false");
+        sut.getFilter().put("ALL.maximumExpectedValue","PT7S");
+        sut.getFilter().put("ALL.minimumExpectedValue","10000000");
+        sut.getFilter().put("ALL.histogramBufferLength","100");
+        sut.getFilter().put("ALL.durationExpiry","PT60S");
+        sut.getFilter().put("ALL.sla","PT0.1S,PT1S");
+        sut.getFilter().put("jvm.enabled","true");
+        sut.getFilter().put("jvm.my.daemon.enabled","false");
+        sut.getFilter().put("jvm.memory.used","enabled");
```
checketts: Implementation details of interest:

* I didn't support wildcards, since I discussed with some team mates and we couldn't find a very good usecase
* The only special value is `ALL`, so you have a 'root' option that applies to all metrics
* Keys are case insensitive
* Made it so `jvm.memory.used.enabled: true` and `jvm.memory.used: enabled` both work since the second one seems to be a common usecase (and read much cleaner when flattened with many other config values)

jkschneider: I think wildcard will be useful to turn off or configure metrics emanating from a system that has a common prefix (e.g. `hikari.foo`, `hikari.bar`)

jkschneider: Boy, you're right. The PTxS syntax looks ugly in the context of SLAs.

checketts: Hierarchical config is supported `hikari.enabled: false` works, `hikari.*.thing.enabled: false` won't.

I'm very tempted to code a simple parser that only supports a number and a unit. (100ms, 10s).

micrometer-core/src/test/java/io/micrometer/core/instrument/util/TimeUtilsTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+class TimeUtilsTest {
+    @Test
+    void simpleParse() {
+        assertThat(TimeUtils.simpleParse("5ns")).isEqualByComparingTo(Duration.ofNanos(5));
+        assertThat(TimeUtils.simpleParse("700ms")).isEqualByComparingTo(Duration.ofMillis(700));
+        assertThat(TimeUtils.simpleParse("1s")).isEqualByComparingTo(Duration.ofSeconds(1));
+        assertThat(TimeUtils.simpleParse("10m")).isEqualByComparingTo(Duration.ofMinutes(10));
+        assertThat(TimeUtils.simpleParse("13h")).isEqualByComparingTo(Duration.ofHours(13));
+        assertThat(TimeUtils.simpleParse("5d")).isEqualByComparingTo(Duration.ofDays(5));
```
checketts: Here is the simple parsing.

micrometer-spring-legacy/src/samples/resources/application-prometheus.yml
```diff
@@ -14,4 +14,12 @@
...
-spring.metrics.prometheus.enabled: true
\ No newline at end of file
+management.security.enabled: false
+
+spring.metrics:
+  prometheus.enabled: true
+  overrides:
+    ALL.enabled: false
+    jvm.enabled: true
+    jvm.memory.used: disabled
```
checketts: @jkschneider I'm liking the value `overrides` to represent the configuration. You can see the example above.

I also added in config validation/error propagation. Let me know if there are any other needs.

jkschneider:
Merged with d2beebf8ba7203638c7310a013acc2fa834a953c.

checketts:
The config validation is gone. Are we OK with that?

jkschneider:
Type coercability validation is done by `ConversionService`. Anything else needed?

checketts:
Mainly an error message lof `ConfigRule 'system.cpu.enabled' has an invalid value of 'flase'` or something to help users locate bad config

jkschneider:
Spring's conversion service presents an error message like `Caused by: java.lang.IllegalArgumentException: Invalid boolean value 'flase'`. I don't think we need any more custom logic than that?

Issue #180, opened 2017-10-27
Expose Hibernate statistics


snicoll:
We had a [PR to expose Hibernate statistics](https://github.com/spring-projects/spring-boot/pull/3596) in Spring Boot but the move to micrometer made it irrelevant. Perhaps we would consider adding this to micrometer?

cc @mdeinum

mdeinum:
I might have some time on my hands to migrate the original PR to work with micrometer. But I need some help getting started. @jkschneider could you give me some pointers?

checketts:
Pointers:

* Have `HibernatePublicMetrics` implement `MeterBinder`
* Each of your `addMetrics` lines, instead of returning a result, will become a `Gauge` that return a single measurement.

For example:
```
addMetric(metrics, prefix + "sessions.open", stats.getSessionOpenCount());
```
Would be something like
```
Gauge.builder("hibernate.sessions.open", stats, stats -> stats.getSessionOpenCount())
  .tags("entityManagerFactory", emfName)
  .register(registry);
```
Check out the [GuavaCacheMetrics](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/GuavaCacheMetrics.java) for comparison

mdeinum:
Pull request created. I'm of course open to comments :). 

I had doubts about using a `Counter` or a `Gauge` and we also have some execution times in there should that be a `TimeGauge` or ... ?

PR #195, opened 2017-11-06
When Hibernate is on the classpath and an EntityManagerFactory is

mdeinum:
configured then you can register a `HibernateMetrics `bean so that the
statistics from hibernate get collected.

The exposed statistics are the same as those that would be exposed using
`Statistics.logSummary` from hibernate itself.

Issue: gh-180

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -20,6 +20,7 @@ dependencies {
     compile 'net.sf.ehcache:ehcache:2.10.3', optional
     compile 'javax.cache:cache-api:1.0.0', optional
     compile 'com.hazelcast:hazelcast:3.8.+', optional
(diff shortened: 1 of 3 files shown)


jkschneider:
Merged with 601a3636569daef2993a27571107e9fb5f3add7b. 

Like you suspected, many of the gauges were changed to `FunctionCounter`s. The value of this counter is that it takes a global or accumulated value from hibernate and breaks it down into the accumulations that occurred over a particular interval (often the publish interval of a registry implementation). This allows you to set alerts/take action on changes over intervals.

The two max stats coming out of Hibernate were changed to `TimeGauge`s in order that the millisecond values reported by Hibernate are able to be scaled to the base time unit of each monitoring system.

Thanks for putting this together @mdeinum!

mdeinum:
Thanks for the update (those were actually the changes I was preparing as well :) ).

PR #197, opened 2017-11-07
NoOpTimer fails to call `record(Runnable)`

checketts:
This manifests in unit tests because CompositeTimer is empty.


checketts:
I don't suppose we could get an `rc4` to address this regression?

jkschneider:
Is it important right now or can we do an rc4 on Friday?

checketts:
It can wait til Friday.

Issue #200, opened 2017-11-09
@Beans in MeterBindersConfiguration should be optional


michaelom:
version: 1.0.0-rc.3

I would like to disable all `MeterBinder`s defined in `MeterBindersConfiguration` (JVM, Uptime, Processor, Logback). Here is an example:
```
    @Bean
    @ConditionalOnMissingBean(JvmMemoryMetrics.class) //   <=== bad
    public JvmMemoryMetrics jvmMemoryMetrics() {
        return new JvmMemoryMetrics();
    }
```
These MeterBinders are conditional beans, yet its very hard to disable them. The only way I could come up with was, to declare a custom NoOp-Implementation:

```
    @Bean
    public UptimeMetrics disabledUptimeMetrics() {
        return new UptimeMetrics() {
            @Override
            public void bindTo(MeterRegistry registry) {
            }
        };
    }
```

~This approach is bad and what's more, it is not working reliably: `MeterBindersConfiguration` is not hooked into Springs AutoConfiguration mechanism (via `spring.factories`).~ edit: This is nonsense. ;-)

I would prefer to disable `MeterBinder`s defined in `MeterBindersConfiguration` using a environment property. 

```
    @Bean
    @ConditionalOnProperty(some.property)
    public JvmMemoryMetrics jvmMemoryMetrics() {
        return new JvmMemoryMetrics();
    }
```

checketts:
Great feedback!

spencergibb:
I think the conditional on missing bean should stay.

michaelom:
That was quick.

> I think the conditional on missing bean should stay.

Works for me, I don't mind.
But I think it's very unlikely someone would subclass `JvmMemoryMetrics` or `ProcessorMetrics`. I would rather disable the beans and implement a custom `MetricsBinder` to measure memory or processor stats.

spencergibb:
:-) I didn't say, the conditional on property shouldn't be added. I think there should be both.

checketts:
Oh, right in case the user has defined their own.

jkschneider:
Fixed with #201. Thanks for the detailed report @michaelom!

PR #201, opened 2017-11-09
Change to ConditionalOnProperty, also prefix the cpu count metric

checketts:
Adresses #200

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
@@ -43,7 +43,7 @@ public class ProcessorMetrics implements MeterBinder {
     public void bindTo(MeterRegistry registry) {
         Runtime runtime = Runtime.getRuntime();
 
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
```diff
@@ -43,7 +43,7 @@ public ProcessorMetrics(Iterable<Tag> tags) {
     public void bindTo(MeterRegistry registry) {
         Runtime runtime = Runtime.getRuntime();
 
-        Gauge.builder("cpu", runtime, Runtime::availableProcessors)
+        Gauge.builder("system.cpu.count", runtime, Runtime::availableProcessors)
```
checketts: Can we also rename this metric @mweirauch @jkschneider so it uses the hierarchy?

checketts:
@spencergibb I re-added `@ConditionalOnBean` and left the class out, since it is implied by the return type. Is that correct?

spencergibb:
Yes

jkschneider:
Thanks @checketts.

PR #203, opened 2017-11-10
Unparsabled values are reported with details of the error

checketts:
I'm OK if you think this is un-needed and should be closed. I mainly did it to inspect that the API would allow the detail if I needed it.

At first I was worried we would need this due to how config is parsed lazily. But it turns out that is actually beneficial, because the user's code the is calling `Timer...register()` will be in the stack, so the metrics name will be locate-able. 

This change just points directly to the config line if the `my.threadpool` config was bad for the `my.threadpool.counter`, so the user won't have to look very hard


checketts:
Also I really like how you solved the global config.

jkschneider:
Seems better, since the effort has been made :)

Issue #28, opened 2017-05-29
Add ability to emit tracing and logging when metering


checketts:
Inspired by this talk: https://youtu.be/P8dc-rLnLr0?t=5h31m5s at Monitorama.

What if I were able to create a timer and when it runs it could check if there was an active slueth/zipkin trace and it could create a span. Also at the same time it could log out details. This way  app code would be instrumented by a single line and we could emit the log/metric/trace info on their behalf.

PR #206, opened 2017-11-12
Add OpenTracing to Timers

checketts:
@jkschneider This is regarding #28 (@adriancole I would love your feedback too)

Initially I took a 'listener' approach, but it wasn't useful for tracing since you need to have the active span so child spans can display properly.

Here are my findings:
* Listener approach doesn't work
* Only `Timer`s make sense for creating traces
* While I was (kind of) able to hack in support for the non-callback `record` method, those spans weren't useful, they ended up disconnected from the rest of their trace
* Being able to automatically expose the metric's tags in the span was really cool

Since traces were only useful for a sub set of metrics and you had to use the Timer the exactly correct way to get a useful measurement, I'm not in love with this feature.

image

I'm going to investigate solving this a different way with a `TracingTimer` so users could opt into which timers they wanted traced.  But I am interested to hear if you had any other feedback or suggestions, or if you think the behavior is useful.

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -10,6 +10,8 @@ apply plugin: 'nebula.optional-base'
 dependencies {
     compile 'org.hdrhistogram:HdrHistogram:2.1.10'
     compile 'org.latencyutils:LatencyUtils:2.0.3'
+    compile 'io.opentracing:opentracing-api:0.30.0'
+    compile 'io.opentracing:opentracing-noop:0.30.0'
 
     // instrumentation options
     compile 'io.dropwizard.metrics:metrics-core:3.+', optional
--- a/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/AtlasMeterRegistry.java
+++ b/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/AtlasMeterRegistry.java
@@ -117,7 +117,7 @@ public class AtlasMeterRegistry extends MeterRegistry {
             PercentileTimer.get(registry, spectatorId(id));
         }
 
-        SpectatorTimer timer = new SpectatorTimer(id, internalTimer, clock, histogramConfig);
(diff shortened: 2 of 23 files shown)


micrometer-core/build.gradle
```diff
@@ -10,6 +10,8 @@ apply plugin: 'nebula.optional-base'
 dependencies {
     compile 'org.hdrhistogram:HdrHistogram:2.1.10'
     compile 'org.latencyutils:LatencyUtils:2.0.3'
+    compile 'io.opentracing:opentracing-api:0.30.0'
```
adriancole: hmm adding a core dependency? probably there's another way. (ps 0.31 is api breaking)

adriancole:
I guess my feedback is this is a good example of why census has tracing apis separate from metrics apis. You don't get into the tagging trouble as you can access propagated tags independently from metrics or tracing. There's also a data model which means you can derive metrics later and forward them to micrometer that way. http://opencensus.io/

caveat: I've not tried to replicate exactly what you are doing here, but having to make a hard compile dep on one of the couple tracing libraries to get metrics hints a broader problem.

adriancole:
if you go this way, basically insulating from library drift is important (I wouldn't in good faith advise adding OpenTracing as a core dep to any library). For example, in census' agent they have a special type to avoid classpath problems. This exposes the minimum surface needed which can be usable  even with a library change https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/UrlInstrumentation.java

(pps instrumentation via agent or aspectj can be an approach to avoid having to do things like this.. it might be a matter of aligning methods so that they are easy to reach)

checketts:
@adriancole Thanks for the feedback. I'm very new to the distributed tracing realm (you can measure it in hours). I think I was under the incorrect assumption that OpenTracing has achieved a common/stable API like slf4j for tracing.

I'll definitely look into making it an optional dependency.

Largely my experiments in this area were to exercise the API and see if there were any pre-1.0 changes we should pursue.

PR #207, opened 2017-11-12
Add TracingTimer

checketts:
Alternative solution to #28 

This creates a `TracingTimer` that could be used to automatically generate tracing spans.

Instead of instrumenting at the registry level, I wrap the `Timer` that is returned by the `register` call.

FYI @adriancole

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -36,6 +36,11 @@ dependencies {
 
     compile 'com.squareup.okhttp3:okhttp:3.9.0',optional
 
+    //tracing
+    compile 'io.opentracing:opentracing-api:0.30.0', optional
+    testCompile 'io.opentracing:opentracing-mock:0.30.0'
+    testCompile 'io.opentracing:opentracing-util:0.30.0'
+
     testCompile 'io.projectreactor:reactor-test:3.1.0.RELEASE'
     testCompile 'io.projectreactor.ipc:reactor-netty:0.7.1.RELEASE'
 
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -37,6 +37,8 @@ dependencies {
     compile 'com.squareup.okhttp3:okhttp:3.9.0',optional
 
     //tracing
+    compile 'io.opencensus:opencensus-api:0.9.1', optional
+    testCompile 'io.opencensus:opencensus-impl:0.9.1'
     compile 'io.opentracing:opentracing-api:0.30.0', optional
     testCompile 'io.opentracing:opentracing-mock:0.30.0'
     testCompile 'io.opentracing:opentracing-util:0.30.0'
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
@@ -143,7 +143,7 @@ public interface Timer extends Meter {
         private String description;
         private final HistogramConfig.Builder histogramConfigBuilder;
 
-        Builder(String name) {
+        protected Builder(String name) {
             this.name = name;
             this.histogramConfigBuilder = new HistogramConfig.Builder();
             minimumExpectedValue(Duration.ofMillis(1));
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
@@ -143,7 +143,7 @@ public interface Timer extends Meter {
         private String description;
         private final HistogramConfig.Builder histogramConfigBuilder;
 
-        private Builder(String name) {
+        Builder(String name) {
             this.name = name;
             this.histogramConfigBuilder = new HistogramConfig.Builder();
             minimumExpectedValue(Duration.ofMillis(1));
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/TracingTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/TracingTimer.java
@@ -1,0 +1,124 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument;
+
+import io.opentracing.ActiveSpan;
+import io.opentracing.Tracer;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+public class TracingTimer implements Timer{
+
+    static Builder builder(String name, Tracer tracer) {
+        return new Builder(name, tracer);
+    }
+
+    static class Builder extends Timer.Builder{
+        private final Tracer tracer;
+
+        Builder(String name, Tracer tracer) {
+            super(name);
+            this.tracer = tracer;
+        }
+
+        @Override
+        public Timer register(MeterRegistry registry) {
+            return new TracingTimer(super.register(registry), tracer);
+        }
+    }
+
+    private final Timer delegate;
+    private final Tracer tracer;
+
+    TracingTimer(Timer delegate, Tracer tracer) {
+        this.delegate = delegate;
+        this.tracer = tracer;
+    }
+
+    private ActiveSpan createSpan() {
+        Tracer.SpanBuilder spanBuilder = tracer.buildSpan(getId().getName());
+        getId().getTags().forEach(t -> spanBuilder.withTag(t.getKey(), t.getValue()));
+        return spanBuilder.startActive();
+    }
+
+
+    @Override
+    public <T> T record(Supplier<T> f) {
+        try(@SuppressWarnings("unused") ActiveSpan span = createSpan()){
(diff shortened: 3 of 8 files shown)


micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/components/PersonController.java
```diff
@@ -30,19 +32,35 @@
...
 
     @GetMapping("/api/people")
     @Timed(percentiles = {0.5, 0.95, 0.999}, histogram = true)
     public List<String> allPeople() {
-        try {
-            Thread.sleep(200);
-        } catch (InterruptedException e) {
-            e.printStackTrace();
-        }
+        TracingTimer.builder("outerTimer", tracer)
```
checketts: Usage example here

micrometer-spring-legacy/src/samples/java/io/micrometer/spring/samples/PrometheusSample.java
```diff
@@ -25,4 +28,27 @@
     public static void main(String[] args) {
         new SpringApplicationBuilder(PrometheusSample.class).profiles("prometheus").run(args);
     }
+
+    @Bean
+    public Tracer tracer() {
+        Tracer tracer = new com.uber.jaeger.Configuration(
```
adriancole: why do you need to use uber's tracer.. I thought opentracing has a test variant?

checketts: This is in the samples and won't be checked in long term. 

It is essentially a spring boot app that I run to play around and explore. Thanks for the point about using the test variant in the unit tests.

I left it in the PR to demonstrate if anyone wanted to check it out and try running it.

adriancole:
cool. in the future, maybe add a comment for temporary code, so it is
obvious what to review

checketts:
Removed temp code. Added in correct unit test

micrometer-core/src/main/java/io/micrometer/core/instrument/TracingTimer.java
```diff
@@ -0,0 +1,124 @@
...
+package io.micrometer.core.instrument;
+
+import io.opentracing.ActiveSpan;
+import io.opentracing.Tracer;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+public class TracingTimer implements Timer{
```
shakuzen: Perhaps `OpenTracingTimer` is a better name as it is more specific. Users who do tracing and are interested in this concept but aren't using OpenTracing might be confused by the current name.

micrometer-core/src/main/java/io/micrometer/core/instrument/TracingTimer.java
```diff
@@ -0,0 +1,124 @@
...
+    static class Builder extends Timer.Builder{
+        private final Tracer tracer;
+
+        Builder(String name, Tracer tracer) {
+            super(name);
+            this.tracer = tracer;
+        }
+
+        @Override
+        public Timer register(MeterRegistry registry) {
```
jkschneider: any reason we can't return the more specific type `TracerTimer`?

jkschneider:
@adriancole I'm trying to understand a few comments from the earlier incarnation of this PR (#206).

> I guess my feedback is this is a good example of why census has tracing apis separate from metrics apis. You don't get into the tagging trouble as you can access propagated tags independently from metrics or tracing. There's also a data model which means you can derive metrics later and forward them to micrometer that way. http://opencensus.io/

I'm comfortable with an optional dependency that provides opentracing support _and also_ provide opencensus support separately -- all in the spirit of there's no need for Micrometer to pick winners. That said, I don't know enough about the tracing world to know: how entrenched is opentracing even if it is objectively lacking in some way? Is long-term support of it worthwhile?

> You don't get into the tagging trouble as you can access propagated tags independently from metrics or tracing.

What is the tagging trouble you are referring to?

micrometer-core/build.gradle
```diff
@@ -36,6 +36,11 @@ dependencies {
 
     compile 'com.squareup.okhttp3:okhttp:3.9.0',optional
 
+    //tracing
+    compile 'io.opentracing:opentracing-api:0.30.0', optional
```
jkschneider: re: @adriancole earlier comment that 0.31.0 has a breaking API change.

checketts: I just upgrade to `0.31.0-RC1` and it does break the API I was using. Should I program against the RC? It appears the `Scope` API is the way forward (and matches the OpenCensus examples I looked at)

micrometer-core/src/main/java/io/micrometer/core/instrument/TracingTimer.java
checketts: Good idea. I had a flashback of Java 4 not supporting covariant return types.

micrometer-core/build.gradle
jkschneider: I really don't know. Let's hang tight and wait for Adrian's input.

checketts: This PR (https://github.com/opentracing/opentracing-java/issues/189) covers the changes. Looks like they will include a backwards compatible shim. So coding against `0.30.0` is probably the right decision. But I agree that @adriancole 's input will be good.

adriancole:
Main thing is to not compile against a dependency which can cause thrash in
core. For example, adding a class compiled against a type that can drift
means any release now includes a type that can conflict yet has nothing to
do with core. Even optional has this problem as you cannot remove this type
from the jar. It is quite unusual to be expected to

On 14 Nov 2017 12:55 am, "checketts" <notifications@github.com> wrote:

*@checketts* commented on this pull request.
------------------------------

In micrometer-core/build.gradle
<https://github.com/micrometer-metrics/micrometer/pull/207#discussion_r150599780>
:

> @@ -36,6 +36,11 @@ dependencies {

     compile 'com.squareup.okhttp3:okhttp:3.9.0',optional

+    //tracing
+    compile 'io.opentracing:opentracing-api:0.30.0', optional

This PR (opentracing/opentracing-java#189
<https://github.com/opentracing/opentracing-java/issues/189>) covers the
changes. Looks like they will include a backwards compatible shim. So
coding against 0.30.0 is probably the right decision. But I agree that
@adriancole <https://github.com/adriancole> 's input will be good.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/micrometer-metrics/micrometer/pull/207#discussion_r150599780>,
or mute the thread

adriancole:
Wrt tags the opentracing api forces many things through the tracer api. For
example to access propagated tags in census you do not need to have a
reference to a span model. In opentracing the span is a heavy reference and
much bigger api than needed just to access tags.

On 13 Nov 2017 10:16 pm, "Jon Schneider" <notifications@github.com> wrote:

@adriancole <https://github.com/adriancole> I'm trying to understand a few
comments from the earlier incarnation of this PR (#206
<https://github.com/micrometer-metrics/micrometer/pull/206>).

I guess my feedback is this is a good example of why census has tracing
apis separate from metrics apis. You don't get into the tagging trouble as
you can access propagated tags independently from metrics or tracing.
There's also a data model which means you can derive metrics later and
forward them to micrometer that way. http://opencensus.io/

I'm comfortable with an optional dependency that provides opentracing
support *and also* provide opencensus support separately -- all in the
spirit of there's no need for Micrometer to pick winners. That said, I
don't know enough about the tracing world to know: how entrenched is
opentracing even if it is objectively lacking in some way? Is long-term
support of it worthwhile?

You don't get into the tagging trouble as you can access propagated tags
independently from metrics or tracing.

What is the tagging trouble you are referring to?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/micrometer-metrics/micrometer/pull/207#issuecomment-343931532>,
or mute the thread

jkschneider:
> Seems very odd to me that this code isn't in a contrib module

I think this should be in a contrib module. 

I want to avoid the explosion of dependencies that something like Prometheus has as I believe there is a discoverability problem inherent in that. So I'm comfortable having optional deps on libraries where drift is unlikely (e.g. Guava cache). But a pre-1.0 library with demonstrated churn is not a good fit for core.

shakuzen:
Has it been considered to ask the OpenTracing Java maintainers if they'd be interested in hosting this code in their library or org? It would seem easier for them to maintain it and keep it in sync with their API, as the assumption is that micrometer's API is going to be more stable.

jkschneider:
Good point @shakuzen. We'll be comitted to 3 year support of the 1.0 line.

checketts:
I'm totally fine with this code being hosted elsewhere. The one finding I wanted to expose pre-1.0 is changing the visibility of the `Timer.Builder` so it can be sub-classes/extended as in this usecase, allowing extensions like this to be possible.

If that isn't desire-able, then I could investigate based on that feedback.

I'm going to refactor this into a separate package to see if there are other visibility issues that need resolution. (Or move it to a contrib module if that exists)

adriancole:
I'm sure someone will approve creating a repo here btw
https://github.com/opentracing-contrib/

thanks for considering this pov, almost all dependencies can bite you
eventually. Ex even jsr305 can bite you in java 9!

micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
```diff
@@ -143,7 +143,7 @@ static Builder builder(String name) {
         private String description;
         private final HistogramConfig.Builder histogramConfigBuilder;
 
-        private Builder(String name) {
+        protected Builder(String name) {
```
checketts: @jkschneider This single line change is the only really 'required' change in Micrometer that will allow creating the other Tracing timer classes in other repos.

I would love to leave the Opentracing/openCensus implementations in Micrometer as Optional deps. I did research and the breaking API change that Adrian brought up will be properly shimmed, and should actually behave as a deprecation for a good while.

checketts:
Yes. This PR was just exploring to ensure the underlying timer was extensible enough to allow the user to extend it. If we could make the `builder` protected then that would allow a user to subclass Timer and make their own `TracingTimer`.  I think closing this for now is fine.

PR #209, opened 2017-11-12
Make TimeWindowHistogram and TimeWindowLatencyHistogram thread-safe

trustin:
Motivation:

TimeWindow(Latency)Histogram is not thread-safe because:

- accumulatedHistogram and intervalHistogram are not thread-safe and
  they are accessed without proper synchronization.
- rotate() can be invoked from multiple threads simultaneously, causing
  a race between them.

Modifications:

- Protect the access to accumulatedHistogram and intervalHistogram with
  synchronization.
- Make sure bucket rotation is performed by only one thread.
- Catch IndexOutOfBoundsException instead of ArrayIndexOutOfBoundsException
  because HdrHistogram sometimes throws IOOBE.

Result:

- Thread safety

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/histogram/TimeWindowHistogram.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/histogram/TimeWindowHistogram.java
@@ -20,23 +20,31 @@ import io.micrometer.core.instrument.Clock;
 import org.HdrHistogram.DoubleHistogram;
 import org.HdrHistogram.DoubleRecorder;
 
-import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


trustin:
Side note: It seems to me that there's no need to use `LatencyUtils` for `TimeWindowLatencyHistogram`, because we just use it for recording the values as they are and we do not do any pause detection, etc. What about just using `Recorder` instead of `LatencyStats`? I can update this PR to replace `LatencyStats` with `Recorder` if @jkschneider is fine. Thoughts?

jkschneider:
Really appreciate the work on this @trustin

trustin:
Thanks for merging! What do you think about replacing LatencyUtils with Recorder?

PR #210, opened 2017-11-13
Deduplicate the code between TimeWindowHistogram and TimeWindowLatencyHistogram

trustin:
Motivation:

TimeWindowHistogram and TimeWindowLatencyHistogram shares majority of
their logic. A change in one place needs the same change in the other
counterpart, which is error-prone.

Modifications:

- Extract the common logic between TimeWindowHistogram and
  TimeWindowLatencyHistogram into TimeWindowHistogramBase
- Use the same number of significant value digits for both histograms
- Add NoopHistogram and use it to reset LatencyStats
- Add TimeWindowRotationTest to verify that sliding window works as
  expected
- Miscellaneous:
  - Use JDK TimeUnit.convert() instead of TimeUtils.convert() when
    there's no need for casting to double

Result:

- Less code duplication
- Increased code coverage


trustin:
Please do not merge yet. Looking into a way not to create a new `LatencyStats` on rotation.

trustin:
Ready to be merged. `LatencyStats` doesn't need to be created on each rotation now.

trustin:
@jkschneider Thanks a lot for merging!

PR #212, opened 2017-11-13
Wait for metrics to appear before asserting on them being present

checketts:
This will make the hystrix tests less flaky

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/hystrix/MicrometerMetricsPublisherCommandTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/hystrix/MicrometerMetricsPublisherCommandTest.java
@@ -28,8 +28,10 @@ import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Flux;
 
 
+import java.time.Duration;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/hystrix/MicrometerMetricsPublisherCommandTest.java
```diff
@@ -71,16 +74,24 @@ void testCumulativeCounters() throws Exception {
...
+
+        assertExecutionMetric(registry, "success", 24.0);
+        assertThat(registry.find("hystrix.execution").tags(tags).tags("event", "timeout").functionCounter().map(FunctionCounter::count)).hasValue(3.0);
+        assertThat(registry.find("hystrix.execution").tags(tags).tags("event", "failure").functionCounter().map(FunctionCounter::count)).hasValue(6.0);
         assertThat(registry.find("hystrix.execution").tags(tags).tags("event", "short_circuited").functionCounter().map(FunctionCounter::count)).hasValue(0.0);
         assertThat(registry.find("hystrix.circuit.breaker.open").tags(tags).gauge().map(Gauge::value)).hasValue(0.0);
     }
 
+    private void assertExecutionMetric(SimpleMeterRegistry registry, String eventType, double count) {
+        Flux.interval(Duration.ofMillis(50))
```
jkschneider: @checketts Polished this up a bit. If I understand the logic in the previous iteration, it would have looped indefinitely if the condition was never met.

jkschneider: Also, pretty good example of `Flux.interval` ;)

checketts: It would have looped indefinitely? I had it check `while(System.currentTimeMillis() - startTime < waitTimeMs) {` so it should have exited correctly. 

Thanks for the `Flux.interval`! The more I see examples like this the more hopeful I am that I'll make the reactive jump eventually. ;)

checketts: Hmm, one drawback with the flux, is we lost some details on failure. Like the build just failed, but it isn't clear why. 30 seconds should have been plenty for the metric to persist.

PR #215, opened 2017-11-14
Fix thread safety issue in MeterRegistry

### LINKED ISSUE #208: Thread safety issue in MeterRegistry.getMeters()
https://github.com/micrometer-metrics/micrometer/blob/v1.0.0-rc.3/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L185

A user can:

- modify `MeterRegistry.meterMap` directly by removing an element from the returned collection:

  ```java
  register.getMeters().clear();
  ```

- iterate over `MeterRegistry.meterMap` while other threads add a new `Meter` to the map, causing the iteration loop throw `ConcurrentModificationException`.

`MeterRegistry.getMeters()` should 1) return a copy of the map values or 2) `MeterRegistry.meterMap` should be a `ConcurrentHashMap`.

trustin:
Motivation:

MeterRegistry.getMeters() returns meterMap.values(), which exposes the
value collection without any protection. It means:

- The caller can remove a meter from the map without restriction.
- The caller may get ConcurrentModificationException while iterating
  over the collection.

Modifications:

- Use ConcurrentHashMap instead of HashMap for meterMap
  - Add some comment about why ConcurrentHashMap.computeIfAbsent() is
    not used
- Improve the efficiency of Search.meters()
- Overall code clean up
  - Removal of redundant package names of imported classes
  - Removal of redundant class names of static-imported methods
  - Add missing curly braces
  - Fix incorrect location of Search.tags(String...) Javadoc
  - Use EnumMap instead of HashMap if applicable

Result:

- Fixes #208
- Thread safety
- Potentially less contention
- Closes the loophole that allowed a user to remove a Meter from a registry

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -16,6 +16,7 @@
 package io.micrometer.core.instrument;
 
 import io.micrometer.core.annotation.Incubating;
+import io.micrometer.core.instrument.Meter.Id;
 import io.micrometer.core.instrument.config.MeterFilter;
 import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.histogram.HistogramConfig;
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -48,11 +53,12 @@
 public abstract class MeterRegistry {
     protected final Clock clock;
 
-    public MeterRegistry(Clock clock) {
+    protected MeterRegistry(Clock clock) {
         this.clock = clock;
     }
 
-    private final Map<Meter.Id, Meter> meterMap = new HashMap<>();
+    private final ConcurrentMap<Id, Meter> meterMap = new ConcurrentHashMap<>();
```
jkschneider: This has caused deadlock in the past, will have to walk this back I think.

trustin: It happens when you do nested registration in the builder function. Described at: https://bugs.openjdk.java.net/browse/JDK-8062841

PR #217, opened 2017-11-14
Add a way to take a snapshot from DistributionSummary and Timer

### LINKED ISSUE #211: Improve histograms to return multiple percentiles or histogram buckets with one call
cc / @trustin

trustin:
Motivation:

Calling percentile(double) or histogramCountAtValue(long) for all
monitored percentiles or buckets are inefficient because each call may
involve an accumulation of two histograms when a new value was recorded
in-between, which is often the case when a user application is handling
a lot of requests simultaneously.

Modifications:

- Add HistogramSnapshot, ValueAtPercentile and CountAtValue to hold the
  snapshot of a Timer of a DistributionSummary
- Add Timer.takeSnapshot() and DistributionSummary.takeSnapshot()
- Use takeSnapshot() for:
  - PrometheusMeterRegistry
  - InfluxMeterRegistry
  - DatadogMeterRegistry
  - Note: Other MeterRegistry implementations were not updated since it
    registers an individual gauge for each percentile, which needs
    non-trivial changes.

Result:

- Fixes #211

### DIFF
--- a/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
+++ b/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
@@ -143,34 +143,36 @@ public class DatadogMeterRegistry extends StepMeterRegistry {
     }
 
     private Stream<String> writeTimer(Timer timer) {
-        long wallTime = clock.wallTime();
+        final long wallTime = clock.wallTime();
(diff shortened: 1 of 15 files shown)


jkschneider:
Good practical balance. Other backends won't ship most histogram buckets anyway because they don't support aggregable percentiles). In these cases, it's most likely that folks will only configure the publication of a handful of client-side percentiles anyway.

trustin:
@jkschneider Thanks a lot for swift review and merge. Looking forward to the next release!

PR #220, opened 2017-11-15
Validate HistogramConfig before creating TimeWindowHistogramBase

trustin:


jkschneider:
The basic purpose of this is to prevent a selection of `histogramExpiry` and `histogramBufferLength` such that `histogramExpiry / histogramBufferLength < 1 ms`.

PR #222, opened 2017-11-15
CumulativeDistributionSummary - prevent losing the maximum in case of data-race 

vladimir-bukhtoyarov:
Currently there is data-race inside CumulativeDistributionSummary when two threads coming in parallel to ```recordNonNegative``` method, as result max value can be lost by missed write.

Fail scenario:
- Preconditions: Thera are two thread **T1** which wants to write 42 and **T2** which wants to write 666. Current max is 13.
- Both **T1** and **T2** reached the recordNonNegative in same time, both got 13 as current maximum.
- **T2** seen that 666 greater than 13 and did the write.
- **T1** seen that 42 greater than 13 and did the overwrite.
- As result actual maximum 666 is missed.


jkschneider:
Thanks for the detailed analysis and solution @vladimir-bukhtoyarov.

PR #223, opened 2017-11-15
CumulativeTimer - prevent losing the maximum in case of data-race

vladimir-bukhtoyarov:
The problem description is same with [this request](https://github.com/micrometer-metrics/micrometer/pull/222)


checketts:
Thanks!

PR #227, opened 2017-11-15
Fix thread safety in MeterRegistry

trustin:
- Remove the double checked locking idiom on meterMap.get() in
  registerMeterIfNecessary() because it is not safe to do so.
  Unsynchronized meterMap.get() may enter into a weird state such as
  infinite loop.
- Add missing synchronization on MeterRegistry.meterMap.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -333,31 +333,33 @@ public abstract class MeterRegistry {
         }
 
         public Collection<Meter> meters() {
-            Stream<Entry<Id, Meter>> entryStream =
-                meterMap.entrySet().stream().filter(e -> e.getKey().getName().equals(name));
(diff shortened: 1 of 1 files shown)


trustin:
https://stackoverflow.com/questions/33350489/double-checked-locking-with-regular-hashmap

jkschneider:
Section on *Does the new memory model fix the "double-checked locking" problem?*:
http://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html

jkschneider:
Post JDK4, if the map value is immutable, double checked locking is safe. It's probably a stretch to assume every Meter implementation is immutable, though in practice all of them are currently.
http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

trustin:
DCL only works when the field being accessed is volatile. HashMap's table elements are not volatile. Also, HashMap update involves potentially multiple writes, giving the reader a chance of observing the partial state of the internal structure, and it's dangerous.

PR #228, opened 2017-11-15
WIP: Jersey2 support

mweirauch:
A work in progress PR. Comments welcome!

## Known Issues/Todos
* [ ] Spring auto-configuration
* [ ] include applicationPath in uri when it is set?
* [ ] annotation based Timer configuration
* [ ] configuration based Timer configuration
* [ ] WebMvc `MetricsFilter` also records requests in certain configurations

  Read as follows: jersey-type, jersey-filter-order, applicationPath

  * servlet, *initial*, /jersey
    * records /hello/{name} (Jersey)
    * records /hello/peter (MVC)
  * servlet, -2147483648, /jersey
    * records /hello/{name} (Jersey)
    * records /hello/peter (MVC)
  * filter, *initial*, /jersey
    * records /hello/{name} (Jersey)
    * records / (MVC)
  * filter, -2147483648, /jersey
    * records /hello/{name} (Jersey)
    * nothing recorded (MVC)

  The following setup with SB 1.5 is used:
  ```
  server:
    port: 7001

  management:
    context-path: /mgmt
    security:
      enabled: false

  endpoints:
    metrics:
      enabled: false
      filter:
        enabled: false
        
  spring:
    application:
      name: ${project.artifactId}
    autoconfigure:
      exclude:
      - org.springframework.boot.actuate.autoconfigure.PublicMetricsAutoConfiguration
    metrics:
      web:
        server:
          auto-time-requests: true
          record-request-percentiles: false
    jersey:
      filter:
        order: -2147483648
      type: filter
      application-path: /jersey
      servlet:
        load-on-startup: 0
  ```


mweirauch:
I think this could go under some initial review.
The similarity to WebMVC instrumentation is undeniable. A great template/source. Perhaps some parts (`TimerConfig` or `Timed` annotation extraction) could be made generally available in core.

mweirauch:
Tested with Spring-Boot `2.0.0-M7` and manual configuration of the listener. Works.

jkschneider:
FYI @wilkinsona.

jkschneider:
@twicksell - Does this help you at all? Thoughts if it does?

twicksell:
Sadly I don't think so, we don't much usage of Jersey 2. Jersey 1, yes, but I don't know if there is any other real demand for that.

micrometer-jersey2/build.gradle
```diff
@@ -0,0 +1,10 @@
+dependencies {
+    compile project(':micrometer-core')
+    compile 'org.glassfish.jersey.core:jersey-server:2.25.1'
```
twicksell: Shouldn't this be an optional dependency? I believe the typical pattern in Boot is to provide functionality guarded by ConditionalOnClass, but not express opinions on dependency versions except in starter poms.

mweirauch: This "lib" provides standalone Jersey2 instrumentation support for micromter. The Spring-Boot auto-configuration is handled in `spring-boot-legacy` and `spring-boot` (2.x) respectively. Thanks for looking!

jkschneider: Yep, we'll make it optional.

mweirauch: I don't get it. If you don't pull in this lib (`micrometer-jersey2`) as a dependency in your project, the auto-configuration won't kick in as it does `@ConditionalOnClass(MicrometerApplicationEventListener.class)`. (Remember, this instrumentation support is not embedded in `micrometer-core`.)

jkschneider: Presumably in a Jersey app you want instrumented, your app already has an opinion about the Jersey version it is using. Suppose you were using a version < 2.25.1. We don't want conflict resolution on a "latest version conflict resolution" build tool like Gradle to pull up your Jersey version simply because you wanted to instrument it.

mweirauch: Ok, made it optional. Perhaps I am so used to maven and it's "nearest dependency wins" behaviour and I accept that apparently not every developer is checking his dependency tree.

mweirauch:
As promised in slack Jon :)

Apart from the fact that CircleCI is constantly failing, rc.6 being out, what needs to be done getting this in?

The instrumentation code is running in our production for a little more then a week. No issues so far.

Is there still some discussion going on about the non-core/external instrumentation package/project naming? Or should this even be part for core in the meantime? Just let me know.

jkschneider:
Merged in a big mess that ended with 9ae63906cc78d44c444dbb43340b5b287d6b645e. Phew...

If you don't mind @mweirauch, please sanity check `MetricsRequestEventListener` (renamed from `MicrometerRequestEventListener`). It incorporates the new timer and long task timer `Sample` types, and some other errata.

jkschneider:
Boot 2 changes related to generalized logic that is shared between the Jersey implementation and Servlet implementation.

PR #233, opened 2017-11-18
Remove unused imports

izeye:
This PR also applies Checkstyle plugin and adds its `UnusedImports` module to prevent unused imports from sneaking into codebase again.


jkschneider:
@izeye Thanks for the checkstyle addition!

Issue #234, opened 2017-11-19
CumulativeCounter imports Google AtomicDouble


philwebb:
I think [this import](https://github.com/micrometer-metrics/micrometer/blob/afeb04598c03e1c204e57d4aa0895308f7366edf/micrometer-core/src/main/java/io/micrometer/core/instrument/simple/CumulativeCounter.java#L18) is probably wrong and it should be the JDK `AtomicDouble`.

philwebb:
FWIW we also hit this a lot in Spring Boot and ended up setting a [checkstyle rule](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-parent/src/checkstyle/checkstyle.xml#L76) to prevent it.

jkschneider:
Fixed with #238

PR #238, opened 2017-11-19
Remove unintended import of Guava classes

checketts:
Addresses #234


checketts:
FYI @philwebb

jkschneider:
Thanks Clint

Issue #239, opened 2017-11-19
Expose Hystrix metrics on time taken by http requests


fiunchinho:
I'm using Feign and Hystrix to make http calls to other services. I'd like to expose metrics on the time taken by those requests so I can graph the latency between services, having different projections for different HTTP status codes and paths.

checketts:
Currently the Hystrix metrics only track success/fallback counts. We need to add in the timing metrics. 

See https://github.com/ahus1/prometheus-hystrix/blob/master/src/main/java/com/soundcloud/prometheus/hystrix/HystrixPrometheusMetricsPublisherCommand.java#L145

jkschneider:
This all makes good sense to support, given that Spring Cloud has Feign as a supported path.

Issue #237, opened 2017-11-19
Add Autoconfiguration support for Hystrix metrics


checketts:
The Hystrix metrics publisher doesn't currently autoconfigure itself. It should add a `@ConditionalOnClass` config.

jkschneider:
This is solved with #240 right?

checketts:
Yes

PR #240, opened 2017-11-21
Use direct counter/timer instrumentation for hystrix

checketts:
FYI @fiunchinho

@spencergibb I followed the example from https://github.com/ahus1/prometheus-hystrix/blob/1c71a2c400342bfd9bd09e0f1e346b4e73ce513b/src/main/java/com/soundcloud/prometheus/hystrix/HystrixPrometheusMetricsPublisherCommand.java#L193 that instruments directly via Micrometer.  This will allow histograms to be used, or other configuration via MeterFilter configuration.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/hystrix/HystrixMetricsBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/hystrix/HystrixMetricsBinder.java
@@ -1,0 +1,30 @@
+package io.micrometer.core.instrument.binder.hystrix;
+
+import com.netflix.hystrix.strategy.HystrixPlugins;
+import com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategy;
+import com.netflix.hystrix.strategy.eventnotifier.HystrixEventNotifier;
+import com.netflix.hystrix.strategy.executionhook.HystrixCommandExecutionHook;
+import com.netflix.hystrix.strategy.properties.HystrixPropertiesStrategy;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+public class HystrixMetricsBinder implements MeterBinder {
+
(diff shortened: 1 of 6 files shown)


checketts:
@jkschneider This PR addresses #239 and #237

checketts:
@ahus1 I would love your feedback. This is largely inspired by your Prometheus metric as mentioned above. Some additional changes are that I broke the execution counters into 3 groups: Execution, Fallback and Other. Similar to the ones described in https://github.com/Netflix/Hystrix/wiki/Metrics-and-Monitoring#command-execution-event-types-comnetflixhystrixhystrixeventtype

I'm hoping this will allow a Prometheus query like `sum(hystrix_execution_total{group="agroup", key="akey"}) by (event)` to give a useful breakdown of command executions. (And keep falback and the other events separate)

checketts:
@jkschneider I think we can move forward with this. I don't think we'll get any further comments.

fiunchinho:
These changes show metrics now! Nice :) How would I go about adding percentiles? At the moment I can only see the `*_max` metric for 100 percentile, and `*_sum` / `*_count` metrics for the average.

checketts:
@fiunchinho The Prometheus example sets some percentiles see: https://github.com/micrometer-metrics/micrometer/commit/df908db57de0ec51d5a6cb7eef070af9135be429

Config:
```
spring.metrics.filter:
  hystrix:
   percentiles: 0.50,0.95,0.999
```

It adds the following percentiles:

```
hystrix_latency_execution_duration_seconds{group="PersonController",key="allPeople",quantile="0.5",} 0.205520895
hystrix_latency_execution_duration_seconds{group="PersonController",key="allPeople",quantile="0.95",} 0.239075327
hystrix_latency_execution_duration_seconds{group="PersonController",key="allPeople",quantile="0.999",} 0.239075327
```

It also means you could enable histograms so you can do cross-instance aggregates. It doesn't support setting percentiles on individual commands, but that can be added via your own custom `MeterFilter`.

Issue #241, opened 2017-11-21
Use immutable map 


padilo:
Checking the code I saw the following:

https://github.com/micrometer-metrics/micrometer/blob/afeb04598c03e1c204e57d4aa0895308f7366edf/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L653-L667

If I use a meter that has a dynamic tag for instance getting the counter as follows:

```
counter = registry.counter("my.metric.custom", "my-tag", tagValue);
```

That would end up using a lock to just check the counter already exists and created if not, if using its cache, or just create my own counter cache, which I think it shouldn't be necessary.

I don't know if there is way to use this kind of metrics that are a bit dynamic without ending in the contention it will give you a lock... If there isn't, I suggest to implement that Map using immutable maps, which will allow to use the double check and just locking in case the Map needs to be modified (reconstructed).

I can PR that you want.

jkschneider:
@padilo Looks like your example was cut off. Meter ids are already immutable. Could you elaborate?

padilo:
Yep, sorry. I rewrite the whole thing.

Yes, the problem is the `meterMap` attribute on the `MetricRegistry`, that isn't immutable and to register a new metric or even check if there is one to add it, it will end up on a synchronized block. Which may be fine if register is only use it once, but not if you have metrics that are not completely static or you need to rely on the list of Meter the `MetricRegistry` has.

Sorry If I don't explain it too well

jkschneider:
@padilo There was an extensive look at the synchronization behavior in #227, and we concluded that this is the best we can do. I don't know what immutable map construct you have in mind, but if you even want to just sketch this out roughly, I'd be interested in looking at it.

@smaldini suggested that using an atomic guard with a `for ;;` might be better, but I haven't wrapped my mind around that pattern yet.

PR #242, opened 2017-11-22
Move metermap to unmodificable instance

padilo:
This is what I tried to explain on #241

This way we can use a double check, at cost of O(N) on register a new instance on the `MeterRegistry` which I think is much better than locking in case of having dynamic meter (N=meterMap.length()).

Keep in mind that before this PR using `MeterRegistry.registerMeterIfNecessary(...)` it will always cause a lock. Now it will only lock in case the `meterMap` needs to be updated.

I don't know if I miss something :)

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -56,7 +56,7 @@ public abstract class MeterRegistry {
         this.clock = clock;
     }
 
-    private Map<Id, Meter> meterMap = Collections.emptyMap();
+    private volatile Map<Id, Meter> meterMap = Collections.emptyMap();
     private final List<MeterFilter> filters = new ArrayList<>();
 
     /**
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -56,6 +56,7 @@ public abstract class MeterRegistry {
         this.clock = clock;
     }
 
+    private final Object meterMapLock = new Object();
     private volatile Map<Id, Meter> meterMap = Collections.emptyMap();
     private final List<MeterFilter> filters = new ArrayList<>();
 
@@ -658,7 +659,7 @@ public abstract class MeterRegistry {
         Meter m = meterMap.get(mappedId);
 
         if (m == null) {
-            synchronized (this) {
+            synchronized (meterMapLock) {
                 m = meterMap.get(mappedId);
 
                 if (m == null) {
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -56,7 +56,7 @@ public abstract class MeterRegistry {
         this.clock = clock;
     }
 
-    private final Map<Id, Meter> meterMap = new HashMap<>();
+    private Map<Id, Meter> meterMap = Collections.emptyMap();
     private final List<MeterFilter> filters = new ArrayList<>();
 
     /**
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -674,6 +654,32 @@ public More more() {
         return (M) m;
     }
 
+    private Meter getOrCreateMeter(HistogramConfig config, BiFunction<Id, HistogramConfig, Meter> builder, Id mappedId) {
+        Meter m = meterMap.get(mappedId);
+
+        if (m == null) {
+            synchronized (this) {
+                m = meterMap.get(mappedId);
```
jkschneider: `meterMap` needs to be `volatile`, at least.

jkschneider:
I suspect that this kind of double-checking is still not as efficient as an atomic guard. We were optimizing for correctness over performance in rc4, but do need to revisit this.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
padilo: Sure

padilo:
Probably there is a better way to do that, but honestly I don't know how to do it. I agree that this should be revisited, but I think this PR will improve it a lot, because the meters aren't going to grow forever, and probably if there are many call to `registerMeterIfNecessary` they are going to return an already existent `Meter` instead of creating it. With this implementation this more likely operation should be more efficient.

jkschneider:
If you've got a moment, could you double check this @trustin? Pun intended...

trustin:
This is not correct. Making a HashMap field volatile doesn't make an unsynchronized read correct at all. If you do not want synchronized reads then you must use `ConcurrentHashMap` or equivalent implementation.

padilo:
I think unsynchronized reads are ok here. As long as we don't have
thread-safety problem.

I did use too much volatile, but what it seems to ensure is that the
reference to the hashmap will not be cached by threads, seems to be a good
idea to put it, if not, under some circumstances we may see different
hashmap on threads.

Please correct me if I'm wrong.

trustin:
Nope. Unsynchronized reads on a HashMap, whose internal structure is updated by multiple threads, is not thread-safe.

padilo:
Yes, I know. In this case this won't happen, as the Map instance is unmodificable, and It's recreated each time it needs to be modified.

trustin:
I see. Sorry I misread the code. I'm curious whether this is a good idea though because it can produce quite a bit of garbage when an application starts up and registers a lot of meters.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -674,6 +654,32 @@ public More more() {
...
+                }
+            }
+        }
+        return m;
+    }
+
+    private void register(Id id, Meter meter) {
+        HashMap<Id, Meter> newMeterMap = new HashMap<>();
+
+        newMeterMap.putAll(meterMap);
```
trustin: Probably faster if we use `HashMap.clone()` here.

trustin: Nevermind. It seems like there's no difference according to the code.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
vladimir-bukhtoyarov: @padilo please make one of:
- Add to javadocs that MeterRegistry does synchronization by itself instead of private non-shared lock, it is good practice to always document that synchronization performed on "this" because users will be better informed about potential deadlocks.
- Or just use private non-shared lock.

trustin: +1 for private non-shared lock

vladimir-bukhtoyarov:
Request looks correct, now registry acts like CopyOnWrite collections from JDK.

jkschneider:
Thanks for the reviews everyone. Very clever idea @padilo!

padilo:
Thank you all guys. :+1:

PR #245, opened 2017-11-23
Remove duplication in MeterRegistry.Search

izeye:
This PR removes duplication in `MeterRegistry.Search`.


jkschneider:
Thanks @izeye

PR #252, opened 2017-11-27
If a controller has async operation, e.g. which returns DeferredResul…

pavel-golub-epam:
…t<T>, then MetricsFilter.doFilterInternal is executed twice: when request handling starts and when completes. So called preHandle method should not reset TIMING_REQUEST_ATTRIBUTE value the second time.


jkschneider:
Good catch, thanks!

Issue #257, opened 2017-11-30
Add TomcatPublicMetrics equivalent for MicroMeter


checketts:
This was uncovered via #256

wilkinsona:
I think this may need to go in Boot (at least for Boot 2.0) as getting hold of the embedded Tomcat instance requires use Boot classes. If Micrometer did that, we’d have a circular dependency between the projects.

checketts:
The Binder will exist in micrometer. The autoconfig would exist in boot. I'll create one autoconfig in the spring-legacy module and each out to @wilkinsona or @jkschneider for the Boot 2 one.

jkschneider:
Your intuition is spot on @checketts.

wilkinsona:
Micrometer, a dependency of Boot, shouldn't know about the embedded containers that Boot supports. If it does, we have a conceptual cycle between projects. This issue is proposing Tomcat-specific metrics, but we may also want to provide metrics for Jetty and Undertow. That would broaden Micrometer's knowledge of containers that Boot supports which feels wrong to me.

#260 also illustrates a further problem with this approach. It's building against Tomcat 9.0.1, but Boot's using 8.5.x (8.5.23 at the moment). Having to keep Boot and Micrometer's Tomcat versions in sync further reinforces the cycle.

I think we should consider all of this functionality going in Boot.

checketts:
Tomcat isn't the first usecase. Notice Jetty metrics also exist in Micrometer (https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyStatisticsMetrics.java) and are used by non-SpringBoot enabled apps. (I'm not sure if SpringBoot 2 is leveraging that yet, I'm a bit behind on that adoption)

Eventually, as Micrometer grows in acceptance I would love to see the instrumentation added to the individual projects (as Hikari did)

I downgraded the Tomcat dep version. The `Manager` class that the metrics derive from is a stable interface that isn't changing. 

I'm not opposed to leaving the Tomcat binder out of Micrometer. But if we can provide support for metrics that are decoupled from SpringBoot I want to encourage that, especially if it improves MicroMeter adoption.

jkschneider:
[...]

PR #260, opened 2017-11-30
Add tomcat session metrics

checketts:
Addresses #257 

@jkschneider It looks like Tomcat has many metrics available via JMX. Would it make sense to pull those values from JMX and expose as metrics?

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -263,18 +263,6 @@ public abstract class MeterRegistry {
             return this;
         }
 
-        public Optional<Double> firstValue(){
-            return meters().stream().map(m -> {
-                Iterator<Measurement> measurements = m.measure().iterator();
-                if(measurements.hasNext()) {
-                    return measurements.next().getValue();
-                } else {
-                    return null;
-                }
-            }).findFirst();
-        }
-
-
         public Optional<Timer> timer() {
             return findAny(Timer.class);
         }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -263,6 +263,18 @@ public abstract class MeterRegistry {
             return this;
         }
 
+        public Optional<Double> firstValue(){
+            return meters().stream().map(m -> {
+                Iterator<Measurement> measurements = m.measure().iterator();
+                if(measurements.hasNext()) {
+                    return measurements.next().getValue();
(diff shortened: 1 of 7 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -263,6 +263,18 @@ public Search value(Statistic statistic, double value) {
             return this;
         }
 
+        public Optional<Double> firstValue(){
```
jkschneider: I know you've tried to create this before :). My objection to it is it is only useful for counters and gauges that have a single measurement.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        StandardSession expiredSession = new StandardSession(manager);
+        expiredSession.setId("third");
+        expiredSession.setCreationTime(System.currentTimeMillis()- 10_000);
+        manager.remove(expiredSession, true);
+
+        List<Tag> tags = Tags.zip("metricTag", "val1");
+        TomcatMetrics.monitor(registry, manager, tags);
+
+        assertThat(registry.find("tomcat.sessions.active.max").tags(tags).firstValue()).hasValue(3.0);
```
jkschneider: Alternative to `firstValue`:
`assertThat(registry.find("tomcat.sessions.active.max").tags(tags).gauge().map(Gauge::value)).isEqualTo(3.0);`

checketts:
@jkschneider Sorry I wasn't trying to sneak that in. I meant to ask for the alternative. I've removed it now.

As you can tell I really prefer getting values in the test to get feedback that way instead of `was not present` errors when using `.value(Statistic.Count, 3)`. In fact if you are on board I would be happy to remove that method and update the unit tests to use the `map()` method.

That being said I'm also not completely in love with the `map` approach since it isn't intuitive/easy to discover, but if all the tests use it, that should improve.

jkschneider:
> It looks like Tomcat has many metrics available via JMX
@checketts Are these in addition to what you have wired through `Manager`?

checketts:
The ones available via JMX are ones that would give it parity with the Jetty metrics. See here: https://www.datadoghq.com/blog/monitor-tomcat-metrics/

As in `tomcat.error_count`, `tomcat.request_count`, `tomcat.max_time`, and `tomcat.processing_time` and more

jkschneider:
Yep, seems like that would be useful to have. No overlap with the current gauges.

checketts:
OK I'll pursue the JMX metrics as a separate PR, if that is OK.

PR #261, opened 2017-12-01
Favor asserting directly against values instead of just optional presence

checketts:


micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -258,11 +257,6 @@ public Search tags(String... tags) {
             return tags(zip(tags));
         }
 
-        public Search value(Statistic statistic, double value) {
```
checketts: I removed `value()` since none of our tests are using it anymore. I can add it back if you think it belongs in the public API.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetricsTest.java
```diff
@@ -51,9 +51,9 @@ void cacheExposesMetricsForHitMissAndEviction() throws Exception {
         cache.put("user3", "Third User");
         cache.put("user4", "Fourth User");
 
-        assertThat(registry.find("c.requests").tags("result", "hit").tags(userTags).value(Count, 1.0).meter()).isPresent();
-        assertThat(registry.find("c.requests").tags("result", "miss").tags(userTags).value(Count, 2.0).meter()).isPresent();
-        assertThat(registry.find("c.evictions").tags(userTags).value(Count, 2.0));
+        assertThat(registry.find("c.requests").tags("result", "hit").tags(userTags).functionCounter().map(FunctionCounter::count)).hasValue(1.0);
+        assertThat(registry.find("c.requests").tags("result", "miss").tags(userTags).functionCounter().map(FunctionCounter::count)).hasValue(2.0);
+        assertThat(registry.find("c.evictions").tags(userTags).functionCounter().map(FunctionCounter::count)).hasValue(2.0);
```
checketts: This improved a lot of the assertions since before they just went off `meter()`

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetricsTest.java
```diff
@@ -19,18 +19,32 @@
...
+
+    @BeforeEach
+    void setup(){
+        Config config = new Config();
+        h = Hazelcast.newHazelcastInstance(config);
+    }
+
+    @AfterEach
+    void cleanup(){
+        h.shutdown();
```
checketts: Hopefully shutting down hazelcast will help a little with the memory problems we've seen in the test runs. Probably not much...

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetricsTest.java
```diff
@@ -39,7 +53,7 @@ void cacheMetrics() {
         map.put("key", "value");
         map.get("key");
 
-        assertThat(registry.find("cache.gets").value(Statistic.Count, 1.0).meter()).isPresent();
-        assertThat(registry.find("cache.puts").value(Statistic.Count, 1.0).meter()).isPresent();
+        assertThat(registry.find("cache.gets").functionTimer().map(FunctionTimer::count)).hasValue(1L);
+        assertThat(registry.find("cache.puts").functionTimer().map(FunctionTimer::count)).hasValue(1L);
```
checketts: `FunctionTimer`'s `count` is unusual that is returns a `long` instead of a `double`. I think it is fine that they are different but wanted to point out the difference in case it was unintentional.

jkschneider: Thanks for pointing it out. Unlike a normal counter whose units are unknown, the counters inside of timers always count occurrences. Depending on the unit, it isn't unreasonable to expect that a regular counter might increment by a fractional value. Occurrences are always integral, however.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetricsTest.java
jkschneider: Good call!

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
jkschneider: Yep, it's good to put this one to bed.

jkschneider:
I think this goes part of the way towards making test assertion failures more obvious. Naturally it depends on what part of the assertion fails. `assertThat(registry.find("my.timer").tags("k", "v").gauge().map(Gauge::value)).isEqualTo(n)` still fails with a less-than-obvious cause if either the name or tags are off.

jkschneider:
Still some tests in `micrometer-test` that would need to be changed to compensate for the now-defunct `value` function.

checketts:
OK I'll get those that I missed.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -258,49 +257,55 @@ public Search tags(String... tags) {
...
+            if(meter.isPresent()) {
+                return meter.get();
+            }
+
+            String tagDetail = "";
+            if (!tags.isEmpty()) {
+                tagDetail = " with Tags:[" + tags.stream().map(t -> t.getKey()+":"+t.getValue()).collect( Collectors.joining( "," ) ) + "]";
+            }
+
+            throw new AssertionError("Unable to locate a meter named '"+name+"'"+tagDetail+" of type "+clazz);
```
checketts: Figured I'd take a crack at the 'ideal' approach. I removed the `Optional` and throw a descriptive error.

This also allows removing the `map` method call. I really like it.

micrometer-spring-legacy/src/test/java/io/micrometer/spring/web/servlet/MetricsFilterTest.java
```diff
@@ -94,7 +94,7 @@ public void timedMethod() throws Exception {
 
         assertThat(this.registry.find("http.server.requests")
             .tags("status", "200", "uri", "/api/c1/{id}", "public", "true")
-            .value(Statistic.Count, 1.0).timer()).isPresent();
+            .timer().count()).isEqualTo(1L);
```
checketts: This test it the only one I updated to the new API. Figured I would get your approval @jkschneider  before touching all those tests one more time.

micrometer-spring-legacy/src/test/java/io/micrometer/spring/web/servlet/MetricsFilterTest.java
```diff
@@ -103,33 +103,32 @@ public void subclassedTimedMethod() throws Exception {
...
 
     @Test
     public void untimedMethod() throws Exception {
         this.mvc.perform(get("/api/c1/untimed/10")).andExpect(status().isOk());
 
-        assertThat(this.registry.find("http.server.requests")
-            .tags("uri", "/api/c1/untimed/10").timer()).isEmpty();
+        assertThatThrownBy(() -> this.registry.find("http.server.requests")
+            .tags("uri", "/api/c1/untimed/10").timer())
+            .hasMessageContaining("Unable to locate a meter named 'http.server.requests'");
```
checketts: Here is how we verify that metrics are missing. I could make the exception more strongly typed instead of just a long message if you like.

micrometer-core/src/test/java/io/micrometer/core/instrument/MeterRegistryTest.java
```diff
@@ -45,8 +45,8 @@ void idTransformingMeterFilter() {
         registry.config().meterFilter(MeterFilter.ignoreTags("k1"));
 
         registry.counter("my.counter", "k1", "v1");
-        assertThat(registry.find("my.counter").counter()).isPresent();
-        assertThat(registry.find("my.counter").tags("k1", "v1").counter()).isNotPresent();
+        registry.mustFind("my.counter").counter();
+        assertThat(registry.find("my.counter").tags("k1", "v1").counter()).isNull();
```
checketts: Here you can see `mustFind` throws an `AssertionError` if the metric name/tags/type aren't found in the registry.

The previous `find` still exists. But since it is the less frequently used, it could probably use a different name, so `mustFind` is the easier to find.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetricsTest.java
```diff
@@ -51,9 +51,9 @@ void cacheExposesMetricsForHitMissAndEviction() throws Exception {
         cache.put("user3", "Third User");
         cache.put("user4", "Fourth User");
 
-        assertThat(registry.find("c.requests").tags("result", "hit").tags(userTags).value(Count, 1.0).meter()).isPresent();
-        assertThat(registry.find("c.requests").tags("result", "miss").tags(userTags).value(Count, 2.0).meter()).isPresent();
-        assertThat(registry.find("c.evictions").tags(userTags).value(Count, 2.0));
+        assertThat(registry.mustFind("c.requests").tags("result", "hit").tags(userTags).functionCounter().count()).isEqualTo(1.0);
```
checketts: Here is the new recommended way to assert metric values.

When calling the type (`functionCounter()` in this line) it checks for the name, tags, and type. If it isn't present the error message is: `Unable to locate a meter named 'c.requests' with Tags:[result:hit,userTagKey:userTagValue] of type io.micrometer.core.instrument.FunctionCounter`

So summary of changes:
* `find` and `mustFind` are the ways to look for metrics. 
* The metric type methods (`timer()`, `gauge()`, etc) now return the actual type instead of an `Optional` and throw an exception if the metric is not found (in the case of `mustFind`) or `null` (in the case of `find`)
* All tests are updated to leverage this simpler API.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetricsTest.java
```diff
@@ -41,45 +41,38 @@ void memoryMetrics() {
...
-        final Tag directBufferTag = Tag.of("id", "direct");
-        final Tag mappedBufferTag = Tag.of("id", "mapped");
-        assertMeters(registry.find("jvm.buffer.count").meters(), false, null, directBufferTag,
-                mappedBufferTag);
-        assertMeters(registry.find("jvm.buffer.memory.used").meters(), true, "bytes",
-                directBufferTag, mappedBufferTag);
-        assertMeters(registry.find("jvm.buffer.total.capacity").meters(), false, "bytes",
-                directBufferTag, mappedBufferTag);
+        assertJvmBufferMetrics(registry, "direct");
+        assertJvmBufferMetrics(registry, "mapped");
```
checketts: @mweirauch  I simplified the `JvmMemoryMetricsTest` to assert pretty much the same info, but it is hopefully a little easier to read now.

micrometer-test/src/main/java/io/micrometer/core/tck/GaugeTest.java
```diff
@@ -71,15 +68,14 @@ default void mapSizeGauge(MeterRegistry registry) {
...
+        Gauge g = registry.mustFind("my.gauge").gauge();
         assertThat(g.value()).isEqualTo(1);
     }
 
     @Test
     @DisplayName("gauges that reference an object that is garbage collected report NaN")
     default void garbageCollectedSourceObject(MeterRegistry registry) {
         registry.gauge("my.gauge", emptyList(), (Map) null, Map::size);
-        assertThat(registry.find("my.gauge").value(Statistic.Value, 0).gauge()).isPresent();
+        assertThat(registry.mustFind("my.gauge").gauge().value()).matches(val -> Double.isNaN(val) || val == 0.0);
```
checketts: This test's output changed in an interesting way. StatsD reports a `0` while Prometheus returns a `NaN` Since I wasn't familiar with the `RegistryCompatibilityTest` I wasn't sure how to override the StatsD `GaugeTest` from the others.

checketts:
Rebased against master. Found a couple of faulty tests along the way even.

Once I rebased in the XMX changes, Circle also is performing much better.

jkschneider:
Thanks for going through this meat grinder @checketts. I know it wasn't fun!

checketts:
My pleasure. Now I can focus on the Nullability/Kotlin story. (Unless you would like a hand with any specific issues.)

micrometer-spring-legacy/src/test/java/io/micrometer/spring/async/ThreadPoolTaskExecutorMetricsTest.java
```diff
@@ -15,7 +15,10 @@
  */
 package io.micrometer.spring.async;
 
-import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.MockClock;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
```
checketts: I love that this test went from a wildcard and the compatibility kit went back to a wildcard import.

Are you sure you don't want to adopt a style of `no wildcard imports except for static imports`?

jkschneider: Both follow the code style:

image

The compatibility kit has 5 imports in `io.micrometer.core.instrument` and this test has 4. I guess I understand the way it looks in a Git diff when you go from 5 imports to 4 and see the import exploded, but I just don't find that distracting enough to warrant a custom code style that I have to remember to apply.

It just uses IntelliJ defaults right now. As discussed earlier this week with @mweirauch, I know this is somewhat inconvenient for Eclipse users, but this is one of those "damned if you do, damned if you don't" sort of things.

jkschneider:
Changed `mustFind` to `get` for consistency with other similar APIs around the Spring ecosystem, and because I'm nice to my friends @philwebb and @wilkinsona :)

PR #264, opened 2017-12-03
MetricsFilter does not correctly tag http 500s

fiunchinho:
I noticed that when my application was showing errors, like 500 HTTP status codes, the metrics exposed by micrometer were still tagged with `status=200`.
Adding a try/catch block to my controller endpoint, and make it return an `ResponseEntity` object with an status code of 500 when catching an exception, metrics are tagged as expected with `status=500`. But if I let exceptions to be thrown, then the status tag contains the value 200, even though my site is showing an error 500 status code.

I don't really know how to fix it, so I've pushed this PR trying to demonstrate the error. If you give me some hints, I could give it a shot and try to fix it.

### DIFF
--- a/micrometer-spring-legacy/src/test/java/io/micrometer/spring/web/servlet/WebMvcMetricsIntegrationTest.java
+++ b/micrometer-spring-legacy/src/test/java/io/micrometer/spring/web/servlet/WebMvcMetricsIntegrationTest.java
@@ -81,7 +81,7 @@ public class WebMvcMetricsIntegrationTest {
 
         clock.add(SimpleConfig.DEFAULT_STEP);
         assertThat(this.registry.find("http.server.requests")
-            .tags("exception", "Exception1").value(Statistic.Count, 1.0).timer())
+            .tags("exception", "Exception1", "status", "500").value(Statistic.Count, 1.0).timer())
             .isPresent();
     }
 
@@ -92,7 +92,7 @@ public class WebMvcMetricsIntegrationTest {
 
         clock.add(SimpleConfig.DEFAULT_STEP);
         assertThat(this.registry.find("http.server.requests")
-            .tags("exception", "Exception2").value(Statistic.Count, 1.0).timer())
+            .tags("exception", "Exception2", "status", "500").value(Statistic.Count, 1.0).timer())
             .isPresent();
     }
 


checketts:
@fiunchinho Please add an explanation around what you are trying to demonstrate.

It appears you are trying to indicate that the second test (`rethrownExceptionIsRecordedInMetricTag`) is incorrect because the tag is `200`, but the actual status was `500`.

micrometer-spring-legacy/src/test/java/io/micrometer/spring/web/servlet/WebMvcMetricsIntegrationTest.java
```diff
@@ -92,7 +92,7 @@ public void rethrownExceptionIsRecordedInMetricTag() throws Exception {
 
         clock.add(SimpleConfig.DEFAULT_STEP);
         assertThat(this.registry.find("http.server.requests")
-            .tags("exception", "Exception2").value(Statistic.Count, 1.0).timer())
+            .tags("exception", "Exception2", "status", "500").value(Statistic.Count, 1.0).timer())
```
checketts: This actually has a tag of `status=200`

micrometer-spring-legacy/src/test/java/io/micrometer/spring/web/servlet/WebMvcMetricsIntegrationTest.java
```diff
@@ -81,7 +81,7 @@ public void handledExceptionIsRecordedInMetricTag() throws Exception {
 
         clock.add(SimpleConfig.DEFAULT_STEP);
         assertThat(this.registry.find("http.server.requests")
-            .tags("exception", "Exception1").value(Statistic.Count, 1.0).timer())
+            .tags("exception", "Exception1", "status", "500").value(Statistic.Count, 1.0).timer())
```
checketts: This succeeds for me. So no problem here.

fiunchinho: Correct. I just added it because I believe it needs to be asserted that the right tags are being applied.

micrometer-spring-legacy/src/test/java/io/micrometer/spring/web/servlet/WebMvcMetricsIntegrationTest.java
fiunchinho: Exactly. That's the bug I wanted to demonstrate.

fiunchinho:
I've updated the description.

micrometer-spring-legacy/src/test/java/io/micrometer/spring/web/servlet/WebMvcMetricsIntegrationTest.java
jkschneider: Look to the test name here. The point of the test is to prove that we record a handled exception that still returns a 200. This change is invalid.

fiunchinho:
Thanks!!

somayaj:
Can this be worked? figures the WebMvcMetricsFilter is setting the 500 error on ln: 104
catch (NestedServletException ex) {
			response.setStatus(HttpStatus.INTERNAL_SERVER_ERROR.value());
			record(timingContext, request, response, ex.getCause());
			throw ex;

checketts:
@somayaj Can you please consider opening a new issue for your question?

PR #268, opened 2017-12-04
Improve registering meters with single tag

maciejwalkowiak:
The goal of this PR is to improve registering metrics with single tag.

Currently there are two ways of registering metrics - lets take `Counter` as an example:

```java
meterRegistry.counter("message.received", "tagKey", "tagValue")
```
This method does not give compilation time safety - if number of `String` parameters passed as tags is not even then code will throw exception in runtime.

```java
metrics.counter("message.received", Arrays.asList(Tag.of("tagKey", "tagValue")))
```
This method adds extra boilerplate.

This PR simplifies it to:
```java
metrics.counter("message.received", Tag.of("tagKey", "tagValue"))
```
which guarantees compilation time safety and eliminates boilerplate.

I guess similar methods could/should be added to `Metrics` class.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -345,13 +345,6 @@ public abstract class MeterRegistry {
 
     /**
      * Tracks a monotonically increasing value.
-     */
-    public Counter counter(String name, Tag tag) {
-        return counter(name, Collections.singletonList(tag));
-    }
-
-    /**
-     * Tracks a monotonically increasing value.
      *
      * @param name The base metric name
      * @param tags MUST be an even number of arguments representing key/value pairs of tags.
@@ -369,13 +362,6 @@ public abstract class MeterRegistry {
 
     /**
      * Measures the sample distribution of events.
-     */
-    public DistributionSummary summary(String name, Tag tag) {
-        return summary(name, Arrays.asList(tag));
-    }
-
-    /**
-     * Measures the sample distribution of events.
      *
      * @param name The base metric name
      * @param tags MUST be an even number of arguments representing key/value pairs of tags.
@@ -389,13 +375,6 @@ public abstract class MeterRegistry {
      */
     public Timer timer(String name, Iterable<Tag> tags) {
         return Timer.builder(name).tags(tags).register(this);
-    }
-
-    /**
-     * Measures the time taken for short tasks and the count of these tasks.
-     */
-    public Timer timer(String name, Tag tag) {
-        return timer(name, Collections.singletonList(tag));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


jkschneider:
@maciejwalkowiak I think you have a valid point about their being a gap.

WDYT about adding a signature to `Tags` like `singletonList(String key, String value)` that mirrors Java's `Collections#singletonList` instead? Then we wouldn't have to add additional signatures for all the different meter types?

maciejwalkowiak:
The drawback of having this method in `Tags` is that library users will have to know that this method exist in order to use it - `Cmd+Space` will not suggest it to them. If we go this way I think the existence of such a method should be at least mentioned in the reference guide.

I am fine with both approaches, pick the one you prefer and I will adjust the PR.

maciejwalkowiak:
I understand your concern about growing number of API methods. I was thinking if I am not making another `RestTemplate` kind of class while I was adding these methods. 

The alternative would be a fluent builder (similar to [QueueBuilder](https://github.com/spring-projects/spring-amqp/blob/master/spring-amqp/src/main/java/org/springframework/amqp/core/QueueBuilder.java) from Spring AMQP).

```
metrics.counter("my.counter")
       .tag("key1","value1")
       .tag("key2","value2")
       .build()
```

For *Gauges* it would make even bigger difference:

```
metrics.gauge("my.gauge")
       .tag("key1","value1")
       .obj(...)
       .func(...)  
       .build()
```

This would unfortunately break current API. Since it's already RC I guess it's too late for such changes.

I will go on with `Tags#singletonList`. It can be always improved later in the future.

checketts:
I wish `singletonList` was shorter.

maciejwalkowiak:
Do you mean the method name? I agree. How about `Tags.of(key, value)`?

checketts:
`Tags.of(key, val)` is great! The `Tags` prefix implies it is plural.

maciejwalkowiak:
I'll wait for @jkschneider. I guess he has a decision power here..?

checketts:
>  I guess he has a decision power here..?

Absolute power ;)

jkschneider:
> I guess he has a decision power here..?
Yikes! I guess it's more like I'll have to live with it ;)

> Tags.of(key, val) is great!
Sounds good to me.

jkschneider:
Closed with d47b4df3da5371eb4ef76a674067b28ba522bd08. Thanks for the input @maciejwalkowiak!

jkschneider:
@maciejwalkowiak Also, forgot to mention that there are already fluent builders for all the meter types:

```java
Gauge.builder("my.gauge")
   .tags("k", "v")
   .register(registry);
```

I could see value in going back to each builder and adding a `tag(String key, String value)` signature.

maciejwalkowiak:
Thanks @jkschneider, was a pleasure to contribute!

PR #271, opened 2017-12-06
Fix max in CumulativeTimer.measure()

izeye:
This PR fixes max value in `CumulativeTimer.measure()`.


jkschneider:
Sheesh...that was kind of emabarassing ;)

PR #277, opened 2017-12-08
Improve CloudWatch configuration and sanitize data

osialr:
After trying out the CloudWatch registry, here's some bugfixes and new capabilities.  

Bugfixes:
* Use batch size of 20 if not configured in spring-legacy
* Clamp values to allowable range

Features:
* Support dry-run mode
* ~Support tagging metrics with EC2 metadata~
* ~Support tagging metrics with custom tags~

Let me know what you think.  Thanks!

### DIFF
--- a/implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchConfig.java
+++ b/implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchConfig.java
@@ -43,6 +43,14 @@ public interface CloudWatchConfig extends StepRegistryConfig {
         return v;
     }
 
+    /**
+     * Returns whether to create the CloudWatchMetrics without performing the PUT operation
+     */
+    default boolean dryRun() {
+        String v = get(prefix() + ".dryRun");
+        return Boolean.parseBoolean(v);
+    }
+    
     @Override
     default int batchSize() {
         String v = get(prefix() + ".batchSize");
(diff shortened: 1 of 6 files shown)


osialr:
The tagging part might be duplicative since I see there is `registry.config().commonTags` capability

osialr:
Dropped everything related to tags.  I can get what I need through commonTags

Now it's just the two bugfixes and a dry-run mode

jkschneider:
Was dry run used primarily to test changes to this implementation, or do you envision this being something end users want?

osialr:
As a user, it's nice have a way to see how the library behaves before committing to it.  Every PUT costs a little and metrics can't be deleted once posted in CloudWatch (They autoexpire after some retention period)

So in this case, it's helpful for vetting micrometer configurations when integrating it into an app.   Because my fear when using libraries that talk to a pay-per-access service is that I misconfigure the library and it blasts the service.

checketts:
Micrometer supports multiple implementations at the same time. 

You might consider enabling the Prometheus registry first, which will give you an endpoint for viewing the metrics that will be published. Then you can configure specific metrics to be pushed to Cloudwatch. 

I think a 'CappedMetrics' filter is a great idea that will accept a certain number of metrics then reject the rest.

jkschneider:
> Because my fear when using libraries that talk to a pay-per-access service is that I misconfigure the library and it blasts the service.

@osialr I've thought for some time that we should provide a meta-metric of sorts that describes the cost in terms of total time series that you are going to incur with a particular setup. Some data that is accessible, say, from the root Boot Actuator metrics endpoint.

Something like the total time series (unique combination of name and tags). Something that really highlights at runtime when high cardinality tag values are happening, since that isn't statically verifiable. 

Any other useful things like this that you can imagine?

@checketts `MeterFilter.maxMetrics(n)` is a great idea, and has some prior art (exists in Spectator).

Also, not totally opposed to the dry run, but this discussion is cool, because it's already brought out two new requirements :)

osialr:
After thinking about my use case at now, I could get rid of dry-run.   I can put the configuration of the CloudWatch object into my app and pass in  Mock CloudWatch that logs any calls to Put.  

That leaves just the value clamping and the bug with the default batch size being 1000 instead of 20.   I can rework the PR into two commits since they affect different modules in the project.

jkschneider:
Fixed by #281

PR #279, opened 2017-12-09
Add a Checkstyle module for checking license headers

izeye:
Looking at the commit history, there are a few license header-only commits. This PR adds a Checkstyle module for checking license headers.


jkschneider:
Thanks @izeye, but license header checking is already part of the build via `io.spring.license`, but does a little better than checkstyle because there is a command to auto-remediate.

When there is a license header commit, it's because I or others simply haven't ran the build before committing.

izeye:
@jkschneider Ah, didn't know 😅 Thanks for letting me know it!

PR #282, opened 2017-12-09
Clamp cloudwatch values to allowable range

osialr:
"Although the Value parameter accepts numbers of type Double, CloudWatch
rejects values that are either too small or too large. Values must be in
the range of 8.515920e-109 to 1.174271e+108 (Base 10) or 2e-360 to 2e360
(Base 2). In addition, special values (for example, NaN, +Infinity,
-Infinity) are not supported."

http://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html

### DIFF
--- a/implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchMeterRegistry.java
+++ b/implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchMeterRegistry.java
@@ -158,7 +158,7 @@ public class CloudWatchMeterRegistry extends StepMeterRegistry {
                 .withMetricName(metricName)
                 .withDimensions(toDimensions(tags))
                 .withTimestamp(new Date(wallTime))
-                .withValue(value)
+                .withValue(CloudWatchUtils.clampMetricValue(value))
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchMeterRegistry.java
```diff
@@ -158,11 +161,11 @@ private MetricDatum metricDatum(Meter.Id id, long wallTime, double value) {
                 .withMetricName(metricName)
                 .withDimensions(toDimensions(tags))
                 .withTimestamp(new Date(wallTime))
-                .withValue(value)
+                .withValue(CloudWatchUtils.clampMetricValue(value))
                 .withUnit(toStandardUnit(id.getBaseUnit()));
     }
 
-    private StandardUnit toStandardUnit(String unit) {
+    private static StandardUnit toStandardUnit(String unit) {
```
checketts: Why are these changed to `static`?

osialr:
SonarLint and Intellij were having a field day in that file and I was trying to clean some of it up.  But I just dropped all of those changes.

PR #285, opened 2017-12-10
Better counter naming for allocated and promoted bytes between GCs

mweirauch:
The `allocation.rate` and `promotion.rate` namings suggested a rate
where this is actually the increment (in bytes) of allocated and
promoted bytes between garbage collections.


checketts:
Ayy tests that need to be updated?

mweirauch:
There's no tests to be updated because there is none.

PR #287, opened 2017-12-11
Fix first meter update in AbstractCompositeMeter.remove()

izeye:
`meterHolder` holds the last meter. This PR makes it hold the first one.


jkschneider:
Looks fine, but not sure why? Does this fix an issue or is it an efficiency play?

izeye:
@jkschneider I don't have an issue with it but just noticed it when reading this commit: 3f7ade1d1faeef1a4ec06bce41b1cdbbe58be845

Issue #288, opened 2017-12-11
Counters are incorrectly created as Gauges in Datadog


khelenek:
To reproduce: 
Create a new metric, for example: Metrics.counter("datadog.test", "key", "value").increment(100)

In datadog, go to the metrics explorer. Search for "datadog.test" and confirm the new metric is of type Gauge.

khelenek:
Looks like this might be a limitation of the DataDog API method for sending metrics. Perhaps we should highlight that in the docs? For many that is probably a deal breaker and the statsd route is the correct option. 
https://docs.datadoghq.com/metrictypes/

checketts:
@khelenek What is the deal breaker you are referring to? 

#289 fixes the `count` problem you had noted.

khelenek:
Sorry, I could absolutely be in the wrong here. I thought the docs suggested it just wasn't possible to  create anything but a Gauge using the API. Note that I didn't test using the agents/statsd approach which likely works fine. I'm happy to test your fix again though if you found a solution. Thanks again for your work on this.

jkschneider:
I'm happy to report _more accurate_ metadata to Datadog, at least to the extent it is possible, but I do not understand the practical impact of mislabeling a particular metric. Seems like the display is the same, same functions available, etc. Can you elaborate on what might be a dealbreaker if we couldn't fix this perfectly?

jkschneider:
Ah, I see there was an explanation on #289:

> Yeah it affects the usage, not just the categorization. For a gauge you can't do things like total count for the last hour/day/etc. It treats it more like a sampling of the data at that time.

jkschneider:
Also to verify: base units are added to metadata correctly.

jkschneider:
...and description text.

jkschneider:
Boot 2 change: We added `DatadogConfig#descriptions()`

jkschneider:
@khelenek For this to work, you'll have to set `spring.metrics.export.datadog.applicationKey` in addition to `spring.metrics.export.datadog.apiKey` (or a similar property binding to `DatadogConfig` if you aren't using Spring). Datadog doesn't allow us to adjust metric metadata without that additional key.

PR #289, opened 2017-12-11
Add type data to datadog API calls

checketts:
Addresses #288 

@khelenek I'm not too familiar with [Datadog metric types](https://help.datadoghq.com/hc/en-us/articles/206955236-Metric-types-in-Datadog). 

Could you please hep answer if `count` or `gauge` is sufficient? I'm not certain when a `rate` would be appropriate.

### DIFF
--- a/implementations/micrometer-registry-datadog/build.gradle
+++ b/implementations/micrometer-registry-datadog/build.gradle
@@ -6,4 +6,5 @@ dependencies {
     compile 'org.slf4j:slf4j-api:latest.release'
 
     testCompile project(':micrometer-test')
+    testCompile 'com.fasterxml.jackson.core:jackson-databind:2.9.3'
 }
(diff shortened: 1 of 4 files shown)


checketts:
I also added `unit`, But that isn't clear if that is part of the input json, based on the docs.

khelenek:
Much appreciated the quick turnaround. I added a note to the issue: I'm not entirely sure it's possible to control this through the API; the docs seem to suggest that any metric through the API gets created as a Gauge type. I apologize that I did not include that I was using the HTTP method, not the statsd/agent method. I did not test with the agent. Did you solve using the HTTP method, or did you also discover it affected the agent method?

I am also uncertain on when rate would be more appropriate, but counter vs gauge seems to me the most important distinction.

checketts:
This is for the HTTP API. Since it allows the type to be passed, it seems like this PR will fix it.

What are the tradeoffs right now with metrics appearing as a gauge instead of count? Does it limit which functions can be used with it? Or just exposes it incorrectly in the the explorer?

khelenek:
Yeah it affects the usage, not just the categorization. For a gauge you can't do things like total count for the last hour/day/etc. It treats it more like a sampling of the data at that time.

jkschneider:
Merged with 489f13fabcaaba5c872a4fe874fccb91d33f3d4c

PR #263, opened 2017-12-03
Add Tomcat metrics that are tracked via JMX

checketts:
This adds in several additional Tomcat metrics that are tracked/exposed via JMX.

Cache Metrics (counters)
```
tomcat_cache_hit_total 0.0
tomcat_cache_access_total 0.0
```

Threadpool metrics
```
tomcat_threads_busy{name="http-nio-8080",} 1.0
tomcat_threads_current{name="http-nio-8080",} 10.0
tomcat_threads_config_max{name="http-nio-8080",} 200.0
```

Servlet/Global metrics
```
tomcat_servlet_request_duration_seconds_count{name="default",} 0.0
tomcat_servlet_request_duration_seconds_sum{name="default",} 0.0
tomcat_servlet_request_duration_seconds_count{name="dispatcherServlet",} 7.0
tomcat_servlet_request_duration_seconds_sum{name="dispatcherServlet",} 0.182
tomcat_servlet_error_total{name="default",} 0.0
tomcat_servlet_error_total{name="dispatcherServlet",} 0.0
tomcat_servlet_request_max_seconds{name="default",} 0.0
tomcat_servlet_request_max_seconds{name="dispatcherServlet",} 0.144

tomcat_global_request_max_seconds{name="http-nio-8080",} 0.0
tomcat_global_request_duration_seconds_count{name="http-nio-8080",} 4.0
tomcat_global_request_duration_seconds_sum{name="http-nio-8080",} 7.317
tomcat_global_sent_bytes_total{name="http-nio-8080",} 0.0
tomcat_global_received_bytes_total{name="http-nio-8080",} 0.0
```

The `name` label is the `mbean` `name` property since it is possible to have multiple servlets, and possible to have multiple globals (I'm guessing by having multiple Tomcat instances embedded on separate ports in the same JVM)

I wasn't able to use the JMX Seraver in the unit tests and ended up mocking them out. I also don't love the extra thread that polls to register once it detects the JMX registration has occurred. I'm including it for now, until I can propose an improvement to Micrometer that will allow for a better hook.

FYI @jkschneider @mweirauch

PR #291, opened 2017-12-11
Make `tomcat.sessions.[expired,rejected]` FunctionCounters

mweirauch:
This PR is just here in case #263 doesn't make it into `rc.6`.
It contains a meter type "fix" for Tomcat session metrics which made it into master with #257 which is part of the `rc.6` milestone.

/cc @checketts


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
```diff
@@ -64,11 +61,11 @@ public Context getContext() {
...
 
     @Test
-    void stats() throws IOException, ServletException {
+    void stats() {
         manager.createSession("first");
         manager.createSession("second");
-        Session thirdSession = manager.createSession("third");
-        try{manager.createSession("forth");} catch(TooManyActiveSessionsException exception) {
+        manager.createSession("third");
+        try{manager.createSession("fourth");} catch(TooManyActiveSessionsException exception) {
```
checketts: lol. Thanks for cleaning up the test and fixing the ordering.

PR #292, opened 2017-12-11
Close HTTP connections to prevent connection leak

tjg184:
Noticed InfluxMeterRegistry was not always closing its HTTP connection.  I went ahead and created a pull request for this along with DatadogMeterRegistry and NewRelicMeterRegistry.  If I need to create an Issue first or need clarification on the fix, let me know.


jkschneider:
Good catch ;) Thanks Troy.

PR #297, opened 2017-12-14
Remove AtomicDouble usage from Guava

izeye:
Using `AtomicDouble` looks unintentional as `guava` is `optional`. This PR adds a simplified version of `AtomicDouble` from Guava and replaces `AtomicDouble` from Guava with it.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/cumulative/CumulativeDistributionSummary.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/cumulative/CumulativeDistributionSummary.java
@@ -15,7 +15,6 @@
  */
 package io.micrometer.core.instrument.cumulative;
 
-import com.google.common.util.concurrent.AtomicDouble;
 import io.micrometer.core.instrument.AbstractDistributionSummary;
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.Measurement;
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/cumulative/CumulativeDistributionSummary.java
```diff
@@ -15,16 +15,23 @@
  */
 package io.micrometer.core.instrument.cumulative;
 
-import com.google.common.util.concurrent.AtomicDouble;
 import io.micrometer.core.instrument.AbstractDistributionSummary;
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.Measurement;
 import io.micrometer.core.instrument.Statistic;
 import io.micrometer.core.instrument.histogram.HistogramConfig;
+import io.micrometer.core.instrument.util.AtomicDouble;
```
checketts: Good catch! I think you should use `java.util.concurrent.atomic.DoubleAdder` instead

izeye: @checketts Thanks for the feedback! `DoubleAdder` came to mind first but I didn't figure out how to replace `compareAndSet()` with it. Any suggestion?

izeye: @checketts I updated this PR in e785aaf as you suggested but for `max` I tried another way.

micrometer-core/src/main/java/io/micrometer/core/instrument/cumulative/CumulativeDistributionSummary.java
```diff
@@ -69,12 +80,13 @@ public double max() {
...
     private void updateMax(double amount) {
+        long amountLongBits = Double.doubleToLongBits(amount);
         while (true) {
-            double currentMax = max.get();
+            double currentMax = Double.longBitsToDouble(maxLongBits);
             if (currentMax >= amount) {
                 return;
             }
-            if (max.compareAndSet(currentMax, amount)) {
+            if (maxLongBitsUpdater.compareAndSet(this, maxLongBits, amountLongBits)) {
```
checketts: Wow I learned a bit from this code. Very interesting! Would it be simpler to just use `java.util.concurrent.atomic.AtomicLong` to hold the bits? That would combine the updater and the `volatile` field?

izeye: @checketts Ah, Thanks! `AtomicLong` is much better way. I'll update.

izeye: @checketts I updated this PR in 0f54a0e760895263dc272c2cc82d869ce4c6efb0 as you suggested.

checketts:
Thanks!

PR #298, opened 2017-12-14
Add IllegalImportCheck Checkstyle module against Guava

izeye:
This PR adds `IllegalImportCheck` Checkstyle module against Guava except `cache` package to prevent them from sneaking into codebase again like https://github.com/micrometer-metrics/micrometer/pull/296, https://github.com/micrometer-metrics/micrometer/pull/297.


jkschneider:
Clever use of the regex ;)

PR #299, opened 2017-12-14
Introduce a field for size() in MeterPartition, MetricDatumPartition

izeye:
The value for `size()` in `MeterPartition`, `MetricDatumPartition` doesn't change, so recalculation on `size()` invocation isn't necessary. This PR introduces a field for it and initializes it in its constructor.

And the original `size` field means the size of a partition but it looks confusing with `size()` which means the number of partiions. So this PR also renames it to `partitionSize` to avoid confusion with `size()` method.


jkschneider:
Nice improvement.

Issue #123, opened 2017-09-15
Any plan to add jsr 305 support for null check


linux-china:
Now Spring 5 & Spring Boot 2 both add jsr305 support for null check, could micrometer add this feature. It will be easy to use io.micrometer.core.instrument.Metrics with null check, such as Metrics.counter("xxx.request") will be Nonnull, and it is friendly to Kotlin language and IDE. 

```
  <groupId>com.google.code.findbugs</groupId>
  <artifactId>jsr305</artifactId>
  <version>3.0.2</version>
```

jkschneider:
I'll have to check with the team. I'm pretty sure last time I heard a talk on this that `com.google.code.findbugs:jsr305` breaks the module system in JDK 9 because `Nonnull` is in `javax.annotation`, leading to a split package scenario with the core JDK.

linux-china:
@jkschneider you are right.   jsr305 will break module system in JDK.  I think that Spring 5 has the same problem to implement null check in JDK 9.  How did they fix this problem?  it's so complicated.

checketts:
Interestingly at https://docs.spring.io/spring-framework/docs/5.0.0.BUILD-SNAPSHOT/spring-framework-reference/core.html#null-safety it states: `It is not necessary nor recommanded to have JSR 305 dependency in user project classpath.`

So perhaps we can just use the annotation and leave them as a provided or optional dependency?

jkschneider:
I'll need to probe a bit deeper. It looks like Spring added a `Nonnull`-like annotation that just applies JSR305 `Nonnull` as a meta-annotation. I can't remember the details around this, but I think the runtime allows meta-annotations missing from the classpath to slip by without blowing up? If this is true, then the trick is to use it indirectly, effectively enabling the behavior in the presence of JSR305 and disabling it otherwise.

jkschneider:
At any rate, disappointing half-solution from the Java team: https://stackoverflow.com/a/37911663/1113486

jkschneider:
> It is not necessary nor recommanded to have JSR 305 dependency in user project classpath.

@checketts I guess not recommended because even with the decision to remove the annotations module from the module path by default, the split package problem still means you can only one of JSR305 _or_ the core annotations module (think `@PostConstruct`) at a time.

zhaoxi1988:
I do not agree that we use google findbugs. Java Code Conventions & Spring 5 is more elegant.

[...]

PR #310, opened 2017-12-20
Initial nullable api support (for Kotlin)

checketts:
This adds initial Nullability support.

This PR is a work in progress as I add `Nullable` to any places where we return any nulls.

I'm mimicking the Spring annotation/approach. Any recommendations or tips @sdeleuze?


jkschneider:
Was there a talk on this that I could listen to? Seems like you are safe to use optional dependencies via meta annotations that just add a layer of indirection?

jkschneider:
Addresses #123

micrometer-core/src/main/java/io/micrometer/core/lang/Nullable.java
```diff
@@ -0,0 +1,49 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A common annotation to declare that annotated elements can be {@code null} under
+ * some circumstance. Leverages JSR 305 meta-annotations to indicate nullability in Java
+ * to common tools with JSR 305 support and used by Kotlin to infer nullability of Spring API.
```
jkschneider: Reference to Spring.

micrometer-core/src/main/java/io/micrometer/core/lang/NonNullFields.java
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.annotation.Target;
+import javax.annotation.Nonnull;
+import javax.annotation.meta.TypeQualifierDefault;
+
+/**
+ * A common annotation to declare that fields are to be considered as
+ * non-nullable by default for a given package.
+ *
+ * <p>Leverages JSR-305 meta-annotations to indicate nullability in Java to common
+ * tools with JSR-305 support and used by Kotlin to infer nullability of Spring API.
```
jkschneider: Reference to Spring

micrometer-core/build.gradle
```diff
@@ -8,6 +8,10 @@ apply plugin: 'me.champeau.gradle.jmh'
 apply plugin: 'nebula.optional-base'
 
 dependencies {
+    // JSR-305 only used for non-required meta-annotations
+    compileOnly("com.google.code.findbugs:jsr305:3.0.2")
```
jkschneider: Use optional instead of provided?

checketts:
Unfortunately the talks where it was mentioned weren't exhaustive. They typically had a slide that said 'just add package_info.java' and went on. So that is why I inspected how Spring did it and mimicked them. That is also why I used `compileOnly` instead of making it `optional`, since that is how Spring approached it.

The *Idiomatic Interop* talk got me the most excited: https://youtu.be/zlggmRMadsU?t=25m49s (I tried to link to the slide part (around 25:40).

I believe it was also covered in *The Road to KotlinTown* https://youtu.be/OmwjrVawHqA

checketts:
I removed the Spring references and switched to the optional compile dep. Kotlin seems to still respect the non-null setting that way.

As I'm checking each class to see if they return null, how should I approach nullability in codebase?

For example with the `AbstractMeter` class, is a null is passed in as the ID then it could return a null from `getId()`, should I begin adding in NonNull assertions?

```
public abstract class AbstractMeter implements Meter {
    private final Meter.Id id;

    public AbstractMeter(Id id) {
        this.id = id;
    }

    @Override
    public Id getId() {
        return id;
    }
}
```

jkschneider:
> For example with the AbstractMeter class, if a null is passed in as the ID then it could return a null from getId(), should I begin adding in NonNull assertions in the constructor?

Yes.

micrometer-core/src/main/java/io/micrometer/core/instrument/DistributionSummary.java
```diff
@@ -133,7 +137,7 @@ public Builder publishPercentileHistogram() {
          * systems that have query facilities to do so (e.g. Prometheus' {@code histogram_quantile},
          * Atlas' {@code :percentiles}).
          */
-        public Builder publishPercentileHistogram(Boolean enabled) {
+        public Builder publishPercentileHistogram(boolean enabled) {
```
checketts: Switched to primitives to avoid null checks

micrometer-core/src/main/java/io/micrometer/core/instrument/HistogramSnapshot.java
```diff
@@ -98,11 +101,11 @@ public String toString() {
         buf.append(mean());
         buf.append(", max=");
         buf.append(max);
-        if (percentileValues.length > 0) {
+        if (percentileValues != null && percentileValues.length > 0) {
```
checketts: Uncovered some potential null pointer exceptions

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -111,16 +114,17 @@ protected String getConventionName(Meter.Id id) {
         return id.getConventionName(config().namingConvention());
     }
 
+    /**
+     * @return the registry's base TimeUnit. Must not be null.
+     */
     protected abstract TimeUnit getBaseTimeUnit();
 
     private String getBaseTimeUnitStr() {
-        if (getBaseTimeUnit() == null) {
```
checketts: All registries don't return null. Removed the check to clean up the API

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -597,7 +610,9 @@ public More more() {
         return (M) m;
     }
 
-    private Meter getOrCreateMeter(HistogramConfig config, BiFunction<Id, HistogramConfig, Meter> builder, Id mappedId) {
+    private Meter getOrCreateMeter(@Nullable HistogramConfig config,
+                                   BiFunction<Id, /*Nullable Generic*/ HistogramConfig, Meter> builder,
```
checketts: I'm not certain how to annotate the generic to indicate that `HistogramConfig` could be null. Perhaps we shouldn't allow it to be `null` ever?

micrometer-core/src/main/java/io/micrometer/core/instrument/histogram/HistogramConfig.java
```diff
@@ -57,50 +58,52 @@ public HistogramConfig merge(HistogramConfig parent) {
     }
 
     public boolean isPublishingHistogram() {
-        return percentileHistogram || sla.length > 0;
+        return (percentileHistogram != null && percentileHistogram) || (sla != null && sla.length > 0);
```
checketts: Fixed potential NPE

micrometer-core/src/main/java/io/micrometer/core/instrument/histogram/HistogramConfig.java
```diff
@@ -57,50 +58,52 @@ public HistogramConfig merge(HistogramConfig parent) {
...
-        if(percentileHistogram && supportsAggregablePercentiles) {
+        if(percentileHistogram != null && percentileHistogram && supportsAggregablePercentiles) {
             buckets.addAll(PercentileHistogramBuckets.buckets(this));
             buckets.add(minimumExpectedValue);
             buckets.add(maximumExpectedValue);
         }
 
-        for (long slaBoundary : sla) {
-            buckets.add(slaBoundary);
+        if(sla != null) {
```
checketts: fixed potential NPE

micrometer-core/src/main/java/io/micrometer/core/instrument/util/Assert.java
```diff
@@ -0,0 +1,19 @@
...
+public class Assert {
+
+    /**
+     * Assert that an object is not {@code null}.
+     * <pre class="code">Assert.notNull(object, "{field} must not be null");</pre>
+     * @param object the object to check
+     * @param field the exception message to use if the assertion fails
+     * @throws IllegalArgumentException if the object is {@code null}
+     */
+    public static void notNull(@Nullable Object object, String field) {
```
checketts: This is the assertion check code

micrometer-core/src/main/java/io/micrometer/core/instrument/Gauge.java
```diff
@@ -46,10 +49,13 @@ default Type type() {
...
         private final ToDoubleFunction<T> f;
         private final List<Tag> tags = new ArrayList<>();
-        private String description;
-        private String baseUnit;
+        private @Nullable String description;
+        private @Nullable String baseUnit;
 
         private Builder(String name, T obj, ToDoubleFunction<T> f) {
+            Assert.notNull(name,"name");
+            Assert.notNull(obj,"gauge object");
```
checketts: This change is causing the dropwizard null gauge test to fail. Should `null`s be allowed for `Gauge`s @jkschneider ?

jkschneider: It is incorrect. The obj can become dereferenced and, since our gauge implementations maintain a weak reference to it, thus `null`. I think we want to maintain that behavior.

In general, I'm not keen on adding assert code anywhere in the main source set.

checketts: > I'm not keen on adding assert code anywhere in the main source set.

Sorry I thought you had requested that I add the asserts. I'm fine removing them. Or by `main source set` are you saying to leave them out of the common code paths but leave them in other places like binders?

jkschneider: > should I begin adding in NonNull assertions in the constructor?

Sorry, I thought you meant adding `@NonNull` annotations on constructor parameters.

PR #313, opened 2017-12-22
Do not create a new CompositeMeterRegistry when MetricsProperties.isUseGlobalRegistry() is true

trustin:
Motivation:

When MetricsProperties.isUseGlobalRegistry() is true, a MeterRegistry
exported by MetricsExporter should be added to Metrics.globalRegistry.

Modifications:

- Make MetricsAutoConfiguration.compositeMeterRegistry() return
  Metrics.globalRegistry when MetricsProperties.isUseGlobalRegistry() is
  true
- Make sure a CompositeMeterRegistry is not added to itself

Result:

- Efficiency


trustin:
Thanks for reviewing and merging, @jkschneider !

Issue #315, opened 2017-12-23
Metrics does not format correctly with Swedish locale


thall:
Metrics are not formatted correctly when Swedish locale is used. Problem is that decimal symbol in Sweden is comma and not dot. 

Other locales could have the same problem and the formatting should always use dot as decimal symbol.

Example log output, notice **value=0,2**:
```
2017-12-22 14:43:19 - ERROR - failed to send metrics: {"error":"partial write: unable to parse 'system_load_average_1m,host=dev,metric_type=gauge value=0,2 1513950198938': invalid field format dropped=0"} 
```

PR #316, opened 2017-12-23
Format metrics independent of system locale

thall:
This fix enforces that English locale is used to ensure dot is used as decimal symbol.

Fix for #315


jkschneider:
Thanks @thall!

thall:
No problem @jkschneider :)

PR #318, opened 2017-12-25
Adds Counter.decrement() helper method

nurkiewicz:
I'll add unit tests to this PR if you find this method useful.


jkschneider:
Unfortunately, we cannot do this. Most monitoring systems define counters (whether they be reported as cumulative or rate-normalized) as monotonically increasing values.

The `Gauge` is the conceptual meter type that allows a value to go up and down.

checketts:
@nurkiewicz Depending on your use case you could create a Gauge that is tracking an `AtomicDouble` that you increment and decrement yourself, or even a `LongTaskTimer`

Issue #339, opened 2018-01-12
Support "TimerContext" in the Timer interface


sunnychanwork:
In DropWizard they have a style for timer usage like this:

TimerContext tc = Timer.start();
/*Do some work*/
tc.stop() (record the time) 

This style is useful for cases like timing a JTA transaction where a transaction is start at transaction object creation, and we stop timing when someone call commit() or rollback(). 

see https://docs.oracle.com/javaee/7/api/javax/transaction/Transaction.html

jkschneider:
Thanks for the feedback @sunnychanwork

PR #340, opened 2018-01-12
Add TimerContext that record a duration after the fact.

checketts:
TimerContext::record returns the duration as nanos to be used in cases like logging

Address #339 

Usage:

```
TimerContext timer = registry.timer("myTimer").time();
// do work
t.record();
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
@@ -72,6 +72,11 @@ public abstract class AbstractTimer extends AbstractMeter implements Timer {
     }
 
     @Override
+    public TimerContext time() {
+        return new TimerContext(clock, this);
+    }
+
+    @Override
     public final void record(long amount, TimeUnit unit) {
         if(amount >= 0) {
             histogram.recordLong(TimeUnit.NANOSECONDS.convert(amount, unit));
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
@@ -108,6 +108,13 @@ public interface Timer extends Meter {
     }
 
     /**
+     * Begin timing an operation.
+     *
+     * @return TimerContext
+     */
+    TimerContext time();
+
+    /**
      * The maximum time of a single event.
      */
     double max(TimeUnit unit);
@@ -135,6 +142,30 @@ public interface Timer extends Meter {
     @Override
     default Type type() {
         return Type.Timer;
+    }
+
+
+    class TimerContext {
+        private final long startTime;
+        private final Clock clock;
+        private final Timer timer;
+
+        public TimerContext(Clock clock, Timer timer) {
+            this.clock = clock;
+            this.timer = timer;
+            this.startTime = clock.monotonicTime();
+        }
+
+        /**
+         * Records the duration of the operation
+         *
+         * @return The duration that was record in nanoseconds
+         */
+        public long record(){
+            long durationNs = clock.monotonicTime() - startTime;
+            timer.record(durationNs, TimeUnit.NANOSECONDS);
+            return durationNs;
+        }
     }
 
     static Builder builder(String name) {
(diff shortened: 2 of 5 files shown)


checketts:
Test pass locally. CircleCI is limiting the memory and that fails the build

jkschneider:
This changed again:

Looks something like this now:

```
Timer.Sample sample = Timer.Sample.start(registry);
sample.stop(registry.timer("my.timer", "response", response.getStatus());

LongTaskTimer.Sample sample = registry.more().longTaskTimer("my.long.task").start();
sample.stop();
```

For short timers, we require that you provide the timer meter at the end. The idea is that many cases that will make use of this construct also want to dynamically commute tags. Pushing tag value determination until `stop` achieves that.

Long task timers now always return a `Sample` instead of a task id when they are started. This is especially useful when all you want to do later is indicate when the task is complete. Previously when tracking multiple long task timers for a given operation, you had to manually hold both the long task timer _and_ the task id on a per timer basis.

checketts:
What is the benefit of including the registry?

Over the weekend I realized we could simplify it to something by creating a simple `StopWatch`:

```
StopWatch watch = StopWatch.start()

registry.timer("my.timer", "response", response.getStatus()).record(watch.stop())
```

We could just make `watch.stop()` return a duration. That way it isn't a different way to recording or calling a timer, but allows the flexibility of late tag evaluation. It doesn't address the `LongTaskTimer`, so keeping that separate is fine in my opinion, but a class name like `LongTask`instead of `Sample` might be a bit cleaner there.

jkschneider:
> What is the benefit of including the registry?

We need access to the clock. There is an overload for starting with a `Clock` instance or the `MeterRegistry`, since in most cases you would just call `Timer.Sample.start(registry.config().clock());` anyway.

jkschneider:
Whether it's:

```registry.timer("my.timer", "response", response.getStatus()).record(watch.stop())```

or

```watch.stop(registry.timer("my.timer", "response", response.getStatus()))```

seems kind of like six of one, half dozen of another. The latter is slightly shorter, and I guess is slightly more discoverable (once you've discovered that this type exists in the first place).

**Comment by jkschneider** [comment]
It would be awesome if we could name the construct the same in `Timer` and `LongTaskTimer`, even if it doesn't function exactly the same way. Once you know of one, you kind of naturally look for the other. If one is called `StopWatch` and the other `LongTask`, I think that reduces co-discoverability?

**Comment by checketts** [comment]
The `StopWatch` name is more intended as a simple general purpose timer. Since `Timer::record()` can take a `Duration`.  For discoverability we could create a subclass of `StopWatchWithTimerContext` (similar to the original PR implementation). 

`Timer.Sample.start(registry.config().clock()) ` is not intuitive at all, especially when the name `Sample` usually reflects a sampling of data (a result). If we want to keep it, perhaps `Timing` is clearer.

The construct only needs to be discoverable if it isn't returned by a Meter's method. Since `LongTastTimer::start` returns the the type it doesn't need to be found by just looking at types. 

Lastly the main reason I recommend this approach is to leverage the existing `record` method that already exists on `Timer`, instead of adding `watch.stop(timer)` being a new way to record results.

**Comment by jkschneider** [comment]
I don't think we can use `StopWatch` as there is already:

* `org.springframework.util.StopWatch` (uses `System#currentTimeMillis`, which makes me shudder)
* `org.junit.rules.Stopwatch`
* `com.google.common.base.Stopwatch`
* `org.apache.commons.lang.time.StopWatch`
* `org.apache.commons.lang3.time.StopWatch`
* Others in Jersey, hsqldb, etc.

It's an import nightmare.

`Timing` is pretty disambiguated, and so I don't really care whether we use `Sample` or `Timing`. With 661e540adc0184a4fa8d5303fce50066ef62b49b we have the static `Timer#start(clock or registry)`, so the type will be generally be written on the left side by the IDE anyway.

**Comment by checketts** [comment]
I'm fine with however you want to take it.

```
org.springframework.util.StopWatch stopwatch = //use stopwatch

//do work

registry.timer("mytimer","my","tag").record(stopwatch.stop(), TimeUnit.MILLISECONDS)
```

Is totally fine by me if the use wants a stopwatch. (I'm just guessing how they would use it)

PR #342, opened 2018-01-12
Add xmx limit for circleci builds

checketts:
Docs: https://circleci.com/docs/1.0/oom/#out-of-memory-errors-in-android-builds

Hopefully this will keep circle from killing the builds sporatically


jkschneider:
Thanks @checketts and @mweirauch for looking at this. We were just kind of limping along weren't we? Will be great to hopefully stop seeing those problems.

PR #347, opened 2018-01-16
Modifiable Graphite meter registry

eutkin:
Added the ability to change the naming convension (I think it is necessary) and add a prefix for the names of metrics

### DIFF
--- a/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteMeterRegistry.java
+++ b/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteMeterRegistry.java
@@ -20,11 +20,14 @@ import com.codahale.metrics.graphite.GraphiteReporter;
 import com.codahale.metrics.graphite.GraphiteSender;
 import com.codahale.metrics.graphite.PickledGraphite;
 import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry;
 import io.micrometer.core.instrument.util.HierarchicalNameMapper;
 
 import java.net.InetSocketAddress;
 import java.util.concurrent.TimeUnit;
+
+import static java.util.Objects.requireNonNull;
 
 public class GraphiteMeterRegistry extends DropwizardMeterRegistry {
 
@@ -58,6 +61,7 @@ public class GraphiteMeterRegistry extends DropwizardMeterRegistry {
         this.reporter = GraphiteReporter.forRegistry(getDropwizardRegistry())
                 .convertRatesTo(config.rateUnits())
                 .convertDurationsTo(config.durationUnits())
+                .prefixedWith(config.metricPrefix())
                 .build(sender);
 
         if(config.enabled())
@@ -70,5 +74,10 @@ public class GraphiteMeterRegistry extends DropwizardMeterRegistry {
 
     public void start() {
         this.reporter.start(config.step().getSeconds(), TimeUnit.SECONDS);
+    }
+
+    public GraphiteMeterRegistry setNamingConvention(NamingConvention namingConvention) {
+        this.config().namingConvention(requireNonNull(namingConvention));
+        return this;
     }
 }
--- a/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteConfig.java
+++ b/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteConfig.java
@@ -31,10 +31,8 @@ public interface GraphiteConfig extends DropwizardConfig {
     /**
      * Get the value associated with a key.
      *
-     * @param k
-     *     Key to lookup in the config.
-     * @return
-     *     Value for the key or null if no key is present.
+     * @param k Key to lookup in the config.
+     * @return Value for the key or null if no key is present.
      */
     String get(String k);
 
@@ -74,16 +72,23 @@ public interface GraphiteConfig extends DropwizardConfig {
     }
 
     /**
+     * Prefix all metric names with the given string.
+     */
+    default String metricPrefix() {
+        return get(prefix() + ".metric-prefix");
+    }
+
+    /**
      * Protocol to use while shipping data to Graphite.
      */
     default GraphiteProtocol protocol() {
         String v = get(prefix() + ".protocol");
 
-        if(v == null)
+        if (v == null)
             return GraphiteProtocol.Pickled;
 
         for (GraphiteProtocol flavor : GraphiteProtocol.values()) {
-            if(flavor.toString().equalsIgnoreCase(v))
+            if (flavor.toString().equalsIgnoreCase(v))
                 return flavor;
         }
 


implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteMeterRegistry.java
```diff
@@ -71,4 +75,9 @@ public void stop() {
     public void start() {
         this.reporter.start(config.step().getSeconds(), TimeUnit.SECONDS);
     }
+
+    public GraphiteMeterRegistry setNamingConvention(NamingConvention namingConvention) {
```
checketts: I don't know that we expose the underlying config on any other registries in this way. Is the benefit to just make it easier to find config options?

I believe the config classes were originally created as an effort to make it clear where configuration was set on registries and what was used for metrics.

eutkin: spring boot does not allow you to change the naming convention. The proposed implementation does not meet the requirements in my company (we use dot convention instead of camelCase for storage in graphite). I hope I understood the question correctly?

checketts: You can modify the meter registry configuration by creating a `MeterConfigurer` bean

```
@Bean
public MeterConfigurer graphiteConfigurer(MeterRegistry reg) {
  if(reg instanceof GraphiteMeterRegistry) {
    reg.config().namingConvention(new namingConvention("myprefix"));
  }
}
```

jkschneider:
@checketts Provided the mechanism we intend for folks to use when they want to change the naming convention on an individual registry:

```
@Bean
public MeterConfigurer graphiteConfigurer(MeterRegistry reg) {
  if(reg instanceof GraphiteMeterRegistry) {
    reg.config().namingConvention(new namingConvention("myprefix"));
  }
}
```

implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteMeterRegistry.java
eutkin: IMHO, the use of "instance of" does not look very elegant. The example with MeterConfigurer is useful, but it seems to me that this method is not very obvious, but it's better to make the prefix in the names of metrics more friendly to programmers. To be honest, you did not convince me. I agree that this example will work, but it looks like a kludge. sorry for my English

eutkin: MeterConfigurer? I not found this class in my spring-boot-2.0.0.m7 application

jkschneider: I agree that a type-safe `MeterConfigurer` would be awesome. Not totally sure it's possible in Spring, but would be awesome.

Prefixing as a thing is a very graphite-specific thing unfortunately, so maybe we could get on board with a `GraphiteConfig#prefix()`. On the other hand, we haven't done this historically because the prefix folks add to Graphite metrics tend to be something like application name which really is a _tag_ in a dimensional system. So our advice has been to add a common tag with (following this example) application name, and then override the naming convention to place that tag value in a prefix position in the name. By adding a common tag, the metric is instantly portable to more capable dimensional monitoring systems for which nobody really does prefixing.

That said, we get this request _a lot_ for Graphite users. So much so, that I think it's safe to say the majority of Graphite users want some sort of tag-like prefix on their metrics. So perhaps a `GraphiteConfig#tagAsPrefix(tagKey)` would be even better. Guide the user to adding a common tag that is portable, and imbue the default Graphite naming convention with a knowledge of how to turn the specified tag key into a prefix when it is present.

checketts: I think @eutkin wants to provide his own `GraphiteMeterRegistry` that is already preconfigured with the naming convention.

```
@Bean
public GraphiteMeterRegistry graphiteRegistry(Clock clock, @Value("${myprefix}") String prefix){
  GraphiteMeterRegistry reg = new GraphiteMeterRegistry(clock)
  reg.setPrefix(prefix)
  return reg;
}
```

That, in fact, seems like more typical Spring Boot autoconfig. Does it make sense to move the `config` class up into the main class? Or are those configs kept separate for the benefit for `@ConfigurationProperties`? I seem to think other boot classes add `@ConfigurationProperties` to the bean in question, instead of it being separate.

eutkin: imho, the whole problem is that we can not use our own naming convention and configure the reporter. Any way to get a link to graphiteMeterRegistry and change the state of this link is a pretty bad decision. Another problem in my opinion is a violation of DI in the class GraphiteMeterRegistry, because it creates its own dependency GraphiteReporter (I believe that the reason in the method is "getDropwizardRegistry ()", that is, it is impossible to get a reference to the internal MetricRegistry). DropwizardMeterRegistry also does not allow you to change your internal meterRegistry, which is not very good for the heirs (as a solution, in the DropwizardMeterRegistry instead of the field reference, use the getter, which you can override). I believe that the problem is somewhat deeper than I thought was originally, and it was a good opportunity to use the external MetricRegistry in the DropwizardMeterRegistry and the heirs, according to the principle of DI

jkschneider: I'm having a little trouble following all of that, but I take the point that we should provide an alternate constructor that allows you to provide your own `GraphiteReporter`, and that is I think fair.

eutkin: Maybe so? https://gist.github.com/eutkin/ecfe6f50b7a50a7792d04d83ec09be68 Added setter for MetricRegistry in DropwizardMeterRegistry and other setters in GraphiteMeterRegistry. Not very nice, but it does not break compatibility with spring boot (but there I would make changes, for example, announcements bean MetricExporter for GraphiteMeterRegistry)

jkschneider:
@eutkin Based on your input, we've now got a few more options:

1. A `GraphiteMeterRegistry` constructor that allows you to provide your own `GraphiteReporter` and `MetricRegistry`, if you choose. We don't think you should have to go to this level of effort to accomplish the common requirement of adding a prefix for Graphite metrics, so...

2. A new `GraphiteConfig#tagsAsPrefix()` option (paired with a new default `GraphiteHierarchicalNameMapper` that consumes it):

So you can configure:

```
management.metrics.export.graphite.tagsAsPrefix=application,host
```

Then, add a common tag to your metrics like you normally would:

```
registry.config().commonTags("application", "MYAPP", "host", "MYHOST");
```

Then when you create a new metric:

```
registry.counter("my.counter", "othertag", "othervalue");
```

It appears in Graphite like so:

```
MYAPP.MYHOST.myCounter.othertag.othervalue
```

jkschneider:
Like @checketts said earlier, there is no need for a setter for naming convention, because there already is one: `registry.config().namingConvention(YOUR_CONVENTION_HERE)`

PR #357, opened 2018-01-18
Support for system and process CPU usage

mweirauch:
micrometer-load-and-usage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,5 @@ out/
 .classpath
 .project
 .settings/
+.sts4-cache/
 bin/
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java
@@ -19,7 +19,11 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import org.junit.jupiter.api.Test;
 
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
+
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 class ProcessorMetricsTest {
     @Test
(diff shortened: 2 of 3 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java
```diff
@@ -32,7 +36,38 @@ void cpuMetrics() {
...
+        new ProcessorMetrics().bindTo(registry);
+
+        assertThat(registry.mustFind("system.cpu.usage").gauge().value()).isGreaterThanOrEqualTo(0);
+        assertThat(registry.mustFind("process.cpu.usage").gauge().value()).isGreaterThanOrEqualTo(0);
+    }
+
+    @Test
+    void openJ9CpuMetrics() {
+        OperatingSystemMXBean osBean = ManagementFactory.getOperatingSystemMXBean();
+        assumeTrue(osBean.getClass().getName().equals("com.ibm.lang.management.internal.UnixExtendedOperatingSystem"));
```
checketts: Change to `assertThat(osBean.getClass().getName()).isEqualTo("com.ibm.lang.management.internal.UnixExtendedOperatingSystem");`

jkschneider:
@mweirauch Do you know of a way to break down CPU usage by thread? That would be amazing.

checketts:
What OS/JVM support this value?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java
mweirauch: JUnits assumeTrue() will skip any further processing and ignore the test: http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/Assumptions.html#assumeTrue-boolean-

The idea is to not let the test fail, like with this proposal :)

mweirauch:
> Do you know of a way to break down CPU usage by thread? That would be amazing. Somehow I think htop is capable of that.

`ThreadMXBean` and `ThreadInfo` came to my mind. Here's an example I quickly looked for: https://stackoverflow.com/questions/29931391/obtaining-cpu-thread-usage-in-java

Doable but might also be expensive from a JVM perspective. Also, some JVM management threads might be hogging your CPU (allthough unlikely; e.g. the CI Compiler threads) which I'd assume aren't listed there. Reading the `ThreadMXBean`s JavaDoc gives you a feel if it is worthwhile exploring this further.

brharrington:
FWIW servo has a utility for computing the thread cpu usage you could look at:

https://github.com/Netflix/servo/blob/master/servo-core/src/main/java/com/netflix/servo/util/ThreadCpuStats.java

It gets used internally on the app admin to show a summary of cpu usage by thread, for example:

screen shot 2018-01-18 at 3 21 25 pm

Most of our apps have too many threads so we don't create metrics as tag values. In some cases teams that are more disciplined about thread naming have done so and grouped the threads. So instead of reporting a metric for 500+ tomcat threads it would create an aggregate for all tomcat threads. It allows you to see which group of threads is using the most cpu. It wasn't something we found useful enough to push for better standardization and adoption on though.

.gitignore
```diff
@@ -11,4 +11,5 @@ out/
 .classpath
 .project
 .settings/
+.sts4-cache/
```
jkschneider: Perhaps you should add this to your global .gitignore?

jkschneider:
Merged with 39c56bb44832f67b37970c89b12c80e8727e177d

.gitignore
mweirauch: It's from the upcoming SpringToolsSuite4. If everybody should have IDE specifc gitignores global, why do we see so many IDE specific project ignores everywhere ... and here.

wuguanyu:
This cpu usage is correct ?

PR #359, opened 2018-01-19
Null tag values break Prometheus export

checketts:
Unfortunately null tag values can be created and they break the Prometheus export. The behavior is pretty ugly since the code that creates the null doesn't explode, only the reporting endpoint blows up.

Here is one solution which assumes that null values are intended and the user can declare what they want `null` to be represented in the output tag.

A simpler approach would be to deny nulls and add an `assert.noNull()` in `ImmutableTag`'s constructor.


jkschneider:
I think the simpler approach makes more sense?

checketts:
Sounds good to me, the tradeoff is the option in this PR avoids runtime exceptions and converts them into valid tags. I'll submit a new PR (and also fix the failing tests).

checketts:
Nevermind. Looks like you fixed it in master already 👍

PR #360, opened 2018-01-19
Add configuration for domain and object name factory into JmxConfig

ITman1:
I am currently missing possibility to specify `domain` and `ObjectNameFactory` for `JmxReporter` inside `JmxMeterRegistry`. Do you think we can add it into `JmxConfig`?


jkschneider:
Yep, no problem.

jkschneider:
Since we want to keep our config interfaces to things that are serializable, I've removed the object name factory config prop. However, also added a new constructor for `JmxMeterRegistry` such that you can provide your own `JmxReporter` as the last argument, and thereby configure your own object name factory on that if you wish.

Hopefully that isn't too much trouble!

PR #362, opened 2018-01-21
Simplify record(Supplier) in LongTaskTimer

szpak:
There is no need to declare throwing checked exception. It seems to be copied from `recordCallable()` placed above. The same method in `Timer` doesn't do that.


jkschneider:
Good catch. Thanks @szpak!

PR #377, opened 2018-01-29
Wavefront - beta push

howardyoo:
New configuration : mode between proxy and direct mode.
apitoken and apihost to define api end point when direct mode is used (default is proxy mode).

IMPORTANT NOTE: Wavefront's direct mode uses JSON as its message payload, and each metric being reported looks like this:
`
{
  "aws.rds.ext.metric1": {
    "value": 2,
    "tags": {
      "accountId": "293993588345",
      "DatabaseClass": "db.m3.medium"
    }
  },
  "aws.rds.ext.metric2": {
    "value": 1000,
    "tags": {
      "accountId": "293993588345",
      "DatabaseClass": "db.m3.medium"
    }
  }
}
`
There exists a potential problem when using direct load, that if the name of the metrics are identical, it would cause JSON parsing error, as each object name should be unique. However, I've seen there are metrics which gets reported together that has same name, but with different tag values. Various approaches can be made to solve that problem, but I would like to have it as an ongoing discussion, as I also would like to hear from others' opinions on that.

### DIFF
--- a/implementations/micrometer-registry-wavefront/build.gradle
+++ b/implementations/micrometer-registry-wavefront/build.gradle
@@ -2,6 +2,7 @@ apply plugin: 'org.junit.platform.gradle.plugin'
 
 dependencies {
     compile project(':micrometer-core')
+    compile 'com.fasterxml.jackson.core:jackson-databind:latest.release'
 
     testCompile project(':micrometer-test')
 }
--- a/implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
+++ b/implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
@@ -25,6 +25,9 @@ import java.time.Duration;
 public interface WavefrontConfig extends StepRegistryConfig {
     WavefrontConfig DEFAULT = k -> null;
 
+    // mode to specify whether to use wavefront proxy or submit directly
+    public enum mode {proxy, direct};
+
     @Override
     default Duration step() {
         String v = get(prefix() + ".step");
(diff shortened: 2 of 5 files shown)


checketts:
That is a valid problem. There will be many cases of metric names being repeated, only differentiated by tags.

I'm interested to hear your ideas on how to work around the problem. The most straightforward seems like the JSON format would need to change (use an array instead of a map), but that would require Wavefront itself to allow for that, which I doubt would be possible.

howardyoo:
Right,
I also need to discuss this w my engineering team to look for options, will get back to you!

howardyoo:
Hi, can someone please fix this checks being failed?

checketts:
Here is the error. You should be able to reproduce it locally using `./gradlew clean test`:

```/home/ubuntu/micrometer/samples/micrometer-samples-boot2/src/main/java/io/micrometer/boot2/samples/SimpleSample.java:20: error: cannot find symbol
import org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryConfigurer;
                                                             ^
  symbol:   class MeterRegistryConfigurer
  location: package org.springframework.boot.actuate.autoconfigure.metrics
/home/ubuntu/micrometer/samples/micrometer-samples-boot2/src/main/java/io/micrometer/boot2/samples/SimpleSample.java:34: error: cannot find symbol
    public MeterRegistryConfigurer configurer() {
           ^
  symbol:   class MeterRegistryConfigurer
  location: class SimpleSample
2 errors
:micrometer-samples-boot2:compileJava FAILED```

howardyoo:
Ok, it seems the class MeterRegistryConfigurer class should have been removed from the code. Removed the code and the checks have now passed.

jkschneider:
Merged in 8cc63a3025cfc311927d97522bde78ad8a570858

PR #385, opened 2018-01-31
Initialize CompositeMeterRegistry with Clock.SYSTEM (fixes #384)

### LINKED ISSUE #384: Autoconfigured CompositeMeterRegistry has null clock
I've updated to rc8 micrometer-spring-legacy.  I'm now registering multiple `MeterRegistry` beans directly instead of through `MeterExporter`.  The `CompositeMeterRegistry` that holds all of my individual `MeterRegistry`s is created with a null clock which results in an NPE.

michael-barker:


jkschneider:
Fixed in 72066831401a4b74f704fd29b94e3c76be36d770

PR #386, opened 2018-01-31
Prometheus actuator endpoint produces application/json

marcoschulte:
The prometheus actuator endpoint is registered by spring as producing application/json. Therefore it throws HTTP 406 if "Accept: text/plain" is added as header, even though it eventually produces text/plain.

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusExportConfiguration.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusExportConfiguration.java
@@ -69,5 +69,10 @@ public class PrometheusExportConfiguration {
             CollectorRegistry collectorRegistry) {
             return new PrometheusScrapeEndpoint(collectorRegistry);
         }
+
+        @Bean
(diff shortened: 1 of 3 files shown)


jkschneider:
Is this for Boot 1.5.x?

marcoschulte:
Yes, therefore in the legacy module.

jkschneider:
@marcoschulte I guess we should document this, but it is simply a known limitation of Boot 1.5. In Boot 2, we can set the correct header in actuator endpoints. My experience with the Boot 1.5 endpoint is that Prometheus scrapes it successfully in spite of the wrong content type.

Do you have something in the middle that is being pickier? Or a specific version of Prometheus that fails?

marcoschulte:
The problem is Spring itself. If the request to the prometheus endpoint is send with "Accept: text/plain", Spring answers with HTTP 406 as it assumes the endpoint will produce application/json. We have a scraper that sets the Accept header and always fails, therefore we noticed.

I'm not sure whether you noticed I created a pull request? The problem is already solved, you can override what actuator endpoints produce in Boot 1.5.

jkschneider:
Ah, shame on me. Somehow I totally missed that this was a PR and not an issue.

jkschneider:
Merged in c8b798728175d932603934fca2c9ebad6a5b1d22

marcoschulte:
No worries, glad I could help. And thanks for the amazing work you guys do!

PR #387, opened 2018-01-31
Added auto-configuration for AsyncRestTemplate beans

sergei-ivanov:
This adds metrics instrumentation to all `AsyncRestTemplate` bean instances registered in the spring context.
Previously, outbound asynchronous REST calls would not be automatically covered by metrics.

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/web/client/RestTemplateMetricsConfiguration.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/web/client/RestTemplateMetricsConfiguration.java
@@ -27,17 +27,22 @@ import io.micrometer.spring.web.client.MetricsRestTemplateCustomizer;
 import io.micrometer.spring.web.client.RestTemplateExchangeTagsProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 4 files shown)


jkschneider:
Cool!

sergei-ivanov:
PR Notes:
Unfortunately, `AsyncRestTemplate` has always been an unloved stepchild in Spring and Spring Boot. It does not enjoy the same level of support as the standard `RestTemplate`: there is no builder, no customisers, no default test instance. This PR simply backports some stuff I had to implement manually while dabbling with Prometheus metrics in one of the projects. Initially it was a completely separate set of configuration and customiser classes, but apart from implementing the `Async...` interfaces, it was 80-90% copy-paste of the original code, so I decided to simply add the missing bits to the existing configuration code (there should be no classpath issues, as `RestTemplate` and `AsyncRestTemplate` have been bundled together for a few Spring releases). I hope someone finds it useful.

jkschneider:
@sergei-ivanov I've merged this to the `micrometer-spring-legacy` backport, but will not be carrying it forward to Boot 2 because they've already deprecated `AsyncRestTemplate`. So I think it is useful for you to have this instrumentation in a version of Boot _prior_ to the deprecation, and your work was well done.

Here is some detail from Rossen Stoyanchev on why the deprecation happened:

> It's nice to see a series or articles covering JAX-RS, Vert.x, and Spring 5 side by side in one issue of an Oracle Java magazine:
http://www.javamagazine.mozaicreader.com/JanFeb2018#&pageSet=15&page=0

> For anyone wondering why we deprecated the AsyncRestTemplate, just take a look at the JAX-RS style of scatter-gather, remote calls with CompletionStage. The result is incredibly verbose (see pages 27-28) and is also limiting in terms of concurrency (must get full List of locations before firing nested operations for temperature for any of the locations). And this is CompletableFuture, never mind working with the plain ListenableFuture that the AsyncRestTemplate returns. I had actually not seen any code taking CompletionStage to the task of composing asynchronous logic, and I must say it looks much worse than I thought it might.

PR #394, opened 2018-02-03
add autoconfiguration for prometheus push-gateway

davidkarlsen:
Support for auto-configuration of Prometheus pushgateway. 
Please review it.


mweirauch:
Some questions:
* Are all configuration properties (including `fixedRateInMs`) auto-completable in the IDE? (Unfortunately my Eclipse left me out in the rain attracting the new properties...)
* How are you dealing with multiple instances of the same job?
  We are using the Pushgateway to send metrics from multiple instances of the same application. In our internal implementation I am sending the `${spring.application.name}` as the `job` with additional `groupingKey` `instance=<hostname>:<server.port>` where `hostname` is retrieved via `InetAddress.getLocalHost().getHostName()` so that multiple instance of the same app could even run on the same host. (Allthough we don't do this. The value of `instance` is also overridable via config props.) On the Prometheus side I am using the recommended label replacement settings, so that job and instance are kept instead of being used from the scraped Pushgateway instance itself. (`honor_labels: true`)

davidkarlsen:
I am unsure how IDEs do their auto-completion.

I can add support for adding groupingkeys - but leave it to the user to specify the actual k/v's (e.g. not dictate instance etc) - I think that makes sense.

mweirauch:
A note regarding `hostname`. It works for me on Linux but some colleague had problems on his mac where the ENV var was not set. So I solved that programmatically like mentioned above. Perhaps a little `InstanceProvider` could be used programmatically where required with a fallback to a `instance` property where this is enough?

davidkarlsen:
Added support for grouping keys. I don't think we should add instance ootb - as this is up to the user. It could be two instances on the same box, and then they should configure it accordingly.
Also note that the pushgateway will add the originating IP as instance automatically, if not explicitly given: https://github.com/prometheus/pushgateway

jkschneider:
Almost done working on this. Can someone explain why you would want to delete the job and grouping key when the app shuts down?

jkschneider:
Merged in a033f561. If we need to add the delete logic back, we can do that in a follow-up commit.

davidkarlsen:
The reason for the delete is: https://prometheus.io/docs/practices/pushing/ - see bulletpoint 3. I think it makes sense to at least add it back with an option, deleteMetricsAtShutdown, default false?

mweirauch:
I was just about to, but you beat me to it @davidkarlsen :)

If you can name any reason to not drop metrics on shutdown due to the points listed in the documentation (and also my personal experience) then have it default to off. I can't find any to keep them in there. (I am periodically - no automation yet - manually dropping stale metric groups from our pushgateway when apps didn't shut down orderly.)

davidkarlsen:
@mweirauch The only reason I can think of is that they get deleted before scraped one last time. For short lasting jobs (and infrequent scrapes) you could risk the metrics being deleted before scraped at all. Also if you have a job with stable identifiers it should not matter too much. If they have identifiers that change it matters more, as your pg fills up with garbage. Maybe it should be true by default - but definitively the option should be there - so the job pushing the metrics can also clean up after itself.

PR #395, opened 2018-02-03
aspectJ aspect for monitoring @Timed types and methods

### LINKED ISSUE #361: Support for @Timed in any Spring-managed bean
[As of v1.0.0-rc.7](https://github.com/micrometer-metrics/micrometer/issues/24#issuecomment-304378205) `@Timed` put on a method of an arbitrary Spring-managed bean is silently ignored. It would be useful to be able to generate metrics for any kind of beans (e.g. `Service`s with non-`RestTemplate` remote system calls) without writing a custom AOP logic.

davidkarlsen:
Fixes #361 .
Please review.
Should I maybe add some auto-configuration for it too?

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/aop/MicrometerAspect.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/aop/MicrometerAspect.java
@@ -1,0 +1,112 @@
+package io.micrometer.spring.aop;
+
+import io.micrometer.core.annotation.Timed;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.annotation.Pointcut;
+import org.springframework.jmx.export.annotation.ManagedAttribute;
+import org.springframework.jmx.export.annotation.ManagedResource;
+
+import javax.annotation.Nonnull;
+
+import static org.springframework.util.StringUtils.isEmpty;
+
+/**
+ * AspectJ aspect for intercepting types or method annotated with @Timed.
+ *
+ * @author <a href="mailto:david@davidkarlsen.com">David J. M. Karlsen</a>
+ */
+@Aspect
+@ManagedResource
+public class MicrometerAspect {
+    private static final String DEFAULT_ID = "method_monitor";
+
+    private final MeterRegistry meterRegistry;
+    private final String description = getClass().getCanonicalName() + " monitor";
+
+    private boolean isActive = true;
+
+    public MicrometerAspect(@Nonnull MeterRegistry meterRegistry) {
+        this.meterRegistry = meterRegistry;
+    }
+
+    @ManagedAttribute
+    public void setActive(boolean active) {
+        isActive = active;
+    }
+
+    @ManagedAttribute
+    public boolean isActive() {
+        return isActive;
+    }
+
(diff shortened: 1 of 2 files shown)


jkschneider:
Thanks @davidkarlsen. 

It may make sense to remove `@ManagedResource` and move this to `micrometer-core` so other frameworks (Vert.x, Armeria, etc.) can take advantage of the same aspect. What impact would that have on you?

I'll talk with the Boot team about whether they are willing to autoconfigure a `@Timed` aspect in Boot 2. I'm not sure what they'll think. If not, at least we can document how to configure it.

davidkarlsen:
I'd prefer to keep it, but not a big deal, I can always extend and override it in-house if needed. I could swap it over to a plain configuration attribute - but you want it free from any spring dependencies?

jkschneider:
Yep, the idea would be to free it from Spring dependencies.

davidkarlsen:
Fixed

jkschneider:
cc / @philwebb

jkschneider:
We should take a look at `org.springframework.cache.annotation.Cacheable`. It's probable that this can be accomplished without AOP (i.e. `BeanPostProcessor` (BPP) that's looking for annotations on methods via reflection).

Puts us in a weird position, because a BPP + reflection approach would only work for Spring, but even in the Spring environment we don't want to have to rely on AOP as it isn't a default-on thing for Boot. Perhaps we wind up with a special Spring implementation with the BPP approach and an aspect for everyone else?

davidkarlsen:
OK, I guess that is doable... 
It seems they offer both approaches (aspectJ & Aop-alliance): 
https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/aspectj/AspectJCachingConfiguration.html
https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/interceptor/CacheInterceptor.html
but I have not looked very close into it.

Could you have a look at https://github.com/micrometer-metrics/micrometer/pull/394 in the mean time?

jkschneider:
Merged in b690f3d4. The aspect has been moved to `micrometer-core` and reduced in scope a bit. To configure manually in Boot:

```
@Configuration
@EnableAspectJAutoProxy
public class AutoTimingConfiguration {
   @Bean
   public TimedAspect timedAspect(MeterRegistry registry) {
      return new TimedAspect(registry);
   }
}
```

izeye:
@sensei-wu I have [a working sample](https://github.com/izeye/sample-micrometer-spring-boot/tree/timed-annotation), so I guess there's some issue in your configuration. Check the sample and then if you still have an issue, you'd better ask it to StackOverflow. Also note that there's [some known issue](https://github.com/micrometer-metrics/micrometer/issues/780).

jiang-wei:
> @jkschneider I could not get `@Timed` working for a simple service method. I have `TimedAspect` like you mentioned and my service method is annotated with `@Timed("mymetric")`. But I do not get `mymetric` under `/actuator/metrics` endpoint. I am using Spring Boot `2.0.4.RELEASE`

I noticed this commit is in package micrometer-spring-legacy
Do I have to include this as a dependency before using this feature?

ioannisvasilas:
Hello @jkschneider . I am facing the same issue here, as I cannot see my annotated metrics under the /actuator/metrics endpoint.  I have TimedAspect like you mentioned and my service method is annotated with @Timed. I have also included the below dependency:
`        <dependency>
            <groupId>org.aspectj</groupId>
            <artifactId>aspectjweaver</artifactId>
            <version>1.9.6</version>
        </dependency>`

Additionally, i tried to add the spring-boot-starter-aop dependency, but the result was the same. Is there something else that i am missing here?

Thank you in advance.

EnGoPy:
I'm facing the same problem, but in my case I've found in [micometer documentation](https://micrometer.io/docs/concepts#_histograms_and_percentiles) under 10.3 Chapter where which says:
```
Warning | Micrometer’s Spring Boot configuration does not recognize @Timed on arbitrary methods.
```
I tried to annotate public method, but the service wasn't bean at all.

PR #396, opened 2018-02-05
Fix sending metrics metadata to Datadog

maciejwalkowiak:
Sending metadata to Datadog results in error message:

> i.m.datadog.DatadogMeterRegistry         : failed to send metric metadata: {"errors": ["request data must be valid JSON"]}

due to sending unescaped quotes in JSON strings, for example:

> {"type":"gauge","description":"The "recent cpu usage" for the whole system"}


GreyTeardrop:
Does it make sense to escape other [JSON-special characters](https://tools.ietf.org/html/rfc7159#section-7) here?

PR #401, opened 2018-02-06
Guard against ClassNotFoundException on GarbageCollectionNotificationInfo in GC metrics binder

### LINKED ISSUE #391: Guard against ClassNotFoundException in JvmGcMetrics
See https://github.com/spring-projects/spring-boot/issues/11879

vpavic:
This resolves #391.

The original fix (see 85d01ec) didn't work because `JvmGcMetrics` was registered as a bean in Spring Boot's config, and as such was subject to introspection. Since introspection also looks at private methods, and `JvmGcMetrics#isConcurrentPhase` had `GarbageCollectionNotificationInfo` in its signature which caused things to blow up before even reaching `JvmGcMetrics#bindTo`.

Additionally, I've added some logging to make users aware that GC notifications won't be available, rather than just swallowing `ClassNotFoundException`.

Also since `JvmGcMetrics#bindTo` gets hit quite a few times in Boot app, I've extracted guard to initialization phase to prevent multiple calls to `Class.forName` as well as multiple logging entries. Check itself is moved to (what it seems to me) a better place to prevent registration of multiple notification listeners that won't be doing anything due to `GarbageCollectionNotificationInfo` not being available.

Let me know what you think @jkschneider. I've tested this in WildFly 10.1 both in scenarios where `com.sun.management` is and is not available.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -22,6 +22,8 @@ import io.micrometer.core.instrument.binder.MeterBinder;
 import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
 import io.micrometer.core.lang.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.management.NotificationEmitter;
 import javax.management.openmbean.CompositeData;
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -120,23 +127,19 @@ public void bindTo(MeterRegistry registry) {
...
+            for (GarbageCollectorMXBean mbean : ManagementFactory.getGarbageCollectorMXBeans()) {
+                if (mbean instanceof NotificationEmitter) {
+                    ((NotificationEmitter) mbean).addNotificationListener((notification, ref) -> {
+                        final String type = notification.getType();
                         if (type.equals(GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION)) {
                             CompositeData cd = (CompositeData) notification.getUserData();
                             GarbageCollectionNotificationInfo notificationInfo = GarbageCollectionNotificationInfo.from(cd);
 
-                            if (isConcurrentPhase(notificationInfo)) {
+                            if (isConcurrentPhase(notificationInfo.getGcCause())) {
```
jkschneider: Good catch! Thanks

jkschneider:
Appreciate your time @vpavic. It was a big help in this last push toward 1.0 GA.

PR #404, opened 2018-02-07
Pushgateway adjustments

mweirauch:
Adresses the most recent discussions on slack regarding the Pushgateway behaviour and setup.

//cc @davidkarlsen 

Relates to #394

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusExportConfiguration.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusExportConfiguration.java
@@ -26,12 +26,13 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.actuate.autoconfigure.ManagementContextConfiguration;
 import org.springframework.boot.actuate.endpoint.AbstractEndpoint;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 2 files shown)


micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusExportConfiguration.java
```diff
@@ -111,9 +130,17 @@ public PrometheusScrapeMvcEndpoint prometheusMvcEndpoint(PrometheusScrapeEndpoin
             this.pushgatewayProperties = prometheusProperties.getPushgateway();
             this.pushGateway = new PushGateway(pushgatewayProperties.getBaseUrl());
             this.environment = environment;
-            this.pushFuture = Executors.newSingleThreadScheduledExecutor()
-                    .scheduleAtFixedRate(this::push, pushgatewayProperties.getPushRate().toMillis(),
-                            pushgatewayProperties.getPushRate().toMillis(), TimeUnit.MILLISECONDS);
+            final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor(
```
davidkarlsen: Should the service be shutdown orderly when destroying app?

davidkarlsen:
LGTM!

micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusExportConfiguration.java
mweirauch: It does so when the daemon thread dies. Arguably it would be cleaner to `shutdownNow()` just didn't wan't to change the status-quo.

davidkarlsen: Just worried things can stick around when doing hot-reload.

mweirauch:
@davidkarlsen good hint regarding context reload. Fixed. Could have made the thread non-daemon again, but I think it doesn't hurt having it classified as a background one.
@jkschneider as per your request, I already toggled the push-on-shutdown to true by default.

jkschneider:
Thanks for the help once again @mweirauch.

PR #405, opened 2018-02-08
Gauge.java: use generics instead of raw types to avoid type warnings

JnRouvignac:


jkschneider:
Good catch, thanks @JnRouvignac!

PR #406, opened 2018-02-08
StatsD: Always publish unchanged meters 

### LINKED ISSUE #403: StatsdGauge not sending updates if there's no change
Hello!

I used Micrometer if one of our projects and I'm very happy with the library. I however stumbled over one issue with `StatsdGauge`: the `poll` method checks old and new value and only sends the metric to StatsD server if the value changed. This doesn't work well with infrequent changes and `deleteIdleStats` StatsD configuration option. This option removes metric if it hasn't been updated in a 10s interval. If the gauge value doesn't change locally in Micrometer in 10 seconds, `statsd` simply deletes the metric.

The `deleteIdleStats` option is important, because it prevents false positives - without it the last value is reported even if the app crashes for example and it's not sending values anymore. 

```java
    @Override
    public void poll() {
        double val = value();
        if(lastValue.getAndSet(val) != val) {
            publisher.onNext(lineBuilder.gauge(val));
        }
    }
```

https://github.com/micrometer-metrics/micrometer/blob/388fc3ea5b00905e48868defc4e8ec065b22b5f0/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdGauge.java#L51-L57

---

Could this check for last value be optional/configurable? I'm happy to contribute with a PR if you kick me into the right direction :-)

bob983:
`StatsdGauge` and `StatsdLongTaskTimer` were only publishing values when the values changed. This was in conflict with `deleteIdleStats` config option in StatsD server which enables deleting of stale metrics (metrics that weren't updated in last 10 seconds).

This PR adds `publishUnchangedMeters` setting that is propagated to both meters mentioned above. Default value is `true` which means by default these meters will publish values on every poll.

I wasn't sure how to test this so for now I just created a simple unit test for the `StatsdGauge` meter. I'm happy to add more tests if needed.

Fixes #403 

PS I will of course rebase and squash the commits to one once we're done here :)

### DIFF
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -8,4 +8,5 @@ dependencies {
     testCompile project(':micrometer-test')
     testCompile 'io.projectreactor:reactor-test:3.1.2.RELEASE'
     testCompile 'org.junit.jupiter:junit-jupiter-params:5.0.0'
+    testCompile 'org.mockito:mockito-core:2.11.0'
 }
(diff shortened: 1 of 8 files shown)


jkschneider:
@bob983 Well done, thanks!

Issue #175, opened 2017-10-24
file descriptors (Prometheus metric name)


danielmenezesbr:
According to [Prometheus documentation](https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics), the metric name to collect file descriptors is `process_open_fds` and `process_max_fds`. Micrometer uses `process_fds_open` and `process_fds_max`

related to #170

checketts:
The micrometer direction is intentional. There is a hierarchical naming that we are encouraging. Similar to how logging levels work. Someday (soon) I expect to have a `filter` I can set to disable metrics.

Something like:
```
metrics:
  process:
    cpu: enabled
    fds: disabled
```

The Prometheus metric name doesn't allow for this approach.

jkschneider:
If we had [added base units](https://github.com/micrometer-metrics/micrometer/blob/716980cc7aa6a1f6941c799f86b19167fe7f67fc/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetrics.java#L63) to the gauge, the Prometheus naming convention would be adding "fds" to the end anyway.

Since we should really have base units on any meter that conceptually has them in our out-of-the-box instrumentation, perhaps this forces us back to `process_open_fds` and `process_max_fds` like it is written in the Prom docs?

jkschneider:
Fixed with 736fd95ca5a9fb9846699489877eb4021882b9df

Issue #170, opened 2017-10-22
process_starttime_seconds (Prometheus metric name)


danielmenezesbr:
According to [Prometheus documentation](https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics), the metric name to collect process start time is `process_start_time_seconds`. Micrometer exposes it metric with the name `process_starttime_seconds`.

danielmenezesbr:
Prometheus client java for reference: https://github.com/prometheus/client_java/blob/ce76a494acd4f4077f01b6cea40e4329b0d54fac/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java#L72

mweirauch:
I knew this would crop up some day or the other. The motivation was to have a "one-word"-thingy as with "uptime".
In some sense, the micrometer Prometheus integration is a client implementation, on the other hand micrometer itself is a system on it's own exposing metric names as it sees fit for itself. The Prometheus integration is just one of the many integrations for the supported metric backends.
So changing the name just to fit Prometheus' client recommendations will change the metric name for any other backend integration mircometer supports.
Personally I am unbiased on the one or the other naming. "start_time" is likely the better english.

jkschneider:
Few should be reliant on the Micrometer name yet, and `process.start.time` is reasonable in the naming convention independent form. So I'd support changing it to follow the existing name Prom users are accustomed to.

@mweirauch correctly points out that we aren't bound to follow Prometheus names for particular pieces of instrumentation where they don't make sense generally, but I don't think this is one of those cases.

PR #412, opened 2018-02-10
Rename file descriptor metrics

shakuzen:
(Hopefully) makes it more clear that the `open` and `max` part of the metric refers to file descriptors. This may also help querying these metrics since their prefix is now the same.

The `FileDescriptorMetrics` are also never being used, but I will send a separate pull request for that.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMetricRenameFilter.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMetricRenameFilter.java
@@ -1,0 +1,43 @@
+/**
+ * Copyright 2018 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.prometheus;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.MeterFilter;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Converts known metric names from Micrometer's preferred name to Prometheus' preferred name.
+ *
+ * @author Tommy Ludwig
+ */
+public class PrometheusMetricRenameFilter implements MeterFilter {
+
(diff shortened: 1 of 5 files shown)


mweirauch:
I initially implemented it like that due to the IMHO better hierarchical naming. Then there was some discussion regarding "Prometheus client implementation" conformance in #175 and #170.

In the meantime I've seen different projects exposing Prometheus metrics not adhering to any Prometheus-proposed naming conventions I whonder if it's really worth it. In one comment I said that Micrometer is the system, not that one registry implementation for Prometheus.

Anyways, I consider the metric names part of the API, so with whatever we go with should be stable before `1.0.0` hits the road.

shakuzen:
@mweirauch Thank you for sharing the background, and my apologies I didn't search to find the previous discussion before proposing the changes.

> In one comment I said that Micrometer is the system, not that one registry implementation for Prometheus.

I think I agree, and maybe we are better off registering a filter when Prometheus registry implementation is used that renames the metrics that don't match what Micrometer otherwise prefers to call them. This seems more practical since other systems will likely have their own preferences.

> Anyways, I consider the metric names part of the API, so with whatever we go with should be stable before `1.0.0` hits the road.

I agree with this also.

jkschneider:
Agree with all of the above. And this is a good idea:

> maybe we are better off registering a filter when Prometheus registry implementation is used that renames the metrics that don't match what Micrometer otherwise prefers to call them

wilkinsona:
I originally opened #425, but @shakuzen pointed out that it was largely a duplicate of this PR so I've closed it.

In addition to the renaming proposed here, could changing `fds` to something more descriptive also be considered? Perhaps `files`:

- `process.files.open`
- `process.files.max`

shakuzen:
@jkschneider I have added a `PrometheusMetricRenameFilter` and configured it in auto-configuration. Let me know if this is what you had in mind. I can also apply @wilkinsona's suggested name change to `process.files.*` if you agree.

mweirauch:
Must admit, I like the renaming proposal to `process.files.*`. What I somehow wouldn't like to see is a metric renaming just to be "compliant" with the [docs](https://prometheus.io/docs/instrumenting/writing_clientlibs/). IMHO it's rather unintuitve that we rename it. Leads to confusion why it's *this* in registry A, and *that* in registry B.

> If a language or runtime doesn't expose one of the variables it'd just not export it.

Let's say we "don't export it" (officially) and we stick to the (re-proposed) hierarchical name?

Adding to that, I don't see any big overlaps in metric naming e.g. between Micrometer, Dropwizard and Prometheus-Java. These are heterogenous per-se.

checketts:
I also like the renaming of `f`ile`d`escriptor`s` to `files`. I think the `PrometheusMetricRenameFilter` is a good idea to have. I'm also OK if it is off by default.

While it is nice to have metrics remain the same between languages (go and java apps for example), they truly diverge very quickly (tracking coroutines versus threads for example). Keeping them the same is only helpful if the metric names are non-intuitive to begin with (`fds` for example)

jkschneider:
I think we'll leave the rename filter off by default to hopefully lessen the surprise, but document its presence for those that care to follow the Prometheus "standard" names.

PR #414, opened 2018-02-11
Fix Jigsaw compatibility for FileDescriptorMetrics and ProcessorMetrics

shakuzen:
jigsaw crash

I noticed that some of the JVM-gathered metrics were not Jigsaw compatible and were effectively disabled (or worse, throwing exceptions) on Java 9+. I've done my best to make these compatible with jigsaw, which should allow them to work with Java 9 and beyond.

- using reflection on only public, exported classes' methods (not methods in internal implementation classes)
- do not `import` a specific JVM class

This thread on the same issue was enlightening: http://mail.openjdk.java.net/pipermail/jigsaw-dev/2015-September/004568.html

I don't have a J9 JDK installation to properly test out that support (is there any easy way to get one on OSX?)

Another consequence of this change is that we have to explicitly support JVM flavors by adding their public, exported interface class to the list (which hopefully uses the same method names). Before, if the code ran on Java 8, it could work with any JVM implementation class that used the expected method names. This isn't viable post-Jigsaw unless the implementation class returned from `ManagementFactory.getOperatingSystemMXBean()` is exported, which is not generally expected.

It might be a good idea to consider introducing something like https://github.com/policeman-tools/forbidden-apis and/or switching the build to run with JDK9 and use the compiler flag `release` to target `1.8`, which [is what JUnit 5 does](https://github.com/junit-team/junit5/blob/a1463d903ab9d4ce249f463081e27bd2035643f4/gradle.properties#L13-L17).

/cc @jkschneider @mweirauch

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetrics.java
@@ -27,19 +27,38 @@ import java.lang.management.ManagementFactory;
 import java.lang.management.OperatingSystemMXBean;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
 
 import static java.util.Collections.emptyList;
 
 /**
- * File descriptor metrics.
+ * File descriptor metrics gathered by the JVM.
+ * <p>
+ * Supported JVM implementations:
+ * <ul>
+ *     <li>HotSpot</li>
+ *     <li>J9</li>
+ * </ul>
  *
  * @author Michael Weirauch
+ * @author Tommy Ludwig
  */
 @NonNullApi
 @NonNullFields
 public class FileDescriptorMetrics implements MeterBinder {
+
+    /** List of public, exported interface class names from supported JVM implementations. */
+    private static final List<String> UNIX_OPERATING_SYSTEM_BEAN_CLASS_NAMES = Arrays.asList(
+        "com.sun.management.UnixOperatingSystemMXBean", // HotSpot
+        "com.ibm.lang.management.UnixOperatingSystemMXBean" // J9
+    );
+
     private final OperatingSystemMXBean osBean;
     private final Iterable<Tag> tags;
+
+    @Nullable
+    private final Class<?> osBeanClass;
 
     @Nullable
     private final Method openFdsMethod;
@@ -60,6 +79,7 @@ public class FileDescriptorMetrics implements MeterBinder {
         this.osBean = osBean;
         this.tags = tags;
 
+        this.osBeanClass = getFirstClassFound(UNIX_OPERATING_SYSTEM_BEAN_CLASS_NAMES);
         this.openFdsMethod = detectMethod("getOpenFileDescriptorCount");
         this.maxFdsMethod = detectMethod("getMaxFileDescriptorCount");
     }
@@ -91,12 +111,26 @@ public class FileDescriptorMetrics implements MeterBinder {
 
     @Nullable
     private Method detectMethod(String name) {
-        try {
-            final Method method = osBean.getClass().getDeclaredMethod(name);
-            method.setAccessible(true);
-            return method;
-        } catch (NoSuchMethodException | SecurityException e) {
+        if (osBeanClass == null) {
             return null;
         }
+        try {
+            // ensure the Bean we have is actually an instance of the interface
+            osBeanClass.cast(osBean);
+            return osBeanClass.getDeclaredMethod(name);
+        } catch (ClassCastException | NoSuchMethodException | SecurityException e) {
+            return null;
+        }
+    }
+
+    @Nullable
+    private Class<?> getFirstClassFound(List<String> classNames) {
+        for (String className : classNames) {
+            try {
+                return Class.forName(className);
+            } catch (ClassNotFoundException ignore) {
+            }
+        }
+        return null;
     }
 }
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetricsTest.java
@@ -20,17 +20,15 @@ import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import org.junit.jupiter.api.Test;
 
-import java.lang.management.ManagementFactory;
 import java.lang.management.OperatingSystemMXBean;
-import java.lang.reflect.InvocationTargetException;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assumptions.assumeFalse;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static org.mockito.Mockito.mock;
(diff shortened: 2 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetrics.java
```diff
@@ -27,20 +27,39 @@
...
+ * File descriptor metrics gathered by the JVM.
+ * <p>
+ * Supported JVM implementations:
+ * <ul>
+ *     <li>HotSpot</li>
+ *     <li>J9</li>
+ * </ul>
  *
  * @author Michael Weirauch
+ * @author Tommy Ludwig
```
shakuzen: I'm not sure the conventions for `@author` tags in the project. For instance, `ProcessorMetrics` doesn't have any `@author` tags, but this class did, so I added my name.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetrics.java
```diff
@@ -91,12 +111,26 @@ private double invoke(@Nullable Method method) {
 
     @Nullable
     private Method detectMethod(String name) {
+        if (osBeanClass == null) {
+            return null;
+        }
         try {
-            final Method method = osBean.getClass().getDeclaredMethod(name);
```
shakuzen: `osBean.getClass()` will return the underlying implementation class, to which we don't have reflective access post-Jigsaw.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetricsTest.java
```diff
@@ -20,76 +20,55 @@
...
  */
 class FileDescriptorMetricsTest {
     private MeterRegistry registry = new SimpleMeterRegistry();
 
-    @Test
-    @SuppressWarnings("restriction")
-    void fileDescriptorMetricsRuntime() {
-        // currently only supporting HotSpot JVM
-        final OperatingSystemMXBean osBean = ManagementFactory.getOperatingSystemMXBean();
-        assumeTrue(osBean instanceof com.sun.management.UnixOperatingSystemMXBean);
```
shakuzen: Can't use `instanceof` on a class that will only be there when the JVM implementation is HotSpot. It would throw an exception on a different JVM.
HotSpot and J9 and supported now. Setting up a build that runs on J9 might be a good idea.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetricsTest.java
```diff
@@ -20,76 +20,55 @@
...
 
     @Test
-    void fileDescriptorMetricsSupportedOsBeanMock() {
-        final HotSpotLikeOperatingSystemMXBean osBean = mock(
-            HotSpotLikeOperatingSystemMXBean.class);
-        when(osBean.getOpenFileDescriptorCount()).thenReturn(Long.valueOf(512));
-        when(osBean.getMaxFileDescriptorCount()).thenReturn(Long.valueOf(1024));
-        new FileDescriptorMetrics(osBean, Tags.of("some", "tag")).bindTo(registry);
+    void unixFileDescriptorMetrics() {
+        assumeFalse(System.getProperty("os.name").toLowerCase().contains("win"));
```
shakuzen: If anyone has a better suggestion for the Unix detection logic here, please let me know. I didn't put much thought into it.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetricsTest.java
```diff
@@ -20,76 +20,55 @@
...
         assertThat(registry.get("process.max.fds").tags("some", "tag")
-            .gauge().value()).isEqualTo(1024.0);
+            .gauge().value()).isGreaterThan(0);
     }
 
     @Test
-    void fileDescriptorMetricsInvocationException() {
-        final HotSpotLikeOperatingSystemMXBean osBean = mock(
-            HotSpotLikeOperatingSystemMXBean.class);
-        when(osBean.getOpenFileDescriptorCount()).thenThrow(InvocationTargetException.class);
```
shakuzen: I didn't fully understand the case in which this happens, but let me know if we should somehow test this still.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java
```diff
@@ -42,16 +39,7 @@ void cpuMetrics() {
...
-        boolean jdk9 = false;
-        try {
-            Class.forName("java.lang.reflect.InaccessibleObjectException");
-            jdk9 = true;
-        } catch (ClassNotFoundException ignored) {
-        }
-
-        OperatingSystemMXBean osBean = ManagementFactory.getOperatingSystemMXBean();
-        assumeTrue(!jdk9);
-        assumeTrue(osBean.getClass().getName().equals("sun.management.OperatingSystemImpl"));
```
shakuzen: This line didn't work on Java 9 because the implementation class is not accessible, but we need only check for the HotSpot-specific accessible interface in this test. This will work on Java 8 and 9.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java
```diff
@@ -62,8 +50,7 @@ void hotspotCpuMetrics() {
 
     @Test
     void openJ9CpuMetrics() {
-        OperatingSystemMXBean osBean = ManagementFactory.getOperatingSystemMXBean();
-        assumeTrue(osBean.getClass().getName().equals("com.ibm.lang.management.internal.UnixExtendedOperatingSystem"));
```
shakuzen: In addition to access issues, I thought it best to not make our tests depend on the underlying, hidden implementation class. Rather, I thought it would be better to use the public, exported interface to check the JVM implementation. In this way, a future version of the JVM can change and remove implementation classes without us being affected.

checketts:
sdkman might have a simple Java 9 option: http://sdkman.io

mweirauch:
First of all, thanks for looking into this. I've always "avoided" getting in touch with Java9 as I haven't had the use-case :)

I'll try to give it a go with different Runtimes. I was about to propose https://adoptopenjdk.net/ but there is no build of a JDK with openJ9 on OSX yet. Likely due [this.](https://github.com/eclipse/openj9/issues/36)

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetricsTest.java
jkschneider: This is more or less what Apache commons-lang does, so seems fine.

jkschneider:
Thanks for doing all this, @shakuzen.

jkschneider:
We really should build against Java 9, but may not get prioritized until after 1.0

mweirauch:
Of course you've been faster with merging ... :)

Tested on Fedora27:
```
OpenJDK8u161-Hotspot -> OK
OpenJDK8u152-OpenJ9 -> OK
OracleJDK9.0.4-Hotspot -> OK
OpenJDK9+181-OpenJ9 -> OK
```

PR #418, opened 2018-02-12
Added auto-configuration for ClassLoaderMetrics

sergei-ivanov:
For some reason, `ClassLoaderMetrics` was left out of the default configuration.
I assume it was simply an omission rather than a deliberate exclusion.

I also had to enforce source encoding as UTF-8 in Gradle, because `WavefrontNamingConventionTest` containing Japanese characters kept failing on my Windows PC.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,9 @@ subprojects {
     apply plugin: 'java'
     apply plugin: 'checkstyle'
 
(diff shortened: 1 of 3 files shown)


jkschneider:
Merged with f6578c8fc048373bdc152339a271498cb6f08a26. Thanks!

Issue #154, opened 2017-10-11
Support Elastic


jkschneider:
See [the docs](https://www.elastic.co/guide/en/beats/metricbeat/6.0/metricbeat-module-http.html). This was raised in https://github.com/spring-projects/spring-boot/issues/10449.

antonyb:
Metricbeat 6.2 will introduce a "dedot.enabled" flag to the HTTP module.
It'll convert dot (".") in field names to underscore ("_").

https://www.elastic.co/guide/en/beats/metricbeat/6.2/metricbeat-metricset-http-json.html#_dedot_enabled

punkstarman:
From a technical point of view, the problem for MetricBeat + ElasticSearch will be solved by this `dedot.enabled` flag.

However, from a "business" viewpoint, wouldn't it be possible to solve the problem by adding a suffix (e.g. `.total`, `.max`, `.value`) to a key whenever other keys contain the first as a prefix. For instance, change `mem` to `mem.total` when `mem.free` is present.

Just as a reminder, the problem arises from the fact that ElasticSearch interprets dots in the keys as map indexing (e.g. `mem.total = 4, mem.free = 2` becomes `{ mem: { total = 4, free = 2 } }`). Thus after reading `mem = 4, mem.free = 2`, ElasticSearch create an object with value `{ mem: 4 }` then tries to introduce a sub field named `free` into the `mem` field. This fails because `mem` contains a scalar type when it needs to contain an object type.

I feel that replacing dots with underscores is just a workaround because we then lose the hierarchical information which is very useful to most users.

jkschneider:
Yikes. I wonder if we have OOTB binders that already exhibit this problem?

jkschneider:
> we then lose the hierarchical information which is very useful to most users.

I wonder how useful it really is. In hierarchical metrics systems, hierarchical information is essentially used to encode tag information. But in dimensional systems, naturally that isn't the case.  I don't know which MetricBeat + Elasticsearch is having not looked closely at this yet.

In a dimensional system, you just select a metric name and go from there.

punkstarman:
[...]

PR #429, opened 2018-02-14
Add Elastic support

tkp1n:
Proposal for issue #154 

Basically a re-implementation of the [elasticsearch-metrics-reporter-java](https://github.com/elastic/elasticsearch-metrics-reporter-java) for micrometer.

@jkschneider


tkp1n:
So I just took the elastic implementation for a spin.

Running a local instance of both elasticsearch and kibana is as simple as:
`docker run -p 5601:5601 -p 9200:9200 -p 5044:5044 -it --name elk sebp/elk:600`

Running the `CounterSample` from the repository pushed the following information into ES and kibana:
countersample_data

A visualization of the rate at which the counters arrive:
countersample_rateviz

And a visualization of the acutal value (the max value per 30 seconds interval/bucket):
countersample_valueviz


Running the `TimerSample` from the repository pushed the following information into ES and kibana:
timersample_data

A visualization of timer values (the max timer value per 30 seconds interval/bucket):
timersample_maxmeanper30sec

I'm sure some kibana champion could produce nicer charts, sorry about that.
(Both samples were run with the default configuration)

jkschneider:
Working on merging this now. Some changes coming:

1. We will make `_type` = 'metric' for all meter types going forward, as in ES 6, indexes may now no longer have more than one distinct `_type`. See [this](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/removal-of-types.html#_why_are_mapping_types_being_removed).

jkschneider:
> Why not "_type": "doc"?

Changed to "doc" as you suggested. Unless you see it as really important, I think we'll stay away from making it configurable for now, as the whole concept of type is going away in future ES releases.

jkschneider:
Merged in 3ec44d958736b0e4c6979479cbc2982fd491384b. I expect we'll make further tweaks, but we have at least some implementation on master now!

tkp1n:
Did you really intent to add a System.out.println(); here?

https://github.com/micrometer-metrics/micrometer/blob/3ec44d958736b0e4c6979479cbc2982fd491384b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepFunctionTimer.java#L79

Thanks for your work on the PR!

jkschneider:
Oops. Thanks!

quzhixue-Kimi:
Hi dear, 
Do you mean that if I would like to use the es as the metrics storage, the version of spring boot must be 2.1.0 or greater?
Currently, spring boot 2.0.5 is used but, failed.

BR
Kimi

izeye:
@quzhixue-Kimi Yes, as you can see, the Elasticsearch meter registry support has been added since 1.1.0 (1.1.0-rc.1) and Spring Boot 2.1.0 is compatible with Micrometer 1.1.0, so you should use Spring Boot 2.1.0 or later.

quzhixue-Kimi:
@izeye, thank you for your feedback.

cykl:
Hey @quzhixue-Kimi. If 2.1 is not yet an option for you, I successfully backported ES autoconfiguration (and micrometer 1.1.0) to Boot 2.0. IIRC it took me less than an hour and was quite easy.

quzhixue-Kimi:
@cykl , it is appreciated that you may share your git repository as you mention.

BR
Kimi

PR #455, opened 2018-02-21
Specify urlTemplate is nullable (fixes #453)

### LINKED ISSUE #453: urlTemplate not marked as nullable for RestTemplateExchangeTagsProvider
When upgrading to 1.0.0 we ran into an issue where urlTemplate was not marked nullable for `RestTemplateExchangeTagsProvider::getTags`.  As of 1.0.0 nulls are now passed for urlTemplate when `RestTemplate` is used with a `URI`.  This resulted in a NPE as runtime in our `RestTemplateExchangeTagsProvider` implementation.

michael-barker:
urlTemplate was previously not nullable but as of 1.0.0
a null is now provided to RestTemplateExchangeTagsProvider
when a URI is used with RestTemplate.

#453

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/client/DefaultRestTemplateExchangeTagsProvider.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/client/DefaultRestTemplateExchangeTagsProvider.java
@@ -32,9 +32,9 @@ public class DefaultRestTemplateExchangeTagsProvider
     implements RestTemplateExchangeTagsProvider {
 
     @Override
-    public Iterable<Tag> getTags(String urlTemplate, HttpRequest request,
+    public Iterable<Tag> getTags(@Nullable String urlTemplate, HttpRequest request,
                                  @Nullable ClientHttpResponse response) {
-        Tag uriTag = StringUtils.hasText(urlTemplate)
+        Tag uriTag = urlTemplate != null && StringUtils.hasText(urlTemplate)
             ? RestTemplateExchangeTags.uri(urlTemplate)
             : RestTemplateExchangeTags.uri(request);
         return Arrays.asList(RestTemplateExchangeTags.method(request), uriTag,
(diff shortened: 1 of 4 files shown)


micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/client/DefaultRestTemplateExchangeTagsProvider.java
```diff
@@ -32,9 +32,9 @@
     implements RestTemplateExchangeTagsProvider {
 
     @Override
-    public Iterable<Tag> getTags(String urlTemplate, HttpRequest request,
+    public Iterable<Tag> getTags(@Nullable String urlTemplate, HttpRequest request,
                                  @Nullable ClientHttpResponse response) {
-        Tag uriTag = StringUtils.hasText(urlTemplate)
+        Tag uriTag = urlTemplate != null && StringUtils.hasText(urlTemplate)
```
michael-barker: Do we still need to check `StringUtils.hasText(urlTemplate)`?

jkschneider:
Wow, that was exceptionally fast.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/client/DefaultRestTemplateExchangeTagsProvider.java
jkschneider: `StringUtils.hasText` checks for null and returns `false` if the input is `null`. The tests still pass without the additional `urlTemplate != null` check. Not sure we have captured the case that yields an NPE?

michael-barker: I figured just `urlTemplate != null` without the has text check would be preferred but didn't know if there was a case where we would still want to use the request URI if `urlTemplate` is blank.

jkschneider: It does at least remove an IDE warning since `hasText` doesn't have a `@Nullable` marker until Spring 5.

PR #460, opened 2018-02-22
influx: create db inside try (fixes #379)

### LINKED ISSUE #379: Influx Metric Registry Publish Can Fail Silently
I had a bug in my logic to create timer metrics which passed in 'null' as one of the values of a tag. The next execution of the publish method would then throw a NullPointerException when invoking writeTimer. The exception wasn't handled in the publish method, bubbled back to the Executor which then failed silently and as result all no more metrics were ever sent.

Currently, the publish method implementation inside InfluxMetricRegistry only catches MalformedURLException and IOException ie.
```
        } catch (MalformedURLException e) {
            throw new IllegalArgumentException("Malformed InfluxDB publishing endpoint, see '" + config.prefix() + ".uri'", e);
        } catch (IOException e) {
            logger.warn("failed to send metrics", e);
        }
```

I am unsure of the best solution for this however I believe that the NullPointerException should be handled and the appropriate log message should be displayed.

alebar:
the `createDatabaseIfNecessary()` method should also be wrapped with try-catch

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -90,9 +90,9 @@ public class InfluxMeterRegistry extends StepMeterRegistry {
 
     @Override
     protected void publish() {
-        createDatabaseIfNecessary();
-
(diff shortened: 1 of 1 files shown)


jkschneider:
Merged the alternative #461. Thanks!

PR #467, opened 2018-02-25
Influx registry improvements

alebar:
Here I solve two problems that I encountered when trying to inject the environment variable with retention policy.

My goal is to allow to specify retention policy through an env var, called for example INFLUX_RET_POLICY. And default policy should be used if there's no such var defined. To do that I firstly tried (knowing that it's wrong, just for test) to set the property like that:
`management.metrics.export.influx.retentionPolicy=${INFLUX_RET_POLICY}` and tested what happens when there's no such variable defined.

As a result the placeholder was not resolved and the `config.retentionPolicy()` was eqal to "${INFLUX_RET_POLICY}". Which in turn resulted in a wrong URL and publishing metrics failed. But as my logging level for `InfluxMeterRegistry` was set to `ERROR`, I saw nothing in logs. So first commit is to fix that.

Then I changed the property to `${INFLUX_RET_POLICY:}` which was resolved as empty string. For example in Telegraf that's the way for using default policy. But not here. I had to change null check for empty-string check.

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -22,6 +22,8 @@ import io.micrometer.core.instrument.step.StepMeterRegistry;
 import io.micrometer.core.instrument.util.DoubleFormat;
 import io.micrometer.core.instrument.util.MeterPartition;
 import io.micrometer.core.lang.Nullable;
+
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -22,6 +22,8 @@
 import io.micrometer.core.instrument.util.DoubleFormat;
 import io.micrometer.core.instrument.util.MeterPartition;
 import io.micrometer.core.lang.Nullable;
+
+import org.apache.commons.lang.StringUtils;
```
jkschneider: Won't be able to use this because:

```
commons-lang:commons-lang:2.6
\--- commons-configuration:commons-configuration:1.8
     \--- com.netflix.archaius:archaius-core:0.4.1
          \--- com.netflix.hystrix:hystrix-core:1.5.12
               \--- project :micrometer-core
                    \--- compileClasspath
```

PR #483, opened 2018-03-08
TimedAspect should support @Timed histogram

ahus1:
If you use TimedAspect, it doesn't use the histogram attribute. 

This PR adds the functionality. I would add a test case but I don't know how. 

I tried _TimedAspectTest_ and _registry.get("something").timer()_, but wasn't able to extract the necessary information from the timer instance.

A hint would be helpful - either a code snippet or a pointer to a test case that already achieves this.


jkschneider:
@ahus1 As a workaround for now, you can use a property-based meter filter in Boot:

`management.metrics.distribution.percentile-histogram.MY.TIMER: true`

Or this can be done [programmatically](http://micrometer.io/docs/concepts#_configuring_distribution_statistics) as well with a MeterFilter.

But you are certainly right, this should be supported by the aspect. Good catch!

PR #495, opened 2018-03-15
Have a default metric name when not overridden in annotation. Fixes #494

### LINKED ISSUE #494: TimedAspect improvement #2 - default metric name
Suggest to make the aspect provide a default metric name if not provided in the annotations default() value - this helps on metricname uniformity - which again helps on reusable templated dashboards. As of today the aspect will just proceed() without doing any metric collections if there is no value() - which forces the user to provide a value - which again will make the non-uniform.

davidkarlsen:


micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
```diff
@@ -60,16 +61,13 @@ public TimedAspect(MeterRegistry registry, Function<ProceedingJoinPoint, Iterabl
     public Object timedMethod(ProceedingJoinPoint pjp) throws Throwable {
         Method method = ((MethodSignature) pjp.getSignature()).getMethod();
         Timed timed = method.getAnnotation(Timed.class);
-
-        if (timed.value().isEmpty()) {
-            return pjp.proceed();
-        }
+        final String metricName = timed.value().isEmpty() ? DEFAULT_METRIC_NAME : timed.value();
```
checketts: Why not build it from the package/class/method name? With using a default, it will keep the cardinality down, but I'm wondering how useful it will be to the user

jkschneider: I think this speaks to the intended use, which is why I hesitated at a default before. If you are timing a specific code path, then plotting the name alone (exploded by a tag that represents the package/class/method) provides a cumulative aggregate. This is similar to what we did in spring-integration:

image

If they all had different names, you'd have to construct this chart manually.

Keep in mind, the number of distinct time series is the same if you vary the name or a tag.

davidkarlsen: "Why not build it from the package/class/method name? With using a default, it will keep the cardinality down, but I'm wondering how useful it will be to the user": Because then it is close to impossible to template in grafana, with a common metric name it can be a dashboard, with templated dropdowns for class/method you want via tags.

jkschneider:
Could you provide a default `tagsBasedOnJoinpoint` as well that tags with class/method or whatever you are doing?

davidkarlsen:
Ah - you beat me to it - yes - it absolutely makes sense to provide a default implementation of Function<ProceedingJoinPoint, Iterable<Tag>> tagsBasedOnJoinpoint which tags method/class - that's what I provide in my internal framework which uses micrometer.

davidkarlsen:
It's already there:
```
 public TimedAspect(MeterRegistry registry) {
        this(registry, pjp ->
                Tags.of("class", pjp.getStaticPart().getSignature().getDeclaringTypeName(),
                        "method", pjp.getStaticPart().getSignature().getName())
        );
    }
```

gavlyukovskiy:
Shouldn't `method_timed` be written with dot convention?

jkschneider:
@gavlyukovskiy Good catch. Fixed in 0061a4ae

PR #500, opened 2018-03-16
Introduced TimedInterceptor, fixes #493, revisit #494

### LINKED ISSUE #493: Be able to target classes and not only methods with TimedAspect
Suggest to make the aspect target also types (in addition to methods), so that by annotating a class, you target all public methods in that class too (due to the way spring works it does not target private/default/protected scopes).

gavlyukovskiy:
I saw discussion in gitter yesterday regarding `@Timed` enhancements and we've already implemented something like timed to meter specific classes with custom metric name and tags extractors.
I didn't extend existing `TimedAspect` and created spring aware `TimedMethodInterceptor` because aspect doesn't work with annotated interfaces thus is not usable for spring data repositories and feign clients.

In my implementation `@Timed` on method level has higher precedence over class level, but tags are merged from both class level and method level annotations.

cc @davidkarlsen

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedAnnotationAdvisor.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedAnnotationAdvisor.java
@@ -1,0 +1,52 @@
+package io.micrometer.spring.advice;
+
+import io.micrometer.core.annotation.Timed;
+import org.aopalliance.aop.Advice;
+import org.aopalliance.intercept.MethodInterceptor;
+import org.springframework.aop.Pointcut;
+import org.springframework.aop.support.AbstractPointcutAdvisor;
+import org.springframework.aop.support.ComposablePointcut;
+import org.springframework.aop.support.annotation.AnnotationMatchingPointcut;
+import org.springframework.aop.support.annotation.AnnotationMethodMatcher;
+import org.springframework.core.annotation.AnnotationUtils;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+
+public class TimedAnnotationAdvisor extends AbstractPointcutAdvisor {
+
+    private final Pointcut pointcut;
+    private final MethodInterceptor advice;
+
+    public TimedAnnotationAdvisor(TimedMethodInterceptor timedMethodInterceptor) {
+        Pointcut cpc = new AnnotationMatchingPointcut(Timed.class, true);
+        Pointcut mpc = new ComposablePointcut(new InheritedAnnotationMethodMatcher(Timed.class));
+        this.pointcut = new ComposablePointcut(cpc).union(mpc);
+        this.advice = timedMethodInterceptor;
+    }
+
+    @Override
+    public Pointcut getPointcut() {
+        return pointcut;
+    }
+
+    @Override
+    public Advice getAdvice() {
+        return advice;
+    }
+
+    private static class InheritedAnnotationMethodMatcher extends AnnotationMethodMatcher {
+
+        private Class<? extends Annotation> annotationType;
+
+        public InheritedAnnotationMethodMatcher(Class<? extends Annotation> annotationType) {
+            super(annotationType);
+            this.annotationType = annotationType;
+        }
+
+        @Override
+        public boolean matches(Method method, Class<?> targetClass) {
+            return AnnotationUtils.findAnnotation(method, annotationType) != null || super.matches(method, targetClass);
+        }
+    }
+}
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
@@ -1,6 +1,7 @@
 package io.micrometer.spring.advice;
 
 import io.micrometer.core.annotation.Timed;
+import io.micrometer.core.aop.TimedAspect;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.Timer;
@@ -13,7 +14,6 @@ import org.springframework.util.ReflectionUtils;
 import java.lang.reflect.Method;
 
 public class TimedMethodInterceptor implements MethodInterceptor {
-    public static final String DEFAULT_METRIC_NAME = "method_timed";
     private final MeterRegistry registry;
     private final TimedMetricNameResolver timedMetricNameResolver;
     private final TimedTagsResolver timedTagsResolver;
@@ -57,7 +57,7 @@ public class TimedMethodInterceptor implements MethodInterceptor {
             metricName = classLevelTimed.value();
         }
         else {
-            metricName = DEFAULT_METRIC_NAME;
+            metricName = TimedAspect.DEFAULT_METRIC_NAME;
         }
         Builder builder = Timer.builder(timedMetricNameResolver.apply(metricName, invocation));
         if (methodLevelTimed != null) {
@@ -72,13 +72,12 @@ public class TimedMethodInterceptor implements MethodInterceptor {
         }
 
         if (classLevelTimed != null) {
-            builder.tags(classLevelTimed.extraTags())
-                .tags(timedTagsResolver.apply(invocation));
+            builder.tags(classLevelTimed.extraTags());
         }
         if (methodLevelTimed != null) {
-            builder.tags(methodLevelTimed.extraTags())
-                .tags(timedTagsResolver.apply(invocation));
+            builder.tags(methodLevelTimed.extraTags());
         }
+        builder.tags(timedTagsResolver.apply(invocation));
         try {
             return invocation.proceed();
         } finally {
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
@@ -1,0 +1,88 @@
+package io.micrometer.spring.advice;
+
+import io.micrometer.core.annotation.Timed;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.Timer.Builder;
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
+import org.springframework.core.annotation.AnnotatedElementUtils;
+import org.springframework.util.ReflectionUtils;
+
+import java.lang.reflect.Method;
+
+public class TimedMethodInterceptor implements MethodInterceptor {
+    public static final String DEFAULT_METRIC_NAME = "method_timed";
+    private final MeterRegistry registry;
+    private final TimedMetricNameResolver timedMetricNameResolver;
+    private final TimedTagsResolver timedTagsResolver;
+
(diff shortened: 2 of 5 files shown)


micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class TimedMethodInterceptor implements MethodInterceptor {
+    public static final String DEFAULT_METRIC_NAME = "method_timed";
+    private final MeterRegistry registry;
+    private final TimedMetricNameResolver timedMetricNameResolver;
+    private final TimedTagsResolver timedTagsResolver;
+
+    public TimedMethodInterceptor(MeterRegistry registry) {
+        this(registry, invocation ->
+            Tags.of("class", invocation.getMethod().getDeclaringClass().getSimpleName(),
```
davidkarlsen: If however not super-likely - I think it would be better with the fully qualified classname (or another tag for package) - so that we ensure there is no possible clash.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
```diff
@@ -0,0 +1,88 @@
...
+import io.micrometer.core.instrument.Timer.Builder;
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
+import org.springframework.core.annotation.AnnotatedElementUtils;
+import org.springframework.util.ReflectionUtils;
+
+import java.lang.reflect.Method;
+
+public class TimedMethodInterceptor implements MethodInterceptor {
+    public static final String DEFAULT_METRIC_NAME = "method_timed";
```
davidkarlsen: should this possibly be shared with the aspectJ based metric name?

gavlyukovskiy: I was thinking about replacing `TimedAspect` with this implementation, but if not I'll change it.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
gavlyukovskiy: I would like to go with another tag for package to not blow up size of class tag. However I don't think someone will put `@Timed` on two different classes with the same name.

jkschneider:
Thanks @gavlyukovskiy.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedAnnotationAdvisor.java
```diff
@@ -0,0 +1,52 @@
...
+import java.lang.reflect.Method;
+
+public class TimedAnnotationAdvisor extends AbstractPointcutAdvisor {
+
+    private final Pointcut pointcut;
+    private final MethodInterceptor advice;
+
+    public TimedAnnotationAdvisor(TimedMethodInterceptor timedMethodInterceptor) {
+        Pointcut cpc = new AnnotationMatchingPointcut(Timed.class, true);
+        Pointcut mpc = new ComposablePointcut(new InheritedAnnotationMethodMatcher(Timed.class));
```
gavlyukovskiy: ` InheritedAnnotationMethodMatcher` looks for `@Timed` annotation not only on implementation method, but also goes to interface method. In Spring Framework 5.0 it can be replaced by `new AnnotationMatchingPointcut(null, Timed.class, true)`.

jkschneider:
@huehnerlady Still planned for 1.1.0.

huehnerlady:
@jkschneider do you have an idea when this will be released approximately?

micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
```diff
@@ -0,0 +1,87 @@
...
+                .publishPercentiles(classLevelTimed.percentiles().length == 0 ? null : classLevelTimed.percentiles());
+        }
+
+        if (classLevelTimed != null) {
+            builder.tags(classLevelTimed.extraTags());
+        }
+        if (methodLevelTimed != null) {
+            builder.tags(methodLevelTimed.extraTags());
+        }
+        builder.tags(timedTagsResolver.apply(invocation));
```
arogulin: Currently, if method-level annotation has an extra tag with name `method`, it will be `added` to the list of tags, producing 2 tags with the same name, which leads to an error on `influxdb` side. Is it possible to merge the tags in a way that method-level tags will `override` the tags defined by `TimedMethodInterceptor`? It will probably require a change in type of io.micrometer.core.instrument.Timer.Builder#tags variable from `List` to `Set`.

gavlyukovskiy: I can use set here to ensure tags uniqueness, but I think it's better to change builder and/or `Meter.Id` where all builders supply tags. 
I think problem is in `Meter.Id`:
```
this.tags = Collections.unmodifiableList(stream(tags.spliterator(), false)
                .sorted(Comparator.comparing(Tag::getKey))
                .distinct()
                .collect(Collectors.toList()));
```
here `distinct()` checks both key and value, not only key. @jkschneider what do you think?

theigl:
@jkschneider Since 1.1.0 has already been released, do you have any idea when this PR will make it to the master?

jkschneider:
@theigl Look for an announcement coming out soon explaining a move to a faster minor release cadence with LTS and non-LTS minors.

huehnerlady:
@jkschneider where can I find this announcement and what does that mean for this particular issue? When will it be released?

micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedMethodInterceptor.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+        else if (classLevelTimed != null && !classLevelTimed.value().isEmpty()){
+            metricName = classLevelTimed.value();
+        }
+        else {
+            metricName = TimedAspect.DEFAULT_METRIC_NAME;
+        }
+        Builder builder = Timer.builder(timedMetricNameResolver.apply(metricName, invocation));
+        if (methodLevelTimed != null) {
+            builder.description(methodLevelTimed.description().isEmpty() ? null : methodLevelTimed.description())
```
OleksiiBondar: If-else statement looks exactly the same except source timed class. Can be extracted into separated method where `Timed` passed as argument. Or alternatively, having local variable like: 

```
        Timed sourceTimed = methodLevelTimed ! = null ? methodLevelTimed : classLevelTimed;
        builder.description(sourceTimed.description().isEmpty() ? null : sourceTimed.description())
        builder.publishPercentileHistogram(sourceTimed.histogram())
        builder.publishPercentiles(sourceTimed.percentiles().length == 0 ? null : methodLevelTimed.percentiles());
```

micrometer-spring-legacy/src/main/java/io/micrometer/spring/advice/TimedAnnotationAdvisor.java
```diff
@@ -0,0 +1,52 @@
+package io.micrometer.spring.advice;
```
OleksiiBondar: please add copyrights across all newly added classes

micrometer-spring-legacy/src/test/java/io/micrometer/spring/advice/TimedMethodInterceptorTest.java
```diff
@@ -0,0 +1,239 @@
...
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.EnableAspectJAutoProxy;
+import org.springframework.context.annotation.Import;
+import org.springframework.stereotype.Service;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class TimedMethodInterceptorTest {
```
OleksiiBondar: Should this be public as per JUnit 4 conventions? As for sake of consistency with other tests within this module?

binkley:
Privet @gavlyukovskiy !

I found this concern with the PR: In a Spring Boot context, when I enable the bean for `TimedMethodInterceptor`, I lose the ability to configure histograms with Spring configuration (`application.yml`), including for metrics not using `@Timed` (eg, `http.server.requests`).  This is evidenced by the lack of `http_server_requests_seconds_bucket` values when using Prometheus to export.

This project commit illustrates my troubles: https://github.com/binkley/spikes/tree/a8ffb2433fa26596be6da269394f0347c30aea71/spring-boot-logging

```
$ cd spring-boot-logging
$ ./gradlew clean bootJar && java -Dloggy.run-once=false -jar build/libs/spring-boot-logging-0.0.1-SNAPSHOT.jar
```

After logging says "DONE!", from another terminal try:

```
http :8080/actuator/prometheus | grep http | grep bucket
```

(Obviously `curl` works as well.)

To see buckets, comment out the bean for `TimedMethodInterceptor` in `src/main/java/x/loggy/FeignConfiguration.java`.

gavlyukovskiy:
@binkley it's been a while since I touched this code... I looked through your examples and it looks like the declaration of `TimedMethodInterceptor` as a bean triggers `PrometheusMeterRegistry` instantiation before `MeterRegistryPostProcessor` instantiation, it results in `PrometheusMeterRegistry` not being post processed so that all customizers, binders and filters (one of which applies percentiles) are not applied to the registry. 
It first looked like a Spring Boot/Framework bug to me, because somehow defining `Advice` bean that requires `MeterRegistry` in the context breaks auto-configuration order specified on `PrometheusMetricsExportAutoConfiguration`:
```
@AutoConfigureAfter(MetricsAutoConfiguration.class)
public class PrometheusMetricsExportAutoConfiguration {
   ...
}
```
But then I tracked it down to `AnnotationAwareAspectJAutoProxyCreator` which requests all advices (?) to possibly instrument all future `BeanPostProcessor`'s (??). Maybe someone more familiar with framework can comment on this. @wilkinsona I saw you masterful explanations on such issues 😃 

As a workaround you can lazily request `MeterRegistry` once method fires, but probably there is a better way. Anyway this PR doesn't seem to be moving anywhere so I don't see a reason updating it.

binkley:
@gavlyukovskiy Thanks for the quick reply!

I tried a different route, and now wonder what the PR is addressing (I did read through, but do not have enough context to form an opinion).

I read through https://micrometer.io/docs/concepts#_the_timed_annotation, and used the "incubating" `TimedAspect` bean.  Everything worked as I had hoped:

- It timed methods (not classes -- so there's a reason to use your PR) in classes and interfaces
- Worked through proxied interfaces (such as Feign and Spring Data repositories)
- Respected configuration in `application.yml` for `management.metrics...`

My takeaway is that this PR supports class-level use `@Timed`.

berngp:
@jkschneider any thoughts on supporting class level annotation for `@Timed`? think there a few PRs opened now with this intent but its unclear to me which is the suggested approach.

* https://github.com/micrometer-metrics/micrometer/pull/1487
* https://github.com/micrometer-metrics/micrometer/pull/1279

shakuzen:
Thank you for the pull request, and sorry for us sitting on this for literal years. At this point, the micrometer-spring-legacy module has been deprecated and removed. There is no more Spring-specific instrumentation in this repository. Something Spring specific like this would need to be proposed to Spring Boot instead. Would you be willing to reach out to the Boot team to see their thoughts on incorporating the functionality offered here?

PR #505, opened 2018-03-19
reworking Hystrix binder  …

### LINKED ISSUE #503: Collection of executions for Hystrix
In #240 counters were implemented for Hystrix. Part of the implementation was taken for a library I maintain.

I found the previous counter implementation difficult to use, therefore it changed in a recent release: https://github.com/ahus1/prometheus-hystrix/issues/19

TL;DR: The total and error counters are dropped and a _terminal_ tag is added to each event type. This allows for more meaningful aggregations than the previous setup.

Feedback welcome. I'm happy to submit a PR. 

cc: @jkschneider, @checketts, @fiunchinho

ahus1:
to re-use counters retrieved from registry, Updated tests for integrative testing. Adding terminal tag to separate different events. Tests are now re-activated and seem to be stable (both local an on CI).

fixes #503


ahus1:
I know this changes several metrics names, but I think it is difficult to group the events in groups of "success" and "failure", as "fallback_success" might be a success for some, but not for others. And a "short_circuited" might not be a failure as it never touched the remove side. Therefore it's for the observer to decide on how to group these.

The one thing that Hystrix provides is the following: every execution leads to the increment of exactly one terminal event counter. Other non-terminal event counters might be updated on the way.

It also creates (or look up) the counters only once from the registry. This should avoid GC and runtime overhead, but I don't know what this is the way to do it within micrometer.io.

As always, feedback is welcome.

An example set of metrics published by Micrometer to Prometheus now looks like this in one of my example applications: 

````
# HELP hystrix_circuit_breaker_open  
# TYPE hystrix_circuit_breaker_open gauge
hystrix_circuit_breaker_open{group="interfaceOne",key="externalCall",} 0.0
# HELP hystrix_latency_total_seconds  
# TYPE hystrix_latency_total_seconds summary
hystrix_latency_total_seconds_count{group="interfaceOne",key="externalCall",} 1.0
hystrix_latency_total_seconds_sum{group="interfaceOne",key="externalCall",} 0.059
hystrix_latency_total_seconds_max{group="interfaceOne",key="externalCall",} 0.059
# HELP hystrix_latency_execution_seconds  
# TYPE hystrix_latency_execution_seconds summary
hystrix_latency_execution_seconds_count{group="interfaceOne",key="externalCall",} 1.0
hystrix_latency_execution_seconds_sum{group="interfaceOne",key="externalCall",} 0.055
hystrix_latency_execution_seconds_max{group="interfaceOne",key="externalCall",} 0.055
# HELP hystrix_threadpool_concurrent_execution_rolling_max  
# TYPE hystrix_threadpool_concurrent_execution_rolling_max gauge
hystrix_threadpool_concurrent_execution_rolling_max{group="interfaceOne",key="externalCall",threadpool="interfaceOne",} 0.0
# HELP hystrix_threadpool_concurrent_execution_current  
# TYPE hystrix_threadpool_concurrent_execution_current gauge
hystrix_threadpool_concurrent_execution_current{group="interfaceOne",key="externalCall",threadpool="interfaceOne",} 0.0
# HELP hystrix_execution_total Execution results. See https://github.com/Netflix/Hystrix/wiki/Metrics-and-Monitoring#command-execution-event-types-comnetflixhystrixhystrixeventtype for type definitions
# TYPE hystrix_execution_total counter
hystrix_execution_total{event="emit",group="interfaceOne",key="externalCall",terminal="false",} 0.0
hystrix_execution_total{event="success",group="interfaceOne",key="externalCall",terminal="true",} 0.0
hystrix_execution_total{event="failure",group="interfaceOne",key="externalCall",terminal="false",} 1.0
hystrix_execution_total{event="timeout",group="interfaceOne",key="externalCall",terminal="false",} 0.0
hystrix_execution_total{event="bad_request",group="interfaceOne",key="externalCall",terminal="true",} 0.0
hystrix_execution_total{event="short_circuited",group="interfaceOne",key="externalCall",terminal="false",} 0.0
hystrix_execution_total{event="thread_pool_rejected",group="interfaceOne",key="externalCall",terminal="false",} 0.0
hystrix_execution_total{event="semaphore_rejected",group="interfaceOne",key="externalCall",terminal="false",} 0.0
hystrix_execution_total{event="fallback_emit",group="interfaceOne",key="externalCall",terminal="false",} 0.0
hystrix_execution_total{event="fallback_success",group="interfaceOne",key="externalCall",terminal="true",} 0.0
hystrix_execution_total{event="fallback_failure",group="interfaceOne",key="externalCall",terminal="true",} 1.0
hystrix_execution_total{event="fallback_rejection",group="interfaceOne",key="externalCall",terminal="true",} 0.0
hystrix_execution_total{event="fallback_missing",group="interfaceOne",key="externalCall",terminal="true",} 0.0
hystrix_execution_total{event="exception_thrown",group="interfaceOne",key="externalCall",terminal="false",} 1.0
hystrix_execution_total{event="response_from_cache",group="interfaceOne",key="externalCall",terminal="true",} 0.0
hystrix_execution_total{event="cancelled",group="interfaceOne",key="externalCall",terminal="true",} 0.0
hystrix_execution_total{event="collapsed",group="interfaceOne",key="externalCall",terminal="false",} 0.0
hystrix_execution_total{event="command_max_active",group="interfaceOne",key="externalCall",terminal="false",} 0.0
````

File: micrometer-core/src/main/java/io/micrometer/core/instrument/binder/hystrix/MicrometerMetricsPublisherCommand.java [hunk 1]
```diff
@@ -133,64 +103,23 @@ public void initialize() {
...
-                        }
-
-                        if (executionEvents.contains(hystrixEventType)) {
-                            getExecutionCounter(hystrixEventType).increment(count);
-                        } else if (fallbackEvents.contains(hystrixEventType)) {
-                            getFallbackCounter(hystrixEventType).increment(count);
-                        } else {
-                            getOtherExecutionCounter(hystrixEventType).increment(count);
-                        }
+                        eventCounters.get(hystrixEventType).increment(count);
```
jkschneider: You can benchmark it if you'd like, but just registering the counter here creates one if it doesn't exist, retrieves an existing one of it does. The cost of doing the map lookup in `eventCounters` is comparable then to the cost of calling `register`.

ahus1: I'm creating the counters at be beginning of the method anyway as it is a good practice to initialize all metrics for a command with zero to ease the aggregations/graphs afterwards.

This was already in the previous code. I update the PR to also keep the comment that was there in the previous version.

Therefore I have the meter already at hand, therefore I think it's good to use it. It also avoid to re-create the list of tags including the toLowerCase() operations. 

Therefore I'd rather keep it in the hashmap.

jkschneider: Great. Thanks.

checketts: I like how this simplifies the metric incrementing and also that they are all still set to zero initially. 

I don't feel comfortable about changing the metric breakdown though. `hystrix.execution` represented more or less how you want the 'terminal' tag to appear in that the tags all represented parts of the whole.  This change will allow duplicate counts and the user would need to know the specific tag to exclude on. I would feel better if those non-terminal executions were counted separately, under a separate name.  

We found that users of our Prometheus/Grafana dashboard got buried in the options and wanted to be able to answer specific questions:

  1. Are any of my circuits open? 
      - Un-changed
  1. Is my Hystrix code being exercised (as in is the path being used) 
      - Also unchanged, success indicates that path is being used
  2. Are fallbacks being triggered? 
      - This is a little more difficult, But achievable for Prometheus, using the regex selector on tags, but I don't want to for users to look for it. Also I don't think filtering in the same way is as easy for other backends.
  3. If they are being triggered, what is the outcome of the fallback being triggered? 
      - Same as the last point, but it is interested in filtering and displaying the results just for fallbacks

As I've mentioned before, my Hystrix experience is still pretty minimal. And the questions my dashboard were addressing probably reflect that (essentially the 'is it configured to work?' data)

Perhaps most Hystrix users would want to answer the question of 'How was my request handled in the end?' That matches with the semantics around it being `terminal`. Is it correct to say that question was impossible to answer before (or at least much more difficult)?

jkschneider: @checketts A grafana dash export of your hystrix setup would go far in explanatory power. Would you mind dropping that somewhere?

ahus1: @checketts thank you for your extensive comment. It took my some days to follow up due to ongoing work projects and conferences. 

I write here to find out what a counter about fallbacks could contain. The docs are here: https://github.com/Netflix/Hystrix/wiki/Metrics-and-Monitoring#hystrix-event-types

You phrased the user's question as: "Are fallbacks being triggered"
Looking at all events that start with "fallback":
* fallback_emit -> non-terminal, should be ignored -> "no"
* fallback_success -> the fallback was triggered, and successful) -> "yes"
* fallback_failure -> the fallback was triggered, but was unsuccessful -> "yes"
* fallback_rejection -> the fallback was not attempted due to capacity problems -> "no" in a "triggered sense"? Or "yes" anyway?
* fallback_missing -> fallback doesn't exist -> "no" 

If you could help to agree on the yes/no we could wrap that into a counter.

IMHO the Hystrix events break things down into tiny pieces that are non-trivial to put together. The result depends on the question asked.

My personal conclusion with Hystrix events: the only question that can be answered straight forward is the following: how do the events relate to the total number of commands I sent. This will give you a breakdown like: 
* 50% SUCCESS (terminal)
* 50% FAILURE (non-terminal)
* 25% FALLBACK_SUCCESS (terminal)
* 25% FALLBACK_FAILURE (terminal)

Again, all terminal events sum up to 100%. Adding a counter that represents the sum of all terminal events would ease the calculation.

As a prometheus query this would look like (switch irate for rate as preferred):

````
irate(hystrix_command_event_total [1m])
/ ignoring(event) group_left
sum(irate(hystrix_command_event_total{terminal='true'} [1m]))
  by (command_group, command_name, instance, job, terminal)
````

If you want to ignore the events that never happened, add the following: 

````
... AND hystrix_command_event_total > 0 
````

Again, adding a counter that represents the sum of all terminal events would ease the calculation. I propose it to be called "hystrix_command_event_terminal_total"

ahus1: @checketts - Maybe we can take this step by step. Let me know where we agree and I will split this into two PRs. We can also have a face-to-face (skype/hangout?) for the second part.

jkschneider: @ahus1 @checketts. Or if this PR represents an improvement over the status quo we can merge and iterate further?

ahus1: To make voting easier: I could rewrite it to keep all old metrics, and just add new metrics/tags. Some metrics will be deprecated (noted in the description). A switch (default off) will allow to skip all deprecated metrics. This will allow easy migration for existing users. This switch will require some documentation.

ahus1:
a short update: I pushed some changes, it's still work in progress. You can now register multiple HystrixMetricsPublisherCommand instances and cascade them. 

I went with "Deprecated10x" as a class name suffix and also as package name (in all lowercase characters). There is also a new property "management.metrics.binders.hystrixDeprecatedMetrics10x.enabled" that is enabled by default and shows all the deprecated metrics. 

Comments on naming welcome.

ahus1:
This is no longer work in progress. The changes are complete from my point of view. They have been rebased and squashed. 

The deprecated metrics can be added using the following piece of code (already included in spring-legacy, but needs to be added manually for spring boot 2.0 application as there is no autoconfiguration in micrometer itself in spring boot 2.0 AFAIK)

    @Bean
    @ConditionalOnClass(name = "com.netflix.hystrix.strategy.HystrixPlugins")
    @ConditionalOnProperty(value = "management.metrics.binders.hystrixDeprecatedMetrics10x.enabled", matchIfMissing = true)
    public HystrixMetricsBinderDeprecated10x hystrixMetricsBinderDeprecatedMetrics10x() {
        return new HystrixMetricsBinderDeprecated10x();
    }

Please comment, test, reject, or merge.

spencergibb:
The auto-configuration in boot 2 land lives in spring-cloud-netflix

checketts:
I'm good with this moving forward. Unfortunately I'm not really in a place to give too much more feedback now-a-days, since I'm not in a hystrix-enabled codebase for the time being. 

Thanks for the updates @ahus1 and sorry it has languished for so long!

jkschneider:
Removing deprecated metrics. Since this is going in 1.1.0, I think it is reasonable to expect that dashboards can change accordingly.

PR #507, opened 2018-03-20
Apply Eclipse Gradle plugin

mweirauch:
Helps setting up the missing JMH dependencies (jmh, colt).
Now I don't need to re-add them to the classpath after
importing or updating the Gradle project.

Tested with Eclipse Oxygen SR2 (4.7.2).

Assumption: No influence on IDEA users. If somebody could give it a try?


jkschneider:
Does the restructuring of the benchmark code in dce54264eb9533ae8d9f40a29b65de33048b278e obviate the need for this?

mweirauch:
That does indeed solve it. Thanks!

PR #510, opened 2018-03-22
prometheus java client 0.3.0

pjfanning:
Missed these when I did https://github.com/micrometer-metrics/micrometer/pull/481

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,10 @@ allprojects {
         println "I'm building $project.name with version $project.version"
     }
     group = 'io.micrometer'
+
+    ext {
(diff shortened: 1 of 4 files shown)


checketts:
Should they be extracted into a property?

checketts:
Thanks!

jkschneider:
If it's OK, I'd like to stick to just using the dependency lock mechanism, as that's the purpose of it. Periodically we can just run `./gradlew updateLocks`, test again, and we've updated our dependencies.

You did identify some places where we had incorrectly specified first order dependencies on fixed versions though, and that has been incorporated in e6c114c8.

PR #514, opened 2018-03-23
Fixed typo in hystrix binding conditional

gavlyukovskiy:
Changed `management.metrics.export.hystrix.enabled` to `management.metrics.binders.hystrix.enabled` to follow general convention.


jkschneider:
Fixed in 5abd05c4. Good catch @gavlyukovskiy!

PR #523, opened 2018-03-27
Seperate _max metric of histograms and summaries to gauge to fix #519

### LINKED ISSUE #519: Timer prometheus metric _max is not according to Prometheus spec
The current [prometheus implementation](https://github.com/micrometer-metrics/micrometer/blob/e30323829185dcd016ad9291316a1f54cb2615a9/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L215) for the Timer metric is not according to the Prometheus specification for a summary metric.

This results in issues with the official prometheus client libraries and should be moved to a seperate gauge metric:

https://github.com/prometheus/client_python/pull/264

Is that a known problem, or rather, is there any interest for a fix? We might be able to offer a PR in the coming days.

marcoschulte:
Fix for #519


marcoschulte:
Just noticed this fix duplicates the "# TYPE" and "# HELP" comments. Closed this PR for now

Issue #554, opened 2018-04-12
Add configuration property to disable MetricsRestTemplateCustomizer


mkrichevsky:
I think it would be more convenient way of disabling

jkschneider:
I think the Spring Boot team is going to point out that you can do this with an exclude:

```java
@SpringBootApplication(exclude = RestTemplateMetricsAutoConfiguration.class)
```

It's unlikely that they'll add another property additionally.

mkrichevsky:
Can micrometer has its own property in `RestTemplateMetricsConfiguration`?
I think it's a little bit inconsistent that some meters can be disabled by properties and the others by exclusion in annotations.
Maybe I'm missing something here?

jkschneider:
There is only one metric coming out of rest template metrics: `http.client.requests`. So disabling it just requires:

```
management.metrics.enable[http.client]=false
```

The binders, OTOH create several metrics, which is why I suppose they added a single disablement property. But even that was probably unnecessary as all OOTB binders use a consistent name prefix internally.

```
management.metrics.enable[hystrix]=false
management.metrics.enable[jvm.gc]=false
management.metrics.enable[jvm.memory]=false // notice you could collapse this with just [jvm]
```

mkrichevsky:
I can tell about `logback` meter, as I debugged the behavior with different configurations:
With `management.metrics.enable[logback]=false` I indeed don't have any metrics about logs, but [decide method](https://github.com/micrometer-metrics/micrometer/blob/09f3ea7963da6135cd253f6e26de8a910273c8fa/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java#L126) is still invoked on every log.
However, with `management.metrics.binders.logback.enabled=false` property the [bean](https://github.com/micrometer-metrics/micrometer/blob/09f3ea7963da6135cd253f6e26de8a910273c8fa/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/MeterBindersConfiguration.java#L73) is not created at all, thus the `decide` method is never called.
I don't know why the behavior is different. If you want, tomorrow I can debug other binders. Or maybe you know why this happens?

mkrichevsky:
Tbh, I would prefer the ability to configure through properties. (I like how `endpoints` configuration is done, where I can disable all of them and enable only those that I want/need).
But I'm not sure how easy it would be to implement the same here and if it would make sense to do such thing at all.
Any other thoughts?

[...]

PR #556, opened 2018-04-13
added properties to enable/disable scheduled aspect and rest client

mkrichevsky:
Following question [#554](https://github.com/micrometer-metrics/micrometer/issues/554) I added some properties to be able to disable certain binders. Tried following the convention of naming properties for other binders


jkschneider:
I would accept a revised PR that turns the rest client metrics configuration into an autoconfiguration class that matches what is in Spring Boot 2 so that it can be disabled via the normal property-based autoconfiguration disablement mechanism of Spring Boot.

Adding these properties has no chance of being accepted in Spring Boot 2, and I think we'd like to keep the legacy module as close as possible to Spring Boot 2 to ease migration.

mkrichevsky:
@jkschneider I see what you mean. What about scheduled aspect?

Issue #272, opened 2017-12-06
Support dynatrace


jkschneider:
https://www.dynatrace.com/support/help/dynatrace-api/custom-devices-and-metrics/what-does-the-custom-network-devices-and-metrics-api-provide/

jkschneider:
Also uses the API directly: https://github.com/Dynatrace/dynatrace-cli

jkschneider:
Shows use of custom metrics: https://www.youtube.com/watch?v=nwK9T22uQac

obarcelonap:
@jkschneider is this subject open for contribution?

jkschneider:
@obarcelonap It sure is!

obarcelonap:
Starting to look into the topic. Some interesting resources below.
Blog post:
https://www.dynatrace.com/news/blog/register-custom-network-devices-smartscape-topology/
Samples:
https://github.com/Dynatrace/dynatrace-api/tree/master/custom-device-metrics-F5-example
https://github.com/Dynatrace/dynatrace-api/tree/master/custom-device-metrics-Raspberry-example

obarcelonap:
@jkschneider did some progress on this, mainly creating all the boiler plate and defining minimal configuration properties.
Branch here:
https://github.com/micrometer-metrics/micrometer/compare/master...obarcelonap:dynatrace-support
Next week I'll start to work in the meter registry, I'll keep you posted!

jkschneider:
Looking good!

grabnerandi:
Hi. Just got off a call with Josh Long who highlighted this project. I am the Dev Advocate from Dynatrace and happy to help with any questions you have around the APIs. Seems you already found the Timeseries API which is great to push data. Soon we will provide other options to pull data into Dynatrace if that is of interest.

obarcelonap:
Hi guys, I've been working in the topic, the progress can be checked in the branch mentioned to comments above: https://github.com/micrometer-metrics/micrometer/compare/master...obarcelonap:dynatrace-support

I've implemented all the bolierplate code needed for the implementation and also part of the meter registry, which has the following behavior:
1. For each meter creates the custom metric using `https://{your-environment-id}.live.dynatrace.com/api/v1/timeseries/<metric_ID>`. The names of metrics created through the API are kept in an in-memory collection so this action would be only executed the first time the publish function is executed.
2. For each meter builds the json body (series) of the metrics to be pushed into the API. 
[...]

PR #558, opened 2018-04-15
Dynatrace support

obarcelonap:
Adds a new meter registry which enables support for Dynatrace's custom metrics (#272)


jkschneider:
Merged with a5272df.

PR #559, opened 2018-04-16
better performance of creating web mvc tags

mkrichevsky:
Moving some tags to constants instead of creating every time should be obvious.
Creation of response status is slightly faster, as we're not casting anymore. More details in [benchmark](https://gist.github.com/mkrichevsky/ade345b10f1540c6a7a84f78e9248991) and [results](https://gist.github.com/mkrichevsky/c5f8ac70509c220ce2a38bb18149da65)


jkschneider:
@mkrichevsky Could you do me a favor and rebase this against 1.0.x?

mkrichevsky:
@jkschneider done :)

jkschneider:
Interestingly, I think your benchmark demonstrates that there is no real difference between casting the int and `Integer.toString(..)`, but the code is more succinct anyway.

PR #560, opened 2018-04-17
Fixed: Pass percentile precision to child Timer and DistributionSummary

kkamenev:
When creating a timer with builder API (as below), `precisionDigits` is not passed to child meters

```
        Timer.builder("my_timer")
            .percentilePrecision(2)
            .register(Metrics.globalRegistry);
```


jkschneider:
Good catch! Thanks @kkamenev.

PR #565, opened 2018-04-18
Kafka consumer metrics

wardhapk:


manderson23:
Do you intend to also support producer and streams metrics?

lorenzo-nutmeg:
Why are all `FunctionCounter`? These metrics are not monotonically increasing. 
Shouldn't be `Gauge`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+    public void bindTo(MeterRegistry reg) {
+        registerConsumerFetchMetrics(reg);
+        registerConsumerCoordinatorMetrics(reg);
+    }
+
+    private void registerConsumerFetchMetrics(MeterRegistry registry) {
+
+        registerMetricsEventually("type", "consumer-fetch-manager-metrics", (name, allTags) -> {
+
+                FunctionCounter.builder("kafka.records.lag.max", mBeanServer,
```
jkschneider: Should be `Gauge`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+                    s -> safeDouble(() -> s.getAttribute(name, "records-lag-max")))
+                    .tags(allTags)
+                    .register(registry);
+
+                FunctionCounter.builder("kafka.fetch.latency.avg", mBeanServer,
```
jkschneider: Should be `Gauge`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+                    s -> safeDouble(() -> s.getAttribute(name, "fetch-latency-avg")))
+                    .tags(allTags)
+                    .register(registry);
+
+                FunctionCounter.builder("kafka.bytes.consumed.rate", mBeanServer,
```
jkschneider: Should be `Gauge`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+                    s -> safeDouble(() -> s.getAttribute(name, "bytes-consumed-rate")))
+                    .tags(allTags)
+                    .baseUnit("bytes")
+                    .register(registry);
+
+                FunctionCounter.builder("kafka.fetch.size.max", mBeanServer,
```
jkschneider: Should be `Gauge`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+                    s -> safeDouble(() -> s.getAttribute(name, "fetch-size-max")))
+                    .tags(allTags)
+                    .baseUnit("bytes")
+                    .register(registry);
+
+                FunctionCounter.builder("kafka.records.consumed.rate", mBeanServer,
```
jkschneider: Should be `Gauge`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+                    .register(registry);
+            }
+        );
+    }
+
+    private void registerConsumerCoordinatorMetrics(MeterRegistry registry) {
+
+        registerMetricsEventually("type", "consumer-coordinator-metrics", (name, allTags) -> {
+
+                FunctionCounter.builder("kafka.assigned.partitions", mBeanServer,
```
jkschneider: Should be `Gauge`. Maybe?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+                    s -> safeDouble(() -> s.getAttribute(name, "assigned-partitions")))
+                    .tags(allTags)
+                    .register(registry);
+
+                FunctionCounter.builder("kafka.commit.latency.avg", mBeanServer,
```
jkschneider: Should be `Gauge`.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+                    s -> safeDouble(() -> s.getAttribute(name, "commit-latency-avg")))
+                    .tags(allTags)
+                    .register(registry);
+
+                FunctionCounter.builder("kafka.commit.latency.max", mBeanServer,
```
jkschneider: Should be `Gauge`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,164 @@
...
+                    s -> safeDouble(() -> s.getAttribute(name, "commit-latency-max")))
+                    .tags(allTags)
+                    .register(registry);
+
+                FunctionCounter.builder("kafka.commit.rate", mBeanServer,
```
jkschneider: Should be `Gauge`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -0,0 +1,182 @@
...
+                    .baseUnit("bytes")
+                    .register(registry);
+
+                Gauge.builder("kafka.fetch.size.max", mBeanServer,
+                    s -> safeDouble(() -> s.getAttribute(name, "fetch-size-max")))
+                    .tags(allTags)
+                    .baseUnit("bytes")
+                    .register(registry);
+
+                Gauge.builder("kafka.records.consumed.rate", mBeanServer,
```
pobodysnerfect: kafka publishes "rate" values as well as corresponding "total" values: 

https://kafka.apache.org/documentation/#monitoring

The rate captures the per/second  rate whereas the "total" is a running total (monotonically increasing). the problem with using the per second rates with a gauge is that you miss data, if you have a burst for a couple seconds that may not be reflected in the gauge stats. so i think in this case using the "total" values with a function counter actually makes more sense (this applies to all rate values below as well).

wardhapk: Checking this further..

quaff:
@wardhapk  `safeDouble` is not safe, you should skip `NaN` and `Infinite`
```
	private double safeDouble(Callable<Object> callable) {
		try {
			double value = Double.parseDouble(callable.call().toString());
			if (Double.isNaN(value) || Double.isInfinite(value)) {
				return 0.0;
			}
			return value;
		} catch (Exception e) {
			return 0.0;
		}
	}
```

```
i.m.i.InfluxMeterRegistry ERROR failed to send metrics: {"error":"partial write: unable to parse 'kafka.commit.latency.max,consumer=\\\"test-0\\\",topic=all,metric_type=gauge value=-∞ 1533540268845': invalid number dropped=0"}
```
And should consumer tag strip double quotes? should topic tag absent instead of "all"?

wardhapk:
@quaff Converting NaN and Infinite to zeros leads to misrepresentation of the data, for instance when ``records-lag-max`` is converted to zero, it tells that the lag is zero when it actually is not. These values should be left as is and is up to the metrics consumer to interpret/plot according to the specific use case. 
The topic tags are as it is published from kafka itself, it make sense to keep the tags, if you are consuming from multiple topics from the same consumer thread. This gives you the ability to expose a per topic level metrics and, also at a per consumer level using the "all" tag.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
jkschneider: Wherever it is possible to use a total rather than a rate coming from Kafka, we should do that. Micrometer already ships rates for monotonically increasing counters for those monitoring systems that don't provide query-side rate functions.

jkschneider:
> you'd also want to change the attributes to bytes-consumed-total, records-consumed-total, and commit-total respectively

`bytes.consumed`, `records.consumed` would be preferable. In the case of Prometheus, `total` will be appended to the name as part of the naming convention. But most monitoring systems don't accept cumulative counts like Prometheus, so the name would be misleading elsewhere.

rammygit:
should this binder should have kafka.consumer:type=consumer-metrics ? as this is also part of consumer apart from fetch metrics and coordinator metrics.

wardhapk:
@jkschneider - are there any additional changes you would need on this?

rammygit:
@wardhapk are you adding this "kafka.consumer:type=consumer-metrics" ?

wardhapk:
@rammygit What all metrics under consumer-metrics would you be interested in?

rammygit:
@wardhapk  Thanks for asking, I am looking for connection-count, request-rate, response rate, incoming-byte-rate, outgoing-byte-rate. It will be nice to have them there. Thanks

jkschneider:
Merged with bd2b4aa7. There were quite a few changes as you can see... Metrics have been carefully separated to determine which ones were gauges, function counters, and time gauges.

Added consumer-metrics like @rammygit suggests.

Also, I added all the consumer metrics that were added in Kafka 2 but not documented anywhere ;). These metrics have a version guard so the binder is backwards compatible to Kafka 0.11.0.

PR #568, opened 2018-04-20
Remove StatsdConfig.queueSize()

izeye:
This PR removes `StatsdConfig.queueSize()` as it's not used since d1edf7d907d88f30bd8f26cebdafb1ded18945e5.


jkschneider:
I suppose we should deprecate instead. Technically this would be a breaking change, though clearly no value is derived from it.

izeye:
@jkschneider Thanks for the feedback! I created another PR for deprecation against the 1.0.x branch.

shakuzen:
@izeye can you rebase this on 2.0.x now that we have a branch for 2.0 work? Thanks!

izeye:
@shakuzen Thanks for letting me know! I rebased this on the `2.0.x` branch.

jonatan-ivanov:
Thanks! :)

PR #570, opened 2018-04-22
Suppress Java compiler warnings

izeye:
This PR changes to use `@SuppressWarnings` to suppress Java compiler unchecked warnings as `noinspection` line comment works only for IntelliJ.


jkschneider:
Feel free to commit these directly to master ;)

Issue #577, opened 2018-04-24
[Feature Request] Add Postgres Meter Binder


kristofdepypere:
Postgres collects out of the box very relevant statistics to monitor the performance of the database. It is a matter of doing the correct queries and exposing these statistics the correct way in order to support postgres statistics through micrometer. 

I think the following article is a very nice summary of the postgres metrics that are very important and I think we should use it as a guideline to collect the correct metrics: https://www.datadoghq.com/blog/postgresql-monitoring/

izeye:
This looks resolved via #578 .

shakuzen:
Thanks @izeye. Duplicates #578

PR #578, opened 2018-04-24
Add Postgres meter binder

kristofdepypere:
This PR provides the functionality for ticket #577 to have correct monitoring for PostgreSQL using Micrometer

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/jdbc/db/PostgreSQLDatabaseMetrics.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/jdbc/db/PostgreSQLDatabaseMetrics.java
@@ -83,7 +83,7 @@ public class PostgreSQLDatabaseMetrics implements MeterBinder {
             .description("Percentage of blocks in this database that were shared buffer hits vs. read from disk")
             .register(registry);
         FunctionCounter.builder("postgres.transactions", postgresDataSource,
-            dataSource -> refreshableFunctionalCounter("postgres.transactions", this::getTransactionCount))
+            dataSource -> resettableFunctionalCounter("postgres.transactions", this::getTransactionCount))
             .tags(tags)
             .description("Total number of transactions executed (commits + rollbacks)")
             .register(registry);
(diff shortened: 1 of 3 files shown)


checketts:
I like these metrics. However, they might be best run as a sidecar to your database though since otherwise if you have 10 servers talking to you DB you'll get this metric 10 times. (That is how the [Prometheus mysql exporter](https://github.com/prometheus/mysqld_exporter) is setup)

kristofdepypere:
That's true.. In our case we actually have a 'agent' webapp running next to our postgres instances to manage them. So we can integrate it in that location. For now we like the flexibility of choosing our target reporting tool cloud vs on premise so we don't focus on prometheus specific solutions yet. I can understand this is perhaps not a good candidate to include in the lib.

checketts:
I'm not saying they aren't a good candidate for the lib, but rather that we probably wouldn't want to add an auto-configuration for them.

jkschneider:
Still a useful PR, I think, and something we should add.

1. The sidecar may actually just be a Micrometer-enabled Java process.
2. To a point, it doesn't really matter if multiple instances are reporting the same stats. Surely Postgres can handle it or there are going to be problems elsewhere where the app is interacting with the DB.

A few things to think about @kristofdepypere:

1. Many of the meters that are reported as `Gauge` right now should be `FunctionCounter`, since the counts are generally increasing over time. By using a `FunctionCounter` instead, we can reason about the number of inserts in an interval in a way that is portable to all monitoring systems (e.g. Datadog has a `rate` function, but many monitoring systems do not). Examples include rows `fetched`, `inserted`, `updated`. In contrast, connection count _is_ certainly a `Gauge`, as that goes up and down.

2. `runQuery` can return a null value right now, and its result is being boxed to a `long` in a few places. There is an NPE exposure here.

3. I think in every place where we call `runQuery(query, Integer.class)`, it should probably be `runQuery(query, Long.class)`. We should check on the relative maximums of a Java long and a Postgres bigint.

4. We need to compensate somehow for calls to the Postgres `pg_stat_reset()` function and friends.

kristofdepypere:
Ok, I'll update the PR to reflect the remarks. 

I think I'm also missing some crucial metrics still so I intend to add some more enhancements

kristofdepypere:
What still bothers me is the pg_stat_reset() compensation. The timestamp is stored in each table so I know when it's called, however what should be done with the FunctionCounters ? They will have to be reset in some way.

jkschneider:
@kristofdepypere Sorry, missed your latest comment earlier! I think we're going to somehow just have to locally keep track of the last value we saw prior to `pg_stat_reset()`, and if the latest call to it ever drops below the last value we knew of, keep accumulating the current value to the old.

kristofdepypere:
Added the safety net for `pg_stat_reset()` calls

jkschneider:
Merged with 4eb83398!

Made a few minor changes to metric names, and broke apart ratios into constituent parts.

PR #591, opened 2018-05-05
Upgrade to Gradle Wrapper 4.7

izeye:
Along the way this PR also resolves some of deprecation warnings.


jkschneider:
Thanks. Split this up a bit and merged most of it to 1.0.x, then fixed the annotationProcessor bit in the benchmark build file in a separate commit. The last commit on the master branch that completes this is d79f639a.

jkschneider:
Been working a bit on the build today anyway, so this was timely. Some updates to the release side for:

1. When OSSRH got slow, multiple maven central syncs from JCenter were getting added to the same staging repo and failing closure.
2. Hope we have a fix for daemon state issues that occasionally cause locks to fail.
3. Gradle 4.7 ;)
4. Publishing 1.0.x snapshots to repo.spring.io now.
5. Fixed ordering of upload to artifactory (POM always first now) to fix Artifactory's attribution of a source/javadoc jar to a prior snapshot.

PR #593, opened 2018-05-07
Add Jetty's ThreadPool metrics

matsumana:
I want to monitor whether Jetty's ThreadPool setting is sufficient so I'd like to add several metrics.

# Usage

Example In Spring Boot 2.0.x

```java
@Bean
public JettyServletWebServerFactory webServerFactory(MeterRegistry registry) {
    JettyServletWebServerFactory factory = new JettyServletWebServerFactory();
    tags = Collections.singletonList(Tag.of("id", "0"));
    factory.setThreadPool(new InstrumentedQueuedThreadPool(registry, tags));
    return factory;
}
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.jetty;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+public class JettyServerThreadPoolMetrics implements MeterBinder {
+
+    private final InstrumentedQueuedThreadPool threadPool;
+    private final Iterable<Tag> tags;
+
+    public JettyServerThreadPoolMetrics(InstrumentedQueuedThreadPool threadPool, Iterable<Tag> tags) {
+        this.threadPool = threadPool;
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        Gauge.builder("jetty.threads.min", threadPool, InstrumentedQueuedThreadPool::getMinThreads)
+             .tags(tags)
+             .description("The number of min threads")
+             .register(registry);
+        Gauge.builder("jetty.threads.max", threadPool, InstrumentedQueuedThreadPool::getMaxThreads)
+             .tags(tags)
+             .description("The number of max threads")
+             .register(registry);
+        Gauge.builder("jetty.threads.live", threadPool, InstrumentedQueuedThreadPool::getThreads)
+             .tags(tags)
+             .description("The current number of live threads")
+             .register(registry);
+        Gauge.builder("jetty.threads.idle", threadPool, InstrumentedQueuedThreadPool::getIdleThreads)
+             .tags(tags)
+             .description("The current number of idle threads")
+             .register(registry);
+        Gauge.builder("jetty.threads.busy", threadPool, InstrumentedQueuedThreadPool::getBusyThreads)
+             .tags(tags)
+             .description("The current number of busy threads")
+             .register(registry);
+    }
+}
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
@@ -40,9 +40,9 @@ public class JettyServerThreadPoolMetrics implements MeterBinder {
              .tags(tags)
              .description("The number of max threads")
              .register(registry);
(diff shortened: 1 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
```diff
@@ -0,0 +1,56 @@
...
+    public void bindTo(MeterRegistry registry) {
+        Gauge.builder("jetty.threads.min", threadPool, InstrumentedQueuedThreadPool::getMinThreads)
+             .tags(tags)
+             .description("The number of min threads")
+             .register(registry);
+        Gauge.builder("jetty.threads.max", threadPool, InstrumentedQueuedThreadPool::getMaxThreads)
+             .tags(tags)
+             .description("The number of max threads")
+             .register(registry);
+        Gauge.builder("jetty.threads.live", threadPool, InstrumentedQueuedThreadPool::getThreads)
```
checketts: How does this compare to the Tomcat equivalent? Wondering if `active` is a better name or if it differs from busy? Or is it merely `idle + busy`?

matsumana: I decided the name `live` with reference to JvmThreadMetrics.
https://github.com/micrometer-metrics/micrometer/blob/v1.0.3/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java#L57

For the meaning of the metric please refer to the Jetty's source.
https://github.com/eclipse/jetty.project/blob/jetty-9.4.8.v20171121/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java#L428-L455

mweirauch:
Just a thought: Is the `jetty.threads.idle` superfluous as it could be calculated from `jetty.threads.current` subtracted by `jetty.threads.busy`? Wonder who would graph the "idle" metric in a plot with "busy" and "current" and "max".

matsumana:
> Just a thought: Is the jetty.threads.idle superfluous as it could be calculated from jetty.threads.current subtracted by jetty.threads.busy?

@mweirauch 

Good point.
I'll delete `jetty.threads.idle` and unify metric name to the same it as TomcatMetrics.

- jetty.threads.config.max
- jetty.threads.busy
- jetty.threads.current

matsumana:
Updated

mweirauch:
Just tested against Spring Boot 1.5 with
```
	@Bean
	public JettyEmbeddedServletContainerFactory webServerFactory(MeterRegistry registry) {
		JettyEmbeddedServletContainerFactory factory = new JettyEmbeddedServletContainerFactory();
		factory.setThreadPool(new InstrumentedQueuedThreadPool(registry, Collections.emptyList()));
		return factory;
	}
```
LGTM.

Perhaps some auto-configuration for `micrometer-spring-legacy` and Spring Boot 2 should be tracked in follow up tickets?

matsumana:
@mweirauch Thank you for the confirmation.
I implemented `JettyMetricsConfiguration.

matsumana:
Would you merge this PR?
If this PR is merged, I'll create an issue about `JettyMetricsConfiguration` on Spring Boot's repo.

jkschneider:
Merged in 21a375a9d8b5fd6ab5758ea6df52465206be5650

mweirauch:
I fear the `JettyMetricsConfiguration::instrumentJettyThreadPool` is kicking in too late. `EmbeddedWebApplicationContext` is calling the factory before afformentioned method is triggered. 

Haven't found any other solution except actually defining the factory ourselfs. Dunno if the setup has changed with Spring Boot 2 and it becomes easier.

In case one really needs to roll his own factory and still wants threadpool instrumentation, one needs to exclude the `JettyMetricsConfiguration` from the auto-configuration. (e.g. with `spring.autoconfigure.exclude`)

matsumana:
@mweirauch Thank you for the opinion.
I'll investigate.

wilkinsona:
I'm looking at [incorporating this change into Spring Boot 2.0](https://github.com/spring-projects/spring-boot/issues/14591). What's the benefit of `InstrumentedQueuedThreadPool`? I'm concerned about the need to set Jetty's thread pool as it will overwrite a user's customisation of the thread pool. It also limits the usefulness of `JettyServerThreadPoolMetrids`. Could it not take a `QueuedThreadPool` or even a `ThreadPool`? It could then adapt the meters that it binds based on the pool's concrete type.

mweirauch:
So, perhaps we could actually ask for the `ThreadPool` in the current [JettyMetricsPostProcessor](https://github.com/micrometer-metrics/micrometer/blob/1399b07d78d5a599bb945ec7332dd791bf7cc612/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/web/jetty/JettyMetricsPostProcessor.java#L47).

Would it be allowed to register a new bean within `postProcessAfterInitialization()` on the held `ApplicationContext`? In this case a `ThreadPool`-adaptive variant of the `JettyServerThreadPoolMetrics` which can later be picked up by the `MeterRegistry` postprocessing where all `MeterBinder` are collected and bound.

wilkinsona:
> Would it be allowed to register a new bean within `postProcessAfterInitialization()` on the held `ApplicationContext`.

Not really, no. A `BeanPostProcessor` is intended for post-processing individual beans. A `BeanFactoryPostProcessor` would be a more appropriate place to perform bean registration. However, I'd prefer that this is done without any post-processing. It should be possible to do it in a manner similar to Boot's [`TomcatMetricsAutoConfiguration`](https://github.com/spring-projects/spring-boot/blob/a20046f8a373a6a3afe081e2e64c52cf165cb6b5/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/tomcat/TomcatMetricsAutoConfiguration.java).

matsumana:
How does this look like?
Although I've only tried Spring MVC yet, it seems to work fine

```java
@Configuration
@ConditionalOnClass({ Server.class, Loader.class, WebAppContext.class })
public class JettyMetricsAutoConfiguration {

	@Bean
	public WebServerFactoryCustomizer<JettyServletWebServerFactory> jettyFactoryCustomizer(
			MeterRegistry registry) {
		return (factory) -> factory.setThreadPool(
				new InstrumentedQueuedThreadPool(registry, Collections.emptyList()));
	}
}
```

wilkinsona:
That's better in terms of the conditions, but it's still setting the factory's thread pool and using `InstrumentedQueueThreadPool`. I think both of those should be avoided for the reasons I [tried to explain above](
https://github.com/micrometer-metrics/micrometer/pull/593#issuecomment-424105592).

wilkinsona:
I've opened https://github.com/micrometer-metrics/micrometer/issues/911 to describe and further discuss what I think needs to be done.

Issue #602, opened 2018-05-15
Micrometer stops updating metrics when maximumAllowableTags is reached


vgaddavcg:
We are running into issue where micrometer stops updating all the metrics when the resttemplate maxUriTags(**management.metrics.web.client.maxUriTags**) is reached. 

After some debugging, figured out **onMaxReached** meterFilter is triggered for every metric when observedTagValues are greater than maxTagValues. 

<img width="726" alt="screen shot 2018-05-15 at 2 43 17 pm" src="https://user-images.githubusercontent.com/26545621/40076989-5abb5aae-584e-11e8-9513-96adcd27a38c.png">

I think the check should be inside the if condition where we check the meterName.

<img width="723" alt="screen shot 2018-05-15 at 2 53 15 pm" src="https://user-images.githubusercontent.com/26545621/40077469-bc4555e4-584f-11e8-9f50-42601d5f2ea6.png">

jkschneider:
Thanks for the report and sorry for the trouble!

jkschneider:
Sure!

PR #603, opened 2018-05-15
stops updating metrics when maximumAllowableTags is reached

vgaddavcg:
#602 stops updating metrics when maximumAllowableTags is reached


jkschneider:
Thanks for the fix @vgaddavcg!

PR #607, opened 2018-05-17
Fixed javadoc: `Duration` instead of `Long`

benhubert:
Javadoc told that a value can be given as `Long` or `Long` (twice), but
actually meant `Long` or `Duration`. Fixed that.


jkschneider:
Thanks!

PR #608, opened 2018-05-18
Add URIUtils.toURL()

izeye:
This PR adds `URIUtils.toURL()` to avoid a check exception which is a bit verbose to handle.


jkschneider:
Death to checked exceptions! 🎉

PR #609, opened 2018-05-18
Shutdown thread pool while close StepMeterRegistry

quaff:
Java process cannot shutdown gracefully if any thread pool doesn't terminated.


jkschneider:
Would you mind rebasing this against 1.0.x please?

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -52,19 +52,20 @@ public void start() {
...
 
     public void start(ThreadFactory threadFactory) {
-        if (publisher != null)
+        if (scheduledExecutorService != null)
             stop();
 
         if (config.enabled()) {
-            publisher = Executors.newSingleThreadScheduledExecutor(threadFactory)
-                    .scheduleAtFixedRate(this::publish, config.step().toMillis(), config.step().toMillis(), TimeUnit.MILLISECONDS);
+            scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
```
checketts: Can you also update the ThreadFactory to name the thread so it is obvious it is a micrometer thread? I had started on this very same PR myself and having the named thread would have been useful then.

quaff: I think you should create another PR after this one approved.

PR #610, opened 2018-05-18
Remove org.apache.commons.text.StringEscapeUtils references

izeye:
`org.apache.commons.text.StringEscapeUtils` looks accidentally used as there is `io.micrometer.core.instrument.util.StringEscapeUtils`. So this PR replaces it with the intended one.

This PR also does:
- add a Checkstyle rule to prevent classes in `org.apache.commons.text` from sneaking into codebase again.
- update the related tests as `org.apache.commons.text.StringEscapeUtils` escapes "/" but AFAICT it doesn't look right because "/" doesn't need to be escaped for JSON.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -7,7 +7,7 @@
 
         <!-- Imports -->
         <module name="IllegalImportCheck" >
-            <property name="illegalPkgs" value="com.google.common.(?!cache).*"/>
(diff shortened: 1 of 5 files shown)


jkschneider:
You're right. I also concluded that "/" doesn't need to be escaped, and didn't include this in the semi-copied implementation.

izeye:
@jkschneider Thanks for the quick feedback :+1:

PR #615, opened 2018-05-21
Add IOUtils.toString()

izeye:
This PR adds `IOUtils.toString()`.

This also does:

- replace `IOUtils.toString()` from Apache Commons IO
- use the newly added `IOUtils.toString()`
- polish logging

When I was working on this, I noticed there is charset inconsistency (default vs. UTF-8) among meter registries. If it's not intentional, it would be good to make them consistent. I usually prefer to use UTF-8 but I'm not sure what is the best here. Maybe making it configurable might be necessary.


izeye:
I removed the "polish logging" part as it's not directly related to this change and makes things look more convoluted than necessary. I'll create another PR for it later.

PR #622, opened 2018-05-24
Polish logging in meter registries

izeye:
This PR polishes logging in meter registries, mostly by adding log level guards to where any computation is necessary.


jkschneider:
Wow, thanks for taking care of that.

PR #633, opened 2018-05-31
Fix Jetty metrics instrumentation

mweirauch:
The factory is being asked to create a server before our configuration
has a chance to kick in.

Relates to #593.


jkschneider:
Thanks @mweirauch!

Issue #632, opened 2018-05-31
GraphiteMeterRegistry does not pass Clock to self created GraphiteReporter


markusheiden:
The clock provided to the GraphiteMeterRegistry constructors is not passed to GraphiteMeterRegistry#defaultGraphiteReporter. So the reporter always uses Clock.defaultClock().

PR #641, opened 2018-06-05
Pass clock to default graphite reporter

markusheiden:
This is the PR for issue #632.


jkschneider:
Thanks @markusheiden!

PR #649, opened 2018-06-09
Fix etsy NamingConvention

cristianoperez:
Change Etsy NamingConvention to Dot instead of camelCase.
Didnt find any source saying that Etsy should be camel case


jkschneider:
We selected camel case as the default naming convention for Etsy StatsD because tags get appended with dots after the metric name. We are purposefully trying to avoid accidental wildcard matches in things like Graphite. A contrived example:

Suppose we had three metrics:
1. Named `jvm.memory.used` with a tag `id=oldGen`.
2. Named `jvm.memory` with a tag `used.process`. 
3. Named `jvm.memory` with a tag `used.system`.

With the camel case naming convention this becomes:

```
jvmMemoryUsed.id.oldGen
jvmMemory.used.process
jvmMemory.used.system
```

With dot naming convention it becomes:

```
jvm.memory.used.id.oldGen
jvm.memory.used.process
jvm.memory.used.system
```

See how there is accidental overlap because of the naming convention?

You won't find any documentation that suggests one convention or another, because the spec doesn't have an opinion. Feel free to override the naming convention for your registry if the default doesn't suit you.

cristianoperez:
I understand and that was exactly what i thought, but it looks more like a workaround since etsy dont support tags

jkschneider:
You're right, it doesn't support tags explicitly. The default is designed for hierarchical systems which ingest metrics with Etsy StatsD like Graphite that have a wildcarding mechanism that folks use like tags, and the most common pattern is for tag-like information to be embedded into the name with dot separators.

PR #652, opened 2018-06-11
Make AbstractTimer.pauseDetector nullable

izeye:
This PR changes `AbstractTimer.pauseDetector` nullable as it looks making things a bit simpler.


jkschneider:
I'd have to check the logic again, but I'm pretty sure passing a null `PauseDetector` implementation to HdrHistogram results in it creating its default `PauseDetector` implementation which is _not_ a no-op.

izeye:
@jkschneider AFAICT when `pauseDetector` is `NoopPauseDetector`, the only effective reference is with `shutdown()` which I think can be removed. The other possible references are guarded with `if` condition of `pauseDetector instanceof SimplePauseDetector`, so they are unreachable. Am I missing something?

jkschneider:
Looks like you are right. `Noop` was important when we were using `LatencyUtils` histograms, but that has decomposed in the latest head.

PR #653, opened 2018-06-14
Upgrade to PCollections 3.0.2

izeye:
This PR upgrades to PCollections 3.0.2.


jkschneider:
Moved to the lock mechanism included in Gradle 4.8. The upgrade to pcollections 3.0.2 was effective with 6501d447.

PR #662, opened 2018-06-16
Polish JettyMetricsAutoConfiguration

izeye:
This PR polishes `JettyMetricsAutoConfiguration` as follows:

- Removes the Tomcat-specific condition as it doesn't seem to be right.
- Makes `@ConditionalOnClass` align with `EmbeddedServletContainerAutoConfiguration.EmbeddedJetty`.
- Tightens conditions with `@AutoConfigureAfter` and `@ConditionalOnBean` similar to other metrics auto-configurations.

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/web/jetty/JettyMetricsAutoConfiguration.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/web/jetty/JettyMetricsAutoConfiguration.java
@@ -16,8 +16,14 @@
 package io.micrometer.spring.autoconfigure.web.jetty;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.spring.autoconfigure.MetricsAutoConfiguration;
+import io.micrometer.spring.autoconfigure.export.simple.SimpleMetricsExportAutoConfiguration;
+
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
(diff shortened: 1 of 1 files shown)


micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/web/jetty/JettyMetricsAutoConfiguration.java
```diff
@@ -27,12 +33,17 @@
  * @author Michael Weirauch
  */
 @Configuration
-@ConditionalOnClass(name = "org.eclipse.jetty.server.Server")
-@ConditionalOnMissingClass("org.apache.catalina.startup.Tomcat")
```
jkschneider: @mweirauch Any insight into why this was here?

mweirauch: It used to be there before I last touched this. But you are right, @izeye has it - as usual - right ;)
Thanks for spotting!

mweirauch:
As for the PR,  works good.

But could somebody double check if a `MeterRegistryCustomizer` setting common tags still works?

When I run my testapp with tomcat, common tags are applied to my metrics. Running with jetty (having the `JettyMetrics(Auto)Configuration` kick in - no matter if on `master` or this PR) common tags are not applied here. Perhaps the `MeterRegistry` shouldn't be passed to `JettyMetrics(Auto)Configuration` but instead the `ApplicationContext` from which we then pull the `MeterRegistry` within the `JettyMetricsPostProcessor`. Assume injecting the `MeterRegistry` at the current "location" constructs it too early?

If you can reproduce then the my initial PR exposed that problem. Not this PR. If nobody beats me to it, I can have a look. (A bit loaded currently, though.)

mweirauch:
I've added PR #665 which is based on this PR which resolves the aforementioned issue of creating the `MeterRegistry` too early. So I'd say go ahead with this PR and afterwards pull in the other.

Issue #592, opened 2018-05-06
Migrate to CircleCI 2.0


shakuzen:
[CircleCI 1.0 is scheduled to be EOL](https://circleci.com/sunset1-0/) in August. If we are going to continue using CircleCI as our CI tool, we should migrate to 2.0. I don't have experience with CircleCI, but they have a [migration guide](https://circleci.com/docs/2.0/migrating-from-1-2/). If someone else has experience and expertise with CircleCI 2.0, feel free to take up this issue. If no one speaks up, I will give a try doing it eventually, since I wanted to add a JDK 10 and possibly 11 EA build to our setup, but I don't want to bother trying to do this in the 1.0 config if we are going to migrate soon.

izeye:
This looks resolved via #663 and [the latest builds with CircleCI 2.0 look okay](https://circleci.com/gh/micrometer-metrics/micrometer).

PR #663, opened 2018-06-16
CircleCI 2.0 migration

shakuzen:
Migrates the CircleCI 1.0 configuration to the CircleCI 2.0 format.

This is an initial attempt to have the same build we had previously but using the new configuration format and new infrastructure. Let me know if anything stands out as different.

See #592


.circleci/config.yml
```diff
@@ -0,0 +1,47 @@
...
+          name: collect test reports
+          command: |
+            mkdir -p ~/micrometer/junit/
+            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/micrometer/junit/ \;
+
+      - store_test_results:
+          path: ~/micrometer/junit/
+
+      - store_artifacts:
+          path: build/libs
```
shakuzen: This isn't working as before, but I'm not actually sure what "artifacts" were being uploaded and what it effectively does. I don't see any "Artifacts" tab on our previous builds, but maybe that is a difference between CircleCI 1.0 and 2.0. Documentation: https://circleci.com/docs/2.0/artifacts/

I'm wondering if we should just delete this step. As far as I know, we do not have any use case to download artifacts from a specific build from CircleCI or view them in the CircleCI UI.

shakuzen: If we wanted to store the JARs we would first need to collect them to a single folder like the JUnit reports.

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+      - run:
+          name: run tests
+          command: ./gradlew test
+
+      - run:
+          name: collect test reports
+          when: always
+          command: |
+            mkdir -p ~/micrometer/test-results/junit/
+            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/micrometer/test-results/junit/ \;
```
shakuzen: This is an existing issue so I don't think it needs to be solved in this pull request, but this will copy all of the test result XML files into a single folder. This would not be a problem if they were all uniquely named, but JUnit Jupiter test report files are all named `TEST-junit-jupiter.xml` so they overwrite each other in the destination folder, losing test results. I propose we change the logic to make a `test-results` directory somewhere and then make a directory under it for each sub project that has test results. All test results XML files for a sub project are put under its directory. Then CircleCI will list the results for each project, instead of just a generic `junit` like it does not. Also, none of the results would be lost. If my bash foo were up to par, I would have just written the code to do this...

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+      - store_test_results:
+          path: ~/micrometer/test-results/
+
+      - store_artifacts:
+          path: build/libs
+
+      - deploy:
+          name: Deployment
+          command: |
+            if [ "$CIRCLE_BRANCH" = "master" ] || [ "$CIRCLE_BRANCH" = "1.0.x" ]; then
```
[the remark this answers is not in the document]
shakuzen: Alternatively, we can move the deployment to a separate job and use workflows, which allows configuring branch filters.
https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs

shakuzen: This also doesn't prevent the publishing from happening on forks (pull requests) that use the branches `master` or `1.0.x`, which we should fix, but is also an existing issue.

jkschneider:
Let's see if this works ;)

PR #666, opened 2018-06-19
Make step size configurable for SignalFx

izeye:
This PR makes step size configurable for SignalFx as it looks missed.


jkschneider:
Rebased on 1.0.x and merged with e90c8da9.

PR #667, opened 2018-06-20
Fixed typo when pushing timeseries.

obarcelonap:
Finally!! I've found the issue that was preventing the time series to appear in the platform: it was a typo in the json payload (facepalm)


obarcelonap:
Sorry, I was so happy to find the issue that I didn't pay attention to the test, now should be fixed :)

jkschneider:
Awesome @obarcelonap! I get a lot of correspondence from folks wanting dynatrace support. Thanks for all your hard work on it.

Issue #658, opened 2018-06-15
Support for Jersey MappableException in 'DefaultJerseyTagsProvider'


jvalloire:
I am using Micrometer in my Jersey application and i have noticed that the `DefaultJerseyTagsProvider` can't handle `MappableException`.

Here is an extract from the `DefaultJerseyTagsProvider` code:
```
 private static int statusCode(RequestEvent event) {
        Throwable exception = event.getException();
        if (exception != null) {
            if (exception instanceof WebApplicationException) {
                return ((WebApplicationException) exception).getResponse().getStatus();
            }
            return Response.Status.INTERNAL_SERVER_ERROR.getStatusCode();
        }
        return event.getContainerResponse().getStatus();
    }
```

With that code, whenever we have an exception that is not a `WebApplicationException`, we will have a **500** status code...

When using `MappableException`, application exceptions are mapped to Jersey Response (with an appropriate response status).
For example, the `ValidationExceptionMapper` from Jersey should return a **400** (_Bad Request_) when request parameters validation fails.
As a consequence, when using **micrometer-jersey2**, i am expecting a `http.server.requests` metric with a _{status = 400}_ tag when my request is not valid but i only have _{status = 500}_ tags.

Should it be possible to add support for `MappableException` so as the status tag value would be the same than the response status code?

jvalloire:
Yes, i would be glad to help.
Do you have any idea of the way `MappableException`s can be handled in the `DefaultJerseyTagsProvider` ?

mweirauch:
As I said on Slack, I can work on this. We just need to coordinate in case you are having a go at it.

The challenge is extracting a useful statuscode in this case. As we are post (or at least optionally post) `EXCEPTION_MAPPING_FINISHED` we can go by checking if the exception was successfully mapped.

@jvalloire Do you have a sample Jersey resource exposing the described problem? (Alternatively I think what is required for replicating is throwing a custom exception for which an `ExceptionMapper` is registered. So we can get passed a `MappableException` and can go on extracting the exceptionmapper-mapped statuscode from the response in `DefaultJerseyTagsProvider`.)

jvalloire:
@mweirauch i'm working on it now.

I add a use-case in `TestResource` test class with an endpoint throwing a custom exception with a `ExceptionMapper` registered for it.

[...]

PR #668, opened 2018-06-20
Simplify computation of the status code

jvalloire:
Given that the status code is computed for `RequestEvent` of type `FINISHED`, we can simply use the `ContainerResponse` to figure out the status code.
That way we handle both `WebApplicationException` and `MappableException` with registered `ExceptionMapper` (see https://github.com/micrometer-metrics/micrometer/issues/658).

In cases in which the `ContainerResponse` is not there, we fallback to `Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()`.


mweirauch:
Thanks for the PR! LGTM and runs as expected.

(Apparently as a collaborator, I can't approve PRs ...)

PR #670, opened 2018-06-20
Make DatadogNamingConvention.tagKey() respect its delegate

izeye:
This PR changes `DatadogNamingConvention.tagKey()` to respect its
`delegate` when a tag key starts with a number.


jkschneider:
Merged with 7cbd2515

PR #673, opened 2018-06-22
Rename DoubleFormat.decimalOrWhole()

izeye:
This PR renames `DoubleFormat.decimalOrWhole()` as both are not mutually exclusive and "whole" is narrower than "decimal". It also aligns with the name (`WHOLE_OR_DECIMAL`) of the constant it uses.


izeye:
@shakuzen I thought the same thing but it seems it's only for internal uses, so I thought it's okay to do so practically.

@jkschneider WDYT?

izeye:
@jkschneider Thanks for the feedback!

If you agree with this renaming, I'll update this as @shakuzen suggested. Please let me know what you think.

izeye:
I changed as suggested and rebased onto `1.0.x`.

micrometer-core/src/main/java/io/micrometer/core/instrument/util/DoubleFormat.java
```diff
@@ -69,8 +69,19 @@ public static String decimalOrNan(double d) {
      * @param d Number to format.
      * @return A stringified version of the number that only uses a decimal representation if the number is not
      * whole.
+     * @deprecated since 1.0.6 in favour of {@link #wholeOrDecimal(double)}
```
shakuzen:
```suggestion
     * @deprecated since 1.0.11 in favour of {@link #wholeOrDecimal(double)}
```

izeye: @shakuzen I applied the suggested change in 81cab5dacf4e10cc6c89e9ad2bf4343caf02ae7d.

PR #677, opened 2018-06-25
StepMeter reports count and sum in intervals of step

daichi-m:


checketts:
Thanks for finding this. Are there any tests that could  help catch this?

Issue #676, opened 2018-06-25
StepTimer and StepDistributionSummary updates count and sum based on expiry


daichi-m:
The `count` and `sum` metrics for `StepMeterRegistry` should be updated every `step` interval. But for `StepTimer` and `StepDistributionSummary` this configuration is not respected, and it is updated every `expiry` interval.

Here is a small gist to demonstrate the above issue:
https://gist.github.com/daichi-m/b8d647520d668b0ec7f64fec339b2e1b

PR #678, opened 2018-06-25
StepMeter does not reports count and sum in intervals of step

daichi-m:
GH Issue: #676


jkschneider:
Thanks for pointing this out and the fix @daichi-m!

jkschneider:
Rebased onto 1.0.x for inclusion in 1.0.6.

PR #679, opened 2018-06-25
Add license headers

izeye:
This PR adds missing license headers.


shakuzen:
Does our license check not check the year (range)?

izeye:
@shakuzen I just followed the other license headers as it looks the codebase consistently uses the start year only although I don't know whether it's correct or not.

For the other projects see https://github.com/spring-projects/spring-kafka/pull/567#discussion_r168630303

jkschneider:
AFAIK, our license check is kind of unintelligent in that it checks for exact matching text. So if we changed the year to 2017-2018 or something similar, we'd have to update all the licenses.

PR #682, opened 2018-06-25
Add a constant for histogram time unit

izeye:
This PR adds a constant for histogram time unit for better readability.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
@@ -34,6 +34,9 @@ import java.util.function.Supplier;
 import static java.util.Objects.requireNonNull;
 
 public abstract class AbstractTimer extends AbstractMeter implements Timer {
+
+    private static final TimeUnit HISTOGRAM_TIME_UNIT = TimeUnit.NANOSECONDS;
+
     private static Map<PauseDetector, org.LatencyUtils.PauseDetector> pauseDetectorCache =
             new ConcurrentHashMap<>();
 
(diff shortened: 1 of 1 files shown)


jkschneider:
I'm a little uncomfortable with this, since the use on 138, 149, 160, and 171 has nothing to do with histograms.

izeye:
@jkschneider Sorry, I just refactored after reading this commit, 3ec8bf277794f97bc6b1e2150090fe7a6811949c. When I was reading it, `TimeUnit.NANOSECONDS` is consistently used across the class, so I thought it's sort of internal time unit and just prefixed with the name using it in the commit. Honestly I'm confused between `TimeUnit.NANOSECONDS` and `baseTimeUnit` as it's never used in the class although it's children uses it. Could you explain the difference between them?

jkschneider:
@izeye Each of the record methods in `AbstractTimer` are really just convenience methods for `record(long amount, TimeUnit unit)`. This in turn calls the abstract `recordNonNegative`. It is in the implementations of `recordNonNegative` that various implementations scale the measurement to the implementation-specific `baseTimeUnit`. 

The histogram data structure is shared across all the implementations, so we use nanos internally in it, and the statistics sampled from the histogram are scaled by each implementation on the way out.

izeye:
@jkschneider Thanks for clarification 👍 

If so, does it make sense renaming it to `INTERNAL_TIME_UNIT`?

jkschneider:
I don't want to be too prescriptive here, but IMO it already uses a constant, `TimeUnit.NANOSECONDS` ;)

izeye:
@jkschneider Okay, thanks for enlightening me on this :smile: Closing.

PR #688, opened 2018-06-27
[InfluxDB] Add a sslVerify boolean options

### LINKED ISSUE #687: [InfluxDB] Add a sslVerify boolean options
When sending metrics to influxdb over https it should be possible to omit verification.

By adding an `sslVerify` boolean to switch on/off certificate chain and hostname verification. This comes in handy for all those that need to send metrics over https to a known and trusted host, without having to setup keystores and truststores.

esarbanis:
Closes #687

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -29,9 +29,14 @@ import io.micrometer.core.lang.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.net.ssl.*;
 import java.io.OutputStream;
 import java.net.*;
 import java.nio.charset.StandardCharsets;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.security.cert.X509Certificate;
 import java.util.Base64;
 import java.util.List;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -74,6 +77,12 @@ private void createDatabaseIfNecessary() {
             con.setConnectTimeout((int) config.connectTimeout().toMillis());
             con.setReadTimeout((int) config.readTimeout().toMillis());
             con.setRequestMethod(HttpMethod.POST);
+
+            if (!config.sslVerify() && "https".equals(queryEndpoint.getProtocol())) {
+                HttpsURLConnection.class.cast(con).setSSLSocketFactory(createSslSocketFactory());
+                HttpsURLConnection.class.cast(con).setHostnameVerifier(new NoopHostnameVerifier());
+            }
```
stefanbirkner: It would be nice to extract a method

    private void doNotVerifySslCeritifcateIfDesired(HttpURLConnection con, URL url) {
      if (!config.sslVerify() && "https".equals(url.getProtocol())) {
        HttpsURLConnection httpsCon = HttpsURLConnection.class.cast(con);
        httpsCon.setSSLSocketFactory(createSslSocketFactory());
        httpsCon.setHostnameVerifier(new NoopHostnameVerifier());
      }
    }

because it makes it easier to understand the code and it avoids duplicate code.

esarbanis: Yes, it does make sense. Thanks!

jkschneider:
Rather than adding config for individual features like this, I think we want to go the direction of allowing you to inject a pre-configured HTTP client into the registry. Then you could take care of skipping SSL verification there. See #842.

PR #692, opened 2018-06-28
Merge replacements into one

izeye:
This PR also changes to use a pre-compiled pattern and polishes names.


jkschneider:
Note to self: when merging this, be very careful. We have added `:` escaping as well.

izeye:
I resolved conflicts, changed the target branch to 1.0.x, and also applied the @jkschneider 's "Note to self" comment above.

PR #696, opened 2018-07-01
Switch to CircleCI workflows

shakuzen:
Separates build step and deploy step, so that the deploy step is only conditionally run on integration branches and version tags.


shakuzen:
Build failure looks related to the new Gradle locks, which seems still broken on the `1.0.x` branch.

jkschneider:
`1.0.x` is finally green again, so here we go...

PR #697, opened 2018-07-01
Bump shadow plugin to 2.0.4 and Mockito 2.18.+ for JDK10 compatibility

shakuzen:
Updates these dependencies and updates locks.

A build passed with JDK10 locally. After merging this, I will work on adding a workflow that tests `master` against JDK10.


jkschneider:
To be clear: prior to this change, the resultant artifact _works_ on JDK 10, just doesn't build with JDK 10 right?

shakuzen:
@jkschneider Yes, that's correct as far as I can tell. I haven't tried every integration with Java 10, but at least the ones I have are already fine at runtime with Java 10. This change makes it so we can run a build with JDK10 to ensure as part of our build everything works with Java 10.

shakuzen:
I rebased on the current master and cleaned up the commit to remove the unrelated (and unnecessary) lock updates.

Issue #243, opened 2017-11-22
Prometheus Push Gateway - error during metrics collection


lupo112:
Hi

I am using prometheus gateway and I have a problem. Gateway reporting this error:

```
An error has occurred during metrics collection:

gathered metric family process_start_time_seconds has help "The start time of the Java virtual machine" but should have "Start time of the process since unix epoch in seconds."
```

So it cannot be scraped any data from push gateway into prometheus server.

Any idea how to solve this?

Used versions (all latest):
```
Micrometer (core, spring-legacy, registry-prometheus) 1.0.0-rc.4
Prometheus Java client (simpleclient, pushgateway) 0.1.0
Prometheus Push Gateway 0.4.0
```

mweirauch:
Ugh... So the Pushgateway is spitting that out you say? Or is it Prometheus actually when it is scraping the Pushgateway?
* Please provide the output from the prometheus endpoint of your app (`/prometheus` by default) including the `TYPE` and `HELP` comments for `process_start_time_seconds`.
* Also the scrape config for the Pushgateway within Prometheus would be interessing. (topic `honor_labels` ...).
* And if available any Pushgateway configuration.
* If possible the output of the `/metrics` endpoint of the Pushgateway itself.

Would like to reproduce what is happening. Perhaps there's a mismatch between the `process_start_time_seconds` and it's help of the Pushgateway itself and that of the pushed metrics to it which are all exposed under the same `/metrics` endpoint. (Sounds likely)

lupo112:
Problem is on the `/metrics` endpoint of the Pushgateway which returns:

```
An error has occurred during metrics collection:

gathered metric family process_start_time_seconds has help "The start time of the Java virtual machine" but should have "Start time of the process since unix epoch in seconds."
```
App produces metrics correctly:
```
# HELP jvm_buffer_total_capacity_bytes An estimate of the total capacity of the buffers in this pool
# TYPE jvm_buffer_total_capacity_bytes gauge
jvm_buffer_total_capacity_bytes{id="direct",} 1.34808549E8
jvm_buffer_total_capacity_bytes{id="mapped",} 0.0
# HELP jvm_memory_used_bytes The amount of used memory
# TYPE jvm_memory_used_bytes gauge
jvm_memory_used_bytes{area="nonheap",id="Code Cache",} 2.6257728E7
jvm_memory_used_bytes{area="nonheap",id="Metaspace",} 8.7563768E7
[...]

PR #717, opened 2018-07-10
Fixed help message to conform to prometheus-pushgateway

tsl-karlp:
Because of https://github.com/micrometer-metrics/micrometer/issues/243, you'll get an error from prometheus-pushgateway:

```
An error has occurred during metrics collection:

gathered metric family process_start_time_seconds has help "The start time of the Java virtual machine" but should have "Start time of the process since unix epoch in seconds."
```

See https://github.com/micrometer-metrics/micrometer/issues/243


jkschneider:
Merged with 799961dc

PR #725, opened 2018-07-15
Make DataSourcePoolMetricsAutoConfiguration conditional on DataSource bean

izeye:
This PR makes `DataSourcePoolMetricsAutoConfiguration` conditional on `DataSource` bean to prevent it from being processed when `DataSource` bean doesn't exist.

This PR also tries to align other conditions and structure with Spring Boot 2.x to easily notice when they are diverged.

As a side note, I noticed the auto-configuration for Hikari is not synched from Spring Boot 2.x.


izeye:
I'm preparing another PR for Hikari one but struggling with dependency locking as I couldn't find an easy way to switch Hikari dependency from `testCompile` to `optional` `compile` without touching other locked dependencies. It would be appreciated if anyone gives me some hint on it.

jkschneider:
I don't have a great suggestion on the locks other than manual editing. Gradle 4.x's locking is kind of disappointing in that way.

izeye:
@jkschneider Thanks for the feedbacks! I just noticed the difference between Spring Boot 1.5.x support and 2.x support and didn't look into it much due to the dependency locking issue. I guess you're right.

PR #727, opened 2018-07-16
Polish UptimeMetrics.bindTo()

izeye:
This PR polishes `UptimeMetrics.bindTo()`.

As a side effect, `process.start.time` has been rounded down but IMHO it looks better as its time unit is seconds.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/UptimeMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/UptimeMetrics.java
@@ -56,12 +56,12 @@ public class UptimeMetrics implements MeterBinder {
 
     @Override
     public void bindTo(MeterRegistry registry) {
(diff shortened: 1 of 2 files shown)


jkschneider:
Further simplified a bit to define the input to `process.start.time` as dimensioned in milliseconds. This way we can use the method reference and not do any truncation. Remember, the value gets scaled to whatever the monitoring system base unit of time is regardless of what the input dimension is.

jkschneider:
Merged with d482a283

izeye:
@jkschneider It looks its time unit doesn't match with its description changed in 799961dc30e250bf002580a061398f6188ca1538. Based on its commit message, we can't change its description due to Prometheus-specific reason?

jkschneider:
Ah, didn't even see that. Removed the "in seconds" part in 5632cd4b.

jkschneider:
I'm at the point now where I'd suggest just renaming process start time when shipping through pushgateway. We could put a meter filter in the pushgateway autoconfiguration itself.

PR #737, opened 2018-07-18
Checks config.enabled() do publish metrics on close

marceloavan:
If config.enabled() is false, the scheduled executor is not activated, so it must not publish on close too.


jkschneider:
Good catch @marceloavan!

jkschneider:
Merged with e3556440.

PR #747, opened 2018-07-20
Make GraphiteNamingConvention respect its delegate NamingConvention

izeye:
This PR makes `GraphiteNamingConvention` respect its `delegate` `NamingConvention` as it delegates all to `tagKey()`.


jkschneider:
Wow! We sure had a lot of bugs in these naming conventions... Kinda embarassing 🤢.

PR #749, opened 2018-07-20
Use pre-compiled patterns in WavefrontNamingConvention

izeye:
This PR changes to use pre-compiled patterns in `WavefrontNamingConvention`.


shakuzen:
Looks like `1.0.x` wasn't merged to master after this, and there was a conflict as the same change with different constant names was already in master. I've opted to keep master with the constant names it had (different from `1.0.x`). Hopefully that's alright.

PR #751, opened 2018-07-21
export percentile values to new relic

jfarr:


jfarr:
You also have to tell the underlying meter to publish percentiles like so:

```
Timer.builder("event.response.time")
        .publishPercentiles(.95, .99)
        .register(meterRegistry);
```
Then this will export them to New Relic.

jkschneider:
Percentiles are already exported to new relic. For the timer `event.response.time`, if a 95th percentile is enabled there will be a metric with name `event.response.time.percentile` and tag `phi=0.95`.

izeye:
@jkschneider Based on your comment and "invalid" label, this meant to be closed?

jkschneider:
I think it's destined for closure, but thought I'd wait a little bit for a response in case there's something we are missing.

jfarr:
I must've missed that, thanks.

izeye:
@jfarr Thanks for the confirmation! Closing.

PR #752, opened 2018-07-23
corrected some copy paste classes in samples

kmarquardsen:


jkschneider:
Thanks @kmarquardsen!

PR #754, opened 2018-07-24
Fix ConditionalOnBean for DataSourcePoolMetricsAutoConfiguration

izeye:
This PR fixes `@ConditionalOnBean` for `DataSourcePoolMetricsAutoConfiguration`.

Closes gh-753

PR #758, opened 2018-07-25
Adjust process.start.time description when using Pushgateway

mweirauch:
Since v0.5.0+ differing descriptions for meters are not allowed anymore.
(If this was intentional or coincident by updating the go client is beyond me.)

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
@@ -17,6 +17,8 @@ package io.micrometer.spring.autoconfigure.export.prometheus;
 
 import io.micrometer.core.annotation.Incubating;
 import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.Meter.Id;
+import io.micrometer.core.instrument.config.MeterFilter;
 import io.micrometer.prometheus.PrometheusConfig;
 import io.micrometer.prometheus.PrometheusMeterRegistry;
 import io.micrometer.spring.autoconfigure.CompositeMeterRegistryAutoConfiguration;
@@ -182,6 +184,28 @@ public class PrometheusMetricsExportAutoConfiguration {
                 job = "spring";
             }
             return job;
+        }
+
+        @Bean
+        MeterFilter pushgatewayCompatMeterFilter() {
+            return new MeterFilter() {
+
+                @Override
+                public Id map(Id id) {
+                    if (!"process.start.time".equals(id.getName())) {
+                        return id;
+                    }
+                    /*
+                     * Starting with prom/pushgateway:v0.5.0 differing meter
+                     * descriptions are not accepted anymore. Pushgateway is
+                     * exposing its own "process_start_time_seconds" metric which
+                     * has a differing description than the one from Micrometer.
+                     */
+                    return new Id(id.getName(), id.getTags(), id.getBaseUnit(),
+                            "Start time of the process since unix epoch in seconds.", id.getType());
+                }
+
+            };
         }
     }
 }


izeye:
Either this PR or renaming the meter as @jkschneider mentioned in https://github.com/micrometer-metrics/micrometer/pull/727#issuecomment-406280886 also needs to be ported to Spring Boot 2.x support as it will break Spring Boot 2.x applications using Prometheus Pushgateway v0.5.0+. Both Spring Boot [2.0.4](https://github.com/spring-projects/spring-boot/milestone/113) and [2.1.0.M1 ](https://github.com/spring-projects/spring-boot/milestone/99)planed to be released in July 30.

izeye:
> Regarding porting to Spring Boot 2.x: There's no Pushgateway support in 2.x. Although I wouldn't mind having it in 2.x.

@mweirauch Ah, sorry. I should have checked it. I just assumed that Spring Boot 1.5.x support in Micrometer 1.0.x aligns with Spring Boot 2.0.x one.

> The question is, are we gonna push out a 1.0.7 before 2018-07-30 and integrate this fix and the DataSource one in 2.x? //cc @jkschneider

I'm not sure I understand this but if "this fix" means this PR and "the DataSource one" refers https://github.com/micrometer-metrics/micrometer/pull/754, they are irrelevant to Spring Boot 2.x as in case of the former you mentioned right before and in case of the latter it's just for Spring Boot 1.5 support. Am I missing something?

mweirauch:
> @mweirauch Ah, sorry. I should have checked it. I just assumed that Spring Boot 1.5.x support in Micrometer 1.0.x aligns with Spring Boot 2.0.x one.

Yeah, I wouldn't mind having it aligned. Would even argue for a PR against `2.0.x` as this feature has been present in Micrometer all the time. Just the question if it'd be accepted for `2.0.x`.

> I'm not sure I understand this but if "this fix" means this PR and "the DataSource one" refers #754, they are irrelevant to Spring Boot 2.x as in case of the former you mentioned right before and in case of the latter it's just for Spring Boot 1.5 support. Am I missing something?

Oh, you are right. These changes are not Spring Boot 2.x relevant.

jkschneider:
> are we gonna push out a 1.0.7 before 2018-07-30

Sorry, wasn't aware that Spring Boot 2.0.4 had slipped that late. I'm sure they'll be a 2.0.5 before it's all said and done.

I wouldn't propose Pushgateway support in Spring Boot 2.0.x. I suppose we could justify it in 2.1.x under the premise that it really is the correct model for Spring Cloud Task, etc. which are definitionally short-lived (this despite the fact that I know folks will use it for long-lived applications as well).

jkschneider:
While the refactoring change in Micrometer 1.0.6 unintentionally broke Pushgateway (again), I agree that requiring description text that is inaccurate for most monitoring systems just to satisfy Pushgateway's description text validation is extreme.

@brianbrazil Does a loosening of Pushgateway's description text validation stand a chance of acceptance?

brian-brazil:
That would be a question for @beorn7, as he maintains the Pushgateway.

Note that the pushgateway is only intended for use by batch jobs, not as a way to switch Prometheus to push.

beorn7:
The restriction isn't really coming from the Pushgateway but from the Prometheus exposition format. Each metric name can only appear once, and only have one `HELP` string. If the Pushgateway is asked to expose two different `HELP` strings for the same metric name, it has to do something about it.

IIRC it will simply log the fact that there are inconsistencies and then randomly pick one to "win". What's the exact kind of error you are encountering here?

mweirauch:
Thanks for the reply @beorn7 !
Starting with `v0.5.0` you get a status 500 on visiting the metrics endpoint of Pushgateway with the following error message: (No warnings or errors are logged to console.)

```
An error has occurred during metrics collection:

gathered metric family process_start_time_seconds has help "Start time of the process since unix epoch." but should have "Start time of the process since unix epoch in seconds."
```
So the Pushgateway metric and any other ingested metric carrying the same name but a different description are escalated. This wasn't the case with any version prior `v0.5.0`. `v0.4.0` is working fine for me.

Perhaps it's related to the prometheus/client_golang updates exposing a stricter validation now? (Just guessing.)

beorn7:
I think I found the reason. See https://github.com/prometheus/pushgateway/issues/194 for details.

jkschneider:
Thanks @beorn7. I'm going to close this for now, with the understanding that a future release of Pushgateway will fix this.

dennysfredericci:
Now pushgateway accepts a different help description, it returns 200 and metrics are published but we have a warning message on pushgateway logs.

`level=info ts=2021-01-25T16:05:04.085Z caller=diskmetricstore.go:173 msg="metric families overlap" err="Metric family has the same name as a metric family used by the Pushgateway itself but it has a different help string. Changing it to the standard help string. This is bad. Fix your pushed metrics!" metric_family="name:\"process_start_time_seconds\" help:\"Start time of the process since unix epoch.\" type:GAUGE metric:<label:<name:\"instance\" value:\"\" > label:<name:\"job\" value:\"lmr-batch-reporting\" > gauge:<value:1.611590696811e+09 > > " standard_help="Start time of the process since unix epoch in seconds."`

1) Should we fix the message in the micrometer to stop these logs?  I have tested it with pushgateway 1.4.0 and micrometer 1.3.2.

2) Is there an easy way to change the help text without a new version of the micrometer?

shakuzen:
@dennysfredericci this was most recently discussed in #2413. There's not much we can do but what I mentioned in https://github.com/micrometer-metrics/micrometer/issues/2413#issuecomment-766516957. If you think that is still worth pursuing, please open a new issue. We can't "fix" the description in general because it is not necessarily seconds for other MeterRegistry implementations.

Issue #646, opened 2018-06-07
Support management.metrics.tags.*


jkschneider:
This support was added in Spring Boot 2.0.2 but wasn't backported to Micrometer 1.0.4?

izeye:
See https://github.com/spring-projects/spring-boot/issues/12933

/cc @snicoll

shakuzen:
@jkschneider just wanted to note that was added to Spring Boot 2.1 (milestone 2.1.0.M1)

manivkm:
Is this feature added in Spring Boot 2.1?

jkschneider:
@manivkm Yes, which is in snapshots until the first milestone on July 19.

manivkm:
Is it possible, somehow to consume the fixed version? For example, 2.0.3.SNAPSHOT or 2.0.3.RELEASE

snicoll:
I am afraid not, that question was already asked on the linked issue, see there for more details.

jkschneider:
@ghost You should be able to consume the fixed version with 2.1.0.SNAPSHOT, unless I misunderstand the question.

amitarora15:
What is way to define these common tag in YML file?

shakuzen:
@amitarora15 the feature was added in Spring Boot 2.1, so you will first need to upgrade to that to use it. As for the format, the second looks right to me. See also the Spring Boot documentation: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-common-tags

amitarora15:
Thanks @shakuzen. Spring Boot 2.0 supports it through MeterRegistryCustomizer. And as you pointed out Boot 2.1 supports it through application configuration by yml.

shakuzen:
I'm not sure what yaml changes you mean - the documentation for 2.0 won't mention changes made in a future release. I think it is going to be easier to get to the bottom of this via chatting on the Micrometer Slack or the Spring Boot Gitter channel.

amitarora15:
@shakuzen No problem. I understand what you were saying is that supporting of common tags through yaml or properties is present in 2.1 version and mentioned in the documentation. 

Spring Boot 2.0 supports it through MeterRegisteryCustomizer

PR #764, opened 2018-07-27
Support management.metrics.tags.*

izeye:
This PR adds support for `management.metrics.tags.*` by almost just synching from https://github.com/spring-projects/spring-boot/commit/602f52fffc63d71e4ab16e3da11c17e24aa4df69.

Closes gh-646

/cc @snicoll

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/PropertiesMeterFilter.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/PropertiesMeterFilter.java
@@ -16,6 +16,7 @@
 package io.micrometer.spring;
 
 import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.config.MeterFilter;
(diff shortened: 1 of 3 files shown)


snicoll:
For the record, I am not super keen to have `2.1` changes backported to a virtual `1.5` support. This makes a strange upgrade path for users using the legacy module as upgrading to a `2.0.x` maintenance means they'll lose the feature.

If we consider this module to be a mirror of the support for Spring Boot `1.5.x`, I  don't think it's legit to have the feature here if it's not available in `2.0.x`.

izeye:
@snicoll Thanks for the feedback! I got your point.

I just assumed as follows:

- Micrometer 1.0.x Spring Boot 1.5.x support aligns with Spring Boot 2.0.x support.
- Micrometer 1.1.x Spring Boot 1.5.x support aligns with Spring Boot 2.1.x support.

But my assumption might be wrong as it's just my observed conclusion. We'd better listen to @jkschneider 's opinion on this.

jkschneider:
I agree with your characterization @izeye that Micrometer 1.0.x = Boot 2.0.x, etc.

PR #766, opened 2018-07-29
Use updated namingConvention in EtsyStatsdLineBuilder

izeye:
This PR changes to use the updated `namingConvention` in `EtsyStatsdLineBuilder` as it looks more consistent.


jkschneider:
I don't think there's a issue here. `config.namingConvention()` is always equal to `this.namingConvention` at the point where this code is called. We only keep track of `this.namingConvention` to know when to update the cache of names, which in turn only exists to create less garbage per StatsD line.

Let me know if I'm missing something.

izeye:
@jkschneider Thanks for the feedback!

> as it looks more consistent.

As I mentioned the above, I didn't mean there's any problem and just thought using the updated one looks more consistent to me but if you disagree, I don't have a strong opinion on this.

Issue #763, opened 2018-07-27
Enhance ExecutorServiceMetrics to record time between task submission and task execution start time


maddenj-ie:
I think it would be useful to expose the duration for which a task was waiting before being serviced.  This allows the ratio between queue time and work time to be quickly determined.

mveitas:
The Metrics project has a good representation of this to capture the number of submitted, running, completed, and the duration of execution: https://github.com/dropwizard/metrics/blob/4.1-development/metrics-core/src/main/java/com/codahale/metrics/InstrumentedExecutorService.java

slovdahl:
I noticed that the idle time between submission and execution start isn't collected if `ExecutorServiceMetrics.monitor` is invoked with an `ExecutorService`. Am I missing something obvious, or was this just an oversight?

PR #775, opened 2018-08-04
Add idle timer to Executor & ExecutorService task metrics

maddenj-ie:
This is by no means the finished article.  This first update is just to get the ball rolling and get some feedback on the approach.  Suggestions welcome.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedExecutor.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedExecutor.java
@@ -26,17 +26,45 @@ import java.util.concurrent.Executor;
  * An {@link Executor} that is timed
  */
 public class TimedExecutor implements Executor {
+    private final MeterRegistry registry;
     private final Executor delegate;
-    private final Timer timer;
+    private final Timer executionTimer;
+    private final Timer idleTimer;
 
     public TimedExecutor(MeterRegistry registry, Executor delegate, String executorName, Iterable<Tag> tags) {
+        this.registry = registry;
         this.delegate = delegate;
-        this.timer = registry.timer("executor", Tags.concat(tags, "name", executorName));
+        this.executionTimer = registry.timer("executor.execution", Tags.concat(tags, "name", executorName));
+        this.idleTimer = registry.timer("executor.idle", Tags.concat(tags, "name", executorName));
     }
 
     @Override
     public void execute(Runnable command) {
-        delegate.execute(timer.wrap(command));
+        delegate.execute(new TimedRunnable(command));
     }
+    
+    class TimedRunnable implements Runnable {
+
+        private final Runnable command;
+        private final Timer.Sample idleSample;
+
+        public TimedRunnable(Runnable command) {
+            this.command = command;
+            this.idleSample = Timer.start(registry);
+        }
+
+        @Override
+        public void run() {
+            idleSample.stop(idleTimer);
+            Timer.Sample executionSample = Timer.start(registry);
+            try {
+                command.run();
+            } finally {
+                executionSample.stop(executionTimer);
+            }
+        }
+
+    }
+
 
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedExecutorService.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedExecutorService.java
@@ -15,16 +15,21 @@
  */
 package io.micrometer.core.instrument.internal;
 
+import static java.util.stream.Collectors.toList;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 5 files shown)


maddenj-ie:
There are some potential drawbacks to this approach.  The calling thread takes the hit on the meter creation/lookup.  There is also the creation of another object that must be passed over a thread boundary.  
I also see that the reactor project may depend on this in the future.  Is it worth separating it from the existing functionality so that the trade offs may be evaluated by the user and they can choose which approach to take?

jkschneider:
@smaldini Any thoughts?

maddenj-ie:
Hi @jkschneider, any progress on this?

jkschneider:
ping @smaldini

micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedExecutor.java
```diff
@@ -15,28 +15,32 @@
...
-    private final Timer timer;
+    private final Timer executionTimer;
+    private final Timer idleTimer;
 
     public TimedExecutor(MeterRegistry registry, Executor delegate, String executorName, Iterable<Tag> tags) {
+        this.registry = registry;
         this.delegate = delegate;
-        this.timer = registry.timer("executor", Tags.concat(tags, "name", executorName));
+        this.executionTimer = registry.timer("executor.execution", Tags.concat(tags, "name", executorName));
+        this.idleTimer = registry.timer("executor.idle", Tags.concat(tags, "name", executorName));
```
perlun: This is the part that addresses #1080.

micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedExecutorService.java
```diff
@@ -15,29 +15,39 @@
...
+    private final Timer idleTimer;
 
-    public TimedExecutorService(MeterRegistry registry, ExecutorService delegate, String executorServiceName, Iterable<Tag> tags) {
+    public TimedExecutorService(MeterRegistry registry, ExecutorService delegate, String executorServiceName,
+            Iterable<Tag> tags) {
+        this.registry = registry;
         this.delegate = delegate;
-        this.timer = registry.timer("executor", Tags.concat(tags ,"name", executorServiceName));
+        this.executionTimer = registry.timer("executor.execution", Tags.concat(tags, "name", executorServiceName));
+        this.idleTimer = registry.timer("executor.idle", Tags.concat(tags, "name", executorServiceName));
```
perlun: ...and this. There are also a few other instances further down.

micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedExecutor.java
```diff
@@ -15,28 +15,32 @@
...
     private final Executor delegate;
-    private final Timer timer;
+    private final Timer executionTimer;
+    private final Timer idleTimer;
 
     public TimedExecutor(MeterRegistry registry, Executor delegate, String executorName, Iterable<Tag> tags) {
+        this.registry = registry;
         this.delegate = delegate;
-        this.timer = registry.timer("executor", Tags.concat(tags, "name", executorName));
+        this.executionTimer = registry.timer("executor.execution", Tags.concat(tags, "name", executorName));
```
shakuzen: Unfortunately we can't just change the name of the metric as this will break users relying on the name being `executor` as it has been. If we want to correct the inconsistency, we'll need to add another timer that times the same thing with the new name. Then in Micrometer 2.0, we can remove the `executor` one.

shakuzen:
@maddenj-ie thank you for the pull request. Sorry it took so long to take action on this and ultimately #1608 superseded this. As commented, we can't change the name of a released metric because it will break users.

maddenj-ie:
Would it not have been more appropriate to still use the pull request and simply ask me to update the metric names?

shakuzen:
> Would it not have been more appropriate to still use the pull request and simply ask me to update the metric names?

Yes, that is the flow we would usually take, and I apologize for not taking that route this time. #1608 added the `TimedScheduledExecutorService` which we also needed and I wanted to get things merged in time for the release I'm trying to make today. If I had more time (which isn't your fault) I would have asked for the changes here, merged this, and merged them into #1608, but that would have resulted in additional merge conflicts.
I do appreciate you taking your time to contribute and I'll try harder to make sure those contributions get integrated in the future.

maddenj-ie:
Hi @shakuzen.  Thank you for taking the time to respond.  Unfortunately this is not the first time I have been frustrated during my attempts to contribute to the micrometer project.  #763 was also resolved in a disappointing manner for me.
I've benefited from open source projects and the patience and kindness of maintainers over the years and so this was my first opportunity to give back.
I appreciate that you're very busy so I won't labour the point any further.  Thanks for all the great work.

shakuzen:
>  Unfortunately this is not the first time I have been frustrated during my attempts to contribute to the micrometer project. #763 was also resolved in a disappointing manner for me.
> I've benefited from open source projects and the patience and kindness of maintainers over the years and so this was my first opportunity to give back.

Sorry about both times. Neither were your fault, and I can understand the frustration you're feeling. I do hope you'll continue to contribute and I'll do what I can to make the next time go smoothly. Thank you for all you've done and your kind words despite the disappointing issues.

PR #779, opened 2018-08-08
Polish PrometheusMetricsExportAutoConfiguration

izeye:
This commit polishes `PrometheusMetricsExportAutoConfiguration` by:

- Replace `PrometheusPushGatewayEnabledCondition` with `@ConditionalOnProperty` on `management.metrics.export.prometheus.pushgateway.enabled` as the other property in the composite condition should be checked in its outer auto-configuration class.
- Make `PrometheusPushGatewayConfiguration` `static`.
- Extract non-configuration logic in `PrometheusPushGatewayConfiguration` into `PushGatewayHandler`.
- Add tests for it.


izeye:
For easy review, see the space-ignored diff: https://github.com/micrometer-metrics/micrometer/pull/779/files?w=1

mweirauch:
LGTM!

Don't ask me why I came up with `AllNestedConditions` initially. Thanks!

PR #782, opened 2018-08-09
Structural pattern matching for meters

pablomusumeci:
Hi 👋 , I've seen an anti-pattern across all the registries which is the use of the following construction `if (meter instanceof XXX)`.

That's why  I want to propose a small enhancement to the current codebase by using a visitor to execute custom logic depending on the object type. Moreover, we can execute the same code for classes that belong to the same hierarchy if needed.

This visitor could be used in all the registries but before implementing such change I wanted to check what you think about this idea 🤓

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/MeterVisitor.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/MeterVisitor.java
@@ -1,0 +1,67 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.FunctionCounter;
+import io.micrometer.core.instrument.FunctionTimer;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.TimeGauge;
+import io.micrometer.core.instrument.Timer;
+
+/**
+ * Base meter visitor. By default it does nothing if the method for the type that's being visited was not overrided.
+ */
+public interface MeterVisitor {
+
+    default void defaultVisit(Meter meter) {
+    }
+
+    default void visitTimer(Timer timer) {
+        defaultVisit(timer);
+    }
+
+    default void visitCounter(Counter counter) {
+        defaultVisit(counter);
+    }
+
+    default void visitGauge(Gauge gauge) {
+        defaultVisit(gauge);
+    }
+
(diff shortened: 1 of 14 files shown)


jkschneider:
Nice design. Adding this to the 1.1.0 milestone as it is additive.

micrometer-core/src/main/java/io/micrometer/core/VisitableMeter.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core;
+
+/**
+ * Accepts a {@link MeterVisitor} implementation.
+ */
+public interface VisitableMeter {
+    void accept(MeterVisitor visitor);
```
jkschneider: I'm OK putting this on `Meter` itself.

micrometer-core/src/main/java/io/micrometer/core/MeterVisitor.java
```diff
@@ -0,0 +1,67 @@
...
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.TimeGauge;
+import io.micrometer.core.instrument.Timer;
+
+/**
+ * Base visitor implementation. By default it does nothing if the method for the type that's being visited was not overrided.
+ */
+public abstract class MeterVisitor {
```
jkschneider: We should be able to do this with an interface + default methods I believe.

micrometer-core/src/main/java/io/micrometer/core/VisitableMeter.java
pablomusumeci: I rather leave this as a separate interface as it just shows the capability to accept a visitor (something which is not really a meter's concern). But if you prefer to avoid the proliferation of classes I can for sure add it the the `Meter`.

micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
```diff
@@ -66,10 +67,10 @@ static Builder builder(Timed timed, String defaultName) {
         }
 
         return new Builder(timed.value().isEmpty() ? defaultName : timed.value())
-                .tags(timed.extraTags())
-                .description(timed.description().isEmpty() ? null : timed.description())
-                .publishPercentileHistogram(timed.histogram())
-                .publishPercentiles(timed.percentiles().length > 0 ? timed.percentiles() : null);
+            .tags(timed.extraTags())
```
dhaval24: Seems like here the IDE played its magic and added automatic spaces. I think should be reverted. Similarly in other meters.

jkschneider:
Been thinking about this. I think I like the Church-encoding approach described [here](http://blog.higher-order.com/blog/2009/08/21/structural-pattern-matching-in-java/) even better, as it is more convenient to use when we want to pattern match over meter type in a functional style. 

It has the disadvantage of requiring `Meter` to know about its subtypes, which may not be such an issue.

It has the advantage of being much more succinct than the visitor, which is quite a bit more verbose than the instanceof checking we were doing.

jkschneider:
> It has the disadvantage of requiring Meter to know about its subtypes

I guess this ship has sailed. Meter already contains the `Meter.Type` enum which exhaustively lists types of metrics anyway, though in slightly coarser terms than the set of `Meter` subtypes.

jkschneider:
One more disadvantage: if we ever add more `Meter` types, it breaks the API contract on `match`.

jkschneider:
This is what it winds up looking like for the New Relic implementation. No way I can mess it up, because I have to provide something for each argument of the method:

image

dhaval24:
Hmm, I think in this case Match API would have to be an exception. However, for the custom cases it might be appropriate to use instanceof . I don't expect to see too many custom cases indeed. Correct me if I am wrong. If Micrometer thinks to add new meter, then yes API would break and it would most likely happen in a feature / major version release as the repo rules.

izeye:
> I hate the name `Matcher`, but it's super late here and I can't think of a better one.

@jkschneider Maybe `MeterWriteMatcher`?

PR #787, opened 2018-08-11
Polish GraphiteHierarchicalNameMapper.toHierarchicalName()

izeye:
This commit polishes `GraphiteHierarchicalNameMapper.toHierarchicalName()` by:

- Changing to use only one` StringBuilder`.
- Changing to use `StringBuilder.append()` for concatenation of intermediate `String`s.


izeye:
I added e8d5bda to rename local variables to make them clear.

jkschneider:
It's a common misconception I think that `StringBuilder.append` is more performant in this case. Examining the bytecode from the original, you'll see that the compiler actually generates `StringBuilder.append` calls for situations like this:

image

izeye:
@jkschneider IIUC the bytecode you provided above means that the part of the code will create an "intermediate" `String` via `StringBuilder`. I meant to eliminate the "intermediate" `String` via a single `StringBuilder`. Am I missing something?

jkschneider:
Ah, gotcha. It's late here ;)

PR #788, opened 2018-08-13
Adjust naming of Subscriber and Processor fields in StatsD classes

breun:
Going through the Micrometer StatsD code I got confused a bit. `Subscriber` and `Publisher` are opposite concepts, but a lot of subscribers are named `publisher`. A `Processor` is both a `Subscriber` and `Publisher`, so I've also changed the `Processor` names to indicate they are a `Processor` and not just a `Publisher`.


jkschneider:
Fair!

Issue #790, opened 2018-08-16
Add extra debugging to WavefrontMeterRegistry


scottmf:
hi,
Right now it is very difficult to trace metrics from my service to wavefront.  It would be really nice if I could turn on debug and trace in order to inspect what is being sent

My ask is for two items:
1. add logger.trace() [here](https://github.com/micrometer-metrics/micrometer/blob/fbca7acb93917767dfc151e25d3df8a5d411c72a/implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java#L79) to show the metrics that are being transmitted to wavefront - proxy or direct
2. add a logger.debug() to show the number of metrics being sent to the wavefront proxy [here](https://github.com/micrometer-metrics/micrometer/blob/fbca7acb93917767dfc151e25d3df8a5d411c72a/implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java#L121)

jkschneider:
@scottmf Out of curiosity, why do you need this? Are you not receiving metrics? Just don't trust it?

scottmf:
@jkschneider, The problem stems from the scenario where I was sending metrics to wavefront but the metrics were being dropped.  I saw that while the directApi outputs some useful information the proxy does not.  Not sure why that is but it seems reasonable to sync that up.
In digging some more it turns out that wavefront drops the entire batch of metrics when the value is NaN.  I got confirmation from the wavefront team on this yesterday.  The only way I figured this was to spend LOTs of time digging which would have been unnecessary if I had this trace-level info.

There is another bug that I eluded above, which I will file now, that the WavefrontReporter should not send metrics with value NaN or else the entire batch is dropped.

Do you have any concerns about adding this to reporter?  It would be quite useful.  Additionally, I would change the logger.info("successfully sent {} metrics to Wavefront", batch.size()) to debug as it is too verbose.  I will have to turn off the logging all together when I use the directApi.

jkschneider:
I appreciate the idea, and sorry it took you so long to debug. I'm not sure I like the idea of adding trace information in all registries, and don't think we should do something special for Wavefront in particular.

Your comment about default debug logging has been captured and will be fixed with #801.

PR #801, opened 2018-08-24
Use debug log level for successful metrics sent events.

izeye:
This PR changes the log level for successful metrics sent event to debug to lower verbosity by default.

### DIFF
--- a/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
+++ b/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
@@ -106,7 +106,7 @@ public class DatadogMeterRegistry extends StepMeterRegistry {
                                         m -> writeMeter(m, metadataToSend))
                                 ).collect(joining(",", "{\"series\":[", "]}")))
                         .send()
-                        .onSuccess(response -> logger.info("successfully sent {} metrics to datadog", batch.size()))
+                        .onSuccess(response -> logger.debug("Successfully sent {} metrics to Datadog.", batch.size()))
                         .onError(response -> logger.error("failed to send metrics to datadog: {}", response.body()));
             }
         } catch (Throwable e) {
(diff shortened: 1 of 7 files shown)


jkschneider:
We've consistently done info level for this same message in all the implementations. Should it be changed elsewhere as well?

Is log level purely subjective, or do you have a guiding principal in mind? Like a good sheep, I just followed the log level I saw in existing instrumentation (e.g. Spectator).

izeye:
@jkschneider Sorry, I didn't notice that the log messages are consistent across all the implementations. When I ran a sample, I just noticed that showing the successful metrics sent events is a bit verbose as a default. I just thought we could assume everything's fine unless there's no warning or error. This might be subjective, so if you disagree with a good reason, feel free to close this one.

jkschneider:
I don't have a strong opinion either way, but I've heard nothing but complaints about the verbosity level. So let's change them to debug 👍.

Example in the last comment on #790.

shakuzen:
For the record: I'm 👍 on switching to DEBUG log level for successful metrics publish because INFO seems too verbose for a normal event in a metrics aspect of an application. I would have already suggested as much if I were using an implementation that published, but I'm using prometheus, so I didn't encounter this.

dhaval24:
Debug makes a whole lot of sense. Specifically in prod excessive logs can be really irritating and also a hit on perf.

izeye:
@jkschneider @shakuzen @dhaval24 Thanks for the feedback! I changed all existing info-level logging to be debug-level and to use consistent messages. I also changed the target branch to `master` as @jkschneider has set the target milestone to the "1.1.0-rc.1".

izeye:
I resolved conflicts and rebased onto the current master branch.

izeye:
I added the same log for Wavefront proxy in e74e9b2e788e08e870e8461a8c15c59cba75c105 as the following argument from @scottmf in https://github.com/micrometer-metrics/micrometer/issues/790#issuecomment-413946714 seems valid to me:

> I saw that while the directApi outputs some useful information the proxy does not. Not sure why that is but it seems reasonable to sync that up.

jkschneider:
Thanks for your patience @izeye. Just wanted to get that `HttpClient` interface in first before considering this.

PR #804, opened 2018-08-24
Polish AbstractTimeWindowHistogram.takeValueSnapshot()

izeye:
This commit merges two `if` statements which target the same purpose in the end but return different values, which is a bit confusing. Decided to use `null` over zero-sized array to avoid object creation.

Along the way, this commit also removes an unused private constant.


jkschneider:
I'm concerned that null vs. zero-sized array is a breaking API change. Given that snapshots are only taken as part of publishing which should be periodic and generates a lot of other garbage along the way, is this optimization worth the potential NPEs?

izeye:
@jkschneider Thanks for the feedback and sorry if I wasn't clear in the first place. This PR intended to try to use `null` to mean no value consistently. The `null` will be used only in [this location](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/HistogramSnapshot.java#L48) and replaced with a constant of zero-sized array there in the end. So there's no additional possibility for NPEs unless I'm missing something.

PR #806, opened 2018-08-24
Azure Application Insights support for Micrometer

dhaval24:
This PR introduces Azure Application Insights Registry to transport Metrics to Azure Monitor. 

Default Publishing Interval : 1 minute
Aggregation on Client Side : Yes

In order to publish metrics to Application Insights you would need to create an Application Insights resource from Azure portal and use the intrumentation key in 

`management.metrics.export.azure.application-insights.instrumentation-key` property.

In case you are using ApplicationInsights SpringBoot starter in your SpringBoot Applications there is no need to specify instrumentation key with this new property. You can continue to provide instrumentation key using - `azure.application-insights.instrumentation-key` property and Spring Auto configuration would be able to use the same for Micrometer.

Please note that there is no separete distinction as Timer, Gauges, etc in Azure. We use a consistent metric data structure as specified in this bond specification -

https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/DataPoint.bond


dhaval24:
@jkschneider please do check on the locks. The CI/CD is failing because of lock state issue.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/azure/AzureProperties.java
```diff
@@ -0,0 +1,38 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
```
SergeyKanzhelev: perhaps the year in copyright can be changed to the 2018?

dhaval24: @jkschneider perhaps its the appropriate suggestion from Sergey. We should update copyright to 2018. Let me know.

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
```diff
@@ -0,0 +1,301 @@
...
+     * @param meterName Name of the Function Timer
+     * @param properties Dimensions of the metric
+     * @param meter FunctionTimer type meter holding rate aggregated metric values
+     */
+    private void trackTimer(String meterName, Map<String, String> properties, FunctionTimer meter) {
+        MetricTelemetry metricTelemetry = createAndGetBareBoneMetricTelemetry(meterName, properties);
+        metricTelemetry.setValue(meter.totalTime(getBaseTimeUnit()));
+        metricTelemetry.setCount((int)meter.count());
+        client.trackMetric(metricTelemetry);
+        logger.trace(String.format("sent metric with name %s", metricTelemetry.getName()));
```
SergeyKanzhelev: it may help troubleshoot if you will include the type of the metric in the message `sent timer metric with the name`

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
```diff
@@ -0,0 +1,301 @@
...
+                            trackTimer(name, properties, ((FunctionTimer) meter));
+                        } else if (meter instanceof DistributionSummary) {
+                            trackDistributionSummary(name, properties, ((DistributionSummary) meter));
+                        } else if (meter instanceof LongTaskTimer) {
+                            trackLongTaskTimer(id, properties, ((LongTaskTimer) meter));
+                        } else {
+                            trackMeter(name, properties, meter);
+                        }
+                } catch (Exception e) {
+                    logger.trace(String.format("Failed to track metric with name %s", getConventionName(meter.getId())));
```
SergeyKanzhelev: should this message be marked as warning or error?

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
```diff
@@ -0,0 +1,301 @@
...
+        if (StringUtils.isEmpty(clientConfig.getInstrumentationKey())) {
+
+            // Pick Instrumentation Key From the Config if not set via instrumentationKey / Environment Variable
+            clientConfig.setInstrumentationKey(config.instrumentationKey());
+        }
+
+        requireNonNull(clientConfig);
+        requireNonNull(clientConfig.getInstrumentationKey());
+
+        this.client = new TelemetryClient(clientConfig);
```
SergeyKanzhelev: set SDK version for this client to ease troubleshooting of metric values. Also register it here https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/SDK-VERSIONS.md

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        if (StringUtils.isEmpty(clientConfig.getInstrumentationKey())) {
+
+            // Pick Instrumentation Key From the Config if not set via instrumentationKey / Environment Variable
+            clientConfig.setInstrumentationKey(config.instrumentationKey());
+        }
+
+        requireNonNull(clientConfig);
+        requireNonNull(clientConfig.getInstrumentationKey());
```
SergeyKanzhelev: this check is unnecessary as `config.instrumentationKey()` will throw anyway.

SergeyKanzhelev: If keeping it here - check also for empty string

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
```diff
@@ -0,0 +1,301 @@
...
+                        } else if (meter instanceof DistributionSummary) {
+                            trackDistributionSummary(name, properties, ((DistributionSummary) meter));
+                        } else if (meter instanceof LongTaskTimer) {
+                            trackLongTaskTimer(id, properties, ((LongTaskTimer) meter));
+                        } else {
+                            trackMeter(name, properties, meter);
+                        }
+                } catch (Exception e) {
+                    logger.trace(String.format("Failed to track metric with name %s", getConventionName(meter.getId())));
+                    client.trackException(e);
```
SergeyKanzhelev: Errors reported to customer telemetry as "SDK errors" using `trackTrace` with the prefix "AI:" and synthetic source set to `SDKTelemetry`. 

Tracking SDK error as custom exception is very confusing

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
dhaval24: A warning makes sense. Error level could be for something fatal happening in the customer code itself. I will change this to warning level. Possibly add a trackTrace() call also for investigation.

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
dhaval24: @SergeyKanzhelev would you suggest having something like this: mm_java:version_numner where "mm" is the prefix ?

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
dhaval24: Makes sense. Fixed it.

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
dhaval24: Removed, thanks its redundant not needed.

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
dhaval24: Please let me know if the new proposed name in the change works. If yes I will make changes to the above link too.

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
dhaval24: updated.

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
dhaval24: Updated.

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
```diff
@@ -0,0 +1,318 @@
...
+     */
+    private void trackTimer(String meterName, Map<String, String> properties, Timer meter) {
+        MetricTelemetry metricTelemetry = createAndGetBareBoneMetricTelemetry(meterName, properties);
+        metricTelemetry.setValue(meter.totalTime(getBaseTimeUnit()));
+        metricTelemetry.setCount((int)meter.count());
+
+        // TODO: Is setting min to 0th percentile value apt workaround?
+        HistogramSnapshot snapshot = meter.takeSnapshot();
+        Optional<ValueAtPercentile[]> opt = Optional.ofNullable(snapshot.percentileValues());
+        opt.ifPresent(u -> { if (u.length > 0) metricTelemetry.setMin(u[0].value(TimeUnit.MILLISECONDS)); });
```
dhaval24: @macrogreg it should be noted that min is not always emitted by Micrometer and it depends if there is percentile snapshot available. Do we want to keep this behavior to emit 0th percentile in place of min when available?

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
SergeyKanzhelev: Prefix like `mm_` in proposed example is used to indicate platform/agent. Please use simply `mmj:version_number`

samples/micrometer-samples-boot1/src/main/resources/application.yml
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-server.port: 8080
+server.port: 8083
```
SergeyKanzhelev: it doesn't look like this change belong to this PR

samples/micrometer-samples-core/build.gradle
```diff
@@ -4,15 +4,16 @@ dependencies {
     compile('ch.qos.logback:logback-classic:1.2.+') {
         force = true
     }
-    compile('org.slf4j:slf4j-api:1.7.25') {
+    compile('org.slf4j:slf4j-api:1.7.5') {
```
SergeyKanzhelev: should it be part of this PR?

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/utils/SampleConfig.java
```diff
@@ -22,6 +22,6 @@
 public class SampleConfig {
     public static MeterRegistry myMonitoringSystem() {
         // Pick a monitoring system here to use in your samples.
-        return SampleRegistries.jmx();
+        return SampleRegistries.azure("");
```
SergeyKanzhelev: looks like some debug changes creeped to the PR

dhaval24: Updated, thanks

samples/micrometer-samples-core/build.gradle
dhaval24: This is because locks were updated. It should be fine. @jkschneider can comment on that.

samples/micrometer-samples-boot1/src/main/resources/application.yml
dhaval24: Fixed

implementations/micrometer-registry-azure/src/main/java/io/micrometer/azure/AzureMeterRegistry.java
dhaval24: Sure, mm:version_number makes sense to me. Will update it.

dhaval24:
Did rebranding of names to AzureMonitor according to the internal discussion within the team.

dhaval24:
@SergeyKanzhelev can you officially sign off the PR as you already mentioned it looks good to you :)

dhaval24:
Completed all rebranding changes.

samples/micrometer-samples-core/build.gradle
izeye: @dhaval24 I have the same question on this with @SergeyKanzhelev . This looks irrelevant to this PR.

implementations/micrometer-registry-azuremonitor/build.gradle
```diff
@@ -0,0 +1,9 @@
+apply plugin: 'org.junit.platform.gradle.plugin'
+
+dependencies {
+    compile project(':micrometer-core')
+    compile 'com.microsoft.azure:applicationinsights-core:2.1.2-SNAPSHOT'
```
izeye: @dhaval24 This needs to have a release version.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -48,7 +57,11 @@ public StepMeterRegistry(StepRegistryConfig config, Clock clock) {
     }
 
     public void start() {
-        start(Executors.defaultThreadFactory());
+        start((r) -> {
```
izeye: @dhaval24 This change looks irrelevant to this and you’ve already raised #814 and #815 , so it’d be good to be reverted here.

dhaval24: @izeye we can either merge #815 before and then merge this one, or merge this before #815 and abonden it. I think it would produce the same result. Whatever is preferred I am fine with it.

izeye:
@dhaval24 I’m not familiar with Azure Application Insights and didn’t look into it much yet but leaved some general comments for a start.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
izeye: @dhaval24 IMHO, any PR would be good to focus on it own purpose in general.

implementations/micrometer-registry-azuremonitor/build.gradle
dhaval24: @izeye yes ofcourse. Soon 2.1.2 version should be released (likely before 1.1.0-rc goes for Micrometer) and I will update this.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/azuremonitor/AzureMonitorProperties.java
```diff
@@ -0,0 +1,38 @@
...
+
+import io.micrometer.spring.autoconfigure.export.StepRegistryProperties;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * {@link ConfigurationProperties} for configuring Azure Application Insights metrics export.
+ *
+ * @author Dhaval Doshi
+ */
+@ConfigurationProperties(prefix = "management.metrics.export.azure.application-insights")
```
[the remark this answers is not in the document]
dhaval24: @izeye I would have definitely liked this. Infact we should do this. @SergeyKanzhelev are you fine with this change? It will just be different from our core SDK. Though there would be no backport issue here. If ApplicationInsights-SpringBoot-Starter is already present on class path then it's autoconfig will take precedence over Micrometer's and the ikey should be resolved with the property there which is `azure.application-insights.instrumentation-key` . Let me know.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
dhaval24: done

izeye:
For the Spring Boot auto-configuration, I’m not sure which one should be the first between Spring Boot 2 and legacy support but if the legacy support intends to be ported from Spring Boot 2, it’d be good to propose against the Spring Boot repo first IMO.

@jkschneider @snicoll Thoughts?

dhaval24:
@izeye regarding auto-config I can file a PR against Boot 2 repo too (infact have the code in local fork tested https://github.com/dhaval24/spring-boot/tree/AzureMonitorRegistryAutoconfig) I will leave it upto boot team to decide on this. However, a thought can be autoconfig can be part of legacy first in rc1 release of Micrometer. Once it has no breaks (hopefully :) ) then it can be ported to boot 2 when stable release happens. However as I said I am fine with whatever boot team thinks.

izeye:
@dhaval24 IMO, as the name (legacy) suggests, it doesn’t sound good to be used for incubating something. But it’d good to wait for feedback on this from @jkschneider or Spring Boot team.

snicoll:
Thanks @izeye. I've already shared my reservation to host this in Spring Boot itself depending on the dependency management that the PR would bring. Besides, Microsoft already has [a Spring Boot integration project](https://github.com/Microsoft/azure-spring-boot) and I was wondering if this was considered instead. 

In any case, I agree with Johnny that we should figure things out before considering the legacy project.

dhaval24:
@snicoll @izeye thanks for your comments and reviews. What would be the way you suggest here Stephane? I understand you would like to review dependency management which will come with this.  We should definitely figure out things. I would let @jkschneider review the registry contribution and also voice his opinion and suggestions on integration too.

PR #808, opened 2018-08-27
Elastic reporter improvements

spinscale:
This commit improves the elastic reporter behaviour in a few ways:

First, it uses way less objects by removing the builder class and the
usage of streams. Second it does not generate one big in memory string
of the metrics, but rather writes the metrics directly to the output
stream - there is still a string builder per metric.

The first line of each of the bulk requests now does not include the index name
and the type as those can be specified in the URL, thus reducing the
size of the bulk request that needs to be sent over the wire.

The encoding of the basic auth data only needs to be done once in the
constructor instead of doing it in each request.

Lastly tests for each different kind of metric have been added.

Reviewers note: I would like to use ` JSONAssert.assertEquals();` but do not have any experience with locking dependencies and I tried and suddently tons of files were updated.  If you help me doing that I am happy to update this PR doing that as well, so we can ensure, the reporter is producing valid JSON per line.


jkschneider:
Thanks for the hard work @spinscale !

PR #815, opened 2018-08-29
Making ThreadFactory in StepMeterRegistry Daemon

### LINKED ISSUE #814: StepMeter Registry Should use Daemon Thread Factory
The current implementation of `StepMeterRegistry` uses `Executors.defaultThreadFactory()` which is a non-demon ThreadFactory and holds a reference.

This prevents applications to terminate unless `registry.close()` method is called explicitly. This is undesired behavior and Applications should shutdown as soon as they complete.

Steps to Reproduce the behavior:

1. Create a console java application in use a registry which extends StepMeterRegistry

A sample main method can be like this: 

```
    MeterRegistry registry = new AzureMeterRegistry(new AzureConfig() {
      public String get(String key) {
        return null;
      }
      public Duration step() {
        return Duration.ofSeconds(10);
      }
    }, null, Clock.SYSTEM);
    Counter counter = registry.counter("sample.counter");
    try {
      counter.increment(10);
      Thread.sleep(2000);
      counter.increment(5);
      Thread.sleep(2000);
      counter.increment(30);
      Thread.sleep(6000);
      //registry.close();
    } catch (InterruptedException ex) {
      ex.printStackTrace();
    }

  }
```

The application doesn't terminate after the main method has completed.

dhaval24:
Fix #814 

This makes default ThreadFactory in `StepMeterRegistry` to be daemon and hence allows the applications to terminate gracefully.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
@@ -15,7 +15,17 @@
  */
 package io.micrometer.core.instrument.step;
 
-import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.FunctionCounter;
+import io.micrometer.core.instrument.FunctionTimer;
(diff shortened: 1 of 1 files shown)


dhaval24:
@jkschneider I hope this is a good enough fix :)  and would be in 1.1.0 RC1

jkschneider:
Merged with fe32bea8.

Issue #812, opened 2018-08-29
Hystrix latency percentile metrics


guxlightyear:
I am currently developing a Spring Boot 2.0 application using Hystrix, and Prometheus for metrics.

I have noticed that the endpoint `/actuator/hystrix.stream` reports correctly a number of metrics, such as various latency percentiles [example hystrix-stream output attached](https://github.com/micrometer-metrics/micrometer/files/2332462/example-hystrix-stream-output.txt), but `/actuator/prometheus` does not contain those [example prometheus output attached](https://github.com/micrometer-metrics/micrometer/files/2332459/example-prometheus-output.txt)

I am under the impression that this commit https://github.com/micrometer-metrics/micrometer/commit/da19a7eceebf7cecddac035b74144b985db1bc17 (merged into master 2 weeks ago) will fix it.

Is that the case? If so, is it going to be included in Micrometer 1.0.7? and when will that be released?

For reference, and in case this problem is simply user error with my configuration, these are the software versions that I am currently using:
```
Spring Boot: 2.0.4.RELEASE
Micrometer: 1.0.6
Java: 1.8
Hystrix: 1.5.12
```

Relevant fragments from my build.gradle:
```
compile 'org.springframework.cloud:spring-cloud-starter-hystrix:1.4.5.RELEASE'
compile "org.springframework.boot:spring-boot-starter-actuator"
compile 'io.micrometer:micrometer-core'
compile 'io.micrometer:micrometer-registry-prometheus'
```

Relevant fragments from application.yml
```
management:
  security:
    enabled: false
  endpoints:
    web:
      exposure:
        include: "*"
  metrics:
    export:
      prometheus:
        enabled: true
    hystrix:
      enabled: true
```

Thanks you,
Gustavo

jkschneider:
The change in da19a7e is targeted for Micrometer 1.1.0. Expect a release candidate in the next two weeks. If you want to ship percentile data for hystrix latencies, you'll want to use one of the configuration options:

```
management.metrics.distribution.percentiles.hystrix=0.95,0.99
management.metrics.distribution.percentiles-histogram.hystrix=true
```

Choose one or the other, depending on your intended use and whether your monitoring system supports aggregable percentiles based on histograms.

guxlightyear:
Excellent, thanks a lot @jkschneider. That was really helpful and worked like a charm.

I had gone through the documentation and hadn't found this feature. Can you please add a link to the relevant documentation for future reference?

[...]

PR #820, opened 2018-08-31
Include useful percentiles for hystrix execution and total latency metrics

guxlightyear:
This PR is intented to solve the thread I opened a few days ago: https://github.com/micrometer-metrics/micrometer/issues/812

The idea is to publish the following percentiles for Hystrix total and execution latencies: https://github.com/Netflix/Hystrix/wiki/Metrics-and-Monitoring#latency-percentiles-hystrixcommandrun-execution-gauge


jkschneider:
Copying from the thread on #812:

As a rule, we never choose to ship percentiles as part of the default instrumentation, but leave it for you to configure later. It isn't free to do so, and percentiles aren't needed everywhere. If you want to ship percentile data for hystrix latencies, we'd like you to use one of the configuration options:

```
management.metrics.distribution.percentiles.hystrix=0.95,0.99
management.metrics.distribution.percentiles-histogram.hystrix=true
```

Choose one or the other, depending on your intended use and whether your monitoring system supports aggregable percentiles based on histograms.

PR #832, opened 2018-09-07
Split method "publishBatch" into smaller method

stefanbirkner:
Reduce size of method "publish". The reader of the method can now get an overview of what the method is doing very fast.

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -97,12 +97,7 @@ public class InfluxMeterRegistry extends StepMeterRegistry {
         createDatabaseIfNecessary();
 
         try {
-            String write = "/write?consistency=" + config.consistency().toString().toLowerCase() + "&precision=ms&db=" + config.db();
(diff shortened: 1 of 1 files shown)


jkschneider:
Thanks for your work on this. Sorry to say that I think the work on #842 will move much of this code out of this registry anyway. Also #782 will remove all the switch cases. I'm going to close this as a duplicate if that's ok?

stefanbirkner:
I looked at both issues/PRs. Can we merge this PR independent of the others? I like to improve `InfluxMeterRegistry` and this two commits will help. #782 in the current state will not be effected by it because it just changes the internals of the now new method `publishBatch`. #842 still needs a method like `buildInfluxPublishUrl`. So no reason for waiting for #842.

stefanbirkner:
This gives you an impression about what I'm heading for: https://github.com/stefanbirkner/micrometer/blob/e8359396e82d632bec8f5bb1c727c5fb45552c15/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java You can see that I'm writing the data directly to the `OutputStream` without arranging it in-memory before.

stefanbirkner:
@jkschneider I rebased the PR against the changes because I think it is still helpful. What do you think?

PR #846, opened 2018-09-12
Upgrade to Gradle Wrapper 4.10.1

izeye:
This PR upgrades to Gradle Wrapper 4.10.1.


jkschneider:
I fear the cache re-download that comes with this... The joys of living in a rural area ;)

Issue #847, opened 2018-09-12
Log4j2 Metrics


steven-sheehy:
Similar to LogbackMetrics, it would be nice to have log4j2 metrics. Incoming PR

PR #848, opened 2018-09-12
Log4j2 metrics

steven-sheehy:
Issue #847. Adds Log4j2 metric support. I've patterned the implementation after the LogbackMetrics as much as possible. I had to update the lockfiles locally to add the log4j2 dependency, but I didn't check those in since I'm not sure how those are handled.

### DIFF
--- a/micrometer-spring-legacy/src/test/java/io/micrometer/spring/autoconfigure/MetricsAutoConfigurationTest.java
+++ b/micrometer-spring-legacy/src/test/java/io/micrometer/spring/autoconfigure/MetricsAutoConfigurationTest.java
@@ -21,6 +21,7 @@ import io.micrometer.core.instrument.binder.MeterBinder;
 import io.micrometer.core.instrument.binder.jvm.ClassLoaderMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmThreadMetrics;
+import io.micrometer.core.instrument.binder.logging.Log4j2Metrics;
 import io.micrometer.core.instrument.binder.logging.LogbackMetrics;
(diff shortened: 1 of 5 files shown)


micrometer-spring-legacy/src/test/java/io/micrometer/spring/autoconfigure/MetricsAutoConfigurationTest.java
```diff
@@ -94,6 +94,7 @@ public void requestMappingIsInstrumented() {
     @Test
     public void automaticallyRegisteredBinders() {
         assertThat(context.getBeansOfType(MeterBinder.class).values())
+            .hasAtLeastOneElementOfType(LogbackMetrics.class)
```
spencergibb: `Log4j2Metrics`?

steven-sheehy: Doh. I'll fix

steven-sheehy:
@spencergibb  So CI failed because of the lockfiles. Am I supposed to run the following command and check in all changed files? Even untracked ones?

`./gradlew classes --update-locks org.apache.logging.log4j:log4j-core`

spencergibb:
@jkschneider or someone else will need to answer that one.

jkschneider:
I'll take care of the locks. I think in the future we'll unlock master by default so folks can add new features without thinking about this. But we are so close to 1.1.0 at this point, that this can be addressed a little later.

jkschneider:
Merged with 77855952. Thanks for the handy feature @steven-sheehy!

dhaval24:
This is awesome feature. @jkschneider we were discussing this couple of days back and here there is a great work! This should complete the story for metrics from logs.

jkschneider:
Yep, @steven-sheehy beat me to it!

Issue #842, opened 2018-09-11
Allow for injection of HTTP interface into push-based registries.


jkschneider:
We'd like to expand the configuration of push-based registries to include a mechanism to supplant the HTTP client interface with your own. Currently all pushed based registries use `java.net.HttpURLConnection` to minimize dependencies, but this leaves gaps in handling of TLS, the ability to skip SSL verification, intermediate proxies, etc. Other HTTP clients contain much more evolved facilities for dealing with these concerns, and often they are already part of the user's classpath (e.g. Netty, Spring's `WebClient`, etc.). Rather than duplicating the efforts in these various clients, we should just allow them to be plugged into the registry.

jkschneider:
For documentation purposes: we should add a Resilience4J retry example as well.

PR #854, opened 2018-09-14
Always convert Strings to byte array using UTF-8

stefanbirkner:


jkschneider:
This is now enforced down at the `HttpClient` interface with #842.

Issue #771, opened 2018-08-03
OkHttpMetricsEventListener default uriMapper


danieljamesscott:
The default uriMapper in OkHttpMetricsEventListener:

https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java#L142

Looks a little strange, it is looking for a header named 'URI_PATTERN'.

In the tests:

https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java#L37

This default is being overridden to extract the path from the request object.

This seems inconsistent, is there a reason why it was done like that?

To make it consistent, either the default should be to extract the path from the request object (my preference), or the test should add a header to the request so that users can use the tests to discover how to add path information to the metrics.

jkschneider:
Thanks @danieljamesscott. Don't know the reason it was created the way it was, but your suggestions are perfectly reasonable.

danieljamesscott:
Do you have a preference for the default? I can do a PR to change it, maybe with a simple static method to restore the old functionality?

jkschneider:
I know this has been a while, but can you shed any light on the reason things are the way they are @bjartek?

bjartek:
The reason it does not fetch the path from the request object is that this path has the url pattern with resolved placeholders.

The test should add a header here to be consistent.

jkschneider:
So it's maybe a little unfortunate for general use, but OkHttp doesn't provide a native way of retrieving the parameterized URI. Anything short of trying to retrieve a custom header would be bad for tag cardinality.

kppullin:
In case someone runs across this like I did, if you're using retrofit on top of OkHttp, here's a bit of code that'll set the URI to the path set the retrofit annotated method (uses a bit of guava and apache commons, but easy to adjust as needed):

```
   private static final ImmutableSet retrofitHttpMethodAnnotationName;

    static {
        retrofitHttpMethodAnnotationName = ImmutableSet.of(
                GET.class.getName(),
                POST.class.getName(),
                PUT.class.getName(),
                DELETE.class.getName(),
                PATCH.class.getName(),
                HEAD.class.getName(),
                OPTIONS.class.getName()
        );
    }

...

              OkHttpMetricsEventListener
[...]

PR #860, opened 2018-09-18
Add a test for OkHttpMetricsEventListener.URI_PATTERN

izeye:
This PR adds a test for `OkHttpMetricsEventListener.URI_PATTERN`.

Along the way, this PR also polishes by:

- Improving Javadoc.
- Adding an assertion to an unexpected path in a test. This revealed that `OkHttpMetricsEventListenerTest.timeNotFound()` has a wrong expectation for `IOException`, so it has been removed.
- Adding missing assertions for `extraTags`.
- Polishing general stuff.

See gh-771


izeye:
I rebased onto the current `1.0.x` to remove the change for a missing license header as it has already been resolved in #862.

jkschneider:
Thanks!

Issue #861, opened 2018-09-18
Metrics are reported even though they are disabled


Edholm:
I have disabled metrics in some environments (tests for example), but they are still reported once when the application is being shut down.

```
java.net.ConnectException: Connection refused (Connection refused)
	at java.net.PlainSocketImpl.socketConnect(Native Method) ~[na:1.8.0_181]
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350) ~[na:1.8.0_181]
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206) ~[na:1.8.0_181]
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188) ~[na:1.8.0_181]
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392) ~[na:1.8.0_181]
	at java.net.Socket.connect(Socket.java:589) ~[na:1.8.0_181]
	at java.net.Socket.connect(Socket.java:538) ~[na:1.8.0_181]
	at java.net.Socket.<init>(Socket.java:434) ~[na:1.8.0_181]
	at java.net.Socket.<init>(Socket.java:244) ~[na:1.8.0_181]
	at javax.net.DefaultSocketFactory.createSocket(SocketFactory.java:277) ~[na:1.8.0_181]
	at com.codahale.metrics.graphite.Graphite.connect(Graphite.java:126) ~[metrics-graphite-3.2.6.jar:3.2.6]
	at com.codahale.metrics.graphite.GraphiteReporter.report(GraphiteReporter.java:240) ~[metrics-graphite-3.2.6.jar:3.2.6]
	at com.codahale.metrics.ScheduledReporter.report(ScheduledReporter.java:243) [metrics-core-3.2.6.jar:3.2.6]
	at io.micrometer.graphite.GraphiteMeterRegistry.close(GraphiteMeterRegistry.java:90) [micrometer-registry-graphite-1.0.6.jar:1.0.6]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_181]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_181]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_181]
[...]

PR #863, opened 2018-09-18
Handle reporter with config.enabled() consistently

izeye:
This PR changes to handle `this.reporter` with `this.config.enabled()` consistently in `GraphiteMeterRegistry`.

Closes gh-861


implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteMeterRegistry.java
```diff
@@ -78,20 +77,26 @@ private static GraphiteReporter defaultGraphiteReporter(GraphiteConfig config, C
...
+        if (this.config.enabled()) {
+            this.reporter.start(config.step().getSeconds(), TimeUnit.SECONDS);
+        }
     }
 
     @Override
     public void close() {
-        if (config.enabled()) {
-            reporter.report();
+        if (this.config.enabled()) {
```
jkschneider: Different from the Spring Boot style, I don't love the excess `this` statements, and prefer `this` to be used only when necessary.

jkschneider: I've let it go to save time for a while, but thought it worth mentioning.

izeye: @jkschneider Ah, sorry. I didn't know the convention here. I prefer to use `this` for fields explicitly as it makes me easily notice they're class member fields without IDE like in GitHub. If you still disagree, I'll revert the change and be happy to comply. Just let me know.

jkschneider: I do understand the reasoning, but still have a preference for dropping extraneous `this` prefixes. I know it's subjective, but personally I find the cost of extra verbosity in the place where I spend most of my time (in the IDE) not worth the benefit of field differentiation where I spend relatively little time (browsing in Github).

izeye: @jkschneider Thanks for the feedback! I'll try to stick with the convention unless I'm confused with other projects. If I break it, just poke me to realise. Then I'll polish to comply. I did the opposite polish for the `this` keyword and hope you're happy with the change now 😄

jkschneider: Thanks for the understanding!

PR #867, opened 2018-09-20
Polish Log4j2Metrics

izeye:
This PR polishes `Log4j2Metrics` by:

- Adding a missing `@since` tag.
- Extracting the meter name to a constant.
- Removing `loggerContext.updateLoggers(configuration);` in `bindTo()` as it doesn't seem to be necessary.

Note that the first commit which is already in the 1.0.x branch is only for passing a PR build.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheus/MicrometerCollectorTest.java
+++ b/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheus/MicrometerCollectorTest.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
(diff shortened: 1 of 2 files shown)


izeye:
@steven-sheehy Ah, it might be an oversight on my side.

At first I noticed that `loggerContext.updateLoggers(configuration);` is missing with `removeFilter()` , so I tried to add it with a test but the test works without it. That led me to think `loggerContext.updateLoggers(configuration);` isn't necessary and when I removed it the existing test passed.

I'll create another PR which restores it and additionally adds `loggerContext.updateLoggers(configuration);` with `removeFilter()`.

@steven-sheehy Does that make sense?

izeye:
@steven-sheehy Thanks for spotting it! I created #872 to restore it.

PR #872, opened 2018-09-22
Restore updateLoggers() in Log4j2Metrics.updateLoggers()

izeye:
This PR also adds `updateLoggers()` after the filter has been removed in `Log4j2Metrics.close()`.

See gh-867


steven-sheehy:
Thanks, looks good to me. Good catch on the updateLoggers for removeFilter.

PR #881, opened 2018-09-29
Add outcome tag to Servlet metrics

izeye:
This PR adds `outcome` tag to Servlet metrics by porting partially from https://github.com/spring-projects/spring-boot/pull/14486.

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/servlet/WebMvcTags.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/servlet/WebMvcTags.java
@@ -31,6 +31,7 @@ import javax.servlet.http.HttpServletResponse;
  *
  * @author Jon Schneider
  * @author Andy Wilkinson
+ * @author Michael McFadyen
  */
 @NonNullApi
 public final class WebMvcTags {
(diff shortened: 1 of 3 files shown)


micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/servlet/WebMvcTags.java
```diff
@@ -47,6 +48,18 @@
 
     private static final Tag STATUS_UNKNOWN = Tag.of("status", "UNKNOWN");
 
+    private static final Tag OUTCOME_UNKNOWN = Tag.of("outcome", "UNKNOWN");
```
checketts: Since this is a derivative of status code wouldn't `statusCodeGroup` be a little clearer?

izeye: @checketts I thought the similar thing (`statusGroup`) but this is just porting from Spring Boot 2.x, so I didn't touch anything but a bit of polishing for compatibility.

/cc @wilkinsona

wilkinsona:
The term "outcome" came from the [original Boot issue](
https://github.com/spring-projects/spring-boot/issues/13801). @jkschneider [liked it](
https://github.com/spring-projects/spring-boot/issues/13801#issuecomment-405941027) and I do too. FWIW, I think it's a better name than "statusCodeGroup" which I find a little cumbersome. I'd also expect the values of "statusCodeGroup" to be 2xx, 3xx, 4xx, or 5xx rather than the more descriptive terms that Boot's using now which describe the outcome of the request.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/servlet/WebMvcTags.java
wilkinsona: The term "outcome" came from the [original Boot issue](
https://github.com/spring-projects/spring-boot/issues/13801). @jkschneider [liked it](
https://github.com/spring-projects/spring-boot/issues/13801#issuecomment-405941027) and I do too. FWIW, I think it's a better name than "statusCodeGroup" which I find a little cumbersome. I'd also expect the values of "statusCodeGroup" to be 2xx, 3xx, 4xx, or 5xx rather than the more descriptive terms that Boot's using now which describe the outcome of the request.

PR #882, opened 2018-09-29
Split method "publishBatch" into smaller method

stefanbirkner:
Reduce size of method "publish". The reader of the method can now get an overview of what the method is doing very fast.

This is #832 again but for the current master.


jkschneider:
Again, this is dependent on #842. That is the current item under development.

stefanbirkner:
Ok. I was not aware that there is ongoing work on #842. I only saw that #782 was fixed.

PR #884, opened 2018-10-01
Add a missing @since for Meter.Type.match()

izeye:
This PR also completes the other missing parts of Javadoc for `Meter.Type.match()`.


izeye:
For readability, I also reordered `if`-`else` statements to align with the order of parameters in 8f3681540b25ea81abbd543375d595b383b52654.

PR #887, opened 2018-10-01
Measurement buffering for Statsd

breun:
The Statsd implementation sends out every individual measurement via UDP. In my company this causes loadbalancers to get flooded with UDP packets. To reduce the load on our Statsd infrastructure I have implemented buffering for Statsd, so multiple measurements can be sent in a single UDP packet. Depending on the maximum packet size used this can result in a very significant reduction of the number of UDP packets sent.


breun:
We have been running a patched version of Micrometer 1.0.6 with this change in production for a few weeks now.

jkschneider:
Thanks, wonderful feature!

PR #888, opened 2018-10-01
Reduce String object creation in ElasticMeterRegistry

izeye:
This PR changes to reduce `String` object creation in `ElasticMeterRegistry`.


izeye:
I also moved `INDEX_LINE` concatenation into `ElasticMeterRegistry.writeDocument()` in 957a2b6 as it reduces `String` object creation.

jkschneider:
Merged with c6025d2f.

PR #894, opened 2018-10-03
Remove invalid DEFAULT from some MeterRegistryConfig subclasses

izeye:
This PR removes invalid `DEFAULT` constants from some `MeterRegistryConfig` subclasses as having the `DEFAULT` constant doesn't seem to make sense if any mandatory property exists.

This PR also updates tests accordingly.


jkschneider:
Trying to think if there is a scenario where we could be more relaxed about mandatory properties like API keys if the registry is initially disabled.

Issue #896, opened 2018-10-03
2018-10-03 19:50:05 io.micrometer.elastic.ElasticMeterRegistry [ERROR] Error connecting to http://localhost:9200: java.net.MalformedURLException: For input string: "9200metrics-2018-10"


andrsuh:
I think It is because of absence of leading slash before the index name in the class ElasticConfig?
Maybe it is worth to add.

izeye:
@andrsuh Thanks for the report!

It seems to have been introduced in 5f198e394b629529824393239932c5013e6c8527. I created #897 to resolve it.

jkschneider:
This has been resolved incidentally with dd62f0c8 for #842.

PR #897, opened 2018-10-03
Add a leading slash for a publishing URI in ElasticMeterRegistry

izeye:
This PR adds a leading slash for a publishing URI in `ElasticMeterRegistry`.

Closes gh-896


jkschneider:
This was resolved incidentally in dd62f0c8.

PR #898, opened 2018-10-04
Add a switch to disable ScheduledMethodMetrics

izeye:
This PR changes the configuration for the `ScheduledMethodMetrics` to align with the other binders like the `JvmGcMetrics` by:

- Adding a switch to disable it.
- Making it back off when a custom one has been defined.

For easy review, the previous `MetricsAutoConfigurationTest` has been renamed to `MetricsAutoConfigurationIntegrationTest` as a new `MetricsAutoConfigurationTest` has been added as unit tests. It seems that Git fails to be aware of the rename although there's no change in the previous one.


izeye:
@jkschneider I intentionally targeted this to the 1.0.x branch as there's no way to disable it in 1.0.x line. Do you prefer this to be resolved in 1.1.0? I just assumed with the milestone you have set.

izeye:
I rebased onto the current 1.0.x branch to resolve conflicts.

PR #905, opened 2018-10-06
Update Wavefront integration

hanwavefront:
Updates to the Wavefront reporter:

- Added a dependency on wavefront-sdk-java to abstract away the sending-to-Wavefront logic and use the latest Wavefront API.
- Added support for Wavefront-specific metric types: Delta Counters and Wavefront Histograms.
- Updated Wavefront properties and config to support the latest Wavefront API and proxy configuration options.

We'll also need to update the following docs due to the updated configuration options:

- https://micrometer.io/docs/registry/wavefront
- https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-export-wavefront


implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontHistogram.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public double max() {
+        return delegate.getMax();
+    }
+
+    @Override
+    public HistogramSnapshot takeSnapshot() {
+        throw new UnsupportedOperationException(
```
jkschneider: Throwing an exception is not OK here. You can return a snapshot that contains the basic distribution statistics like count, mean, etc. at least.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontHistogram.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    public List<WavefrontHistogramImpl.Distribution> flushDistributions() {
+        return delegate.flushDistributions();
+    }
+
+    /**
+     * Fluent builder for Wavefront histograms.
+     */
+    static class Builder {
```
jkschneider: This builder should not be necessary. A user should never build an implementation-specific form of a meter.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
```diff
@@ -94,37 +131,54 @@ public void start(ThreadFactory threadFactory) {
         super.start(threadFactory);
     }
 
+    @Override
+    protected DistributionSummary newDistributionSummary(
+        Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, double scale) {
+        if (isWavefrontHistogram(id)) {
```
jkschneider: If wavefront histograms collect buckets, then here you are only making a decision about whether to use a wavefront histogram or a plainer distribution summary on the basis of whether `distributionStatisticConfig` has percentile histograms or SLA buckets enabled.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
```diff
@@ -183,10 +237,19 @@ private boolean directToApi() {
...
 
         Meter.Id id = summary.getId();
-        addMetric(metrics, id, "sum", wallTime, summary.totalAmount());
-        addMetric(metrics, id, "count", wallTime, summary.count());
-        addMetric(metrics, id, "avg", wallTime, summary.mean());
-        addMetric(metrics, id, "max", wallTime, summary.max());
+
+        if (summary instanceof WavefrontHistogram) {
+            if (wavefrontSender != null) {
+                sendWavefrontHistogram(id, ((WavefrontHistogram) summary).flushDistributions());
```
jkschneider: Are we only using the SDK to send histograms? Seems like we should send either all metrics or none through the SDK. If the former, need to be sure that the SDK supports injecting something like `HttpSender`.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
```diff
@@ -303,8 +421,72 @@ public Builder httpClient(HttpSender httpClient) {
...
+            histogramGranularities.add(HistogramGranularity.DAY);
+            return this;
+        }
+
+        @Nullable
+        private Tags globalTags() {
+            if (applicationTags == null) {
+                return null;
+            }
+            Tags globalTags = Tags.of(
```
jkschneider: We cannot assume any sort of structure around common tags. What makes sense for an IaaS looks totally different on a CaaS, etc.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontHistogram.java
hanwavefront: Updated to return a basic snapshot.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontHistogram.java
hanwavefront: Added public `wavefrontHistogram(...)` methods to `WavefrontMeterRegistry`, so users will be using those instead of the builder.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
hanwavefront: Wavefront histograms are reported differently from summaries and other metrics -- they're reported as distributions instead of points.  So deciding whether to use a wfhistogram based on `distributionStatisticConfig` could break backwards compatibility for existing users.  Hence why `isWavefrontHistogram()` is based on the presence of an injected tag.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConstants.java
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
```
OleksiiBondar: Should it be 2019?

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
hanwavefront: If a `WavefrontSender` is specified, the SDK is used to send all metrics and histograms.  If not specified, we're still using the old method (`HttpSender`, etc) for backwards compatibility, but the old method doesn't support the sending of wfhistograms (due to missing config options).

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConstants.java
```diff
@@ -0,0 +1,47 @@
...
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
+
+/**
+ * Class containing Wavefront-specific constants and related static methods.
+ *
+ * @author Han Zhang
+ */
+public class WavefrontConstants {
```
OleksiiBondar: Please add private constructor and final modifier to the class to prevent instance creation and class inheritance

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConstants.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public static final String WAVEFRONT_METRIC_TYPE_TAG_KEY = "wavefrontMetricType";
+
+    /**
+     * @param id                            The identifier for a metric.
+     * @param wavefrontMetricTypeTagValue   The tag value that identifies a particular
+     *                                      Wavefront-specific metric type.
+     * @return {@code true} if the id identifies the metric type, {@code false} otherwise.
+     */
+    static boolean isWavefrontMetricType(Meter.Id id, String wavefrontMetricTypeTagValue) {
```
OleksiiBondar: IMO, having this method is class named as *Constants looks confusing.
Would it be cleaner to rename class to something like `WavefrontMetricTypeEvaluator` with boolean method `evaluate`?

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontHistogram.java
```diff
@@ -0,0 +1,178 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
```
OleksiiBondar: 2019

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
hanwavefront: `ApplicationTags` are a Wavefront-specific set of pre-defined tags.  They're optional here.

OleksiiBondar:
I'm not seeing any unit test being added/updated. Please add for newly added functionality

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
```diff
@@ -86,6 +107,21 @@ private WavefrontMeterRegistry(WavefrontConfig config, Clock clock, ThreadFactor
         start(threadFactory);
     }
 
+    private WavefrontMeterRegistry(WavefrontConfig config, Clock clock, ThreadFactory threadFactory,
+                                   WavefrontSender wavefrontSender, @Nullable Tags globalTags,
+                                   Set<HistogramGranularity> histogramGranularities) {
+        super(config, clock);
+        this.config = config;
+        this.httpClient = null;
```
OleksiiBondar: Very confusing and error-prone. Both constuctors have a lot in common.
How about define super constructor for both `wavefrontSender` and `httpSender`? Something like that:
```

    public WavefrontMeterRegistry(WavefrontConfig config, Clock clock, ThreadFactory threadFactory,
	    HttpSender httpClient, @Nullable Tags globalTags) {
	this(config, clock, threadFactory, null, httpClient, null, globalTags);
	if (directToApi() && config.apiToken() == null) {
	    throw new MissingRequiredConfigurationException(
		    "apiToken must be set whenever publishing directly to the Wavefront API");
	}

	start(threadFactory);
    }

    public WavefrontMeterRegistry(WavefrontConfig config, Clock clock, ThreadFactory threadFactory,
	    WavefrontSender wavefrontSender, Set<HistogramGranularity> histogramGranularities,
	    @Nullable Tags globalTags) {
	this(config, clock, threadFactory, wavefrontSender, null, histogramGranularities, globalTags);
	start(threadFactory);
    }

    private WavefrontMeterRegistry(WavefrontConfig config, Clock clock, ThreadFactory threadFactory,
	    @Nullable WavefrontSender wavefrontSender, @Nullable HttpSender httpClient,
	    @Nullable Set<HistogramGranularity> histogramGranularities, @Nullable Tags globalTags) {
	super(config, clock);
	this.config = config;
	this.httpClient = httpClient;
	this.wavefrontSender = wavefrontSender;
	this.globalTags = globalTags;
	this.histogramGranularities = histogramGranularities;

	config().namingConvention(new WavefrontNamingConvention(config.globalPrefix()));
    }
```

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConstants.java
hanwavefront: Updated.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConstants.java
hanwavefront: Updated.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConstants.java
hanwavefront: Updated

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontHistogram.java
hanwavefront: Updated

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
hanwavefront: Updated. Thanks for reviewing.

implementations/micrometer-registry-wavefront/src/test/java/io/micrometer/wavefront/WavefrontHistogramTest.java
```diff
@@ -0,0 +1,4 @@
+package io.micrometer.wavefront;
+
+public class WavefrontHistogramTest {
```
OleksiiBondar: Em, I feel I have more comments to this class than lines inside :) 
* copyrights
* as per JUnit 5 conventions class should be package-private
* very minor: where is content ? :)

hanwavefront: Bad commit, filling in the gaps now.

hanwavefront: Updated the access level.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
jkschneider: We need to collapse these two concepts I think. Having to choose between being able to use (for example) a proxy and being able to send a histogram is not a choice we want our users to have to be required to make. I'd like to see an update to the wavefront-sdk to permit sender injection, or that at least for now, we persist in sending all data directly to API.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
jkschneider: I don't see this as a breaking change. It's likely that a Micrometer `DistributionSummary` can result in _both_ a Wavefront histogram and some other summary statistics based on whether or not `percentileHistograms`/`sla` are configured. This doesn't need to be an either/or proposition. See our treatment of Prometheus timers for example. Prometheus timers don't support the concept of max, so Micrometer ships both a full Prometheus timer and also a separate max gauge.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontHistogram.java
jkschneider: We cannot accept vendor-specific meter types. It breaks the portability of the instrumentation.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
jkschneider: This is not OK, leads to vendor-specific configuration of tags that should be common tags for portability.

jkschneider: The furthest we've taken this in other implementations is to define which tag keys, if present, are mapped to some vendor-defined specific tag. See `hostTag` on `DatadogConfig`, for example. We still don't define the value of host in `DatadogConfig`, nor does the implementation assume that the tag key will be present on all meters necessarily.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontMeterRegistry.java
jkschneider: As we discussed offline, we should converge the concepts of `WavefrontSender` and `HttpSender`. That probably means a change to the wavefront SDK. Otherwise, we need to ship directly to API without the SDK until such a time as the SDK supports sender injection. Our users shouldn't be required to choose between sender customization and histogram support.

hanwavefront:
@shakuzen I've updated the PR in response to your comments.  Updated the summary and timer implementations to record and ship histograms per the config.  Please review, thanks.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontTimer.java
```diff
@@ -0,0 +1,62 @@
...
+    protected void recordNonNegative(long amount, TimeUnit unit) {
+        super.recordNonNegative(amount, unit);
+        if (delegate != null) {
+            delegate.update(TimeUtils.convert(amount, unit, baseTimeUnit()));
+        }
+    }
+
+    List<WavefrontHistogramImpl.Distribution> flushDistributions() {
+        if (delegate == null) {
+            return new ArrayList<>();
```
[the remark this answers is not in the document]
hanwavefront: Fixed.

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontDistributionSummary.java
```diff
@@ -0,0 +1,59 @@
...
+    protected void recordNonNegative(double amount) {
+        super.recordNonNegative(amount);
+        if (delegate != null) {
+            delegate.update(amount);
+        }
+    }
+
+    List<WavefrontHistogramImpl.Distribution> flushDistributions() {
+        if (delegate == null) {
+            return new ArrayList<>();
```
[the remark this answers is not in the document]
hanwavefront: Fixed.

implementations/micrometer-registry-wavefront/build.gradle
```diff
@@ -1,6 +1,7 @@
 dependencies {
     compile project(':micrometer-core')
     compile 'org.slf4j:slf4j-api:1.7.+'
+    compile 'com.wavefront:wavefront-sdk-java:1.2'
```
jkschneider: Let's make this `latest.release` so as we move locks we pick the latest versions.

jkschneider:
Merged with e3326fe6. Thanks for all the patience @hanwavefront :)

shakuzen:
@hanwavefront thank you for this. Would you be willing to send a pull request to update the documentation also? https://github.com/micrometer-metrics/micrometer-docs/

hanwavefront:
Thank you for reviewing. @shakuzen I'll open a PR for the documentation.

hanwavefront:
@shakuzen Opened a PR for documentation: https://github.com/micrometer-metrics/micrometer-docs/pull/86

Issue #900, opened 2018-10-05
Align Jersey tags with Web MVC


wilkinsona:
There are a couple of areas where it would be good to align Jersey tags with those for Web MVC:

1. Provide a `JerseyTags` class akin to `WebMvcTags` and change `DefaultJerseyTagsProvider` to be a thin wrapper around it.
2. Add an `outcome` tag as was added for Web MVC in https://github.com/micrometer-metrics/micrometer/pull/881.

izeye:
@wilkinsona Thanks for the suggestion! I created #907 to resolve this.

PR #907, opened 2018-10-06
Align Jersey tags with Web MVC

izeye:
This PR changes to align Jersey tags with Web MVC.

I tried to align it with Web MVC as possible as I can while I'm trying to keep the original semantics but I'm not a Jersey user, so there's chance to misinterpret. It'd be good to be reviewed by the original author, @mweirauch .

Closes gh-900

### DIFF
--- a/micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/JerseyTags.java
+++ b/micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/JerseyTags.java
@@ -169,19 +169,25 @@ public final class JerseyTags {
     public static Tag outcome(ContainerResponse response) {
         if (response != null) {
             int status = response.getStatus();
-            if (status < 200) {
-                return OUTCOME_INFORMATIONAL;
+            switch (Response.Status.Family.familyOf(status)) {
+                case INFORMATIONAL:
+                    return OUTCOME_INFORMATIONAL;
+
+                case SUCCESSFUL:
+                    return OUTCOME_SUCCESS;
+
+                case REDIRECTION:
+                    return OUTCOME_REDIRECTION;
+
+                case CLIENT_ERROR:
+                    return OUTCOME_CLIENT_ERROR;
+
+                case SERVER_ERROR:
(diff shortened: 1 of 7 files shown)


micrometer-jersey2/src/test/java/io/micrometer/jersey2/server/MetricsRequestEventListenerTest.java
```diff
@@ -134,25 +137,23 @@ public void exceptionsAreMappedCorrectly() {
...
             .isEqualTo(1);
     }
 
-    private static Iterable<Tag> tagsFrom(String uri, int status, String exception) {
-        return Tags.of(DefaultJerseyTagsProvider.TAG_METHOD, "GET",
-            DefaultJerseyTagsProvider.TAG_URI, uri, DefaultJerseyTagsProvider.TAG_STATUS,
-            String.valueOf(status), DefaultJerseyTagsProvider.TAG_EXCEPTION,
-            exception == null ? "None" : exception);
+    private static Iterable<Tag> tagsFrom(String uri, String status, String exception) {
+        return Tags.of("method", "GET", "uri", uri, "status", String.valueOf(status),
```
mweirauch: `String.valueOf()` is now superfluous.

micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/JerseyTags.java
```diff
@@ -0,0 +1,189 @@
...
+                : exception.getClass().getName());
+    }
+
+    /**
+     * Creates an {@code outcome} tag based on the status of the given {@code response}.
+     * @param response the container response
+     * @return the outcome tag derived from the status of the response
+     */
+    public static Tag outcome(ContainerResponse response) {
+        if (response != null) {
```
mweirauch: Only want to mention: This *could* be solved by switch-casing over `Response.Status.Family.familyOf()` which also provides `OTHER` for out-of-range code.

izeye:
@mweirauch Thanks for the feedback! I added c7504c8 as suggested.

PR #908, opened 2018-10-08
Add tests for match() and consume() in Meter.Type

izeye:
I saw [the fix](https://github.com/micrometer-metrics/micrometer/commit/80978cebedfb33fec3c37262cb05d5a85fd40aa8) from @jkschneider (Thanks!) for [the bug which I have introduced](https://github.com/micrometer-metrics/micrometer/commit/1399b07d78d5a599bb945ec7332dd791bf7cc612) carelessly. I'd be good to have tests for it and the similar method, `consume()` to prevent it from sneaking into them again somehow.


jkschneider:
Many thanks!

Issue #542, opened 2018-04-06
Add marker `TagsProvider` interface to Spring Boot


jkschneider:
Just a marker superinterface of both `RestTemplateExchangeTagsProvider`, `WebMvcTagsProvider`, and `WebFluxTagsProvider`. Functionally this does nothing, but eases the discovery of tags providers on the classpath.

cc / @twicksell

izeye:
@jkschneider I created #909 to resolve this.

PR #909, opened 2018-10-08
Add TagsProvider as a marker interface

izeye:
This PR adds `TagsProvider` as a marker interface and changes tags provider interfaces to extend it.

Closes gh-542


jkschneider:
@twicksell IIRC you had asked for this at one point? Is this satisfactory?

wilkinsona:
Given the discussion on 
https://github.com/spring-projects/spring-boot/pull/14824, I think it would be good to revert this change so that a `TagsProvider` interface that defines a proper contract can be introduced in the future without having to make breaking changes to the marker interface that was added here.

Issue #543, opened 2018-04-09
Can we get how many thread in waiting state


vinkdong:
Hi:
I have monitoring thread number of living, peak, and daemon, Now I need to monitoring JVM threads state, such as waiting, timed_waiting or running, is there a simple way to add it?

izeye:
@vinkdong Thanks for the suggestion! I created #910 to resolve this.

Issue #807, opened 2018-08-26
Multi-gauge: create multiple metrics from a single function


ochezeau:
Hi,
I tried to create a list of gauge from the result of a query and i didn't find any way to do that.

I have the following query which return the size of elements in database (postgres) :

SELECT N.nspname || '.' || C.relname AS "relation",
	    CASE WHEN reltype = 0
	        THEN pg_total_relation_size(C.oid)
	        ELSE pg_total_relation_size(C.oid)
	    END AS "size"
	FROM pg_class C
	LEFT JOIN pg_namespace N ON  (N.oid = C.relnamespace)
	LEFT JOIN pg_tables T ON (T.tablename = C.relname)
	LEFT JOIN pg_indexes I ON (I.indexname = C.relname)
	WHERE nspname NOT IN ('pg_catalog','pg_toast','information_schema');

My idea was to create a list of gauge with : id : 'db.relation.size' with tag 'relation' and value 'size'.

So that why i wanted to create a single function that perform the sql call and that allow me to create a list of gauge.

Maybe I'm wrong and that's not the right way of doing that or maybe this is contrary to the philosophy of Micrometer.

Thanks in advance of your help.

jkschneider:
This has been frequently requested, and depends on us first being able to expire metrics that go unused. It probably also requires a strong-reference gauge.

kristwaa:
> This has been frequently requested, and depends on us first being able to expire metrics that go unused.

I've tried a very simple version of this, where the observed value "expires" after being read (I made a `DecayingInteger`-class extending `AtomicInteger`). If no new value was provided until the next read, it would return a default value (typically zero in my case). This was also influenced by the lack of being able to remove a meter from the registry. The expiry / decay logic could be more complex if required.

My use-case was counts of _something_ in the form of a `SELECT ... GROUP BY something` query, where you wouldn't get a count of zero if there were no instances of a given value of _something_.

jkschneider:
@kristwaa I think you discovered the implicit dependency of this issue, which is meter removal (which in turn leads to expiration). I'm putting the finishing touches on removal right now. Once done, the `MultiGauge` will I believe be a very similar interface to `TimeGauge`, i.e. one that doesn't need specific implementations for each registry implementation.

ochezeau:
Thank you for the new feature and for the fantastic job done on Micrometer.

checketts:
The solution I needed in the past (previous employer) was either:

[...]

PR #910, opened 2018-10-08
Add thread state metrics

izeye:
This PR adds thread state metrics.

Note that this could be enhanced if there's a way to reuse a result across gauges but I couldn't find an easy way yet. #807 might be helpful here.

Closes gh-543

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
@@ -18,6 +18,7 @@ package io.micrometer.core.instrument.binder.jvm;
 import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.binder.MeterBinder;
 import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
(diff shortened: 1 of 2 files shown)


mweirauch:
I never got around looking into this, but would have thought of 'jvm.thread.states' with 'id' tag for the state. So you'd get all states (arguably the remaining are not that interesting) and this would differentiate this metric from the 'jvm.threads.*' ones. But just thinking loud. Thanks for looking into this!

izeye:
@mweirauch Thanks for the feedback! I just tried to translate the request from the referenced issue into code and didn't think much on this. I'd like to understand your suggestion but I couldn't get it fully. I might miss some context there.

I guess you're suggesting metric per thread and the meter definition is as follows:

- Meter name: `jvm.thread.states`
- Tag: `id` for thread ID

What I didn't get fully is how the state will be represented there. Translating it into numeric value or having another tag like `state`?

mweirauch:
What I meant was actually encoding the state as 'id' tag. I didn't consider the thread id. But what about 'id' and 'state' tags? 

Or does this generate too much cardinality given the fact that every thread id can be in every state?

izeye:
> What I meant was actually encoding the state as 'id' tag. I didn't consider the thread id.

@mweirauch Ah, I'm just confused by the name, `id` with thread IDs. I have no opinion on the meter names and interpreting states into tags. @jkschneider might have an opinion on this.

> But what about 'id' and 'state' tags?

Again, I'm just confused by the name, `id` with thread IDs, so I'm not sure it's worth to be a tag. And TBH I don't know how it could be achieved with `MeterBinder` for now.

mweirauch:
I agree thread id or name might be too much, also regarding the retrieval. And yes regarding 'jvm.threads.live', would be the same just sliced up.

So a distinct metric just with all states available expressed with a tag would do is my understanding from the discussion.

izeye:
@shakuzen @mweirauch Thanks for the feedback! Having `state` tag for thread state value sounds reasonable. I added 1d5ff43 as suggested.

mweirauch:
I know nitpicking, but I think all 'Thread.State' should be part of the metric. So 'NEW' and 'TERMINATED'. Otherwise summing over all states doesn't give the whole set of threads.

izeye:
@mweirauch I just assumed `ThreadMXBean.getAllThreadIds()` won't include threads having `NEW` state based on its Javadoc:

> Returns all live thread IDs. Some threads included in the returned array may have been terminated when this method returns.

But I didn't check more than the Javadoc, so I might be wrong. And I thought `TERMINATED` state is less interesting but it might be subjective.

I just added 8061ccb as suggested.

Issue #911, opened 2018-10-08
Rework Jetty ThreadPool metrics to work with any ThreadPool


wilkinsona:
This has been discussed on #593, but I don't want it to get lost in the push to RC1 as it's important for Boot 2.1 as well.

The short of it is that I think the class that binds the metrics for Jetty's thread pool should be reworked so that it doesn't rely on being able to customise the `ThreadPool` of the `Server`. Such customisation may overwrite something that a user's already done.

I think the `MeterBinder` should look something like this:

```java
public class JettyThreadPoolMetrics implements MeterBinder {

	private final ThreadPool threadPool;

	private final Iterable<Tag> tags;

	public JettyThreadPoolMetrics(ThreadPool threadPool, Iterable<Tag> tags) {
		this.threadPool = threadPool;
		this.tags = tags;
	}

	@Override
	public void bindTo(MeterRegistry registry) {
		if (this.threadPool instanceof SizedThreadPool) {
			SizedThreadPool sizedThreadPool = (SizedThreadPool) this.threadPool;
			Gauge.builder("jetty.threads.config.min", sizedThreadPool,
					SizedThreadPool::getMinThreads)
					.description("The minimum number of threads in the pool")
					.tags(this.tags).register(registry);
			Gauge.builder("jetty.threads.config.max", sizedThreadPool,
					SizedThreadPool::getMaxThreads)
					.description("The maximum number of threads in the pool")
					.tags(this.tags).register(registry);
			if (this.threadPool instanceof QueuedThreadPool) {
				QueuedThreadPool queuedThreadPool = (QueuedThreadPool) this.threadPool;
				Gauge.builder("jetty.threads.busy", queuedThreadPool,
						QueuedThreadPool::getBusyThreads)
						.description("The number of busy threads in the pool")
						.tags(this.tags).register(registry);
			}
		}
		Gauge.builder("jetty.threads.current", this.threadPool,
				ThreadPool::getThreads)
				.description("The total number of threads in the pool")
				.tags(this.tags).register(registry);
		Gauge.builder("jetty.threads.idle", this.threadPool,
				ThreadPool::getIdleThreads)
				.description("The number of idle threads in the pool").tags(this.tags)
				.register(registry);
	}

}
```

Getting hold of the `Server` needed to create `JettyThreadPoolMetrics` is a pain and is something we should improve in Boot. It can be done like this with Boot 2:

```java
@ConditionalOnBean(MeterRegistry.class)
[...]

PR #914, opened 2018-10-11
Rework Jetty ThreadPool metrics to work with any ThreadPool

izeye:
This PR updates `JettyServerThreadPoolMetrics` to work with any Jetty `ThreadPool`.

Closes gh-911


micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/web/jetty/JettyMetricsAutoConfiguration.java
```diff
@@ -46,11 +53,26 @@
         "javax.servlet.Servlet", "org.eclipse.jetty.server.Server", "org.eclipse.jetty.util.Loader",
         "org.eclipse.jetty.webapp.WebAppContext" })
 @Conditional(JettyMetricsAutoConfiguration.JettyMetricsAutoConfigurationConditionalOnBeans.class)
+@ConditionalOnBean(AnnotationConfigEmbeddedWebApplicationContext.class)
```
wilkinsona: Sometimes I am surprised by what's available as a bean. This is one of those times. Hopefully we can come up with a more robust way of dealing with this, but I'm not yet sure what it should be. A custom condition might make the intent clearer if nothing else.

wilkinsona:
I've opened https://github.com/spring-projects/spring-boot/issues/14784.

Another, somewhat related, thought just occurred to me about the difference between Tomcat's metrics and Jetty's metrics. Tomcat's metrics are heavily based on JMX. This, I believe, means that they'll work for applications deploy to a Tomcat container and applications that are using embedded Tomcat. I wonder if a similar approach was considered for Jetty and, if it was, why it was rejected.

micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/web/jetty/JettyMetricsAutoConfiguration.java
izeye: @wilkinsona Thanks for the feedback! I changed to use a custom condition to check whether it's an embedded web application in a32aa8a.

jkschneider:
@wilkinsona I can't say there was much deliberate thought here. I tend to want to get things from a programmatic interface if possible, resorting to JMX when there is no other option.

Issue #915, opened 2018-10-11
failed to send metrics to elastic: io.micrometer.core.ipc.http.HttpResponse@689e5ba


andrsuh:
Could you provide more meaningful message for error?

izeye:
@andrsuh Thanks for the report! I created https://github.com/micrometer-metrics/micrometer/pull/916 to resolve this.

PR #916, opened 2018-10-11
Log response body when invoking onSuccess() with error

izeye:
This changes to log response body when invoking `onSuccess()` with error as `HttpResponse` doesn't provide meaningful information with `toString()` and we already log response body in `onError()`.

Closes gh-915


jkschneider:
Woops, thanks!

PR #917, opened 2018-10-11
Make TomcatMetricsConfiguration auto-configuration

izeye:
This PR changes `TomcatMetricsConfiguration` to auto-configuration and renames it accordingly as there's no way to prevent from creating `TomcatMetrics` bean and it's aligned with [Spring Boot 2.x support](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/tomcat/TomcatMetricsAutoConfiguration.java).


izeye:
Note that I've targeted this to 1.1.0 although it's been there since Spring Boot 2.0.0 as no one has complained this so far and 1.1.0's release seems imminent based on the due date for its milestone. Maybe my other 1.0.x-targeted non-bug-fixing Spring Boot legacy PR(s) need(s) to be retargeted based on this thought.

jkschneider:
> Maybe my other 1.0.x-targeted non-bug-fixing Spring Boot legacy PR(s) need(s) to be retargeted based on this thought.

I'm a fan of keeping both the 1.0.x line and the 1.1.x line as behaviorally close to Spring Boot 2.0.x and 2.1.x respectively, so there is less surprise if folks upgrade suddenly from Boot 1.x to 2.x. That said, I suppose anybody upgrading in the near future from 1.x to 2.x would skip the 2.0.x line anyway.

Issue #918, opened 2018-10-12
Backport: MeterFilter and MeterRegistryCustomizer beans are not ordered


shakuzen:
Backport the fix from Spring Boot 2 to micrometer-spring-legacy for https://github.com/spring-projects/spring-boot/issues/14780

izeye:
@shakuzen I created #919 to resolve this.

shakuzen:
Done via #919. Thanks, Johnny!

PR #919, opened 2018-10-12
Order MeterFilters and MeterRegistryCustomizers

izeye:
This PR changes to order `MeterFilters` and `MeterRegistryCustomizers` by porting the changes from https://github.com/spring-projects/spring-boot/issues/14780.

For easy review, the previous `MetricsAutoConfigurationTest` has been renamed to `MetricsAutoConfigurationIntegrationTest` as a new `MetricsAutoConfigurationTest` has been added as unit tests. It seems that Git fails to be aware of the rename although there's no change in the previous one.

Closes gh-918


shakuzen:
@izeye thanks for the quick turnaround. See also Andy's follow-up polish commit https://github.com/spring-projects/spring-boot/commit/5e3b5ae14fed1cef6bc00a6bf62c8feffc2e1ffd which we might want to apply in Micrometer also to keep things in sync.

izeye:
@shakuzen Thanks for the feedback! I noticed the commit but AFAICT [it could be applied only to Spring framework 5.1 or later](https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/java/org/springframework/beans/factory/ObjectProvider.java#L163-L177). Am I missing something?

shakuzen:
@izeye you are, of course, correct and not missing anything. I should have looked more closely as you did :) Nevermind my comment then.

PR #921, opened 2018-10-13
Make JerseyServerMetricsConfiguration auto-configuration

izeye:
This PR changes `JerseyServerMetricsConfiguration` to auto-configuration and renames it accordingly.

This PR also aligns it with Spring Boot 2.x support as much as possible but I didn't sync the `jerseyMetricsUriTagFilter()` bean definition intentionally as it'd be better to be a separate PR alongside a counterpart for Web MVC IMO.


izeye:
> it'd be better to be a separate PR alongside a counterpart for Web MVC

I forgot the counterpart for Web MVC is already in the 1.0.x branch. Due to the common code from it, it'd be better to add the `jerseyMetricsUriTagFilter()` bean definition after it's merged into the master branch.

PR #923, opened 2018-10-14
Fix spring.aop.auto property reference

izeye:
This PR fixes the `spring.aop.auto` property reference as it seems incorrect based on [the Spring Boot reference](https://docs.spring.io/spring-boot/docs/1.5.x/reference/htmlsingle/#common-application-properties) as follows:

```
spring.aop.auto=true # Add @EnableAspectJAutoProxy.
```

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/MetricsAutoConfiguration.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/MetricsAutoConfiguration.java
@@ -124,7 +124,7 @@ public class MetricsAutoConfiguration {
 
     @Configuration
     @ConditionalOnClass(name = "org.aspectj.lang.ProceedingJoinPoint")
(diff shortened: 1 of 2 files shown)


jkschneider:
I can't explain why the test works currently.

izeye:
@jkschneider Which test did you mean?

jkschneider:
The test that you modified (`backsOffWhenSpringAopEnabledIsFalse`) passes when `spring.aop.enabled=false` and fails when this prop is removed.

jkschneider:
> I confirmed again that it failed with the wrong property

How could this be? The test has existed for some time with `spring.aop.enabled`. I, too, cannot find the property documented anywhere. This is all very strange.

izeye:
@jkschneider I guess you're saying the current 1.0.x branch and the reason why it works is [the current code is relying on the wrong property](https://github.com/micrometer-metrics/micrometer/blob/1.0.x/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/MetricsAutoConfiguration.java#L127). Am I missing something?

jkschneider:
Ah of course, now I understand.

PR #927, opened 2018-10-14
Fix Jersey status and outcome tags on unmapped exceptions

mweirauch:
In case unmapped exceptions are thrown from resources there
is no response object.

This one slipped in with #907.

//cc @izeye (Thank you very much for taking care of this! My review on a mobile screen missed this one.)


jkschneider:
Technically the server container has a chance to catch exceptions and cause any status to be returned, but consistent with what we've done with `WebMvc` and given that the vast majority of the time this will result in a 500, I think this behavior is reasonable.

mweirauch:
A bit late though, nevertheless thanks for sharing your input and reasoning on the matter! (I am just a Jersey user and never stop learning by working with you guys!)

PR #939, opened 2018-10-16
Add a utility for JSON String builder

izeye:
This PR adds a utility for JSON `String` builder and also applies it to `AppOpticsMeterRegistry` as an example.

I was writing a generalized version for this but it seems to make things complicated more than necessary, so I just focused only on `AppOpticsMeterRegistry` for brevity by leaving out other primitive types, escaping, and formatting. We can augment this when necessary.


jkschneider:
I suppose we wind up creating more short term garbage this way?

izeye:
@jkschneider If you meant to reduce intermediate `String`s, yes, I tried to do so.

jkschneider:
I mean we are building up map objects now rather than just writing directly to a `StringBuilder`.

izeye:
@jkschneider Ah, yes. I thought this is a usual approach to create a JSON and as I said we can reduce intermediate `String` object creation with the cost. Do you want to propose the other way?

shakuzen:
Thanks for the PR. We'll keep things the way they are now to avoid creating more garbage when publishing metrics.

PR #947, opened 2018-10-18
Making sure TimedAspect does not disrupt the application flow when there is an exception recording a metric

amanzag:
I found that sometimes the monitoring backend can throw an Exception when recording a metric and the TimedAspect then breaks the application flow.
This happens, for instance, with prometheus when the metric has the wrong number of tags.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -78,14 +78,18 @@ public class TimedAspect {
             exceptionClass = ex.getClass().getSimpleName();
             throw ex;
         } finally {
-            sample.stop(Timer.builder(metricName)
-                    .description(timed.description().isEmpty() ? null : timed.description())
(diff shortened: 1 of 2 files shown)


jkschneider:
@amanzag In theory, registries should never throw exceptions like this. Can you let us know what exception you are seeing?

amanzag:
Sorry, I don't have a stack trace right now, but it's easy enough to reproduce with prometheus.
Just create 2 methods like this:
```java
@Timed(value="metric_name", extraTags = { "tag1", "x" })
public void method1() {}

@Timed(value="metric_name", extraTags = { "tag1", "x", "tag2", "y" })
public void method2() {}
```
If you call method1 and then method2, prometheus throws an exception indicating that a metric should always have the same tags (which is right). Whether that exception should bubble up or be caught somewhere by micrometer, I don't know, but regardless, I think catching any exceptions in the TimedAspect won't hurt.

jkschneider:
Polished and merged with 3e437adc.

Issue #948, opened 2018-10-18
InfluxDB returns "Unable to parse" after upgrading


lukas-krecan-lt:
After upgrading from Spring Boot 2.1.0.M4 to RC1 (micrometer 1.1.0.m1 to 1.1.0.rc1) Influx metric registry started to return following error message:

[...]

PR #950, opened 2018-10-19
Add Content-Encoding header for gzip

izeye:
This PR adds `Content-Encoding` header for gzip.

Fixes gh-948


jkschneider:
Ah, thanks @izeye!

PR #968, opened 2018-10-25
Remove Object creation in StrongReferenceGaugeFunction

izeye:
This PR removes `Object` creation in `StrongReferenceGaugeFunction` as it doesn't seem to be necessary unless I'm missing something.


jkschneider:
Merged with 6b02febe.

jkschneider:
I had a crazy notion initially that the Gauge could receive a null object and somehow still have a strong reference I think. Just a bad idea. Thanks for pointing this out.

PR #973, opened 2018-10-27
Updating Sysdig line builder to sanitize replace all non-alphanumeric characters with an underscore

brmakana:
According to Sysdig they only allow the following for metrics:

> Allowed characters: a-z A-Z 0-9 _ .

Replace everything else with an underscore


jkschneider:
Merged and polished with 4f2d6b39

jkschneider:
Thanks for taking this on @brmakana!

Issue #979, opened 2018-10-30
PostgreSQLDatabaseMetrics#runQuery does not advance JDBC cursor to next row


michaelbull:
The [`runQuery`](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java#L282) method in `PostgreSQLDatabaseMetrics` does not call `resultSet.next()`.

The [`ResultSet`](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html) documentation states:

>Initially the cursor is positioned before the first row. The `next` method moves the cursor to the next row

As `next()` is not called, the `runQuery` method always fails with the following exception:

>ResultSet not positioned properly, perhaps you need to call next.

![](https://i.imgur.com/wCpVfP0.png)

As a result, none of the intended metrics are recorded.

izeye:
Thanks for the report! It seems an oversight. I created #980 to fix this.

PR #980, opened 2018-10-31
Invoke missing ResultSet.next() in PostgreSQLDatabaseMetrics.runQuery()

izeye:
This PR changes to invoke missing `ResultSet.next()` in `PostgreSQLDatabaseMetrics.runQuery()`.

Fixes gh-979

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
@@ -292,7 +292,7 @@ public class PostgreSQLDatabaseMetrics implements MeterBinder {
                 return 0L;
             }
         } catch (SQLException e) {
-            logger.error("Error getting statistic from PostgreSQL database.");
+            logger.error("Error getting statistic from PostgreSQL database.", e);
             return 0L;
         }
     }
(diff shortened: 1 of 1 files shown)


izeye:
I also changed to log `SQLException` in 5aa37a3.

jkschneider:
I wish I had noticed `Logger` before. We need to strike that field altogether, as binders shouldn't assume the presence of a particular logging framework.

izeye:
@jkschneider I'm not sure I understand your comment. Do you suggest removing all logs from meter binders? Or using Apache Commons Logging as Spring Boot does?

jkschneider:
Removing all logs from MeterBinders. Spring Boot may have chosen Apache Commons Logging, but we cannot presume that all other frameworks have made the same choice.

jkschneider:
Actually, I should soften that language a bit. Remove logs from MeterBinders that are observing something that doesn't already have an explicit logging dependency. It's OK in hystrix for example because of:

image

Presumably excluding slf4j would break hystrix in other ways so we can assume it is present.

jkschneider:
Merged with da4aeefc

izeye:
@jkschneider Thanks for the clarification! Based on your change in da4aeef, losing some details from logs is okay, right?

izeye:
@jkschneider Meter registries also uses SLF4J. Are they okay unlike meter binders?

jkschneider:
> Based on your change in da4aeef, losing some details from logs is okay, right?

I think it has to be, unfortunately.

jkschneider:
> Meter registries also uses SLF4J.

Maybe it's worth thinking about this more. 

Registry implementations don't all perform logging. Some do, and I've thought in the past that it is OK because you can always create your own implementation if you don't like the way something works in a particular one. Core though I think should be even less opinionated than any given registry.

In practice, the idea of asking someone to write a custom datadog registry, for example, because they don't like the SLF4J dependency is a bit of an extreme thing to ask.

jkschneider:
@izeye Reactor goes through some hoops to conditionally log in `Flux#log()`, but I haven't looked yet at what they do. It's such a shame to require a wrapper around a wrapper around a wrapper, but I don't see a good alternative.

PR #985, opened 2018-11-02
Remove logging of metrics line on sout

fauscik:


izeye:
Changing the milestone to 1.0.8 as this also needs to be fixed on the 1.0.x branch.

See https://github.com/micrometer-metrics/micrometer/blob/1.0.x/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilder.java#L56

jkschneider:
Merged with 30f6f48b. Thanks @fauscik!

PR #986, opened 2018-11-02
Remove HttpSender.print() usages

izeye:
This PR removes `HttpSender.print()` usages as it doesn't seem to be intentional.


jkschneider:
Merged with 34f79c8b.

PR #988, opened 2018-11-03
Added property to control whether the date is to be appended to the elastic search index name

### LINKED ISSUE #987: ElasticMeterRegistry always adds a date format to the index
The current implementation assumes that the index to be used will be composed of the specified index name plus a date format separated by a hyphen.

From `ElasticMeterRegistry`

`return config.index() + "-" + DateTimeFormatter.ofPattern(config.indexDateFormat()).format(dt)`

In the environment where I am looking to deploy mm, in conjunction with spring boot & elastic, indexes are pre-provisioned such that we only get 1 index for the duration of the index and as such the idea of date rolling the index would not work.

Currently setting the `management.metrics.export.elastic.indexDateFormat=` removes the date portion of the index, but the `-` is still appended meaning that we could not currently match our provided index name.

As a potential fix suggest creating a new property `indexDate` with a default of `true` which will continue with the above behavior, but when set to `false` will not append anything to the provide index name.

If you would like, I will try to raise a PR to fix this over the next couple of days. If you would prefer to fix it yourself within the existing project, please let me know and I will simply leave it with you.

smitp33:
To fix #987

### DIFF
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java
@@ -120,4 +120,15 @@ public interface ElasticConfig extends StepRegistryConfig {
         String v = get(prefix() + ".password");
         return v == null ? "" : v;
     }
+
+    /**
+     * Whether to add the indexDate suffix on the index name.
+     * Default is: {@code true}
+     *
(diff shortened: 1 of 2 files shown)


smitp33:
Could not find any contributing guide so I have following the o

As this is relatively simple change, and due to the fact that the methods affected are not easily exposed I have not currently added any tests.

Assume there needs to be an additional PR to the spring actuator to actually add this new property as part of the auto-config?

PR #1004, opened 2018-11-14
Add Proxy option to HttpUrlConnectionSender

ecornely:
Since agent 6 cannot be a proxy anymore I thought it would be good to be able to use a proxy directly in the HttpSender.
In our company we are forced to use proxies so I had to switch to the statsd approach.

Later we could even think of proxy authentication if you like my proposal

### DIFF
--- a/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
+++ b/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
@@ -73,4 +73,22 @@ public interface DatadogConfig extends StepRegistryConfig {
         String v = get(prefix() + ".descriptions");
         return v == null || Boolean.valueOf(v);
     }
+
+    /**
+     * @return The http proxy used to publish metrics to datadog, or {@code null} if direct connection can 
+     * be used.
+     */
+    @Nullable
+    default String proxyUrl() {
+        return get(prefix() + ".proxyUrl");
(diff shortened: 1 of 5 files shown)


jkschneider:
I'm on board with the idea of adding proxy capability to `HttpUrlConnectionSender`, but not as enthusiastic about the change to `DatadogMeterRegistry`. We're trying to steer folks to inject a custom `HttpSender` into push-based registries when they need more advanced features like this. Otherwise, we'd have to duplicate this proxy configuration option across all push-based registries.

So with your changes to `HttpUrlConnectionSender` you could use a proxy by instantiating `DatadogMeterRegistry` like so:

```
DatadogMeterRegistry
   .builder(datadogConfig)
   .httpClient(new HttpUrlConnectionSender(Duration.ofSeconds(1), Duration.ofSeconds(1), proxy))
   .build()
```

What do you think?

ecornely:
Looks nice as well...

I never used authentication with proxies. Do you think providing the proxy
object will be enough or it requires more optional arguments?

By the way I use now statsd instead of direct access to Datadog API in my
applications.

Kr,

Eric

Le jeu. 29 nov. 2018 16:54, Jon Schneider <notifications@github.com> a
écrit :

jkschneider:
I've never used `java.net.Proxy`, so not sure.

jkschneider:
You really are better off with direct-to-API over statsd once this is merged. Creates less short-term garbage in your app.

PR #1005, opened 2018-11-14
Remove redundant 'System.out.println()' in TelegrafStatsdLineBuilder.

denzelby:
Removed redundant stdout when using Telegraf StatsD line builder.


denzelby:
Same PR as #985, just a bit prettier.

izeye:
@denzelby Thanks for the PR but as you've noticed, this is a duplicate of #985. Closing.

For the polishing part, you can create a PR after #985 is merged.

PR #1007, opened 2018-11-14
Replace spaces with dashes to fix error when sending messages to influx

aurisl:
Fixes error when sending JVM memory metrics to Influx. 
Influx does not support spaces for its tags. 
So the value `PS Eden Space` breaks it. With `Invalid field format unable to parse` error. 
I have replaced space with dash. 

https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_tutorial/#tag-set


jkschneider:
Currently `InfluxNamingConvention` escapes spaces before shipping to Influx. Testing it here, I don't see any problem. The escaped text gets to influx, and is neatly displayed in labels on charts based on Influx queries:

image

aurisl:
Thanks, yes you are right. We are using our own naming convention, my bad.

PR #1012, opened 2018-11-15
Fix a wrong index when copying an array in Tags.and(Tag...)

izeye:
This PR fixes a wrong index when copying an array in `Tags.and(Tag...)`.


jkschneider:
Merged with a5983d44

PR #1029, opened 2018-11-26
Fix that deny filters always take precedence

slomo:
I had some issues with writing MeterFilters. After accepting a Meter with my filter it is still filtered out by a proceeding `MeterFilter.deny`.  According to the [concepts documentation](https://micrometer.io/docs/concepts#_meter_filters):

> ACCEPT - If a filter returns ACCEPT, the meter is immediately registered without interrogating any further filters' accept methods.

So this PR tries to fix this and also provides a test, which I would expect to work (given the above section of the documentation), but it doesn't with the proposed changes.


jkschneider:
Merged with 5494daae.

PR #1030, opened 2018-11-26
When no event occurs don't sent aggregated metrics to cloudwatch

fsec-progacki:
Currently, CloudWatch registry sends metrics like avg or max even if no event happened. In such a case, values are set to 0 what makes those metrics aggregation broken. This PR fixes it by sending avg and max metrics only when at least 1 event occurred

Also added Mockito dependency. That forced me to update locks within micrometer-registry-cloudwatch module.


shakuzen:
@fsec-progacki I'm looking to review this for inclusion in the 1.2 release. Would you be able to rebase your changes on the latest `master`? Note we've remove all lockfiles from the `master` branch to avoid such conflicts on `master`.

krzsztf:
I've rebased this change in https://github.com/micrometer-metrics/micrometer/pull/1404

shakuzen:
Closing in favor of #1404

Issue #1038, opened 2018-11-28
[BUG] Counter object is not removed from Dropwizard registry


mrniko:
Follow code throws `java.lang.IllegalArgumentException: A metric named test already exists` exception. Meter is not removed from Dropwizard registry, only from micrometer registry. Reproducible with Dropwizard 4.0.3 and micrometer 1.1.0.

Code to reproduce the case:
```java
        DropwizardConfig consoleConfig = new DropwizardConfig() {
            @Override
            public String prefix() {
                return "console";
            }
            @Override
            public String get(String key) {
                return null;
            }
        };

        MetricRegistry reg = new MetricRegistry();
        DropwizardMeterRegistry registry = new DropwizardMeterRegistry(consoleConfig, reg, HierarchicalNameMapper.DEFAULT, Clock.SYSTEM) {
            @Override
            protected Double nullGaugeValue() {
                return null;
            }
        };
        
        Counter counter = registry.counter("test");
        registry.remove(counter);
        registry.counter("test");
```

Exception:

```java
java.lang.IllegalArgumentException: A metric named test already exists
	at com.codahale.metrics.MetricRegistry.register(MetricRegistry.java:97)
	at io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry.newCounter(DropwizardMeterRegistry.java:60)
	at io.micrometer.core.instrument.MeterRegistry.lambda$registerMeterIfNecessary$5(MeterRegistry.java:528)
	at io.micrometer.core.instrument.MeterRegistry.getOrCreateMeter(MeterRegistry.java:580)
	at io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:543)
	at io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:528)
	at io.micrometer.core.instrument.MeterRegistry.counter(MeterRegistry.java:245)
	at io.micrometer.core.instrument.Counter$Builder.register(Counter.java:128)
	at io.micrometer.core.instrument.MeterRegistry.counter(MeterRegistry.java:352)
	at io.micrometer.core.instrument.MeterRegistry.counter(MeterRegistry.java:363)
```

izeye:
@mrniko Thanks for the report and the test! I created #1040 to try to fix it.

PR #1040, opened 2018-11-28
Remove a metric from MetricRegistry when removing a meter

izeye:
This PR changes to remove a metric from `MetricRegistry` when removing a meter from `DropwizardMeterRegistry`.

Fixes gh-1038


jkschneider:
Fixed the test assertion (I think) and merged with 57d7a6ac.

Issue #1043, opened 2018-11-29
Statsd buffering does not flush until reaching max packet size in some cases


jonny-miller:
The statsd buffering that was recently introduced doesn't seem to actually write the UDP packets until it reaches the maximum size in situations where metrics are updated at a shorter interval than the polling interval.

I ran a simple test where I started incrementing a counter every second and ran tcpdump watching for packets, and no packets were sent out until the packet was near the threshold.  I tweaked the test to only increment 20 times and no packets were sent at all.  

My test snippet (in Kotlin):

```
        val config = object : StatsdConfig {
            override fun get(key: String) = null
            override fun flavor() = StatsdFlavor.TELEGRAF
            override fun step() = Duration.ofSeconds(1)
        }
        val registry = StatsdMeterRegistry(config, Clock.SYSTEM)
        val counter = registry.counter("counter")

        for (i in 0..20) {
            counter.increment()
            Thread.sleep(1000)
        }

        registry.stop()
```

and running tcp dump watching for packets showed none sent (`sudo tcpdump -i lo0 port 8125 -X`).
If I had a number closer to 50, it would eventually send out a packet when the size was in the 1300-1400 byte range.

I poked into the BufferingFlux and found it appears to be an issue with how the time diff is computed.  In BufferingFlux (https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/BufferingFlux.java#L64) it updates the `lastTime` regardless of whether the data is buffered or not, which means that if you have metrics firing faster than the `pollingInterval` (which is passed in as `maxMillisecondsBetweenEmits`) it always ends up determining that not enough time has passed.

PR #1044, opened 2018-11-29
Fix statsd buffering for metrics updates faster than pollingInterval

jonny-miller:
I found that the time computation for determining whether to buffer or not was not correct (#1043).  It was always computing the time since the last value, not since it was last flushed. If you had metrics that were triggered faster than the pollingInterval (which is the value used to determine how long to buffer events), they would not be written at the correct time and would only flush when reaching the maximum packet size.

I couldn't come up with a way to write this as a test using StepVerifier, I was only able to demonstrate the problem by directly subscribing to the BufferedFlux and toggling a boolean once a value was received.  Without my changes, this would show that no value is emitted after 500ms, where the values were written every 100ms and the BufferedFlux is configured to emit events after 200ms max.


jkschneider:
Thanks for the keen eye and fix @jonny-miller! I changed the test to use a `CountDownLatch` rather than the sleep, which I think does the trick.

jkschneider:
Merged with f491daec

Issue #1042, opened 2018-11-28
meter only root logger log4j2


z0mb1ek:
Hi. I see in Log4j2Metrics that you add only ROOT logger. Why? Can I meter my loggers too?

izeye:
@z0mb1ek Looking at the code, I can see the filter being added on the root logger, so I expect it will meter all loggers. Do you mean it meters only the root logger or you want to have per-logger meters?

z0mb1ek:
@izeye thx for reply. Yes, I tested it and it meters ONLY root logger :( I want to meter all my loggers together and my be it would be useful to configure what logger to meter.

izeye:
@z0mb1ek Looking at [its test](
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java#L53), it seems to work for other loggers. Could you provide a sample project to reproduce it?

z0mb1ek:
@izeye its about configuration of log4j2
For example:
```  Loggers:

      Root:
        level: info
        AppenderRef:
          - ref: Console_Appender

      Logger:
        - name: com.skipz
          additivity: false
          level: debug
          AppenderRef:
            - ref: Console_Appender
```
in this logs from com.skipz package will not meter.

izeye:
@z0mb1ek Thanks for sharing a sample configuration! The test uses [a logger named its own class name](
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java#L37). I’m not sure whether the `additivity` configuration could affect here somehow. I can see your logger levels are different, so could you confirm the application logged something at debug level with the non-root logger?

z0mb1ek:
@izeye yes. Or it can be kafka client with error level only for example, because it writes very much logs.

izeye:
@z0mb1ek I confirmed it's working with a sample: https://github.com/izeye/sample-micrometer-spring-boot/tree/log4j2

The only remaining difference I noticed is the `additivity` flag. I'll try to check this later but if you could provide a sample to reproduce this on GitHub, it'd be appreciated.

z0mb1ek:
@izeye can you show me log4j2 config in this project? Looking at the code this log to root too

izeye:
[...]

PR #1045, opened 2018-11-29
Meter additivity-disabled loggers for Log4j2

izeye:
This PR changes to meter additivity-disabled loggers for Log4j2.

Closes gh-1042


izeye:
This PR doesn't support additivity-disabled loggers created at runtime. I'm not sure there's a way to hook into logger creation points in Log4j2 but if it's possible, we could also support them.

jkschneider:
Merged with 48c9e60b.

PR #1046, opened 2018-11-29
Configurable StatsD buffering

jonny-miller:
This makes the buffering configurable.  Some might want buffering, while others may want packets to be sent out immediately.  I made this default to true since that is the current behaviour.


jkschneider:
Thanks @jonny-miller. We'll have to reserve this for 1.2.0 since it's additive.

shakuzen:
@jonny-miller thank you for the pull request. Would you be willing to rebase this on the latest changes in `master`?

Issue #1037, opened 2018-11-28
HibernateMetrics fails if Hibernate is not on the classpath


snicoll:
Already reported in #518 but I didn't properly follow-up.

The intention that is discussed there ([this code](https://github.com/micrometer-metrics/micrometer/blob/6cdc476985b2f6e57d0c2487dd0745e7d02aabd6/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java#L170-L185)) does not work if Hibernate is not on the classpath.

```
Caused by: java.lang.NoClassDefFoundError: org/hibernate/SessionFactory
    at io.micrometer.core.instrument.binder.jpa.HibernateMetrics.getStatistics(HibernateMetrics.java:180)
    at io.micrometer.core.instrument.binder.jpa.HibernateMetrics.hasStatisticsEnabled(HibernateMetrics.java:166)
    at io.micrometer.core.instrument.binder.jpa.HibernateMetrics.<init>(HibernateMetrics.java:57)
    at org.springframework.boot.actuate.autoconfigure.metrics.orm.jpa.HibernateMetricsAutoConfiguration.bindEntityManagerFactoryToRegistry(HibernateMetricsAutoConfiguration.java:71)
```

There was a faulty condition in Spring Boot that got fixed in https://github.com/spring-projects/spring-boot/issues/15317 but something should probably change here as well. 

To make it more obvious that a `SessionFactory` is required, perhaps the constructor should require that rather than an `EntityManagerFactory` that doesn't enforce anything related to Hibernate. Alternatively, `HibernateMetrics` could gracefully handle the fact that Hibernate is not on the classpath.

izeye:
Looking at the code, it seems to assume that Hibernate is on the classpath as it has [a final member field typed `Statistics`](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java#L45) which comes from Hibernate. As @snicoll suggested, exposing `SessionFactory` might be an option to be clear or updating Javadoc might be another option.

For https://github.com/spring-projects/spring-boot/issues/15317, I created https://github.com/micrometer-metrics/micrometer/pull/1039 to fix in the Spring Boot legacy support.

izeye:
[...]

PR #1053, opened 2018-12-04
Use SessionFactory for public static factory methods and constructor

izeye:
Closes gh-1037

/cc @snicoll

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
@@ -30,39 +30,112 @@ import java.util.function.ToDoubleFunction;
 
 /**
  * A {@link MeterBinder} implementation that provides Hibernate metrics. It exposes the
- * same stats as would be exposed when calling {@code Statistics#logSummary}.
+ * same statistics as would be exposed when calling {@link Statistics#logSummary()}.
  *
  * @author Marten Deinum
  * @author Jon Schneider
+ * @author Johnny Lim
  */
 @NonNullApi
 @NonNullFields
 public class HibernateMetrics implements MeterBinder {
 
+    private static final String SESSION_FACTORY_TAG_NAME = "entityManagerFactory";
+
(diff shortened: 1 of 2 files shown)


snicoll:
@izeye thanks for the poing. Have you considered wrapping the call to hibernate in a static class that we would wrap with a class presence check? IMO, if the input of this class is the JPA interface, we ought to be a bit more defensive about things.

izeye:
@snicoll Thanks for the feedback! I just chose this approach as its implementation seems to assume that Hibernate is on the classpath and Hibernate will be on the classpath in most cases, so I just thought clarifying it might be sufficient.

IIUC, the suggested approaches are as follows:

1. Make `HibernateMetrics` no-op when it's not on classpath.
2. Change parameter type `EntityManagerFactory` to `SessionFactory` on public static factory methods and constructor.

For the first approach, I'm not sure it's worthing allowing `HibernateMetrics` to be created when Hibernate isn't on the classpath as it will bring some code which doesn't seem to be neat.

For the second approach, I prefer this one but it's a breaking change, so I just hesitated to do so but it's a right way to go, I'm happy to update this PR to do so.

I might misunderstand the suggestions. Please correct me if I'm wrong.

snicoll:
Thanks for the feedback. I prefer the second one as well but that ship has sailed for the current line.

> as it will bring some code which doesn't seem to be neat.

I am not sure I got that. I agree that `HibernateMetrics` means exactly that it is supposed to work with Hibernate but the fact the current arrangement allows you to create an instance without an Hibernate-specific construct is a case for making it more lenient. The fact we have to unwrap to `SessionFactory` is an implementation detail after all...

izeye:
@snicoll Thanks for the quick feedback! "some code" meant code for hiding [the `stats` final member field typed `Statistics`](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java#L45) somehow to avoid `ClassNotFoundException`. I'll update this to the second approach.

izeye:
@snicoll I've updated this to the second approach. Let me know it's what you have in mind.

PR #1058, opened 2018-12-05
Exclude libnetty_transport_native_epoll_x86_64.so

izeye:
This PR excludes "libnetty_transport_native_epoll_x86_64.so" to restore the previous behavior.

See gh-1054


jkschneider:
Thanks @izeye! I suppose this is worth releasing 1.1.2 shortly.

Issue #569, opened 2018-04-20
Add 'clear' method to SimpleMeterRegistry to support testing


mkopylec:
There is no implementation of `MeterRegistry` suitable for intergration/e2e tests.
The `SimpleMeterRegistry` is almost OK, but it lacks of one major functionality.
Its `Meter`s cannot be reset and again filled with metrics.
The `close()` method makes the `SimpleMeterRegistry` useless when invoked.
Because the `SimpleMeterRegistry` is not reusable I need to start Spring Boot application every test method which is a simple waste of time.

checketts:
Can you add a little more details about what you need the `SimpleMeterRegistry` to do in an integration test? Are you trying to use it to time your tests?

Also what behavior do you see when calling `close()` that makes it unusable?

mkopylec:
In each test method I need to check that the `Timer` and `Counter` metrics were correctly recorded in the `MeterRegistry` by their name. Every test case collects identically named metrics but different number of times. When `Meter`s in the `MeterRegistry` are not cleared each test method than the `MeterRegistry.count()` method will return increased number each test method and the assertions will fail. There is no implemention of `MeterRegistry` that provides clearing metrics feature. I thought that maybe the `MeterRegistry.close()` clears metrics but after invoking the method the `MeterRegistry` is no longer accepting new metrics.

checketts:
It appears that a `clear()` method would be useful. We've discussed this and the problems for why other registries like the PrometheusMeterRegistry can't support it (yet), but I don't think that should block `SimpleRegistry` from supporting it.

jkschneider:
I think this is well covered by #479? Will close this in favor of the more general solution targeted for 1.1.0

checketts:
I'm fine merging the 2 tickets. #479 seems to be targeted at unregistering individual meters. It would be nice to have an `un-register all` so the user doesn't have to iterate over all meters, they just want a fresh set of measures.

jkschneider:
Agreed. Once we have the ability to "unregister individual" meters, clearing is trivial.

kurtharriger:
This is closed, but I don't yet see a clear() method on SimpleMeterRegistry am I missing something? Thanks

checketts:
@kurtharriger You are correct that `clear()` doesn't yet exist. The referenced tickets like #479 solved removing a single meter, clear would just be a convenience method around fetching all meters and removing each one.

izeye:
[...]

PR #1069, opened 2018-12-14
Add removeAllMeters() to MeterRegistry

izeye:
This PR adds `removeAllMeters()` to `MeterRegistry` to support clearing all meters.

See gh-569

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -53,6 +53,7 @@ import static java.util.Objects.requireNonNull;
  * thread, e.g. it should respond immediately by avoiding IO call, deep stack recursion or any coordination.
  *
  * @author Jon Schneider
+ * @author Johnny Lim
  */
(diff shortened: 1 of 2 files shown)


izeye:
@jkschneider Thanks for the feedback! I was considering `clear()` (the name suggested in the original issue) or `removeAll()` for its name first but it seemed ambiguous to me as there are another mutable parts like meter filters. If you still prefer `removeAll()`, just let me know. I'll update its name accordingly.

izeye:
@jkschneider I updated its since information to align with the target release in 939dff6d0cf304d0c39163326978ea7238e1fc08.

checketts:
Some things to keep in mind:

- `clear()` is used in collections, and since a registry is largely a collection of meters that is usually the first thing a user would look for. 
- `removeAll()` (or `removeAllMeters()`) aligns with `remove`, however if an implementation wanted to do additional resetting in that method then it would be mis-named.

I'm fine with both names, but wanted to call out those points.

jkschneider:
I'm OK with either `clear()` or `removeAll()`, with I suppose a leaning toward `clear()` for the reasons @checketts mentioned.

PR #1072, opened 2018-12-14
print dynatrace message content in debug logs

pjfanning:
I'm hitting issues where the messages are too large and being rejected by Dynatrace.
It's useful to be able to get the message contents so they can be analysed.

`{"code":413,"message":"Entity must not be larger than 15360 bytes"}`


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceTimeSeries.java
```diff
@@ -39,17 +39,24 @@ public String getMetricId() {
     }
 
     String asJson() {
-        String body = "{\"timeseriesId\":\"" + metricId + "\"" +
-                ",\"dataPoints\":[[" + time + "," + DoubleFormat.decimalOrWhole(value) + "]]";
```
pjfanning: I've seen issues in JDK 1.8 where the `DoubleFormat.decimalOrWhole(value)` is not being added as text but some unparseable value

pjfanning:
I published a jar with these changes - https://github.com/kontainers/micrometer-registry-dynatrace

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceTimeSeries.java
jkschneider: Can you elaborate on the conditions that yield unparseable values?

pjfanning: I haven't really got to the bottom of this issue yet. The StringBuilder approach does not make the issue go away.

The metric that this issue happens with has the date and time - a value like 201812311145.

`DoubleFormat.decimalOrWhole(value)` seems to handle this value ok for a few mins but the Dynatrace upload is done every 1 min (by default) and eventually the value starts appearing as soon non-numeric text and Dynatrace rejects the upload as unparseable.

Happens with OpenJDK 1.8.0_171.

pjfanning: The bad double seems to go away with newer JREs (OpenJDK 1.8.0_192 / OpenJDK 11.0.1). The main intent of this Pull Request is to add the logging support that allowed me to debug the double format issue.

pjfanning: @jkschneider the issue with the invalid value was that it was Double.NaN - I logged https://github.com/micrometer-metrics/micrometer/issues/1115

PR #1074, opened 2018-12-15
Reduce memory footprint of histograms

felixbarny:


felixbarny:
You're welcome 🙂 

Have you thought about whether it would be possible to replace the search for the key in `io.micrometer.core.instrument.distribution.TimeWindowFixedBoundaryHistogram.FixedBoundaryHistogram#leastLessThanOrEqualTo` with an algorithm which calculates the index of the value instead of searching for it? I guess that would reduce the complexity of recording a value from O(log(B)) to O(1).

I think the [memory footprint documentation](http://micrometer.io/docs/concepts#_memory_footprint_estimation) should also be adjusted now.

If I'm not mistaken it should now be 

`Fb = Fixed boundary histogram. 8b * B * R`

This, however, neglects the static overhead of `FixedBoundaryHistogram` and `AtomicLongArray` but that does not matter too much, especially with any real-world bucket sizes.

jkschneider:
> with an algorithm which calculates the index of the value instead of searching for it?

Given that buckets are not always describable by a function, I'm not sure how it could be done. Open to suggestions though.

felixbarny:
Yeah, coming up with the underlying function is the tricky part...

In HdrHistogram, they seem to be able to calculate the index instead of searching for it though: https://github.com/HdrHistogram/HdrHistogram/blob/1b51671cf06062671af931f98a8af2a53d335a70/src/main/java/org/HdrHistogram/AbstractHistogram.java#L2248

But the distribution of the histogram buckets works a bit differently there I think.

PR #1079, opened 2018-12-17
Remove Gradle Task Tree plugin

izeye:
This PR upgrades to Gradle Task Tree 1.3.1 which is compatible with Gradle 5.0.


jkschneider:
We can drop this plugin altogether.

izeye:
@jkschneider Thanks for the feedback! I updated this PR to remove the plugin.

PR #1081, opened 2018-12-17
Remove jar.deleteAllActions() usages

izeye:
This PR removes `jar.deleteAllActions()` which has been removed in Gradle 5.0. I failed to see the reason why it exists and a quick search on it didn't give much information. I might miss something here. Please let me know if it must be handled differently with some reason.

### DIFF
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -33,7 +33,7 @@ shadowJar {
     relocate 'org.pcollections', 'io.micrometer.shaded.statsd.org.pcollections'
 }
 
-jar.deleteAllActions()
+jar.enabled = false
 jar.dependsOn shadowJar
 
 publishing {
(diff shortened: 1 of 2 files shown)


izeye:
I think I understand the reason (making an artifact to be a shadow jar, not an original jar) now and it seems to be able to do the same trick with the following configuration:

```
jar.enabled = false
```

I updated this PR accordingly.

jkschneider:
If `jar.enabled=false` then the `jar` task doesn't run though right? Does anything that depends on `jar` still cause `shadowJar` to run even though `jar` is disabled?

izeye:
@jkschneider Yes. If I run `../gradlew clean pTML --console=plain` in `micrometer-core`, I see the following output:

```
...
> Task :micrometer-core:clean
> Task :micrometer-core:generatePomFileForNebulaPublication
> Task :micrometer-core:compileJava
> Task :micrometer-core:processResources NO-SOURCE
> Task :micrometer-core:classes
> Task :micrometer-core:writeManifestProperties
> Task :micrometer-core:shadowJar
> Task :micrometer-core:jar SKIPPED
> Task :micrometer-core:javadoc
> Task :micrometer-core:javadocJar
> Task :micrometer-core:sourceJar
> Task :micrometer-core:publishNebulaPublicationToMavenLocal
> Task :micrometer-core:publishToMavenLocal
...
```

jkschneider:
Interesting...

PR #1085, opened 2018-12-19
Make the static initializers of the Tags class @Nullable

### LINKED ISSUE #1078: Static initializers of the Tags class are not null-safe
It seems that some of the methods in the `Tags` class are null-safe (e.g.: `and(...)`) and some of them are not (`of(...)`). The practical reason behind this is the fact that I bumped into a library which is using some of these methods with null parameters (https://github.com/codecentric/chaos-monkey-spring-boot) which will result in NPE.

This is my proposed change, I would be happy to submit it as a PR: https://github.com/jonatan-ivanov/micrometer/commit/06c76aa15e078e32e9aaa692b5e3b883c10d7fc0

jonatan-ivanov:
Fixes #1078


jkschneider:
Thanks @jonatan-ivanov!

PR #1098, opened 2018-12-23
Utilize try-with-resources in tests

OleksiiBondar:
Close resources and other minor improvements in tests

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetricsTest.java
@@ -15,9 +15,12 @@
  */
 package io.micrometer.core.instrument.binder.kafka;
 
+import io.micrometer.core.instrument.ImmutableTag;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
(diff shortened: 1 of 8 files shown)


OleksiiBondar:
For me it fails with the same error in master branch as well:
`
Execution failed for task ':micrometer-spring-legacy:compileTestJava'.
> Could not resolve all files for configuration ':micrometer-spring-legacy:testCompileClasspath'.
   > Resolved 'org.mozilla:rhino:1.7.7' which is not part of the dependency lock state
`

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetricsTest.java
```diff
@@ -35,44 +40,40 @@
...
-        Properties props = new Properties();
-        props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, BOOTSTRAP_SERVERS);
-        props.put(ConsumerConfig.GROUP_ID_CONFIG, "MicrometerTestConsumer");
-        props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, LongDeserializer.class.getName());
-        props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
-
-        Consumer<Long, String> consumer = new KafkaConsumer<>(props);
-        consumer.subscribe(Collections.singletonList(TOPIC));
-    }
+    private List<Tag> tags = Arrays.asList(new ImmutableTag("app", "myApp"), new ImmutableTag("version", "1"));
```
jkschneider: The right side can all be replaced with `Tags.of("app, "myapp", "version", "1");`

jkschneider:
Don't think about the dependency lock state. I'm trying to wipe out dependency locks on master right now so you won't have to worry about that.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetricsTest.java
OleksiiBondar: Good point, addressed.

Issue #1099, opened 2018-12-26
Incorrect MeterFilter documentation


bisoldi:
The Concepts documentation for MeterFilter shows a convenience method `acceptNameStartsWith`, however MeterFilter contains no such method.

https://micrometer.io/docs/concepts#_convenience_methods
> acceptNameStartsWith(String) - Accept every meter with a matching prefix.

I found this because I was looking for a way to specify common tags on a per Meter basis, rather than for the entire Registry.  Other than knowing the Meter.Id, which doesn't seem easily done because the Id is created internally and can't be created externally, I don't see a way to accept a Meter by name and then set tags.

If it helps, I cross-posted this question on SO.

https://stackoverflow.com/questions/53925641/micrometer-common-tags-for-specific-metrics

Does anyone have a good solution for this, other than creating a separate MeterRegistry?

izeye:
@bisoldi Thanks for the report! For the inconsistency part I created #1100 to align with doc.

For the question part it seems to be resolved in StackOverflow.

PR #1100, opened 2018-12-26
Add MeterFilter.acceptNameStartsWith()

izeye:
This PR adds `MeterFilter.acceptNameStartsWith()` to resolve inconsistency with doc.

Closes gh-1099

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
@@ -290,6 +290,17 @@ public interface MeterFilter {
     }
 
     /**
+     * Meters that start with the provided name should be present in published metrics.
+     *
+     * @param prefix When a meter name starts with the prefix, guarantee its inclusion in published metrics.
+     * @return A filter that guarantees the inclusion of matching meters.
+     * @since 1.2.0
+     */
+    static MeterFilter acceptNameStartsWith(String prefix) {
+        return accept(id -> id.getName().startsWith(prefix));
+    }
+
+    /**
      * Set a maximum expected value on any {@link Timer} whose name begins with the given prefix.
      *
      * @param prefix Apply the maximum only to timers whose name begins with this prefix.
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/MeterFilterTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/MeterFilterTest.java
@@ -224,4 +224,14 @@ class MeterFilterTest {
         assertThat(filter.accept(id1)).isEqualTo(MeterFilterReply.NEUTRAL);
         assertThat(filter.accept(id2)).isEqualTo(MeterFilterReply.DENY);
     }
+
+    @Test
+    void acceptNameStartsWith() {
+        Meter.Id id1 = new Meter.Id("my.counter", Tags.empty(), null, null, Meter.Type.COUNTER);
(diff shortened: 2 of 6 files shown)


izeye:
FTR I'm not sure why the dependency locks were broken locally but I updated dependency locks anyway in b1800e1 as follows:

```
./gradlew clean check
...
* What went wrong:
Execution failed for task ':micrometer-spring-legacy:compileTestJava'.
> Could not resolve all files for configuration ':micrometer-spring-legacy:testCompileClasspath'.
   > Resolved 'org.mozilla:rhino:1.7.7' which is not part of the dependency lock state
...

./gradlew :micrometer-spring-legacy:testClasses --update-locks org.mozilla:rhino

...
Execution failed for task ':micrometer-spring-legacy:test'.
> Could not resolve all files for configuration ':micrometer-spring-legacy:testRuntimeClasspath'.
   > Resolved 'org.mozilla:rhino:1.7.7' which is not part of the dependency lock state
...

./gradlew :micrometer-spring-legacy:testRuntime --update-locks org.mozilla:rhino
```

OleksiiBondar:
Thanks @izeye! I've applied the same change in my PR which was failing with the same error.
I wonder why do we need these locks in general? 

Also, it helped with local build failure, but PR still failing - https://github.com/micrometer-metrics/micrometer/pull/1098? 
Any thoughts?

izeye:
> I wonder why do we need these locks in general?

@OleksiiBondar In Micrometer, dynamic dependency versions like ["latest.release"](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/build.gradle#L11) are used and in maintenance releases dependency upgrades are limited to their maintenance releases only, so dependency locking is necessary here. See https://docs.gradle.org/current/userguide/dependency_locking.html for more details.

> Also, it helped with local build failure, but PR still failing - #1098 even though changeset is exactly the same.
> Any thoughts?

As I said above, this dependency lock failures are not clear to me. To my knowledge I don't expect dependency lock failures without any dependency change. It might be a Gradle bug or feature I'm not aware of. For the failure of your PR in the CI, I don't know why it complains but I confirmed your branch builds successfully with `./gradlew clean check` locally. As @jkschneider suggested in https://github.com/micrometer-metrics/micrometer/pull/1098#issuecomment-449668821, you'd better just ignore the failure for now as the dependency locking seems to be removed in the master branch soon based on his comment.

OleksiiBondar:
@izeye @jkschneider IMO, using explicit dependency versions will address multiple problems including but not limited to:
 - no need to worry about dependency locking;
 - consistent behavior across different PRs against master branch;
 - more controlled/reproducible environment;
 - awareness and accountability from developer' side who updates dependency.

More arguments - https://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/

Don't get me wrong - I'm an advocate of using latest and greatest versions, but I'd prefer to have this as deliberate and verified change rather than blindly upgrade versions until it fail.
Of course, this approach has down side - someone will haveto time to time update dependency versions. But to me, thit is lesser evil. 
Thoughts on this?

izeye:
@OleksiiBondar It's my first-time and only experience on the dependency locking and I still struggle with it as you can see, so I'm comfortable with what you suggested. But I don't have enough experience to express any opinion on the topic. @jkschneider might have a good reason to use it.

Anyway, it's not related to this PR, so if you continue the discussion, it'd be better to create a dedicate issue for it.

OleksiiBondar:
Make sense - I'll create separate issue to keep track of the conversation and output to not litter this PR :)

PR #1102, opened 2018-12-27
Influx: Move histogram/percentile fields into the timer table

machao23:
This PR add quantile fields in writeTimer method when publish to influxdb.

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -18,11 +18,12 @@ package io.micrometer.influx;
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.FunctionTimer;
+import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.LongTaskTimer;
 import io.micrometer.core.instrument.Measurement;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.Timer;
-import io.micrometer.core.instrument.distribution.ValueAtPercentile;
+import io.micrometer.core.instrument.search.MeterNotFoundException;
 import io.micrometer.core.instrument.step.StepMeterRegistry;
 import io.micrometer.core.instrument.util.DoubleFormat;
 import io.micrometer.core.instrument.util.MeterPartition;
@@ -38,6 +39,7 @@ import java.util.ArrayList;
 import java.net.MalformedURLException;
 import java.net.URLEncoder;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
@@ -223,19 +225,18 @@ public class InfluxMeterRegistry extends StepMeterRegistry {
                 new Field("upper", timer.max(getBaseTimeUnit()))
         );
 
-        final ValueAtPercentile[] percentileValues = timer.takeSnapshot().percentileValues();
-        if (percentileValues.length > 0) {
-            List<Field> quantiles = new ArrayList<>(percentileValues.length);
-            for (ValueAtPercentile v : percentileValues) {
-                quantiles.add(new Field("quantile" + v.percentile(), v.value(getBaseTimeUnit())));
-            }
-            totalFields = Stream.concat(fields, quantiles.stream());
+        try {
+            Collection<Gauge> percentilesGauges = this.get(timer.getId().getName() + ".percentile").gauges();
+            List<Field> percentiles = new ArrayList<>(percentilesGauges.size());
+            percentilesGauges.forEach( gauge -> {
+                percentiles.add(new Field("phi" + gauge.getId().getTag("phi"), gauge.value()));
+            });
+            totalFields = Stream.concat(fields, percentiles.stream());
+            return Stream.of(influxLineProtocol(timer.getId(), "histogram", totalFields));
+        } catch (MeterNotFoundException ex) {
+            // Not all timer has percentile gauges
+            return Stream.of(influxLineProtocol(timer.getId(), "histogram", fields));
         }
-        else {
-            totalFields = fields;
-        }
-
-        return Stream.of(influxLineProtocol(timer.getId(), "histogram", totalFields));
     }
 
     private Stream<String> writeSummary(DistributionSummary summary) {
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -22,7 +22,6 @@ import io.micrometer.core.instrument.step.StepMeterRegistry;
 import io.micrometer.core.instrument.util.*;
 import io.micrometer.core.ipc.http.HttpSender;
 import io.micrometer.core.ipc.http.HttpUrlConnectionSender;
-import org.apache.commons.lang.ArrayUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -15,11 +15,14 @@
  */
 package io.micrometer.influx;
 
+import com.google.common.collect.Lists;
 import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.distribution.ValueAtPercentile;
 import io.micrometer.core.instrument.step.StepMeterRegistry;
 import io.micrometer.core.instrument.util.*;
 import io.micrometer.core.ipc.http.HttpSender;
 import io.micrometer.core.ipc.http.HttpUrlConnectionSender;
+import org.apache.commons.lang.ArrayUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -201,6 +204,9 @@ public class InfluxMeterRegistry extends StepMeterRegistry {
     }
 
     private Stream<String> writeTimer(Timer timer) {
+
+        final Stream<Field> totalFields;
+
         final Stream<Field> fields = Stream.of(
                 new Field("sum", timer.totalTime(getBaseTimeUnit())),
                 new Field("count", timer.count()),
@@ -208,7 +214,19 @@ public class InfluxMeterRegistry extends StepMeterRegistry {
                 new Field("upper", timer.max(getBaseTimeUnit()))
         );
 
-        return Stream.of(influxLineProtocol(timer.getId(), "histogram", fields));
+        final ValueAtPercentile[] percentileValues = timer.takeSnapshot().percentileValues();
+        if (percentileValues.length > 0) {
+            List<Field> quantiles = Lists.newArrayListWithExpectedSize(percentileValues.length);
+            for (ValueAtPercentile v : percentileValues) {
+                quantiles.add(new Field("quantile" + v.percentile(), v.value(getBaseTimeUnit())));
+            }
+            totalFields = Stream.concat(fields, quantiles.stream());
+        }
+        else {
+            totalFields = fields;
+        }
+
+        return Stream.of(influxLineProtocol(timer.getId(), "histogram", totalFields));
     }
 
     private Stream<String> writeSummary(DistributionSummary summary) {
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -15,7 +15,6 @@
  */
 package io.micrometer.influx;
 
-import com.google.common.collect.Lists;
 import io.micrometer.core.instrument.*;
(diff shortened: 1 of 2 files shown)


machao23:
I don't know why the dependency lock check was failed,  I haven't add or modify the dependency.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -183,14 +185,29 @@ protected void publish() {
...
+
         final Stream<Field> fields = Stream.of(
                 new Field("sum", timer.totalTime(getBaseTimeUnit())),
                 new Field("count", timer.count()),
                 new Field("mean", timer.mean(getBaseTimeUnit())),
                 new Field("upper", timer.max(getBaseTimeUnit()))
         );
 
-        return Stream.of(influxLineProtocol(timer.getId(), "histogram", fields));
+        final ValueAtPercentile[] percentileValues = timer.takeSnapshot().percentileValues();
```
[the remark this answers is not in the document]
machao23: Thanks @OleksiiBondar ! I have commited as your suggestion.

implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryWriteTest.java
```diff
@@ -0,0 +1,58 @@
+package io.micrometer.influx;
```
OleksiiBondar: Please add copyright header to consistency

implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryWriteTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Test
+    public void timerPercentileTest() throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
+        Timer timer = Timer.builder("test.timer")
+            .publishPercentiles(0.5, 0.95)
+            .register(registry);
+
+        timer.record(Duration.ofSeconds(5));
+        timer.record(Duration.ofSeconds(1));
+        String[] result = doTestWriteTimer(timer).collect(Collectors.toList()).get(0).split(",");
+        assertEquals(timer.takeSnapshot().percentileValues().length, 2);
```
OleksiiBondar: `timer.takeSnapshot().percentileValues()` can be extracted as local variable to improve readability

implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryWriteTest.java
```diff
@@ -0,0 +1,58 @@
...
+            .isEqualTo(1, offset(0.1));
+        assertThat(timer.takeSnapshot().percentileValues()[1].value(TimeUnit.SECONDS))
+            .isEqualTo(5.0, offset(0.1));
+
+        assertEquals(result[0], "test_timer");
+        assertTrue(result[5].startsWith("quantile0.5="));
+        assertTrue(result[6].startsWith("quantile0.95="));
+    }
+
+    /**
```
OleksiiBondar: Probably, typo in method name - should be `writeTmimer` instead of `testWriteTimer`

machao23: Thanks @OleksiiBondar !  Already resolve these problem.

machao23:
@jkschneider , I have changed to get percentils from percentile gauges, but it seems like if  get percentils from the timer without percentile gauge, the requiredSearch instance will throw the  MeterNotFoundException, is it right or any better idea?

jkschneider:
@machao23 The percentile gauges are a rare form of synthetic meter not searchable at the moment. Since we are already shipping percentiles and histogram buckets, I believe this PR can be closed as duplicating an already existing feature?

machao23:
@jkschneider I understand what you mean, but I think it would be better to publish the percentages under the influxdb table of the timer meter. What do you think?

jkschneider:
@machao23 I see what you're getting at now. Thanks. In this case, we should override `newTimer` in `InfluxMeterRegistry` so it doesn't _also_ ship these values as gauges separate from the timer table. And whatever we do for timers, we should also do for `DistributionSummary` which have the same characteristics (just not base units in the time domain).

shakuzen:
@machao23 could you rebase on the latest master so that the build gets run again?

machao23:
> @machao23 could you rebase on the latest master so that the build gets run again?

ok

shakuzen:
There are a lot of commits that shouldn't be there if successfully rebased. Could you check again, please?

machao23:
> There are a lot of commits that shouldn't be there if successfully rebased. Could you check again, please?


I am not very familiar with git. Please tell me how to roll back to the previous rebase and then rebase the latest master correctly

machao23:
> There are a lot of commits that shouldn't be there if successfully rebased. Could you check again, please?

I reset the branch to the previous rebase, then force push, and finally rebase the latest master according to  [this guide](https://help.github.com/articles/syncing-a-fork/) Is this correct?

shakuzen:
The rebase looks good now. Thank you.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -211,7 +243,24 @@ protected void publish() {
...
+        return Stream.of(influxLineProtocol(summary.getId(), "histogram", totalFields));
+    }
+
+    private Stream<Field> writePercentiles(HistogramSupport histogramSupport, Stream<Field> fields) {
+        Stream<Field> totalFields;
+        final ValueAtPercentile[] percentileValues = histogramSupport.takeSnapshot().percentileValues();
+        if (percentileValues.length > 0) {
+            List<Field> quantiles = new ArrayList<>(percentileValues.length);
+            for (ValueAtPercentile v : percentileValues) {
+                quantiles.add(new Field("phi" + v.percentile(), v.value(getBaseTimeUnit())));
```
michaelandrepearce: what is phi? normally prefix percentiles with just a "p" as like in dropwizard console reporters.

michaelandrepearce:
@jkschneider when will this be available?

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
machao23: To consistent with the prefix of percentiles tag registered in the HistogramGauges.registerWithCommonFormat method

shakuzen:
> I think it would be better to publish the percentages under the influxdb table of the timer meter.

@machao23 Could you explain why this would be better or what the issues are with the current way percentiles are written?

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
shakuzen: This includes value information (the percentile) in the field name, which feels to me like an abuse. Is there some precedent of doing this in InfluxDB?

michaelandrepearce:
Because you just want the reporter to report back the percentiles not all the values. This is how dropwizard worked

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
michaelandrepearce: Yes this is how others did thus dropwizard and appmetrics. Though prefix of just p is more typically used.

machao23:
> > I think it would be better to publish the percentages under the influxdb table of the timer meter.
> 
> @machao23 Could you explain why this would be better or what the issues are with the current way percentiles are written?


The  issues with the current way is that timer meter is not to reporter percentiles to influxDB, not friendly enough for some monitoring services that do not have percentiles computing capabilities.

shakuzen:
> Because you just want the reporter to report back the percentiles not all the values.

> The issues with the current way is that timer meter is not to reporter percentiles to influxDB, not friendly enough for some monitoring services that do not have percentiles computing capabilities.

The existing implementation reports configured percentiles, just in a separate measurement from a synthetic meter. For example, the `http.server.requests` Timer metric that times HTTP server requests reports the configured percentiles as synthetic gauges to the `http_server_requests_percentile` measurement like the following:

```
name: http_server_requests_percentile
time                exception method metric_type outcome phi  status uri        value
----                --------- ------ ----------- ------- ---  ------ ---        -----
1561632329850000000 None      GET    gauge       SUCCESS 0.8  200    /count     96.206848
1561632329850000000 None      GET    gauge       SUCCESS 0.9  200    /count     96.206848
```

This pull request moves those percentiles, for example, to the `http_server_requests` measurement as fields, which is a breaking change for users depending on the values being in the `http_server_requests_percentile` measurement. It does seem like it will be more space efficient to have them as fields in one measurement (the same one as the timer/summary), but breaking compatibility is a significant concern.

For properly prioritizing this, I'm trying to understand what is problematic with the current way the percentiles are reported to InfluxDB - that is, in a measurement with `_percentile` appended to the name.
For example, to get all the configured percentiles for `http.server.requests` by URI, I wrote the following InfluxQL for a Grafana panel: `SELECT "value" FROM "http_server_requests_percentile" WHERE $timeFilter GROUP BY "phi", "uri"`.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
shakuzen: @michaelandrepearce thank you. I see a Dropwizard implementation here as you described it: https://github.com/iZettle/dropwizard-metrics-influxdb/blob/f0d0a9567cedd621630d4a9541459e4e3262aa25/metrics-influxdb/src/main/java/com/izettle/metrics/influxdb/InfluxDbReporter.java

It would be nice if InfluxDB documented a preference for this that different implementations could follow to be aligned.

machao23:
> > Because you just want the reporter to report back the percentiles not all the values.
> 
> > The issues with the current way is that timer meter is not to reporter percentiles to influxDB, not friendly enough for some monitoring services that do not have percentiles computing capabilities.
> 
> The existing implementation reports configured percentiles, just in a separate measurement from a synthetic meter. For example, the `http.server.requests` Timer metric that times HTTP server requests reports the configured percentiles as synthetic gauges to the `http_server_requests_percentile` measurement like the following:
> 
> ```
> name: http_server_requests_percentile
> time                exception method metric_type outcome phi  status uri        value
> ----                --------- ------ ----------- ------- ---  ------ ---        -----
> 1561632329850000000 None      GET    gauge       SUCCESS 0.8  200    /count     96.206848
> 1561632329850000000 None      GET    gauge       SUCCESS 0.9  200    /count     96.206848
> ```
> 
> This pull request moves those percentiles, for example, to the `http_server_requests` measurement as fields, which is a breaking change for users depending on the values being in the `http_server_requests_percentile` measurement. It does seem like it will be more space efficient to have them as fields in one measurement (the same one as the timer/summary), but breaking compatibility is a significant concern.
> 
> For properly prioritizing this, I'm trying to understand what is problematic with the current way the percentiles are reported to InfluxDB - that is, in a measurement with `_percentile` appended to the name.
> For example, to get all the configured percentiles for `http.server.requests` by URI, I wrote the following InfluxQL for a Grafana panel: `SELECT "value" FROM "http_server_requests_percentile" WHERE $timeFilter GROUP BY "phi", "uri"`.

Ok, I just know that there is a measurement with a percentile suffix. This meets my needs. But this measurement is so hidden that I have to implement this function myself at first, thank you.

michaelandrepearce:
How does it break that? I dont see any change in that code area? This is just additional no? I didnt write this patch, but from the look of it and applying the patch locally it seems like it.

When having millions of metrics the overhead is massive. Its so ineffcient space wise, and querying is expensive, e.g.  query time is significantly larger when making overlayed graphs where as all in one record/row it is almost 10x faster on query. Making it a performance issue

machao23:
> How does it break that? I dont see any change in that code area? This is just additional no? I didnt write this patch, but from the look of it and applying the patch locally it seems like it.
> 
> When having millions of metrics the overhead is massive. Its so ineffcient space wise, and querying is expensive, e.g. query time is significantly larger when making overlayed graphs where as all in one record/row it is almost 10x faster on query. Making it a performance issue

That's right, the current way does have performance issues, especially in big data scenarios.

shakuzen:
> How does it break that? I dont see any change in that code area? This is just additional no? I didnt write this patch, but from the look of it and applying the patch locally it seems like it.

The current changes stop writing to the `_percentiles` measurement in favor of the added fields in the non-suffixed measurement. This would be a breaking change for those expecting the data in that measurement. That said, we could write to both measurements until Micrometer 2.0 where we could stop writing the `_percentiles` measurement altogether.

> When having millions of metrics the overhead is massive. Its so ineffcient space wise, and querying is expensive, e.g. query time is significantly larger when making overlayed graphs where as all in one record/row it is almost 10x faster on query. Making it a performance issue

Thank you for sharing that. This is the kind of information I was looking for in order to prioritize.
To keep backwards compatibility, we would need to continue writing the `_percentiles` measurement, but if it is a substantial query performance improvement, we could start writing the percentile values as fields in the timer/summary measurements, like proposed in this pull request. Would this be worth doing?

shakuzen:
We are going through old pull requests and deciding what to do on them. We haven't heard back about my last question. It sounds like the most important outcome here is to properly document how percentiles are written to Influx, since that wasn't clear. I have opened https://github.com/micrometer-metrics/micrometer-docs/issues/224 to do that. If you would like, feel free to propose a documentation update for that. Otherwise, we will take care of it.

If the performance of the current implementation is still a problem for users, please open a new issue related to that with details and we will again consider things from that angle. Thanks.

PR #1107, opened 2018-12-31
Revise cache metrics implementations + tests

OleksiiBondar:
Initially, I was planning just cover package `io.micrometer.core.instrument.binder.cache` with unit tests since there was none. As a result of this effort, I've addressed few bugs/improvements:
* removed non-needed null checks from `CacheMeterBinder` - null values defaulted to 0;
* appended common tags to `cache.remoteSize` metric - can't think of a good reason why it shouldn't be there unless accidentally missed;
* added null check for `objectName` before `MbeanServer.getAttribute` since could be initialized as per implementation (i.e. when `cacheManager` is null)
* used proper methods for
** `cache.put`: `cachePutAddedCount` -> `cachePutUpdatedCount`
** `getLocalOffHeapSize` -> `getLocalOffHeapSizeInBytes`
* add unit test coverage:
Initial coverage
image

After these changes:
image

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
@@ -23,7 +23,6 @@ import io.micrometer.core.lang.Nullable;
 
 import java.lang.ref.WeakReference;
 
-
 /**
  * A common base class for cache metrics that ensures that all caches are instrumented
  * with the same basic set of metrics while allowing for additional detail that is specific
@@ -47,17 +46,23 @@ public abstract class CacheMeterBinder implements MeterBinder {
 
     @Override
     public final void bindTo(MeterRegistry registry) {
-        Long cacheSize = size();
-        if (cacheSize != null) {
-            Gauge.builder("cache.size", cache.get(), c -> cacheSize)
+        if (size() != null) {
+            Gauge.builder("cache.size", cache.get(),
+                    c -> {
+                        Long size = size();
+                        return size == null ? 0 : size;
+                    })
                     .tags(tags)
                     .description("The number of entries in this cache. This may be an approximation, depending on the type of cache.")
                     .register(registry);
         }
 
-        Long missCount = missCount();
-        if (missCount != null) {
-            FunctionCounter.builder("cache.gets", cache.get(),  c -> missCount)
+        if (missCount() != null) {
+            FunctionCounter.builder("cache.gets", cache.get(),
+                    c -> {
+                        Long misses = missCount();
+                        return misses == null ? 0 : misses;
+                    })
                     .tags(tags).tag("result", "miss")
                     .description("the number of times cache lookup methods have returned an uncached (newly loaded) value, or null")
                     .register(registry);
@@ -73,9 +78,12 @@ public abstract class CacheMeterBinder implements MeterBinder {
                 .description("The number of entries added to the cache")
                 .register(registry);
 
-        Long evictionCount = evictionCount();
-        if (evictionCount != null) {
-            FunctionCounter.builder("cache.evictions", cache.get(), c -> evictionCount)
+        if (evictionCount() != null) {
+            FunctionCounter.builder("cache.evictions", cache.get(),
+                    c -> {
+                        Long evictions = evictionCount();
+                        return evictions == null ? 0 : evictions;
+                    })
                     .tags(tags)
                     .description("cache evictions")
                     .register(registry);
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
@@ -47,27 +47,23 @@ public abstract class CacheMeterBinder implements MeterBinder {
 
     @Override
     public final void bindTo(MeterRegistry registry) {
-        if (size() != null) {
-            Gauge.builder("cache.size", cache.get(),
-                    c -> {
-                        Long size = size();
-                        return size == null ? 0 : size;
-                    })
-                    .tags(tags)
-                    .description("The number of entries in this cache. This may be an approximation, depending on the type of cache.")
-                    .register(registry);
-        }
+        Gauge.builder("cache.size", cache.get(),
+                c -> {
+                    Long size = size();
+                    return size == null ? 0 : size;
+                })
+                .tags(tags)
+                .description("The number of entries in this cache. This may be an approximation, depending on the type of cache.")
+                .register(registry);
 
-        if (missCount() != null) {
-            FunctionCounter.builder("cache.gets", cache.get(),
-                    c -> {
-                        Long misses = missCount();
-                        return misses == null ? 0 : misses;
-                    })
-                    .tags(tags).tag("result", "miss")
-                    .description("the number of times cache lookup methods have returned an uncached (newly loaded) value, or null")
-                    .register(registry);
-        }
+        FunctionCounter.builder("cache.gets", cache.get(),
+                c -> {
+                    Long misses = missCount();
+                    return misses == null ? 0 : misses;
+                })
+                .tags(tags).tag("result", "miss")
+                .description("the number of times cache lookup methods have returned an uncached (newly loaded) value, or null")
+                .register(registry);
 
         FunctionCounter.builder("cache.gets", cache.get(), c -> hitCount())
                 .tags(tags).tag("result", "hit")
@@ -79,16 +75,14 @@ public abstract class CacheMeterBinder implements MeterBinder {
                 .description("The number of entries added to the cache")
                 .register(registry);
 
-        if (evictionCount() != null) {
-            FunctionCounter.builder("cache.evictions", cache.get(),
-                    c -> {
-                        Long evictions = evictionCount();
-                        return evictions == null ? 0 : evictions;
-                    })
-                    .tags(tags)
-                    .description("cache evictions")
-                    .register(registry);
-        }
+        FunctionCounter.builder("cache.evictions", cache.get(),
+                c -> {
+                    Long evictions = evictionCount();
+                    return evictions == null ? 0 : evictions;
+                })
+                .tags(tags)
+                .description("cache evictions")
+                .register(registry);
 
         bindImplementationSpecificMetrics(registry);
     }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
@@ -47,23 +47,19 @@ public abstract class CacheMeterBinder implements MeterBinder {
 
     @Override
     public final void bindTo(MeterRegistry registry) {
-        Gauge.builder("cache.size", cache.get(),
-                c -> {
-                    Long size = size();
-                    return size == null ? 0 : size;
-                })
-                .tags(tags)
-                .description("The number of entries in this cache. This may be an approximation, depending on the type of cache.")
-                .register(registry);
+        if (size() != null) {
+            Gauge.builder("cache.size", cache.get(), c -> size())
+                    .tags(tags)
+                    .description("The number of entries in this cache. This may be an approximation, depending on the type of cache.")
+                    .register(registry);
+        }
 
-        FunctionCounter.builder("cache.gets", cache.get(),
-                c -> {
-                    Long misses = missCount();
-                    return misses == null ? 0 : misses;
-                })
-                .tags(tags).tag("result", "miss")
-                .description("the number of times cache lookup methods have returned an uncached (newly loaded) value, or null")
-                .register(registry);
+        if (missCount() != null) {
+            FunctionCounter.builder("cache.gets", cache.get(),  c -> missCount())
+                    .tags(tags).tag("result", "miss")
+                    .description("the number of times cache lookup methods have returned an uncached (newly loaded) value, or null")
+                    .register(registry);
+        }
 
         FunctionCounter.builder("cache.gets", cache.get(), c -> hitCount())
                 .tags(tags).tag("result", "hit")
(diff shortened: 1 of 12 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
```diff
@@ -67,7 +67,7 @@ public CaffeineCacheMetrics(Cache<?, ?> cache, String cacheName, Iterable<Tag> t
      * @param <C>       The cache type.
      * @return The instrumented cache, unchanged. The original cache is not wrapped or proxied in any way.
      */
-    public static <C extends Cache> C monitor(MeterRegistry registry, C cache, String cacheName, String... tags) {
+    public static <C extends Cache<?, ?>> C monitor(MeterRegistry registry, C cache, String cacheName, String... tags) {
```
jkschneider: Could you help me understand in what circumstances this helps? It doesn't help the compiler infer generic type arguments on an expression involving `monitor` is evaluated, and the lack of wildcards doesn't generate compiler warnings about missing type arguments.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/EhCache2Metrics.java
```diff
@@ -102,7 +103,7 @@ protected void bindImplementationSpecificMetrics(MeterRegistry registry) {
                 .description("Cache puts resulting in a new key/value pair")
                 .register(registry);
 
-        FunctionCounter.builder("cache.puts.added", stats, StatisticsGateway::cachePutAddedCount)
+        FunctionCounter.builder("cache.puts.added", stats, StatisticsGateway::cachePutUpdatedCount)
```
jkschneider: I wonder now if this meter is appropriately named. Perhaps it should be just `cache.puts`, keeping the result tag that indicates the action verb.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java
```diff
@@ -129,17 +129,19 @@ protected void bindImplementationSpecificMetrics(MeterRegistry registry) {
...
 
     private Long lookupStatistic(String name) {
-        try {
-            List<MBeanServer> mBeanServers = MBeanServerFactory.findMBeanServer(null);
-            for (MBeanServer mBeanServer : mBeanServers) {
-                try {
-                    return (Long) mBeanServer.getAttribute(objectName, name);
-                } catch (AttributeNotFoundException | InstanceNotFoundException ex) {
-                    // did not find MBean, try the next server
+        if (objectName != null) {
```
jkschneider: This guard looks important.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/AbstractCacheMetricsTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @author Oleksii Bondar
+ */
+abstract class AbstractCacheMetricsTest {
+
+    protected abstract Tags getTags();
+
+    /**
+     * Verifies base metrics presense
```
jkschneider: Spelling nit: `presence`

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetricsTest.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Test
+    void doNotReportMetricsForNonLoadingCache() {
+        SimpleMeterRegistry meterRegistry = new SimpleMeterRegistry();
+        Cache<Object, Object> cache = Caffeine.newBuilder().build();
+        CaffeineCacheMetrics metrics = new CaffeineCacheMetrics(cache, "testCache", expectedTag);
+        metrics.bindTo(meterRegistry);
+
+        assertThrows(MeterNotFoundException.class, () -> {
```
jkschneider: You can use `assertThat(meterRegistry.find("cache.load.duration").timeGauge()).isNull())`, which is a bit simpler. `find` is a relaxed version of the stricter `get`. Since the meter won't exist at all, I think it's unnecessary to add the `expectedTag` to the search.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/EhCache2MetricsTest.java
```diff
@@ -0,0 +1,201 @@
...
+
+    private static CacheManager cacheManager;
+    private static Cache cache;
+
+    private Tags expectedTag = Tags.of("app", "test");
+    private EhCache2Metrics metrics = new EhCache2Metrics(cache, expectedTag);
+
+    @Test
+    void reportMetrics() {
+        SimpleMeterRegistry registry = new SimpleMeterRegistry();
```
jkschneider: Left-hand side of assignment can be `MeterRegistry`

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/AbstractCacheMetricsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.search.RequiredSearch;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+
+/**
+ * @author Oleksii Bondar
+ */
+abstract class AbstractCacheMetricsTest {
+
+    protected abstract Tags getTags();
```
jkschneider: Since the same `app=test` tag is defined on all of the subclasses of `AbstractCacheMetricsTest`, can the setup be simplified?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetricsTest.java
```diff
@@ -0,0 +1,191 @@
...
+class HazelcastCacheMetricsTest extends AbstractCacheMetricsTest {
+
+    private static IMap<String, String> cache;
+
+    private Tags expectedTag = Tags.of("app", "test");
+    private HazelcastCacheMetrics metrics = new HazelcastCacheMetrics(cache, expectedTag);
+
+    @Test
+    void reportMetrics() {
+        SimpleMeterRegistry meterRegistry = new SimpleMeterRegistry();
```
jkschneider: Simplify left-hand assignment to `MeterRegistry`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
OleksiiBondar: With out parametrization it actually generates compiler warnings
image

Apart from addressing warnings, it also makes implementation consistent across different cache metrics. Here are few examples where it's already parametrized:
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java#L54
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetrics.java#L62
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetrics.java#L47
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java#L58
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java#L73

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/EhCache2Metrics.java
OleksiiBondar: Good point! I've changed meter name to `cache.puts.updated` and removed qualifying tags since names are unique now.
IMO it's better to have fully qualified names here rather than `cache.puts` since we already have this name defined in https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java#L77
which corresponds to total number of puts per cache not segregated by added/modified flavor.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java
OleksiiBondar: Yes, even though probability is low there is still a chance of having `cacheManager` as null and it's better to be on the safe side.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetricsTest.java
OleksiiBondar: Thanks, addressed across all *CacheMetricsTest classes

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/AbstractCacheMetricsTest.java
OleksiiBondar: Right, moved to `AbstractCacheMetricsTest` and removed abstract method :)

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/EhCache2MetricsTest.java
OleksiiBondar: Thanks, addressed across all *CacheMetricsTest classes

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetricsTest.java
OleksiiBondar: Thanks, addressed

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/AbstractCacheMetricsTest.java
OleksiiBondar: Fixed

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
```diff
@@ -47,27 +47,23 @@ public CacheMeterBinder(Object cache, String cacheName, Iterable<Tag> tags) {
 
     @Override
     public final void bindTo(MeterRegistry registry) {
-        if (size() != null) {
```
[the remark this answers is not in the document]
OleksiiBondar: Make sense. I've also removed piece of code that defaults to 0 when value is null since we already have null check 
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java#L52-L55

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/EhCache2Metrics.java
jkschneider: Ah, I see. So it was a name disambiguation concern originally. I don't find `updated` to be any more accurate than `added`, so perhaps we should revert to `added` to preserve compatibility?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
jkschneider: Thanks. Ah Eclipse. I think it just isn't smart enough to tell that in this context the generic type parameters don't actually help the compiler. IntelliJ doesn't show the same warnings. I'm happy to leave the type wildcards to help our Eclipse friends.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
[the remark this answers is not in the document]
jkschneider: I'd like to preserve the null check inside of the gauge as well, for the event that a JCache implementation supports turning stats on and off at runtime. Just extra safety. Want to be really careful that telemetry code never blows up an application.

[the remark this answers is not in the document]
OleksiiBondar: I see. For this purpose I'd rather extract value into local variable prior to null evaluation so that we don't need to do double null check. Other aspect of it - it allows to cover code with unit tests, where with double null check it's close to impossible. Let me know if any questions/concerns.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/EhCache2Metrics.java
OleksiiBondar: Done

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
```diff
@@ -47,23 +47,17 @@ public CacheMeterBinder(Object cache, String cacheName, Iterable<Tag> tags) {
...
-            Gauge.builder("cache.size", cache.get(),
-                    c -> {
-                        Long size = size();
-                        return size == null ? 0 : size;
-                    })
+        Long cacheSize = size();
```
jkschneider: This introduces a bug, causing the initial cache size to be reported from this point forward.

OleksiiBondar: (facepalm) Thank you for catching this! Addressed

Issue #1108, opened 2018-12-31
changes to AppOpticsMeterRegistry resulting in invalid characters


HunterSherms:
I finally got around to upgrading to the official deployment of MicroMeter with the AppOptics implementation after relying on a custom build of my initial PR.

Something appears to be broken as invalid characters are getting sent to AppOptics. After noticing returned errors of:

```
failed to send metrics to appoptics: {"code":400,"message":"Unable to process JSON"}
```

I dropped a debugger inside of HttpSender.Builder.withJsonContent and saw lines like the following:

```
{"name":"system.cpu.usage","period":30,"value":�,"tags":{"_type":"gauge","hostname":"local","source":"local"}}
```

What could have brought about this change?

izeye:
@HunterSherms Thanks for the report! NaN handling seems to be missing. I created #1109 to resolve it.

OleksiiBondar:
@izeye @jkschneider Should it be closed than if appropriate PR with fix already merged? Or it requires release in order to close issue? Just curious what is a process :)

PR #1109, opened 2019-01-01
Prevent publishing NaN or infinite values to AppOptics

izeye:
This PR prevents publishing NaN or infinite values to AppOptics.

Closes gh-1108

### DIFF
--- a/implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsMeterRegistry.java
+++ b/implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsMeterRegistry.java
@@ -43,6 +43,7 @@ import static java.util.stream.Collectors.joining;
  *
  * @author Hunter Sherman
  * @author Jon Schneider
+ * @author Johnny Lim
  * @since 1.1.0
  */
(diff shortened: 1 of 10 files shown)


jkschneider:
Note to self: rebase on 1.1.x before merging.

izeye:
@jkschneider Sorry, I forgot to change the target branch. Now I changed it to the 1.1.x branch.

jkschneider:
I suppose we'll need to delete the dependency lock changes on the 1.1.x branch.

izeye:
@jkschneider The failing CI task was `downloadDependencies` which I didn't run locally, and I updated dependency locks for it. Now the CI build seems to be okay although I don't understand the root cause of it in the first place. I think it'd be better to include the depedency lock updates to satisfy Gradle unless there's any other way to make it happy.

izeye:
FTR I thought the dependency lock problem happens only in Gradle 5.0 but it also happens in [Gradle 4.10.2 which the 1.1.x branch uses](https://github.com/micrometer-metrics/micrometer/blob/1.1.x/build.gradle#L100) although I'm not sure it's a bug or a feature which I don't know.

jkschneider:
I see, a patch upgrade of Spring Boot, from which I suppose the rhino dependency is derived transitively.

PR #1110, opened 2019-01-01
Migrate to junit5

OleksiiBondar:
Intent is to have consistent JUnit5 testing paradigm across all modules. Apart from zillion locks changes are:
* replace annotations 
@RunWith(SpringRunner.class) -> @ExtendWith(SpringExtension.class)
org.junit.Test ->  org.junit.jupiter.api.Test
image

* change class and method access modifiers to package-private as per JUnit5 conventions;

The only module using JUnit4 is `micrometer-jersey2` referring to open issue of `JerseyTest` incompatibility with JUnit5.
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-jersey2/build.gradle#L15

https://github.com/jersey/jersey/issues/3662

### DIFF
--- a/benchmarks/benchmarks-core/gradle/dependency-locks/annotationProcessor.lockfile
+++ b/benchmarks/benchmarks-core/gradle/dependency-locks/annotationProcessor.lockfile
@@ -1,0 +1,3 @@
+# This is a Gradle generated file for dependency locking.
+# Manual edits can break the build and are not advised.
+# This file is expected to be part of source control.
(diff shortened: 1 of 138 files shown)


jkschneider:
I'm sorry you went through all this work. We have intentionally used JUnit 4 on the `micrometer-spring-legacy` module so that it is easy to migrate tests to and from the Spring Boot 1.5 codebase. This module will eventually be retired along with Spring Boot 1.5.

OleksiiBondar:
@jkschneider no worries. Two lessons learned here:
 - ask before doing;
 - make changes as granular as possible (there were few minor improvements to tests other than just migration - will address as separate PR then).

Happy New Year ! :)

PR #1111, opened 2019-01-01
Add tests for io.micrometer.core.instrument.noop package

OleksiiBondar:
Add unit tests for io.micrometer.core.instrument.noop package

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/noop/NoopTimerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/noop/NoopTimerTest.java
@@ -1,0 +1,103 @@
+/**
+ * Copyright 2019 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/noop/NoopTimerTest.java
```diff
@@ -0,0 +1,103 @@
...
+    private NoopTimer timer = new NoopTimer(id);
+
+    @Test
+    void returnsId() {
+        assertThat(timer.getId()).isEqualTo(id);
+    }
+
+    @Test
+    void recordSupplier() {
+        Supplier<String> supplier = mock(Supplier.class);
```
jkschneider: Rather than mocking a `Supplier`, can we just inline a real `Supplier`? I'd prefer to keep mocking limited to where it is strictly necessary. Same for `Callable` below.

OleksiiBondar: Done

Issue #1115, opened 2019-01-02
omit NaN metrics from dynatrace upload


pjfanning:
I had issues where a metric started to evaluate as Double.NaN and the micrometer-registry-dynatrace DynatraceTimeSeries.asJson was generating invalid JSON for the number.

I suggest that we skip over metrics that eval to NaN instead of adding them to the JSON upload.

My metric was registered like this (in Scala)
```
    meterRegistry.gauge("kontainers.version", Seq(Tag.of("deviceId", deviceId)).asJava,
      new java.lang.Double(201901022345.0))
```

pjfanning:
similar to https://github.com/micrometer-metrics/micrometer/pull/1109

izeye:
@pjfanning Thanks for the report! I created #1116 to resolve this.

PR #1116, opened 2019-01-02
Prevent publishing NaN or infinite values to Dynatrace

izeye:
This PR prevents publishing NaN or infinite values to Dynatrace.

A build for this PR will fail due to the dependency lock problem. After inclusion of the dependency lock updates in #1109, I expect the build will pass.

Closes gh-1115

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
@@ -136,13 +136,8 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
     Stream<DynatraceCustomMetric> writeMeter(Meter meter) {
         final long wallTime = clock.wallTime();
         return StreamSupport.stream(meter.measure().spliterator(), false)
-                .filter(this::isPublishable)
+                .filter(ms -> Double.isFinite(ms.getValue()))
                 .map(ms -> createCustomMetric(meter.getId(), wallTime, ms.getValue()));
-    }
-
-    private boolean isPublishable(Measurement measurement) {
-        double value = measurement.getValue();
-        return !Double.isNaN(value) && !Double.isInfinite(value);
     }
 
     private Stream<DynatraceCustomMetric> writeLongTaskTimer(LongTaskTimer longTaskTimer) {
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
@@ -132,10 +132,17 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
         }
     }
 
-    private Stream<DynatraceCustomMetric> writeMeter(Meter meter) {
+    // VisibleForTesting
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
```diff
@@ -185,11 +187,15 @@ private DynatraceCustomMetric createCustomMetric(Meter.Id id, long time, Number
     }
 
     private DynatraceCustomMetric createCustomMetric(Meter.Id id, long time, Number value, @Nullable DynatraceUnit unit) {
+        double doubleValue = value.doubleValue();
```
OleksiiBondar: Can we add unit tet for this behavior?

izeye: @OleksiiBondar Sure. I thought so but I've been lazy as it was in the middle of the bed time 😅 

Thanks for pointing it out. I added tests in 76cc010.

izeye:
I rebased this onto the current 1.1.x branch to make the build green.

izeye:
I reworked this to try to align with handling in other meter registries.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
```diff
@@ -132,12 +132,19 @@ protected void publish() {
...
-    private Stream<DynatraceCustomMetric> writeMeter(Meter meter) {
+    // VisibleForTesting
+    Stream<DynatraceCustomMetric> writeMeter(Meter meter) {
         final long wallTime = clock.wallTime();
         return StreamSupport.stream(meter.measure().spliterator(), false)
+                .filter(this::isPublishable)
                 .map(ms -> createCustomMetric(meter.getId(), wallTime, ms.getValue()));
     }
 
+    private boolean isPublishable(Measurement measurement) {
```
OleksiiBondar: I've noticed that this is not a first PR where we adding `!Double.isNaN(value) && !Double.isInfinite(value)` check across different MeterRegistries. Would it make sense to move this method into `MeterRegistry` itself or into some utility/helper class and utilize in all registries (Dynatrace, Atlas, Azure, Datadog, InfluxDB, CloudWatch, etc)?

2 things can be addressed by doing this:
* consistency and maintanability since this check logic present in a single place and it's easier to update if neede;
* revise existing registries to make sure all of them have this check in place rather than addressing problems one by one.

Can be addressed as a separate item.

izeye: @OleksiiBondar Thanks for the feedback! I thought the same before but I wasn't sure all the meter registries could be handled in the same way for NaN and infinity values. I created #1123 to check and resolve it if possible.

For the verbosity of the check logic I changed it to use `Double.isFinite()` in d70cce7. Other places can be polished separately or wait for #1123.

OleksiiBondar: Great, thank you!

jkschneider: It's true that some monitoring systems are OK with NaN and Infinity, so we can't handle it in `MeterRegistry`, but the helper utility is welcome.

Issue #1090, opened 2018-12-21
Please add formatter and import order settings 


OleksiiBondar:
Please add code style formatter and import order settings to the repo so that every contributor will use the same settings. 
After this is done, we could also bring spotless plugin (https://github.com/diffplug/spotless) and appropriate step in build process to validate/apply consistent formatting to all java-files.

mweirauch:
To my knowledge, spotless only supports the Eclipse formatter. This would mean trying to export the IDEA settings to Eclipse and then dumping them into SCM. Same goes for the import ordering.

jonatan-ivanov:
+1

Also, according to @jkschneider, the project is using IntelliJ defaults right now.

OleksiiBondar:
@jonatan-ivanov Not all contributors are using Intelij. And it's already misformatted in multiple places. 
It's definitilly has low priority but a good improvement going forward.
If not, who want to volunteer to keep formatting whole repo on some schedule basis (i.e. once a month)?

jonatan-ivanov:
@OleksiiBondar I completely agree with you, that was my observation too. :)
I think adding formatter rules would be very beneficial here.

jkschneider:
Agreed it would be beneficial to have as rules, especially for PRs we merge without noticing import ordering issues.

izeye:
This seems to have been superseded by applying Spring Java Format.

shakuzen:
Superseded by #3156

PR #1124, opened 2019-01-05
WIP - Add spotless plugin with formatting rules 

OleksiiBondar:
Not intended to be merged yet. Want to share current progress on the https://github.com/micrometer-metrics/micrometer/issues/1090.
Feedback is welcomed and highly appreciated.
Some basic info around spotless plugin:
https://github.com/diffplug/spotless/blob/master/plugin-gradle/README.md#applying-to-java-source

I've tried to make formatter/import order settings as much close to existing one, but there are still certain gaps due to inconsitent formatting (not sure what is expected) and incompleteness of formatter itself. 

After settings finalized (maybe even merged to master? or we can do within the same PR) I'd expect someone to run `./gradlew spotlessApply` against repo at a time when impact of merge conflicts is minimal (least open PRs as possible).  When this is done, we need to turn on force validation by removing `enforceCheck false` flag, so that we could prevent misformatting in the future.


build.gradle
```diff
@@ -73,16 +74,30 @@ subprojects {
...
+    
+    spotless {
+        java {
+            target project.fileTree(project.rootDir) {
+                include '**/*.java'
+                exclude '**/package-info.java'
+            }
+            importOrder 'cern', 'com', 'io', 'org', 'reactor', 'java', 'javax', 'static *'
+            eclipse().configFile 'config/checkstyle/micrometer-style.xml'
+            licenseHeaderFile rootProject.file('config/checkstyle/copyright.txt')
```
jkschneider: We do have the hieronymous license plugin applied currently which is doing some work on this. Are they overlapping?

build.gradle
```diff
@@ -73,16 +74,30 @@ subprojects {
...
-        toolVersion = '8.12'
-        configFile = rootProject.file('config/checkstyle/checkstyle.xml')
+    
+    spotless {
+        java {
+            target project.fileTree(project.rootDir) {
+                include '**/*.java'
+                exclude '**/package-info.java'
+            }
+            importOrder 'cern', 'com', 'io', 'org', 'reactor', 'java', 'javax', 'static *'
```
jkschneider: Basically, alphabetical except for `java*`, with statics last.

config/checkstyle/copyright.txt
```diff
@@ -0,0 +1,15 @@
+/**
```
jkschneider: Duplicates `gradle/licenseHeader.txt`

config/checkstyle/micrometer-style.xml
```diff
@@ -0,0 +1,336 @@
+<profiles version="13">
```
jkschneider: I'm a little nervous about this coming out of Eclipse. But I suppose we can test the validity of formatting by applying to the whole project, then opening IntelliJ and formatting the whole project. Should be zero changes on that last step.

jkschneider:
> I've tried to make formatter/import order settings as much close to existing one, but there are still certain gaps due to inconsistent formatting (not sure what is expected) and incompleteness of formatter itself.

* A few gaps are fine I think as long as the formatting output doesn't create a style that is altered by using the default IntelliJ formatter inside the IDE. Incremental improvement is always welcome. 
* If you're unsure what is expected currently, run the IntelliJ default Java formatter on a file. Whatever happens is correct.

jkschneider:
Also, I think before we make this a dependency of `check` that can fail the build I'd like to spend the time ensuring that we have a bot that autoremediates formatting in PRs as necessary. My main concern is reducing the friction or scare points for first time contributors. If they run a test suite from the IDE and everything passes, submit a PR, and the PR gives them a big nasty red X, that feels bad.

build.gradle
```diff
@@ -73,16 +74,30 @@ subprojects {
...
+            eclipse().configFile 'config/checkstyle/micrometer-style.xml'
```
mweirauch: Should this still be in a "checkstyle" subdir instead of "spotless"? (Just asking...)

config/checkstyle/micrometer-style.xml
```diff
@@ -0,0 +1,336 @@
...
+	<profile kind="CodeFormatterProfile" name="MicrometerStyle" version="1">
```
mweirauch: Perhaps just "Micrometer"? Also for the filename, just "micrometer-format.xml"?

build.gradle
OleksiiBondar: Yes, I understand that. I'm not sure that there is other way to specify import order with having `java*` last exept specifying all domains that should be before explicitly

config/checkstyle/micrometer-style.xml
OleksiiBondar: Yes, that was a plan. And as a I said, there are still minor differences that I'm working on.

config/checkstyle/micrometer-style.xml
OleksiiBondar: make sense, updated

build.gradle
OleksiiBondar: Originally, I was thinking of `checkstyle` as library-independent name that could prevent folder/file names change if we would go with other library in future. But since there is product that called the same way, it's probably makes it more confusing. Changed to spotless.

config/spotless/copyright.txt
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright $YEAR Pivotal Software, Inc.
+ * <p>
```
mweirauch: Do we still need these paragraph tags? I don't remember why they were added. I know from the basepom default checkstyle rules, that it nags on the trailing whitespace in a multiline comment ("' * '"), but as I see it checkstyle is no more.
Can you give it a local try if these paragraph tags are still needed? (Or can somebody comment why it is/was needed otherwise?)

mweirauch:
I am having a hard time even with a default IDEA 2018.3 CE not touching anything on format.
`Timer.java` is my test and it even formats the Javadoc to contain an empty line between description and the start of the tags. Additionally the Continuation Indent (8 spaces) seems to be ignored and makes them 4.

```
@@ -57,6 +59,7 @@ public interface Timer extends Meter, HistogramSupport {
 
     /**
      * Start a timing sample.
+     *
      * @param clock a clock to be used
      * @return A timing sample with start time recorded.
      */
@@ -83,10 +86,10 @@ public interface Timer extends Meter, HistogramSupport {
         }
 
         return new Builder(timed.value().isEmpty() ? defaultName : timed.value())
-                .tags(timed.extraTags())
-                .description(timed.description().isEmpty() ? null : timed.description())
-                .publishPercentileHistogram(timed.histogram())
-                .publishPercentiles(timed.percentiles().length > 0 ? timed.percentiles() : null);
+            .tags(timed.extraTags())
+            .description(timed.description().isEmpty() ? null : timed.description())
+            .publishPercentileHistogram(timed.histogram())
+            .publishPercentiles(timed.percentiles().length > 0 ? timed.percentiles() : null);
     }
 
     /**
@@ -182,9 +185,9 @@ public interface Timer extends Meter, HistogramSupport {
     @Override
     default Iterable<Measurement> measure() {
         return Arrays.asList(
-                new Measurement(() -> (double) count(), Statistic.COUNT),
-                new Measurement(() -> totalTime(baseTimeUnit()), Statistic.TOTAL_TIME),
-                new Measurement(() -> max(baseTimeUnit()), Statistic.MAX)
+            new Measurement(() -> (double) count(), Statistic.COUNT),
+            new Measurement(() -> totalTime(baseTimeUnit()), Statistic.TOTAL_TIME),
+            new Measurement(() -> max(baseTimeUnit()), Statistic.MAX)
         );
     }
 
@@ -462,7 +465,7 @@ public interface Timer extends Meter, HistogramSupport {
         public Timer register(MeterRegistry registry) {
             // the base unit for a timer will be determined by the monitoring system implementation
             return registry.timer(new Meter.Id(name, tags, null, description, Type.TIMER), distributionConfigBuilder.build(),
-                    pauseDetector == null ? registry.config().pauseDetector() : pauseDetector);
+                pauseDetector == null ? registry.config().pauseDetector() : pauseDetector);
         }
     }
 }
```
If some IDEA user could re-check that there is no changes applied in your "Default IDE" format?

Regarding the micrometer-format.xml, either my Eclipse 2018-12 fools me or I think it needs some huge love still. Still way to off. (No offense! Thanks for looking into this! I am all for a unified formatting across IDEs and I hate hand formatting stuff just to be compliant.)

OleksiiBondar:
@mweirauch You are totally right - formatter settings are still off. Main reason why is that - I can't come up with consistant formatting that will make least diff after reformatting.

I've tried to download latest Intelij CE and auto-format with default config whole repo and this action gave me 600+ classes changed with various differences including but not limited to javadocs, builder formatting, import order, etc.
All these makes me think that:
1. at some point we definitely need to reformat whole repo at least with some settings - it's far off from consistant state;
2. I'm guessing that different versions of Intelij may have different default settings;

I've stopped looking into this for time being - need some confirmation/input of Intelij default settings. Maybe having someone to re-format whole repo as a pre-requisite is right path forward. So that, we need to create xml file that won't produce any diff - that's could be a acceptance criteria.

OleksiiBondar:
Closing PR for a time being

Issue #1104, opened 2018-12-27
dynatrace limits uploads to 15k


pjfanning:
I suggest that we can support a dynatrace.maxMessageSize property.

https://github.com/kontainers/micrometer-registry-dynatrace/commit/48e27c7d428812f773978854b93692dc34319408 has an implementation where if the limit is reached, the upload data is split over multiple messages.

jkschneider:
Does dynatrace blow up if the 15k limit is exceeded?

pjfanning:
Dynatrace rejects the message if it exceeds 15k.

jkschneider:
Good to know. Is there any value then in making the max size configurable?

pjfanning:
In https://github.com/kontainers/micrometer-registry-dynatrace/commit/48e27c7d428812f773978854b93692dc34319408 - I split the message into mutiple messages so that none exceeds the maxMessageSize.

pjfanning:
I guess we could hardcode 15k. I can put together an MR in the coming days.

PR #1126, opened 2019-01-05
WIP: Split Dynatrace messages if they are too large

pjfanning:
If the Dynatrace message is greater than 15360 bytes, create multiple messages.

See #1104


jkschneider:
We'll need to rebase this on 1.1.x

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/Tuple.java
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
```
OleksiiBondar: 2019?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/Tuple.java
```diff
@@ -0,0 +1,25 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.dynatrace;
+
+class Tuple<X, Y> {
```
OleksiiBondar: Can we use `javafx.util.Pair` instead of creating new class?

pjfanning: https://www.infoworld.com/article/3305073/java/removed-from-jdk-11-javafx-11-arrives-as-a-standalone-module.html - seems like javafx may not be a good choice

commons-lang3 has a Tuple class - would that dependency be ok

the advantage of my current approach is that it limits dependencies on outside classes

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/Tuple.java
pjfanning: the rest of the dynatrace code was added in 2018, still has 2017 in the license files though

I'll change this one

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/Tuple.java
shakuzen: I don't think we want to depend on anything just for a Tuple class. Ideally, we would eliminate the need for a Tuple class or we can keep a simple internal class for it.

shakuzen:
As Jon stated before, this needs to be rebased on 1.1.x and changed to target that branch. Some unrelated changes are showing up in the diff currently.

Overall the approach looks right to me.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
```diff
@@ -231,21 +235,75 @@ void putCustomMetric(final DynatraceMetricDefinition customMetric) {
...
+                        .onError(response -> logger.error("failed to send metrics to dynatrace: {}", response.body()));
+            }
         } catch (Throwable e) {
             logger.error("failed to send metrics to dynatrace", e);
         }
     }
 
+    // VisibleForTesting
+    List<Tuple<String, Integer>> createPostMessages(String type, List<DynatraceTimeSeries> timeSeries) {
+        final StringBuilder sb = new StringBuilder(1024);
```
shakuzen: I don't think there is any benefit to using a `StringBuilder` here. The Java compiler will optimize String concatenation in simple cases like this ("string" + a + "string" + b) to use a `StringBuilder` in the bytecode. I think only in the case of concatenating in a loop is an explicit `StringBuilder` needed because it isn't optimized automatically.

pjfanning: new StringBuilder() uses a 16 char array so initialising the stringbuilder with a bigger array avoids the runtime from having to create new arrays as the string gets longer

```
   public StringBuilder() {
        super(16);
    }
```

pjfanning:
opened new PR (https://github.com/micrometer-metrics/micrometer/pull/1147)

PR #1127, opened 2019-01-05
Add dynamic tag resolution via OkHttp request tags for OkHttpMetricsEventListener.

### LINKED ISSUE #1106: OkHttp deduce dynamic tags from headers
`OkHttpMetricsEventListener` is supplied to OkHttp client via constructor and unfortunately there is no way of customisation of tags apart from url mapping in some request context. 

Would you consider to add another way of customised tags like getting from some special header `X-Micrometer-OkHttp-Tags-{tag name}` and its value from header value (we can discuss further how it should be done)? I can send a pr if you wish.

ping @bjartek

nyilmaz:
Resolves #1106

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
@@ -41,6 +41,7 @@ import java.util.function.Function;
  *
  * @author Bjarte S. Karlsen
  * @author Jon Schneider
+ * @author Nurettin Yilmaz
  */
 @NonNullApi
 @NonNullFields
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
```diff
@@ -100,15 +101,20 @@ public void responseHeadersEnd(Call call, Response response) {
...
 
     private void time(CallState state) {
+        Request request = state.request;
+        boolean requestAvailable = request != null;
+
         String uri = state.response == null ? "UNKNOWN" :
-            (state.response.code() == 404 || state.response.code() == 301 ? "NOT_FOUND" : urlMapper.apply(state.request));
+            (state.response.code() == 404 || state.response.code() == 301 ? "NOT_FOUND" : urlMapper.apply(request));
+
+        Tags dynamicTags = requestAvailable ? request.tag(Tags.class) : Tags.empty();
```
jkschneider: Out of curiosity, what happens if a user places a tag on the request that is not of type `Tags`?

nyilmaz: Then `request.tag(Tags.class)` returns `null`. `Request` holds a map as tag which typed as `<Class<T>, Object>` whose object is expected to have type `T`:
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Request.java#L94

nyilmaz: @jkschneider do you think there should be another marker class?

jkschneider: Perfect. Thanks!

jkschneider:
Neat feature. Thanks @nyilmaz!

bjartek:
What happends here if there is a uri request tag? Will that override the generated tag?

Since this is client metrics and you control them you should be able to set a uri even if the request fails?

jkschneider:
> What happens here if there is a uri request tag? Will that override the generated tag?

@bjartek Yes

PR #1135, opened 2019-01-08
#1134: Do not send "Infinity" values to elastic meter registry

### LINKED ISSUE #1134: Elastic: Failed to send metrics when using KafkaConsumerMetrics in 1.1 with default settings
For mm 1.1.1 with Kafka 2.1:

I have bound the KafakConsumer metrics like so:

```
new KafkaConsumerMetrics().bindTo(globalRegistry)
```

Also I have added the ElasticMeterRegistry to the globalRegistry.
Now I get the following error when micromter tries to send the metrics to my local elastic:

```
failed to send metrics to elastic: {"took":11,"errors":true,"item
.....
,"status":400,"error":{"type":"mapper_parsing_exception","reason":"failed to parse","caused_by":{"type":"json_parse_exception","reason":"Non-standard token '-Infinity': enable JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS to allow\n at [Source: org.elasticsearch.common.bytes.BytesReference$MarkSupportingStreamInputWrapper@273c3274; line: 1, column: 242]"}}}},
```

So it looks like some of the metrics is "-Infinity" which elastic won't accept.

Any ideas on how to start fixing this? I would be willing to help.

fkoehler:
This resolves #1134

### DIFF
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
@@ -184,7 +184,7 @@ public class ElasticMeterRegistry extends StepMeterRegistry {
     // VisibleForTesting
     Optional<String> writeGauge(Gauge gauge) {
         Double value = gauge.value();
-        if (!value.isNaN()) {
+        if (Double.isFinite(value)) {
             return Optional.of(writeDocument(gauge, builder -> {
                 builder.append(",\"value\":").append(value);
             }));
@@ -195,7 +195,7 @@ public class ElasticMeterRegistry extends StepMeterRegistry {
     // VisibleForTesting
     Optional<String> writeTimeGauge(TimeGauge gauge) {
         Double value = gauge.value();
-        if (!value.isNaN()) {
+        if (Double.isFinite(value)) {
             return Optional.of(writeDocument(gauge, builder -> {
                 builder.append(",\"value\":").append(gauge.value(getBaseTimeUnit()));
             }));
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
@@ -169,16 +169,21 @@ public class ElasticMeterRegistry extends StepMeterRegistry {
 
     // VisibleForTesting
     Optional<String> writeCounter(Counter counter) {
-        return Optional.of(writeDocument(counter, builder -> {
-            builder.append(",\"count\":").append(counter.count());
-        }));
+        return writeCounter(counter, counter.count());
     }
 
     // VisibleForTesting
     Optional<String> writeFunctionCounter(FunctionCounter counter) {
-        return Optional.of(writeDocument(counter, builder -> {
-            builder.append(",\"count\":").append(counter.count());
-        }));
+        return writeCounter(counter, counter.count());
+    }
+
+    private Optional<String> writeCounter(Meter meter, Double value) {
+        if (Double.isFinite(value)) {
+            return Optional.of(writeDocument(meter, builder -> {
+                builder.append(",\"count\":").append(value);
+            }));
+        }
+        return Optional.empty();
     }
 
     // VisibleForTesting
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
```diff
@@ -165,22 +169,27 @@ private String indexName() {
...
     // VisibleForTesting
     Optional<String> writeFunctionCounter(FunctionCounter counter) {
-        return Optional.of(writeDocument(counter, builder -> {
-            builder.append(",\"count\":").append(counter.count());
-        }));
+        return writeCounter(counter, counter.count());
+    }
+
+    private Optional<String> writeCounter(Meter meter, Double value) {
+        if (Double.isFinite(value)) {
```
izeye: @fkoehler Could counters also be NaN or infinity? If so, other meter registries which handle NaN and infinity gauges should also handle NaN and infinity counters (not in this PR).

jkschneider: `FunctionCounters` could be. Regular `Counters` could not I think because of the signature `void increment(double amount)`.

fkoehler: As @jkschneider said, I found this while looking at the kafka metrics and sending them to elastic. It happens for function counters. So other registries should check for that as well.

fkoehler:
Should I rebase this onto the 1.1.x branch to get this merged or is master ok?

jkschneider:
> Should I rebase this onto the 1.1.x branch to get this merged or is master ok?

Yes, please rebase onto 1.1.x

implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
```diff
@@ -123,23 +123,27 @@ protected void publish() {
...
+                        this::writeLongTaskTimer,
+                        this::writeTimeGauge,
+                        this::writeFunctionCounter,
+                        this::writeFunctionTimer,
+                        this::writeMeter))
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .collect(joining("\n", "", "\n"));
+
+                logger.trace("Sending metrics json to elastic: {}", jsonContent);
```
jkschneider: Could we avoid the trace call here?

fkoehler: Sure, I just though it is very helpful. With this I was able to spot the bug. We have the same trace calls/pattern i.e. in https://github.com/micrometer-metrics/micrometer/blob/2a59cd638739edcb2cb36ccc72e0a5520c158f97/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java#L117-L131 so I thought it makes sense to add it here as well.

Just let me know if to remove or not. Performance wise it should not add any overhead when tracing is disabled.

implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
izeye: I created #1138 for other meter registries.

izeye:
> I find it misleading that all counters return 0 in the assertion although we pass a value like 123.0 or NaN in. But I also could not find the place where this NoopCounter, I reckon, comes from.
> 
> Wouldn't it be easier to read and understand if we would use concrete values?

@fkoehler I think it's because `StepFunctionCounter`, not `NoopCounter`. To see the value a step needs to be passed, so the mock clock needs to be updated as follows: https://github.com/micrometer-metrics/micrometer/pull/1138/files#diff-6a01c57c7c86b0a478bbfc756612dc3bR98

I think it'd be better to update tests accordingly.

fkoehler:
> @fkoehler I think it's because `StepFunctionCounter`, not `NoopCounter`. To see the value a step needs to be passed, so the mock clock needs to be updated as follows: https://github.com/micrometer-metrics/micrometer/pull/1138/files#diff-6a01c57c7c86b0a478bbfc756612dc3bR98
> 
> I think it'd be better to update tests accordingly.

Thanks for the info. I did update the tests where appropriate and added now a test for the `Double.NaN` case.

The build fails but it's a dependency resolve error which is not related to my change

jkschneider:
Merged with 3a562260 and 40867d6660a12fb8165200499207a88298b19f0c. Thanks for the help @fkoehler!

Issue #878, opened 2018-09-26
Register Kafka "records-lag" metric per topic-partition


alexanderabramov:
Affects: master version of micrometer-core (SNAPSHOT-1.1.0-20180921.212031-73).

This is in a way related to #877.

I've been trying to get "records-lag" metric (or "records-lag-max" would have been fine too) per topic-partition to show up in /actuator/prometheus/.
I'm not sure if current KafkaConsumerMetrics is expected to register it or not, at least [KafkaConsumerMetrics.java#L236](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java#L236) hints that it's interested in per topic metrics, but currently it does not, and I think it is a very valuable metric for applications consuming more than 1 topic.

After digging around in code, I believe there are some issues with how current implementation of KafkaConsumerMetrics registers the metrics.

1. It tries to register all metrics with all combinations of tags, instead of only using existing combinations.
At [KafkaConsumerMetrics.java#L201](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java#L201) JXM NotificationListener is called first with only client-id (many times actually, but maybe that's kafka-clients issue), then with [client-id, topic], then with [client-id, topic, partition] and each time it goes through the whole list at [KafkaConsumerMetrics.java#L76](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java#L76).

For the list of metric-tags combinations actually coming from Kafka via JMX, see:
[kafka FetcherMetricsRegistry.java#L66](https://github.com/apache/kafka/blob/1.1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/FetcherMetricsRegistry.java#L66)
[kafka Fetcher.java#L1326](https://github.com/apache/kafka/blob/1.1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L1326)

2. It does add "topic" tag, but not "partition" tag, whereas e,g, "records-lag-max" is only exported by Kafka per topic-partition, not per topic. So it's not clear how meaningful "records-lag-max" per topic will be even if it does get added to Prometheus registry.
[...]

PR #1136, opened 2019-01-08
Addresses #878: Reading kafka records-lag metrics with topic and partition name

fkoehler:
I verified that this works as intended by creating a lag locally. I received the correct metrics via JMX and micrometer.

Caveat:
- I was not able to formulate a test for this, I also tried running the current tests for the KafkaConsumer and I believe they are not really testing anything. Most of the tests have no assertions and no real Kafka is being used when they run. But it might be that I miss some important parts

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
@@ -21,6 +21,8 @@ import io.micrometer.core.instrument.binder.MeterBinder;
 import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
 import io.micrometer.core.lang.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.management.*;
 import java.lang.management.ManagementFactory;
@@ -51,6 +53,9 @@ import static java.util.Collections.emptyList;
 @NonNullApi
 @NonNullFields
 public class KafkaConsumerMetrics implements MeterBinder {
+
+    private final Logger logger = LoggerFactory.getLogger(KafkaConsumerMetrics.class);
+
     private static final String JMX_DOMAIN = "kafka.consumer";
     private static final String METRIC_NAME_PREFIX = "kafka.consumer.";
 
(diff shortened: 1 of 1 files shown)


fkoehler:
@jkschneider Does this look reasonable to you?
I will publish this for myself and use it in our app via a real dashboard to verify the implementation further. Anything else I should add/change?

fkoehler:
We used this in a deployed app now and get the correct metrics in our datadog dashboard.
So the functionality seems to be there. 

Unfortunately kafka jmx metrics do not publish the consumer group but in our case we do not really need that information

jkschneider:
cc / @shakuzen
Looks reasonable to me.

shakuzen:
The changes look fine, but I would like to have some tests around the behavior this is expected to change, even if we do that after merging this. I don't have a good setup (kafka + config + consumer) to manually test this right now. In the future, we can maybe use Testcontainers to have integration tests with Kafka running.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -241,15 +250,15 @@ private void registerNotificationListener(String type, BiConsumer<ObjectName, Ta
         NotificationListener notificationListener = (notification, handback) -> {
             MBeanServerNotification mbs = (MBeanServerNotification) notification;
             ObjectName o = mbs.getMBeanName();
+            logger.trace("Got a JMX notification for object: {} with properties: {}", o.getCanonicalName(), o.getKeyPropertyListString());
```
shakuzen: We might want to guard this with a `isTraceEnabled` conditional. It looks like `getKeyPropertyListString` may not be the cheapest call to make, with the following comment in its method body:

```
// BEWARE : we rebuild the propertyliststring at each call !!
```

jkschneider: Removed the logging.

jkschneider:
> Most of the tests have no assertions and no real Kafka is being used when they run. But it might be that I miss some important parts

@fkoehler The tests do have assertions, they probably just don't look like it. `registry.get(..)...gauge()` will throw an exception if no matching meters are found.

jkschneider:
Merged with 176c71d80ecc0493fb5444e93c790a21de9be162.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
fkoehler: :) Funny, I just wanted to do the same. Thanks.

larsduelfer:
@fkoehler @jkschneider We also upgraded to micrometer 1.1.2 now but we still do not get the tags for topic and partition on the new records lag metric. While debugging I saw that the tag names are extracted though from the JMX Bean. In additiom, the kafka major version is also not determined properly. We are using a 2.x kafka client. Any idea? We will further debug this and provide more information then. Thanks!

fkoehler:
Hey Lars,

Great that you spot the Bug. So it was probably random luck that it worked
for me. I can also provide the PR if you send me a snippet or I try to
figure out the if‘s by myself.

Lars Duelfer <notifications@github.com> schrieb am Mi. 16. Jan. 2019 um
12:28:

larsduelfer:
Hi Fabian, I am already working on a fix. So far I found one that does the job but while investigating the code deeper I am looking into a better solution still. The code is not ideal for the structure of the mbeans in the kafka metrics provider IMHO. Latest tomorrow I will drop a note here about possible ways to improve this so we can discuss this. Regards, Lars

larsduelfer:
@jkschneider We (@fkoehler and myself) are currently working on a fix and will provide a pull request to get the consumer metrics working now.

larsduelfer:
@rafal-dudek we provided a pull request to fix this issue (#1166)

PR #1143, opened 2019-01-10
Set checkedForIndexTemplate to true only when succeeded

izeye:
This PR changes to set `checkedForIndexTemplate` to `true` only when the index template has been created successfully.


jkschneider:
I think the original behavior was intentional, i.e. if I can't manually add a template, see if Elastic will do it automatically for me when I send the first payload. May not have been the best idea...

izeye:
@jkschneider Thanks for the quick feedback! If it's intentional and works that way, we can close this.

PR #1158, opened 2019-01-15
"records-lead" in KafkaConsumerMetrics should not be a TimeGauge

quaff:
"records-lead" is similar to "records-lag", not time based gauge.


shakuzen:
@quaff Thank you for noticing and fixing this. This seems like a bug to me, so could you target the changes at the `1.1.x` branch instead?

quaff:
@shakuzen updated.

shakuzen:
Note: depending on the registry used, this change will break those relying on the published time series, but it was wrong from the beginning and therefore warrants breaking compatibility, IMO. For instance, Prometheus would add `_seconds` at the end of the metric name, and instead now it will add `_records` like the "records-lag" metric.

PR #1162, opened 2019-01-16
 Refactor CircleCI configuration with command

izeye:
This PR refactors CircleCI configuration with a command to reduce duplicate configuration.

This PR also upgrades to CircleCI 2.1 to use commands.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   build:
     working_directory: ~/micrometer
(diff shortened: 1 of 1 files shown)


shakuzen:
I like the reduction in duplicated code.
Here is a blog post describing the new features of the CircleCI 2.1 config: https://discuss.circleci.com/t/circleci-2-1-config-overview/26057
We need to `Enable build processing (preview)` for this to work. I'd like to test it without affecting all branches and pull requests, but I'm not sure how to do that.

shakuzen:
Given the change in #1230 and my concerns about the current "preview" status of this CircleCI feature, I think we can close this for now and revisit it in the future. Let me know if you think otherwise.

PR #1165, opened 2019-01-17
Migration of registry-cloudwatch to AWS SDK 2.x

PyvesB:
The [AWS SDK for Java 2.0](https://github.com/aws/aws-sdk-java-v2) reached general availability back in November (more information on their [official blog post](https://aws.amazon.com/blogs/developer/aws-sdk-for-java-2-x-released/)) and has de facto become the preferred way of integrating with AWS using Java.

Amongst other benefits, using it rather than the old version improves performance and allows to write cleaner code.

This pull request migrates the _registry-cloudwatch_ implementation to the new SDK.

### DIFF
--- a/implementations/micrometer-registry-cloudwatch/build.gradle
+++ b/implementations/micrometer-registry-cloudwatch/build.gradle
@@ -4,7 +4,7 @@ apply plugin: 'nebula.optional-base'
 dependencies {
     compile project(':micrometer-core')
     compile 'org.slf4j:slf4j-api:1.7.+'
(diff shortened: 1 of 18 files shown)


shakuzen:
This would break all existing users of the `CloudWatchMeterRegistry` due to the constructor change. I don't think we want to do that in a minor release. Maybe it is better to make a separate module using the new AWS SDK v2 and mark the current one as deprecated for removal in our next major version.

PyvesB:
Done. I called the new module _micrometer-registry-cloudwatch2_ to follow the same naming convention as with the Jersey implementation.

Issue #512, opened 2018-03-22
Add Kafka binder


jkschneider:
![image](https://user-images.githubusercontent.com/1697736/37796103-68285f00-2de4-11e8-9b2e-393787ae5863.png)

wardhapk:
Let me know if this works for your intent, can be extended to include other consumer level (or producer level) metrics and be added to the library. These are captured from the default Kafka metrics (org.apache.kafka.common.metrics.Metrics).

```
import io.micrometer.core.instrument.FunctionCounter;
import io.micrometer.core.instrument.MeterRegistry;
import io.micrometer.core.instrument.Tag;
import io.micrometer.core.instrument.Tags;
import io.micrometer.core.instrument.binder.MeterBinder;
import io.micrometer.core.lang.NonNullApi;
import io.micrometer.core.lang.NonNullFields;
import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
import org.springframework.context.annotation.Configuration;

import javax.management.InstanceNotFoundException;
import javax.management.MBeanServer;
import javax.management.MBeanServerDelegate;
import javax.management.MBeanServerFactory;
import javax.management.MBeanServerNotification;
import javax.management.MalformedObjectNameException;
import javax.management.NotificationFilter;
import javax.management.NotificationListener;
import javax.management.ObjectName;
import java.lang.management.ManagementFactory;
import java.util.List;
import java.util.Set;
import java.util.concurrent.Callable;
import java.util.function.BiConsumer;

import static java.util.Collections.emptyList;

@NonNullApi
@NonNullFields
@Configuration
@ConditionalOnClass(org.apache.kafka.clients.consumer.KafkaConsumer.class)
public class KafkaConsumerMetrics implements MeterBinder {

    private final MBeanServer mBeanServer;

    private final Iterable<Tag> tags;

    public KafkaConsumerMetrics() {
        this(getMBeanServer(), emptyList());
    }

    public KafkaConsumerMetrics(Iterable<Tag> tags) {
        this(getMBeanServer(), tags);
    }

    public KafkaConsumerMetrics(MBeanServer mBeanServer, Iterable<Tag> tags) {
        this.tags = tags;
        this.mBeanServer = mBeanServer;
    }

    public static MBeanServer getMBeanServer() {
        List<MBeanServer> mBeanServers = MBeanServerFactory.findMBeanServer(null);
[...]

PR #1166, opened 2019-01-17
#878: Fix kafka consumer fetch manager metrics to be reported per topic and…

larsduelfer:
… partition (#878)

The kafka consumer fetch manager metrics have been registered too many times and therewith
also for the wrong mbeans. Hence, when using prometheus as a registry, for example, only
the metrics registered for the wrong mbean were shown with NaN as values. With this fix
the metrics are registered for the right mbeans and in case they are provided for different levels
like per consumer and topic as well as per consumer, topic and partition they are only registered
once for the lowest level of details. Aggregations can be done with the monitoring tools.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
@@ -85,28 +85,32 @@ public class KafkaConsumerMetrics implements MeterBinder {
     @Override
     public void bindTo(MeterRegistry registry) {
         registerMetricsEventually("consumer-fetch-manager-metrics", (o, tags) -> {
-            registerGaugeForObject(registry, o, "records-lag", tags, "The latest lag of the partition", "records");
-            registerGaugeForObject(registry, o, "records-lag-avg", tags, "The average lag of the partition", "records");
-            registerGaugeForObject(registry, o, "records-lag-max", tags, "The maximum lag in terms of number of records for any partition in this window. An increasing value over time is your best indication that the consumer group is not keeping up with the producers.", "records");
-            registerGaugeForObject(registry, o, "fetch-size-avg", tags, "The average number of bytes fetched per request.", "bytes");
-            registerGaugeForObject(registry, o, "fetch-size-max", tags, "The maximum number of bytes fetched per request.", "bytes");
-            registerGaugeForObject(registry, o, "records-per-request-avg", tags, "The average number of records in each request.", "records");
-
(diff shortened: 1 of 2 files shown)


fkoehler:
Looks good to me.
I think we should get this one in but just reading https://github.com/micrometer-metrics/micrometer/issues/512#issuecomment-450988958 I think that's eventually the better solution as the JMX handling seems to be non intuitive and error prone

larsduelfer:
Agreed. This is for the next round of improvement once we at least have a working solution in place.

OleksiiBondar:
@fkoehler @larsduelfer Please check https://github.com/micrometer-metrics/micrometer/pull/1173 whether it sufice your needs.

larsduelfer:
@jkschneider How are we going to proceed with this PR? Will it be integrated for 1.1.3? Or do you want to wait until #1173 is finished and replace the JMX based implementation with the new one? The current KafkaConsumerMetrics bean is autowired with spring boot. In case we replace the this one with the new implementation done by @OleksiiBondar we could replace the bean and there is no change required in spring boot. In a step-wise approach, we could integrate this PR, then continue working and testing on #1173 and replace it in the next step. In this case, the alternative implementation would have to be configured manually.

Either way, it would be great if we have a working solution for the consumer metrics when 1.1.3 is released. What do you think? Thanks!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -211,10 +215,12 @@ int kafkaMajorVersion(Tags tags) {
                             String version = (String) mBeanServer.getAttribute(new ObjectName(JMX_DOMAIN + ":type=app-info,client-id=" + clientId.getValue()), "version");
                             return Integer.parseInt(version.substring(0, version.indexOf('.')));
                         } catch (Throwable e) {
-                            return -1; // should never happen
+                            // this can happen during application bootstrapping
+                            // in this case, JMX bean is not available yet and version cannot be determined yet
+                            return null;
```
shakuzen: I noticed the issue of this block being used and the result cached also, but returning `null` here will result in a NPE.

larsduelfer:
@shakuzen: I can work on this on friday, maybe also tomorrow. Regarding the NPE you mentioned: I have not noticed any issue with this so far. I am using this branch on a deployment and it works fine. However, I will check that. A generally better solution is worked on anyway in ticket #1173 so my pull request will provide a working solution for now and a code improvement will then be provided with #1173

larsduelfer:
@shakuzen @jkschneider Any news here? I rebased and fixed the mentioned problem.

jkschneider:
Thanks @larsduelfer! Who knew Kafka metrics would be so tough :)

PR #1168, opened 2019-01-17
Drop tags that do not have values before sending to Influx

### LINKED ISSUE #1167: InfluxMeterRegistry sends tags with empty values which influx then rejects
InfluxMeterRegistry simply collects up all tags for a given meter/metric and joins them together. Influx doesn't allow for tags with empty/null values and as such returns an error failing to parse the line.

```
    private String influxLineProtocol(Meter.Id id, String metricType, Stream<Field> fields) {
        String tags = getConventionTags(id).stream()
                .map(t -> "," + t.getKey() + "=" + t.getValue())
                .collect(joining(""));
```

I've run into the issue due to the project reactor integration where it defines an 'exception' tag with a default value of "" - somewhere along the line that "" becomes a null ( though either way influx would have an issue ). 

I believe that the reason for the pre-defined tag from the reactor perspective is due to prometheus requiring all the meters/metrics to have the same tags.

Given the above, I'd imagine that the most 'correct' fix would be to filter out the tags with empty/null values prior to mapping them. 

Thoughts? I need to see about both a very short term fix, as well as obviously hoping that a fix would be integrated here in the somewhat near future.

sean-brandt:
Fixes #1167 

Filter out tags with no value

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -216,6 +216,7 @@ public class InfluxMeterRegistry extends StepMeterRegistry {
 
     private String influxLineProtocol(Meter.Id id, String metricType, Stream<Field> fields) {
         String tags = getConventionTags(id).stream()
+                .filter(t -> StringUtils.isNotBlank(t.getValue()))
                 .map(t -> "," + t.getKey() + "=" + t.getValue())
                 .collect(joining(""));
 
--- a/implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryTest.java
+++ b/implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryTest.java
@@ -110,4 +110,15 @@ class InfluxMeterRegistryTest {
         assertThat(meterRegistry.writeCounter(counter.getId(), Double.NEGATIVE_INFINITY)).isEmpty();
     }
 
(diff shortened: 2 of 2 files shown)


sean-brandt:
Failure appears to be in Hibernate? Not code that I touched/changed:

```
Failures (1):
  JUnit Jupiter:HibernateMetricsTest:shouldExposeMetricsWhenStatsEnabled()
    MethodSource [className = 'io.micrometer.core.instrument.binder.jpa.HibernateMetricsTest', methodName = 'shouldExposeMetricsWhenStatsEnabled', methodParameterTypes = '']
    => org.opentest4j.AssertionFailedError: 
Expecting:
 <0.0>
to be equal to:
 <42.0>
but was not.
```

shakuzen:
> If we're going to this direction, the implementation looks good to me. It'd be good to target for the 1.0.x branch if we merge this.

I agree.
@sean-brandt thank you for the pull request and apologies on the delay in processing it. Would you be willing to rebase and target the `1.0.x` branch? You can force push that to update this pull request.

sean-brandt:
I'm looking at the rebase now, there's a lot of conflicts/changes to work through. Wonder if it'd be easier to just locally rename the branch, create a new one of 1.0.x and cherry-pick the single change across, then force push.

shakuzen:
I was working on doing the rebase but we have an interesting issue... in the 1.0.x branch, the `writeX` methods are private, so none of the tests in `InfluxMeterRegistryTest` can be written.

shakuzen:
I'm going to split it into two commits and then cherry-pick them (test commit only to 1.1.x).

shakuzen:
More fun: StringUtils doesn't exist in the 1.0.x branch :D

Issue #1096, opened 2018-12-23
Add support for KafkaStreamsMetrics (similar to KafkaConsumerMetrics)


OleksiiBondar:
It would be grate to publish Kafka streams metrics which already available through JMX beans.
List of available attributes - https://kafka.apache.org/documentation/#kafka_streams_monitoring
ttps://www.madewithtea.com/monitoring-kafka-streams-metrics-via-jmx.html

shakuzen:
Thank you everyone for the interest in this. This has been implemented by #1835 as an incubating feature in 1.4 (see the `KafkaStreamsMetrics` binder). Please try it out (available in snapshots now, soon to be released) and give feedback. With enough feedback, we can promote this to a non-incubating feature.

Issue #1095, opened 2018-12-23
Add support for KafkaProducerMetrics (similar to KafkaConsumerMetrics)


OleksiiBondar:
It would be grate to publish Kafka producer metrics which already available through JMX beans.
List of attribute names - https://docs.confluent.io/current/kafka/monitoring.html#producer-metrics

OleksiiBondar:
If no objections, I'd like to work on this improvement.

jkschneider:
Sounds great. Part of the reason we started with consumer metrics is the tagging is difficult to figure out sometimes. It's possible it will be easier for producer side metrics. See the lengthy discussion on #878 for how to infer partition tags.

larsduelfer:
@OleksiiBondar Please check the PR #1166 to fix the consumer metrics. Have you started on this yet? I can support you on this if you want.

OleksiiBondar:
@larsduelfer I've started implementing ProducerMetrics using JMX (similar to Consumer) but as a part of discussion https://github.com/micrometer-metrics/micrometer/issues/512#issuecomment-450988958 it was suggested to use more native way (using `org.apache.kafka.common.metrics.MetricsReporter` interface) which seems to be less error-prone and requires less maintanance in a future. I was able to create prototype and verify briefly consumer metrics, but haven't extend this implementation to support both producer and streams. 
I've put this activity on hold for a little bit, should get back to this within a week or so.
You can look into this if interested

mmaia:
I can try helping here with this implementation. But after reading the linked issues and discussions it's still not clear to me which is the final decision. JMX or implementing [MetricsReporter](http://people.apache.org/~nehanarkhede/kafka-0.9-producer-javadoc/doc/org/apache/kafka/common/metrics/MetricsReporter.html) directly? 
Indeed it seems nice to be possible to bypass JMX but in this case the kafka client will need to be added as an optional dependency of the micrometer core and clients will have to explicitly enable the metrics configuring their kafka clients (consumers, producers or streams). If this is ok then it looks like the best approach indeed.

shakuzen:
> I can try helping here with this implementation. But after reading the linked issues and discussions it's still not clear to me which is the final decision. JMX or implementing [MetricsReporter](http://people.apache.org/~nehanarkhede/kafka-0.9-producer-javadoc/doc/org/apache/kafka/common/metrics/MetricsReporter.html) directly?

[...]

PR #1173, opened 2019-01-22
Kafka binder via metrics reporter

OleksiiBondar:
This PR addresses both: 
https://github.com/micrometer-metrics/micrometer/issues/1095
https://github.com/micrometer-metrics/micrometer/issues/1096

For more details please check https://github.com/micrometer-metrics/micrometer/issues/512 for more details


fkoehler:
As discussed in #512 I am in favor of a purely tag based solution without extra prefixes for producer, consumer and streams. The Kafka prefix should stay. 

The reason is that when we add a prefix we loose the ability to aggregate i.e. over all bytes-sent metrics regardless if this comes from a producer or consumer. 

Secondly tags are the de facto standard to slice and dice the dimensions. 

Lastly I still like the idea of adding the additional information if it came via producer , consumer or stream in order to pregroup them. But instead of using a prefix we could also invent a new tag for that.  
In addition to that it might make sense to look into the „configure“ callback to check if there is more information worth adding via tags, i.e. the consumer Group

One other question: shouldn’t we remove metrics in the „removeMetric“ callback via the mm api?

larsduelfer:
> One other question: shouldn’t we remove metrics in the „removeMetric“ callback via the mm api?
@fkoehler I think we have to in order to remove them when consumers are rebalanced. This works with the current implementation using JMX beans and should behave the same once we have a new implementation here using the metrics reporter interface. 
@OleksiiBondar What do you think?

OleksiiBondar:
@larsduelfer Good point, I'm going to address this shortly.

OleksiiBondar:
@fkoehler I understand that there could be use cases when you want to see aggregated data points across multiple APIs. 
Since implementation provides capability to extend it, you could probably have something specific for this use case and inject into all APIs you want to inspect/harvest data from:
```
public class KafkaAggregatedApiMetrics extends AbstractKafkaMetrics {

    @Override
    public String getMetricPrefix() {
        return "kafka.";
    }

}
```
I'm happy to include this as a part of library if people are interested (also have similar requirements).

>The reason is that when we add a prefix we loose the ability to aggregate

Not sure what registry/system you are using, but few which I worked with are providing ability to represent data as formula. I.e. we can easily represent consolidated graph/dashboard/alert having value represented as (kafka.consumer.bytes.total + kafka.producer.bytes.total + kafka.streams.bytes.total). 

IMO, having prefixing mechanism has wider audience (at least in a scope of projects I've worked with) to guarantee clear separation of concerns. We used to harvest and represent/visualize data in more granular way (per API, per topic, per partition, per host, etc) rather than seeing total bytes sent/received across all APIs. Yes, it still can be achieved with having aggregated metric representation, but to me (might be very subjective) it looks little bit misleading and adds additional overhead in maintanence since requires to have very narrow list of tags and takes longer to find metric I'm particulary interested in. 

As for having tag to represent API name itself, we already have `metric.group` tag. We can probably expose new tag like `kafka.api` to represent `consumer/producer/streams` value, but for segregated metrics (`KafkaConsumerApiMetrics`, `KafkaConsumerApiMetrics`, `KafkaConsumerApiMetrics`) it looks redundant. 
Depending on where this conversation end up, there are options :)

fkoehler:
Well, I think having one solution to filter data is more appropriate. When we would add the `kafka.api` tag there is no need to prefix the metrics you can just use the tag for filtering. I would also not need to use metrics formulas to combines these metrics.  I would be interested in other opinions still.

OleksiiBondar:
@fkoehler Please check recent commit - I've tried to address both use cases (having prefixing and consolidated + api tag).
Don't think we need to go deep in debate and creating poll to vote which approach is more reasonable :) 
It's clear what you are asking for and it makes perfect sense for certain use cases.
From other side, I still have a feeling that there is considerable amount of services/developers that may benefit from having `kafka.(consumer|producer|streams)` prefixes because:
 - it's intuitively cleaner metric definition due to API segregation;
 - it's easier to migrate from existing KafkaConsumerMetrics via JMX (i.e. we have quite a few services/applications using it and we have very low interest in going through exercise of updating all dashboards/alerts to address new naming convention and make sure that no regression introduced. Also worth to mention that all historical data will be gone after switching to new naming)

Hope this addresses what you've asked for.

flozano:
Does this work on specific version of Kafka? latest Kafka versions (from 2.0 I think) has changes in how metrics are organized: https://cwiki.apache.org/confluence/display/KAFKA/KIP-272%3A+Add+API+version+tag+to+broker%27s+RequestsPerSec+metric

fkoehler:
@OleksiiBondar This looks good to me. I forgot that the current binder also prefixes the metrics with the type of api like `consumer`. I generally like the new approach and that we have the option to do both. 

Would it make sense to get this merged into the 1.1.x branch?

Let's what the maintainers have to say about this PR.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/AbstractKafkaMetrics.java
```diff
@@ -0,0 +1,105 @@
...
+            for (KafkaMetric kafkaMetric : metrics) {
+                metricChange(kafkaMetric);
+            }
+        }
+    }
+
+    public void metricChange(KafkaMetric metric) {
+        MetricName metricNameRef = metric.metricName();
+        String metricName = prepareMetricName(metricNameRef.name());
+        if (!observableMetrics.containsKey(metricName) && metric.metricValue() instanceof Double) {
```
andaag: I don't see how this would work.

this is called with metricChange(global metric), and then afterwards with per topic metrics. You'll be getting the global ones in some cases and the non global ones in other cases? Depending on which is called first (which is arguably most likely the global ones, still..)

I'd suspect the more interesting ones are the per topic metrics.

The tag's for per topic metrics and global metrics are also not the same. So it'd require some magic to deal with node-id only existing in half of the numbers reported here for example.

gysel:
Is there anything I can help with? We'd like to add KafkaProducerMetrics to our application.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/AbstractKafkaMetrics.java
shakuzen: This is indeed a problem. Some metrics won't be registered with the current implementation, and if all were registered, at least the Prometheus registry would throw an error because Prometheus does not allow different sets of tag keys for the same metric name.

shakuzen:
> Is there anything I can help with?

We need to figure out how best to solve the issues mentioned in https://github.com/micrometer-metrics/micrometer/pull/1173#discussion_r264639711 before we can integrate this.

andaag:
The way that we've solved this is ..ugly.

Whitelist of metrics we track per topic:
        "kafka.record.send.total",
        "kafka.records.consumed.total",
        "kafka.records.lag",
        "kafka.record.send.rate",
        "kafka.skipped.records.total",
        "kafka.record.size.avg",
        "kafka.record.error.total",
        "kafka.record.retry.rate",
        "kafka.records.per.request.avg",
        "kafka.record.error.rate",
        "kafka.commit.rate",
        "kafka.fetch.size.avg"

whitelist of metrics we track globally (aka if they dont have a topic tag):
        "kafka.request.size.max",
        "kafka.request.size.avg",
        "kafka.connection.count",
        "kafka.io.ratio",
        "kafka.punctuate.total"

blacklist of tags that we remove:
        "node-id", // request.size contains both metrics with and without node-id
        "processor-node-id",
        "task-id"
  
It's a really ugly solution, and it won't be reliable if upstream changes which tags/tracking they do. We have unit tests to ensure this works when kafka gets updated - which you can't really do in a generic way.

I'd suggest that the only way to get this reliable is a whitelist that whitelists both metrics / global metrics and specific tags that are allowed. Maybe a configurable one so people can add new things if the library gets updated slowly? Ugly solution, but should solve the 90% case and be reliable.

jeqo:
Hi everyone. 

I've playing with another way to solve this ([source code](https://github.com/micrometer-metrics/micrometer/compare/master...jeqo:kafka-metrics?expand=1)). 

Kafka APIs give access to metrics via `metrics()` method. Compared with the `MetricsReporter` approach attempted in this PR, we can initialize binder as others: `new KafkaConsumerMetrics(kafkaConsumer).bindTo(...)`.

@shakuzen This still has the issue with Prometheus mentioned here https://github.com/micrometer-metrics/micrometer/pull/1173#discussion_r264639711 
I have also check if removing this exception throw leads to a Prometheus library exception, but it doesn't happen; so instead of throwing, when tags sets are different I create another `MicrometricCollector`. Could this issue with prometheus client been solved in latest versions?

kgunnerud:
Any progress on this one?

shakuzen:
Sorry this slipped, but I need more time to dig into this, so it's not going to make it into 1.3. I'm moving it to the 1.4 milestone and I'll work with @jeqo on the remaining issues/changes.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/AbstractKafkaMetrics.java
cykl: Why not define `observableMetrics` as `Set<MetricName>` ?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/AbstractKafkaMetrics.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private static final Object STUB_VALUE = new Object();
+
+    // used to keep track of already registered/observable metrics to prevent registering existing metrics
+    private Map<String, Object> observableMetrics = new ConcurrentHashMap<>();
+
+    public abstract String getMetricPrefix();
+
+    public void configure(Map<String, ?> configs) {
+        // nothing needed
```
RikJansen81: Do we want the client to be able to override `METRIC_GROUP_NAME`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/AbstractKafkaMetrics.java
RikJansen81: Is this PR still hanging on this?
Is it an option to define 2 separate binders, one for _global_ and one for _topic_? Users can bind them both/individually through `metric.reporters`.

I have a feeling this issue should ideally be solved at a higher level than this. Surely, this pattern (different tags on different occasions) is not Kafka-specific?

jonatan-ivanov:
As far as I understand this PR was superseded by https://github.com/micrometer-metrics/micrometer/pull/1835 and all the issues it refers are closed now so I'm closing this, please let us know if this is incorrect and we should reopen.

PR #1179, opened 2019-01-24
Forbid instance creation for stateless utility classes

OleksiiBondar:
Define private constructors and mark class as final to protect class from extension


shakuzen:
This is technically a breaking change (if anyone was instantiating or extending these), but it is probably an omission that we didn't do this originally. I just want to confirm with @jkschneider that we want to make this change outside a major release.

jkschneider:
I agree that it's a safe bet.

PR #1183, opened 2019-01-25
Add jupiter params to all registries (inherited from :micrometer-test…

OleksiiBondar:
Add jupiter params to all registries (inherited from :micrometer-test) and add missing tests


shakuzen:
This looks like it should target the `1.0.x` branch since it is adding tests for Datadog (available in 1.0.x) and only affects tests.

OleksiiBondar:
Ok, let me rebase this change and reopen against proper branch

PR #1184, opened 2019-01-25
Remove @Disabled on BufferingFluxTest

izeye:
This PR removes `@Disabled` on `BufferingFluxTest` as there's no explanation for it and all the tests in it pass.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/BufferingFluxTest.java
+++ b/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/BufferingFluxTest.java
@@ -15,7 +15,6 @@
  */
 package io.micrometer.statsd.internal;
 
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import reactor.core.publisher.DirectProcessor;
 import reactor.core.publisher.Flux;
@@ -26,7 +25,6 @@ import java.time.Duration;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
-@Disabled
 class BufferingFluxTest {
 
     @Test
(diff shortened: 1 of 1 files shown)


izeye:
It's failing on the CI, so I guess that's why the tests are disabled. I didn't look into it yet, so if there's anyone who already knows the reason, please let me know. Otherwise, I'll look into it later.

shakuzen:
Looks like it has been disabled since the original pull request (#887). Maybe @breun knows more about why the test fails on CI?

breun:
Mysterious… I don’t remember why I would have disabled those unit tests before sending the PR. I don’t recall having issues with them in CI either, but it’s been a while.

shakuzen:
I'm looking into the test failure.

shakuzen:
The failure was caused by a race condition between the last element of the Flux passed being processed and the heartbeat flux publishing an element before ending. The failure happens (sometimes) when a heartbeat empty String is processed after the final test element is buffered. This tends to happen when the delay of the last element is a multiple of the `maxMillisecondsBetweenEmits`. With added logging the timeline looks like this for Flux elements (100ms heartbeat, 200ms delay on last element):

```
diff: 0 ms, line: twelve bytes
diff: 0 ms, line: fourteen bytes
diff: 0 ms, line: twelve bytes
diff: 104 ms, line: 
batch: [twelve bytes, fourteen bytes, twelve bytes]
diff: 100 ms, line: fourteen bytes
diff: 102 ms, line: 
batch: [, fourteen bytes]
batch: []
```

The last batch containing only a heartbeat empty String ideally shouldn't happen since the last element has already been emitted, but between emitting the last element and stopping the heartbeat Flux, a heartbeat (theoretically designed to happen at the same time if all processing time were 0) is emitted and processed.

I haven't done anything to change the race since it only potentially happens to produce one extra blank string at the very end, and instead, I opted to hopefully avoid the race with more carefully picked timings.

izeye:
@shakuzen Thanks for the analysis and fix 👍

PR #1185, opened 2019-01-25
addd unit tests for DD config and move juniper params to micrometer-t…

OleksiiBondar:
Addd unit tests for DD config and move juniper params to micrometer-test module


OleksiiBondar:
Maybe it's easier to re-open https://github.com/micrometer-metrics/micrometer/pull/1183 against master instead of adding zillion lock files which already removed in master and this change does not bring functional changes?

shakuzen:
Either way is fine with me. If someone wants to add a parameterized test to a module in any non-master branch, we're going to run into affecting a bunch of lock files eventually anyways.

OleksiiBondar:
Agree, updated locks

shakuzen:
Thanks for this. I have merged this in https://github.com/micrometer-metrics/micrometer/commit/7d28dff9a9d379dd075b8440c5b35c4ee5a6f841 with some polish (JUnit Jupiter 5.4 now offers a simpler `junit-jupiter` dependency)

PR #1203, opened 2019-02-01
Align tests' class names with their test target class names

izeye:
This PR changes to align tests' class names with their test target class names.


izeye:
@shakuzen Thanks for the quick review and merging 👍

PR #1206, opened 2019-02-01
Restore interrupted status in CloudWatchMeterRegistry.publish()

izeye:
This PR changes to restore interrupted status in `CloudWatchMeterRegistry.sendMetricData()`.


izeye:
@shakuzen Thanks for the feedback! I updated not to affect the rest of batches.

Issue #1208, opened 2019-02-01
Timer#wrap(Supplier)


chrylis:
The `Timer` class includes convenience `record` methods for several functional signatures and `wrap` for `Callable` and `Runnable`, but no `wrap(Supplier)`. It would be helpful to have the last wrapper, especially for use with Vavr and resilience4j.

PR #1213, opened 2019-02-06
Add wrap for supplier

OleksiiBondar:
Addressing request https://github.com/micrometer-metrics/micrometer/issues/1208


OleksiiBondar:
@chrylis please check if that's something you are looking for

PR #1216, opened 2019-02-07
Datadog: log when applicationKey is not configured

### LINKED ISSUE #1181: Units not being sent to datadog for Timer metrics
Right now Timer metrics report their base unit as "seconds". The datadog registry maintains a whitelist of units that datadog supports and that includes "second" (not pluralized). Since "seconds" isn't in the whitelist, no units get sent for Timer metrics.

I would guess we just want a mapping from micrometer units to datadog units.

nickmbailey:
Fixes #1181


shakuzen:
@nickmbailey thank you for this. Could you rebase and target this to the `1.0.x` branch, so we can easily merge this change forward for inclusion in all maintenance versions?

shakuzen:
I'll rebase it now.

PR #1217, opened 2019-02-07
Add documentation to 'value' property of @Timed

renannprado:


micrometer-core/src/main/java/io/micrometer/core/annotation/Timed.java
```diff
@@ -22,6 +22,9 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Inherited
 public @interface Timed {
+    /**
+     * The name of the metric
```
shakuzen: This is good. "name of the Timer" would be even more precise.

renannprado: What about "The name of the timer metric"?

shakuzen: Sounds good to me

shakuzen:
Thank you for this change @renannprado. I have cherry-picked your commit to the 1.0.x branch and merged forward.

Issue #1161, opened 2019-01-15
Incorrect time conversions in FunctionTimer via JmxReporter


C-B-B:
I've played with the FunctionTimer today, to measure the duration of an operation that affects multiple records. When exposing the metrics to JMX and reading through the jconsole, I've noticed that the values were a million time lower than what I would see when using a standard Timer (and operating on a single record at a time).
Looking into the code, I can see that we're converting from the base time unit (nanoseconds) to the registry time unit (milliseconds) in two places within the same call stack, resulting in the one million factor difference (in DropwizardFunctionTimer.Snapshot.getMean() as well as JmxReporter.JmxTimer.getMean())

[JmxReporter.JmxTimer](https://github.com/dropwizard/metrics/blob/4.1-development/metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java#L421):
```
        this.durationFactor = 1.0 / durationUnit.toNanos(1);
```
```
        public double getMean() {
            return metric.getSnapshot().getMean() * durationFactor;
        }
```
[DropwizardFunctionTimer.Snapshot](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardFunctionTimer.java#L82)
```
        public double getMean() {
            double count = count();
            return count == 0 ? 0 : totalTime(baseTimeUnit()) / count;
        }
```
```
        public double totalTime(TimeUnit unit) {
            T obj2 = ref.get();
            if (obj2 == null)
                return lastTime;
            return (lastTime = TimeUtils.convert(totalTimeFunction.applyAsDouble(obj2),
                totalTimeFunctionUnit,
                unit));
        }
```

Should the conversion in DropwizardFunctionTimer be removed? Can it?

Bit new to the group - let me know if I'm not following the right process.

Many thanks in advance,
CBB

checketts:
@C-B-B Are you supplying the correct TimeUnit where you create your function timer?

Or are you saying the unit is correct in the non-JMX representation, but the JMX one is off?

C-B-B:
@checketts, thanks for your reply.

Indeed I was careful to provide the right time unit (capturing time with System.nanoTime() and specifying Time Unit.NANOSECONDS. Also tried with milliseconds for both and got the same result - I think the issue is agnostic to that parameter).

[...]

PR #1222, opened 2019-02-09
Use nanoseconds for Dropwizard timer snapshot mean

izeye:
This PR changes to use nanoseconds for Dropwizard timer snapshot mean.

Closes gh-1161

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardFunctionTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardFunctionTimer.java
@@ -110,7 +110,8 @@ public class DropwizardFunctionTimer<T> extends AbstractMeter implements Functio
                     @Override
                     public double getMean() {
                         double count = count();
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardFunctionTimer.java
```diff
@@ -104,7 +110,7 @@ public long getMax() {
                     @Override
                     public double getMean() {
                         double count = count();
-                        return count == 0 ? 0 : totalTime(baseTimeUnit()) / count;
+                        return count == 0 ? 0 : totalTime(TimeUnit.NANOSECONDS) / count;
                     }
```
checketts: You might want to add a comment for why the hardcoded base unit is needed

izeye: @checketts Thanks for the feedback! I added a comment in c478cc41c419c5edf06c2e9c220467c48646a308 as suggested.

izeye:
I rebased onto the current 1.0.x branch and squashed.

Issue #533, opened 2018-04-03
Apache HTTP client instrumentation


shakuzen:
It looks like there is already [OkHttp instrumentation](https://github.com/micrometer-metrics/micrometer/blob/d30ca20c13e0682e6e78fa4c4c32dc75c5f9c1ac/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java) (which isn't documented as far as I can tell). It would be nice if there were also instrumentation for the [Apache HTTP client](https://hc.apache.org/). This would help achieve HTTP client instrumentation in cases where people are using the HTTP client directly (instead of through `RestTemplate` in a Spring application) or for things like the [Elasticsearch Java Low Level REST](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/6.2/java-rest-low-usage-initialization.html).

ravin-singh:
@benhubert How it's going with the testing of pool monitoring? I would like to use for one of our app in case it can be made available.

benhubert:
@ravin-singh Sorry. We are working on some other big changes at the moment, which are causing delays for this particular task. But it looks like we are able to continue with this next week, so I should be able to prepare the pull request soon.

ravin-singh:
Thanks @benhubert. For mean time I have created a workaround solution that exposed pool stats to micrometer. 
https://gist.github.com/ravin-singh/8654b654163c508cdb227a621fe078f0

shakuzen:
With the merged pull requests, I'm considering this enhancement request complete. There is now `MicrometerHttpRequestExecutor` which can be used to time requests made from an Apache `HttpClient`, and there is `PoolingHttpClientConnectionManagerMetricsBinder` which provides metrics on the connection pool for a `PoolingHttpClientConnectionManager`.

I've opened up follow-up issues to track other ideas for metrics for Apache HttpClient. See #1616 and #1617.

j-gurda:
Thank you for your contribution!

I just wonder how I may use `PoolingHttpClientConnectionManagerMetricsBinder` with HTTP Client constructed with `HttpClientBuilder`. The problem is that  `HttpClientBuilder` internally instantiates `PoolingHttpClientConnectionManager` baseing on a bunch of configuration properties (see `HttpClientBuilder::build`. Maybe I miss something but I believe it's impossible to access such `PoolingHttpClientConnectionManager` in a "civilized way" and I need to use reflection:

```java
try {
[...]

PR #1223, opened 2019-02-09
Add instrumentation for pools of HttpComponents HttpClient

benhubert:
Hi! This PR adds instrumentation for monitoring thread pools and limits of Apache HttpComponents `PoolingHttpClientConnectionManager` for #533 .

For now I just tested it locally, but we will soon use these metrics in one of our applications, where the `HttpClient` instances have to deal with very high traffic.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
@@ -201,13 +201,13 @@ public class PoolingHttpClientConnectionManagerMetricsBinder implements MeterBin
         poolRoutePendingGauge.register(routesToRows(routes, PoolStats::getPending));
     }
 
-    private Set<MultiGauge.Row> routesToRows(Set<HttpRoute> routes, Function<PoolStats, Integer> valueFunction) {
+    private Iterable<MultiGauge.Row<?>> routesToRows(Set<HttpRoute> routes, Function<PoolStats, Integer> valueFunction) {
         return routes.stream()
             .map((route) -> routeToRow(route, () -> valueFunction.apply(connectionManager.getStats(route))))
             .collect(Collectors.toSet());
     }
 
-    private MultiGauge.Row routeToRow(HttpRoute route, Supplier<Number> valueFunction) {
+    private MultiGauge.Row<Supplier<Number>> routeToRow(HttpRoute route, Supplier<Number> valueFunction) {
         Tags tags = Tags.of(
             "target.host", route.getTargetHost().getHostName(),
             "target.port", String.valueOf(route.getTargetHost().getPort()),
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
@@ -1,0 +1,202 @@
+package io.micrometer.core.instrument.binder.httpcomponents;
+
+import com.google.common.base.Preconditions;
+import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.binder.MeterBinder;
+import io.micrometer.core.lang.NonNull;
+import org.apache.http.conn.HttpClientConnectionManager;
+import org.apache.http.conn.routing.HttpRoute;
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.apache.http.pool.PoolStats;
+
+import javax.annotation.PreDestroy;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * Collects metrics from a {@link PoolingHttpClientConnectionManager}.
+ *
+ * It monitors the overall connection pool and can also be used to monitor
+ * connection pools per route.
+ *
+ * @author Benjamin Hubert (benjamin.hubert@willhaben.at)
+ * @since 1.2.0
+ */
+public class PoolingHttpClientConnectionManagerMetricsBinder implements MeterBinder {
+
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_TOTAL_MAX             = "httpcomponents.httpclient.pool.total.max";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_TOTAL_AVAILABLE       = "httpcomponents.httpclient.pool.total.available";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_TOTAL_LEASED          = "httpcomponents.httpclient.pool.total.leased";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_TOTAL_PENDING         = "httpcomponents.httpclient.pool.total.pending";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_DEFAULT_MAX_PER_ROUTE = "httpcomponents.httpclient.pool.route.max.default";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_ROUTE_MAX             = "httpcomponents.httpclient.pool.route.max";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_ROUTE_AVAILABLE       = "httpcomponents.httpclient.pool.route.available";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_ROUTE_LEASED          = "httpcomponents.httpclient.pool.route.leased";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_ROUTE_PENDING         = "httpcomponents.httpclient.pool.route.pending";
+
+    private final ScheduledExecutorService routeUpdateTask = Executors.newScheduledThreadPool(1);
+    private final PoolingHttpClientConnectionManager connectionManager;
+    private final Iterable<Tag> tags;
+    private final boolean monitorRoutes;
+
+    private MultiGauge poolRouteMaxGauge;
+    private MultiGauge poolRouteAvailableGauge;
+    private MultiGauge poolRouteLeasedGauge;
+    private MultiGauge poolRoutePendingGauge;
+
+    /**
+     * Creates a metrics binder for the given pooling connection manager.
+     * This instance will not monitor pool statistics of specific routes.
+     *
+     * @param connectionManager The connection manager to monitor.
+     * @param name Name of the connection manager. Will be added as tag with the
+     *             key "httpclient".
+     * @param tags Tags to apply to all recorded metrics. Must be an even number
+     *             of arguments representing key/value pairs of tags.
+     */
+    public PoolingHttpClientConnectionManagerMetricsBinder(HttpClientConnectionManager connectionManager, String name, String... tags) {
+        this(connectionManager, name, Tags.of(tags));
+    }
+
+    /**
+     * Creates a metrics binder for the given pooling connection manager.
+     * This instance will not monitor pool statistics of specific routes.
+     *
+     * @param connectionManager The connection manager to monitor.
+     * @param name Name of the connection manager. Will be added as tag with the
+     *             key "httpclient".
+     * @param monitorRoutes When true, this metrics binder will monitor every
+     *                      route (target scheme, host and port). Be careful:
+     *                      Set this to true if and only if you are sure that
+     *                      your service accesses only a limited number of
+     *                      target schemes, hosts and ports. DO NOT ENABLE THIS
+     *                      FEATURE if your HttpClient accesses an unlimited
+     *                      number of routes (i.e. if the target host depends on
+     *                      the user input).
+     * @param tags Tags to apply to all recorded metrics. Must be an even number
+     *             of arguments representing key/value pairs of tags.
+     */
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,176 @@
+package io.micrometer.core.instrument.binder.httpcomponents;
+
```
OleksiiBondar: Please add copyrights

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * Collects metrics from a {@link PoolingHttpClientConnectionManager}.
+ *
+ * It monitors the overall connection pool and can also be used to monitor
+ * connection pools per route.
+ *
+ * @author Benjamin Hubert (benjamin.hubert@willhaben.at)
+ * @since 1.2.0
+ */
+public class PoolingHttpClientConnectionManagerMetricsBinder implements MeterBinder {
```
OleksiiBondar: Please add unit tests

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,204 @@
...
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_TOTAL_AVAILABLE       = "httpcomponents.httpclient.pool.total.available";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_TOTAL_LEASED          = "httpcomponents.httpclient.pool.total.leased";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_TOTAL_PENDING         = "httpcomponents.httpclient.pool.total.pending";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_DEFAULT_MAX_PER_ROUTE = "httpcomponents.httpclient.pool.route.max.default";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_ROUTE_MAX             = "httpcomponents.httpclient.pool.route.max";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_ROUTE_AVAILABLE       = "httpcomponents.httpclient.pool.route.available";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_ROUTE_LEASED          = "httpcomponents.httpclient.pool.route.leased";
+    private static final String NAME_HTTPCLIENT_CONNECTION_POOL_ROUTE_PENDING         = "httpcomponents.httpclient.pool.route.pending";
```
brunocl: Wouldn't it be better to define metric name just as "httpcomponents.httpclient.pool" and let the rest be just tags? What allows users of this metric to define hierarchical metric names as they need/wish using micrometer's name mappers.
The tags would be:
total/route/max/available/leased/pending/max-default

checketts: I'm not too familiar with the connection pools, however you want to treat tags a parts of a whole. For example is available+leased+pending = total then we would use those tags and eliminate total as a tag (or keep total as a separate name). 

Using that logic is how you would decide if 'max' would belong in the same metric name as a tag. I suspect max is a high water mark, or max available so mixing it in as a tag probably wouldn't be a good idea.

benhubert: The values I named `total` are the values you get with `getTotalStats`. `total` is **not** the sum of `available` + `leased` + `pending`, but as far I understand, its values should be the sum of the corresponding `route` stats. In other words:

 * sum of all `route.max` = `total.max`
 * sum of all `route.leased` = `total.leased`
 * ...

So, it might be possible to remove `total` stats and calculate its values using all `route` stats, but as noted in line 176, you should not always enable route stats. Thus, you don't always have stats about all routes.

So, I'd personally keep the total stats.

----

Yes, `max` is the configured maximum, so a high water mark, which I would keep as separate metric name. Same for `route.max.default`.

----

Regarding tags vs. metrics for `available`, `leased` and `pending`: These numbers come from [PoolStats](https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/pool/PoolStats.html), which describes them as follows:

 * `available` ... Gets the number idle **persistent connections**.
 * `leased` ... Gets the number of **persistent connections** tracked by the connection manager currently being used to execute requests.
 * `pending` ... Gets the number of **connection requests** being blocked awaiting a free connection.

So, it might make sense to publish `available` and `leased` together in one metric with tags, because both are a count of `persistent connections`. If I get it right, `pending` does **not** count the is the number of `persistent connections`, so I think this should stay as a separate metric.

----

What do you think now? Should I change something?

checketts: Thanks for the explanations of the various metrics. The only change I would recommend is to capture your explanations in the `description` of the various gauge builders, or via comments in the code.

I like how adding `available` and `leased` as tags could express they are part of the same pool. You could also get a similar message by unifying their metrics prefixes to include 'persistent'. I would favor tags for that case, but it wouldn't give me heartburn if you didn't:

```
httpcomponents.httpclient.pool.persistent.available
httpcomponents.httpclient.pool.persistent.leased
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,222 @@
...
+        return routes.stream()
+            .map((route) -> routeToRow(route, () -> valueFunction.apply(connectionManager.getStats(route))))
+            .collect(Collectors.toSet());
+    }
+
+    private MultiGauge.Row<Supplier<Number>> routeToRow(HttpRoute route, Supplier<Number> valueFunction) {
+        Tags tags = Tags.of(
+            "target.host", route.getTargetHost().getHostName(),
+            "target.port", String.valueOf(route.getTargetHost().getPort()),
+            "target.scheme", route.getTargetHost().getSchemeName()
```
benhubert: When I tried to build a dashboard, I saw that separating the route values into three different tags might cause some difficulties. For example, let's say your client accesses the following routes:

 * `http`, `foo.example.com`, `80`
 * `http`, `bar.example.com`, `80`
 * `https`, `foo.example.com`, `443`

You might now want to create a board, which shows statistics for all three routes, meaning three combinations of the values `http`/`https` + `foo.example.com`/`bar.example.com` + `80`/`443`. But that's where I'm stuck now. Combining all tags, would also create the route `https`, `bar.example.com`, `80`, which is never really called by the application.

The initial idea of putting these values into separate tags was, to be able to create statistics for all `http` calls, or statistics for all calls to a specific host. But the more I think about it, I guess that in most cases this is not really needed, and that it would be enough to put these values just as a single `route` tag, for example:

 * `http://foo.example.com:80`
 * `http://bar.example.com:80`
 * `https://foo.example.com:443`

What do you think? Do you have an idea, how to query the original solution in Prometheus? How is it with other monitoring backends?

mweirauch: I don't really have the resulting metrics at hand, but what about:
```
sum(httpcomponents_httpclient_pool_route_connections) by (target.scheme)
... by (target.host,target.port)
```

benhubert: I want to create a Grafana board, containing a *repeating graph panel*, so one panel for each route. I can only iterate through one of the `target_*` tags, but not through all.

Do you have a solution for this?

checketts: You could create a composite tag:

* in Micrometer via a metrics filter, 
* in Prometheus as a relabeling step on collection
* in Prometheus via a recording rule

benhubert:
@mweirauch From my point of view, this PR is done since February. Do I have to do something, so that it gets merged?

mweirauch:
Hi @benhubert I currently don't have the time for review and/or testing how it "feels" from a metric point of view. Perhaps one of the maintainers or other users which also need this instrumentation can chime in.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,222 @@
...
+
+    private Iterable<MultiGauge.Row<?>> routesToRows(Set<HttpRoute> routes, Function<PoolStats, Integer> valueFunction) {
+        return routes.stream()
+            .map((route) -> routeToRow(route, () -> valueFunction.apply(connectionManager.getStats(route))))
+            .collect(Collectors.toSet());
+    }
+
+    private MultiGauge.Row<Supplier<Number>> routeToRow(HttpRoute route, Supplier<Number> valueFunction) {
+        Tags tags = Tags.of(
+            "target.host", route.getTargetHost().getHostName(),
```
shakuzen: I don't feel strongly about it, but `target` feels a little redundant since this is a tag on metrics for an HTTP client. That is, without `target` it might be clear enough what is meant, given the context. Anyone else have an opinion on it?

tgeens: With a simple `host` tag it might get confusing if you tag all (non-http-client-related) metrics with the hostname of your own service ?  For example, I have multiple instances of my service and I tag all the metrics with `host` = hostname to differentiate between my instances.

Then again, OkHttpClient instrumentation seems to be using `target` too.

shakuzen: Good point. Let's keep as is for now, then.

benhubert:
I think I don't have the permission to merge it. @shakuzen Will you merge it?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,222 @@
...
+     */
+    @SuppressWarnings("WeakerAccess")
+    public PoolingHttpClientConnectionManagerMetricsBinder(HttpClientConnectionManager connectionManager, String name, String... tags) {
+        this(connectionManager, name, Tags.of(tags));
+    }
+
+    /**
+     * Creates a metrics binder for the given pooling connection manager.
+     *
+     * For convenience this constructor will take care of casting the given
```
shakuzen: I think it would be better to ask users to cast, so it is more clear what is required here without reading the JavaDoc - that is, enforced by the compiler. What is the scenario in which you're expecting to save users from casting themselves?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,222 @@
...
+     *
+     *         &#x40;Scheduled(fixedRate = 10000)
+     *         public void updateRoutes() {
+     *             metricsBinders.forEach(PoolingHttpClientConnectionManagerMetricsBinder::updateRoutes);
+     *         }
+     *     }
+     * </pre>
+     *
+     * BE CAREFUL! Call this if and only if you are sure that your http client
+     * accesses only a limited number of remote schemes, hosts and ports. DO NOT
```
shakuzen: Spring HTTP client metrics mitigates this by capping the number of tags that will be created. Perhaps we should advise how to do the same kind of thing here. It seems like the risk is mostly if IP addresses and ephemeral ports are used, cardinality will explode.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,222 @@
...
+     * USE THIS METHOD if your HttpClient accesses a big or unlimited number of
+     * routes (i.e. if the target host depends on the user input).
```
shakuzen: Is there any way that we can capture path variables in a template? That would allow us to tag a path with less worry about tag explosion.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Collects metrics from a {@link PoolingHttpClientConnectionManager}.
+ *
+ * It monitors the overall connection pool and can also be used to monitor
+ * connection pools per route.
+ *
+ * @author Benjamin Hubert (benjamin.hubert@willhaben.at)
+ * @since 1.2.0
```
shakuzen: We'll need to update this to 1.3, but I can do this when merging.

PR #1160, opened 2019-01-15
Handle empty histogram distribution in Stackdriver registry

itzg:
The meter `jvm.gc.pause` was causing an `ArrayIndexOutOfBoundsException` when initially exporting since the `StepTimer` did not contain any histogram buckets.

The id of the meter that failed in my case was 

```
MeterId{name='jvm.gc.pause', tags=[tag(action=end of minor GC),tag(cause=Allocation Failure)]}
```

and contained a snapshot of
```
HistogramSnapshot{count=0, total=0.0, mean=0.0, max=2.0E7}
```

There are not any existing unit tests that I could update to verify my change, so I instead verified it by hand with a Spring Boot application. I verified the GCP Stackdriver metrics request included all expected entries especially the `jvm/gc/pause`:
```
time_series {
  metric {
    labels {
      key: "cause"
      value: "Allocation Failure"
    }
    labels {
      key: "action"
      value: "end of minor GC"
    }
    type: "custom.googleapis.com/jvm/gc/pause/max"
  }
  resource {
    type: "global"
    labels {
      key: "project_id"
      value: "...."
    }
  }
  metric_kind: GAUGE
  value_type: DOUBLE
  points {
    interval {
      end_time {
        seconds: 1547588380
        nanos: 376000000
      }
    }
    value {
      double_value: 29.0
    }
  }
}
```

Fixes #1001

PR #1226, opened 2019-02-12
Handle empty histogram distribution in Stackdriver registry

### LINKED ISSUE #1001: Stackdriver metrics not ending up in Stackdriver
Micrometer version: 1.1.0
Spring boot version: 5.1.0
JDK version: 11

There seem to be multiple issues with the Stackdriver integration of Micrometer:
1. Metrics are created, but are not under the Global resource it seems. (The ones with global were created using our current solution) See also the screenshot below.
missing-global
2. When debugging, the `StackdriverMeterRegistry#publish` call is just done once. With the `LoggingMeterRegistry`, the publish-method is called every interval.
3. Any way of using Spring integration with actuator?
4. Feature request: Can we use the prefix for the custom metric, so instead of `custom/jvm/buffer/etc` something like `custom/{prefix}/jvm/buffer/etc`? This makes it much easier to distinguish between services running in the same project.
5. I'm also missing the documentation right now. ;-)

See also the demo zip I have added for my sample code.
[monitoringdemo.zip](https://github.com/micrometer-metrics/micrometer/files/2576517/monitoringdemo.zip)

itzg:
The meter `jvm.gc.pause` was causing an `ArrayIndexOutOfBoundsException` when initially exporting since the `StepTimer` did not contain any histogram buckets.

This PR re-submits https://github.com/micrometer-metrics/micrometer/pull/1160 but against the 1.1.x branch.

The id of the meter that failed in my case was 

```
MeterId{name='jvm.gc.pause', tags=[tag(action=end of minor GC),tag(cause=Allocation Failure)]}
```

and contained a snapshot of
```
HistogramSnapshot{count=0, total=0.0, mean=0.0, max=2.0E7}
```

There are not any existing unit tests that I could update to verify my change, so I instead verified it by hand with a Spring Boot application. I verified the GCP Stackdriver metrics request included all expected entries especially the `jvm/gc/pause`:
```
time_series {
  metric {
    labels {
      key: "cause"
      value: "Allocation Failure"
    }
    labels {
      key: "action"
      value: "end of minor GC"
    }
    type: "custom.googleapis.com/jvm/gc/pause/max"
  }
  resource {
    type: "global"
    labels {
      key: "project_id"
      value: "...."
    }
  }
  metric_kind: GAUGE
  value_type: DOUBLE
  points {
    interval {
      end_time {
        seconds: 1547588380
        nanos: 376000000
      }
    }
    value {
      double_value: 29.0
    }
  }
}
```

Unit testing is also difficult as discussed here https://github.com/micrometer-metrics/micrometer/pull/1160#issuecomment-457350068 where it is currently not feasible to mock the Stackdriver client interactions.

Fixes #1001 

cc @shakuzen

### DIFF
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
@@ -369,6 +369,9 @@ public class StackdriverMeterRegistry extends StepMeterRegistry {
 
         private Distribution distribution(HistogramSnapshot snapshot, boolean timeDomain) {
             CountAtBucket[] histogram = snapshot.histogramCounts();
+            if (histogram.length == 0) {
+                return Distribution.newBuilder().build();
+            }
 
             // selected finite buckets (represented as a normal histogram)
             AtomicLong truncatedSum = new AtomicLong();


shakuzen:
Thank you for opening this against 1.1.x. I was testing this locally and it looks like it may have another issue. I'm getting errors like the following in the logs:

```
failed to send metrics to stackdriver: INVALID_ARGUMENT: One or more TimeSeries could not be written: Field timeSeries[2].points[0].value had an invalid value: Bucket options must be specified for Distribution metric.: timeSeries[2]; Field timeSeries[47].points[0].value had an invalid value: Bucket options must be specified for Distribution metric.: timeSeries[47]; Field timeSeries[56].points[0].value had an invalid value: Bucket options must be specified for Distribution metric.: timeSeries[56]; Field timeSeries[66].points[0].value had an invalid value: Bucket options must be specified for Distribution metric.: timeSeries[66]; Field timeSeries[68].points[0].value had an invalid value: Bucket options must be specified for Distribution metric.: timeSeries[68]
```

As mentioned in the [Stackdriver API documentation](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#Distribution), the `BucketOptions` is "Required in the Stackdriver Monitoring API v3. Defines the histogram bucket boundaries." So I think we either need to send with the one empty bucket or not send a `TimeSeries` for the Distribution when there are no buckets. I'll try some things and test locally.

shakuzen:
I fixed in a different way in 52579c2bb45f079c78a0061c7cb48dc373012d79 which didn't have the error I previously mentioned. There may still be some other issues, but we'll have to tackle those as we find them - getting some more testing of this code will help a lot, I think.

itzg:
That’s strange I didn’t see the new issue you found, but in any case I like your approach better anyway.

PR #1230, opened 2019-02-14
Simplify build configuration for JDK versions > 8

shakuzen:
This adapts the setup that the JUnit 5 build uses to achieve the same (see https://github.com/junit-team/junit5/blob/master/build.gradle.kts) Java 8 compatibility without needing a build with JDK 8.

The build continues to work with JDK 8 and 11, but Java 8 compatible artifacts can be reliably produced with just a JDK 11 build using the new Java compiler arg `--release` (see [JEP 247](https://openjdk.java.net/jeps/247)). This opens up the opportunity for us to increase our build's JDK version while maintaining compatibility with Java 8 and not increasing complexity. If we want to use Java 9+ language features in test code, that is also a possibility now, but will require all contributors to have at least that version of the JDK installed to build locally.

---

Let me know what people think of this. It feels like a better future-looking direction to go.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
     environment:
       GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
     docker:
-      - image: circleci/openjdk:8-jdk
+      - image: circleci/openjdk:11-jdk
     steps:
 
       - checkout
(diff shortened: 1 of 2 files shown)


jkschneider:
Agreed. At least simplifying the build is a good step. We can consider the impact of Java 9 features in the code when the need arises I suppose.

eleganty:
is this project able to build with jdk7?

izeye:
@eleganty No, you need JDK 8+.

eleganty:
> @eleganty No, you need JDK 8+.

OK, thank you

PR #1233, opened 2019-02-16
Remove inaccurate statement from MeterRegistry#gauge JavaDoc

### LINKED ISSUE #1176: MeterRegistry#Gauge javadoc is incorrect- states that gauges will aggregate with same id
The [javadoc for `MeterRegistry#Gauge`](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L425) states:

> If multiple gauges are registered with the same id, then the values will be aggregated and the sum will be reported. For example, registering multiple gauges for active threads in a thread pool with the same id would produce a value that is the overall number of active threads.

However, the actual behavior, it appears, is that only one of the gauges will be used to collect metrics from, and no aggregation is performed. From my quick look through the code, I found no trace of such aggregation anywhere but in the javadoc. So it would appear that it is not accurate. Moreover, the javadoc is referencing "gauge id", which is not a term used anywhere else in the javadoc of this class.

**Steps to reproduce:**

1. Create multiple gauges with the same name:
```(java)
meterRegistry.gauge("test.metric.multiple", new AtomicLong(-1));
meterRegistry.gauge("test.metric.multiple", new AtomicLong(0));
meterRegistry.gauge("test.metric.multiple", new AtomicLong(1));
```
2. Get the value for the test metric
3. Expected result: `0`
4. Actual result: `-1` (or whatever the first gauge instance's value is)

gvsmirnov:
As stated in https://github.com/micrometer-metrics/micrometer/issues/1176, the behaviour described in the javadoc never existed.

Fixes #1176

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -407,12 +407,6 @@ public abstract class MeterRegistry implements AutoCloseable {
      * Register a gauge that reports the value of the object after the function
      * {@code f} is applied. The registration will keep a weak reference to the object so it will
      * not prevent garbage collection. Applying {@code f} on the object should be thread safe.
-     * <p>
-     * If multiple gauges are registered with the same id, then the values will be aggregated and
-     * the sum will be reported. For example, registering multiple gauges for active threads in
-     * a thread pool with the same id would produce a value that is the overall number
-     * of active threads. For other behaviors, manage it on the user side and avoid multiple
-     * registrations.
      *
      * @param name          Name of the gauge being registered.
      * @param tags          Sequence of dimensions for breaking down the name.


shakuzen:
Thank you for the pull request. Would you be willing to rebase it on and target it at the `1.0.x` branch so that we can easily merge it forward for inclusion in all the next maintenance releases?

gvsmirnov:
@shakuzen looks like I did it. Full disclosure: I rarely use git, so my rebase skills are mediocre at best. Please let me know if this is indeed what you meant.

shakuzen:
> Please let me know if this is indeed what you meant.

It looks right to me. Thank you!

gvsmirnov:
Should it be merged, then?

shakuzen:
Thank you for the pull request and bearing with me @gvsmirnov. This is now merged and will be in the next release.

gvsmirnov:
Yay! Glad to help

PR #1234, opened 2019-02-17
Use Maven-compatible versions

izeye:
This PR changes to use Maven-compatible versions as the current Gradle build shows as follows:

```
...
Maven publication 'nebula' contains dependencies that will produce a pom file that cannot be consumed by a Maven client.
  - org.eclipse.jetty:jetty-server:9.2.+ declared with a Maven incompatible version notation
  - org.apache.logging.log4j:log4j-core:2.+ declared with a Maven incompatible version notation
  - io.dropwizard.metrics:metrics-core:4.0.+ declared with a Maven incompatible version notation
  - ch.qos.logback:logback-classic:1.2.+ declared with a Maven incompatible version notation
  - org.aspectj:aspectjweaver:1.8.+ declared with a Maven incompatible version notation
  - org.apache.tomcat.embed:tomcat-embed-core:8.+ declared with a Maven incompatible version notation
...
```

### DIFF
--- a/implementations/micrometer-registry-appoptics/build.gradle
+++ b/implementations/micrometer-registry-appoptics/build.gradle
@@ -1,6 +1,6 @@
 dependencies {
     compile project(':micrometer-core')
-    compile 'org.slf4j:slf4j-api:1.7.+'
+    compile 'org.slf4j:slf4j-api:[1.7,1.8)'
 
     testCompile project(':micrometer-test')
     testCompile 'org.mockito:mockito-core:latest.release'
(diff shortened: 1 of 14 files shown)


shakuzen:
I'm not sure what effect the current setup has on projects using Micrometer with Maven and I couldn't find much info in the Gradle documentation about this. Do you know what kind of problem the current usage might cause? The maven compatible versions will be more of a pain to maintain but if there's some benefit to users, it's of course worth it. Even if not, it's generally nice to avoid warnings in the build.

izeye:
@shakuzen Thanks for the feedback! I didn't look into it, either. I just noticed Gradle is printing them and wanted not to print.

jkschneider:
It should be OK I believe. IIRC the nebula-publish plugin replaces these kinds of constraints with Maven-compatible version numbers.

izeye:
@jkschneider Thanks for the feedback! I just opened this as the messages bother me when building, so we can close this if it's a false positive and the as-is is preferred.

jonatan-ivanov:
I'm closing this since it the warning has disappeared (we a new version of the plugin).

Issue #1235, opened 2019-02-17
AppOpticsMeterRegistry throws error when zero metrics are available to send


CandiedCode:
When there are zero metrics to send to appoptics, the api throws an error.

```
ERROR 2019-02-17 09:51:45,933 i.m.a.AppOpticsMeterRegistry - failed to send metrics to appoptics: {"errors":["Payload must include at least 1 measurement"]}

izeye:
@CandiedCode Thanks for the report! I created #1237 to resolve this.

PR #1237, opened 2019-02-18
Prevent publishing metrics to AppOptics when having no metrics

izeye:
This PR changes to prevent publishing metrics to AppOptics when having no metrics.

Closes gh-1235

### DIFF
--- a/implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsMeterRegistry.java
+++ b/implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsMeterRegistry.java
@@ -48,7 +48,9 @@ import static java.util.stream.Collectors.joining;
  */
 public class AppOpticsMeterRegistry extends StepMeterRegistry {
     private static final ThreadFactory DEFAULT_THREAD_FACTORY = new NamedThreadFactory("appoptics-metrics-publisher");
-    private static final String BODY_NO_MEASUREMENTS = "{\"measurements\":[]}";
+    private static final String BODY_MEASUREMENTS_PREFIX = "{\"measurements\":[";
+    private static final String BODY_MEASUREMENTS_SUFFIX = "]}";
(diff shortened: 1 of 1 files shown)


implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsMeterRegistry.java
```diff
@@ -96,24 +96,27 @@ public void start(ThreadFactory threadFactory) {
...
+                                this::writeLongTaskTimer,
+                                this::writeTimeGauge,
+                                this::writeFunctionCounter,
+                                this::writeFunctionTimer,
+                                this::writeMeter)
+                        )
+                        .filter(Optional::isPresent)
+                        .map(Optional::get)
+                        .collect(joining(",", "{\"measurements\":[", "]}"));
+                if (body.equals("{\"measurements\":[]}")) {
```
CandiedCode: @izeye  Thanks for jumping on this so quick!

does it make sense to have a Static String for this?  

```
private Static final String EMPTY_MEASURES_BODY = "{\"measurements\":[]}"; 
...
if (body.equals(EMPTY_MEASURES_BODY)) {

izeye: @CandiedCode Thanks for the feedback! I updated in d9377b8 as suggested.

shakuzen: From the bytecode perspective, it is a constant string either way. If we want to try to make sure we don't update one part and not the other (resulting in a bug), it'd be better to make the prefix and suffix constants and then use the concatenation of those in the `equals` method.

izeye: @shakuzen Sounds good to me. I pushed d90696f.

Issue #1180, opened 2019-01-24
Support for Prometheus-like collector in the Micrometer API


nahguam:
We have use cases where there are no objects to pre-bind metrics to at registration time. To use the Micrometer API we would have to build a lot of boilerplate to pre-scrape the information on a schedule so we can sync the metrics with Micrometer, constantly registering and removing metrics as object are born and die. 

Currently we use the Prometheus APIs `Collector` directly to achieve this. At measurement time we programmatically scrape metrics from the live objects and any objects that have died since the previous measurement will have no metric reported this time around. Do you plan to add support for similar functionality to Micrometer?

izeye:
This sounds like a duplicate of #1114.

nahguam:
It's not a duplicate (scrape vs expiration) but it would be a way of handling our use cases.

checketts:
It is a duplicate of #807 

Migration from Prometheus `Collector`s is the exact usecase it was  addressing. So check out `MultiGauge`. The semantics aren't identical, since collectors execute on demand when the Prometheus endpoint is called, but with a `MultiGauge` I believe you need to schedule a thread to update your gauge yourself.  I haven't used it yet in any of my code, (just reviewed the PR when it came up)

nahguam:
Great. Thanks for the pointers all!

shakuzen:
@nahguam let us know if the `MultiGauge` works for your use case. It sounds like it should. We can discuss further on Slack also, if you would like.

nahguam:
@shakuzen It seems to be doing the job. The only thing I'm not keen on is that all combinations of `MultiGauge.Row#of` leave me with compiler warnings due to the omission of the generic in the return type.

nahguam:
@shakuzen Just pushed out to our lab environment. Looks good! Code is nice and simple.

shakuzen:
@nahguam Thank you for the follow-up. I'm glad it is working well for you.

mweirauch:
Just out of curiosity (and I think I asked about this in another issue as well), wouldn't there be a need for this collector style for counters and distribution-summaries/timers as well?

PR #1240, opened 2019-02-19
Avoid using raw types for MultiGauge if possible

izeye:
This PR changes to avoid using raw types in `MultiGauge` if possible. Unfortunately there's one use internally which seems to be inevitable to me.

See https://github.com/micrometer-metrics/micrometer/issues/1180#issuecomment-457519352

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java
@@ -51,16 +51,17 @@ public class MultiGauge {
         return new Builder(name);
     }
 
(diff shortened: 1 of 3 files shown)


shakuzen:
@izeye is this intentionally targeted at `master` instead of `1.1.x`? If you think there's some risk associated to this change, I'm fine keeping it targeted at master, but it seems like it is probably fine from a quick glance.

izeye:
@shakuzen I just wanted to be safe but I didn't find any breaking change, either. Changing to target to the 1.1.x branch.

izeye:
@shakuzen Changed to target to the 1.1.x branch.

PR #1245, opened 2019-02-21
Run release script and maven central sync from CI for release tags

shakuzen:
We had the build configuration and scripts available to do this already but the CI config was not setup for it. This makes it so maintainers do not have to run the release process locally and can instead make a release/tag in GitHub, which will start the new workflow. There is a manual approval stage between publishing to Jcenter and syncing to Maven Central so release artifacts can be consumed before final sync.

---

I also wrote a wiki page to explain the release process so it is documented and easier for anyone to understand/do it. See https://github.com/micrometer-metrics/micrometer/wiki/Releasing-Micrometer

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,9 +53,21 @@ jobs:
           name: Deployment
           command: sh ./gradle/deploy.sh
 
+  maven-central-sync:
+    docker:
+      - image: circleci/openjdk:8-jdk
+    steps:
+      - checkout
+      - run:
+          name: Decrypt properties
+          command: openssl aes-256-cbc -d -in gradle.properties.enc -out gradle.properties -k $KEY -md md5
+      - deploy:
+          name: Maven Central sync
+          command: ./gradlew mavenCentralSync -Prelease.useLastTag=true -x bintrayPublish -x bintrayUpload -x bintrayCreateVersion -x bintrayCreatePackage
+
 workflows:
   version: 2
-  build_and_deploy:
+  build_prs_deploy_snapshots:
     jobs:
       - build
       - deploy:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -66,5 +75,31 @@ workflows:
...
+      - build:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v\d+\.\d+\.\d+(-rc\.\d+)?$/
+      - deploy:
+          requires:
+            - build
+          filters:
```
shakuzen: I don't know if we need to repeat this same filter in every job since they all require (directly or transitively) the `build` job which has the filter.

.circleci/config.yml
```diff
@@ -66,5 +78,31 @@ workflows:
...
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v\d+\.\d+\.\d+(-rc\.\d+)?$/
+      - staging-approval:
+          type: approval
```
shakuzen: TODO: Test what this approval job looks like in CircleCI.

shakuzen: Here are some screenshots from the deploy pipeline I'm working on for the docs site:

image

image

When you click on the approval type job, you get a pop-up like the following:
image

shakuzen: The official docs don't specify who has access to approve `approval` type jobs, but I did ask @adriancole as a test to try approving it, which seemed to do nothing. I am able to approve and cause the workflow to proceed so maybe it is limited to users on the CircleCI team for the GitHub org?
https://circleci.com/docs/2.0/workflows/#holding-a-workflow-for-a-manual-approval

shakuzen: see https://circleci.com/workflow-run/d0c4004c-0334-4237-83ac-1f5f385d881c

Issue #1254, opened 2019-03-02
AsyncCache support (caffeine 2.7)


ben-manes:
Micrometer provides convenient bindings for `Cache`, `LoadingCache`, and `AsyncLoadingCache`. `AsyncCache` was recently split off from `AsyncLoadingCache` based on integrations where the mapping function isn't provided up front (e.g. Akka's http cache) or for those who need to capture non-key parameters in the given lambda. These behaviors were supported by `AsyncLoadingCache` but a little clunky.

I believe all that is required is changing `CaffeineCacheMetrics` from
```java
public static <C extends AsyncLoadingCache> C monitor(...) {}
```
to
```java
public static <C extends AsyncCache> C monitor(...) {}
```

Due to the library maintaining strict semver, it should be safe to upgrade the required version clients use.

izeye:
@ben-manes Thanks for the suggestion! I created #1255 to try to resolve this as suggested.

shakuzen:
We'll want to note in our 1.2.0 release notes that Caffeine 2.7 or later is now required for the Caffeine caching metrics support.

PR #1255, opened 2019-03-02
Support AsyncCache in CaffeineCacheMetrics

izeye:
This PR changes to support `AsyncCache` in `CaffeineCacheMetrics`.

Closes gh-1254


izeye:
This change will break compatibility with Caffeine prior versions to 2.7.0.

PR #1258, opened 2019-03-03
Fix conditional on beans for HibernateMetricsAutoConfiguration

izeye:
This PR fixes `@ConditionalOnBean` usage for `HibernateMetricsAutoConfiguration` as elements in `@ConditionalOnBean` `value` array attribute are evaluated with "or" in Spring Boot 1.5 or prior whereas they are evaluated with "and" since 2.0.


shakuzen:
Reference note from the Spring Boot 2.0 migration guide: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#conditionalonbean-semantic-change

PR #1260, opened 2019-03-04
Start line sink only when enabled in StatsD meter registry

izeye:
This PR changes to start line sink only when enabled in StatsD meter registry.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -50,7 +50,21 @@ import java.util.function.ToDoubleFunction;
 import java.util.function.ToLongFunction;
 
 /**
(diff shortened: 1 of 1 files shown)


izeye:
For reviews, see https://github.com/micrometer-metrics/micrometer/pull/1260/files?w=1

shakuzen:
I decided to write a test for this (see e258b0d4629b9c3c988dde9ad11a248420dfa70b) and then realized that the line sink wouldn't actually be called before either due to

```java
@Override
public void onNext(String line) {
    if (started.get()) {
        lineSink.accept(line);
    }
}
```

The test is actually failing now due to the assertion that the queue gets drained. When the registry is disabled, it seems the queue will continue to grow since nothing is subscribed. This is a (big) problem.

izeye:
@shakuzen Thanks for adding the test and spotting the problem! Are you working on it? Otherwise, I can try to resolve it.

izeye:
@shakuzen I created #1263 to try to resolve it just in case. If you already have a fix for it locally, feel free to close it.

PR #1266, opened 2019-03-08
Add a test for multiple registrations of the same name of gauge

izeye:
This PR adds a test for behavior on multiple registrations of the same name of gauge.

See gh-1176


izeye:
@shakuzen Thanks for the feedback! I created https://github.com/micrometer-metrics/micrometer/pull/1273 to add Javadoc.

Issue #1268, opened 2019-03-08
Fails to push metrics to Stackdriver with short step interval and large batch size


GlebMendrul:
Hi, I'm using micrometer version 1.1.3 and spring boot 2.1.0.RELEASE. StackdriverMeterRegistry is defined as a spring bean :

```
@Bean
  public static StackdriverMeterRegistry stackdriver() {
    return StackdriverMeterRegistry.builder(new StackdriverConfig() {
      @Override
      public String projectId() {
        return "test-project";
      }

      @Override
      public String get(String key) {
        return null;
      }

      @Override
      public Duration step() {
        return Duration.ofSeconds(5);
      }
    }).build();
  }
```

However, when I ran the app on GCP I keep seeing the following warning without any stacktrace : 
`failed to send metrics to stackdriver: INTERNAL: http2 exception`

It's strange because I can push metrics from the same application and environment using just google-cloud-monitoring library: 
https://cloud.google.com/monitoring/docs/reference/libraries#client-libraries-usage-java

Could you please let me know if I'm missing something in configuration etc.

Thanks,
Gleb

izeye:
@mglebka Thanks for the report! I didn't look into this but no stack trace is not aligned with other meter registries, so I created #1269 to resolve it.

izeye:
I reproduced the error with [a sample](https://github.com/izeye/sample-micrometer-spring-boot/tree/stackdriver) and the stack trace was as follows:

```
2019-03-09 05:23:56.139  WARN 82634 --- [trics-publisher] i.m.s.StackdriverMeterRegistry           : failed to send metrics to stackdriver

com.google.api.gax.rpc.InternalException: io.grpc.StatusRuntimeException: INTERNAL: http2 exception
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:67) ~[gax-1.34.0.jar:1.34.0]
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:72) ~[gax-grpc-1.34.0.jar:1.34.0]
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:60) ~[gax-grpc-1.34.0.jar:1.34.0]
[...]

PR #1269, opened 2019-03-08
Include stack trace when failed to send metrics to Stackdriver

izeye:
This PR changes to include stack trace when failed to send metrics to Stackdriver. This change is aligned with other meter registries.

See gh-1268


mglebka:
@izeye , thanks for this merge request, could you please also print stacktrace here https://github.com/micrometer-metrics/micrometer/blob/348346a9e3665f45f5404a22583157eec8600f66/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java#L357

izeye:
@mglebka Thanks for the feedback! Sounds reasonable and it's aligned with Dynatrace meter registry at least. I updated as suggested in 43fc64b.

izeye:
I also noticed there's no logging for successful batch publishing, so I added it in 619a6b3.

Issue #417, opened 2018-02-12
Use specific exceptions for configuration problems


wilkinsona:
#409 made the various registries fail fast when a configuration problem is detected. If a custom exception was used (rather than `IllegalStateException`), we could take this a step further in Spring Boot and provide a `FailureAnalyzer` ([introduced in 1.4](https://spring.io/blog/2016/07/28/spring-boot-1-4-released#startup-failure-message)) that provides some user-friendly guidance on how to fix the problem and avoid showing the user a stacktrace that isn't particularly useful.

A quick look at the code suggests there's room for two exceptions:

- One for some missing configuration
- One for some configuration that's invalid

There may well be room for more, but the more specific the exception the easier it'll be for Boot the failure analysis without having to resort to exception message parsing.

jkschneider:
* One for some missing configuration - `MissingRequiredConfigurationException`
* One for some configuration that's invalid - `InvalidConfigurationException`

wilkinsona:
Thanks!

PR #1270, opened 2019-03-08
Consistently throw MissingRequiredConfigurationException from MeterRegistry implementations

izeye:
This PR changes to fail fast when project ID which is mandatory is missing in Stackdriver meter registry.

### DIFF
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
@@ -53,6 +53,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toCollection;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.StreamSupport.stream;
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
```diff
@@ -87,6 +88,8 @@ public StackdriverMeterRegistry(StackdriverConfig config, Clock clock) {
     private StackdriverMeterRegistry(StackdriverConfig config, Clock clock, ThreadFactory threadFactory,
                                      Callable<MetricServiceSettings> metricServiceSettings) {
         super(config, clock);
+        requireNonNull(config.projectId());
```
shakuzen: We should probably use the `MissingRequiredConfigurationException` class we have like at least `WavefrontMeterRegistry` does for #417.

izeye: @shakuzen Thanks for the feedback! I created https://github.com/micrometer-metrics/micrometer/pull/1277 to resolve it against the 1.0.x branch. I'll push another commit on this PR for the 1.1.x branch.

izeye: @shakuzen I pushed e59137b.

shakuzen:
I think we'll need some changes (at least in tests) in Spring Boot now that some repositories fail fast on required properties that were not before. I have applied similar changes to the remaining repositories in 1.1.x that have required properties.

PR #1244, opened 2019-02-20
Added an integration of ElasticSearch ingest pipelines

cambierr:
Changes:

- Added an integration of ElasticSearch [ingest pipelines](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/pipeline.html) to be able to pre-processes micrometer metrics
- Added a way to customize the separator between the index name and the date part (not all of us are using `-`)

PR #1272, opened 2019-03-10
Polish Elasticsearch ingest pipeline and index separator support

izeye:
This PR polishes Elasticsearch ingest pipeline and index separator support.

This PR also adds a test showing that gh-987 is supported now.

See gh-1244
Closes gh-987
Closes gh-988


izeye:
As a side note, I tempted to merge `indexDateFormat` and `indexDateSeparator` into one like `indexDateSuffixFormat`.

shakuzen:
> As a side note, I tempted to merge `indexDateFormat` and `indexDateSeparator` into one like `indexDateSuffixFormat`.

Yea, I'm not thrilled with the current options either. We should think about what the better properties are now that we have more feedback from ElasticMeterRegistry users, and then how we can migrate to those (deprecation/co-existing current and new property).

PR #1275, opened 2019-03-12
Add continuation indent size to .editorconfig

izeye:
This PR adds continuation indent size for Java sources to the `.editorconfig` file.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,3 +3,4 @@ root = true
 [*.java]
 indent_style = space
 indent_size = 4
+continuation_indent_size = 8


shakuzen:
I'm not sure how widely supported the property is, but it at least seems it is supported by IntelliJ: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#ideas-for-domain-specific-properties

izeye:
@shakuzen Thanks! FTR It's also working with GitHub for code view.

shakuzen:
After I merged this, I noticed that IntelliJ complains about it being an unknown key. I'm not sure if the plugin is just not aware of the key or if it doesn't work in IntelliJ. Either way, not much harm in it being there as long as it helps somewhere.

izeye:
> FTR It's also working with GitHub for code view.

@shakuzen Sorry. I was confused with `indent_size` when commenting. I only checked it's working in IntelliJ.

PR #1277, opened 2019-03-12
Use MissingRequiredConfigurationException consistently for 1.0.x

izeye:
This PR changes to use `MissingRequiredConfigurationException` consistently for the 1.0.x branch.

See https://github.com/micrometer-metrics/micrometer/pull/1270#discussion_r264550810


shakuzen:
@izeye After merging this, I noticed that all of those implementations throw the `MissingRequiredConfigurationException` from the *Config class. I'm not sure which is better but it feels like overkill to have it thrown in both places, and IntelliJ even warns that the getter method never returns `null`.

izeye:
@shakuzen We can subclass them, so I think the checks are valid. Am I missing something?

shakuzen:
> We can subclass them, so I think the checks are valid.

That's a good point. I mainly mean that things are again inconsistent as not all Config classes throw `MissingRequiredConfigurationException`.

izeye:
@shakuzen Did you find any missing handling for `MissingRequiredConfigurationException` in any configuration class?

shakuzen:
> Did you find any missing handling for `MissingRequiredConfigurationException` in any configuration class?

Now that I look more closely, I cannot find any. Ignore me about that, then.

PR #1279, opened 2019-03-12
Time method calls within a class when class is annotated with @Timed

### LINKED ISSUE #493: Be able to target classes and not only methods with TimedAspect

sabouri:
Assuming that when a class is annotated @Timed, the expected behavior is to measure all methods of the class, this PR adds an advice to perform just that.

Resolves #493

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -150,6 +150,15 @@ public class TimedAspect {
         this.shouldSkip = shouldSkip;
     }
 
+    @Around("@within(io.micrometer.core.annotation.Timed)")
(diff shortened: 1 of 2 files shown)


themistoklik:
any news/updates if this is going to be merged?

makowskimateusz:
@shakuzen @izeye What do you think about it? Are you planning to pull-in this? Would be super useful for me and not only for me probably :)

MaxZhuravsky:
any update on this one?

makobernal:
Hello all. Is there any chance this PR can make it into the branch? I find it extremely useful, and it's much less confusing to make `@Timed` apply to classes as well, instead of methods only.

shakuzen (approved):
Looks good to me. Thanks @sabouri. Would you be willing to sign the CLA so we can accept your contribution?

Issue #1282, opened 2019-03-13
Multiplied value of metric when using CompositeMeterRegistry


rafal-dudek:
Hi,

In our application we have 2 registries: PrometheusMeterRegistry and JmxMeterRegistry.
There is created automatically CompositeMeterRegistry.

Some metrics are doubled e.g. "logback.events" (io.micrometer.core.instrument.binder.logging.LogbackMetrics) and "jvm.gc.pause" (io.micrometer.core.instrument.binder.jvm.JvmGcMetrics).

It looks like MeterBinder is binded to all 3 MeterRegistries -> counters in PrometheusMeterRegistry  and JmxMeterRegistry are incremented twice.

It is similar to issue: https://github.com/micrometer-metrics/micrometer/issues/546 

We are using spring-boot-actuator-autoconfigure:2.1.3.RELEASE, micrometer-registry-prometheus:1.1.3 and micrometer-registry-jmx:1.1.3.
[demo.zip](https://github.com/micrometer-metrics/micrometer/files/2960474/demo.zip)

izeye:
@rafal-dudek Thanks for the report! I didn't look into `JvmGcMetrics` yet but confirmed `LogbackMetrics` has an issue with `CompositeMeterRegistry`. I created #1283 to try to resolve it.

For `JvmGcMetrics` I suspect there's the same problem if any. I'll look into it.

izeye:
This is a general issue with `Counter` as all `MeterBinder`s are bound to all `MeterRegistry`s including `CompositeMeterRegistry` in Spring Boot.

With a quick search the following `MeterBinder`s seem to be affected:

- `MicrometerMetricsPublisherCommand`
- `JvmGcMetrics`
- `Log4j2Metrics`
- `LogbackMetrics`

I can fix them like #1283 but I think we need a better fix as `MeterBinder` can be defined outside of this repo although I don't know how at the moment.

izeye:
I created https://github.com/spring-projects/spring-boot/pull/16221 to try resolve this on Spring Boot side. If the PR got merged, we can close #1283.

shakuzen:
> This is a general issue with `Counter` as all `MeterBinder`s are bound to all `MeterRegistry`s including `CompositeMeterRegistry` in Spring Boot.

[...]

PR #1283, opened 2019-03-13
Change CompositeMeterRegistry not to increment counters for Logback

izeye:
This PR changes `CompositeMeterRegistry` not to increment counters for Logback.

See gh-1282

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/LogbackMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/LogbackMetricsTest.java
@@ -23,6 +23,7 @@ import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.MockClock;
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
 import io.micrometer.core.instrument.cumulative.CumulativeCounter;
 import io.micrometer.core.instrument.simple.SimpleConfig;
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/LogbackMetricsTest.java
```diff
@@ -86,6 +93,22 @@ void removeFilterFromLoggerContextOnClose() {
...
 
+    @Issue("#1282")
+    @Test
+    void compositeMeterRegistryShouldNotIncrementCounter() {
+        SimpleMeterRegistry simpleMeterRegistry = new SimpleMeterRegistry();
+        CompositeMeterRegistry compositeMeterRegistry = new CompositeMeterRegistry();
+        compositeMeterRegistry.add(simpleMeterRegistry);
+
+        new LogbackMetrics().bindTo(compositeMeterRegistry);
+        new LogbackMetrics().bindTo(simpleMeterRegistry);
```
checketts: Why do we bind to each registry? It seems like it should just be binding to the composite registry.

izeye: @checketts Creating `LogbackMetrics` twice is a mistake but applying `MeterBinder`s to all meter registries including composite meter registries is the current implementation in Spring Boot.

I think you're right if there's composite meter registry but if there isn't, we need to bind meters to all meter registries.

See https://github.com/micrometer-metrics/micrometer/issues/1282#issuecomment-472375431

shakuzen:
I'll close this until we reach a resolution on the related Spring Boot change. We can reopen this if appropriate then.

PR #1289, opened 2019-03-19
Log payload for failed metric posts

grabnerandi:
We have seen "failed to send metrics to dynatrace" errors with not enough indication on the actual root cause. Logging the actual requestPayload will give us more troubleshooting insights in case this problem happens again

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
@@ -246,7 +246,7 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
                                     postMessage.metricCount, postMessage.payload.getBytes(UTF_8).length);
                             }
                         })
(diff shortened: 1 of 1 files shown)


shakuzen:
I think logging the whole request payload at ERROR log level for an API call failure is too verbose. Perhaps you are running into https://github.com/micrometer-metrics/micrometer/issues/1343. I have opened #1357 instead, which I'll get included in the soon to be released 1.1.4.

danieljamesscott:
Out of interest. Why is logging the request payload too verbose? Surely there should not be that many request failures in normal operation. Also, the failure may be due to invalid request data so being able to see what was attempted is useful.

shakuzen:
> Out of interest. Why is logging the request payload too verbose?

The request payload is expected to be quite large in many cases, and it's not even necessarily helpful. For example, if the target host is down, knowing the request payload is not helpful, and the exception message given should indicate the problem (e.g. host is unreachable). It's fine to make it possible to get the request payload on error when you cannot figure out the problem from the exception message and the request payload would be useful - that's why we opted to log it at DEBUG level. It doesn't flood logs with huge payloads by default (assuming INFO level logging is default), but logging can be configured for those that want to see the payload.

Issue #1291, opened 2019-03-19
LoggingMeterRegistry fails to print Meter measurements


zeratul021:
LoggerMeterRegistry prints ~ **java.util.stream.ReferencePipeline$3@515211d3** instead of concatenated Measurement values due to missing collector at the end of the stream.

I can provide a PR if needed.

checketts:
Good catch! A PR would be great!

PR #1297, opened 2019-03-21
#1291 - expanded printing of Meter measurements

zeratul021:
PR for issue #1291 

I attempted to guess which Statistics would be candidates for value handling with baseUnit.
I didn't include ACTIVE_TASKS even though it's used in that way on line 136. Not sure if that's intended. 
Also not sure about COUNTER as it's a rate in this context. Let me know WDYT.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
@@ -152,11 +152,26 @@ public class LoggingMeterRegistry extends StepMeterRegistry {
                                     loggingSink.accept(print.id() + " throughput=" + print.rate(count) +
                                             " mean=" + print.time(timer.mean(getBaseTimeUnit())));
                                 },
-                                meter -> loggingSink.accept(print.id() + StreamSupport.stream(meter.measure().spliterator(), false)
-                                        .map(ms -> ms.getStatistic().getTagValueRepresentation() + "=" + decimalOrNan(ms.getValue())))
+                                meter -> loggingSink.accept(writeMeter(meter, print))
                         );
                     });
         }
+    }
+
+    private String writeMeter(Meter meter, Printer print) {
+        return print.id() + StreamSupport.stream(meter.measure().spliterator(), false)
+                .map(ms -> {
+                    String msLine = ms.getStatistic().getTagValueRepresentation() + "=";
+                    switch (ms.getStatistic()) {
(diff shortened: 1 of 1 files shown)


zeratul021:
I attempted to guess which Statistics would be candidates for value handling with baseUnit.
I didn't include ACTIVE_TASKS even though it's used in that way on line 136. Not sure if that's intended. Let me know WDYT.

shakuzen:
Thank you for catching this bug and sending a pull request to fix it.
Would you update the pull request to target the `1.1.x` branch so we can include this in the upcoming 1.1.4 release? I can more easily merge the change forward to master than I can backwards to a maintenance branch.

> I attempted to guess which Statistics would be candidates for value handling with baseUnit.

Do we need to do this? It looks like using `print.value()` will check the baseUnits and only add them if they are not null anyways, so wouldn't it be easiest to just unconditionally use that?

https://github.com/micrometer-metrics/micrometer/blob/a3e313cc327d2058b5e0bd671e4257d93ea90356/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java#L199-L201

https://github.com/micrometer-metrics/micrometer/blob/a3e313cc327d2058b5e0bd671e4257d93ea90356/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java#L212-L218

micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
```diff
@@ -152,13 +152,28 @@ protected void publish() {
...
+                    switch (ms.getStatistic()) {
+                        case TOTAL:
+                        case MAX:
+                        case VALUE:
+                            return msLine + print.value(ms.getValue());
+                        default:
+                            return msLine + decimalOrNan(ms.getValue());
+                    }
+                })
+                .collect(joining(" ", " ", ""));
```
shakuzen: Perhaps a comma and space (`, `) would more clearly delineate measurements. What do you think?
Could you also share a whole line output for a `Meter` after this change? Or adding a test would be even better but not required.

zeratul021:
```suggestion
                .collect(joining(", ", " ", ""));
```

zeratul021: Sure!

PR #1303, opened 2019-03-21
Added OSHI metrics support

### LINKED ISSUE #1274: Improve system metrics with oshi
https://github.com/oshi/oshi

OSHI is a free JNA-based (native) Operating System and Hardware Information library for Java. It does not require the installation of any additional native libraries and aims to provide a cross-platform implementation to retrieve system information, such as OS version, processes, memory & CPU usage, disks & partitions, devices, sensors, etc.

Merge (with some charges) https://github.com/jrask/micrometer-oshi-binder

@jkschneider - I thought we could keep the discussion here.

jrask:
@checketts @jkschneider 

First version of OSHI metrics. This PR is incomplete since it simply delegates to the "old" jvm version always.

What I want to discuss is:

- metric naming, can we be backwards compatible by swapping jvm with oshi and only adding metrics and not removing any. Or should we use simply new names and always run both in parallell?

- (if not in parallell) We need to decide which to use depending on OS, I have not yet tried on windows only on mac and linux. Not sure if a first version would be ok to use oshi only on linux and mac until windows is tested. I know people run java on windows servers I am just not aware of any ;-)

- I guess I need to have some sanity tests as well.

Resolves #1274

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -28,6 +28,10 @@ dependencies {
     compile 'com.hazelcast:hazelcast:latest.release', optional
     compile 'org.hibernate:hibernate-entitymanager:latest.release', optional
 
+    // oshi
+    compile 'com.github.oshi:oshi-core:3.13.0', optional
+
+
     // server runtime monitoring
     // 9.4+ are incompatible with the latest version of wiremock as of 1/4/2019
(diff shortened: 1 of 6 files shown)


shakuzen:
Closing since there is now support in OSHI: https://github.com/oshi/oshi/tree/master/oshi-metrics

PR #1307, opened 2019-03-22
Upgrade to Checkstyle 8.18

izeye:
This PR upgrades to Checkstyle 8.18.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,7 @@ subprojects {
     }
 
     checkstyle {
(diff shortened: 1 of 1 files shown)


shakuzen:
I think since Checkstyle is a dependency only used during the build, we can upgrade this in all maintenance lines. If you agree, would you change to target `1.0.x`?

izeye:
@shakuzen Sure! Done.

PR #1310, opened 2019-03-23
Use injected clock for auto-configured Stackdriver meter registry

izeye:
This PR changes to use the injected clock for the auto-configured Stackdriver meter registry.


shakuzen:
Does Spring Boot not have auto-config support for Stackdriver yet?

izeye:
> Does Spring Boot not have auto-config support for Stackdriver yet?

@shakuzen [It's marked as incubating](https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java#L61), so I'm not sure it's ready for it at the moment.

izeye:
@shakuzen Thanks for the feedback! I added tests in 5d962f5.

PR #1312, opened 2019-03-25
Add Infinity Buckets to the StatsD MeterRegistry for Meters with an SLA

mpoggy:
I'm a first-time contributor and don't write Java daily, so I'm more than happy to change anything here in favor of a more idiomatic solution, just let me know! I've ensured all existing tests pass while adding a few new ones.

### Problem
The rate at which StatsD reports metrics and expiration time on SLA histogram age buckets is not aligned. This can cause odd values to be observed on monitoring systems e.g. decimal values being reported from a single host or the SLA count exceeding the total count. One of the primary use cases for the SLA feature is to report on violations, this change aims to make that easier.

### Solution
This was discussed briefly in the Micrometer Slack channel. Aligning the registry's reporting time with the distribution statistic configuration isn't straightforward. So, instead we'll add `+Inf` (`Long.MAX_VALUE`) buckets to the underlying histograms on our user-defined buckets. That way we'll be able to calculate violations using the following - 

```
metric{le:+Inf} - metric{le:$sla}
```

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -24,10 +24,7 @@ import io.micrometer.core.instrument.distribution.pause.PauseDetector;
 import io.micrometer.core.instrument.internal.DefaultMeter;
 import io.micrometer.core.instrument.util.HierarchicalNameMapper;
 import io.micrometer.core.lang.Nullable;
-import io.micrometer.statsd.internal.DatadogStatsdLineBuilder;
-import io.micrometer.statsd.internal.EtsyStatsdLineBuilder;
-import io.micrometer.statsd.internal.LogbackMetricsSuppressingUnicastProcessor;
-import io.micrometer.statsd.internal.TelegrafStatsdLineBuilder;
+import io.micrometer.statsd.internal.*;
 import org.reactivestreams.Processor;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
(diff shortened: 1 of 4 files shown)


jkschneider:
@shakuzen Interested in your input. I'm thinking we might want to backport this to 1.0.x. Because of the impossibility to align the StatsD agent's publishing interval with Micrometer's, I think the lack of a +Inf bucket makes StatsD SLAs unusable.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryTest.java
```diff
@@ -312,17 +312,52 @@ void slasOnlyNoPercentileHistogram(StatsdFlavor flavor) {
...
+
+        assertThat(registry.get("my.timer.histogram").tag("le", "+Inf").gauge()).isNotNull();
+    }
+
+    @Test
+    void distributionSummariesWithSlasHaveInfBucket() {
+        StatsdMeterRegistry registry = new StatsdMeterRegistry(configWithFlavor(StatsdFlavor.ETSY), clock);
+        DistributionSummary summary = DistributionSummary.builder("my.distribution").sla(1).register(registry);
+
+        assertThat(registry.get("my.distribution.histogram").tag("le", "+Inf").gauge()).isNotNull();
```
jkschneider: Can remove the `assertThat` because `get` will throw a detailed exception if a matching gauge isn't found.

mpoggy: Done. Thanks!

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -25,6 +25,7 @@
 import io.micrometer.core.instrument.util.HierarchicalNameMapper;
 import io.micrometer.core.lang.Nullable;
 import io.micrometer.statsd.internal.*;
+import org.apache.commons.lang.ArrayUtils;
```
shakuzen: We don't want to make a hard dependency on Apache Commons, but I can fix this while merging.

PR #1315, opened 2019-03-25
Wavefront cumulative values

hanwavefront:
Updates to the Wavefront integration:

- Send cumulative values instead of aggregated rates, since the Wavefront query language provides `rate()` function.
- Update baseTimeUnit to be milliseconds instead of seconds.

### DIFF
--- a/implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
+++ b/implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
@@ -16,7 +16,7 @@
 package io.micrometer.wavefront;
 
 import io.micrometer.core.instrument.config.MissingRequiredConfigurationException;
-import io.micrometer.core.instrument.step.StepRegistryConfig;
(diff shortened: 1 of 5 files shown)


implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
```diff
@@ -31,7 +31,7 @@
  * @author Jon Schneider
  * @since 1.0.0
  */
-public interface WavefrontConfig extends StepRegistryConfig {
+public interface WavefrontConfig extends PushRegistryConfig {
```
shakuzen: @hanwavefront I'm curious what was the motivation for changing this from Step to Push? This is ending up breaking some tests in Spring Boot.

shakuzen: Sorry for the silly question. After refreshing my memory on this change, it all makes sense now. The config type changed to match the change to the MeterRegistry type, in order to stop shipping rate-normalized metrics.

PR #1328, opened 2019-03-26
Make the document type configurable when writing to ES index

### LINKED ISSUE #1326: Make document type configurable in ElasticMeterRegistry
Currently the document type of the metrics is hardcoded to be `doc` in `ElasticMeterRegistry`. It would be nice (in our case even necessary) to make type configurable.

malex211:
Resolves #1326


shakuzen:
Could you explain the motivation for making this configurable? Is this related to or affected by the ["removal of mapping types"](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html)?

Issue #1340, opened 2019-03-28
[Stackdriver] A maximum of 200 TimeSeries can be written in a single request


cbismuth:
Micrometer 1.1.3
Spring Boot 2.1.3.RELEASE

Hi,

We're encountering this quotas-related error below, should this be handled by Micrometer or by ourselves?

Thanks,
Christophe

Error message:

```
2019-03-28 14:53:01.000  WARN 11949 --- [stackdriver-metrics-publisher] i.m.s.StackdriverMeterRegistry           : failed to send metrics to stackdriver: INVALID_ARGUMENT: Field timeSeries had an invalid value: A maximum of 200 TimeSeries can be written in a single request.
```

Code sample:

```java
    @KafkaListener(topics = {
        "${kolsquare.ingest.kafka.topic.post.batch.name}",
        "${kolsquare.ingest.kafka.topic.post.stream.name}"
    })
    public void listen(final List<String> payloads) {
        try {
            Metrics.summary("ingest.loader.listener.size.summary").record(payloads.size());

            final List<Post> posts = resolvePosts(payloads);
            postRepository.updatePosts(posts);

            final Set<User> users = resolveUsers(posts);
            userRepository.updateUsers(users);

            Metrics.counter("ingest.loader.listener.success.counter").increment();
        } catch (final Exception e) {
            LOGGER.error(e.getMessage(), e);

            Metrics.counter("ingest.loader.listener.failure.counter").increment();

            throw e;
        } finally {
            Metrics.counter("ingest.loader.listener.total.counter").increment();
        }
    }
```

References:
* https://cloud.google.com/monitoring/quotas

cbismuth:
I'll try to override the default batch size which is 10 000.

```java
@Component
class StackdriverProperties extends StepRegistryProperties {

    private final String projectId;

    StackdriverProperties(@Value("${kolsquare.ingest.google.projectId}") final String projectId) {
        this.projectId = projectId;
    }

    String getProjectId() {
        return projectId;
    }

    @Override
    public Integer getBatchSize() {
        return 200;
    }
}
```

cbismuth:
After debugging, the `i.m.s.StackdriverMeterRegistry#publish` API finely gets a batch size value of `200`, but we still have the warning in logs and no metric pushed in Stackdriver.

cbismuth:
[...]

PR #1347, opened 2019-03-29
#1340 Add Stackdriver batch size limit

cbismuth:
Stackdriver's API limits us to less than 200 events per call, see documentation below.

* https://cloud.google.com/monitoring/quotas

### DIFF
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
@@ -140,7 +140,8 @@ public class StackdriverMeterRegistry extends StepMeterRegistry {
             return;
         }
 
-        for (List<Meter> batch : MeterPartition.partition(this, config.batchSize())) {
+        // Stackdriver's API limits us to less than 200 events per call
+        for (List<Meter> batch : MeterPartition.partition(this, Math.min(config.batchSize(), 199))) {
             Batch publishBatch = new Batch();
 
             Iterable<TimeSeries> series = batch.stream()
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
@@ -140,9 +140,7 @@ public class StackdriverMeterRegistry extends StepMeterRegistry {
             return;
         }
 
-        // Stackdriver's API limits us to less than 200 events per call
-        int limit = 199;
-        for (List<Meter> batch : MeterPartition.partition(this, Math.min(config.batchSize(), limit))) {
+        for (List<Meter> batch : MeterPartition.partition(this, config.batchSize())) {
             Batch publishBatch = new Batch();
 
             AtomicLong partitioningCounter = new AtomicLong();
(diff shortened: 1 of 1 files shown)


shakuzen:
We can merge this, but one Micrometer meter can translate to more than one Stackdriver TimeSeries. For instance, Timers will have at least a distribution and max time series. To properly fix this we need to do the batching on the converted TimeSeries, I think.

cbismuth:
Thanks @shakuzen, here is another proposal. I've tried to make this change as small as possible, but feel free to ask me for other changes as well.

cbismuth:
I'll fix the Guava illegal import.

cbismuth:
Alright, we're back to green, I stay available for any further change.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
```diff
@@ -140,10 +140,14 @@ protected void publish() {
             return;
         }
 
-        for (List<Meter> batch : MeterPartition.partition(this, config.batchSize())) {
+        // Stackdriver's API limits us to less than 200 events per call
+        int limit = 199;
+        for (List<Meter> batch : MeterPartition.partition(this, Math.min(config.batchSize(), limit))) {
```
shakuzen: We don't need this partitioning anymore since the subsequent grouping is already splitting the `TimeSeries`.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
```diff
@@ -140,10 +140,14 @@ protected void publish() {
             return;
         }
 
-        for (List<Meter> batch : MeterPartition.partition(this, config.batchSize())) {
+        // Stackdriver's API limits us to less than 200 events per call
+        int limit = 199;
```
shakuzen: I think 200 is right. Perhaps before you had mostly Micrometer meters that translated to 1 Stackdriver `TimeSeries` each, and 1 meter that translated to 2 `TimeSeries` which put it over the limit when batching 200 meters?

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
cbismuth: Good catch, thanks, fixed in c99df29b4a30cfb318195241eef0f6e72d01e751.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
cbismuth: I've restored `200` in c99df29b4a30cfb318195241eef0f6e72d01e751 and I'll give it a second try on our systems and let you know.

cbismuth: @shakuzen, I can confirm 200 is right :+1:

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
shakuzen: I mean I don't think we need to use MeterPartition at all with Stackdriver anymore, since we are batching the `TimeSeries` before sending. I can fix this while merging, though.

shakuzen:
Merged as 26be0ac and polished with 7bfe2cf

cbismuth:
Thanks a lot @shakuzen :+1:

Issue #1349, opened 2019-04-01
Memory leak in JvmGcMetrics.bindTo


blaszczyk:
`JvmGcMetrics.bindTo` registers a listener in the Jvm Gc system:
`((NotificationEmitter) mbean).addNotificationListener((notification, ref) -> {...})`
without the possibility to ever remove it. In an application server environment like tomcat, when undeploying an application that includes micrometer, this prevents the applications classloader together with its whole metaspace from being garbage collected.

izeye:
@blaszczyk Thanks for the report! I created https://github.com/micrometer-metrics/micrometer/pull/1352 to try to resolve this.

blaszczyk:
Thank you for fixing this so quickly. One last question: Will the removal of the listeners be triggered automatically, in particular when using micrometer with spring-boot-actuator, or are there actions required in the client code?

izeye:
@blaszczyk I'd expect the Spring framework invokes its `close()` on context close.

PR #1352, opened 2019-04-01
Clean up JMX notification listeners for JVM GC metrics

izeye:
This PR changes to clean up JMX notification listeners for JVM GC metrics after binding meters similar to `TomcatMetrics`.

Closes gh-1349

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -25,14 +25,18 @@ import io.micrometer.core.lang.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.management.ListenerNotFoundException;
 import javax.management.NotificationEmitter;
+import javax.management.NotificationListener;
 import javax.management.openmbean.CompositeData;
 import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
 import java.lang.management.MemoryPoolMXBean;
 import java.lang.management.MemoryUsage;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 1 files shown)


izeye:
For reviews, see https://github.com/micrometer-metrics/micrometer/pull/1352/files?w=1

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -133,64 +136,74 @@ public void bindTo(MeterRegistry registry) {
...
 
-                            if (youngGenPoolName != null) {
-                                final long youngBefore = before.get(youngGenPoolName).getUsed();
-                                final long youngAfter = after.get(youngGenPoolName).getUsed();
-                                final long delta = youngBefore - youngGenSizeAfter.get();
-                                youngGenSizeAfter.set(youngAfter);
-                                if (delta > 0L) {
-                                    allocatedBytes.increment(delta);
+                                try {
+                                    ((NotificationEmitter) mbean).removeNotificationListener(this);
```
mweirauch: Maybe my diff viewing skills are fooling me, but isn't this removing the listener exactly after the first notification?

izeye: @mweirauch I thought the first notification is enough for binding meters similar to Tomcat one but didn't check. Are the subsequent events considered to get meters bound fully?

izeye: @mweirauch Oops. Looking into the code closely, this implementation seems completely wrong. Thanks for pointing this out.

izeye: Unless there's a proper shutdown hook, I'm not sure how to remove the listener at the moment.

izeye: One way might be to make `JvmGcMetrics` `AutoCloseable`.

izeye: @mweirauch Thanks for the feedback! I changed `JvmGcMetrics` to implement `AutoCloseable` although I'm not sure there's a better fix for this.

PR #1359, opened 2019-04-03
Add Dropwizard metrics binder

cbornet:
Dropwizard `MetricRegistryListener` that binds Dropwizard's metrics with Micrometer Meters.
Useful for third-party libraries that expose their metrics only through Dropwizard registries.
Example with cassandra's driver:
```java
    cluster.getMetrics().getRegistry().addListener(new MicrometerMetricRegistryListener(meterRegistry, "cassandra"));

```

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/dropwizard/MicrometerMetricRegistryListenerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/dropwizard/MicrometerMetricRegistryListenerTest.java
@@ -1,0 +1,97 @@
+/**
+ * Copyright 2019 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.dropwizard;
+
+import com.codahale.metrics.*;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class MicrometerMetricRegistryListenerTest {
+
+    private static final String METRIC_PREFIX = "prefix";
+    private static final String METRIC_NAME = "metric";
+    private MeterRegistry meterRegistry = new SimpleMeterRegistry();
+    private MetricRegistry metricRegistry = new MetricRegistry();
+
+    @BeforeEach
+    void setUp() {
+        metricRegistry.addListener(new MicrometerMetricRegistryListener(meterRegistry, METRIC_PREFIX));
+    }
+
+    @Test
+    void shouldBindGauge() {
+        AtomicInteger innerVal = new AtomicInteger();
+        Gauge<Integer> gauge = innerVal::get;
+        metricRegistry.register(METRIC_NAME, gauge);
+        innerVal.set(1);
+        innerVal.set(42);
+        assertEquals(42, meterRegistry.get(String.format("%s.%s", METRIC_PREFIX, METRIC_NAME)).gauge().value());
+    }
+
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/dropwizard/MicrometerMetricRegistryListenerTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class MicrometerMetricRegistryListenerTest {
+
+    private static final String METRIC_PREFIX = "prefix";
+    private static final String METRIC_NAME = "metric";
+    private MeterRegistry meterRegistry = new SimpleMeterRegistry();
```
checketts: If the `MeterRegistry` is the dropwizard flavor, will we get into a loop, or will it recover gracefully?

cbornet: If you use the same MetricRegistry in both, yes it will loop. But why would you do that ?
(Note that the JMX and Graphite autoconf create their own MetricRegistry)

checketts: Darn I was hoping it would just hit it twice, but 'add' wouldn't get called for subsequent times. (Just as a safety net).

You wouldn't intentionally do that. But it would be good to test it or note it in the javadoc so users can avoid getting burned.

Are there any autoconfigs that wire up a DropWizard registry? That magic could surprise someone if a registry was created and added to the composite without the user realizing.

cbornet: The system loops because there's a prefix that is added to the metric name. I've pushed a commit to not add a prefix by default. This is however not enough since now you'll get another exception `throw new IllegalArgumentException("A metric named " + name + " already exists");`.
I don't see how to completely prevent this but also I don't think this can really happen in real life...

rreimann:
Is that a change that might make it into release 1.1.5 and if yes is there an approximate release date for 1.1.5?

shakuzen:
> Is that a change that might make it into release 1.1.5 and if yes is there an approximate release date for 1.1.5?

Since this is a purely additive feature, it will go in the next feature release made from when it is merged. Most likely 1.2 or 1.3. Tentative timelines for releases are updated in the milestones page as they are decided.

jkschneider:
I think this kind of thing might make sense as a separate jar like `dropwizard-over-micrometer`, analagous to `log4j-over-slf4j`. I think it's hard to determine how robust this will be, especially how to reliably deconstruct a hierarchical name into a dimensional one. Would be nice to be able to rev this independently.

asarkar:
I've the exact same use case as the author (Cassandra metrics). Is there an agreement on how to proceed (separate JAR vs core) - the author hasn't responded since April.

cbornet:
I'm still there 😀. Do I have to do something ?

asarkar:
@cbornet 
> Do I have to do something ?

I'd think so, otherwise the PR would've been merged.

cbornet:
Just tell me then. I can do the separate jar if that's the way to go.

asarkar:
@cbornet I'm just a guy, not a collaborator on this project. You probably wanna contact the other people who commented. You probably already know, but they've a Slack channel http://slack.micrometer.io/

asarkar:
@cbornet I found a problem with your code. One of the meter exposes is `public com.codahale.metrics.Timer getRequestsTimer()` which you're converting to a counter, thus dropping the other attributes of a timer on the floor. Among them, latencies, very important.
```
public com.codahale.metrics.Timer getRequestsTimer()
Returns metrics on the user requests performed on the Cluster.
This metric exposes

the total number of requests.
the requests rate (in requests per seconds), including 1, 5 and 15 minute rates.
the mean, min and max latencies, as well as latency at a given percentile.
Returns:
a Timer metric object exposing the rate and latency for user requests.
```

cbornet:
A Timer holds values in a bucket/reservoir in which the timestamp of the records is important so I don't think it's possible to copy them from one Timer to another. But maybe there's a way ?

asarkar:
Filed https://github.com/dropwizard/dropwizard/issues/2840.

If dealing with Cassandra only, a hackish workaround can be found in the ticket I filed against Spring Data Cassandra: https://jira.spring.io/browse/DATACASS-675.

At this time, I think this PR serves no general purpose, and doesn't belong in the Micrometer core. I recommend closing it.

cbornet:
Unless you know a better way, this PR still has value. For instance for Cassandra, you get total number of requests which can be derived to request rate. Indeed you lose latencies but it's better than nothing.
About moving to another jar, as I said,  I'm ready to do it. Let the maintainers decide.

asarkar:
I do have a better idea; doing a PR for the Dropwizard issue. I’ll do that when I’ve time. As for this PR, as you said, the maintainers can decide but as far as I’m concerned, it’s not even close to a general solution.

PR #1362, opened 2019-04-08
Polish NamingConvention

izeye:
This PR polishes `NamingConvention`.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
@@ -81,10 +81,10 @@ public interface NamingConvention {
                     conventionName.append(str);
                 } else {
                     final char firstChar = str.charAt(0);
-                    if (Character.isTitleCase(firstChar)) {
+                    if (Character.isUpperCase(firstChar)) {
                         conventionName.append(str); // already capitalized
                     } else {
-                        conventionName.append(String.valueOf(Character.toTitleCase(firstChar))).append(str.substring(1));
+                        conventionName.append(Character.toUpperCase(firstChar)).append(str.substring(1));
                     }
                 }
             }
@@ -100,11 +100,7 @@ public interface NamingConvention {
         }
 
         private String capitalize(String name) {
-            if (name.length() == 0) {
-                return name;
-            }
-
-            if (Character.isUpperCase(name.charAt(0))) {
+            if (name.length() == 0 || Character.isUpperCase(name.charAt(0))) {
                 return name;
             }
 


micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
```diff
@@ -81,10 +81,10 @@ private String toCamelCase(String value) {
                     conventionName.append(str);
                 } else {
                     final char firstChar = str.charAt(0);
-                    if (Character.isTitleCase(firstChar)) {
+                    if (Character.isUpperCase(firstChar)) {
```
shakuzen: What was the motivation to switch to using the `*UpperCase` methods instead? Reading the JavaDoc, it seems there is a difference between the behavior of these two methods with some characters (e.g. `Ǉ`, `ǈ`). Using title-case seems more correct to me for camelCase. Whether we need to support the limited characters that would be problematic or not, I'm not sure, though.
https://www.compart.com/en/unicode/U+01C8

shakuzen: Looking into it more, I see a single uppercase letter character returns `false` for `isTitleCase`. So currently, for almost all cases, it is going to the else block here.

shakuzen:
While looking into this change, I noticed that the camelCase naming convention does not lowercase the first character in a name like I would expect. I wonder if that is intentional or not.
/cc @jkschneider 
For example, this test fails:
```java
@Test
void camelCase2() {
    String name = NamingConvention.camelCase.name("Some.Name.with.Words", Meter.Type.COUNTER);
    assertThat(name).isEqualTo("someNameWithWords");
}
```

micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
izeye: @shakuzen I just wanted to align it with [the `upperCamelCase` implementation](https://github.com/micrometer-metrics/micrometer/blob/1.0.x/micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java#L95-L115) and it seemed reasonable to me to use them as I thought meter names and tag keys should be alphanumeric except delimiters. If my assumption is incorrect, this change should be reverted.

izeye:
@shakuzen It seems to be a bug to me. I can try to fix it with another commit in the PR if it's appropriate.

shakuzen:
> I can try to fix it with another commit in the PR if it's appropriate.

I think it's better as a separate PR

izeye:
@shakuzen On second thought it seems intentional as you suspected above. It does seem to rely on Micrometer's default lowercase dot notation.

izeye:
[This code](https://github.com/micrometer-metrics/micrometer/blob/1.0.x/micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java#L84) is confusing me. If my assumption above is correct, it’d be better to remove the code and update the test accordingly.

micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
shakuzen: `isTitleCase` is definitely not what we want, now that I understand what it does.

shakuzen:
As for the failing test case I mentioned, while it would be more correct in general for a camelCase naming convention to lowercase the first character, in the context of Micrometer, it is expected that the input will be in the defined lowercase dot-notation. Since the issue wasn't brought up by a user, let's leave it the way it is currently. If there is somewhere to better document this expectation on the input that doesn't already, let's update that.
In the meantime, this polish you've proposed is good to be merged.

PR #1363, opened 2019-04-08
Polish MeterFilter

izeye:
This PR polishes `MeterFilter`.

PR #1365, opened 2019-04-08
Polish MeterFilter against 1.0.x

izeye:
This PR polishes `MeterFilter` against the 1.0.x branch.

See https://github.com/micrometer-metrics/micrometer/pull/1363

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
@@ -24,7 +24,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
```diff
@@ -301,7 +300,7 @@ static MeterFilter maxExpected(String prefix, Duration max) {
         return new MeterFilter() {
             @Override
             public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticConfig config) {
-                if (Meter.Type.TIMER.equals(id.getType()) && id.getName().startsWith(prefix)) {
+                if (id.getType() == Meter.Type.TIMER && id.getName().startsWith(prefix)) {
```
shakuzen: This change removes an inherent null-check, but Id's `type` is not marked as Nullable, so it is probably okay.

izeye: I'm not sure I understood your comment correctly but it should work as before regardless its nullability. Am I missing something?

shakuzen: Sorry, I was wrong. You're right, it will work the same.

PR #1369, opened 2019-04-09
Make mutable fields volatile in MeterRegistry

izeye:
This PR changes mutable fields to be `volatile` in `MeterRegistry` as they could suffer from visibility issue theoretically.


shakuzen:
I think we should measure the cost (if anything) of this change before we consider it, since no users have reported issues due to this and it would be an unusual case (from use cases we currently know about) for this to be a problem now.

shakuzen:
Let's leave things as they are now. We generally only expect these to be set before creating/using meters. I don't want everyone to pay the price of volatile access for a scenario we haven't had reported as an issue so far.

Issue #1370, opened 2019-04-09
Micrometer logs error message when partial error happens due to Infinity value using Elasticsearch


mmaia:
Micrometer core and  micrometer-elastic version: 1.1.3

Springboot version: 2.1.3.RELEASE

Hi, Using micrometer getting metrics from springboot application using `micrometer-registry-elastic` 

Micrometer keeps logging error messages: 

`ERROR [market-fetcher,,,] 8106 --- [trics-publisher] i.m.elastic.ElasticMeterRegistry         : failed to send metrics to elastic:`   

When checking elastic index it's working, index get's updated with new values but I noticed one of the entries in there reports an error: 

`{"index":{"_index":"market-fetcher-metrics-2019-04","_type":"doc","_id":"XyEp_WkB2oH6K2ScUlT5","status":400,"error":{"type":"mapper_parsing_exception","reason":"failed to parse [value]","caused_by":{"type":"illegal_argument_exception","reason":"[float] supports only finite values, but got [Infinity]"}}}}`

which is probably what's causing the error log entry being generated. I guess it should instead be able to wrap / handle Infinity? Also not sure where the error is exactly from as we don't explicitly wrap any of the metrics in primitive float, still investigating that. 

Find below a  full entry of the log error message and pls let me know if it needs further details.

'2019-04-08 15:35:37.761 ERROR [market-fetcher,,,] 8106 --- [trics-publisher] i.m.elastic.ElasticMeterRegistry         : failed to send metrics to elastic: {"took":24,"errors":true,"items":[{"index":{"_index":"market-fetcher-metrics-2019-04","_type":"doc","_id":"CCEp_WkB2oH6K2ScUlT5","_version":1,"result":"created","_shards":{"total":2,"successful":1,"failed":0},"_seq_no":322,"_primary_term":1,"status":201}},{"index":{"_index":"market-fetcher-metrics-2019-04","_type":"doc","_id":"CSEp_WkB2oH6K2ScUlT5","_version":1,"result":"created","_shards":{"total":2,"successful":1,"failed":0},"_seq_no":328,"_primary_term":1,"status":201}},
... 
[...]

PR #1374, opened 2019-04-10
Drop custom meters non-finite measurements for Elasticsearch

izeye:
This PR drops custom meters having any non-finite measurement for Elasticsearch.

Closes gh-1370

### DIFF
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
@@ -32,7 +32,6 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Date;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
```diff
@@ -254,9 +257,21 @@ private String indexName() {
 
     // VisibleForTesting
     Optional<String> writeMeter(Meter meter) {
+        Iterable<Measurement> measurements = meter.measure();
+        List<Double> values = new ArrayList<>();
+        for (Measurement measurement : measurements) {
+            double value = measurement.getValue();
+            if (!Double.isFinite(value)) {
+                return Optional.empty();
```
shakuzen: Could we just skip the measurement with the non-finite value, rather than all of the measurements?

izeye: @shakuzen I chose this way due to [this internal restriction](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java#L50-L51) but publishing it sounds like a different concern, so I pushed 7700492 as you suggested.

implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
```diff
@@ -254,9 +256,23 @@ private String indexName() {
...
+            double value = measurement.getValue();
+            if (!Double.isFinite(value)) {
+                continue;
+            }
+            names.add(measurement.getStatistic().getTagValueRepresentation());
+            values.add(value);
+        }
+        if (names.isEmpty()) {
+            return Optional.empty();
+        }
```
shakuzen: The current implementation is correct and works, but I wondered if maybe something like the following is more straightforward.

```java
List<Measurement> measurements = StreamSupport.stream(meter.measure().spliterator(), false)
        .filter(measurement -> Double.isFinite(measurement.getValue()))
if (measurements.isEmpty()) {
    return Optional.empty();
}
```

Then the code below can stay the same also.

izeye: @shakuzen The reason I chose that way is due to the possibility of changes on the values in-between as you pointed it out in another issue.

shakuzen: Oh how short my memory is :D
We should probably leave a comment to make it more apparent why the implementation is that way, then, so someone like forgetful me doesn't reintroduce a bug like that.

izeye: @shakuzen Thanks for the feedback! I added a comment in 1f78437867c60ff994a5f430d044310f5e684af1.

Issue #1389, opened 2019-04-18
TelegrafStatsdLineBuilder appears to have a race condition


bw-tom:
As raised here https://stackoverflow.com/questions/55631373/telegrafstatsdlinebuilder-appears-to-have-a-race-condition, it seems we can consistently get metrics published to StatsD without common tags applied which _do_ appear on the Actuator Metrics endpoint, or when the Counter's `increment` is invoked by only a single thread at a time.

I've made a small Spring Boot app which seems to reliably reproduce the problem: https://github.com/bw-tom/concurrent-metrics-publisher/

izeye:
@bw-tom [`@Autowired` with an infinite loop](https://github.com/bw-tom/concurrent-metrics-publisher/blob/master/src/main/java/com/brandwatch/metrics/App.java#L28-L33) seems to prevent from completing to start the application. Is this intentional or am I missing something?

bw-tom:
@izeye Should only start the infinite loop once the application is ready, from my reading of https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/context/event/ApplicationReadyEvent.html!

izeye:
@bw-tom As mentioned above, `@Autowired` seems to kick in first to me, so I’m not sure how it works for you. By the way why do you need the annotation there?

bw-tom:
Ah, I think the Autowired actually does nothing! An earlier version of the method had parameters that needed to be wired in, I'll remove it 👍

izeye:
@bw-tom Thanks for the report and the sample! I reproduced it thanks to your sample. I created #1391 to try to resolve this.

bw-tom:
Happy to help, and thank you very much for the rapid response!

PR #1174, opened 2019-01-22
Handle NaN and infinities for Humio meter registry

izeye:
This PR changes to handle NaN and infinities for Humio meter registry.

This PR also polishes Humio support along the way.

Note: I added a separate commit updating depedency locks to pass a build. FTR I did the following to update dependency locks:

```
./gradlew clean check
...
* What went wrong:
Could not resolve all dependencies for configuration ':micrometer-spring-legacy:compileClasspath'.
> Dependency lock state for configuration 'compileClasspath' is out of date: Resolved 'javax.inject:javax.inject:1' which is not part of the lock state
...

./gradlew :micrometer-spring-legacy:classes --update-locks javax.inject:javax.inject

...
* What went wrong:
Could not resolve all dependencies for configuration ':micrometer-samples-boot1:compileClasspath'.
> Dependency lock state for configuration 'compileClasspath' is out of date: Resolved 'javax.inject:javax.inject:1' which is not part of the lock state
...

./gradlew :micrometer-samples-boot1:classes --update-locks javax.inject:javax.inject
```

If the dependency lock updates are inappropriate, we can drop the commit when merging.

PR #1391, opened 2019-04-20
Prevent race condition when updating tags for StatsD Telegraf flavor

izeye:
This PR prevents race condition when updating tags for StatsD Telegraf flavor.

This PR also contains a preceding commit updating dependency lockfiles to resolve build failures similar to done in https://github.com/micrometer-metrics/micrometer/pull/1174. I didn't include unversioned dependency lockfiles intentionally to reduce verbosity in diff and hope to be okay without them.

And I noticed `DatadogStatsdLineBuilder` has the same concern, so if this approach is okay, I'll prepare a similar one for it.

Closes gh-1389


shakuzen:
`downloadDependencies` task is failing due to dependency locking after merging this... the dependency lock time bandit strikes again.

shakuzen:
After locally running `./gradlew downloadDependencies --update-locks javax.inject:javax.inject` and `./gradlew build --update-locks javax.inject:javax.inject`, it seems to have undone your lockfile changes and the build passes now.

izeye:
@shakuzen Strange... It fails to me without the dependency lockfiles updates locally although I didn't include all the changes from it, so I might miss some pieces from the unversioned changes.

PR #1392, opened 2019-04-21
Backporting changes for issue #1282

Imaskar:
For SpringBoot 2.x changes needs to be fixed from SpringBoot side and
they are proposed by izeye here: https://github.com/spring-projects/spring-boot/pull/16221

However, for SpringBoot 1.5.x micrometer-spring-legacy is in use and
changes shall be ported there. This is a goal of this PR.

### DIFF
--- a/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/MeterRegistryConfigurer.java
+++ b/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/MeterRegistryConfigurer.java
@@ -18,6 +18,7 @@ package io.micrometer.spring.autoconfigure;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.binder.MeterBinder;
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
 import io.micrometer.core.instrument.config.MeterFilter;
 
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


izeye:
@Imaskar Thanks for the PR! Unlike Spring Boot 2.0.x, Micrometer 1.0.x is still active, so I think it'd be better to change the target branch to the 1.0.x.

Imaskar:
This bug does not exist for 1.0.x branch.

izeye:
@Imaskar I didn't look into the branch in details but just looking at it, the same code seems to live there: https://github.com/micrometer-metrics/micrometer/blob/1.0.x/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/MeterRegistryConfigurer.java#L59

Am I missing something?

Imaskar:
Probably the Meter classes implementation was changed between the branches. We noticed this bug when we tried to migrate from 1.0.5 to 1.1.3. I checked it again and I don't see this bug on 1.0.5.

BTW, the original pull request was recently merged with a bit of style changes. I will port them too soon to keep codebases as close as possible.

izeye:
@Imaskar Thanks for the quick feedback! I’d like to understand why regardless of this fix as the fix doesn’t cover nested CompositeMeterRegistry hierarchy.

Imaskar:
> I’d like to understand why regardless of this fix as the fix doesn’t cover nested CompositeMeterRegistry hierarchy.

I'm sorry, I don't understand. Do you mean why is 1.0.x branch fine without the fix?

izeye:
@Imaskar Sorry, it wasn't clear. Yes, that was what I meant.

shakuzen:
I have incorporated the polish commit into this pull request and rebased it for merging into the 1.0.x branch.

PR #1401, opened 2019-04-28
Tag URI with OkHttp client when call times out

### LINKED ISSUE #1285: Uri (UNKNOWN)  is incorrect when okhttp call ends with timeout
When okhttp's call ends due to timeout error then published metric looks like this
`GET - UKNOWN - IO_ERROR` (pattern: `{{method}} - {{uri}} - {{status}}`)

That's enough if I call just one uri - so I know what uri cases the problem. However this information has no sense when I have more than uri to call.

For example:
- I am calling 2 endpoints:
 1.`GET example.com/xxx`
 2.` GET example.com/yyy`

- Assuming that some calls ended correctly and on failed due to timeout then in Prometheus it looks like:

`GET /xxx 200`
`GET /yyy 200`
`GET UNKNOWN IO_ERROR`

And i have no guess which one timedout.

Is there sense keeping it as it's or PR is welcomed?

play-io:
Resolves #1285

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
@@ -21,6 +21,7 @@ import io.micrometer.core.instrument.MockClock;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.simple.SimpleConfig;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import java.util.function.Function;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


play-io:
@shakuzen PR based on [this](https://github.com/micrometer-metrics/micrometer/issues/1285)

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
```diff
@@ -76,7 +82,7 @@ void timeNotFound(@WiremockResolver.Wiremock WireMockServer server) throws IOExc
         client.newCall(request).execute().close();
 
         assertThat(registry.get("okhttp.requests")
-                .tags("foo", "bar", "uri", "NOT_FOUND")
+                .tags("foo", "bar", "status", "404", "uri", URI_EXAMPLE_VALUE)
```
shakuzen: This behavior is problematic because it can lead to the cardinality of tag combinations exploding. A nearly infinite URIs do not exist, and if we tag them, there's no guarantee on a reasonable cardinality. That's why 404 responses used `NOT_FOUND` for the URI previously. We need to maintain this behavior. Same for the previous logic about 301 status codes.

play-io: I could imaging that this could be a problem for crawlers developers, but then i am not really sure they are using this stats. Could you please elaborate on this or give a link?

shakuzen: Here's a [warning in the documentation](http://micrometer.io/docs/concepts#_tag_values):

> Beware of the potential for tag values coming from user-supplied sources to blow up the cardinality of a metric. You should always carefully normalize and bound user-supplied input. Sometimes the cause is sneaky. Consider the URI tag for recording HTTP requests on service endpoints. If we don’t constrain 404’s to a value like NOT_FOUND, the dimensionality of the metric would grow with each resource that can’t be found.

An example is if I access the non-existent URI `/resource/{n}` where `n` is in the range `0..9999999` then this would tag with 10 million unique tag values (of course any larger number or different URI pattern can be used as well). This is not scalable, and will cause performance issues in most metric backends. If a user controls the client and doesn't have this kind of concern, they would need to make their own implementation to have such tagging. We need to be safe in what we provide by default.

PR #1407, opened 2019-05-02
Support long task timer with Timed annotaion 

### LINKED ISSUE #604: TimedAspect support of LongTaskTimer
As of now [TimedAspect](https://github.com/micrometer-metrics/micrometer/blob/0061a4aeba8c59ba018307cfa03435739c973827/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java) doesn't care about longTask value set to [@Timed](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/annotation/Timed.java).
Do you have any plan to switch to long task timers in case longTask value is set to true ? 

```
    @Around("execution (@io.micrometer.core.annotation.Timed * *.*(..))")
    public Object timedMethod(ProceedingJoinPoint pjp) throws Throwable {
        Method method = ((MethodSignature) pjp.getSignature()).getMethod();
        Timed timed = method.getAnnotation(Timed.class);
        final String metricName = timed.value().isEmpty() ? DEFAULT_METRIC_NAME : timed.value();

        Timer.Sample sample = Timer.start(registry);
        try {
            return pjp.proceed();
        } finally {
            sample.stop(Timer.builder(metricName)
                    .description(timed.description().isEmpty() ? null : timed.description())
                    .tags(timed.extraTags())
                    .tags(tagsBasedOnJoinpoint.apply(pjp))
                    .publishPercentileHistogram(timed.histogram())
                    .publishPercentiles(timed.percentiles().length == 0 ? null : timed.percentiles())
                    .register(registry));
        }
    }
```

nejckorasa:
Support `LongTaskTimer` with `@Timed` annotation.

Closes #604

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -128,7 +128,7 @@ public class TimedAspect {
             return pjp.proceed();
         } finally {
             try {
-                longTaskTimer.stop(sample.task());
+                sample.stop();
             } catch (Exception e) {
                 // ignoring on purpose
             }
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -17,6 +17,7 @@ package io.micrometer.core.aop;
 
 import io.micrometer.core.annotation.Incubating;
 import io.micrometer.core.annotation.Timed;
+import io.micrometer.core.instrument.LongTaskTimer;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
@@ -84,6 +85,19 @@ public class TimedAspect {
         Timer.Sample sample = Timer.start(registry);
         String exceptionClass = "none";
 
+        if (!timed.longTask()) {
+            return processWithTimer(pjp, timed, metricName, sample, exceptionClass);
+        } else {
+            return processWithLongTaskTimer(pjp, timed, metricName);
+        }
+    }
+
+    private Object processWithTimer(
+            ProceedingJoinPoint pjp,
+            Timed timed,
+            String metricName,
+            Timer.Sample sample,
+            String exceptionClass) throws Throwable {
         try {
             return pjp.proceed();
         } catch (Exception ex) {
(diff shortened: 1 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
```diff
@@ -87,13 +97,33 @@ public Object timedMethod(ProceedingJoinPoint pjp) throws Throwable {
...
+                .tags(tagsBasedOnJoinPoint.apply(pjp))
+                .register(registry);
+
+        LongTaskTimer.Sample sample = longTaskTimer.start();
+
+        try {
+            return pjp.proceed();
+        } finally {
+            try {
+                longTaskTimer.stop(sample.task());
```
shakuzen: Can `sample.stop()` be used instead here? Then adding the `task()` method would not be needed.

nejckorasa: Yes, that's much cleaner, I've changed it.

nejckorasa:
@shakuzen I've added tests for it. Also wrapped the creation of `LongTaskTimer` so as not to disrupt the application flow in case of exception.

nejckorasa:
> Thank you very much for the contribution. This looks good to me. Do you have interest/time to add tests also? If not, I can do it. Let me know.

I've tested it... Do the added tests and changes look ok?

micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
```diff
@@ -92,16 +104,46 @@ public Object timedMethod(ProceedingJoinPoint pjp) throws Throwable {
...
-                        .description(timed.description().isEmpty() ? null : timed.description())
-                        .tags(timed.extraTags())
-                        .tags(EXCEPTION_TAG, exceptionClass)
-                        .tags(tagsBasedOnJoinPoint.apply(pjp))
-                        .publishPercentileHistogram(timed.histogram())
-                        .publishPercentiles(timed.percentiles().length == 0 ? null : timed.percentiles())
-                        .register(registry));
+                                    .description(timed.description().isEmpty() ? null : timed.description())
+                                    .tags(timed.extraTags())
+                                    .tags(EXCEPTION_TAG, exceptionClass)
```
shakuzen: It looks like the exception tag isn't applied to the LongTaskTimer. I can add this post-merge, though, unless there was some reason it wasn't included.

shakuzen:
> I've tested it... Do the added tests and changes look ok?

Looks good. Thank you!

micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
shakuzen: Nevermind. I see now why we can't tag the exception class.

Issue #1071, opened 2018-12-14
Elastic indexing counters/timers can fail due to different numeric types


jkschneider:
See https://github.com/micrometer-metrics/micrometer/issues/498#issuecomment-446816808

izeye:
In Elasticsearch 6.7.2 (the latest 6.x version ATM), a floating point number (`count`) in JSON is typed as `float` as follows:

```
$ curl -XDELETE http://localhost:9200/metrics-2019-05
{"acknowledged":true}
$ curl -XPOST --header "Content-Type: application/json" http://localhost:9200/metrics-2019-05/doc/_bulk --data-binary '
> { "index" : {} }
> {"@timestamp":"2019-05-14T15:31:46.313Z","name":"jvm_gc_memory_promoted","type":"counter","count":0.0}
> '
{"took":105,"errors":false,"items":[{"index":{"_index":"metrics-2019-05","_type":"doc","_id":"l7_-tmoBVqC16kvPmpLF","_version":1,"result":"created","_shards":{"total":2,"successful":1,"failed":0},"_seq_no":0,"_primary_term":1,"status":201}}]}
$ curl http://localhost:9200/metrics-2019-05/_mapping?pretty
{
  "metrics-2019-05" : {
    "mappings" : {
      "doc" : {
        "_all" : {
          "enabled" : false
        },
        "properties" : {
          "@timestamp" : {
            "type" : "date"
          },
          "count" : {
            "type" : "float"
          },
          "name" : {
            "type" : "keyword"
          },
          "type" : {
            "type" : "text",
            "fields" : {
              "keyword" : {
                "type" : "keyword",
                "ignore_above" : 256
              }
            }
          }
        }
      },
      "_default_" : {
        "_all" : {
          "enabled" : false
        },
        "properties" : {
          "name" : {
            "type" : "keyword"
          }
        }
      }
    }
  }
}
$
```

An integer number (`count`) in JSON is typed as `long` as follows:

```
$ curl -XDELETE http://localhost:9200/metrics-2019-05
{"acknowledged":true}
$ curl -XPOST --header "Content-Type: application/json" http://localhost:9200/metrics-2019-05/doc/_bulk --data-binary '
> { "index" : {} }
> {"@timestamp":"2019-05-14T15:31:46.313Z","name":"jvm_gc_pause","type":"timer","action":"end of major GC","cause":"Metadata GC Threshold","count":0,"sum":0.0,"mean":0.0,"max":0.0}
[...]

PR #1418, opened 2019-05-14
Improve partial error logging for Elasticsearch meter registry

izeye:
When the HTTP status is considered as success but its body has `errors` property, there's no logging for request body unlike unsuccessful HTTP statuses. This PR changes to log HTTP request errors in Elasticsearch meter registry consistently.


shakuzen:
Are you aware of Elasticsearch documentation on this? I would like to understand if they have documented differences of when Elasticsearch will return an unsuccessful status code versus a successful status code with an unsuccessful body.

izeye:
@shakuzen I didn't find any doc on it but when I was looking into #1071 I noticed the `errors` property has been set when there's partial failure of a bulk request.

With regards to the `error` property case I'd like to change the log message to show numbers for both successes and failures as the current message seems to indicate a total failure. If you agree with this, I can update this PR or prepare another one for it.

shakuzen:
> I didn't find any doc on it but when I was looking into #1071 I noticed the `errors` property has been set when there's partial failure of a bulk request.
> 
> With regards to the `error` property case I'd like to change the log message to show numbers for both successes and failures as the current message seems to indicate a total failure. If you agree with this, I can update this PR or prepare another one for it.

Yes that seems to make sense. Otherwise partial failures and complete failures might look the same.

izeye:

> If you agree with this, I can update this PR or prepare another one for it.

Is it okay to update this PR accordingly?

shakuzen:
> Is it okay to update this PR accordingly?

Yes, please.

izeye:
@shakuzen I updated as discussed.

PR #1419, opened 2019-05-14
Add an empty constructor to TimedAspect

luisrosilva:
Having an empty constructor available on TimedAspect would make it easier to use load time weaving via aspectj.
As referenced in the documentation (https://micrometer.io/docs/concepts, on section 3.2), Micrometer provides a static global registry (Metrics.globalRegistry) that can be used to achieve this.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -21,6 +21,7 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.lang.NonNullApi;
 import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.annotation.Around;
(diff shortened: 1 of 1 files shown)


shakuzen:
Sorry for my ignorance on load time weaving, but you could explain more how this constructor makes that easier, perhaps by showing the configuration required before and after this change?

luisrosilva:
No problem, I'm no expert on that either and documentation is rather scarce, so I can be wrong on this. However, here is what I know:

To use load time weaving, we have to provide these additional properties on jvm launch:
```
-javaagent:/Users/luissilva/.m2/repository/org/springframework/spring-instrument/5.1.5.RELEASE/spring-instrument-5.1.5.RELEASE.jar
-javaagent:/Users/luissilva/.m2/repository/org/aspectj/aspectjweaver/1.9.2/aspectjweaver-1.9.2.jar
```

The issue is that aspect instantiation happens in a different place, not controlled by spring. Aspectj weaver doesn't know about spring and as such it can't do injection. Also the spring context doesn't know about the aspect, so it can't help there. According to documentation (https://www.eclipse.org/aspectj/doc/next/progguide/semantics-aspects.html), aspects are instanciated upon need by the weaver, and thus the default empty constructor is used. I've tested these changes locally after finding about Metrics.globalRegistry and it worked ok.

The advantage of using this method is that aspect usage is independent of spring proxies and so, private methods or bean recursive calls can be intercepted with those as well.

shakuzen:
@luisrosilva Thank you for the contribution. This is merged now and will be included in the upcoming 1.2.0 release.

Issue #1420, opened 2019-05-14
Elastic Shipper non finite value


sabareeshkkanan:
@izeye 
I am still facing this issue even with the new snapshot. It produced several metric where elastic might consider it as not finite. 
Spring boot 2.1.4
Spring cloud GreenwichSR1
Java11
Spring Cloud Kafka Streams

```
+--- io.micrometer:micrometer-registry-elastic:1.1.5-SNAPSHOT
|    \--- io.micrometer:micrometer-core:1.1.5-SNAPSHOT (*)

```

```
{ "index" : {} }
{"@timestamp":"2019-05-13T16:50:18.778Z","name":"kafka_consumer_records_lead_min","type":"gauge","activeProfile":"dev","client_id":"consumer-4","partition":"0","region":"","serviceName":"request-handler","topic":"DV_REQUEST_HANDLER","value":1.7976931348623157E308}
{ "index" : {} }
{"@timestamp":"2019-05-13T16:50:18.778Z","name":"kafka_consumer_records_lead_min","type":"gauge","activeProfile":"dev","client_id":"consumer-4","partition":"1","region":"","serviceName":"request-handler","topic":"DV_REQUEST_HANDLER","value":1.7976931348623157E308}
{ "index" : {} }
{"@timestamp":"2019-05-13T16:50:18.778Z","name":"kafka_consumer_records_lead_min","type":"gauge","activeProfile":"dev","client_id":"consumer-4","partition":"2","region":"","serviceName":"request-handler","topic":"DV_REQUEST_HANDLER","value":1.7976931348623157E308}
{ "index" : {} }
{"@timestamp":"2019-05-13T16:50:18.778Z","name":"kafka_consumer_records_lead_min","type":"gauge","activeProfile":"dev","client_id":"consumer-4","partition":"3","region":"","serviceName":"request-handler","topic":"DV_REQUEST_HANDLER","value":1.7976931348623157E308}
{ "index" : {} }
{"@timestamp":"2019-05-13T16:50:18.778Z","name":"kafka_consumer_records_lead_min","type":"gauge","activeProfile":"dev","client_id":"consumer-4","partition":"4","region":"","serviceName":"request-handler","topic":"DV_REQUEST_HANDLER","value":1.7976931348623157E308}
{ "index" : {} }
[...]

PR #1421, opened 2019-05-14
Use double for values in Elasticsearch meter registry

izeye:
This PR changes to use `double` for values in Elasticsearch meter registry.

Closes gh-1071
Closes gh-1420


izeye:
There's a caveat. If the `metrics_template` template already exists, it needs to be deleted first before running. Another name could be chosen but the old one should be cleaned up anyway.

shakuzen:
> There's a caveat. If the `metrics_template` template already exists, it needs to be deleted first before running. Another name could be chosen but the old one should be cleaned up anyway.

I've tagged this with `release notes` so it can be mentioned in the release notes. Thank you for pointing it out.

PR #1423, opened 2019-05-15
InfluxDB 2.0 Meter Registry

bednar:
This PR introduces support for sending metrics to [InfluxDB 2.0](https://v2.docs.influxdata.com/v2.0/get-started/).

The `InfluxMeterRegistry` and `Influx2MeterRegistry` sharing `LineProtocolBuilder` that is used to build [Line protocol](https://v2.docs.influxdata.com/v2.0/reference/line-protocol/).

The InfluxDB 2.0 documentation: https://v2.docs.influxdata.com/v2.0/get-started/

Full API documentation is build in `influx` service:

1. start InfluxDB 2.0 via : `docker run -p 9999:9999 quay.io/influxdb/influx:nightly`
1. to view the API documentation visit the http://localhost:9999/docs in a browser


bednar:
The InfluxDB 2.0 documentation: https://v2.docs.influxdata.com/v2.0/get-started/

Full API documentation is build in `influx` service:

1. start InfluxDB 2.0 via : `docker run -p 9999:9999 quay.io/influxdb/influx:nightly`
1. to view the API documentation visit the http://localhost:9999/docs in a browser

bednar:
Hi @jkschneider, @shakuzen, @izeye 

What do you think about this PR? 

Jon and I discussed on the `Slack` channel that it would be better merge `Influx2MeterRegistry` into `InfluxMeterRegistry`. This combination will be better for user because they will be easily switch from InfluxDB 1 to InfluxDB 2.

It is a right way for use how to proceed the support of InfluxDB 2 into `Micrometer`?

bednar:
The conflicts with #1446 has been resolved.

implementations/micrometer-registry-influx/src/shared/java/io/micrometer/influx/internal/LineProtocolBuilder.java
```diff
@@ -0,0 +1,185 @@
...
+                new Field("count", timer.count()),
+                new Field("mean", timer.mean(baseTimeUnit))
+        );
+
+        return Stream.of(influxLineProtocol(timer.getId(), "histogram", fields));
+    }
+
+    public Stream<String> writeTimer(Timer timer) {
+        final Stream<Field> fields = Stream.of(
+                new Field("sum", timer.totalTime(baseTimeUnit)),
```
michaelandrepearce: What about percentiles of the histogram?

michaelandrepearce: see https://github.com/micrometer-metrics/micrometer/pull/1102

bednar: I would like to follow the master branch. I will rebase #1423 after merge the #1102 into master.

shakuzen:
The documentation seems to indicate that InfluxDB 2.0 is still in alpha status. I'm hesitant to release support for the alpha version, even if we mark it as incubating.

bednar:
Yes, the [InfluxDB 2.0 OSS](https://www.influxdata.com/products/influxdb-overview/influxdb-2-0/) is currently in alpha, but the [InfluxDB 2.0 Cloud](https://www.influxdata.com/products/influxdb-cloud-2-0/) is in beta status. The OSS and Cloud version has identical API.

The beta version of **InfluxDB 2.0 OSS** comes in short time so we could wait with merging.

I would like to have a prepared and reviewed support for InfluxDB 2.0 for the merging after the beta status.

What do you think? Can we continue this way?

shakuzen:
> I would like to have a prepared and reviewed support for InfluxDB 2.0 for the merging after the beta status.
> 
> What do you think? Can we continue this way?

If we have some guarantee that things won't break when the GA version is released, we can get this in a release. I'm not sure on the timing of the InfluxDB 2.0 GA release, but perhaps we can get this in Micrometer 1.3 if the timing works out.

bednar:
OK, I figure out information about state of InfluxDB 2.0 API from @influxdata engineers.

rhajek:
We have confirmation from InfluxData engineering from @russorat, that basic authentication and write API in InfluxDB 2.0 will not be changed in non backward compatible way. After fixing of failing unit test, the pull request should be ready for merging to master.

bednar:
Unit tests are fixed and also the PR is rebased.

bednar:
Hi @shakuzen,

The [InfluxDB 2.0 Cloud](https://www.influxdata.com/products/influxdb-cloud-2-0/) is in beta phase.

Do you think that this PR will be a part of the 1.3 milestone?

shakuzen:
> The InfluxDB 2.0 Cloud is in beta phase.

Thank you for the update.

> Do you think that this PR will be a part of the 1.3 milestone?

It is hard to say right now. I'll certainly try to get to this in time for 1.3 but there's a lot to do.

timhallinflux:
InfluxDB Cloud 2.0 is GA.  InfluxDB OSS 2.0 is in beta.  We have a few folks really interested in this integration to leverage it with our Cloud offering.

jkschneider:
Copying from #1974 ...

> @bednar Without having a look at all the code, can we make the distinction between Influx 1 and 2 fully transparent to the user, i.e. can Micrometer somehow check for v2 support via some API endpoint and just use it and fall back on v1 otherwise?

> Supposing this is not possible, we could introduce an InfluxConfig option for Influx version (carefully validated to an acceptable range of values).

> Only if the other configuration options differ substantially between Influx 1 and 2 would I consider a separate MeterRegistry implementation. In other words, I want to avoid two disjoint sets of config options in one InfluxConfig interface where it isn't clear to the user what they need to configure for which version.

bednar:
Closing this PR due the preferred solution to https://github.com/micrometer-metrics/micrometer/issues/1974

Issue #1408, opened 2019-05-03
Issue when creating Elasticsearch Index


kevcodez:
Using Spring Boot with actuator and micrometer-registry-elastic.

Versions:

* Spring Boot 2.1.4.RELEASE
* Micrometer registry elastic 1.1.4
* Elasticsearch 7.0

I am getting the following exception ONCE when submitting the metrics to Elasticsearch:

```
failed to add metrics template to elastic: {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"Root mapping definition has unsupported parameters:  [_default_ : {_all={enabled=false}, properties={name={type=keyword}}}]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: Root mapping definition has unsupported parameters:  [_default_ : {_all={enabled=false}, properties={name={type=keyword}}}]","caused_by":{"type":"mapper_parsing_exception","reason":"Root mapping definition has unsupported parameters:  [_default_ : {_all={enabled=false}, properties={name={type=keyword}}}]"}},"status":400}

```

I am getting data into the index, not sure if anything is missing due to the previous error.

shakuzen:
I am guessing it is an issue with the index template containing something no longer supported in Elasticsearch 7.0. If metrics are still saved successfully, that's good there's no effective impact, but certainly we'll want to update things to avoid the warning. A pull request from anyone would be appreciated, keeping compatibility with previous versions of Elasticsearch also.

pete-woods:
Some insight from looking into a related project (Zipkin), this is how they are handling the different versions of ElasticSearch:

https://github.com/apache/incubator-zipkin/blob/bf9350eceef8d4b58fb0c7e63ad83e277de4db00/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java

You can see the diff here:
https://github.com/apache/incubator-zipkin/pull/2398/files

pete-woods:
I see, significantly:
- `if (templates.version() >= 7) builder.addQueryParameter("include_type_name", "true");` on the URL
- In the payload: `"\"_default_\": { \"_all\": {\"enabled\": false} },\n"` for < 5 and not at all for 5+
- Also for >=7 `Colons are no longer allowed in index names`

izeye:
I created #1428 to try to resolve this.

PR #1428, opened 2019-05-19
 Make index template compatible with Elasticsearch 7

izeye:
This PR changes to make the index template compatible with Elasticsearch 7 by removing usages of [`_default` mapping](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/default-mapping.html) and [`_all` field](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/mapping-all-field.html) which have been deprecated since Elasticsearch 6 and removed in Elasticsearch 7.

This change affects only with Elasticsearch 7 or later and Elasticsearch 6 or prior will remain as-is.

I've tested this change against Elasticsearch 7 (7.0.1), 6 (6.7.2), and 5 (5.6.15).

This change has been built on top of #1427 as it's necessary to get the version of Elasticsearch.

Closes gh-1408


shakuzen:
@izeye what do you think about setting up some integration tests with testcontainers so we can ensure via CI that this works with all the versions we say? https://www.testcontainers.org/modules/elasticsearch/
Of course, we can make a separate issue for it so as not to block this, but it made me think it is probably time we integration test this.

izeye:
@shakuzen Sounds great and I'd like to but I'm not familiar with it, so it might take some time for me. I think it'd be good to have a separate issue for it. If you're already familiar with it, giving me some direction could boost the progress. Or you could do it yourself if you prefer.

izeye:
I rebased this onto the current 1.1.x branch having #1427.

izeye:
@shakuzen Sorry, I missed authentication part for Elasticsearch version fetching. I'll create a fix for it shortly.

PR #1436, opened 2019-05-24
Append a delimiter at the end of all BufferingFlux messages

sdemart-rei:
The `BufferingFlux` only joins lines with the delimiter `\n`, but does not add a newline (or other delimiter) at the end of the message. I think that behavior might be ok with UDP  (if message are only parsed per packet) but with TCP it causes the first line of the next message sent from the buffer to be joined with the last line of the previous on one line. 

So multiple messages would be mashed together like:

```
...
example.stat:1|c
example.stat:1|c
example.stat:1|cexample.stat:1|c
example.stat:1|c
example.stat:1|c
...
```

and `example.stat:1|cexample.stat:1|c` is not a valid statsd line.

This PR fixes that

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/BufferingFlux.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/BufferingFlux.java
@@ -15,12 +15,12 @@
  */
 package io.micrometer.statsd.internal;
 
-import reactor.core.publisher.DirectProcessor;
-import reactor.core.publisher.Flux;
-
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+
+import reactor.core.publisher.DirectProcessor;
+import reactor.core.publisher.Flux;
 
 public class BufferingFlux {
 
@@ -93,7 +93,7 @@ public class BufferingFlux {
                     }, true)
                     .map(lines -> {
                         lines.removeIf(String::isEmpty); // Ignore empty messages
-                        return String.join(delimiter, lines);
+                        return String.join(delimiter, lines) + delimiter;
                     });
         });
     }
--- a/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/BufferingFluxTest.java
+++ b/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/BufferingFluxTest.java
@@ -15,15 +15,23 @@
  */
 package io.micrometer.statsd.internal;
 
-import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.IntStream;
+
 import reactor.core.publisher.DirectProcessor;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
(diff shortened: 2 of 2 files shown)


shakuzen:
> but with TCP it causes the first line of the next message sent from the buffer to be joined with the last line of the previous on one line.
> 
> So multiple messages would be mashed together like:

I've tried playing around with this locally and I'm not able to reproduce this behavior with UDP or TCP. Have you noticed this behavior reported as bad lines from a StatsD agent? Note that the `flushOnEach` option is set, so if I am understanding correctly, each buffered set of one or more lines should be sent in one packet.

https://github.com/micrometer-metrics/micrometer/blob/4e5f395c65726bd68b11e46e49aebba7741b6056/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java#L244

sdemart-rei:
Yes we have been using statsite for our statsd implementation and have been seeing this, basically resulting in missing data. I think for UDP that `flushOnEach` works fine, but for TCP I don't think the assumptions hold. If the statsd server that listens on TCP parses the data on a socket connection level rather than individual messages, then they will be seeing this behavior since the messages could be mashed together on the receiving server (with no newline delimiter).

And testing locally I started up `nc -l 8025` and pointed micrometer at that and you can see the mashed mashed lines. 

Does that make sense?

shakuzen:
> And testing locally I started up `nc -l 8025` and pointed micrometer at that and you can see the mashed mashed lines.

I was able to see them this way. I was testing using the plain [StatsD Daemon](https://github.com/statsd/statsd), which also showed the mashed lines in the log (I just didn't catch them), but it did not count them as a "bad line" and instead did a weird combining of values/types.

Thank you for reporting the issue, providing a fix, and working with me to reproduce it. We'll get this fix in the upcoming 1.2.0 release (which is the first release with TCP as an option for statsd).
(great branch name, by the way!)

edwardsre:
@shakuzen We just identified this issue happening for us as well, but we are still on the Spring Boot 2.1.9 using Micrometer 1.1.7. It is causing us a lot of grief right now. Can this fix be back ported to the 1.1.x line for the next release, or can we just manage the Micrometer version to 1.2.2?

shakuzen:
> We just identified this issue happening for us as well using UDP, but we are still on Spring Boot 2.1.9 using Micrometer 1.1.7. It is causing us a lot of grief right now. Can this fix be back ported to the 1.1.x line for the next release, or is it possible for us to just pin to Micrometer 1.2.2?

@edwardsre I would expect you can upgrade to Micrometer 1.2.2 (or better yet 1.3.1) without any issues, but of course I would test when doing so. If you would like a backport to 1.1.x, please open a separate issue referencing this with steps to reproduce. I tried but was not able to reproduce the issue with UDP.

edwardsre:
> @edwardsre I would expect you can upgrade to Micrometer 1.2.2 (or better yet 1.3.1) without any issues, but of course I would test when doing so. If you would like a backport to 1.1.x, please open a separate issue referencing this with steps to reproduce. I tried but was not able to reproduce the issue with UDP.

Thanks for the reply. I have been able to reproduce it with UDP and traced it to the same problem fixed in this pull request. I used Wireshark to capture the packets and applied a packet filter of `udp.dstport == 8125`. If you then use Analyze | Follow | UDP Stream,  you see the lines concatenated on every packet boundary:

```
mycounter:1|c|#statistic:count
mycounter:1|c|#statistic:count
mycounter:1|c|#statistic:count
mycounter:1|c|#statistic:count
mycounter:1|c|#statistic:count
mycounter:1|c|#statistic:countmycounter:1|c|#statistic:count
mycounter:1|c|#statistic:count
```
We do not use a StatsD agent, but use a Splunk forwarder listening on udp:8125. The forwarder buffers the packet data and then forwards to Splunk's indexers. There isn't any special processing happening for each packet on the forwarder.

Issue #1441, opened 2019-05-30
CompositeMeters don't ignore NoopMeters when forwarding value get calls to child.


pivotal-jbarrett:
For example, consider a `CompositeCounter` made up of `Counter`s from two registries where one registry may have filtered or otherwise produces a `NoopMeter`. Invoking `CompositeCounter.count()` may result in either `0` from the `NoopMeter` or some other value from the other `Counter`.

The problem appears to be in `AbstractCompositeCounter.firstChild()` as it doesn't filter out `NoopMeters` when delegating to the child. Alternatively a `CompositeMeter` could avoid placing any `NoopMeter`s in its collection of child meters.

peterfouquet0001:
Please merge the PR to the main branch to fix the problem.
The problem still occurs for us.

PR #1442, opened 2019-05-30
Ignore NoopMeter in firstChild scan in CompositeMeters

jake-at-work:
Make CompositeMeters return a meter value if at least one of the child
Meters is not a NoopMeter.

See #1441


shakuzen:
Also, since it sounds like this is a bug and users should not be relying on this behavior, could you rebase the changes on the `1.7.x` branch so we can include it in upcoming maintenance releases?

PR #1443, opened 2019-05-30
Ignore NoopMeter from child Registry

jake-at-work:
If a child Registry returns a NoopMeter then don't add it to the map of
child Meters.

See #1441


pivotal-jbarrett:
I honestly can't recall why from nearly 3 years ago. I think it has to do with the registry holding references multiple references to {{NoopMeter}} instances. these two PRs are solutions to the same issue. Sorry I can't be of more help anymore. Since they aren't a concern for my downstream projects as far as I can tell anymore I suppose we can just close them both.

Issue #1444, opened 2019-05-30
Exception thrown from gauge stops scheduled task


engineal:
I am using micrometer with the CloudWatchMeterRegistry. I have registered a gauge like so:

`registry.gauge("my_repository.count", Tags.empty(), myRepository, MyRepository::count);`

The CloudWatchMeterRegistry, being a child of the PushMeterRegistry, schedules the publish task at a fixed rate. However, if `MyRepository::count` throws an exception, metrics stop being published until the application is restarted.

I have followed the exception and it appears to make it into the `CloudWatchMeterRegistry.publish()` function, which contains a try...finally block without a catch. Once the exception makes it to the scheduled executor service, it will no longer schedule the publish method anymore.

While I could make sure every gauge has a try...catch block, it would be cleaner if a catch block could be added to the `CloudWatchMeterRegistry.publish()` function.

izeye:
@engineal [Any `Throwable` from `PushMeterRegistry.publish()` should be caught in `PushMeterRegistry.publishSafely()`](https://github.com/micrometer-metrics/micrometer/blob/bbd4113990061f09d8a97696be38d5ebbbd912c0/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java#L43-L52) since #1228, so 1.1.4+ should be okay. Which version of Micromter are you using?

engineal:
Thanks for your comment, I am using 1.1.3. I will test 1.1.4 to verify that my bug is fixed there.

shakuzen:
@engineal if with the logging of the exception you've found an issue in Micrometer code, please let us know and we'll try to get it fixed.

Issue #1228, opened 2019-02-13
PushMeterRegistry implementations can silently stop sending metrics 


shakuzen:
See https://github.com/micrometer-metrics/micrometer/issues/1001#issuecomment-463023832 for what prompted me to create this issue. I see that previously we guarded against this in push-based registries at the time in cfbf1b0cbddc4575332218c25a7268b82cde3bf9. Now that we have the `PushMeterRegistry` abstraction, perhaps it is better that we apply this guard at that level, so it is consistent and new registry implementations do not miss this feature.

The risk of not having this is that when an uncaught exception happens, it is not logged and the `publish` task will no longer be scheduled. This makes it hard to notice/investigate these bugs. However, in cases like the issue in #1001, every publish would throw this error and it would never stop (nor would any metrics be published) since there was a bug in the logic preparing the data to publish. Logging a stack trace every publish forever seems a bit annoying for a user affected by this waiting for a fixed version (though they could update their setup to not enable/configure the registry once they realize this issue). Maybe there is a balance we could strike of catching and logging for this while not being too noisy?

juliannzta:
I 100% agree on this and ideally should be high priority. The risk of silently losing monitoring far outweighs the annoyance of logging an error every time the scheduler runs. Quitting the schedule means no recovery other than an app restart and as a result application monitoring is unavailable.

Perhaps to avoid too much log noise, it would be useful to log a ERROR or WARN if without the full stacktace, but also log the entire stacktrace with some informative information on a DEBUG level. Additional configuration options would also be helpful for this as you mention.

#1001  is an example of how that situation can make this library unusable for a production environment. I (and others as noted on that ticket) spent considerable time scratching heads and debugging through code to get any understanding of why metrics aren't being published. As a sort of side not, the statsd registry has a similar issue, it logs nothing on success or failure. I still never got to the bottom of why it never worked in my production environment, I eventually changed approach.

Thanks of your great work on this library!!

PyvesB:
We're using the CloudWatch implementation and suddenly observed that the metrics stopped being published. As the CloudWatch implementation relies on `PushMeterRegistry`, may the problem also be related to the issue described here?

Thanks for looking into this!

shakuzen:
[...]

PR #1445, opened 2019-05-30
Add catch statement to CloudWatchMeterRegistry.publish

engineal:
Fix for bug #1444

I added a catch statement to CloudWatchMeterRegistry.publish, so that in the case where a gauge throws an exception, the scheduled executor won't be halted.


engineal:
Fixed in 1.1.4+ by #1228 instead.

PR #1447, opened 2019-06-04
Declarative Annotation Based API for Counters

### LINKED ISSUE #1435: Declarative Annotation Based API for Counters
Synopsis
--------
Although using `Timer` and `DistributionSummary` we can publish a count of events in addition to other measurements, sometimes it just makes sense to use a plain `Counter`. 

For timers, in order to keep track of how long a particular task is running, we have two options:
 1. Manually starting and stopping the `Timer`.
 2. Declarative time tracking using the `@Timed`.

These two options are complementary to each other. That is, the declarative approach is there to abstract the cross-cutting concern in a nice and familiar enterprise-ish way but we still have the option to manually manage the `Timer`s when we can't use the declarative API.

_For plain `Counter`s, there is no such declarative approach._ 

Recurring Pattern
-------------------
It's a recurring theme to keep track of:
 - A total number of attempts for a particular task.
 - A number of successful attempts.
 - A Number of failed attempts.

For example. when I receive a job submission request, I increment the `job.submission.attempts` by one. If I managed to successfully submit the job, the `job.submission.successful` would be incremented by one. Otherwise, I'd increment the `job.submission.failed` metric.

Suggestion
------------
Alongside the current `Counter` API, we can provide a declarative API to manage counters. For example, we can offer an annotation, e.g. `@Counted`, as the entry point to such API:
```java
/**
 * We expose the following metrics for the methods annotated with this annotation:
 * <ul>
 *     <li>{@code metric.initialSuffix} to represent the number of times the method is executed</li>
 *     <li>{@code metric.succeedSuffix} to represent the number of successful executions.</li>
 *     <li>{@code metric.failedSuffix} to represent the number of failed executions.</li>
 * </ul>
 */
@Inherited
@Target(ElementType.METHOD)
@Retention(RetentionPolicy.RUNTIME)
public @interface Counted {

    /**
     * Determines the basic metric name, e.g. {@code job.submission}.
     * 
     * @return The prefix for all metrics.
     */
    String metric();

    /**
     * Determines the suffix for total attempts.
     * 
     * @return The total attempts suffix.
     */
    String initialSuffix() default "attempts";

    /**
     * Determines the suffix for successful attempts.
     * 
     * @return The successful attempts suffix.
     */
    String succeedSuffix() default "succeed";

    /**
     * Determines the suffix for failed attempts.
     *
     * @return The failed attempts suffix.
     */
    String failedSuffix() default "failed";
}
```

Then we can implement the counting logic using an aspect much like the `TimedAspect`:
```java
@Aspect
public class CountedAspect {

    private static final String EXCEPTION_TAG_NAME = "exception";
    
    private final MeterRegistry meterRegistry;

    public CountedAspect(MeterRegistry meterRegistry) {
        this.meterRegistry = meterRegistry;
    }
    
    @Around("@annotation(counted)")
    public Object interceptAndCount(ProceedingJoinPoint pjp, Counted counted) throws Throwable {
        try {
            Counter.builder(initialName(counted)).register(meterRegistry).increment();
            Object result = pjp.proceed();
            Counter.builder(successfulName(counted)).register(meterRegistry).increment();
            
            return result;
        } catch (Throwable e) {
            Counter.builder(failedName(counted))
                .tag(EXCEPTION_TAG_NAME, e.getClass().getSimpleName())
                .register(meterRegistry).increment();
                
            throw e;
        }
    }
    
    private String initialName(Counted counted) {
        return String.format("%s.%s", counted.metric(), counted.initialSuffix());
    }

    private String successfulName(Counted counted) {
        return String.format("%s.%s", counted.metric(), counted.succeedSuffix());
    }

    private String failedName(Counted counted) {
        return String.format("%s.%s", counted.metric(), counted.failedSuffix());
    }
}
```

If you guys are OK with the idea, I'm more than happy to provide a PR.

Cheers 🍻

alimate:
A few points:
 - I thought there is no use case for making the `Counted` annotation a *Repeatable* one. If you guys think otherwise, please let me know.
 - Same argument for the `extraTags`. Are they useful for `Counted`?

Resolves #1435

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
@@ -19,8 +19,9 @@ import java.lang.annotation.*;
 
 /**
  * Annotated methods would expose a few counter metrics about their execution status.
- * By default, only failure attempts would be reported. Switching the {@link #successfulAttempts()}
- * to {@code true} would instruct the corresponding aspect to record successful attempts, too.
+ * By default, both failed and successful attempts would be reported. Switching the
+ * {@link #successfulAttempts()} to {@code false} would instruct the corresponding aspect
+ * to only record the failed attempts.
  *
  * @author Ali Dehghani
  * @see io.micrometer.core.aop.CountedAspect
@@ -39,8 +40,8 @@ public @interface Counted {
     String value() default "";
 
     /**
-     * By default, successful attempts won't be recorded. Switch it to {@code true} in order to
-     * record successful attempts alongside with failed attempts.
+     * By default, both failed and successful attempts are recorded. Switch it to {@code false} in
+     * order to only record failed attempts.
      *
      * @return Whether to report successful attempts or not.
      */
--- a/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
@@ -31,11 +31,12 @@ import java.lang.annotation.*;
 public @interface Counted {
 
     /**
-     * Represents the metric name for the to-be-recorded counters.
+     * Represents the metric name for the to-be-recorded counters. The default value is
+     * {@code method.counted}.
      *
      * @return The metric name.
      */
-    String value();
+    String value() default "";
 
     /**
      * By default, successful attempts won't be recorded. Switch it to {@code true} in order to
--- a/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
@@ -20,7 +20,7 @@ import java.lang.annotation.*;
 /**
  * Annotated methods would expose a few counter metrics about their execution status.
  * By default, both failed and successful attempts would be reported. Switching the
- * {@link #successfulAttempts()} to {@code false} would instruct the corresponding aspect
+ * {@link #recordFailuresOnly()} to {@code true} would instruct the corresponding aspect
  * to only record the failed attempts.
  *
  * @author Ali Dehghani
@@ -40,12 +40,12 @@ public @interface Counted {
     String value() default "";
 
     /**
-     * By default, both failed and successful attempts are recorded. Switch it to {@code false} in
+     * By default, both failed and successful attempts are recorded. Switch it to {@code true} in
      * order to only record failed attempts.
      *
-     * @return Whether to report successful attempts or not.
+     * @return Whether to only report failures.
      */
-    boolean successfulAttempts() default true;
+    boolean recordFailuresOnly() default false;
 
     /**
      * An optional description for what the underlying counter is going to record.
--- a/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
@@ -1,0 +1,54 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.annotation;
+
+import java.lang.annotation.*;
+
+/**
+ * Annotated methods would expose a few counter metrics about their execution status.
+ * By default, only failure attempts would be reported. Switching the {@link #successfulAttempts()}
+ * to {@code true} would instruct the corresponding aspect to record successful attempts, too.
+ *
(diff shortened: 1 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    String value() default "";
+
+    /**
+     * By default, successful attempts won't be recorded. Switch it to {@code true} in order to
+     * record successful attempts alongside with failed attempts.
+     *
+     * @return Whether to report successful attempts or not.
+     */
+    boolean successfulAttempts() default false;
```
checketts: I think defaulting this to true would be better. My comment on #1435 about defaulting to false was about the 'submitted' count (which you have eliminated)

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -0,0 +1,127 @@
...
+        try {
+            Object result = pjp.proceed();
+            if (counted.successfulAttempts()) {
+                counter(pjp, counted).tag(STATUS_TAG, "succeed").register(meterRegistry).increment();
+            }
+
+            return result;
+        } catch (Throwable e) {
+            counter(pjp, counted)
+                    .tag(EXCEPTION_TAG, e.getClass().getSimpleName())
```
checketts: You can't have different tags for the same metric, so you'll need to add a blank `EXCEPTION_TAG` to the success counter

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -0,0 +1,127 @@
...
+                    .tag(STATUS_TAG, "failed")
```
checketts: Since `succeed` and `failed` aren't grammatically parallel, I would recommend changing them to be similar to the tags used in [CaffeineCacheMetrics](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java#L167): make the key `result` with `success`/`failure`

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
alimate: 🙈 My bad!

micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
alimate: Does it make sense to make this behavior configurable?

checketts: I could see that some users would want to `recordFailuresOnly`, if they only care about exceptions/failures. So I think there might be a good reason to have some configuration, but renaming it might be a good idea.

alimate: I've renamed to `recordFailuresOnly`, as you suggested.

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
alimate: When recording successful attempts, I tag the metric with `None` exception. Is it OK?

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
alimate: ✅

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param counted The annotation.
+     * @return Whatever the intercepted method returns.
+     * @throws Throwable When the intercepted method throws one.
+     */
+    @Around("@annotation(counted)")
+    public Object interceptAndRecord(ProceedingJoinPoint pjp, Counted counted) throws Throwable {
+        try {
+            Object result = pjp.proceed();
+            if (!counted.recordFailuresOnly()) {
+                record(pjp, counted, "None", "success");
```
checketts: Looks like the `@TimedAspect` used a lowercase `none`

alimate: ✅

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
checketts: Make it lowercase so it is similar to the `@TimedAspect`

alimate: Done!

checketts:
Anyone else want to look this over before merging? @jkschneider or @davidkarlsen (since you both contributed to the `@TimedAspect`)?

alimate:
@checketts Is it a good idea to add a `result` tag to `TimedAspect`? I know we can infer the execution status by the `exception` tag but using something like a `result` tag would simplify the process.

checketts:
Making them consistent wouldn't be a bad idea. Perhaps adding it in a separate PR would be a good idea so it doesn't distract the conversation in this PR.

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
shakuzen: It is unfortunate that `WebMvcTags` in the Spring support uses `None`, though. We should probably align these in 2.0.

PR #1449, opened 2019-06-05
AppOptics Implementation - Set metric time to clock time.

HunterSherms:
and provide the option to floor times to synchronize between nodes.

If no time is provided by the client AppOptics assumes the time upon receiving the metrics, which can lead to unwanted jitter in metrics.

This sets the time value to the nodes' clock time and provides the ability to floor the provided time to synchronize across nodes, which should help to fix charts like this:

### DIFF
--- a/implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsConfig.java
+++ b/implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsConfig.java
@@ -65,6 +65,13 @@ public interface AppOpticsConfig extends StepRegistryConfig {
         return v == null ? "https://api.appoptics.com/v1/measurements" : v;
     }
 
+    /**
+     * @return whether or not to ship a floored time - synchronizes reporting across nodes
+     */
+    default boolean floorTimes() {
+        return get(prefix() + ".floorTimes") == "true";
+    }
+
     @Override
     default int batchSize() {
         String v = get(prefix() + ".batchSize");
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsConfig.java
```diff
@@ -65,6 +65,13 @@ default String uri() {
         return v == null ? "https://api.appoptics.com/v1/measurements" : v;
     }
 
+    /**
+     * @return whether or not to ship a floored time - synchronizes reporting across nodes
+     */
+    default boolean floorTimes() {
+        return get(prefix() + ".floorTimes") == "true";
```
shakuzen: For more lenient and conventional parsing, I would suggest the following.
```suggestion
        return Boolean.parseBoolean(get(prefix() + ".floorTimes"));
```

HunterSherms:
@shakuzen https://github.com/micrometer-metrics/micrometer/pull/1449/commits/0f6cd15dfa2e2a35476c26b441bc0579de17fc7e

shakuzen:
For reference of anyone looking through this pull request, here is the AppOptics documentation describing the `time` property and this flooring technique: https://docs.appoptics.com/api/#measurements

> `time` - Unix Time (epoch seconds). This defines the time that a measurement  is recorded at. It is useful when sending measurements from multiple  hosts to align them on a given time boundary, eg. time=floor(Time.now,  60) to align samples on a 60 second tick.

PR #1450, opened 2019-06-06
Add Timer count to Stackdriver registry

### LINKED ISSUE #1371: Timer counts not published to Stackdriver
The [Micrometer documentation](https://micrometer.io/docs/concepts#_counters) recommends

> Never count something you can time with a Timer or summarize with a DistributionSummary! Both Timer and DistributionSummary always publish a count of events in addition to other measurements.

However, the Stackdriver backend doesn't seem to publish timer counts. I can't find them in Stackdriver and I don't see where they would be created in the code.

Can we add an additional Stackdriver metric (`my.timer.count`) when we publish timer durations?

ormanli:
Timer already supports count of measured duration. Added another TimeSeries to feed Stackdriver.

Fixes #1371


shakuzen:
@ormanli thank you for the pull request. This will be included in the upcoming 1.1.5 release.

PR #1453, opened 2019-06-10
Prevent race condition when updating tags for StatsD Datadog flavor

jlisam:
This PR prevents race condition when updating tags for StatsD Datadog flavor. 

Similar to https://github.com/micrometer-metrics/micrometer/issues/1389, we have noticed unusual behavior when multiple threads interact with our counter. Although we can't prove that it is indeed a race condition yet, in https://github.com/micrometer-metrics/micrometer/pull/1391, the author mentions updates to `DatadogStatsdLineBuilder` which were never done. 

> And I noticed DatadogStatsdLineBuilder has the same concern, so if this approach is okay, I'll prepare a similar one for it.


shakuzen:
@jlisam Thank you for the pull request. I have rebased it on the `1.0.x` branch so that we can easily include it in the final 1.0.x release and the 1.1.5 release.

Issue #1454, opened 2019-06-13
HazelcastCacheMetrics latency values reported as nanos where they are retrieved as millis


wneild:
HazelcastCacheMetrics latency values are reported as nanos where Hazelcast (versions >= 3.9) stats report them as milliseconds.

e.g. 
```
cache -> cache.getLocalMapStats().getTotalGetLatency(), TimeUnit.NANOSECONDS)
```
Simply need to correct the TimeUnit to milliseconds but this will break for anyone using an old version of Hazelcast.

shakuzen:
Do you mean that Hazelcast changed the units it returns for these stats, in a non-backwards compatible way? If so, could you link to something from Hazelcast that mentions this change?
Or do you mean that the units were always milliseconds and the code in Micrometer has always been wrong?

wneild:
@shakuzen I'm not entirely clear what the case was before 3.9 but the best reference I could find is that they changed the code off of this pull request and comment explicitly describes a change from nanos to millis being reported https://github.com/hazelcast/hazelcast/pull/10930#issuecomment-315681738. They haven't documented it as a breaking change as far as I can tell even though IMO it looks like it should merit it.

izeye:
If I read the reference PR correctly, it seems to change its internal time resolution from milliseconds to nanoseconds. Hence, there seems to be no change on getters for latencies which `HazelcastCacheMetrics` depends on. They seem to be in milliseconds from the beginning.

I created #1456 to resolve it.

PR #1456, opened 2019-06-14
Fix time unit for latencies in HazelcastCacheMetrics

izeye:
This PR fixes time unit for latencies in `HazelcastCacheMetrics`.

Closes gh-1454


checketts:
Should we upgrade the version of hazelcast to match the breaking change version?

Or add a way for older hazelcast users to opt out of the unit change?

izeye:
@checketts If I understand it correctly, it is a bug and this PR fixes it. The time unit seems to be milliseconds from the beginning and `HazelcastCacheMetrics` seems to interpret it wrongly.

See https://github.com/micrometer-metrics/micrometer/issues/1454#issuecomment-502299656

shakuzen:
I know it will break graphs depending on the units being nanoseconds now, but since this is a bug and the nanoseconds reported now are wrong, I think we should change this for `1.0.x` also. Does anyone disagree?

izeye:
@shakuzen Oop, I created this PR from `1.0.x` branch but forgot to select it when creating this. I changed the base branch from `master` to `1.0.x` now.

Issue #1458, opened 2019-06-18
Support OpenJ9 GC policies


jschmied:
OpenJ9 (Java11b04) uses different names for GC:
nursery-allocate/nursery-survivor/tenured-SOA/tenured-LOA

Not sure how this fits into the Old/New GcGenerationAge.

https://www.eclipse.org/openj9/docs/gc/

jschmied:
I did a rough implementation of JvmGcMetrics for OpenJ9.
[JvmGcMetrics.java.txt](https://github.com/micrometer-metrics/micrometer/files/3301460/JvmGcMetrics.java.txt)

izeye:
@jschmied Thanks for the suggestion!

I have no experience on Eclipse OpenJ9 but just based on the doc you shared, "nursery area" seems to belong to "young generation" and "tenured area" seems to belong to "old generation". I think it'd be better to map them to the current corresponding concepts if possible rather than creating a dedicated implementation for it.

jschmied:
Hmm, the behavior is different. 
- there are two old pools (small/large)
- there is no "ConcurrentPhase"
Even if we map both GC to same counter/gauges, we have to switch between implementations. This could be done by the memory-pool names. Maybe we can have a shared set of counter/gauges and if a gc has special values the can create them additionally.

izeye:
@jschmied I didn't look into `JvmGcMetrics` in details but there are similar discrepancies among various garbage collectors in Oracle's JVM and I expect it covers them, so it might be possible to fit Eclipse OpenJ9 into it somehow.

jschmied:
"jvm.gc.max.data.size" gives back -1 others work

PascalSchumacher:
It would be great if at least "jvm.gc.live.data.size.bytes" was added.

DanHeidinga:
I'm excited to see the interest in support OpenJ9 in micrometer.  As part of the OpenJ9 developer community, I can help out with this.

PR #1460, opened 2019-06-18
Add support for OpenJ9 gc (#1458)

jschmied:
See ticket #1458

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -34,6 +34,8 @@ import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
 import java.lang.management.MemoryPoolMXBean;
 import java.lang.management.MemoryUsage;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -62,8 +64,7 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
     @Nullable
     private String youngGenPoolName;
 
-    @Nullable
-    private String oldGenPoolName;
+    private Collection<String> oldGenPoolNames = new ArrayList<>();
 
     private final List<Runnable> notificationListenerCleanUpRunnables = new CopyOnWriteArrayList<>();
 
@@ -76,7 +77,7 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
             if (isYoungGenPool(mbean.getName()))
                 youngGenPoolName = mbean.getName();
             if (isOldGenPool(mbean.getName()))
-                oldGenPoolName = mbean.getName();
+                oldGenPoolNames.add(mbean.getName());
         }
         this.tags = tags;
     }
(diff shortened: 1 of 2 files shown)


shakuzen:
Thank you for this. Would you be willing to make some unit tests for this? You can look at `ProcessorMetricsTest` to see a way we make a test conditional on being run on OpenJ9. Maybe it is time we add a build with OpenJ9.

jschmied:
Added test

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -203,11 +203,11 @@ private boolean isConcurrentPhase(String cause) {
     }
 
     private boolean isOldGenPool(String name) {
-        return name.endsWith("Old Gen") || name.endsWith("Tenured Gen");
+        return name.endsWith("Old Gen") || name.endsWith("Tenured Gen") || name.equals("tenured-SOA");
```
mweirauch: If I am reading the referenced documentation right, "tenured-LOA" should also be considered. As objects are either moved from "nursery" to SOA or LOA if the exceed a certain size. Please correct me if I am wrong.

> Within the tenure area, new objects are allocated into the small object area (SOA), which is illustrated in the earlier diagram (see 3.). A large object area (LOA) is set aside for objects greater than 64 KB that cannot be allocated into the SOA to minimize fragmentation.

This would mean though, that we just can't keep the 'oldGenPoolName' and use it, but rather query more then pool name from the 'GCInfo.getMemoryUsage*Gc()' maps on every garbage collection notification. (Again, under the assumption the both, SOA and LOA make up the metric for "promoted" bytes.)

jschmied: Yes you are right of course. I just read the LOA area since I did not like to rewrite the whole class. LOA is more important, actually in my applications (average spring boot stuff) SOA is 0 bytes...
To be 100% correct we would have to consider multiple old gen pools.

I'll see if I can implement this without changing to much.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -143,21 +144,39 @@ public void bindTo(MeterRegistry registry) {
...
+                                    }
+
+                                    // Some GC implementations such as G1 can reduce the old gen size as part of a
+                                    // minor GC. To track the live data size we record the value if we see a
+                                    // reduction in the old gen heap size or after a major GC.
+                                    if (oldAfter < oldBefore || GcGenerationAge.fromName(notificationInfo.getGcName()) == GcGenerationAge.OLD) {
+                                        isOldGenGarbageCollected = true;
+                                        final long oldMaxAfter = after.get(oldGenPoolName).getMax();
+
+                                        if (liveDataSizeValue > 0) {
```
shakuzen: This condition will always be false because the variable is initialized with 0 and only set inside this conditional.

shakuzen (commented):
@jschmied according to the documentation shared, there are several different GC policies and collectors listed for OpenJ9. Which are covered by these changes?

shakuzen:
Closing as this has been superseded by #2562 and #2514 now.

PR #1461, opened 2019-06-18
Add support for per-query metrics in Hibernate statistics(#1006)

### LINKED ISSUE #1006: Add support for per-query metrics in Hibernate statistics
Current implementation of **HibernateMetrics** exposed counters & also time gauges for max execution queries, however no tags are provided in order to know about which query those metrics are related.

I believe this could be very useful and could be done using current Hibernate **Statistics** class implementation.

For example the TimeGauge **hibernate.query.executions.max** could be tag with a **query** tag and the value could be provided by `Statistics.getQueryExecutionMaxTimeQueryString()`.

Also the other counter metrics could be retrieved by query using:

- `Statistics.getQueries()`
- `Statistics.getQueryStatistics(String query)`

What do u think ?

pawel2401:
Resolves #1006

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
@@ -295,7 +295,7 @@ public class HibernateMetrics implements MeterBinder {
             for (String query : statistics.getQueries()) {
                 QueryStatistics queryStatistics = statistics.getQueryStatistics(query);
                 if (Objects.nonNull(queryStatistics)) {
-                    String queryName = query.replace(" ", "_").toLowerCase();
+                    String queryName = query.replace(" ", "_").replace(".", "_").toLowerCase();
                     if (Objects.isNull(Search.in(meterRegistry).tags("query", queryName).functionCounter())) {
 
                         FunctionCounter.builder("hibernate.query.cache.requests", queryStatistics, QueryStatistics::getCacheHitCount)
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
@@ -17,14 +17,20 @@ package io.micrometer.core.instrument.binder.jpa;
 
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.binder.MeterBinder;
+import io.micrometer.core.instrument.search.Search;
 import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
 import io.micrometer.core.lang.Nullable;
 import org.hibernate.SessionFactory;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.event.service.spi.EventListenerRegistry;
+import org.hibernate.event.spi.*;
+import org.hibernate.stat.QueryStatistics;
 import org.hibernate.stat.Statistics;
 
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.PersistenceException;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import java.util.function.ToDoubleFunction;
 
@@ -35,6 +41,7 @@ import java.util.function.ToDoubleFunction;
  * @author Marten Deinum
  * @author Jon Schneider
  * @author Johnny Lim
+ * @author Pawel Stepien
  */
 @NonNullApi
 @NonNullFields
@@ -46,6 +53,9 @@ public class HibernateMetrics implements MeterBinder {
 
     @Nullable
     private final Statistics statistics;
+
+    @Nullable
+    private final SessionFactory sessionFactory;
 
     /**
      * Create {@code HibernateMetrics} and bind to the specified meter registry.
@@ -109,6 +119,7 @@ public class HibernateMetrics implements MeterBinder {
         this.tags = Tags.concat(tags, SESSION_FACTORY_TAG_NAME, sessionFactoryName);
         Statistics statistics = sessionFactory.getStatistics();
         this.statistics = statistics.isStatisticsEnabled() ? statistics : null;
+        this.sessionFactory = sessionFactory;
     }
 
     /**
@@ -122,6 +133,7 @@ public class HibernateMetrics implements MeterBinder {
     public HibernateMetrics(EntityManagerFactory entityManagerFactory, String entityManagerFactoryName, Iterable<Tag> tags) {
         this.tags = Tags.concat(tags, SESSION_FACTORY_TAG_NAME, entityManagerFactoryName);
         SessionFactory sessionFactory = unwrap(entityManagerFactory);
+        this.sessionFactory = sessionFactory;
         if (sessionFactory != null) {
             Statistics statistics = sessionFactory.getStatistics();
             this.statistics = statistics.isStatisticsEnabled() ? statistics : null;
@@ -136,10 +148,10 @@ public class HibernateMetrics implements MeterBinder {
         }
 
         FunctionCounter.builder(name, statistics, f)
-            .tags(tags)
-            .tags(extraTags)
-            .description(description)
-            .register(registry);
+                .tags(tags)
+                .tags(extraTags)
+                .description(description)
+                .register(registry);
     }
 
     @Override
@@ -154,31 +166,31 @@ public class HibernateMetrics implements MeterBinder {
 
         // Transaction statistics
         counter(registry, "hibernate.transactions", "The number of transactions we know to have been successful",
-            Statistics::getSuccessfulTransactionCount, "result", "success");
+                Statistics::getSuccessfulTransactionCount, "result", "success");
         counter(registry, "hibernate.transactions", "The number of transactions we know to have failed",
-            s -> s.getTransactionCount() - s.getSuccessfulTransactionCount(), "result", "failure");
+                s -> s.getTransactionCount() - s.getSuccessfulTransactionCount(), "result", "failure");
         counter(registry, "hibernate.optimistic.failures", "The number of StaleObjectStateExceptions that have occurred",
-            Statistics::getOptimisticFailureCount);
+                Statistics::getOptimisticFailureCount);
 
         counter(registry, "hibernate.flushes", "The global number of flushes executed by sessions (either implicit or explicit)",
-            Statistics::getFlushCount);
+                Statistics::getFlushCount);
         counter(registry, "hibernate.connections.obtained", "Get the global number of connections asked by the sessions " +
-            "(the actual number of connections used may be much smaller depending " +
-            "whether you use a connection pool or not)", Statistics::getConnectCount);
+                "(the actual number of connections used may be much smaller depending " +
+                "whether you use a connection pool or not)", Statistics::getConnectCount);
 
         // Statements
         counter(registry, "hibernate.statements", "The number of prepared statements that were acquired",
-            Statistics::getPrepareStatementCount, "status", "prepared");
+                Statistics::getPrepareStatementCount, "status", "prepared");
         counter(registry, "hibernate.statements", "The number of prepared statements that were released",
-            Statistics::getCloseStatementCount, "status", "closed");
+                Statistics::getCloseStatementCount, "status", "closed");
 
         // Second Level Caching
         counter(registry, "hibernate.second.level.cache.requests", "The number of cacheable entities/collections successfully retrieved from the cache",
-            Statistics::getSecondLevelCacheHitCount, "result", "hit");
+                Statistics::getSecondLevelCacheHitCount, "result", "hit");
         counter(registry, "hibernate.second.level.cache.requests", "The number of cacheable entities/collections not found in the cache and loaded from the database",
-            Statistics::getSecondLevelCacheMissCount, "result", "miss");
+                Statistics::getSecondLevelCacheMissCount, "result", "miss");
         counter(registry, "hibernate.second.level.cache.puts", "The number of cacheable entities/collections put in the cache",
-            Statistics::getSecondLevelCachePutCount);
+                Statistics::getSecondLevelCachePutCount);
 
         // Entity information
         counter(registry, "hibernate.entities.deletes", "The number of entity deletes", Statistics::getEntityDeleteCount);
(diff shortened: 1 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -250,4 +264,97 @@ private SessionFactory unwrap(EntityManagerFactory entityManagerFactory) {
...
+                                .description("The number of putting the query into cache")
+                                .register(meterRegistry);
+
+                        FunctionCounter.builder("hibernate.query.executions", queryStatistics, QueryStatistics::getExecutionCount)
+                                .tags(tags)
+                                .tags("query", queryName)
+                                .description("Total number of query executions")
+                                .register(meterRegistry);
+
+                        TimeGauge.builder("hibernate.query.executions.total", queryStatistics, TimeUnit.MILLISECONDS, QueryStatistics::getExecutionTotalTime)
```
checketts: Can you combine `hibernate.query.executions.total` and `hibernate.query.executions` into a `FunctionTimer`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -250,4 +264,97 @@ private SessionFactory unwrap(EntityManagerFactory entityManagerFactory) {
...
+                                .description("Query maximum execution time")
+                                .register(meterRegistry);
+
+                        TimeGauge.builder("hibernate.query.executions.min", queryStatistics, TimeUnit.MILLISECONDS, QueryStatistics::getExecutionMinTime)
+                                .tags(tags)
+                                .tags("query", queryName)
+                                .description("Query minimum execution time")
+                                .register(meterRegistry);
+
+                        FunctionCounter.builder("hibernate.query.executions.rows", queryStatistics, QueryStatistics::getExecutionRowCount)
```
checketts: Is this a counter (meaning this number only goes up)? Or is it a gauge of the latest rows returned by a query?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -250,4 +264,97 @@ private SessionFactory unwrap(EntityManagerFactory entityManagerFactory) {
...
+        }
+
+        @Override
+        public void onPostLoad(PostLoadEvent event) {
+            registerQueryMetric(event.getSession().getFactory().getStatistics());
+        }
+
+        void registerQueryMetric(Statistics statistics) {
+            for (String query : statistics.getQueries()) {
+                QueryStatistics queryStatistics = statistics.getQueryStatistics(query);
```
checketts: Is there any concern about this `queryStatistics` object being garbage collected? (since Micrometer favors weak references it won't keep the object around on its own)

At least that is how gauges work, not certain about `FunctionCounter`s

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
pawel2401: Yes, this is the total cumulatively increasing number of rows returned by specific query, so I think counter is ok

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
pawel2401: The hibernate StatisticsImpl are keeping references to QueryStatistics in the Map, the entries aren't removed in any point of time. And since we are obtaining Statistics directly from SessionFactory which is created once during startup there no need to worry about their garbage collection, but it was good point didn't think about checkin' it before!

`public class StatisticsImpl implements StatisticsImplementor, Service {
      
private final ConcurrentMap<String, QueryStatisticsImpl> queryStatsMap = new ConcurrentHashMap();
}`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
pawel2401: Yup, sure will adjust

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -295,7 +295,7 @@ void registerQueryMetric(Statistics statistics) {
             for (String query : statistics.getQueries()) {
                 QueryStatistics queryStatistics = statistics.getQueryStatistics(query);
                 if (Objects.nonNull(queryStatistics)) {
-                    String queryName = query.replace(" ", "_").toLowerCase();
+                    String queryName = query.replace(" ", "_").replace(".", "_").toLowerCase();
```
checketts: Dots are OK in tags, so they shouldn't cause a problem (at least in Prometheus).

pawel2401: Graphite is treating dots as next tag

checketts: Could you create an issue about the graphite registry not escaping dots in tags?

pawel2401: Sure, will create

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -47,6 +54,9 @@
     @Nullable
     private final Statistics statistics;
 
+    @Nullable
```
shakuzen: Is this Nullable? If the `sessionFactory` passed in the constructor is `null` then a NPE will happen: `Statistics statistics = sessionFactory.getStatistics();`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
shakuzen: If the intent is to generically make the query fit the convention of the MeterRegistry, the NamingConvention configured for the MeterRegistry can be used: `meterRegistry.config().namingConvention().tagValue(query)`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -250,4 +264,91 @@ private SessionFactory unwrap(EntityManagerFactory entityManagerFactory) {
...
+        @Override
+        public void onPostLoad(PostLoadEvent event) {
+            registerQueryMetric(event.getSession().getFactory().getStatistics());
+        }
+
+        void registerQueryMetric(Statistics statistics) {
+            for (String query : statistics.getQueries()) {
+                QueryStatistics queryStatistics = statistics.getQueryStatistics(query);
+                if (Objects.nonNull(queryStatistics)) {
+                    String queryName = query.replace(" ", "_").replace(".", "_").toLowerCase();
```
shakuzen: What kind of cardinality can be expected for this? I suppose it depends on whether bind variables are used and if the query is somehow normalized.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
pawel2401: You're right. WIll change to @NonNull

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
pawel2401: You're asking for example? It can be:
`Select generatedAlias0 from Table as generatedAlias0 where generatedAlias0.param0 :val0`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
pawel2401: //EDIT
This field is Nullable when using the other Deprecated Constructor. NPE in this constructor won't happen because of NonNullApi annotation.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
shakuzen: I mean how many unique values for this queryName can we expect in applications that use Hibernate? I'm thinking it might be a lot and if that is the case, maybe this should be an opt-in feature that can be turned on for those that want to track query-level metrics.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -250,4 +264,91 @@ private SessionFactory unwrap(EntityManagerFactory entityManagerFactory) {
         }
     }
 
+    /**
+     * Register hibernate loadEvent listener - event trigger the metrics registration,
+     * only select queries are recorded in QueryStatistics {@link org.hibernate.stat.spi.StatisticsImplementor}
+     * @param meterRegistry meterRegistry to use
+     */
+    private void registerListeners(MeterRegistry meterRegistry) {
+        if (Objects.nonNull(sessionFactory) && sessionFactory instanceof SessionFactoryImplementor) {
```
shakuzen: nit: `instanceof` works as a null check so the `Objects.nonNull(sessionFactory)` is not necessary.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
pawel2401: @shakuzen I think that it would make sense to make it opt-in feature, since number of these metrics can be quite high

pawel2401: @shakuzen Do you have any strategy for making a metrics configurable? So far I've seen that all are enabled by default without any additional configuration

shakuzen: Adding a boolean flag to the constructor would be one way, but there are already a lot of arguments in the constructor and this doesn't scale if we wanted more opt-in metrics, so I don't think we should go this route. Making it a separate `MeterBinder` is probably the easiest way to make it not automatically included with the current `MeterBinder`.

pawel2401: Moved it to separate MeterBinder

pawel2401:
@shakuzen are you fine with the changes? Should I change smth else?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
```diff
@@ -0,0 +1,181 @@
...
+                                .description("The number of putting the query into cache")
+                                .register(meterRegistry);
+
+                        FunctionTimer.builder("hibernate.query.executions.total", queryStatistics, QueryStatistics::getExecutionCount, QueryStatistics::getExecutionTotalTime, TimeUnit.MILLISECONDS)
+                                .tags(tags)
+                                .tags("query", queryName)
+                                .description("Function tracked total number of query executions during time")
+                                .register(meterRegistry);
+
+                        TimeGauge.builder("hibernate.query.executions.avg", queryStatistics, TimeUnit.MILLISECONDS, QueryStatistics::getExecutionAvgTime)
```
shakuzen: This is redundant because the `FunctionTimer` can already provide the average or it can be derived from the totalTime / count.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
```diff
@@ -0,0 +1,181 @@
...
+                                .tags(tags)
+                                .tags("query", queryName)
+                                .description("Query average execution time")
+                                .register(meterRegistry);
+
+                        TimeGauge.builder("hibernate.query.executions.max", queryStatistics, TimeUnit.MILLISECONDS, QueryStatistics::getExecutionMaxTime)
```
shakuzen: Out of curiosity, is this the max for the lifetime of the application or does it decay at all?

shakuzen: This will clash with the metric by the same name in `HibernateMetrics`: https://github.com/micrometer-metrics/micrometer/blob/516f9c51bd6f4ed7e7b418a98c04a2d5c432f940/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java#L216-L219

If we have per query metrics, we don't need the overall metric from `HibernateMetrics`, but we won't know if both binders are going to be used or not.

pawel2401: Max for the lifetime of the SessionFactory - which is created once on startup of the application

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
pawel2401: Yup, that was my intention, since graphite problem is resolved we can proceed with this generic approach

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
pawel2401: Won't this clash be resolved by query tag?
The general metrics for hibernate will be put under:
hibernate/query/executions/max
The query metrics will be put under:
hibernate/query/executions/max/QUERY_TAG

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
pawel2401: Removed

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
```diff
@@ -0,0 +1,181 @@
...
+                                .description("The number of query cache hits")
+                                .register(meterRegistry);
+
+                        FunctionCounter.builder("hibernate.query.cache.requests", queryStatistics, QueryStatistics::getCacheMissCount)
+                                .tags(tags)
+                                .tags("result", "miss", "query", queryName)
+                                .description("The number of query cache miss")
+                                .register(meterRegistry);
+
+                        FunctionCounter.builder("hibernate.query.cache.puts", queryStatistics, QueryStatistics::getCacheMissCount)
```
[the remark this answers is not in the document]
pawel2401: Changed

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
shakuzen: Having the same metric name with a different set of tag names is not allowed with the Prometheus backend, at least. I need to double check if that restriction is still necessary as there was a suggestion it may no longer be.

pawel2401: Okey, let's resolve this clash then. I've changed .executions -> .execution which kinda make sense since it is a single query.

pawel2401:
@shakuzen Anything else do you see that would be good to change?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
```diff
@@ -0,0 +1,175 @@
...
+        public void onPostLoad(PostLoadEvent event) {
+            registerQueryMetric(event.getSession().getFactory().getStatistics());
+        }
+
+        void registerQueryMetric(Statistics statistics) {
+            for (String query : statistics.getQueries()) {
+                QueryStatistics queryStatistics = statistics.getQueryStatistics(query);
+                if (Objects.nonNull(queryStatistics)) {
+                    String queryName = meterRegistry.config().namingConvention().tagValue(query);
+                    if (Objects.isNull(Search.in(meterRegistry).tags("query", queryName).functionCounter())) {
```
shakuzen: Is the point of this check because there could be multiple of the same query returned by `statistics.getQueries()`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
```diff
@@ -0,0 +1,175 @@
...
+import org.hibernate.event.spi.PostLoadEventListener;
+import org.hibernate.stat.QueryStatistics;
+import org.hibernate.stat.Statistics;
+
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A {@link MeterBinder} implementation that provides Hibernate query metrics. It exposes the
+ * same statistics as would be exposed when calling {@link Statistics#getQueryStatistics(String)}.
```
shakuzen: We should have a warning about potential high cardinality depending on the variety of queries one's application executes. I can add this as a polish commit post merge.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public void bindTo(MeterRegistry registry) {
+
+        registerListeners(registry);
+
+    }
+
+    /**
+     * Register hibernate loadEvent listener - event trigger the metrics registration,
+     * only select queries are recorded in QueryStatistics {@link org.hibernate.stat.spi.StatisticsImplementor}
```
shakuzen: If we think it would be helpful to users to have this information, we should add it to one of the public methods or the class JavaDoc. On a private method, it will not be noticed by most users.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetricsTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private MeterRegistry registry;
+
+    @BeforeEach
+    void setup() {
+        registry = new SimpleMeterRegistry(SimpleConfig.DEFAULT, new MockClock());
+    }
+
+    private HibernateQueryMetrics.MetricsEventHandler createMetricsEventHandlerMock() {
+        HibernateQueryMetrics hibernateQueryMetricsMock = Mockito.mock(HibernateQueryMetrics.class);
```
shakuzen: It feels like we're mocking too much when we are even mocking the class under test. I will try to rework the tests a bit to get something that feels closer to real usage. We could even make tests that use an in-memory database and normal Hibernate with that. This can be done post-merge, though.

shakuzen: After looking at this more closely, I see we're essentially testing the `MetricsEventHandler` in isolation. That's fine for this. I'll open an issue to add tests with not-mocked Hibernate classes.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
pawel2401: Emm If I remember correctly there should not, it is a little bit redundant check

PR #1463, opened 2019-06-19
Polish JvmGcMetrics

izeye:
This PR polishes `JvmGcMetrics`.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -111,81 +111,84 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
             .description("Incremented for an increase in the size of the young generation memory pool after one GC to before the next")
             .register(registry);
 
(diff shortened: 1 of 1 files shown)


izeye:
@mweirauch Thanks for the feedback! I like the suggestion and applied it in 9024b70e9d79dc306a022d2325880413915ad333.

For the `this.managementExtensionsPresent` guard when I was preparing this PR I was thinking moving it up on the top of the method so that no metrics are registered but I didn't include it initially as it seems more than polishing. If it's okay or preferred to be included here, I can push another commit for it. Otherwise, I'll create another PR for it.

mweirauch:
@izeye I would keep it as is, as not registering them would be too much of a change. Perhaps it's worthwhile considering this for a `2.0`.

We already have the case of 0-value-metrics with present mangement extension on a Hotspot VM with ZGC. So in this case we register the meters, but would still need to add support collecting the right metrics. It's the same status-quo.

Thanks for considering my proposal.

izeye:
@mweirauch Thanks for the additional feedback! I thought making it no-op when it's not supported is better. I'll prepare a dedicate PR for it for further discussion as it'll change the current behavior anyway.

PR #1469, opened 2019-06-21
Adding Custom Device Group name property to Dynatrace.

### LINKED ISSUE #1464: Support Dynatrace custom device group name property.
I see the custom metrics being reported from spring application to dynatrace under the name which I mentioned as device id `management.metrics.export.dynatrace.device-id=@project.artifactId@` but it is under some random custom device group name.

image

Can we add a property in DynatraceConfig class to set the group name like `management.metrics.export.dynatrace.group=mycustomgroup`?

I am cool with issuing a PR.

s425267:
The group name property helps us to add custom device group name in Dynatrace from micrometer.

Resolves #1464

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
@@ -15,7 +15,14 @@
  */
 package io.micrometer.dynatrace;
 
-import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.FunctionTimer;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Measurement;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Timer;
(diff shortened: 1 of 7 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
```diff
@@ -265,8 +274,13 @@ private void postCustomMetricValues(String type, List<DynatraceTimeSeries> timeS
     }
 
     // VisibleForTesting
-    List<DynatraceBatchedPayload> createPostMessages(String type, List<DynatraceTimeSeries> timeSeries) {
-        final String header = "{\"type\":\"" + type + '\"' + ",\"series\":[";
+    List<DynatraceBatchedPayload> createPostMessages(String type, String group, List<DynatraceTimeSeries> timeSeries) {
+        final String header;
+        if (StringUtils.isNotBlank(group)) {
```
jkschneider: Just a suggestion: this could be simplified with a ternary `StringUtils.isNotBlank(group) ? ... : ""` inside of the string concatenation so the other parts of the header aren't repeated. Maybe less chance of drift later.

s425267: That's a good suggestion. Made the changes. Thanks for reviewing @shakuzen @jkschneider

jkschneider: Thanks for the quick turnaround!

PR #1475, opened 2019-06-26
Make PushMeterRegistry work without slf4j-api

izeye:
`PushMeterRegistry` has references to classes (`Logger` and `LoggerFactory`) from `slf4j-api`, so without the dependency, the following error occurs:

```
java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory

	at io.micrometer.core.instrument.push.PushMeterRegistry.<clinit>(PushMeterRegistry.java:30)
```

This PR changes to make `PushMeterRegistry` work without `slf4j-api` by falling back to the standard output. If this fallback is not appropriate, another option might be logging nothing. I'm not sure if there's any better option for this. Maybe the easiest option might be just using a specific logging library like SLF4J explicitly.

`micrometer-core` has two more references in `JvmGcMetrics` and `MicrometerMetricsPublisherCommand`. If this PR finds a reasonable way, I'll prepare another PR for them or can update this to accommodate them.

I can see `implementations` has many references, so they already depends on SLF4J. So making Micrometer depend on SLF4j seems to be the easiest fix overall but IIRC @jkschneider doesn't want Micrometer to depend on any logging library.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -18,8 +18,8 @@ package io.micrometer.core.instrument.push;
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.lang.Nullable;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import io.micrometer.core.util.internal.logging.InternalLogger;
+import io.micrometer.core.util.internal.logging.InternalLoggerFactory;
 
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -27,7 +27,7 @@ import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
 public abstract class PushMeterRegistry extends MeterRegistry {
-    private final static Logger logger = LoggerFactory.getLogger(PushMeterRegistry.class);
+    private final static InternalLogger logger = InternalLoggerFactory.getInstance(PushMeterRegistry.class);
     private final PushRegistryConfig config;
 
     @Nullable
--- a/micrometer-core/src/main/java/io/micrometer/core/util/internal/logging/AbstractInternalLogger.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/util/internal/logging/AbstractInternalLogger.java
@@ -1,0 +1,255 @@
+/**
+ * Copyright 2019 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * Copyright 2012 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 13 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -85,4 +85,17 @@ public void close() {
...
+
+    private static void warn(String message, Throwable t) {
+        try {
+            LoggerFactory.getLogger(PushMeterRegistry.class).warn(message, t);
+        }
+        catch (Throwable ex) {
+            StringWriter writer = new StringWriter();
+            t.printStackTrace(new PrintWriter(writer));
+            String stackTrace = writer.toString();
+            System.out.println(message + System.lineSeparator() + stackTrace);
```
jkschneider: Do we want to just go to no-op here? JUL?

izeye: @jkschneider Thanks for the feedback!

Looking at the [Netty's implementation for internal loggers](https://github.com/netty/netty/blob/78adeb540833251c3bf31b5c2d92cb3343c84461/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java#L41-L61), it has the following precedence for logging frameworks: SLF4J > Log4J > Log4J2 > JUL

I think we can have a simplified version of it for now, for example: SLF4J > JUL

Please let me know which direction you prefer. Then I'll update this accordingly.

izeye:
> I can see `implementations` has many references, so they already depends on SLF4J.

I confirmed that `implementations` using SLF4J classes all depend on `slf4j-api` which is not optional unlike `micrometer-core`.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: Something like Netty's delegation to an available logging framework is preferable to `System.out.println`, I think. Let's go that direction.

shakuzen: Also starting with a simplified version that does SLF4J > JUL as suggested sounds good to me. Do you want to update this pull request with that, or shall we make a separate issue to track it, @izeye ?

izeye: @shakuzen Thanks for the feedback! I'd like to copy and paste the code for internal logging in Netty rather than reinvent the wheel as it seems to be complete and well written already. IANAL, but both Netty and Micrometer are licensed under the Apache License, Version 2.0, so it seems okay. Is it okay to go this route?

shakuzen: IANAL either, but I believe that's okay. As a courtesy and for attribution, we can mention in the JavaDocs that it was taken from the Netty project.

izeye: @shakuzen Thanks for the quick feedback 👍 I'll update this PR with a simplified version (SLF4J > JUL) then.

izeye:
@shakuzen I updated this as discussed. Along the way, I noticed Netty uses [multiple copyright headers](https://github.com/netty/netty/blob/3099bbcc137dc9bde200e072416ce6d9ac81c766/common/src/main/java/io/netty/util/internal/logging/InternalLogger.java#L1-L39). It seems to be reasonable and feels safe, so I chose the same approach.

shakuzen:
> I updated this as discussed.

Thanks a lot for the quick action.

> Along the way, I noticed Netty uses [multiple copyright headers](https://github.com/netty/netty/blob/3099bbcc137dc9bde200e072416ce6d9ac81c766/common/src/main/java/io/netty/util/internal/logging/InternalLogger.java#L1-L39). It seems to be reasonable and feels safe, so I chose the same approach.

This makes me a bit nervous since we don't have anything else like this in the project so far and I've never noticed this type of thing, but IANAL. I'll try to check with someone on this.

izeye:
@shakuzen FYI, I imagined the reason why it has the second copyright header was the following statement in it:

```
 * The  above  copyright  notice  and  this permission  notice  shall  be
 * included in all copies or substantial portions of the Software.
```

shakuzen:
Sorry for the delay on this. I'm checking internally for an answer on if the extra copyright is an issue or not.

shakuzen:
As a follow-up thought: do we want to add a Checkstyle rule to forbid third-party logger imports/usage outside of the corresponding LoggerFactory? If it's straightforward enough to do, it may be worth it to do this in a subsequent change to this.

izeye:
@shakuzen Sounds good to me. I created https://github.com/micrometer-metrics/micrometer/pull/1610 for it.

Issue #1465, opened 2019-06-20
Graphite registry - escape dots in tags


pawel2401:
Hello,
The graphite-registry is not escaping dots in tags - when dot is present in tag name/value it will be treated as next tag value, e.g.
Tag<tagName.subName, tagValue.subValue> will be treated as <tagName, subName, tagValue, subValue>

One of the possible solutions is to escape the dots, for example by replacing them with underscore will fix the issue.
/Regards,
Pawel

checketts:
Are there any other characters that are problematic? Like spaces?

pawel2401:
From graphite doc:

> Graphite understands messages with this format:
>metric_path value timestamp\n
>metric_path is the metric namespace that you want to populate.
>value is the value that you want to assign to the metric at this time.
>timestamp is the number of seconds since unix epoch time.

It is parsing the metrics basing on space split, so I think that spaces should be escaped as well.

izeye:
I created #1498 to try to resolve this and improve Graphite name sanitization in general.

msmsimondean:
@checketts would it be possible to sanitize `?` question marks too?   Spring Boot puts question marks in metric names in some situations.  Here's an example: 

`httpClientRequests.clientName.some.host.name.method.POST.status.200.uri._path?key__clientKey_.count`

If it helps, here are the characters that HostedGraphite (as an example of a hosted Graphite provider) allow in a Graphite metric path: 

` a-zA-Z0-9:\._$%#=\-\[\] `

msmsimondean:
In the example above, the question mark comes from a URI that's embedded in the metric path via a tag.

checketts:
@msmsimondean Could you add that info to @izeye 's PR: https://github.com/micrometer-metrics/micrometer/pull/1498/files#diff-6bc8bb17eeb45b5274912aaa910ac3d7R37 ?

shevyf:
> If it helps, here are the characters that HostedGraphite (as an example of a hosted Graphite provider) allow in a Graphite metric path:
> 
> `a-zA-Z0-9:\._$%#=\-\[\]`

For information, that's a slightly larger list of characters than normal Graphite allows (HG uses different storage). Vanilla Graphite doesn't allow the colon character either, as Whisper stores metrics as files so it would have issues in any OS that prohibits colons in file names (e.g. Windows). The backslashes shown there are just escape characters, for the same reason.

PR #1498, opened 2019-07-05
Improve Graphite name sanitization

izeye:
This PR improves Graphite name sanitization by:

- sanitizing tag values from tagsAsPrefix.
- sanitizing spaces from meter names.
- sanitizing dots from tag keys and values.

Closes gh-1465

### DIFF
--- a/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteNamingConvention.java
+++ b/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteNamingConvention.java
@@ -33,7 +33,8 @@ public class GraphiteNamingConvention implements NamingConvention {
      * A list that probably is blacklisted: https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/grammar.py#L48-L55.
      * Empirically, we have found others.
      */
-    private static final Pattern blacklistedChars = Pattern.compile("[{}(),=\\[\\]/]");
+    private static final Pattern PATTERN_NAME_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ ]");
+    private static final Pattern PATTERN_TAG_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ .]");
     private final NamingConvention delegate;
 
     public GraphiteNamingConvention() {
@@ -46,17 +47,17 @@ public class GraphiteNamingConvention implements NamingConvention {
 
     @Override
     public String name(String name, Meter.Type type, @Nullable String baseUnit) {
-        return sanitize(this.delegate.name(normalize(name), type, baseUnit));
+        return sanitizeName(this.delegate.name(normalize(name), type, baseUnit));
     }
 
     @Override
     public String tagKey(String key) {
-        return sanitize(this.delegate.tagKey(normalize(key)));
+        return sanitizeTag(this.delegate.tagKey(normalize(key)));
     }
 
     @Override
     public String tagValue(String value) {
-        return sanitize(this.delegate.tagValue(normalize(value)));
+        return sanitizeTag(this.delegate.tagValue(normalize(value)));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteNamingConvention.java
```diff
@@ -33,7 +33,8 @@
      * A list that probably is blacklisted: https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/grammar.py#L48-L55.
      * Empirically, we have found others.
      */
-    private static final Pattern blacklistedChars = Pattern.compile("[{}(),=\\[\\]/]");
+    private static final Pattern PATTERN_NAME_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ ]");
+    private static final Pattern PATTERN_TAG_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ .]");
```
msmsimondean: @izeye  would it be possible to sanitize `?` question marks in names and tags too? Spring Boot puts question marks in metric names in some situations. Here's an example:

`httpClientRequests.clientName.some.host.name.method.POST.status.200.uri._path?key__clientKey_.count`

In this example, the question mark comes from a URI that's embedded in the metric path via a tag.

izeye: @msmsimondean Thanks for the info! Is there any reference on it like their doc or similar?

msmsimondean: @izeye here's the docs for Spring Boot: https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html#production-ready-metrics-http-clients.  The question mark can appear in the URI tag.

izeye: @msmsimondean Sorry, my question doesn't seem to be clear. I meant to ask the reference for the following comment in https://github.com/micrometer-metrics/micrometer/issues/1465#issuecomment-508806231:

> If it helps, here are the characters that HostedGraphite (as an example of a hosted Graphite provider) allow in a Graphite metric path: 
> 
> ` a-zA-Z0-9:\._$%#=\-\[\] `

msmsimondean: Hi @izeye.  The list of characters comes from an email conversation with a HostedGraphite support agent.  The email conversation was only 3 days ago, so info should be up-to-date.  I've done a quick web search and I can't find a page on the hostedgraphite.com site that shows the list of characters.

izeye: @msmsimondean Thanks for the quick feedback! I'm not a regular Graphite user, so I couldn't judge that sanitizing "?" (question mark) is a good idea without a reference. I can update this PR accordingly if anyone can convince me on it or we can discuss on it in a separate issue.

tens91: Hi @izeye , I'm the agent at hosted graphite who @msmsimondean had been emailing. The information they've supplied is correct, those characters are the ones which we accept in graphite names. My colleague also added more information here https://github.com/micrometer-metrics/micrometer/issues/1465#issuecomment-509169319

If documentation is needed for you to update the PR, I can add this to our docs at https://hostedgraphite.com/docs - let me know.

izeye: @tens91 Thanks for the feedback! Based on the comment from @shevyf I pushed 3ffcf9d to be considered. IMHO it'd be good for users to have the info in the doc regardless of the change.

PR #1504, opened 2019-07-09
Polish CloudWatch modules

izeye:
This PR polishes CloudWatch modules.


implementations/micrometer-registry-cloudwatch/src/test/java/io/micrometer/cloudwatch/CloudWatchUtilsTest.java
```diff
@@ -24,9 +24,6 @@
  */
 class CloudWatchUtilsTest {
 
-    private static final double EXPECTED_MIN = 8.515920e-109;
```
shakuzen: I prefer the duplication so that we can more easily catch if we change the value for some reason (tests will start to fail).

izeye: @shakuzen Thanks for the feedback! I usually prefer to use constants from production code in tests as this way, changes on constants from production code won't trigger updates on tests. If testing constant values matters, I'd prefer to have dedicated tests only for verifying them. It might be possible to update them that way but I just reverted them for now instead as I'm okay with as-was.

shakuzen: > If testing constant values themselves matters, I'd prefer to have dedicated tests only for verifying them.

I'd be fine with this as long as it is tested, since if the constant were changed, it would be a change in behavior from the perspective of Micrometer users. But this can be done in a separate pull request anyways.

Issue #1505, opened 2019-07-09
[Regression] IllegalArgumentException in ElasticMeterRegistry.getMajorVersion()


mixcloud-downloader:
The latest Spring Update pulled in micrometer-registry-elastic-1.1.5.jar, which fails now with this exception:

> java.lang.IllegalArgumentException: Unexpected response body: {"status":200,"name":"Sematext-Logsene","cluster_name":"elasticsearch","cluster_uuid":"anything","version":{"number":"5.3.0","build_hash":"3adb13b","build_date":"2017-03-23T03:31:50.652Z","build_snapshot":false,"lucene_version":"6.4.1"},"tagline":"You Know, for Search"}
Jul 09 19:46:59 ubuntu downloader.jar[13292]:         at io.micrometer.elastic.ElasticMeterRegistry.getMajorVersion(ElasticMeterRegistry.java:253)

If I check the offending ElasticMeterRegistry.java:253, I see [this pattern](https://github.com/micrometer-metrics/micrometer/blob/d41fbf531c03739940980eaa0fa94c7577b503b8/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java#L104) `"\"number\" : \"([\\d]+)"`. The spaces don't match the reported json: `"number":"5.3.0"`.

Here would be the PR: https://github.com/micrometer-metrics/micrometer/pull/1507

See also: https://github.com/micrometer-metrics/micrometer/issues/1479

izeye:
@mixcloud-downloader Thanks for the report! Your Elasticsearch seems to be configured to disable pretty-print for the info response somehow unlike the default installation as follows:

```
$ curl http://localhost:9200
{
  "name" : "xQIv0LN",
  "cluster_name" : "elasticsearch",
  "cluster_uuid" : "HqK07rS8S6elsmhXL-JkmA",
  "version" : {
    "number" : "5.3.0",
    "build_hash" : "3adb13b",
    "build_date" : "2017-03-23T03:31:50.652Z",
    "build_snapshot" : false,
    "lucene_version" : "6.4.1"
  },
  "tagline" : "You Know, for Search"
}
$ curl http://localhost:9200?pretty=false
[...]

PR #1507, opened 2019-07-09
Fix IAE in ElasticMeterRegistry.getMajorVersion() (#1505)

mixcloud-downloader:
This commit fixes this error:

> java.lang.IllegalArgumentException: Unexpected response body: …
>   at io.micrometer.elastic.ElasticMeterRegistry.getMajorVersion(ElasticMeterRegistry.java:253)

See also: https://github.com/micrometer-metrics/micrometer/issues/1505

### DIFF
--- a/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticMeterRegistryTest.java
+++ b/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticMeterRegistryTest.java
@@ -257,4 +257,11 @@ class ElasticMeterRegistryTest {
         assertThat(ElasticMeterRegistry.getMajorVersion(responseBody)).isEqualTo(5);
     }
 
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticMeterRegistryTest.java
```diff
@@ -257,6 +257,12 @@ void getVersionWhenVersionIs5() {
         assertThat(ElasticMeterRegistry.getMajorVersion(responseBody)).isEqualTo(5);
     }
 
+    @Issue("#1505")
+    @Test
+    void getVersionWhenVersionIs5_3_0() {
```
izeye: It'd be good to rename to something the test method does like`getVersionWhenVersionIs5AndNotPrettyPrinted`.

Issue #1494, opened 2019-07-04
Gauges do not send units to CloudWatch


danieljamesscott:
It looks like guages do not send units to cloudformation:

https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchMeterRegistry.java#L159

Would it be possible to pull the baseunit from the gauge and call the overloaded `metricDatum` message with the units?

izeye:
@danieljamesscott Thanks for the suggestion! I created #1523 to try to resolve it generally.

PR #1523, opened 2019-07-19
Allow base units from meters to be used in CloudWatch meter registry

izeye:
This PR changes to allow base units from meters to be used in CloudWatch meter registry.

Closes gh-1494


shakuzen:
@danieljamesscott As a CloudWatch user, can you shed any light on my concern?

> I just want to confirm that this won't break any existing users. I looked through [the documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/working_with_metrics.html) and it seems like units being there or not should not break anyone's queries/graphs.

danieljamesscott:
Sorry, I only just saw your question. The queries/graphs would probably be OK because they reference the metric by name/dimension. Alarms based on the metrics should also be OK because they just use a raw number for the threshold and make no mention of the units.

PR #1524, opened 2019-07-19
[Stackdriver] Allow to set custom MonitoredResouce

### LINKED ISSUE #1489: [Stackdriver] Allow to set custom MonitoredResource
Currently, StackdriverMeterRegistry has hardcoded MonitoredResource with type "global" and label "project_id".

There is no possibility to set different resource type other than "global" and put more labels.
Please enhance the code to allow customization of MonitoredResource

YashchenkoN:
Fixes #1489

### DIFF
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverConfig.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverConfig.java
@@ -15,6 +15,8 @@
  */
 package io.micrometer.stackdriver;
 
+import com.google.api.MonitoredResource;
+
 import io.micrometer.core.instrument.config.MissingRequiredConfigurationException;
 import io.micrometer.core.instrument.step.StepRegistryConfig;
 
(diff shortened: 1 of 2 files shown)


YashchenkoN:
#1489

IRus:
I feel like you shouldn't expose:
1. Monitored resource to user, instead define method that allow to customize resource as simple string
2. RESOURCE_TYPE to public

Are you sure, that labels part, can't be solved through common tags?

YashchenkoN:
@IRus
1. Agree, fixed in a subsequent commit
2. Fixed

Yes, I checked and it seems that labels might be solved via common tags

YashchenkoN:
Hi @shakuzen,
Could you please take a look and merge if it LGTY? Thanks

YashchenkoN:
@shakuzen I've changed the base branch. Could you please merge this PR? It would be nice to see this change in the next version of the library.

shakuzen:
> It would be nice to see this change in the next version of the library.

The plan is to include this in the 1.3 release. I'm just focusing on getting out patch releases for 1.1.x and 1.2.x before that. I will circle back to this soon.

Issue #1531, opened 2019-07-25
GraphiteMeterRegistry prints stacktrace at shutdown time.


ekdxhrl0096:
I experienced stacktrace every time I quit the spring application. 
It seems that when trying to close a non-Singleton value, an error occurs. 
How can I fix it?

*application.yml*
```
management:
  metrics:
    export:
      graphite:
        host: 127.0.0.1
        enabled: true
        port: 2003
        step: 20s
        protocol: plaintext
        prefix: "prefix"
```

*Stacktrace*
```
org.springframework.beans.factory.BeanCreationNotAllowedException: Error creating bean with name 'nullChannel': Singleton bean creation not allowed while singletons of this factory are in destruction (Do not request a bean from a BeanFactory in a destroy method implementation!)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:208)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:318)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:199)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBeansOfType(DefaultListableBeanFactory.java:606)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBeansOfType(DefaultListableBeanFactory.java:594)
	at org.springframework.context.support.AbstractApplicationContext.getBeansOfType(AbstractApplicationContext.java:1226)
	at org.springframework.integration.config.IntegrationManagementConfigurer.lambda$registerComponentGauges$1(IntegrationManagementConfigurer.java:407)
	at io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry.lambda$newGauge$0(DropwizardMeterRegistry.java:79)
	at com.codahale.metrics.graphite.GraphiteReporter.reportGauge(GraphiteReporter.java:368)
	at com.codahale.metrics.graphite.GraphiteReporter.report(GraphiteReporter.java:268)
	at com.codahale.metrics.ScheduledReporter.report(ScheduledReporter.java:237)
	at io.micrometer.graphite.GraphiteMeterRegistry.close(GraphiteMeterRegistry.java:92)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #1532, opened 2019-07-25
Handle exception for gauge value function in DropwizardMeterRegistry

izeye:
This PR changes to handle exception for gauge value function in `DropwizardMeterRegistry`. This will give other metrics a chance to be published although there's any exception from gauge functions. There's one caveat that the exception will be lost without any logging as there's no proper logging method for Micrometer core module yet.

See gh-1531

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardMeterRegistry.java
@@ -76,7 +76,11 @@ public abstract class DropwizardMeterRegistry extends MeterRegistry {
         Gauge<Double> gauge = () -> {
             T obj2 = ref.get();
             if (obj2 != null) {
-                return valueFunction.applyAsDouble(obj2);
+                try {
+                    return valueFunction.applyAsDouble(obj2);
+                } catch (Throwable ex) {
+                    return nullGaugeValue();
+                }
             } else {
                 return nullGaugeValue();
             }


checketts:
We need to log something. Even printing a stacktrace would be better than nothing.  How about an interface that just has `logError(String message, Throwable exception)` that we forward to the correct logging implementation based on the provided registry?

izeye:
@checketts Thanks for the feedback! Sounds good to me.

As it's technically a breaking change, I'd like to know which target branch should be used for this.

@shakuzen If you agree with the changes, could you please give me a direction to choose a right target branch?

shakuzen:
> that we forward to the correct logging implementation based on the provided registry?

I don't think we're clear on what logging dependencies we have where, and I'd like to clear that up first. @izeye brought this up in https://github.com/micrometer-metrics/micrometer/pull/1475#issuecomment-506583936. I'll make an issue and link it here.

shakuzen:
The conclusion of #1566 is that we'll introduce some delegating logging like suggested by @checketts in https://github.com/micrometer-metrics/micrometer/pull/1532#issuecomment-515117095. That said, if we need something in the meantime here, Dropwizard does depend on slf4j, so it is probably safe enough for us to use that temporarily in areas that will only be used from an application with a Dropwizard dependency.

shakuzen:
> We need to log something. Even printing a stacktrace would be better than nothing.

I'm a bit torn here. I agree we should let users know they've registered a gauge with a function that is throwing an exception. But if we do this every time metrics are reported, it feels like we're being too verbose about a single metric.

checketts:
In similar situations, I've seen code print out the warning on the first time, and note that further violations will be printed at debug level. Then we could just use a boolean to track if it is the first occurrence or not.

PR #1533, opened 2019-07-25
Update javadocs for @Timed annotation

tauparticle:
extraTags{} should be supplied with an even key-value pair list, which isn't clear in the annotation comments.


shakuzen:
Thank you for this. I have rebased it onto `1.1.x` so it can be applied to that maintenance branch and merged forward.

PR #1544, opened 2019-08-15
Move to new Humio structered API.

mwl:
New structured API doesn't require a repo reference so was removed as well.

---
References

* [Ingest tokens](https://docs.humio.com/operations-guide/configuration/sending-data-to-humio/ingest-tokens/)
* [Ingest API](https://docs.humio.com/developer-resources/api/ingest-api/)

### DIFF
--- a/implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
+++ b/implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
@@ -46,14 +46,6 @@ public interface HumioConfig extends StepRegistryConfig {
     }
 
     /**
-     * @return The repository name to write metrics to.
-     */
-    default String repository() {
-        String v = get(prefix() + ".repository");
-        return v == null ? "sandbox" : v;
-    }
-
-    /**
      * Humio uses a concept called "tags" to decide which datasource to store metrics in. This concept
      * is distinct from Micrometer's notion of tags, which divides a metric along dimensional boundaries.
      * All metrics from this registry will be stored under a datasource defined by these tags.
--- a/implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
+++ b/implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
@@ -46,6 +46,15 @@ public interface HumioConfig extends StepRegistryConfig {
     }
 
     /**
+     * @return The repository name to write metrics to.
+     * @deprecated No longer used as repository is resolved from the api token
+     */
+    @Deprecated
+    default String repository() {
+        return "";
+    }
+
+    /**
      * Humio uses a concept called "tags" to decide which datasource to store metrics in. This concept
      * is distinct from Micrometer's notion of tags, which divides a metric along dimensional boundaries.
      * All metrics from this registry will be stored under a datasource defined by these tags.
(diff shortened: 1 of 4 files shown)


mwl:
CI checks currently failing due to apparently unrelated Atlas issue. Please advice

shakuzen:
@mwl I've opened #1545 regarding the build failure.

shakuzen:
@mwl if you rebase this branch on `master`, it should have the previous build issue resolved.

mwl:
Tested against upcoming Humio 1.6.0 but should also be compatible with everything from [1.2](https://docs.humio.com/release-notes/#2018-12-11) onwards

shakuzen:
> Documentation might need updating too. What's the process for that?

Documentation is at https://github.com/micrometer-metrics/micrometer-docs/
Feel free to send a pull request over there with doc updates. Thanks.

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
```diff
@@ -45,14 +45,6 @@ default String uri() {
         return v == null ? "https://cloud.humio.com" : v;
     }
 
-    /**
-     * @return The repository name to write metrics to.
-     */
-    default String repository() {
```
shakuzen: Rather than suddenly deleting this, which may break compatibility for users, let's mark it as deprecated and note that it is no longer used and will be removed.

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioMeterRegistry.java
```diff
@@ -91,7 +91,7 @@ public void start(ThreadFactory threadFactory) {
     protected void publish() {
         for (List<Meter> meters : MeterPartition.partition(this, config.batchSize())) {
             try {
-                HttpSender.Request.Builder post = httpClient.post(config.uri() + "/api/v1/dataspaces/" + config.repository() + "/ingest");
+                HttpSender.Request.Builder post = httpClient.post(config.uri() + "/api/v1/ingest/humio-structured");
```
shakuzen: How is the repository determined without it being in the URL? Is it determined by the API token configured?

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
mwl: We _can_ put it back in as a deprecated method, but it will be completely ignore in the code. Your call.

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioMeterRegistry.java
mwl: There have been some major refactoring of Humio so that ingest keys are now created outside of the repository but have a default repository in the ingest routing layer.

mwl:
@shakuzen I for one (as an Humio employee) do not want to break any dashboards etc. either, so I'm very carefull when I make these changes and by the end of the day these changes are just reflecting what is actually going on in the server where dataspace/repository is being ignored and resolved from the ingest token.
The API used in current version of Micrometer has been [deprecated some time ago](https://docs.humio.com/release-notes/#2019-02-07) so to make sure Micrometer works as usual going forward we should move to the latest api as soon as possible.
Currently there a re no plans to delete the API in question.

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
shakuzen: Yes, that would be preferable. If you make this change but revert the change in the other classes, they result in compiler errors because they claim to `@Override` a method not declared in `HumioConfig` anymore. This is what users would experience, but it is not a great upgrade experience. We can deprecate it in 1.3 (and stop using it for anything) and remove it in 1.4 so users will have been able to notice the deprecation and remove its usage for a smooth upgrade to 1.4.

shakuzen:
Thank you for the explanation and care taken here. I've assigned this to the 1.3 milestone so we can get this change out in the next feature release.

shakuzen:
In our Spring support, we'll want to make sure that the `repository` property gets marked as deprecated. I've added the `spring-boot change` label to remind me of this.

PR #1550, opened 2019-08-19
adjust to recent spectator changes

### LINKED ISSUE #1545: Build broken since Atlas version 0.93.0
Noticed by @mwl https://github.com/micrometer-metrics/micrometer/pull/1544#issuecomment-521543413

Since the `master` branch doesn't use dependency locks and the dependency is declared with `latest.release` as the version, we get the latest version of the `com.netflix.spectator:spectator-reg-atlas` dependency on each build. Since version `0.93.0` the tests are failing. They continue to fail with the latest version (at the time of opening this issue) `0.94.0`.

We can pin to an older version to fix the build temporarily, but we need to figure out the root cause of things starting to fail and perhaps report it to the maintainers of the dependency.

The last version the tests pass with is `0.92.2`

brharrington:
Two changes to fix tests that started breaking after recent
changes to AtlasRegistry:

1. Update AtlasConfig objects in tests to make the hi-res
   step used for on-demand streams match the publish step.
   When using tests with a synthetic clock this avoids
   missing some updates in the consolidated value published.
2. `AtlasMeterRegistryTest` was using reflection to call a
   private method that was renamed. This has been changed to
   just call stop on the AtlasRegistry. AtlasRegistry as of
   0.95.0 will handle the flush of the final interval.

Fixes #1545.

### DIFF
--- a/implementations/micrometer-registry-atlas/build.gradle
+++ b/implementations/micrometer-registry-atlas/build.gradle
@@ -1,7 +1,6 @@
 dependencies {
     compile project(':micrometer-core')
-    // FIXME see https://github.com/micrometer-metrics/micrometer/issues/1545
-    compile 'com.netflix.spectator:spectator-reg-atlas:0.92.2'
+    compile 'com.netflix.spectator:spectator-reg-atlas:latest.release'
 
     testCompile project(':micrometer-test')
 }
--- a/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/AtlasMeterRegistry.java
+++ b/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/AtlasMeterRegistry.java
@@ -33,8 +33,6 @@ import io.micrometer.core.instrument.step.StepFunctionTimer;
 import io.micrometer.core.instrument.util.DoubleFormat;
 import io.micrometer.core.lang.Nullable;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.function.ToDoubleFunction;
@@ -88,15 +86,6 @@ public class AtlasMeterRegistry extends MeterRegistry {
 
     @Override
     public void close() {
-        try {
-            @SuppressWarnings("JavaReflectionMemberAccess")
-            Method collectData = registry.getClass().getDeclaredMethod("collectData");
-            collectData.setAccessible(true);
-            collectData.invoke(registry);
-        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
-            // oh well, we tried
-        }
-
         stop();
         super.close();
     }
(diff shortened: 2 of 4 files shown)


shakuzen:
Thank you for the quick turnaround on changes both here and in spectator-reg-atlas @brharrington.

I'm not very familiar with the Spectator registry for Atlas implementation details, but I noticed that `lwcEnabled()` defaults to false, which would give me the impression that `lwcStep` shouldn't affect things unless LWC is enabled.
Also, I'm wondering if users will run into the same issue our tests had when upgrading versions of `spectator-reg-atlas` if they don't make the same changes. What makes the problem specific to using a mocked clock?

Was it considered to make `lwcStep()` default to the value of `step()` like changed here, for backwards compatibility?

brharrington:
The `lwcEnabled` defaults to false because it is new and users can choose whether or not to opt in. It is setup so that it can be dynamically changed at runtime, other details have to be in place for it work without requiring a restart.

It shouldn't impact users running the registry normally. In the test cases you are not actually starting it and having the background threads process the data as they normally would.

shakuzen:
> It shouldn't impact users running the registry normally. In the test cases you are not actually starting it and having the background threads process the data as they normally would.

Perhaps the reporting of metrics is still fine (I haven't checked), but programmatically accessing the metrics doesn't seem the same as before still. I tried running the tests without explicitly setting the `lwcStep` (as I expect most users' configuration to be when upgrading) and a `step` of 10 seconds with a 10 second sleep so background threads can do their work before checking the value of a Counter's `count`. The count is still not what it was before because the step used by the Counter is the `lwcStep`, I think.

Here's a simplified version of one of the tests that fails when upgrading without configuring the `lwcStep` but replacing the mocked clock with a real one and sleeping for the `step` time. This demonstrates the difference in behavior a user might be affected by if programmatically accessing meters in their application.

```java
static final Duration STEP = Duration.ofSeconds(10);
AtlasMeterRegistry registry = new AtlasMeterRegistry(new AtlasConfig() {
    @Override
    public boolean enabled() {
        return false;
    }

    @Override
    public String get(String k) {
        return null;
    }

    @Override
    public Duration step() {
        return STEP;
    }
}, Clock.SYSTEM);

@Test
void findMetersByValue() throws InterruptedException {
    Counter c = registry.counter("counter");
    c.increment();

    Thread.sleep(STEP.toMillis());

    assertThat(registry.get("counter").counter().count()).isEqualTo(1.0);
    registry.close();
}
```

brharrington:
The amount of time for accessing a value directly from a meter is an implementation detail for the registry. From the [javadocs](https://static.javadoc.io/com.netflix.spectator/spectator-api/0.95.0/com/netflix/spectator/api/Counter.html#count--):

> The cumulative count since this counter was last reset. How often a counter is reset depends on the underlying registry implementation.

We have always tried to keep the guarantees on that to a minimum so we have flexibility to change and optimize the implementation of the registry over time. In practice running with these changes we haven't seen issues with this behavior. Testing is the main place I can see where it comes up and the user is highly concerned about when a particular value will be shown as opposed to a representative snapshot of recent behavior. However, Micrometer usage is a bit of an anomaly for that usage as it is bound specifically to the AtlasRegistry and testing some implementation details of that registry. For most uses we see, [testing](http://netflix.github.io/spectator/en/latest/intro/testing/) is done by using the simple DefaultRegistry implementation which doesn't have any timing or reset behavior to worry about.

shakuzen:
Thank you for educating me. I think I have the whole picture now. We were relying on behavior that was never guaranteed. We may want to consider whether we want to restore that behavior by wrapping the Spectator implementation or making our own, but that's something we can consider separately.
Thank you again for noticing the issue about the build and quickly making the changes necessary.

PR #1556, opened 2019-08-24
Remove SendOptions usage

izeye:
This PR removes `SendOptions` usage as it has been removed in https://github.com/reactor/reactor-netty/issues/766.


shakuzen:
Do you know when the changes to the send behavior were made? Depending on that, maybe we want to backport this to a maintenance version as well.

izeye:
@shakuzen Based on https://github.com/reactor/reactor-netty/pull/791, it has been deprecated since 0.8.10.RELEASE but I don't know when the mentioned changes have been applied.

shakuzen:
Alright. Let's just merge this for 1.3 for now, then. Micrometer users will have issues trying to use reactor-netty 0.9 with older versions of Micrometer, but we don't want to make the change and affect behavior for users using an early version of reactor-netty.

shakuzen:
I was reminded afterwards that we are shading reactor-netty, so the version used by Micrometer users won't matter here. Given that, we can backport this safely.

PR #1565, opened 2019-08-29
Upgrade to Gradle Wrapper 5.6.1

izeye:
This PR upgrades to Gradle Wrapper 5.6.1.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -109,7 +109,7 @@ subprojects {
 }
 
 wrapper {
-    gradleVersion = '5.6'
+    gradleVersion = '5.6.1'
 }
 
 defaultTasks 'build'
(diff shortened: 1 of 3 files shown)


shakuzen:
The build failed with the following error. I have seen this before but haven't looked into the cause of it. I don't think it is related to the changes here. Can you rerun the build just to double check?
We should probably figure out why this happens so we don't keep losing time to this failure.
```
Execution failed for task ':micrometer-spring-legacy:test'.
> Process 'Gradle Test Executor 23' finished with non-zero exit value 137
  This problem might be caused by incorrect test process configuration.
  Please refer to the test execution section in the User Manual at https://docs.gradle.org/5.6.1/userguide/java_testing.html#sec:test_execution
```

izeye:
@shakuzen I force-pushed and the build succeeded this time. Just with a quick search, it might relate to OOM killer in CircleCI based on [this SO question's accepted answer](https://stackoverflow.com/questions/38967991/why-are-my-gradle-builds-dying-with-exit-code-137).

PR #1575, opened 2019-09-04
Metrics collector for Undertow

### LINKED ISSUE #1227: Feature: Undertow metrics
I would very much like to measure Undertow using Micrometer, but I notice that there is no `UndertowMetrics` class like there is for Tomcat (`TomcatMetrics`) or Jetty (`JettyStatisticsMetrics` et.al).

Has anyone already built this somewhere else, or are there any ongoing efforts to build this?

I noticed that there are multiple issues mentioning Undertow metrics, but none that come to any conclusion.

Thank you for considering my request.

dharmeshjogadia:
fixed #1227


micrometer-core/build.gradle
```diff
@@ -32,6 +32,9 @@ dependencies {
     // 9.4+ are incompatible with the latest version of wiremock as of 1/4/2019
     compile 'org.eclipse.jetty:jetty-server:9.2.+', optional
     compile 'org.apache.tomcat.embed:tomcat-embed-core:8.+', optional
+    // https://mvnrepository.com/artifact/io.undertow/undertow-servlet
+    compile 'io.undertow:undertow-servlet:2.0.+', optional
```
IRus: `undertow-core` should be enough (I can't see if any servlets used down here). And please delete ad comment.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/undertow/UndertowMetrics.java
```diff
@@ -0,0 +1,94 @@
...
+import static io.undertow.server.handlers.MetricsHandler.MetricResult;
+
+/**
+ * {@link MetricsHandler} to collect metrics for {@link io.undertow.Undertow}
+ *
+ * @author Dharmesh Jogadia
+ * */
+@NonNullApi
+@NonNullFields
+public class UndertowMetrics implements MetricsCollector, MeterBinder {
```
IRus: Why not implement it as more general `MetricsHandler`? So it can be used outside of servlet api (which more generic)

dharmeshjogadia: Ok, I'll check and get back, if there's another way to scrap undertom metrics. 
Any more suggestions or references are welcome...!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/undertow/UndertowMetrics.java
```diff
@@ -0,0 +1,94 @@
...
+                .tags(tags)
+                .tag("servlet_name", servletName)
+                .description(desc)
+                .register(registry);
+    }
+
+    private void bindTimeGauge(MeterRegistry registry, String name, String desc, String servletName, MetricResult metricResult, ToDoubleFunction<MetricResult> consumer) {
+        TimeGauge.builder(name, metricResult, TimeUnit.MILLISECONDS, consumer)
+                .tags(tags)
+                .tag("servlet_name", servletName)
```
shakuzen: Tag names should also follow meter naming convention (lowercase dot notation) so they can be properly converted for each metric backend implementation. In this case, `servlet.name`.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/undertow/UndertowMetrics.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.ToDoubleFunction;
+import java.util.function.ToLongFunction;
+
+import static io.undertow.server.handlers.MetricsHandler.MetricResult;
+
+/**
+ * {@link MetricsHandler} to collect metrics for {@link io.undertow.Undertow}
```
shakuzen: Some information about how to use this class in the JavaDoc would be good. For instance, from the test code, it seems this needs to be set as the metricsCollector on the DeploymentInfo, and it probably needs to be bound after `Undertow` is started?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/undertow/UndertowMetrics.java
dharmeshjogadia: I did some research on Undertow generic metrics and here a mail [thread](https://lists.jboss.org/pipermail/undertow-dev/2017-October/002123.html) I found from undertow-dev.

shakuzen: @IRus is the idea that we would use the Undertow `MetricsHandler` somehow, or would we make our own `HttpHandler` implementation that does something similar but with Micrometer?

IRus: [MetricsHandler](https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/server/handlers/MetricsHandler.java) doesn't preserve request paths and methods, so yes, I think will be better to have in-house one

shakuzen: @dharmeshjogadia would that work for the metrics you want to make available for Undertow? If so, would you be willing to update this pull request with that implementation?
I notice you're tagging the servlet name in the current implementation. I wonder if we would be able to get that kind of information from an `HttpHandler` implementation. If not, maybe we will need the current implementation in addition to an `HttpHandler` one - but users would have to be careful to not configure both if they're using servlet.

shakuzen: Still looking for feedback on my previous comment if folks have time.

shakuzen:
Closing this pull request for now. Ideally we would want something that works for servlet and non-servlet environments, if we can get the info we need. That said, I've left a comment on https://github.com/micrometer-metrics/micrometer/issues/1227 asking for Micrometer and Undertow users looking for metrics to explore having the instrumentation in the Undertow project instead of here.

PR #1578, opened 2019-09-06
Human-readable SimpleMeterRegistry toString

### LINKED ISSUE #1527: Human-readable output for SimpleMeterRegistry
When doing first steps with micrometer it's not that straight forward to understand what is going on. It would simplify the learning curve massively if you can just output metrics you've collected in a human-readable form like Dropwizard is doing:

```
-- Gauges ----------------------------------------------------------------------
testLatencyHistogram.le.100
             value = 30.0
testLatencyHistogram.le.12
             value = 4.0
testLatencyHistogram.le.200
             value = 50.0
testLatencyHistogram.le.24
             value = 9.0
testLatencyHistogram.le.50
             value = 15.0
testLatencyPercentile.phi.0.9
             value = 150.994432
testLatencyPercentile.phi.0.97
             value = 150.994432

-- Timers ----------------------------------------------------------------------
testLatency
             count = 50
         mean rate = 11.59 calls/second
     1-minute rate = 0.00 calls/second
     5-minute rate = 0.00 calls/second
    15-minute rate = 0.00 calls/second
               min = 0.02 milliseconds
               max = 156.63 milliseconds
              mean = 83.62 milliseconds
            stddev = 47.23 milliseconds
            median = 90.30 milliseconds
              75% <= 124.76 milliseconds
              95% <= 148.10 milliseconds
              98% <= 148.20 milliseconds
              99% <= 156.63 milliseconds
            99.9% <= 156.63 milliseconds
```

I know you can [hook up](https://micrometer.io/docs/guide/consoleReporter) dropwizard registry, which is convenient for those, who are familiar with dropwizard metrics, but why bother if you can just override `SimpleMeterRegistry`'s `toString` method (or add another method) and print all the metrics in a human-readable form. So far I came up with something like:

```
Gauge test.latency.histogram [tag(le=12)] value=4.0
Gauge test.latency.histogram [tag(le=24)] value=9.0
Gauge test.latency.histogram [tag(le=50)] value=15.0
Gauge test.latency.histogram [tag(le=100)] value=30.0
Gauge test.latency.histogram [tag(le=200)] value=50.0
Timer test.latency []
	count=50, total=4189.092825, mean=83.7818565, max=156.630944
	percentileValues:
		90.0%: 150.994432
		97.0%: 150.994432
	histogramCounts:
		12.0: 4.0
		24.0: 9.0
		50.0: 15.0
		100.0: 30.0
		200.0: 50.0
Gauge test.latency.percentile [tag(phi=0.9)] value=150.994432
Gauge test.latency.percentile [tag(phi=0.97)] value=150.994432
```

But it can be further improved, by properly rounding and indenting numbers, adding units and adding mode to deduplicate technical gauges.

There is also `LoggingMeterRegistry`, but as far as understand it is designed to save metrics to logs, so one line – one meter, also you need to lunch it and it will output metrics regularly, which is inconvenient for my use cases.

Two main use cases I can see:

1. Learning micrometer, doing first steps, playing with metrics and output when you need it.
1. Instrument some local test project, where you want to benchmark different approaches and output human-readable stats.

The implementation I have in mind is to override `toString` method for `SimpleMeterRegistry`.

Let me know, what do you think, maybe I'm missing something and it already exists. If you like an idea – I can submit a merge request pretty soon.

artjock:
Example:

Timer response.time [tag(action=create)]
        count = 300
        total = 44.850000
         mean = 0.149500
          max = 0.299000
    percentileValues:
        75.0 <= 0.226459648
        90.0 <= 0.285179904
        95.0 <= 0.285179904
        99.0 <= 0.30195712
        99.9 <= 0.30195712
    histogramCounts:
        0.02 <= 21.0
        0.05 <= 51.0
         0.1 <= 101.0

closes #1527

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/simple/SimpleMeterRegistryTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/simple/SimpleMeterRegistryTest.java
@@ -16,18 +16,26 @@
 package io.micrometer.core.instrument.simple;
 
 import io.micrometer.core.Issue;
+import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.FunctionCounter;
 import io.micrometer.core.instrument.FunctionTimer;
 import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Measurement;
 import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Meter.Type;
 import io.micrometer.core.instrument.MockClock;
+import io.micrometer.core.instrument.Statistic;
 import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.TimeGauge;
 import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.cumulative.CumulativeFunctionCounter;
 import io.micrometer.core.instrument.cumulative.CumulativeFunctionTimer;
 import io.micrometer.core.instrument.step.StepFunctionCounter;
 import io.micrometer.core.instrument.step.StepFunctionTimer;
+import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicLong;
 import org.junit.jupiter.api.Test;
 
 import java.time.Duration;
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/simple/SimpleMeterRegistryTest.java
```diff
@@ -101,6 +109,133 @@ public void newFunctionCounterWhenCountingModeIsStepShouldReturnStepFunctionCoun
...
+                + "        99.0 <= 303.9375\n"
+                + "       99.99 <= 303.9375\n"
+                + "    histogramCounts:\n"
+                + "       250.0 <= 256.0\n"
+                + "       500.0 <= 300.0\n"
+                + "      1000.0 <= 300.0\n"
+                + "      2000.0 <= 300.0\n"
+                + "      4000.0 <= 300.0\n"
+                + "      6000.0 <= 300.0\n"
+                + "Gauge response.size.histogram {le=250} value=256.000000\n"
```
shakuzen: It seems like it'd be nice if we could deduplicate these gauges since this info is already present in the above histogramCounts.

jkschneider:
@artjock The `LoggingMeterRegistry` already attempts this sort of human readable formatting. An example is included here. Do you think this works or could make necessary improvements to it instead?

image

Issue #1543, opened 2019-08-13
Micrometer with ehcache and datadog throws InvocationTargetException


cscheffel:
We have integrated micormeter in a spring boot 2.1.6-App which use ehcache 2.10.6 as caching provider. The runtime enviroment is open jdk 11. After adding micrometer 1.1.5 (and as a try 1.2.0) and a custom specifc counter everything works well at the beginning. 

Micrometer transfers the custom counter to datadog.

After a view seconds (i.e. timerbased), micrometer trys to transmit metering data for ehcache which results into a InvocationTargetException. At this moment, all transmitting is stopped. Even the custom counter is not transmitted any more.

Stacktrace:
```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
        at org.terracotta.statistics.PassThroughStatistic.value(PassThroughStatistic.java:65)
        at net.sf.ehcache.statistics.extended.AbstractStatistic.value(AbstractStatistic.java:57)
        at net.sf.ehcache.statistics.extended.SemiExpiringStatistic.value(SemiExpiringStatistic.java:33)
        at net.sf.ehcache.statistics.StatisticsGateway.getLocalHeapSizeInBytes(StatisticsGateway.java:327)
        at io.micrometer.core.instrument.internal.DefaultGauge.value(DefaultGauge.java:40)
        at io.micrometer.core.instrument.Measurement.getValue(Measurement.java:39)
        at io.micrometer.datadog.DatadogMeterRegistry.lambda$writeMeter$12(DatadogMeterRegistry.java:210)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.Collections$2.tryAdvance(Collections.java:4745)
        at java.base/java.util.Collections$2.forEachRemaining(Collections.java:4753)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

PR #1582, opened 2019-09-10
Make value function defensive for exceptions in DefaultGauge

izeye:
This PR changes value function defensive for exceptions in `DefaultGauge`. This way, we can keep publishing other meters even if value functions from some `DefaultGauge`s have failed.

See gh-1543


izeye:
I'd like to add some log there, but there's no proper way to do so yet. Once #1475 has been merged, some log could be added.

PR #1583, opened 2019-09-10
Eat RuntimeExceptions in DataDog-EHCache transmissions to avoid errors

cscheffel:
This fixes issue Micrometer with ehcache and datadog throws InvocationTargetException #1543


shakuzen:
I believe we have solved this in a more general way for gauges which will help all registries, not just Datadog. Please let us know if this is still an issue in the latest supported versions. I don't think this change is needed due to https://github.com/micrometer-metrics/micrometer/pull/1582.

PR #1585, opened 2019-09-12
Fix NewRelicMeterRegistry.writeMeter() publishing duplicate keys/metrics

neiljpowell:
New Relic Insights API docs (REST API and Java Agent) state that a
standard map of key value pairs (number, string) are expected and
Object/Array value types are not supported.

See Insights REST API (JSON Format Guidelines):
https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/send-custom-events-event-api

@shakuzen 
@breedx-nr

### DIFF
--- a/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
+++ b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
@@ -15,7 +15,34 @@
  */
 package io.micrometer.newrelic;
 
-import io.micrometer.core.instrument.*;
+import static io.micrometer.core.instrument.util.StringEscapeUtils.escapeJson;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.FunctionCounter;
+import io.micrometer.core.instrument.FunctionTimer;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Measurement;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.TimeGauge;
+import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.config.MissingRequiredConfigurationException;
 import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.step.StepMeterRegistry;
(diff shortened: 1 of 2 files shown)


neiljpowell:
@checketts If you could review this simple fix also, it'd be much appreciated.

breedx-nr:
I also took a look and this looks good, @neiljpowell .  Thanks for the contribution.

neiljpowell:
Did you look at my second PR?  It was more involved but straightforward to optionally write all metrics to same eventType.

- Neil Powell

On Sep 17, 2019, at 11:05 AM, Jason Plumb <notifications@github.com> wrote:

I also took a look and this looks good, @neiljpowell . Thanks for the contribution.

implementations/micrometer-registry-new-relic/src/test/java/io/micrometer/newrelic/NewRelicMeterRegistryTest.java
```diff
@@ -151,7 +151,7 @@ void writeMeterWhenCustomMeterHasMixedFiniteAndNonFiniteValuesShouldSkipOnlyNonF
         Measurement measurement5 = new Measurement(() -> 2d, Statistic.VALUE);
         List<Measurement> measurements = Arrays.asList(measurement1, measurement2, measurement3, measurement4, measurement5);
         Meter meter = Meter.builder("my.meter", Meter.Type.GAUGE, measurements).register(this.registry);
-        assertThat(registry.writeMeter(meter)).contains("{\"eventType\":\"myMeter\",\"value\":1,\"value\":2}");
```
shakuzen: Could we add a test case specifically for what this pull request is fixing just so it is more explicit?

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
```diff
@@ -194,18 +208,19 @@ protected void publish() {
...
-        List<Attribute> attributes = new ArrayList<>();
+        Map<String, Attribute> attributes = new HashMap<>();
         for (Measurement measurement : meter.measure()) {
             double value = measurement.getValue();
             if (!Double.isFinite(value)) {
                 continue;
             }
-            attributes.add(new Attribute(measurement.getStatistic().getTagValueRepresentation(), value));
+            String name = measurement.getStatistic().getTagValueRepresentation();
+            attributes.put(name, new Attribute(name, value));
```
shakuzen: This results in a last duplicate attribute name wins behavior. I'm wondering what was the effective behavior before when we were sending all to New Relic? Was it resulting in an error or only saving one of the attributes with the duplicate name, or something else?

shakuzen:
If this is a bug, and it sounds like it is (but I'd like to understand the current behavior and its effect), then we'll want to merge this into the `1.1.x` branch and then I'll merge it forward to other branches. That way the fix gets applied to all maintenance branches.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
neiljpowell: I'll investigate. If you haven't had issues, and error isn't returned by the client and/or non-deterministic behavior occurs. I came across this while working on my Java-Agent based registry PR.  The Agent Insights API takes a Map and forwards to an API. I assume the same or similar. When reading the docs, it made sense to align to a similar client side Map behavior for consistency.  @breedx-nr would you know?

implementations/micrometer-registry-new-relic/src/test/java/io/micrometer/newrelic/NewRelicMeterRegistryTest.java
neiljpowell: Certainly.

neiljpowell:
Hmmm. PrometheusMetricsExportAutoConfigurationTest failures?

shakuzen:
> Hmmm. PrometheusMetricsExportAutoConfigurationTest failures?

I just merged the fix for that in #1601 and I've rebased this pull request, so it should be good now.

Issue #1462, opened 2019-06-19
Run tests on OpenJ9


shakuzen:
In addition to HotSpot, we try to support OpenJ9 where applicable and feasible. We, however, don't run our test suite on an OpenJ9 JVM to proactively catch issues and ensure compatibility.
We should consider adding an OpenJ9 build.

marcingrzejszczak:
Is this still something we want to do @shakuzen ?

shakuzen:
We claim to support JVM metrics on OpenJ9, so yes, it would generally be good that we actually run tests on OpenJ9. I think it will require creating a custom container image to run tests on or otherwise use different images than the convenience images offered by CircleCI since they rejected the request to add OpenJ9 variants: https://github.com/CircleCI-Public/cimg-openjdk/issues/78

PR #1589, opened 2019-09-16
Add Compilation JVM metric binder

gquintana:
Add a metric binder to get JVM compilation time

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
@@ -15,9 +15,10 @@
  */
 package io.micrometer.core.instrument.binder.jvm;
 
-import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.FunctionCounter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.binder.MeterBinder;
 import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
@@ -44,8 +45,8 @@ public class CompilationMetrics implements MeterBinder {
     public void bindTo(MeterRegistry registry) {
         CompilationMXBean compilationBean = ManagementFactory.getCompilationMXBean();
         if (compilationBean.isCompilationTimeMonitoringSupported()) {
-            Gauge.builder("jvm.compilation.time.total", compilationBean, CompilationMXBean::getTotalCompilationTime)
-                    .tags(tags)
+            FunctionCounter.builder("jvm.compilation.time.total", compilationBean, CompilationMXBean::getTotalCompilationTime)
+                    .tags(Tags.concat(tags, "compiler", compilationBean.getName()))
                     .description("The approximate accumulated elapsed time spent in compilation")
                     .baseUnit("ms")
                     .register(registry);
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
@@ -1,0 +1,54 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.jvm;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.MeterBinder;
+import io.micrometer.core.lang.NonNullApi;
+import io.micrometer.core.lang.NonNullFields;
+
+import java.lang.management.CompilationMXBean;
+import java.lang.management.ManagementFactory;
+
+import static java.util.Collections.emptyList;
+
+@NonNullApi
+@NonNullFields
+public class CompilationMetrics implements MeterBinder {
+    private final Iterable<Tag> tags;
+
+    public CompilationMetrics() {
+        this(emptyList());
+    }
+
+    public CompilationMetrics(Iterable<Tag> tags) {
+        this.tags = tags;
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public CompilationMetrics(Iterable<Tag> tags) {
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        CompilationMXBean compilationBean = ManagementFactory.getCompilationMXBean();
+        if (compilationBean.isCompilationTimeMonitoringSupported()) {
+            Gauge.builder("jvm.compilation.time.total", compilationBean, CompilationMXBean::getTotalCompilationTime)
```
checketts: Since this is an accumulated total, you'll want to use a `FunctionCounter`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
```diff
@@ -0,0 +1,54 @@
...
+                    .tags(tags)
```
checketts: It might be useful to expose the JIT compiler name as a tag via `getName`

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/CompilationMetricsTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+class CompilationMetricsTest {
+    @Test
+    void compilationTimeMetric() {
+        MeterRegistry registry = new SimpleMeterRegistry();
+        new CompilationMetrics().bindTo(registry);
+
+        assertThat(registry.get("jvm.compilation.time.total").gauge().value()).isGreaterThan(0);
```
checketts: Will this test fail on builds that run on alternate JDKs?  I'm not too familiar with the build setup in that regard @shakuzen

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+import io.micrometer.core.lang.NonNullFields;
+
+import java.lang.management.CompilationMXBean;
+import java.lang.management.ManagementFactory;
+
+import static java.util.Collections.emptyList;
+
+@NonNullApi
+@NonNullFields
+public class CompilationMetrics implements MeterBinder {
```
checketts: @shakuzen @jkschneider Should this class be named `JvmCompilationMetrics` since these are metrics around the JVM?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public CompilationMetrics(Iterable<Tag> tags) {
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        CompilationMXBean compilationBean = ManagementFactory.getCompilationMXBean();
+        if (compilationBean.isCompilationTimeMonitoringSupported()) {
+            FunctionCounter.builder("jvm.compilation.time.total", compilationBean, CompilationMXBean::getTotalCompilationTime)
```
mweirauch: I am unsure about the naming and the baseUnit usage below.
The PrometheusNamingConvention adds the unit and then the "_total" suffix again due to how it works for `Counter`s:
`jvm_compilation_time_total_ms_total`

Unfortunately we can't make this a `Timer` due to not being triggered by the JVM so we don't really have any counts on compilation frequency.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/CompilationMetricsTest.java
mweirauch: I was of the false memory that we already had some kind of matrix build which don't just test the different Hotspot JVMs. Apparently that isn't the case. Was quickly testing a `./gradlew test` with 8u222-openJ9 and there's quite some test failures before I could reach this one. (Hope I didn't screw up.)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
gquintana: I don't mind renaming the class but there is already a  `ClassLoaderMetrics` which should be renamed `JvmClassLoaderMetrics` in that case.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
gquintana: What about `jvm.compilation.ms.total` ?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/CompilationMetricsTest.java
gquintana: Is it a problem in the end?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
mweirauch: That will become `jvm_compilation_ms_total_ms_total` because the `PrometheusNamingConvention` will first add the base unit and then the `_total`.

How about just `jvm.compilation.time` and any NamingConvention specific to a registry implementation will make sure that it generates whatever is best practice for that backend. For Prometheus this would then become `jvm_compilation_time_ms_total`.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
shakuzen: I agree `JvmCompilationMetrics` is more clear here. `ClassLoaderMetrics` is already released public API so we can't change that without breaking all users of it.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/CompilationMetricsTest.java
shakuzen: > I was of the false memory that we already had some kind of matrix build which don't just test the different Hotspot JVMs. Apparently that isn't the case.

I opened an issue for it but never got around to doing it - #1462. If a test is specific to a JVM, you can use JUnit Assumptions to skip it when those assumptions aren't meant. We do this for example in the `ProcessorMetricsTest`: https://github.com/micrometer-metrics/micrometer/blob/49162a881c4c80976980d6ea0ef45d2d59c4ca14/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java#L53

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/CompilationMetrics.java
shakuzen: `jvm.compilation.time` sounds good to me.

shakuzen:
Polish in 114d62be80c51768031a0672b021bb30a917c615

PR #1592, opened 2019-09-17
Polish ReactorNettySender

izeye:
This PR polishes `ReactorNettySender` by avoiding unnecessary `HttpClient` creation and String conversion.

See https://github.com/micrometer-metrics/micrometer/commit/5998b8e50eec591ce0c494a9a9d273f59cabaaa0#r34920004


BartoszKaszewczuk:
This change brings back original fix from https://github.com/micrometer-metrics/micrometer/pull/1552 and takes it further by omitting redundant allocation of `String` objects during message creation.

Well done 👍🏻

PR #1595, opened 2019-09-20
Add Tomcat ThreadPool connectionCount metric.

### LINKED ISSUE #1598: Missing connectionCount metric in TomcatMetrics
We have recently found that Micrometer does not expose 'connectionCount' metrics for Tomcat, as a part of the PR: https://github.com/micrometer-metrics/micrometer/pull/1595 we also spotted that there is no effective test for metrics derived from 'mBeanServer' (in contrast to Manager object) since invalid metrics name would result in zero value rather than exception during initialization.

rafal-dudek:
This PR adds metric for connectionCount attribute from "Tomcat:name=*,type=ThreadPool" mBean.

Fixes #1598

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
@@ -147,6 +147,12 @@ public class TomcatMetrics implements MeterBinder {
                     .tags(allTags)
                     .baseUnit(BaseUnits.THREADS)
                     .register(registry);
+
+            Gauge.builder("tomcat.threads.active", mBeanServer,
+                    s -> safeLong(() -> s.getAttribute(name, "connectionCount")))
+                    .tags(allTags)
+                    .baseUnit(BaseUnits.CONNECTIONS)
+                    .register(registry);
         });
     }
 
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
```diff
@@ -147,6 +147,12 @@ private void registerThreadPoolMetrics(MeterRegistry registry) {
                     .tags(allTags)
                     .baseUnit(BaseUnits.THREADS)
                     .register(registry);
+
+            Gauge.builder("tomcat.threads.active", mBeanServer,
```
shakuzen: Is the connectionCount necessarily equal to the number of active threads in the thread pool? Starting an application locally and looking at the MBean, I see `maxThreads` is 200, but `maxConnections` is 10,000.

rafal-dudek: I set baseUnit to "connections", so whole metric is "tomcat_threads_active_connections".
Maybe you have a suggestion for a better name?

shakuzen:
Sorry to not follow-up on this pull request so I could merge this. Thanks for proposing it, but I have recently merged #2248 which supersedes this one. It includes the connectionCount as well as a couple other metrics.

Issue #1080, opened 2018-12-17
Possible inconsistence between timer metrics for Executor and ExecutorService


fragonib:
Hi!

The latest version of Micrometer (1.1.1) supports per-execution timer metrics. However, Is micrometer registering different metrics, related to task timming, for different type of executors?

* **java.util.concurrent.Executor:** both timer metrics `executor.execution` and `executor.idle`
* **java.util.concurrent.ExecutorService:** just timer metric `executor` 

If so, what's the reason behind this?

# Involved code:

**io.micrometer.core.instrument.internal.TimedExecutor**
```java
public class TimedExecutor implements Executor {
    private final MeterRegistry registry;
    private final Executor delegate;
    private final Timer executionTimer;
    private final Timer idleTimer;

    public TimedExecutor(MeterRegistry registry, Executor delegate, String executorName, Iterable<Tag> tags) {
        this.registry = registry;
        this.delegate = delegate;
        this.executionTimer = registry.timer("executor.execution", Tags.concat(tags, "name", executorName));
        this.idleTimer = registry.timer("executor.idle", Tags.concat(tags, "name", executorName));
    }
``` 

**io.micrometer.core.instrument.internal.TimedExecutorService**
```java
public class TimedExecutorService implements ExecutorService {
    private final ExecutorService delegate;
    private final Timer timer;

    public TimedExecutorService(MeterRegistry registry, ExecutorService delegate, String executorServiceName, Iterable<Tag> tags) {
        this.delegate = delegate;
        this.timer = registry.timer("executor", Tags.concat(tags ,"name", executorServiceName));
    }
```

fragonib:
Hi, Is this difference by design? What are the reasons behind that different behaviour?

Thank you.

mweirauch:
If I am not totally mistaken it's missing the idle timings. (Apart from the differing metric name for the execution timer: `executor.executions` vs `executor`. Unfortunately.)
So all `Callable`s and `Runnable`s would need to be wrapped again to get the "idle" timings for the scheduled/submitted tasks. Is that what is your question about or are you just irritated about the different metric names?

fragonib:
I'm just concerned by both:

- Same timer metric concept, but with different metric name on both executors: `Executor` and `ExecutorService`.
- Support for task timming in `ScheduledExecutorService` as showed on https://github.com/micrometer-metrics/micrometer/issues/1021 for specific methods of this executor .

Is there any plan to address both?

[...]

PR #1608, opened 2019-09-29
ScheduledExecutorService timings

### LINKED ISSUE #1021: Support ScheduledExecutorService per-execution monitoring
Hi!

The latest version of Micrometer (1.1.0) supports per-execution `ExecutorService` monitoring via `TimedExecutorService`.

In Project Reactor, we use `ScheduledExecutorService` (extended version of `ExecutorService`).
Would be nice if Micrometer will add it to `ExecutorServiceMetrics`, so that we can transparently reuse the infrastructure provided by Micrometer.

See https://github.com/reactor/reactor-core/pull/1281

shakuzen:
Adds support specifically for `ScheduledExecutorService`.

Resolves #1021

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedScheduledExecutorService.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedScheduledExecutorService.java
@@ -1,0 +1,148 @@
+/**
+ * Copyright 2019 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedScheduledExecutorService.java
```diff
@@ -0,0 +1,149 @@
...
+    private final ScheduledExecutorService delegate;
+    private final Timer executionTimer;
+    private final Timer idleTimer;
+    private final Counter scheduledOnce;
+    private final Counter scheduledRepetitively;
+
+    public TimedScheduledExecutorService(MeterRegistry registry, ScheduledExecutorService delegate, String executorServiceName, Iterable<Tag> tags) {
+        this.registry = registry;
+        this.delegate = delegate;
+        // TODO: IMO, it would make more sense to name this timer "executor.execution", but that's a breaking change
```
shakuzen: This is discussed in #1080 as we're not consistent on the naming. However, we do not want to break existing users. In Micrometer 2.0, we can correct this. In the meantime, we either need to live with the inconsistency and not as specific name, or we need to duplicate the timer so there's both `executor` and `executor.execution` until 2.0 when we can remove `executor`.

shakuzen:
@bsideup feel free to review this if you have time.

shakuzen:
Merged via 43fe77e

PR #1620, opened 2019-10-01
Use JMX to expose tomcat sessions metrics

baptistemesta:
This avoid using the manager and allows to have sessions related metrics
even if the application is not a spring application.

This is a Work in progress.
If maintainers think it's a good idea I can work on that and:

- [ ] Replace Manager by JMX completely
- [ ] Update tests and add new test to verify the behavior with JMX
- [ ] Update the autoconfigurations

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/Pair.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/Pair.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright 2019
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


shakuzen:
> This avoid using the manager and allows to have sessions related metrics
> even if the application is not a spring application.

Could you elaborate more on why using the Tomcat `Manager` makes it not usable outside of Spring applications?
Retrieving metrics from JMX queries has a track record of being fragile and error-prone (a recent example is #1778). When the same metrics are available from a strongly typed and versioned API, we generally prefer that.

baptistemesta:
Hi,
Thank you for taking the time to check that.

What I meant is that, if you are in an application where the tomcat is not embedded (i.e. an old school WAR application), the application does not have access to the instance of `org.apache.catalina.Manager`
So it not possible to use any `tomcat sessions metrics`

When using Spring boot, Spring starts the tomcat for you. It keeps the reference to the `Manager` and everything works well.

I did not found a way to retrieve this Manager in a classic WAR application, maybe I just missed something obvious.

shakuzen:
> What I meant is that, if you are in an application where the tomcat is not embedded (i.e. an old school WAR application), the application does not have access to the instance of `org.apache.catalina.Manager`

I see what you mean now. Thanks for the clarification.

> I did not found a way to retrieve this Manager in a classic WAR application, maybe I just missed something obvious.

I'm not sure either. We'll have to check with someone more knowledgeable about that than me. I suspect the answer from Tomcat might be that the way to do it is to use the JMX query, though.

PR #1623, opened 2019-10-02
Use internal logging for LoggingMeterRegistry

izeye:
Now that we have proper internal logging, it'd be good to use internal logging in `LoggingMeterRegistry` for consistency.


micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
```diff
@@ -50,6 +51,8 @@
  */
 @Incubating(since = "1.1.0")
 public class LoggingMeterRegistry extends StepMeterRegistry {
+    private final static InternalLogger log = InternalLoggerFactory.getInstance(PushMeterRegistry.class);
```
shakuzen:
```suggestion
    private final static InternalLogger log = InternalLoggerFactory.getInstance(LoggingMeterRegistry.class);
```

izeye: @shakuzen Thanks! 😅

Issue #1622, opened 2019-10-02
Lost Counter increments when registering Counters from multiple threads using CompositeMetricsRegistry with Prometheus


mcekovic:
There is an issue in the Micrometer 1.3.0 (also checked from 1.1.4 onward, but did no check earlier versions) that manifests in the lost Counter increments when registering Counters from multiple threads using CompositeMetricsRegistry with Prometheus.

Here is the minimal example code:

```
package test.micrometer;

import java.util.concurrent.*;

import org.junit.jupiter.api.*;

import io.micrometer.core.instrument.*;
import io.micrometer.core.instrument.composite.*;
import io.micrometer.prometheus.*;

import static org.assertj.core.api.Assertions.*;

class MicrometerCounterRegistrationTest {

	private MeterRegistry meterRegistry = createMeterRegistry();

	private static MeterRegistry createMeterRegistry() {
		CompositeMeterRegistry meterRegistry = new CompositeMeterRegistry();
		meterRegistry.add(new PrometheusMeterRegistry(PrometheusConfig.DEFAULT));
		return meterRegistry;
	}

	@Test
	void testTaggedCounterConcurrently() throws InterruptedException {
		int count = 10000;
		int tagCount = 100;
		var executor = Executors.newFixedThreadPool(10);
		for (int i = 0; i < count; i++) {
			int tagValue = i % tagCount;
			executor.execute(() ->
				createCounter(tagValue).increment()
			);
		}
		executor.shutdown();
		executor.awaitTermination(1L, TimeUnit.DAYS);
		for (int i = 0; i < tagCount; i++)
			assertThat(createCounter(i).count()).isEqualTo(count / tagCount);
	}

	private Counter createCounter(int tagValue) {
		return Counter.builder("test.counter").tag("test.tag", String.valueOf(tagValue)).register(meterRegistry);
	}
}
```

Running the above test couple of times will provoke the issue at least once.
The issue seems to **not** manifest using PrometheusMetricsRegistry directly, **nor** also seems to manifest when using SimpleMetricsRegistry inside the CompositeMeterRegistry.
The sample code runs with JDK 11, JUnit 5, AssertJ and Micrometer (Core and Prometheus) as only dependencies, though issue manifests with JDK 8 too.
Pre-registering Counters in the main thread eliminates the issue.

izeye:
@mcekovic Thanks for the report and the test! I could reproduce this locally with your test, even with `SimpleMeterRegistry`.

[...]

PR #1628, opened 2019-10-03
Fix a race condition in MeterRegistry.getOrCreateMeter()

izeye:
This PR resolves a race condition between meter addition and
composite meter's children addition via meter-added event listener in `MeterRegistry.getOrCreateMeter()`.

Closes gh-1622

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/composite/CompositeMeterRegistryTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/composite/CompositeMeterRegistryTest.java
@@ -468,7 +468,7 @@ class CompositeMeterRegistryTest {
             });
         }
         executor.shutdown();
-        assertThat(executor.awaitTermination(1L, TimeUnit.DAYS)).isTrue();
+        assertThat(executor.awaitTermination(1L, TimeUnit.SECONDS)).isTrue();
         for (int i = 0; i < tagCount; i++) {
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/composite/CompositeMeterRegistryTest.java
```diff
@@ -445,4 +448,31 @@ void stopTrackingMetersThatAreRemoved() {
...
+        for (int i = 0; i < count; i++) {
+            int tagValue = i % tagCount;
+            executor.execute(() -> {
+                Counter counter = Counter.builder(meterName).tag(tagName, String.valueOf(tagValue))
+                        .register(this.composite);
+                counter.increment();
+            });
+        }
+        executor.shutdown();
+        assertThat(executor.awaitTermination(1L, TimeUnit.DAYS)).isTrue();
```
shakuzen: Let's not block for 1 day in a test because this will require manual intervention to stop builds, or rely on the build system to cancel build jobs that are running for too long. I know there should be no reason the tasks don't complete in a timely manner, but I think it'll be better to be safe here with a more reasonable wait time like 1 second.

izeye: @shakuzen Thanks for the feedback! I changed it in ed62622 as suggested.

PR #1632, opened 2019-10-05
Add log for gauge's value function application failure

izeye:
This PR adds log for gauge's value function application failure in `DropwizardMeterRegistry`. For the verbosity concern from @shakuzen, I chose [an approach from Spring framework](https://github.com/spring-projects/spring-framework/blob/972337e8f9424051d90a20e4680eb61037fe5f36/spring-beans/src/main/java/org/springframework/beans/factory/support/DisposableBeanAdapter.java#L261) to lower the verbosity, but I'm open to any suggestion. An approach suggested from @checketts in https://github.com/micrometer-metrics/micrometer/pull/1532#issuecomment-527944141 also seems good to me.

See gh-1532

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/dropwizard/DropwizardMeterRegistry.java
@@ -31,6 +31,7 @@ import io.micrometer.core.util.internal.logging.InternalLoggerFactory;
 
 import java.lang.ref.WeakReference;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.ToDoubleFunction;
 import java.util.function.ToLongFunction;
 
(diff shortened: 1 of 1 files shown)


shakuzen:
I think this approach is still too verbose. A key difference between the referenced Spring framework code and this code is that the framework logging will only happen when an error occurs on destroying a bean, which generally only happens on application shutdown. In our case, the logging would happen every time the gauge value is read, which is generally every publish. I wouldn't mind having that detailed logging at debug level, though. What does everyone else think?

izeye:
@shakuzen Thanks for the feedback! Some systems might rely on metrics for notifications, so failures on collecting metrics might be crucial for them. That's why I chose the warn level for it. Maybe, @checketts's suggested approach might work for both. What do you think about the approach which uses a warn level at first and then uses a debug or trace level for the rest?

shakuzen:
> What do you think about the approach which uses a warn level at first and then uses a debug or trace level for the rest?

That sounds like a good solution to me.

izeye:
@shakuzen Thanks for the feedback! I changed in f34df3e as discussed.

Issue #1629, opened 2019-10-03
Configurable _source.enabled Elastic mapping property


lnorregaard:
From Elastic's documentation:
https://www.elastic.co/guide/en/elasticsearch/reference/7.3/mapping-source-field.html

> The metrics use case is distinct from other time-based or logging use cases in that there are many small documents which consist only of numbers, dates, or keywords. There are no updates, no highlighting requests, and the data ages quickly so there is no need to reindex. Search requests typically use simple queries to filter the dataset by date or tags, and the results are returned as aggregations.
> In this case, disabling the _source field will save space and reduce I/O.

This means that it is not possible to see the fields in the index and Kibana view will not show the metrics. 

I would make a pull request to change to enabled: true for Elasticsearch 7.3 and up

robachmann:
Thank you, Lars, for offering to make a PR.
This [Stack Overflow](https://stackoverflow.com/a/57430154/3105453) answer also concludes that setting `enabled: true` would solve the issue of not seeing metrics in Kibana. I think this would be a quick-fix until there is native support for Elastic APM as requested in https://github.com/micrometer-metrics/micrometer/issues/793

shakuzen:
The quoted section of the documentation explains why disabling the `_source` field is a good idea for metrics, but you're proposing enabling it.

Just so we're on the same page, the fields are there (you can check them in the Kibana index pattern, for instance) and you can query/aggregate them, but they don't show the raw values in the Kibana Discovery page for each document. I can see how this makes ad-hoc querying the metrics difficult unless you know the metric names/tags already. It would be nice if Kibana could still auto-complete the field names/values for this purpose.

The proposal to enable the  `_source` field would remedy this pain, but is it worth everyone paying the cost of including the `_source` field for metrics data? I'm not sure how much that cost is, to be honest. I wonder if there is some other way we can enable easier querying of the data without the overhead of enabling the `_source` field.
@xeraa sorry to ping you suddenly, but I'd love to get any thoughts from you on the above, if you have time to help.

Other Elasticsearch/Kibana users also feel free to give your input on this, as the proposed change would affect the default behavior for all users of the Micrometer Elasticsearch registry.

---

The above being said, there is the workaround currently of specifying your own template that uses `"_source": { "enabled": true }`. Even if we decide keeping the current default is best, another option would be to make a configuration property for whether to enable the `_source` field, so users don't have to copy the template just to modify that.

[...]

PR #1633, opened 2019-10-06
Changed Elastic Template source to true for Version 7 and above (#1629)

lnorregaard:
After 7.3 the _source needs to be enabled for getting the fields visible on the index mapping.


shakuzen:
I don't think `_source` set to `true` is a good default. It almost certainly should not be used in production, as it adds unnecessary overhead. For that reason, we'll not be accepting this pull request. Thank you for taking the time to propose it, nonetheless.

PR #1638, opened 2019-10-10
Add Javadoc indicating that SLA units are nanoseconds when they are meant for a Timer

### LINKED ISSUE #1414: Using a `MeterFilter` to modify the SLAs of an existing timer is not intuitive
`Timer.Builder::sla` takes a `Duration...`, which is self-explanatory.

If you need to create/modify SLA boundaries for that `Timer`, (I think) you need to use a `MeterFilter` to create a new `DistributionStatisticConfig` and merge it with the existing one.

`DistributionStatisticConfig.Builder::sla` takes a `long...`, and the Javadoc provides no guidance on the units.  Only by reading the `Timer.Builder::sla` source did I discover that the long values were meant to be nanoseconds.  I understand that `Timer`s aren't the only object that use a distribution configuration, but there should still be an easier way to figure out how to update them.

Possible remedies:

* update the `DistributionStatisticConfig.Builder::sla` to mention the unit for timers
* include a `DistributionStatisticConfig.Builder::sla` overload which takes a `Duration...`, with a warning in the Javadoc that this should only be used for `Timer`s.

(In case you're curious why I ran into this scenario, I'm attempting to instrument an existing Dropwizard/Jetty based application with Micrometer and was attempting to set up distribution statistics for a `MetricsApplicationEventListener`/`MetricsRequestEventListener`)

fitzoh:
Fixes #1414

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
@@ -199,7 +199,8 @@ public class DistributionStatisticConfig implements Mergeable<DistributionStatis
     /**
      * Publish at a minimum a histogram containing your defined SLA boundaries. When used in conjunction with
      * {@link #percentileHistogram}, the boundaries defined here are included alongside other buckets used to
-     * generate aggregable percentile approximations.
+     * generate aggregable percentile approximations.  If The {@link DistributionStatisticConfig} is meant for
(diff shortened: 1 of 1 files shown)


shakuzen:
Thanks for the pull request. Could you rebase the changes to target the `1.1.x` branch so we can get this in patch releases for 1.1.x, 1.3.x and the upcoming 1.4? I can also do it when merging, otherwise.

fitzoh:
That what you're looking for @shakuzen ?

shakuzen:
> That what you're looking for @shakuzen ?

Yes. Thank you.

PR #1641, opened 2019-10-14
Polish TimeWindowMax

izeye:
This PR polishes `TimeWindowMax`.


shakuzen:
What's the motivation for this change? I worry a bit if this change will affect GC pressure.

izeye:
@shakuzen I just tried to reduce duplication. Which part makes you worry about GC?

jkschneider:
> Which part makes you worry about GC?

I'm guessing the lambda. Each unique lambda in a class increases its metaspace footprint, e.g. something like this per lambda. So strictly speaking, probably no GC impact, but there is a slight memory footprint impact.

image

shakuzen:
It's hard to remember at this point what I was thinking (sorry I wasn't more specific), but I think my concern was about the addition of lambdas with variable capture on our "hot path" for recording metrics. I'm not sure there's any per-invocation cost associated with that like I must have been thinking. I wrote a JMH test locally to benchmark the changes and didn't notice a significant difference either.

PR #1647, opened 2019-10-17
NewRelicMeterRegistry Java Agent Insights API Support

### LINKED ISSUE #1540: New Relic Java Agent MeterRegistry
@shakuzen We are users of the New Relic Java Agent and would like a MeterRegistry that delegates to the New Relic Insights API available on Agent.
We have created our own delegating implementation that is very similar to the existing NewRelicMeterRegistry, but it provides several additional benefits:
- Avoids the need for separate API key provisioning/management, which is required for the existing implementation (also not supported by the New Relic PCF Tile)
- Can limit security proxy/firewall connectivity to collection endpoints used by Agent vs. the Insights REST API (single agent/client publishing metrics to a single monitoring system)
- Agent provides a reservoir/buffer for metrics processing, similar to Agent APM metrics.

Additionally, we have a MeterRegistryCustomizer/MeterFilter that denies metrics that are already collected by the Agent to avoid duplication, which includes; jvm (buffer, classes, memory, threads), process (cpu, files, start, uptime); system (cpu, load) and http.server.requests.

Finally, we have an MetricsExportAutoConfiguration implementation that load our MeterRegistry based on the presence of the Agent. It also sets the Agent's instanceName and common tags for applicationName and instanceIndex for easier metric correlation via Insights NRQL

neiljpowell:
This PR is for Issue https://github.com/micrometer-metrics/micrometer/issues/1540
NewRelicMeterRegistry support for delegating to the New Relic Java Agent Insights API for previously mentioned benefits. Construction/Builder defaults to original behavior using Insights REST API.

Implementation notes:
- NewRelicClientProvider interface introduced for easy introduction of Agent support. Bulk of the original registry implementation moved to the Http client provider. Separate client provider impls will also work well for conditional Boot auto configuration based on presence of the Agent, etc. It may also help with incorporating future enhancements like dimensional metrics noted in PR https://github.com/micrometer-metrics/micrometer/pull/1588
- MeterRegistryCustomizer/MeterFilter mentioned in the Issue for preventing duplicate metrics collected by the Agent would be included with the accompanying SpringBoot NewRelicMetricsExportAutoConfiguration updates.

@shakuzen @checketts @breedx-nr

Resolves #1540


checketts:
@neiljpowell I'm not too familiar with New Relic's offerings. This PR has a lot of changes for the NewRelic registry, however New Relic has created their own dimensional registry (https://github.com/newrelic/micrometer-registry-newrelic#new-relic-micrometer-registry) and has mentioned that it supersedes this one.

Can you summarize the options/approaches at a high level the New Relic users have and how thise PR changes that?

neiljpowell:
@checketts Hi. Yes, this enhancement provides the benefits mentioned in the originating issue noted above.  It allows users of the New Relic Java agent to continue publishing metrics to Insights in a more consolidated way, that doesn't require additional API Key provisioning. Delegating to the agent also provides buffering not provided by the current implementation. It can potentially simplify security proxy config by reducing the outbound destinations needed. @shakuzen 's preference for incorporation into the current meter registry resulted in the large amount of change. I was originally going to contribute a separate meter registry. However, the approach taken left most of the original code intact.

Regarding @breedx-nr 's announcement in my other PR, I had previously discussed this Agent based enhancement with him and other New Relic folks. They supported this Insights publishing option, as it may benefit many of their users which have a high likelihood of also using the agent with their apps.

neiljpowell:
@checketts @shakuzen This PR changed the registry's baseTimeUnit to milliseconds. It was one of the few registry impls with seconds, which is an eternity when monitoring all the things. Any concerns?

checketts:
> which is an eternity when monitoring all the things 

Would this change break user's dashboards they had already created? Does it actually impact the outputted data since you still have the decimals in seconds? Isn't 10 millis and 0.01 seconds the same?

I'm not opposed to making the change if we don't need to worry about breaking existing users. Or we should wait for a 2.x release to do the break.

neiljpowell:
@checketts To ensure backward compatibility, baseTimeUnit has been reverted. It affects output.

shakuzen:
I've added the linked issue to the 1.4 milestone. I'll come back to more thoroughly review this in time for that. Others feel free to leave reviews in the meantime.

implementations/micrometer-registry-new-relic/build.gradle
```diff
@@ -1,6 +1,7 @@
 dependencies {
     compile project(':micrometer-core')
     compile 'org.slf4j:slf4j-api:1.7.+'
+    compile 'com.newrelic.agent.java:newrelic-api:5.+'
```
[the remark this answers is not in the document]
neiljpowell: Yes, good catch. Done.  
FYI, I had already tried to minimize transitive dependencies. The newrelic-api dependency is light weight, with only interfaces and No-Op defaults/impls.  The newrelic-agent dependency has the actual implementation that is typically provided on a host/in a container.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
```diff
@@ -15,82 +15,79 @@
...
      * @deprecated Use {@link #builder(NewRelicConfig)} instead.
      */
-    @Deprecated
-    public NewRelicMeterRegistry(NewRelicConfig config, Clock clock, ThreadFactory threadFactory) {
-        this(config, clock, threadFactory, new HttpUrlConnectionSender(config.connectTimeout(), config.readTimeout()));
-    }
-
-    // VisibleForTesting
-    NewRelicMeterRegistry(NewRelicConfig config, Clock clock, ThreadFactory threadFactory, HttpSender httpClient) {
+    @Deprecated  // VisibleForTesting
```
shakuzen: No need to mark this deprecated if it is package-private. We marked it before because it was public API. It's been deprecated since 1.1.0, so it is probably fair to remove it from public API at this point.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
```diff
@@ -15,82 +15,79 @@
...
-            throw new MissingRequiredConfigurationException("eventType must be set to report metrics to New Relic");
-        }
-        if (StringUtils.isEmpty(config.accountId())) {
-            throw new MissingRequiredConfigurationException("accountId must be set to report metrics to New Relic");
+        if (clientProvider == null) {
+            throw new MissingRequiredConfigurationException("clientProvider required to report metrics to New Relic");
         }
-        if (StringUtils.isEmpty(config.apiKey())) {
-            throw new MissingRequiredConfigurationException("apiKey must be set to report metrics to New Relic");
+        if (namingConvention == null) {
```
shakuzen: There's no way from a public constructor for `namingConvention` or `threadFactory` to be `null`. `MissingRequiredConfigurationException` is conventionally used for things users can configure that are required but might be missing.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
```diff
@@ -105,181 +102,10 @@ public void start(ThreadFactory threadFactory) {
...
-            httpClient.post(insightsEndpoint)
-                    .withHeader("X-Insert-Key", config.apiKey())
-                    .withJsonContent(events.peek(ev -> totalEvents.incrementAndGet()).collect(Collectors.joining(",", "[", "]")))
-                    .send()
-                    .onSuccess(response -> logger.debug("successfully sent {} metrics to New Relic.", totalEvents))
-                    .onError(response -> logger.error("failed to send metrics to new relic: http {} {}", response.code(), response.body()));
-        } catch (Throwable e) {
-            logger.warn("failed to send metrics to new relic", e);
-        }
+        clientProvider.publish(this, getMeters());
```
shakuzen: Why pass both the MeterRegistry and the list of meters? We can get the latter from the first, and the Http implementation doesn't use the passed list of meters.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicClientProvider.java
```diff
@@ -0,0 +1,87 @@
...
+import io.micrometer.core.instrument.TimeGauge;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.config.NamingConvention;
+
+/**
+ * @author Neil Powell
+ */
+public interface NewRelicClientProvider {
+    //long task timer
+    static final String DURATION = "duration";
```
shakuzen: `static` and `final` modifiers are redundant for interface fields (courtesy of IntelliJ).

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicHttpClientProviderImpl.java
```diff
@@ -0,0 +1,296 @@
...
+        return Arrays.stream(attributes)
+                .map(attr -> 
+                        (attr.getValue() instanceof Number) 
+                            ? ",\"" + attr.getName() + "\":" + DoubleFormat.wholeOrDecimal(((Number)attr.getValue()).doubleValue())
+                            : ",\"" + attr.getName() + "\":\"" + namingConvention.tagValue(attr.getValue().toString()) + "\""
+                )
+                .collect(Collectors.joining("", "{\"eventType\":\"" + escapeJson(eventType) + "\"", tagsJson + "}"));
+    }
+
+    void sendEvents(Meter.Id id, Object attributesObj) {
```
shakuzen: This looks like it could be simplified by changing the type of the second parameter to `Stream<String>` instead of `Object`.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicHttpClientProviderImpl.java
```diff
@@ -0,0 +1,296 @@
...
+        this.insightsEndpoint = config.uri() + "/v1/accounts/" + config.accountId() + "/events";
+    }
+
+    @Override
+    public void publish(NewRelicMeterRegistry meterRegistry, List<Meter> meters) {
+        // New Relic's Insights API limits us to 1000 events per call
+        // 1:1 mapping between Micrometer meters and New Relic events
+        for (List<Meter> batch : MeterPartition.partition(meterRegistry, Math.min(config.batchSize(), 1000))) {
+            for (Meter meter : batch) {
+                sendEvents(meter.getId(),
```
shakuzen: This looks like it'll make an HTTP call for each meter. Previously it made one HTTP call per batch, if my understanding is correct.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
neiljpowell: Of course.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
neiljpowell: Ah, was being overly protective

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
neiljpowell: got it

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicHttpClientProviderImpl.java
neiljpowell: Updated.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicHttpClientProviderImpl.java
neiljpowell: Good catch. Reverted.

neiljpowell:
@shakuzen I've addressed all review comments. I'm happy to look at anything else you may find. Thanks.

shakuzen:
@neiljpowell Sorry for the delay on this. Would you be able to rebase your changes here on the latest master so it is easier for me to merge?

neiljpowell:
@shakuzen Sure. I should be able to get it done later today.

neiljpowell:
@shakuzen Master has been merged in.  I had to adjust build.gradle. Please take a look.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicClientProvider.java
```diff
@@ -0,0 +1,85 @@
...
+    String TOTAL = "total";
+    String AVG = "avg";
+    String COUNT = "count";
+    //timer
+    String TOTAL_TIME = "totalTime";
+    String TIME = "time";
+    //gauge
+    String VALUE = "value";
+    //counter
+    String THROUGHPUT = "throughput";  //TODO Why not "count"? ..confusing if just counting something
```
shakuzen: This is probably an oversight. As the philosophy of Micrometer, we want users to focus on rates of counts rather than the raw count. Accordingly, the intention was probably to ship the rate (throughput) of the counter rather than the step count. I'll open a separate issue to look at that. In this pull request, it is probably best to keep the existing behavior and limit its changes to the purpose of the pull request.

PR #1651, opened 2019-10-18
Add support for writing counters and gauges as single multi-field line in InfluxMeterRegistry

### LINKED ISSUE #949: InfluxMeterRegistry writing counters and gauges as single multi-field line
Would it pose any problems if `InfluxMeterRegistry` wrote counters and gauges as a single line with multiple fields?

The aggregate of all our application instances, sending each counter and gauge as a separate metric with their tags, is consuming a lot of memory in our InfluxDB instance. The thought is that by writing the counters and gauges (which just have one field) as a single line, this would dramatically reduce memory consumption.

sondemar:
This PR adds support for writing counters and gauges as a single multi field line in InfluxMeterRegistry grouping these meters by type, prefix, tags and prefix which is provided as a parameter. This prefix will be target name of a meter. For not grouped meters an existing logic will be applied.

resolves #949

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterPartition.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterPartition.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


jonatan-ivanov:
There was no interest in the original issue (https://github.com/micrometer-metrics/micrometer/issues/949) since 2018. Since that one is closed, we’ll close this PR for now, and if there still will be any interest, please comment on the original issue (https://github.com/micrometer-metrics/micrometer/issues/949) and we’ll consider re-opening it.

PR #1658, opened 2019-10-23
Constructors to customize InstrumentedQueuedThreadPool

### LINKED ISSUE #1639: Add ability to customize InstrumentedQueuedThreadPool
Right now [InstrumentedQueuedThreadPool](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPool.java) can be constructed with default underlying values only by implicitly calling no-arg `QueuedThreadPool` constructor, which looks like this:
```java
public QueuedThreadPool()
{
    this(200);
}
```
While I assume default logic might be useful in some cases, having an opportunity to tweak it is something I would really like to have (I have finely tuned threadpool in production). 

If team is fine with this proposal I can provide a PR, which will add bunch of overloaded constructors.

IgorPerikov:
I decided to provide all possible constructors for convenience and further simplicity of using (although with one big constructor, people might figure out what to pass to the missing parameters, it's still not as good as plenty of 'native' options)

fixes #1639

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPool.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPool.java
@@ -19,12 +19,47 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
 
+import java.util.concurrent.BlockingQueue;
+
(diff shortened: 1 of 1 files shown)


IgorPerikov:
@shakuzen Would you mind taking a look?

IgorPerikov:
@shakuzen @izeye kindly ping

shakuzen:
Sorry for the delay on review here. I plan to start reviewing things planned for 1.4 once I finish the patch releases this week.

michaelbannister:
We have (a subset of) this same code in our app, and it'd be nice to replace it with using the class directly provided by Micrometer.

Issue #1648, opened 2019-10-17
Dynatrace Load average Metrics - Error 400


nandhusriram:
I am using micrometer-registry-dynatrace jar version 1.3.0 with Spring Boot 2.1.9 and  I am seeing error code 400 while pushing  the load average metrics.

failed to create custom metric custom:system.load.average.1m in dynatrace: {"error":{"code":400,"message":"The timeseries id is only allowed to contain alphanumeric characters, '.', '_' and '-'. Also, it is not allowed to start with 'com.ruxit.builtin'."}}

Any ideas , may  be I am passing something wrong ?

izeye:
@nandhusriram Thanks for the report! I'm not a Dynatrace user, but I did a quick try, and I could reproduce it with [a sample](https://github.com/izeye/sample-micrometer-spring-boot/tree/dynatrace).

It seems to reject numeric characters in a time series ID unlike what it says in its error message. I think it'd be better to ask Dynatrace folks about it.

I wonder why this hasn't been reported earlier.

shakuzen:
@nandhusriram Thank you for reporting this, and thank you @izeye for investigating and reproducing the issue.
@nandhusriram let us know if you can reach out to Dynatrace support about this and get an official answer about the discrepancy of the error message as mentioned by @izeye.

izeye:
I got the following answer from the official Dynatrace support channel:

> A numeric character is supported, but it cannot be immediately after the dot.

I created https://github.com/micrometer-metrics/micrometer/pull/1661 to try to work around the constraints by prefixing "n" (from number). I'm not happy with it, but I couldn't find a better solution. We might be better to refrain from using leading numeric characters in meter names for Dynatrace in the future. In the PR, I also changed to sanitize meter names as it seems to be missing.

checketts:
> We might be better to refrain from using leading numeric characters in meter names for Dynatrace in the future.

Do the numbers appear after the dot because of hierarchical naming? Like `status` tag values? I don't think we can refrain in that case.

`n` isn't intuitive, I think a character like `_` is more typical as not having significance and still being widely supported.

izeye:
> > We might be better to refrain from using leading numeric characters in meter names for Dynatrace in the future.
> 
> Do the numbers appear after the dot because of hierarchical naming? Like `status` tag values? I don't think we can refrain in that case.

@checketts  No, it's due to a meter name, not a tag value. Currently, `system.load.average.1m` is the only one that has the problem.

[...]

PR #1661, opened 2019-10-24
Log custom meter names having leading numeric characters for Dynatrace

izeye:
This PR changes to sanitize meter names for Dynatrace.

Closes gh-1648

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceNamingConvention.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceNamingConvention.java
@@ -18,6 +18,7 @@ package io.micrometer.dynatrace;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.lang.Nullable;
+import io.micrometer.core.util.internal.logging.WarnThenDebugLogger;
 
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/DynatraceNamingConventionTest.java
```diff
@@ -30,8 +36,18 @@ void nameStartsWithCustomAndColon() {
         assertThat(convention.name("mymetric", Meter.Type.COUNTER, null)).isEqualTo("custom:mymetric");
     }
 
+    @Test
+    void nameShouldBeSanitized() {
+        assertThat(convention.name("mymetric1._-,", Meter.Type.COUNTER, null)).isEqualTo("custom:mymetric1._-_");
```
shakuzen: Is it okay to start with `_` after a dot?

izeye: @shakuzen No, it will fail actually. As I mentioned in https://github.com/micrometer-metrics/micrometer/issues/1648#issuecomment-546126235, I just focused on handling `system.load.average.1m` as I thought other possible glitches could be handled with renaming them by users. If multiple non-alphabet characters should be handled, another unhappy alphabet prefix like "n" might be necessary. I'm not sure if there's any better solution for this.

shakuzen:
I was trying this out, and it doesn't fail on registration like we were intending. It fails on publishing, which seems significantly less helpful. Maybe a warn-then-debug log and skipping publishing that metric is the easiest thing to do, particularly for 1.1.x and 1.3.x. In the upcoming 1.4, we could try more risky changes like failing on registration. We need to get out a patch release for 1.1.x and 1.3.x today or early tomorrow, and I'd like to have some fix for Dynatrace users on this since it will fail to publish metrics for most users right now, I think.

shakuzen:
I have pushed 712411a for now so that we can get the fix for the known problem in the 1.1.8 release. We can continue to iterate here on letting users know there is a meter with a non-conforming name.

izeye:
@shakuzen Thanks! I changed this to log custom meter names having leading numeric characters rather than throwing exceptions.

PR #1663, opened 2019-10-25
Apply Gradle build cache

izeye:
This PR applies Gradle build cache. I could see that the build time has been reduced more than 20% locally although it won't be able to work in CircleCI unless there's any directory shared across builds.

Before applying Gradle build cache:

```
$ ./gradlew clean check
...
BUILD SUCCESSFUL in 2m 22s
292 actionable tasks: 234 executed, 58 up-to-date
$

$ ./gradlew clean check
...
BUILD SUCCESSFUL in 2m 12s
292 actionable tasks: 240 executed, 52 up-to-date
$

$ ./gradlew clean check
...
BUILD SUCCESSFUL in 2m 22s
292 actionable tasks: 240 executed, 52 up-to-date
$
```

After applying Gradle build cache:

```
$ ./gradlew clean check
...
BUILD SUCCESSFUL in 2m 24s
292 actionable tasks: 199 executed, 41 from cache, 52 up-to-date
$

$ ./gradlew clean check
...
BUILD SUCCESSFUL in 1m 50s
292 actionable tasks: 186 executed, 54 from cache, 52 up-to-date
$

$ ./gradlew clean check
...
BUILD SUCCESSFUL in 1m 47s
292 actionable tasks: 186 executed, 54 from cache, 52 up-to-date
$

$ ./gradlew clean check
...
BUILD SUCCESSFUL in 1m 43s
292 actionable tasks: 186 executed, 54 from cache, 52 up-to-date
$
```


shakuzen:
> it won't be able to work in CircleCI unless there's any directory shared across builds.

We can do that. We already do that with dependencies downloaded during the build. Where are files for the Gradle build cache? We don't do the best job of making a cache key, so it could be good timing to improve it.

https://github.com/micrometer-metrics/micrometer/blob/80ebe9a3ddf365ed26c684984409edcebc5a6ad3/.circleci/config.yml#L13-L23

gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.caching=true
```
shakuzen: During deploys/releases, a `gradle.properties` file is decrypted, so this would be overwritten during deploys. I can add this property to the encrypted file, though, if we end up going this route.

shakuzen: We no longer use an encrypted `gradle.properties` file, so this is not an issue anymore.

shakuzen:
I thought #2583 alone was enough to turn on build caching with Gradle Enterprise, but `org.gradle.caching=true` is still required (or CLI arg `--build-cache`).

PR #1670, opened 2019-10-25
Adds missing tests for mBeanServer in TomcatMetrics

### LINKED ISSUE #1619: Missing tests for mBeanServer in TomcatMetrics
We spotted that there is no effective test for metrics derived from 'mBeanServer' in TomcatMetrics (in contrast to Manager object) since invalid metrics name would result in zero value rather than exception during initialization.

Previously mentioned in https://github.com/micrometer-metrics/micrometer/issues/1598

padilo:
Closes #1619

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
@@ -15,6 +15,11 @@
  */
 package io.micrometer.core.instrument.binder.tomcat;
 
+import io.micrometer.core.instrument.MockClock;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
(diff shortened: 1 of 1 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
```diff
@@ -104,25 +115,63 @@ void mbeansAvailableAfterBinder() throws LifecycleException, InterruptedExceptio
...
+
+                resp.getOutputStream().write("yes".getBytes());
+            }
+        };
+
+        runTomcat(servlet, () -> {
+            TomcatMetrics.monitor(registry, null);
+
+            try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
+                HttpPost post = new HttpPost("http://localhost:61000/");
```
shakuzen: Just a note that it would be better in the future if we can change this to use a random available port rather than being hard-coded to 61000.

Issue #1606, opened 2019-09-26
Source information is not available in signalfx


zeldigas:
Current reporter for SignalFx does not send information about "source", that is by default equal to hostname.

SFX team's dropwizard reporter sends this information as dedicated dimension, appended to every data point: https://github.com/signalfx/signalfx-java/blob/master/signalfx-codahale/src/main/java/com/signalfx/codahale/reporter/AggregateMetricSenderSessionWrapper.java#L195

Micrometer implementation sets `source` field in dataPoint, but looks like it does not work as expected, because there is no dimension with a value having this info.

Expected result: 
1. It is possible to enable behavior that all reported data point has additional dimension with value equal to `source()` property. 
1. It is also should be allowed to define tag name  with `sf_` prefix so that reported data can keep compatibility with already existing data (some legacy examples for signal fx used `sf_source` for this info)

I'm fine if this behavior would be optional and enabled by special config parameter, e.g. `management.metrics.export.signalfx.reportSource`

zeldigas:
I'm ready to send PR for this issue if you are fine to add this improvement

shakuzen:
Right now we're just passing the configured `SignalFxConfig#source` as the `defaultSourceName` parameter to the `AggregateMetricSender` provided by SignalFx's Java client. I'm wondering why that doesn't result in the expected behavior and if the change belongs in Micrometer or SignalFx's Java client.

zeldigas:
Maybe some SFX folks could comment here. @mpetazzoni, @beccatortell, appreciate you can comment on the `source` field in protobuf spec compared to dimension

zeldigas:
@shakuzen I've did some search in SFX docs and while `source` field is present in protobuf descriptor, there is no such field in http api docs as well as other docs:
1. https://developers.signalfx.com/ingest_data_reference.html - API endpoint definition to send metric data
2. https://docs.signalfx.com/en/latest/getting-started/concepts/data-model.html - description of data model
3. https://docs.signalfx.com/en/latest/metrics-metadata/metrics-metadata.html#metrics-metadata - various metadata about metrics only includes dimensions

zeldigas:
@shakuzen do you have any comments? Are you waiting for some SignalFX rep to comment or we can count on some "de facto" behavior?

shakuzen:
Sorry for the delay in response. I was taking some time off and then busy with a conference.

[...]

PR #1674, opened 2019-10-27
Signalfx source dimension

zeldigas:
micrometer-metrics/micrometer#1606 added support for dimension with source value.

Dimension is off by default to keep existing behavior, but can be enabled with `includeSourceTag` property as well as dimension name can be specified with `sourceTag` property (default - `source`)


zeldigas:
@shakuzen when would you have a chance to look at this pr and provide feedback?

zeldigas:
Tumbleweed.gif

shakuzen:
> @shakuzen when would you have a chance to look at this pr and provide feedback?

I'm planning to review it in time for the 1.4 feature release. I have a couple of patch releases to work on before I get to that. So, this week at the earliest.

PR #1681, opened 2019-10-31
Counter value not getting updated correctly. It keeps printing count …

gdganesh:
…always 1 on counter.increment()

httpRequests.statistic.count.uri:1|c
httpRequests.statistic.count.uri:1|c
Fixed: onNext event to use count value rather than amount value. Because Amount is always fixed. After fix counter value increments correctly:
httpRequests.statistic.count.uri:1|c
httpRequests.statistic.count.uri:2|c


jonatan-ivanov:
Is this still a problem? Could you please provide a test that shows the wrong behavior?

shakuzen:
The current behavior is correct for statsd. It expects the amount incremented in the "counting" line, not the total count. See https://github.com/statsd/statsd/blob/master/docs/metric_types.md#counting

PR #1682, opened 2019-11-01
Stackdriver - Ability to configure credentials

saturnism:
Added an additional config so that the credentials can be configured programmatically.


implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
```diff
@@ -85,8 +85,8 @@
     @Nullable
     private MetricServiceClient client;
 
-    public StackdriverMeterRegistry(StackdriverConfig config, Clock clock) {
-        this(config, clock, DEFAULT_THREAD_FACTORY, () -> MetricServiceSettings.newBuilder().build());
+    protected StackdriverMeterRegistry(StackdriverConfig config, Clock clock) {
```
shakuzen: Why the change in access here? This will break code that was using this public constructor before.

shakuzen: @saturnism I'd like to include this feature in the upcoming release. When you have time, let me know about the above question.

saturnism: eak, not quite sure why it was changed ;) i've reverted it back. thnx for catching this!

saturnism:
not sure why this is failing; built locally :/

PR #1702, opened 2019-11-11
Aligning http clients tags between httpclient and okhttp

### LINKED ISSUE #1701: Potential incosistency between httpclient and okhttp metrics
In `OkHttp` metrics event listener, the `host` tag is used to denote the tagret url / host (https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java#L123) which is slightly different to apache httpclient metrics executor (https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/MicrometerHttpRequestExecutor.java#L137-L141) where `target.host` tag is being used. 

I personally find the latter better (especially many - including myself) use `host` tag to denote the machine code is running. I think `target.host` work better to describe the target host

Wouldn't it be nice'ish to to try to keep the tags coherent between different (two) http clients. I've noticed this while migrating httpclient to okhttp in one of the projects.

kubamarchwicki:
See #1701


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
```diff
@@ -114,13 +118,12 @@ private void time(CallState state) {
...
                 ? "NOT_FOUND" : urlMapper.apply(request);
 
-        Tags requestTags = requestAvailable ? getRequestTags(request) : Tags.empty();
+        Tags requestTags = requestAvailable ? getRequestTags(request).and(generateTagsForRoute(request)) : Tags.empty();
 
         Iterable<Tag> tags = Tags.concat(extraTags, Tags.of(
             "method", requestAvailable ? request.method() : "UNKNOWN",
             "uri", requestAvailable ? uri : "UNKNOWN",
-            "status", getStatusMessage(state.response, state.exception),
-            "host", requestAvailable ? request.url().host() : "UNKNOWN"
```
checketts: I like the change and unifying tags, however, we need to include a way for users to opt out and return to the current way to remain backwards compatible with their dashboard, etc.

You could do it either by providing a way to configure the old behavior, or provide something like a `OkHttpHostTagFilter` that is a meter filter that users could use to switch the tag back.

kubamarchwicki: Fair point. When I pushed the code, it occurred to me I haven't rethought backwards compatibility. What is the preferred way for micrometer? To default to the backwards-compatible way and opt-in for new behaviour. Or the other way round: default to the preferred way and have a way to opt-out and fallback to the old approach? 

I'd happily work on that further.

kubamarchwicki: Nevermind, I've reread your post and it contains the answer ;) 
I'll create a way to opt-out and return to the existing way of tagging.

checketts: I would rely on @shakuzen to make the call. But I believe he wants to save backwards incompatible changes for a 2.0 release.

So probably make it so the user would opt-in for the new behavior.

kubamarchwicki:
I've added the filter. Unfortunately the `TomcatMetricsTest` randomly fails? Is it a flaky test? Cause locally it fails once in a while, every hand full of builds.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpRemoveHostTagFilter.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @author Jakub Marchwicki
+ * @see io.micrometer.core.instrument.binder.httpcomponents.MicrometerHttpRequestExecutor#generateTagsForRoute(HttpContext)
+ */
+public class OkHttpRemoveHostTagFilter implements MeterFilter {
+
+    @Override
+    public Meter.Id map(Meter.Id id) {
+        List<Tag> newTags = id.getTags().stream()
+                .filter(t -> !t.getKey().equals("host")).collect(Collectors.toList());
```
checketts: If a user had a commonTag setup of `host='myHost'` is there a risk that this would accidentally strip out the commonTag `host` tag?

kubamarchwicki: That takes a key as a whole (as opposed to `contains()` method. So I think we are good. 
But to be explicite I'll include a custome `myHost` tag

kubamarchwicki: if a user had custom tag `host='myHost'` that would be overwritten by the `OkHttpMetricsEventListener` anyway. That was the exact scenario I was in, where we had some common tags that were overwritten by the metrics listener. In the spring configuration, we had a customizer, which value was overwritten

```java
MeterRegistryCustomizer<MeterRegistry> meterRegistryCustomizer(Environment environment) {
    String hostName = environment.getProperty("JMX_HOSTNAME");
    return registry -> registry.config().commonTags(
        "host", hostName != null ? hostName.split("\\.")[0] : "localhost"
    );
}
```

Maybe try and replace it with a commonTag setup, if present?

kubamarchwicki: I retreat my comment ;) There is no way to retrieve common tags explicitly, which inclines me towards: maybe bluntly overwriting a common tag by `OkHttpMetricsEventListener` is a bug in a first place?

checketts: I agree. I haven't used commonTags much myself since Prometheus has a built in concept that I leverage. However it seems like we should elevate commonTags so they always win and can't be overwritten on accident. (I think that is something we could work on in a separate issue)

Perhaps a better solution than the `OkHttpRemoveHostTagFilter` would be just to add a constructor arg in the `OkHttpMetricsEventListener` to disable it. I don't love it, but it does avoid the potential removing of a tag that was added via `commonTag` config.

shakuzen:
I'm planning to review this for 1.4 with a backwards compatible solution. We can keep the linked issue open for 2.0 to flip things to being consistent by default.

Issue #1711, opened 2019-11-19
apache commons pool2 statistics


chang-chao:
Some libraries(like jedis) are using apache commons pool2, and apache commons pool2 exposes jmx metrics, so it seems to make sense to export those metrics.

Reference:
https://commons.apache.org/proper/commons-pool/apidocs/org/apache/commons/pool2/impl/GenericObjectPool.html

izeye:
Resolved via #1712

Issue #1710, opened 2019-11-18
org.apache.commons.dbcp2.BasicDataSource statistic


skabashnyuk:
It would be nice to see some data from `org.apache.commons.dbcp2.BasicDataSource `
For example:
- maxTotal
- numActive
- maxIdle
- numIdle
- minIdle
- initialSize
- maxWaitMillis
- softMinEvictableIdleTimeMillis
- timeBetweenEvictionRunsMillis
- defaultQueryTimeout
- maxConnLifetimeMillis
- maxOpenPreparedStatements
- minEvictableIdleTimeMillis
- removeAbandonedTimeout
etc.

jkschneider:
Partially implemented with https://github.com/micrometer-metrics/micrometer/pull/1712

PR #1712, opened 2019-11-19
Add CommonsObjectPool2Metrics

chang-chao:
fixes https://github.com/micrometer-metrics/micrometer/issues/1711

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/commonspool2/CommonsObjectPool2Metrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/commonspool2/CommonsObjectPool2Metrics.java
@@ -102,35 +102,36 @@ public class CommonsObjectPool2Metrics implements MeterBinder, AutoCloseable {
             registerMetricsEventually(
                     type,
                     (o, tags) -> {
-                        registerGaugeForObject(registry, o, "NumIdle", tags, "The number of instances currently idle in this pool", "objects");
+                        registerGaugeForObject(registry, o, "NumIdle", "num.idle", tags, "The number of instances currently idle in this pool", "objects");
 
-                        registerGaugeForObject(registry, o, "NumWaiters", tags, "The estimate of the number of threads currently blocked waiting for an object from the pool", "threads");
+                        registerGaugeForObject(registry, o, "NumWaiters", "num.waiters", tags, "The estimate of the number of threads currently blocked waiting for an object from the pool", "threads");
 
-                        registerFunctionCounterForObject(registry, o, "CreatedCount", tags, "The total number of objects created for this pool over the lifetime of the pool", "objects");
-                        registerFunctionCounterForObject(registry, o, "BorrowedCount", tags, "The total number of objects successfully borrowed from this pool over the lifetime of the pool", "objects");
-                        registerFunctionCounterForObject(registry, o, "ReturnedCount", tags, "The total number of objects returned to this pool over the lifetime of the pool", "objects");
-                        registerFunctionCounterForObject(registry, o, "DestroyedCount", tags, "The total number of objects destroyed by this pool over the lifetime of the pool", "objects");
+                        registerFunctionCounterForObject(registry, o, "CreatedCount", "created.count", tags, "The total number of objects created for this pool over the lifetime of the pool", "objects");
+                        registerFunctionCounterForObject(registry, o, "BorrowedCount", "borrowed.count", tags, "The total number of objects successfully borrowed from this pool over the lifetime of the pool", "objects");
+                        registerFunctionCounterForObject(registry, o, "ReturnedCount", "returned.count", tags, "The total number of objects returned to this pool over the lifetime of the pool", "objects");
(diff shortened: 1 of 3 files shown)


skabashnyuk:
Can this code be reuse to implement https://github.com/micrometer-metrics/micrometer/issues/1710 `org.apache.commons.dbcp2.BasicDataSource statistic`?

chang-chao:
> Can this code be reused to implement #1710 org.apache.commons.dbcp2.BasicDataSource statistic?
because dbcp2 uses commons pool2, so I believe yes.all the common metrics of commons pool will be exposed, like `MaxIdle `, etc, and will check and confirm this.

For the metrics specific to dbcp, like `defaultQueryTimeout ` , this fix will not create them.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/commonspool2/CommonsObjectPool2Metrics.java
```diff
@@ -0,0 +1,331 @@
...
+        if (!mBeanServers.isEmpty()) {
+            return mBeanServers.get(0);
+        }
+        return ManagementFactory.getPlatformMBeanServer();
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        for (String type : TYPES) {
+            registerMetricsEventually(
```
checketts: Is there a non-JMX way to register these?

chang-chao: Other ways are not always possible as far as I know, because libraries using commons-pool2(like jedis or dbcp2) keeps the pool objects as internal properties, and we cannot get them.
and commons-pool exposes JMX, so using JMX seems to be a natural fit.

checketts: I agree JMX as the default is a good idea. I was more curious if there was a way if we did have access to the pool directly. I used commons pool in some of my own code and didn't have JMX enabled at the time. It would be nice to offer a way to still gather those metrics if posssible.

chang-chao: As we can see from the following code, JMX registration in commons-pool2 is enabled by default , unless you configure to disable it explicitly, which I don't think a serious application should do.
https://github.com/apache/commons-pool/blob/2d2049233940b10c7dd35df70a353d383f601053/src/main/java/org/apache/commons/pool2/impl/BaseObjectPoolConfig.java#L139

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/commonspool2/CommonsObjectPool2Metrics.java
```diff
@@ -0,0 +1,331 @@
...
+                        registerFunctionCounterForObject(
+                                registry,
+                                o,
+                                "DestroyedByBorrowValidationCount",
+                                tags,
+                                "The total number of objects destroyed by this pool as a result of failing validation during borrowObject() over the lifetime of the pool",
+                                "objects");
+
+                        registerTimeGaugeForObject(
+                                registry, o, "MaxBorrowWaitTimeMillis", "maxBorrowWaitTime", tags, "The maximum time a thread has waited to borrow objects from the pool");
```
jkschneider: Please use dot separated names like `max.borrow.wait.time`, which will be camel-cased by certain `NamingConvention` implementations. By keeping dot-separated names in the core code, we maximize the portability of the metric to many different monitoring systems.

chao-chang-paypay: @jkschneider thanks for pointing it out, fixed in https://github.com/micrometer-metrics/micrometer/pull/1712/commits/32c5c0d576e2b7f9761bfaaa0d2545dfcc5eb089

IcyEagle:
@chao-chang-paypay is this PR considered finished? Can I help you somehow?

I'm actually looking forward to this update 👍

jkschneider:
Thanks for the hard work on this @chao-chang-paypay. I can see that this was a non-trivial implementation, and you did a fantastic job on it! Polished a bit and merged with 2f7f51b43756cb37690e8ba098a9b9ed22418179.

CoderLan0668:
@chang-chao  @jkschneider  

I aready see these code in branch Master，but can not find tnem in the latest release version. What's the problem? and i can not wait anymore.

AnkitKaneri:
@chang-chao @jkschneider I am using 1.6.0 version of `micrometer-core` and `micrometer-registry-prometheus`. I am not able to find jedis metrics under `/actuator/prometheus` endpoint. Am I missing anything here?

izeye:
@AnkitKaneri There's no out-of-the-box support for `CommonsObjectPool2Metrics` from Spring Boot, so you need to configure it yourself.

See https://github.com/spring-projects/spring-boot/issues/23525#issuecomment-703462483

PR #1715, opened 2019-11-21
Add a binder for Jetty connector metrics

### LINKED ISSUE #1704: Add Jetty connector metrics
It's often useful to see how Jetty's connectors are performing, in terms of connection counts and inbound/outbound traffic throughput.

So I'd like to propose adding a meter binder for these metrics. I'm happy to contribute a PR for this (I have some tested code already available).

tomakehurst:
Resolves #1704

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
@@ -19,12 +19,17 @@ import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.binder.MeterBinder;
 import org.eclipse.jetty.io.ConnectionStatistics;
 
+import static java.util.Collections.emptyList;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 public class JettyConnectionMetrics implements MeterBinder {
 
     private final ConnectionStatistics connectionStatistics;
     private final Iterable<Tag> tags;
+
+    public JettyConnectionMetrics(ConnectionStatistics connectionStatistics) {
+        this(connectionStatistics, emptyList());
+    }
 
     public JettyConnectionMetrics(ConnectionStatistics connectionStatistics, Iterable<Tag> tags) {
         this.connectionStatistics = connectionStatistics;
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
@@ -45,6 +45,7 @@ public class JettyConnectionMetrics implements MeterBinder {
                 .tags(tags)
                 .description("The total number of connections")
                 .register(registry);
+
         Gauge.builder("jetty.connector.bytesReceived", connectionStatistics, ConnectionStatistics::getReceivedBytesRate)
                 .tags(tags)
                 .description("The rate of bytes received")
@@ -54,6 +55,22 @@ public class JettyConnectionMetrics implements MeterBinder {
                 .tags(tags)
                 .description("The rate of bytes sent")
                 .baseUnit("bytes/second")
+                .register(registry);
+
+        Gauge.builder("jetty.connector.duration.max", connectionStatistics, ConnectionStatistics::getConnectionDurationMax)
+                .tags(tags)
+                .description("The maximum connection lifetime duration")
+                .baseUnit("milliseconds")
+                .register(registry);
+        Gauge.builder("jetty.connector.duration.mean", connectionStatistics, ConnectionStatistics::getConnectionDurationMean)
+                .tags(tags)
+                .description("The mean connection lifetime duration")
+                .baseUnit("milliseconds")
+                .register(registry);
+        Gauge.builder("jetty.connector.duration.stddev", connectionStatistics, ConnectionStatistics::getConnectionDurationStdDev)
+                .tags(tags)
+                .description("The standard deviation of the connection lifetime duration")
+                .baseUnit("milliseconds")
                 .register(registry);
     }
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
@@ -70,7 +70,7 @@ public class JettyConnectionMetrics implements MeterBinder {
                 .tags(tags)
                 .description("The maximum number of connections")
                 .register(registry);
-        Gauge.builder("jetty.connector.connections.total", connectionStatistics, ConnectionStatistics::getConnectionsTotal)
+        FunctionCounter.builder("jetty.connector.connections.total", connectionStatistics, ConnectionStatistics::getConnectionsTotal)
                 .tags(tags)
                 .description("The total number of connections")
                 .register(registry);
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
@@ -15,12 +15,14 @@
  */
 package io.micrometer.core.instrument.binder.jetty;
 
-import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.FunctionCounter;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.MeterBinder;
 import org.eclipse.jetty.io.ConnectionStatistics;
 
 import static java.util.Collections.emptyList;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
  * {@link MeterBinder} for Jetty's connection metrics.<br><br>
@@ -82,19 +84,6 @@ public class JettyConnectionMetrics implements MeterBinder {
                 .tags(tags)
                 .description("The rate of bytes sent")
                 .baseUnit("bytes")
-                .register(registry);
-
-        TimeGauge.builder("jetty.connector.duration.max", connectionStatistics, MILLISECONDS, ConnectionStatistics::getConnectionDurationMax)
-                .tags(tags)
-                .description("The maximum connection lifetime duration")
-                .register(registry);
-        TimeGauge.builder("jetty.connector.duration.mean", connectionStatistics, MILLISECONDS, ConnectionStatistics::getConnectionDurationMean)
-                .tags(tags)
-                .description("The mean connection lifetime duration")
-                .register(registry);
-        TimeGauge.builder("jetty.connector.duration.stddev", connectionStatistics, MILLISECONDS, ConnectionStatistics::getConnectionDurationStdDev)
-                .tags(tags)
-                .description("The standard deviation of the connection lifetime duration")
                 .register(registry);
     }
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
@@ -1,0 +1,59 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.jetty;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.MeterBinder;
+import org.eclipse.jetty.io.ConnectionStatistics;
+
+public class JettyConnectionMetrics implements MeterBinder {
+
+    private final ConnectionStatistics connectionStatistics;
+    private final Iterable<Tag> tags;
+
+    public JettyConnectionMetrics(ConnectionStatistics connectionStatistics, Iterable<Tag> tags) {
+        this.connectionStatistics = connectionStatistics;
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        Gauge.builder("jetty.connector.connections.current", connectionStatistics, ConnectionStatistics::getConnections)
+                .tags(tags)
+                .description("The current number of open connections")
+                .register(registry);
+        Gauge.builder("jetty.connector.connections.max", connectionStatistics, ConnectionStatistics::getConnectionsMax)
+                .tags(tags)
+                .description("The maximum number of connections")
+                .register(registry);
+        Gauge.builder("jetty.connector.connections.total", connectionStatistics, ConnectionStatistics::getConnectionsTotal)
+                .tags(tags)
+                .description("The total number of connections")
+                .register(registry);
+        Gauge.builder("jetty.connector.bytesReceived", connectionStatistics, ConnectionStatistics::getReceivedBytesRate)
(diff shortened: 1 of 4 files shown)


tomakehurst:
Hmmmm....these two tests failed when I ran the suite from HEAD, so I thought they might just be due to some quirk of my local setup. I guess not.

Any thoughts on why these numbers have changed?

tomakehurst:
OK, it looks like this is due to a new Tomcat patch version being shipped today (8.5.49). When I pinned it to 8.5.47 (in https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/build.gradle#L34) it worked fine.

Is this particular number meaningful?
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java#L227

Or could it just be set to the new value we're getting after the Tomcat version upgrade?

checketts:
What does that value represent? Is that byte sent? I guess a version upgrade might affect the encoding and thus change the bytes. Seems like we would update to match if that is the case.

tomakehurst:
Those tests trigger Tomcat's default `405 Method Not Allowed` response and it looks like this has shed some headers and body content between these versions, so I think that would explain it.

I don't know whether triggering a 405 is intentional, but if not I guess it might be better to generate a specific servlet response so that it's under the test's control and won't change between Tomcat versions.

tomakehurst:
This is the problematic line: https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java#L180

If the servlet was updated to handle GETs too, this problem would go away I think.

Happy to push another commit with this if that suits.

checketts:
Yes let's add that in this pr as a separate commit. That way we could cherry pick it if needed

tomakehurst:
Here you go. f751a159917f19a025b52a455362533dbe8d4f36 (the Tomcat test fix) should be cherry pickable if necessary.

john-karp:
Actually we were going to submit a similar PR. However, we were planning to include connector.connections.duration.* as time gauges. Would you mind adding those metrics too?

tomakehurst:
@john-karp seems like good idea, but looking at the Jetty code I can't see a way to get at the total duration, as it's encapsulated. Have you found a way to get at this (other than unpacking it via reflection)?

tomakehurst:
@john-karp would you be happy with max, mean and stddev as gauges?

john-karp:
@tomakehurst - We were going to make do with mean/max/stddev gauges, actually. I might look at a better way, but I don't want to block this PR on that.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -0,0 +1,76 @@
...
+        Gauge.builder("jetty.connector.duration.mean", connectionStatistics, ConnectionStatistics::getConnectionDurationMean)
+                .tags(tags)
+                .description("The mean connection lifetime duration")
+                .baseUnit("milliseconds")
+                .register(registry);
+        Gauge.builder("jetty.connector.duration.stddev", connectionStatistics, ConnectionStatistics::getConnectionDurationStdDev)
+                .tags(tags)
+                .description("The standard deviation of the connection lifetime duration")
+                .baseUnit("milliseconds")
+                .register(registry);
```
john-karp: Do we want to use https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/TimeGauge.java#L25 ?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        Gauge.builder("jetty.connector.connections.current", connectionStatistics, ConnectionStatistics::getConnections)
+                .tags(tags)
+                .description("The current number of open connections")
+                .register(registry);
+        Gauge.builder("jetty.connector.connections.max", connectionStatistics, ConnectionStatistics::getConnectionsMax)
+                .tags(tags)
+                .description("The maximum number of connections")
```
shakuzen: Is this the maximum number of open connections for the lifetime of the server?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -0,0 +1,76 @@
...
+                .register(registry);
+        Gauge.builder("jetty.connector.connections.total", connectionStatistics, ConnectionStatistics::getConnectionsTotal)
+                .tags(tags)
+                .description("The total number of connections")
```
shakuzen: Is this the total number of connections opened over the lifetime of the server?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -0,0 +1,76 @@
...
+                .register(registry);
+
+        Gauge.builder("jetty.connector.bytesReceived", connectionStatistics, ConnectionStatistics::getReceivedBytesRate)
```
shakuzen: I'm wondering if using a FunctionCounter with `getReceivedBytes` wouldn't be better. We generally say don't gauge something you can count. A rate could be derived from this.

I'm also concerned about the bolded part of this description for `getReceivedBytesRate` (emphasis mine): "Total number of bytes received per second **since the last invocation of this method**". That sounds like it could lead to unexpected values if the method is invoked more than once in a step interval, which we can't guarantee won't happen, I think.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
shakuzen: Yes, I think a TimeGauge will be better.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
tomakehurst: Yes, that's my understanding

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
tomakehurst: Yes, also my understanding

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -0,0 +1,100 @@
...
+        Gauge.builder("jetty.connector.connections.max", connectionStatistics, ConnectionStatistics::getConnectionsMax)
+                .tags(tags)
+                .description("The maximum number of connections")
+                .register(registry);
+        Gauge.builder("jetty.connector.connections.total", connectionStatistics, ConnectionStatistics::getConnectionsTotal)
+                .tags(tags)
+                .description("The total number of connections")
+                .register(registry);
+
+        FunctionCounter.builder("jetty.connector.bytesReceived", connectionStatistics, ConnectionStatistics::getReceivedBytesRate)
```
shakuzen: I'm leaning towards removing `bytes` from the name. We have some precedent for this e.g. Tomcat metrics `tomcat.global.received` and JVM metrics, and we are setting the `baseUnit` to `bytes` which will cause it to be added to the ID depending on the MeterRegistry used. I can make this change as polish after merging, though. I just wanted to leave a note here so it wasn't a surprise and anyone can give feedback on it.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -0,0 +1,100 @@
...
+
+        TimeGauge.builder("jetty.connector.duration.max", connectionStatistics, MILLISECONDS, ConnectionStatistics::getConnectionDurationMax)
+                .tags(tags)
+                .description("The maximum connection lifetime duration")
+                .register(registry);
+        TimeGauge.builder("jetty.connector.duration.mean", connectionStatistics, MILLISECONDS, ConnectionStatistics::getConnectionDurationMean)
+                .tags(tags)
+                .description("The mean connection lifetime duration")
+                .register(registry);
+        TimeGauge.builder("jetty.connector.duration.stddev", connectionStatistics, MILLISECONDS, ConnectionStatistics::getConnectionDurationStdDev)
```
shakuzen: I'm still hoping to get some feedback on whether including mean and standard deviation for duration is even useful. After we release this in a version, it's hard to remove the metrics because it will break i.e. metrics dashboards for users. Maybe it would be better for us to properly instrument this with Micrometer ourselves, rather than only relying on what is available via `ConnectionStatistics`. Or we could request an enhancement in Jetty for better duration statistics.

tomakehurst: I'd have thought mean would be the most useful of the three available metrics, although I'll admit I'm not wholly convinced there's much utility in any of them. I'm struggling to think what tuning decisions I might make based on connection duration.

Perhaps @john-karp can comment?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
tomakehurst: Seems sensible to me.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
shakuzen: On mean latency, here's a [short blog post](http://latencytipoftheday.blogspot.com/2014/06/latencytipoftheday-average-random.html) with the summary of arguments against using it.

tomakehurst: Oh yeah, I get that mean is a poor way to present latency. I'm not sure connection lifetime can be considered a latency metric though.

shakuzen: Maybe I misunderstand what the connection duration means here. Is it the duration the server connection is open? Isn't that essentially the server response time to requests? Or is the duration across more than one request?

tomakehurst: Connections are usually reused, so in most cases it'll be the lifetime of the connection across many request/responses.

john-karp: Our use case is that we're using HTTP as a transport for batches of data. We want to detect if incoming connections are short-lived, since frequent TLS connection (re)establishment reduces our throughput. I agree the standard deviation and max as computed by Jetty are not that useful.

tomakehurst: In the past I've solved this problem by adding a rate metric derived from total connections. Although it's not quite the same as tracking connection lifetime, when compared with the request rate, it gives a good indication of whether connections are being well utilised.

I guess you could even explicitly calculate a ratio of the two in your monitoring tool.

tomakehurst: OK, how's this for a final set of changes?

* Remove the connection duration metrics completely for now
* Rename the rate metrics to remove "bytes" from the names, and add the unit

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -0,0 +1,100 @@
...
+    public void bindTo(MeterRegistry registry) {
+        Gauge.builder("jetty.connector.connections.current", connectionStatistics, ConnectionStatistics::getConnections)
+                .tags(tags)
+                .description("The current number of open connections")
+                .register(registry);
+        Gauge.builder("jetty.connector.connections.max", connectionStatistics, ConnectionStatistics::getConnectionsMax)
+                .tags(tags)
+                .description("The maximum number of connections")
+                .register(registry);
+        Gauge.builder("jetty.connector.connections.total", connectionStatistics, ConnectionStatistics::getConnectionsTotal)
```
john-karp: Would a FunctionCounter make more sense? It would be more consistent with how we're counting bytes sent/received below.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
john-karp: Ok, that's fine. I have one other comment though.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
tomakehurst: Yeah, fair point, I'll change that too.

tomakehurst:
I don't think any of the Micrometer drivers would do that by default. Count -> rate is a transformation that takes place in the monitoring tool usually in my experience.

john-karp:
I'm referring to "Some monitoring systems, like Prometheus, push cumulative values for counters to the backend, but others publish the rate at which a counter is incrementing over the push interval." https://micrometer.io/docs/concepts#_function_tracking_counters

See https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepFunctionCounter.java#L44 , any registries using that will publish increments for each time step, not a total.

checketts:
@john-karp Good point about the differences with backends, I hadn't realized `FunctionCounters` had that feature.

That being said, with StepMeterRegistries I don't think the name `total` is a confusing, since it is the total during that interval and `total` is used on many metrics and it remains consistent with all those (based on the registry it is a total within an interval or a cumulative total).

Do you have a clearer name in mind?

tomakehurst:
Agree with @checketts - calling it `rate` would be wrong for some backends and `count` would be wrong for the others. `total` seems like a reasonable compromise.

john-karp:
If 'total' doesn't imply anything, with respect to cumulative vs incremental, or anything else, then its presence doesn't really add any meaning to the name. If it is meaningful in some way, then we should answer why it is not also `jetty.connector.sent.total and `jetty.connector.received.total`.

Why not just `jetty.connector.connections.opened` / `"Number of connections opened"` or `jetty.connector.connections.new` / `"Number of new connections formed"`?

tomakehurst:
One thing it does communicate is the relationship with the underlying Jetty metric it's exposing.

But TBH at this point I'll happily call it whatever is likely to get the PR merged quickest.

john-karp:
Well, I don't want to bike-shed this. Maybe @shakuzen or @checketts wants to chime in, but go ahead otherwise.

jkschneider:
It appears that Jetty's `ConnectionStatistics` implements `Connection.Listener`, so I think we can avoid much of the naming confusion here and improve the metrics by implementing our own `Connection.Listener`. Then we get true distribution statistics (e.g. percentiles, decaying maxes, histograms, SLA boundaries), and can name appropriately. For example, a single `Timer` called `jetty.connection` represents the max, mean, count, sum, percentiles, etc. all with one meter.

jkschneider:
Yikes, shivering at some of the stuff going on in the Jetty class, like:
```
long elapsed = System.currentTimeMillis() - connection.getCreatedTimeStamp();
```

john-karp:
@jkschneider - do similar concerns also apply to micrometer's `JettyStatisticsMetrics`? It uses `org.eclipse.jetty.server.handler.StatisticsHandler`, which is similar to `org.eclipse.jetty.io.ConnectionStatistics`.

https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyStatisticsMetrics.java#L32
https://git.eclipse.org/c/jetty/org.eclipse.jetty.project.git/tree/jetty-server/src/main/java/org/eclipse/jetty/server/handler/StatisticsHandler.java#n102

jkschneider:
@john-karp Yes, I've replaced that locally here as well. Somehow async handling doesn't work yet the same way it does in the built-in Jetty handlers even though they look the same to me. That's all I've got left and then can commit.

jkschneider:
Squashed Tom's original work in b219c7f39d1dafe80e95e837adf9802fca04abf1, added more work and merged with 0599c167c1fb4a5442b582ae1cac6f9e6b963463. Thanks for the patience on this!

tomakehurst:
Thanks everyone for giving this your time and attention!

PR #1716, opened 2019-11-24
Support metrics for Apache http components async client pools

worldtiki:
Same as the regular PoolingHttpClientConnectionManagerMetricsBinder but for the async client.


worldtiki:
Tests also failing in master

shakuzen:
Sorry about the test flake. It should be fixed on master now. Could you rebase these changes on the latest master now and push it so the tests get re-run?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingNHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,101 @@
...
+import io.micrometer.core.lang.NonNull;
+import org.apache.http.impl.nio.conn.PoolingNHttpClientConnectionManager;
+
+/**
+ * Collects metrics from a {@link PoolingNHttpClientConnectionManager}.
+ *
+ * It monitors the overall connection pool state.
+ *
+ * @author Daniel Albuquerque
+ * @since 1.3.2
```
shakuzen: Since this is a new feature, it would be added to the next feature release, which is currently 1.4.0.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/httpcomponents/PoolingNHttpClientConnectionManagerMetricsBinderTest.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+class PoolingNHttpClientConnectionManagerMetricsBinderTest {
+
+    private MeterRegistry registry = new SimpleMeterRegistry(SimpleConfig.DEFAULT, new MockClock());
+    private PoolingNHttpClientConnectionManager connectionManager;
+    private PoolingNHttpClientConnectionManagerMetricsBinder binder;
+
+    @BeforeEach
+    void setup() {
+        connectionManager = mock(PoolingNHttpClientConnectionManager.class);
```
shakuzen: Would there be any issues instantiating a connection pool for the test? It seems like that would be better than mocking. Though I suppose we would need a wiremock server to make HTTP requests with the HTTP client, and that would add some complexity.

worldtiki: I would love to add testcontainers and get rid of some of these mocks :/

worldtiki:
> Overall it looks good as it is basically the same as the existing non-async binder. That made me wonder though if we couldn't combine the two. I notice that both of the HttpClientConnectionManager implementations implement `ConnPoolControl`, which defines all of the methods used for the metrics currently. Do you think it'd be too complex to handle both these in the existing MeterBinder?

That's a good point. This stops being http specific though.

I like the idea of having a "PoolingHttpClientConnectionManagerMetricsBinder" because it helps to direct users to what they are looking for. For that reason I agree with your suggestion of merging both sync and async into one, but I would not try to make this more generic than that, ie, I would keep this name, even if in practical terms this same implementation could be used for other cases.

Or maybe we can have dulplicate implementation for all apache pools, but still keep this one as "syntatic sugar" or a convenience class.

PR #1717, opened 2019-11-24
Remove unnecessary boxing

worldtiki:
Tiny tiny change.
There is no point in using Booleans just to immediately unbox them to a boolean.


worldtiki:
The TomcatMetricsTest are also failing in the master branch

PR #1729, opened 2019-12-03
Clean up compile warnings in 1.3.x branch

izeye:
This PR cleans up compile warnings in the 1.3.x branch.

PR #1731, opened 2019-12-03
Add illegal import check for Java6Assertions

izeye:
This PR adds a Checkstyle rule for `org.assertj.core.api.Java6Assertions`.

See https://github.com/micrometer-metrics/micrometer/pull/1729#discussion_r353007060

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -14,6 +14,7 @@
         <!-- Imports -->
         <module name="IllegalImportCheck" >
             <property name="illegalPkgs" value="com.google.common.(?!cache).*,org.apache.commons.text.*,org.slf4j.*"/>
+            <property name="illegalClasses" value="org\.assertj\.core\.api\.Java6Assertions\..*"/>
             <property name="regexp" value="true"/>
         </module>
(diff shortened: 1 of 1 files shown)


config/checkstyle/checkstyle.xml
```diff
@@ -27,6 +27,12 @@
                       value="Line has leading tab characters; indentation should be performed with spaces only." />
             <property name="ignoreComments" value="true" />
         </module>
+        <module name="com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineJavaCheck">
```
shakuzen: Could we use the `IllegalImportCheck`'s `illegalClasses` property instead? I think the regex here will also work, but it just seems more straightforward to use the IllegalImportCheck. I'm not sure of its behavior when importing a static method from one of the `illegalClasses`.
https://checkstyle.org/config_imports.html#IllegalImport

izeye: @shakuzen Thanks for the feedback 👍 I tried it with a wrong value before and just assumed that it doesn't work with static imports. I confirmed that it works and updated in 903716d as you suggested.

Issue #1743, opened 2019-12-10
Memory leak with kafka metrics


gstephant:
Hi there,

I have a springboot app wich purpose is to print the lag on all kafka topics/consumer groups. And that's all.

It runs out of memory, and the memory usage is as follows :

![image](https://user-images.githubusercontent.com/22791611/70515272-5ca8bb80-1b35-11ea-9537-c33bedd82519.png)

Code : 

```java
@Service
@ManagedResource
public class KafkaLagWriter {

    private static final Logger LOGGER = LoggerFactory.getLogger(KafkaLagWriter.class);

    private final KafkaAdmin kafkaAdmin;
    private final String bootstrapServers;

    public KafkaLagWriter(final KafkaAdmin kafkaAdmin,
                          @Value("${spring.kafka.bootstrap-servers}") final String bootstrapServers) {
        this.kafkaAdmin = kafkaAdmin;
        this.bootstrapServers = bootstrapServers;
    }

    @ManagedOperation
    @Scheduled(fixedDelay = 60_000)
    public void logKafkaTopicLags() {

        try (final AdminClient adminClient = AdminClient.create(kafkaAdmin.getConfig())) {
            final ListConsumerGroupsResult consumerGroupsResult = adminClient.listConsumerGroups();
            final Collection<ConsumerGroupListing> consumerGroupListing = consumerGroupsResult.all().get();
            for (final ConsumerGroupListing consumerGroup : consumerGroupListing) {
                final String consumerGroupId = consumerGroup.groupId();

                final ListConsumerGroupOffsetsResult consumerGroupOffsetsResult = adminClient.listConsumerGroupOffsets(consumerGroup.groupId());
                final KafkaFuture<Map<TopicPartition, OffsetAndMetadata>> futureMap = consumerGroupOffsetsResult.partitionsToOffsetAndMetadata();
                final Map<TopicPartition, OffsetAndMetadata> offSetByPartition = futureMap.get();
                final Map<String, Map<TopicPartition, OffsetAndMetadata>> topicsToOffSetByPartition = new HashMap<>();

                offSetByPartition.forEach((key, value) -> {
                    Map<TopicPartition, OffsetAndMetadata> offSets = topicsToOffSetByPartition.get(key.topic());
                    if (offSets != null) {
                        offSets.put(key, value);
                    } else {
                        offSets = new HashMap<>();
                        offSets.put(key, value);
                    }
[...]

PR #1745, opened 2019-12-10
Remove Kafka meters on unregistration notification

izeye:
This PR removes Kafka meters on unregistration notification.

Closes gh-1743


shakuzen:
Thank you for the quick turnaround on a fix for this. I'd like to make sure we have a test for this. Would you be willing to add one? Otherwise, I can work on that.

izeye:
@shakuzen Thanks for the feedback! I added tests and handled a missed case thanks to the tests. I also changed the base branch to 1.1.x as it was master accidentally.

shakuzen:
One thing that came to mind is that if Kafka consumers are created and closed before a full step interval, the metrics from it may never be published. I'm not sure that's even an issue we should concern ourselves with, but I don't think we have any global metrics that would show this behavior happening currently. This is one of the things we can hope to fix with the Kafka metrics overhaul. In the meantime, we want this fix so we're not causing memory issues.

PR #1747, opened 2019-12-12
Clean up compile warnings

izeye:
This PR cleans up compile warnings.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsConnectionPoolListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsConnectionPoolListener.java
@@ -91,12 +91,14 @@ public class MongoMetricsConnectionPoolListener extends ConnectionPoolListenerAd
                 .decrementAndGet();
     }
 
+    @SuppressWarnings("deprecation")
     @Override
     public void waitQueueEntered(ConnectionPoolWaitQueueEnteredEvent event) {
         waitQueueSize.get(event.getServerId())
                 .incrementAndGet();
     }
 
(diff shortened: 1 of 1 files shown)


izeye:
@shakuzen Thanks for the feedback! I read the same Javadoc, assumed that there was no replacement, and just thought that we could remove metrics when the methods are removed.

shakuzen:
I'm not sure the timeframe of the Mongo client release that will remove the methods, but I'd like to be as proactive as we can. Suppressing the deprecation warning will lead to a broken build at some point in the future. I'd almost say we might as well just remove these metrics in 1.4 if the Mongo team is signaling that they aren't important by not providing (or at least not mentioning in the deprecation) an alternate way to get them. Of course, I'm willing to be educated otherwise by Mongo users.

izeye:
I found [the related issue](https://jira.mongodb.org/browse/JAVA-3460) and [the commit](https://github.com/mongodb/mongo-java-driver/commit/d3ea5fa73d480eb05551a841dbc5985c44fa2892), but it's not clear to me why they removed the events for removing wait queue size limit.

shakuzen:
Thank you for the references. That has helped me understand the motivation and direction. [Here is a section](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#why-are-waitqueuesize-and-waitqueuemultiple-deprecated) in the referenced specification with some more explanation. Given what I've read, I think your original plan of action makes the most sense after all. While there is a wait queue mechanism in the Mongo driver, we should provide metrics for it. The problem will be if we need to try to support the current major version and the next one, but we'll have to tackle that later.

Thank you for bearing with me. I wanted to make sure I understood what we were doing, why, and what to expect coming up.

Issue #1746, opened 2019-12-11
Signalfx tag name sanitization is incomplete


kjurasovic-dexcom:
Recently we started using SignalFx for our application monitoring.
We integrated to their endpoint using spring Boot 2.2 and micrometer SignalFx integration. 

Integration works well except in one case.

In our application we have Kafka consumers. 
Micrometer and Spring Boot expose all metrics to actuator endpoint and SignalFx. 
The problem is SignalFx rejects all Kafka related metrics.
After few days of research with SignalFx we determined that metrics are rejected because they contain tag with with a dot (.). Specifically tag "client.id".

According to SignalFx documentation tag name must be:
1) UTF-8 string, maximum length 128 characters (512 bytes)
2) Must start with an uppercase or lowercase letter. The rest of the name can contain letters, numbers, underscores (_) and hyphens (-)
3) Must not start with the underscore character (_)
4) Must not start with the prefix sf_, except for dimensions defined by SignalFx such as sf_hires
5) Must not start with any of these prefixes: aws_, gcp_, or azure_.

I've checked class SignalFxNamingConvention and it seems micrometer only enforces rules 1, part 1 of 2, 3, 4 but not 5 and "the rest of the name" rules.

At the moment we created our custom filter that sanitizes on missing rules. 
Would it be possible to add missing rules to SignalFx module?

SignalFx naming convention can be found here [link](https://developers.signalfx.com/metrics/data_ingest_overview.html#_criteria_for_metric_and_dimension_names_and_values).

We are currently using version "io.micrometer:micrometer-registry-signalfx:1.1.4".
I've check the master branch and it has the same issue.

izeye:
@kjurasovic-dexcom Thanks for the report! I created #1748 to resolve this.

shakuzen:
Thanks for the report @kjurasovic-dexcom and sorry for the inconvenience. We'll get this fixed for the next patch release. Once we have a fixed merged, if you could test the snapshots without your current workaround, we'd appreciate the confirmation of the fix.

By the way, Spring Boot 2.2 should manage the version of all Micrometer dependencies for you. Is there any reason you're explicitly using version 1.1.4 instead of the version managed by Spring Boot?

kjurasovic-dexcom:
Thanks for the fix @shakuzen, @izeye. That was quick.
Sure we can test the fix when its ready.

Regarding version 1.1.4. We know about spring dependency management. This was some fix from before what was not removed yet. Will be in some future release.

PR #1748, opened 2019-12-12
Apply missing sanitization for SignalFx tag keys

izeye:
This PR applies missing sanitization for SignalFx tag keys.

Closes gh-1746

### DIFF
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
@@ -22,6 +22,7 @@ import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.util.StringEscapeUtils;
 import io.micrometer.core.instrument.util.StringUtils;
 import io.micrometer.core.lang.Nullable;
+import io.micrometer.core.util.internal.logging.WarnThenDebugLogger;
 
 /**
  * {@link NamingConvention} for SignalFx.
@@ -32,6 +33,8 @@ import io.micrometer.core.lang.Nullable;
  * @author Johnny Lim
  */
 public class SignalFxNamingConvention implements NamingConvention {
+
+    private static final WarnThenDebugLogger logger = new WarnThenDebugLogger(SignalFxNamingConvention.class);
 
     private static final Pattern START_UNDERSCORE_PATTERN = Pattern.compile("^_");
     private static final Pattern SF_PATTERN = Pattern.compile("^sf_");
@@ -72,9 +75,14 @@ public class SignalFxNamingConvention implements NamingConvention {
         conventionKey = SF_PATTERN.matcher(conventionKey).replaceAll(""); // 2
 
         conventionKey = PATTERN_TAG_KEY_BLACKLISTED_CHARS.matcher(conventionKey).replaceAll("_");
-        if (PATTERN_TAG_KEY_BLACKLISTED_PREFIX.matcher(conventionKey).matches()
-                || !START_LETTERS_PATTERN.matcher(conventionKey).matches()) { // 3
+        if (!START_LETTERS_PATTERN.matcher(conventionKey).matches()) { // 3
             conventionKey = "a" + conventionKey;
+        }
+        if (PATTERN_TAG_KEY_BLACKLISTED_PREFIX.matcher(conventionKey).matches()) {
+            logger.log("'" + conventionKey + "' (original name: '" + key + "') is not a valid meter name. "
+                    + "Must not start with any of these prefixes: aws_, gcp_, or azure_. "
+                    + "Please rename it to conform to the constraints. "
+                    + "If it comes from a third party, please use MeterFilter to rename it.");
         }
         return StringUtils.truncate(conventionKey, KEY_MAX_LENGTH); // 1
     }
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxNamingConventionTest.java
```diff
@@ -30,4 +36,16 @@ void tagKey() {
...
+    @Test
+    void tagKeyWhenKeyHasBlacklistedCharShouldSanitize() {
+        assertThat(convention.tagKey("a.b")).isEqualTo("a_b");
+    }
+
+    @Test
+    void tagKeyWhenKeyStartsWithBlacklistedPrefixShouldSanitize() {
+        assertThat(convention.tagKey("aws_a")).isEqualTo("aaws_a");
+        assertThat(convention.tagKey("gcp_a")).isEqualTo("agcp_a");
+        assertThat(convention.tagKey("azure_a")).isEqualTo("aazure_a");
```
shakuzen: This isn't ideal. I feel like it'd be better if we logged a warning the first time to let users know what metric is registered that violates the rules of the metrics system, and then just don't send the metric at all. I don't think we have a proper way of doing that last part now, though. Maybe in the future we need a distinction between converting for convention and violations with no obvious conversion. It's subjective, though.
Anyways, for now, what does everyone think of adding a once logger to warn of the metric name and otherwise keep things as they are?

izeye: @shakuzen Thanks for the feedback! I just chose this way as leading underscores are already replaced with "a". I changed it to use the same approach as `DynatraceNamingConvention` in 813505a. Do you think that leading underscores need to be treated in the same way?

shakuzen: Leading underscores and numbers somehow felt more acceptable to me the way they are now, though the argument could be made for them too. The next question is, if we're not changing these now and attempt to send them as is, does the whole payload fail or are only these metrics not stored by signalfx? We don't want the whole payload to fail because some don't follow the rules.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
```diff
@@ -66,9 +74,16 @@ public String tagKey(String key) {
         conventionKey = START_UNDERSCORE_PATTERN.matcher(conventionKey).replaceAll(""); // 2
         conventionKey = SF_PATTERN.matcher(conventionKey).replaceAll(""); // 2
 
+        conventionKey = PATTERN_TAG_KEY_BLACKLISTED_CHARS.matcher(conventionKey).replaceAll("_");
         if (!START_LETTERS_PATTERN.matcher(conventionKey).matches()) { // 3
             conventionKey = "a" + conventionKey;
         }
+        if (PATTERN_TAG_KEY_BLACKLISTED_PREFIX.matcher(conventionKey).matches()) {
+            logger.log("'" + conventionKey + "' (original name: '" + key + "') is not a valid meter name. "
```
shakuzen: meter → tag

implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxNamingConventionTest.java
shakuzen: If the whole payload fails, it'd be better to restore the behavior of appending the `a` character - at least it is consistent with the other cases you mentioned, and at least we're logging something now so it can be noticed and fixed.

izeye: > The next question is, if we're not changing these now and attempt to send them as is, does the whole payload fail or are only these metrics not stored by signalfx?

@shakuzen I'm not a SignalFx user, so I'm not sure. I just read the following description from @kjurasovic-dexcom in #1746 as the other metrics haven't been rejected:

```
The problem is SignalFx rejects all Kafka related metrics.
```

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
izeye: @shakuzen Thanks! I fixed it in 034f52c.

implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxNamingConventionTest.java
kjurasovic-dexcom: @izeye from the tests we did it seems like SignalFx only rejects metrics that don't follow the rules. Other JVM metrics were displayed correctly in SignalFx. I could check with them directly if needed?

izeye: @kjurasovic-dexcom Thanks for the feedback! If you could confirm it, it would be much appreciated.

kjurasovic-dexcom: @izeye This is the info I got from SignalFx team "we’ll just drop those metrics that have this invalid tag name.  Others should be ok.". This is the behavior we observed in our tests.

izeye: @kjurasovic-dexcom Thanks for the confirmation!

@shakuzen Based on the info, is it okay with as-is?

shakuzen: Looks good to me. Thank you both for the follow-up on this.

PR #1749, opened 2019-12-12
native-image.properties that adds the shaded artifact pcollections to initialize-at-build-time.

jkremser:
GraalVM's [native-image](https://www.graalvm.org/docs/reference-manual/native-image/) tool can create executable self-contained applications out of jar files by doing a lot of magic, like evaluating static initializers during the compilation, making a snapshot of the memory heap, etc. However sometimes some classes' initialization need to be deferred to run-time to pass the build. This PR adds the property file that is read by the native-image tool during the compilation and the arguments are appended to the build tool. It's harmless for the jar itself and will be ignored by all the other tools.

I know you are not owner of the `pcollections` artifact, however you are using the shading and changing its group id, hence the PR. I will be opening also the pr against the upstream repo soon.

### DIFF
--- a/micrometer-core/src/main/resources/META-INF/native-image/io.micrometer.shaded.org.pcollections/pcollections/native-image.properties
+++ b/micrometer-core/src/main/resources/META-INF/native-image/io.micrometer.shaded.org.pcollections/pcollections/native-image.properties
@@ -1,0 +1,1 @@
+Args = --initialize-at-build-time=io.micrometer.shaded.org.pcollections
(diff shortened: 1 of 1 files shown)


shakuzen:
Thank you for the pull request. I'm wondering if this will still be required or not after the upstream PR is merged. Since we are shading and the package is different, I'm guessing it may, but I just wanted to confirm.

shakuzen:
> I'm wondering if this will still be required or not after the upstream PR is merged. Since we are shading and the package is different, I'm guessing it may, but I just wanted to confirm.

Still curious about this. I see the upstream change has been merged and released. We will be upgraded to that version in the same release this change would go in Micrometer. So if it isn't needed at all, we can just close this.

jkremser:
Unfortunately, it is. The `--initialize-at-build-time=foo.bar.package` is prefix based and in the upstream `pcollections` project it adds this: `--initialize-at-build-time=org.pcollections` (this will add anything that starts w/ `org.pcollections`) however this wouldn't match the shaded version in your project, because here it starts w/ different a prefix

jkremser:
Thank you 🎁 🎄

shakuzen:
As a note, we ended up removing our dependency on pcollections, which obviated this change. See https://github.com/micrometer-metrics/micrometer/issues/1791

PR #1750, opened 2019-12-13
Truncate tag values > 1024 for Stackdriver

ramondebeijer:
Spit up from original PR-1724

This PR is about truncating the tag value of Stackdriver metric when >1024 chars according the documenation here:  https://cloud.google.com/monitoring/quotas


shakuzen:
Since this is a bug, we'll want to fix it in maintenance releases also. I'll rebase the pull request on the `1.1.x` branch so that we can include it in the next 1.1.x release and then merge it forward from there.

PR #1752, opened 2019-12-14
Make statsd more configurable

beedub:
We run statsd with some slight customizations in building the line and appending the tags, so these changes would be very helpful.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -466,6 +466,11 @@ public class StatsdMeterRegistry extends MeterRegistry {
             return this;
         }
 
+        public Builder namingConvention(NamingConvention namingConvention) {
+            this.namingConvention = namingConvention;
+            return this;
+        }
+
         public Builder lineSink(Consumer<String> lineSink) {
             this.lineSink = lineSink;
             return this;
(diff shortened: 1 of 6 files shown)


beedub:
I should have been more descriptive in the PR.

Our overall goal is to prepend the `host` as the first term in the line. For example, if the ID's name were `http.requests`, then we'd like the name to actually be `container-2.http.requests`. This is because wavefront (or at least the way in which our wavefront is setup) interprets the third term (because statsd prepends its own `statsd.<type>` to the line) as the host.

Is there a way to achieve this today without any my proposed changes? Or would there be a better improvement than the ones I've suggested? `lineBuilderFunction` currently only passes the `id`, which would be insufficient as my custom `lineBuilder` inherits from `EtsyStatsdLineBuilder` and needs not just the `id`, but `config` and `nameMapper`. I could update `lineBuilderFunction` to pass `id`, config`, and `nameMapper`, but that would be a breaking change (I think). 

(Apologies if I missed something obvious as I'm just getting up to speed with Java)

shakuzen:
I'm very sorry for not getting back to you for so long. We are trying to do better about taking action on pull requests and making the status of them clear.

> Is there a way to achieve this today without any my proposed changes?

You could register a `MeterFilter` on your `StatsdMeterRegistry` that prepends the host to meter names. See [this part](https://micrometer.io/docs/concepts#_transforming_metrics) of the documentation.

If you are using Wavefront, though, we have a WavefrontRegistry that can publish metrics directly to Wavefront or to a Wavefront proxy, without the need for using statsd. See, for example, [this documentation](https://micrometer.io/docs/registry/wavefront).

I'll close this pull request for now, since I think the above alternatives should be sufficient for the mentioned use case. If anyone disagrees or otherwise has a need for this kind of thing, please let us know and we can reconsider.

PR #1757, opened 2019-12-17
Support extra tags with @Counted annotation.

Insutanto:
Support extra tags with @Counted annotation.


jkschneider:
LGTM, thanks @lnsutanto.

Issue #1492, opened 2019-07-04
Autoclosable issue on multiple Metrics implementation


nickstolwijk:
Some Metrics extend AutoCloseable which causes code quality tools to flag those metrics as not being closed.

`
@Configuration
public class MetricsConfiguration {
    private final MeterRegistry registry;

    public MetricsConfiguration(MeterRegistry registry) {
        this.registry = registry;
    }

    @PostConstruct
    public void enableMetrics() {
        // do not change to try-with-resources as the resource will be removed on closing
        new JvmGcMetrics().bindTo(registry);
        new LogbackMetrics().bindTo(registry);
    }
}`

Is there a good reason to make these AutoCloseable? If not, could those be removed? Maybe replace it with a default method on MeterBinder for close?

shakuzen:
Thank you for letting us know about the usability problem here. Indeed those `MeterBinder`s should not be closed after binding, but they generally should be closed on application shutdown, to clean up side effects of the binder. In environments that manage lifecycles (e.g. Spring beans), these being `AutoCloseable` can be helpful as the environment will automatically `close` them on application shutdown.

nickstolwijk:
Would then Closeable be a better solution?

https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html

With regards,

Nick

shakuzen:
I think `Closeable` would not match in these cases based on its JavaDoc:

> A Closeable is a source or destination of data that can be closed.

I'm happy to add to a JavaDoc on these classes or methods to help explain why and when they should be closed. What do you think?

codemedian:
I've just run into the same issue and found it quite confusing. At the very least [this documentation page](https://micrometer.io/docs/ref/jvm) should contain some information about it and recommendations on how to close it.

shakuzen:
@codemedian thank you for the suggestion. I've opened https://github.com/micrometer-metrics/micrometer-docs/issues/117

codemedian:
I've just upgraded my micrometer version and I think the way this is done is a bit odd/inconsistent. 

I understand that you want to close the resources and all, but there's no clean way to hook into micrometer. I.e. the registry doesn't have a shutdown hook and/or cannot be shutdown. This effectively means that in most cases people will do something along those lines

```java
JvmGcMetrics jvmGcMetrics = new JvmGcMetrics();
jvmGcMetrics.bindTo(prometheusRegistry);

[...]

Issue #757, opened 2018-07-25
Autoclosable issue


ankurbhakta:
All meters extend AutoCloseable which  causes code quality  tools to flag all meters as not being closed.
Looks like only timers really need to be AutoCloseable, so can we move the AutoCloseable down to AbstractTimer class only ?

jkschneider:
Interesting feedback, thanks. Does the code quality tool you are referring to look for a `void close()` method or an implementation of `AutoCloseable`? If it is the latter, we can keep the signature and drop the interface extension which doesn't provide any value anyway.

I think we should leave the signature at least on `Meter` though, because while there may not be `close` implementations across the board for different meter types currently, there could be in the future.

ankurbhakta:
Sonar looks at two things, the close statement is called or an AutoClosable resource should be in a try-with-resources statement. This prevents us from reusing the same counter with try-with-resources in other parts of the service. As you see the example below it really doesn't make sense for meters.
eg.

try ( Counter c =Metrics.counter("rest.failed.counter")){
.....
c.increment();
    }

PR #1760, opened 2019-12-18
Clean up notification listeners in TomcatMetrics

### LINKED ISSUE #1424: ClassLoader leak in TomcatMetrics
When a web application is deployed multiple times, the old classloader will not be garbage collected. This is similar to #827 

I debugged with eclipse and see the following behaviour:
registerMetricsEventually will be called 2x (for "type", "StringCache" and "j2eeType", "Servlet") but on redeploy, the listener, which removes itself with `mBeanServer.removeNotificationListener` is only called for the "j2eeType", "Servlet" filter. 
You see also in the debug screenshot, that there are more than one "type", "StringCache" refenece in the listenerList.

Tomcat version is 9.0.20 (tested on windows)
 
Possible fix: Let TomcatMetrics implement DisposableBean and remove all listeners on destroy.

(Is there a simple workaround to disable the TomcatMetrics?)

jprofiler
grafik

eclipse debug
grafik

izeye:
This PR changes `TomcatMetrics` to implement `AutoCloseable` to clean up notification listeners.

Closes #1424

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
@@ -29,6 +29,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 
@@ -41,7 +42,7 @@ import java.util.function.BiConsumer;
  */
 @NonNullApi
 @NonNullFields
-public class TomcatMetrics implements MeterBinder {
+public class TomcatMetrics implements MeterBinder, AutoCloseable {
 
     private static final String JMX_DOMAIN_EMBEDDED = "Tomcat";
     private static final String JMX_DOMAIN_STANDALONE = "Catalina";
@@ -54,6 +55,8 @@ public class TomcatMetrics implements MeterBinder {
 
     private final MBeanServer mBeanServer;
     private final Iterable<Tag> tags;
+
+    private final Set<NotificationListener> notificationListeners = ConcurrentHashMap.newKeySet();
 
     private volatile String jmxDomain;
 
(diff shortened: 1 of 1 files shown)


jkschneider:
@izeye All the `MeterBinder` implementations used to implement `AutoCloseable`, but we had some feedback a while back that it makes for a bad IDE experience, because the block of code that creates the binder isn't usually what closes it, and Intellisense expects a try-with-resources block or explicit close call.

izeye:
@jkschneider Thanks for the feedback! The reason why I thought that implementing `AutoCloseable` was reasonable is that the clean-up method needs to be called explicitly. If a static code analysis tool warns on this, it sounds like a false positive to me.

shakuzen:
Perhaps having a close method without implementing `AutoCloseable` is a reasonable compromise. We can explain in the JavaDoc for the class when a user of the class wants to ensure the close method is called.

shakuzen:
Providing a clean up method is a good idea, but I wonder if we are also registering listeners for things that will never be notified. If we can eliminate that, it would help with memory usage, even in the case where there aren't multiple contexts. We can tackle that in a separate change, though.

izeye:
> Perhaps having a close method without implementing `AutoCloseable` is a reasonable compromise.

We already have four `MeterBinder`s implementing `AutoCloseable` (`JvmGcMetrics`, `KafkaConsumerMetrics`, `Log4j2Metrics`, and `LogbackMetrics`). Do we want to apply the same rule for them?

izeye:
> Providing a clean up method is a good idea, but I wonder if we are also registering listeners for things that will never be notified.

@shakuzen I thought about it, but I didn't do it intentionally as it is a breaking change. Is it okay to remove it on master?

shakuzen:
For reference, here are the previous issues that have been raised for IDE/Sonar warnings on things that are `AutoCloseable`: #1492 and #757.

> We already have four `MeterBinder`s implementing `AutoCloseable` (`JvmGcMetrics`, `KafkaConsumerMetrics`, `Log4j2Metrics`, and `LogbackMetrics`). Do we want to apply the same rule for them?

I forgot about those. It is probably better to be consistent. I'm leaning towards making `TomcatMetrics` `AutoCloseable`, but I'm open to other proposals also.

PR #1764, opened 2019-12-20
Add JOOQ SQL metrics

jkschneider:
Introducing a mechanism to time [JOOQ](https://www.jooq.org/doc/3.12/manual-single-page/) SQL queries, where queries are provided a human-readable name that is used as the `name` tag on the `jooq.query` timer. In this way, you can dimensionally drill-down on the worst-performing queries or set alerts on particular named queries.

The class `JooqMetrics` needs to be added as an `ExecuteListener` to JOOQ's configuration like so:

```
Configuration configuration = new DefaultConfiguration()
  .set(conn)
  .set(SQLDialect.H2)
  .set(new JooqMetrics(meterRegistry));

DSLContext jooq = DSL.using(configuration);
``` 

At this point, there are two methods for ensuring a JOOQ query is timed:

1. `JooqMetrics.timing(jooq, "selectAllAuthors").fetch("SELECT non_existent_field FROM author")`
2. `timing(jooq.select(asterisk()).from("author"), "selectAllAuthors").fetch()`

No global state is maintained in `JooqMetrics`, but rather the tag value and timer samples are added to derived configuration data in each JOOQ `Query`/`ExecutionContext`.

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -57,6 +57,8 @@ dependencies {
 
     compile 'org.mongodb:mongo-java-driver:latest.release', optional
 
+    compile 'org.jooq:jooq-meta:latest.release', optional
+
     testCompile 'io.projectreactor:reactor-test:latest.release'
 
     // JUnit 5
@@ -76,6 +78,8 @@ dependencies {
     testCompile 'org.springframework:spring-context:4.+'
     testCompile 'org.springframework:spring-test:4.+'
     testCompile 'com.google.inject:guice:4.1.0'
+
+    testCompile 'com.h2database:h2:latest.release'
 
     // Uncomment these if you are interested in testing injection with dagger in MeterRegistryInjectionTest
 //    testCompile 'com.google.dagger:dagger:2.11'
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/JooqExecuteListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/JooqExecuteListener.java
@@ -1,0 +1,108 @@
+/**
+ * Copyright 2020 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.db;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.util.StringUtils;
+import org.jooq.ExecuteContext;
+import org.jooq.exception.DataAccessException;
+import org.jooq.impl.DefaultExecuteListener;
+import org.pcollections.HashTreePMap;
+import org.pcollections.PMap;
+
+import java.util.function.Supplier;
+
+class JooqExecuteListener extends DefaultExecuteListener {
+    private final MeterRegistry registry;
+    private final Iterable<Tag> tags;
+    private final Supplier<Iterable<Tag>> queryTagsSupplier;
+
+    private final Object sampleLock = new Object();
+    private volatile PMap<ExecuteContext, Timer.Sample> sampleByExecuteContext = HashTreePMap.empty();
+
+    public JooqExecuteListener(MeterRegistry registry, Iterable<Tag> tags, Supplier<Iterable<Tag>> queryTags) {
+        this.registry = registry;
+        this.tags = tags;
+        this.queryTagsSupplier = queryTags;
+    }
+
+    @Override
+    public void start(ExecuteContext ctx) {
+        startTimer(ctx);
+    }
+
+    @Override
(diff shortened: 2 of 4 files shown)


jkschneider:
@lukaseder WDYT?

micrometer-core/build.gradle
```diff
@@ -57,6 +57,8 @@ dependencies {
 
     compile 'org.mongodb:mongo-java-driver:latest.release', optional
 
+    compile 'org.jooq:jooq-meta:latest.release', optional
```
lukaseder: I think you wouldn't need `jooq-meta` as a dependency. It's the auxiliary library that provides schema meta information to the code generator. It has a transitive dependency on `org.jooq:jooq`, which is the one you probably want to depend on, directly.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/JooqMetrics.java
```diff
@@ -0,0 +1,116 @@
...
+  public JooqMetrics(MeterRegistry registry, Iterable<Tag> tags) {
+    this.registry = registry;
+    this.tags = tags;
+  }
+
+  @Override
+  public void start(ExecuteContext ctx) {
+    if (ctx.configuration().data(CONFIGURATION_KEY) != null) {
+      ctx.query().attach(ctx.configuration().derive());
+      ctx.query().configuration().data(TIMER_SAMPLE_KEY, Timer.start(registry));
```
lukaseder: Instead of using the `Query.configuration().data()` for your context information, you could just use the `ExecuteContext.data()` map, instead. Benefits:

- It is guaranteed to be local to your single query execution
- You don't modify a query that might be produced from a cache in client code

lukaseder:
> 1\. We could add Micrometer instrumentation directly to JOOQ, as we've done in Reactor, HikariCP, Rabbit, etc. Don't know what the release cadence there is like.

I like your approach. I wonder, if there's additional potential for matching generated SQL text using a regex in the `ExecuteListener`? That would work more globally without the need to modify client code to monitor individual queries...

> 2\. Instead of using derived `Configuration`, should we add JOOQ API to contextually label/tag individual queries for use by `ExecuteListener` implementations? Under the covers, may look the same, but it would be nice to be able to inline the tagging statement like
>     `jooq.select(asterisk()).from("author").tagged(JooqMetrics.TAG, "selectAllAuthors").fetch()` instead of having to wrap the statement like
>     `timing(jooq.select(asterisk()).from("author"), "selectAllAuthors").fetch()`

That would be a very easy thing to do with Kotlin or any other language that has extension methods, but in Java / jOOQ, you'd be duplicating a ton of API, which we might occasionally break between minor releases. I mean, in order to add that tagging method at the "end" of the DSL, you'd have to duplicate the entire DSL for every jOOQ version. I don't think that would scale for you or for your users.

Another option would be to tagg queries at the beginning, directly on `DSLContext`, which you could extend or proxy. That would lead to

```java
jooq
  .tagged(TAG, "selectAllAuthors")
  .select(asterisk())
  .from("author")
  .fetch();
```

Perhaps a bit more convenient? You'd have to proxy only a single type, and we won't change that type's design any time soon.

However, do note that while copying a configuration is not too expensive, copying it on a per-query-execution basis means that all of its internal caches (specifically the reflection cache) will be empty again. This might not be an issue in ordinary client code, but given that you're a metrics tool, you probably don't want to add this much overhead to the your own measurements. I'm saying this specifically because you're timing the `ExecuteListener.start()` and `ExecuteListener.end()` events, which include the time spent in the `RecordMapper`, when perhaps the user is mostly interested in the timing between `executeStart()` and `executeEnd()`.

jkschneider:
> That would be a very easy thing to do with Kotlin or any other language that has extension methods, but in Java / jOOQ, you'd be duplicating a ton of API, which we might occasionally break between minor releases. I mean, in order to add that tagging method at the "end" of the DSL, you'd have to duplicate the entire DSL for every jOOQ version. I don't think that would scale for you or for your users.

Sorry, I realize my original comment was ambiguous. I meant adding API to JOOQ, not duplicating the API in Micrometer to add something like `tagged`.

lukaseder:
> Sorry, I realize my original comment was ambiguous. I meant adding API to JOOQ, not duplicating the API in Micrometer to add something like `tagged`.

That's how I understood you, but that's a significant change, and before engaging in the details of it (e.g. deriving a configuration has its own costs, because e.g. reflection caches will not be inherited, which creates much overhead, just to name one issue), I wanted to offer you a workaround that you can quickly implement on your side.

jkschneider:
> I wanted to offer you a workaround that you can quickly implement on your side.

Ok, perfect. Do you think deriving the config once on the top-level `DSLContext` context and using thread locals inside to maintain tags per query would be better?

jkschneider:
Incorporating @lukaseder's suggestions, the new API is:

```
Configuration configuration = new DefaultConfiguration()
        .set(conn)
        .set(SQLDialect.H2);

MetricsDSLContext jooq = withMetrics(DSL.using(configuration), meterRegistry, Tags.empty());

jooq.tag("name", "selectAllAuthors").execute("SELECT * FROM author");
jooq.tag("name", "selectAllAuthors").fetch("SELECT * FROM author");
jooq.tag("name", "selectAllAuthors").select(asterisk()).from("author").fetch();
jooq.tag("name", "updateAuthor").update(table("author")).set(field("author.first_name"), "jon").execute();
jooq.tag("name", "insertAuthor").insertInto(table("author")).values("2", "jon", "schneider").execute();
```

micrometer-core/build.gradle
shakuzen: Changed in f2814031353d0f96e1daf1a3ca204e89f540f6c2. Thanks

lukaseder:
That looks great!

We'll think about this on our side to see how jOOQ API could help here: https://github.com/jOOQ/jOOQ/issues/9709. In any case, having your own API will allow you to keep in control of what it means to store a "tag" in a jOOQ `Configuration.data()` map, so even if we implement #9709, having your own API is not a bad idea.

PR #1767, opened 2019-12-27
Fix running condition for ProcessorMetricsTest.hotspotCpuMetrics()

izeye:
OpenJ9 has `com.sun.management.OperatingSystemMXBean`, and `com.ibm.lang.management.OperatingSystemMXBean` actually extends `com.sun.management.OperatingSystemMXBean` as follows:

```
public interface OperatingSystemMXBean extends com.sun.management.OperatingSystemMXBean {
```

I used the following version of OpenJ9:

```
% java -version
openjdk version "1.8.0_232"
OpenJDK Runtime Environment (build 1.8.0_232-b09)
Eclipse OpenJ9 VM (build openj9-0.17.0, JRE 1.8.0 Mac OS X amd64-64-Bit Compressed References 20191101_372 (JIT enabled, AOT enabled)
OpenJ9   - 77c1cf708
OMR      - 20db4fbc1
JCL      - 3504dff40aa based on jdk8u232-b09)
%
```

I'm not sure if it has been changed at some point, but the lookup order needs to be reversed to work with the latest version of OpenJ9.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
@@ -41,6 +41,11 @@ import static java.util.Objects.requireNonNull;
  *     <li>HotSpot</li>
  *     <li>J9</li>
  * </ul>
+ *
+ * @author Jon Schneider
+ * @author Michael Weirauch
+ * @author Clint Checketts
+ * @author Tommy Ludwig
  */
 @NonNullApi
 @NonNullFields
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java
@@ -22,6 +22,15 @@ import org.junit.jupiter.api.Test;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
+/**
+ * Tests for {@link ProcessorMetrics}.
+ *
+ * @author Jon Schneider
+ * @author Michael Weirauch
+ * @author Clint Checketts
+ * @author Tommy Ludwig
+ * @author Johnny Lim
+ */
 class ProcessorMetricsTest {
     @Test
     void cpuMetrics() {
@@ -39,7 +48,7 @@ class ProcessorMetricsTest {
 
     @Test
     void hotspotCpuMetrics() {
-        assumeTrue(classExists("com.sun.management.OperatingSystemMXBean"));
+        assumeTrue(!isOpenJ9());
 
         MeterRegistry registry = new SimpleMeterRegistry();
         new ProcessorMetrics().bindTo(registry);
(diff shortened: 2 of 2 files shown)


shakuzen:
> the lookup order needs to be reversed to work with the latest version of OpenJ9.

Is that true? We aren't using any methods that are specific to `com.ibm.lang.management.OperatingSystemMXBean`. It should still be working now. We can cast to and access the currently used methods from any instance of `com.sun.management.OperatingSystemMXBean`.

When I ran a sample Spring Boot application on the same version of OpenJ9 as you, I was able to see the process and system CPU usage metrics from the metrics endpoint.

I vaguely remember there being some situation when the sun classes weren't available and only the IBM classes were. Perhaps this was with the IBM JVM rather than OpenJ9. We only claim to have OpenJ9 support, and I'm wondering if OpenJ9 has always had the `com.sun.management.OperatingSystemMXBean` class available or not, now. If `com.sun.management.OperatingSystemMXBean` is always available, then we can simplify the implementation and always use that.

shakuzen:
Some more digging shows that even on the IBM JVM this has been changed in the latest version (8.0.0):
* 7.1: https://www.ibm.com/support/knowledgecenter/en/SSYKE2_7.1.0/com.ibm.java.api.71.doc/com.ibm.lang.management/com/ibm/lang/management/OperatingSystemMXBean.html
* 8.0: https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.api.80.doc/com.ibm.lang.management/com/ibm/lang/management/OperatingSystemMXBean.html

shakuzen:
Looking at the GitHub history of the OpenJ9 repository, it looks like it has always extended `com.sun.management.OperatingSystemMXBean`: https://github.com/eclipse/openj9/blob/openj9-0.8.0/jcl/src/jdk.management/share/classes/com/ibm/lang/management/OperatingSystemMXBean.java

If I'm not missing anything, things should be working as is. We could simplify the implementation to only check for the `com.sun.management.OperatingSystemMXBean` class, but that would break support for users of older versions of the IBM JVM that don't have that class available. It seems like the best course of action would be nothing for now. Let me know if I'm missing something.

izeye:

`ProcessorMetricsTest.hotspotCpuMetrics()` was running on OpenJ9 unintentionally due to the reason described above and failing somehow, so I assumed that it didn't work. I didn't check the reason, so I might miss something here. Does the test pass on OpenJ9 for you?

izeye:
@shakuzen Looking into the test, I can see the difference on assertions. Based on your feedback, just updating tests seem to be sufficient here. Is it okay to do so?

shakuzen:
> `ProcessorMetricsTest.hotspotCpuMetrics()` was running on OpenJ9 unintentionally due to the reason described above and failing somehow

It's failing for me too. The reason is in the comment for the `openJ9CpuMetrics` test.
https://github.com/micrometer-metrics/micrometer/blob/49162a881c4c80976980d6ea0ef45d2d59c4ca14/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/ProcessorMetricsTest.java#L58-L64

I wish we had a better way to test this.

> just updating tests seem to be sufficient here. Is it okay to do so?

Yes, that sounds good. The current assumptions aren't accurate.

shakuzen:
Looking at the OpenJ9 tests, there's some complexity I'm not sure we want, but here is how it looks like they're testing using the two methods we're using:

* `getSystemCpuLoad`: https://github.com/eclipse/openj9/blob/2f295ea91cdd97ff37eb5d164fc63014ea718bc7/test/functional/JLM_Tests/src/org/openj9/test/management/TestGetSystemCpuLoad.java#L56-L91
* `getProcessCpuLoad`: https://github.com/eclipse/openj9/blob/2f295ea91cdd97ff37eb5d164fc63014ea718bc7/test/functional/JLM_Tests/src/org/openj9/test/management/TestOperatingSystemMXBean.java#L395-L443

izeye:
@shakuzen Thanks for the quick feedback! I updated in 9083c3a.

PR #1768, opened 2019-12-27
Make functionTimerUnits() pass in OpenJ9

izeye:
This PR adds a dummy assertion to the end of `MeterRegistryCompatibilityKit.functionTimerUnits()` to make it pass in OpenJ9 as OpenJ9 garbage-collects it eagerly.


izeye:
The reason why I added a dummy assertion was that I saw similar failures in some interface tests. However, it doesn't happen as much as `MeterRegistryCompatibilityKit.functionTimerUnits()` does. So I just changed it to be a member field instead.

PR #1778, opened 2020-01-05
Allow NaN for tomcat.threads.busy in tests

izeye:
The initial value for `tomcat.threads.busy` seems to be 0 all the time in the 1.1.x branch which uses `org.apache.tomcat.embed:tomcat-embed-core:8.5.34`. However, it seems to be possible to be `Double.NaN` in the 1.3.x branch which uses `org.apache.tomcat.embed:tomcat-embed-core:8.5.46` due to the following exception:

```
javax.management.AttributeNotFoundException:  Cannot find attribute currentThreadsBusy for org.apache.tomcat.util.net.SocketProperties@198d6542
	at org.apache.tomcat.util.modeler.ManagedBean.getGetter(ManagedBean.java:435)
	at org.apache.tomcat.util.modeler.BaseModelMBean.getAttribute(BaseModelMBean.java:167)
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.getAttribute(DefaultMBeanServerInterceptor.java:647)
	at com.sun.jmx.mbeanserver.JmxMBeanServer.getAttribute(JmxMBeanServer.java:678)
	at io.micrometer.core.instrument.binder.tomcat.TomcatMetrics.lambda$null$2(TomcatMetrics.java:140)
	at io.micrometer.core.instrument.binder.tomcat.TomcatMetrics.safeDouble(TomcatMetrics.java:313)
	at io.micrometer.core.instrument.binder.tomcat.TomcatMetrics.lambda$null$3(TomcatMetrics.java:140)
	at io.micrometer.core.instrument.internal.DefaultGauge.value(DefaultGauge.java:54)
	at io.micrometer.core.instrument.binder.tomcat.TomcatMetricsTest.checkMbeansInitialState(TomcatMetricsTest.java:251)
	at io.micrometer.core.instrument.binder.tomcat.TomcatMetricsTest.lambda$mbeansAvailableBeforeBinder$3(TomcatMetricsTest.java:201)
	at io.micrometer.core.instrument.binder.tomcat.TomcatMetricsTest.runTomcat(TomcatMetricsTest.java:234)
	at io.micrometer.core.instrument.binder.tomcat.TomcatMetricsTest.mbeansAvailableBeforeBinder(TomcatMetricsTest.java:190)
...
```

This PR changes to allow it to be `Double.NaN` in addition to 0.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
@@ -237,7 +237,7 @@ class TomcatMetricsTest {
         assertThat(registry.get("tomcat.global.request").functionTimer().totalTime(TimeUnit.MILLISECONDS)).isEqualTo(0.0);
         assertThat(registry.get("tomcat.global.request.max").timeGauge().value(TimeUnit.MILLISECONDS)).isEqualTo(0.0);
         assertThat(registry.get("tomcat.threads.config.max").gauge().value()).isGreaterThan(0.0);
-        assertThat(registry.get("tomcat.threads.busy").gauge().value()).isEqualTo(0.0);
+        assertThat(registry.get("tomcat.threads.busy").gauge().value()).isIn(Double.NaN, 0.0);
         assertThat(registry.get("tomcat.threads.current").gauge().value()).isGreaterThan(0.0);
         assertThat(registry.get("tomcat.cache.access").functionCounter().count()).isEqualTo(0.0);
         assertThat(registry.get("tomcat.cache.hit").functionCounter().count()).isEqualTo(0.0);


shakuzen:
It looks like all of the Tomcat metrics that rely on the MBeanServer are not present, from a quick running of a sample application. The Tomcat MBean we're expecting is not being registered anymore, perhaps. Can we do something about that in the test instead?

izeye:
> It looks like all of the Tomcat metrics that rely on the MBeanServer are not present, from a quick running of a sample application.

@shakuzen IIRC, it did work for me. Could you please share the sample?

shakuzen:
> Could you please share the sample?

I generated a Spring Boot app [on Initializr selecting web and actuator dependencies](https://start.spring.io/#!type=maven-project&language=java&platformVersion=2.2.2.RELEASE&packaging=jar&jvmVersion=1.8&groupId=com.example&artifactId=demo&name=demo&description=Demo%20project%20for%20Spring%20Boot&packageName=com.example.demo&dependencies=web,actuator), and adding the property `management.endpoints.web.exposure.include=*` to application.properties and checked the `/actuator/metrics` endpoint.

I noticed there's a `server.tomcat.mbeanregistry.enabled` property that is `false` by default. If I switch this to `true`, then I can see the Tomcat metrics that rely on the MBeanServer in the `/actuator/metrics` endpoint.

The test in question doesn't include Spring Boot, but I'm wondering if something changed in Tomcat around registering its MBean registry.

checketts:
With spring boot 2.2 isn't jmx disabled by default? Is that the problem?

shakuzen:
> With spring boot 2.2 isn't jmx disabled by default? Is that the problem?

Tomcat registers its own MBeanServer, I thought, which I thought would be used in our tests without Spring Boot. We aren't starting an embedded Tomcat, so maybe it doesn't apply either way, but we used to consistently get `0.0` but now are getting `NaN`. I'm just trying to understand what changed in the Tomcat versions between the `1.1.x` and `1.3.x` branches that's causing the behavior difference. I'm also confused why the tests started to fail now after the `Thread.sleep` was added.

izeye:
> I'm also confused why the tests started to fail now after the `Thread.sleep` was added.

@shakuzen It's due to the following change in #1777:

> This PR also adds a missing `checkMbeansInitialState()` invocation in `TomcatMetricsTest.mbeansAvailableBeforeBinder()`.

shakuzen:
I looked into this some more.
https://github.com/micrometer-metrics/micrometer/blob/c3034f9e6369be881d0a3f968686a52fd272ea89/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java#L231

The above line when run for the Thread metrics will return a Set of two `ObjectName` objects.
```
0 = {ObjectName@3248} "Tomcat:type=ThreadPool,name="http-nio-57015""
1 = {ObjectName@3249} "Tomcat:type=ThreadPool,name="http-nio-57015",subType=SocketProperties"
```
We register the metrics using these in the order an iterator returns them. If it is in this order above then the test will pass as is. If it is in the other order, the test will fail because of the error mention in the description.

shakuzen:
As a follow-up to my previous comment, the 1.1.x branch always seems to return a single `ObjectName` when registering the Thread metrics:

```
Tomcat:type=ThreadPool,name="http-nio-57246"
```

adriancole:
would it make sense to merge this with an open issue for a more through investigation?

shakuzen:
> would it make sense to merge this with an open issue for a more through investigation?

Given the build is currently broken due to this, yes, I agree.

Issue #1786, opened 2020-01-10
micrometer-core contains modified code from Guava, Jettison and Commons-lang without proper copyright attribution


tazle:
According to several comments in micrometer-core, it contains modified source code from Guava, Jettison and Commons-lang:
```
$ git grep -C1 'Modified'
micrometer-core/src/main/java/io/micrometer/core/instrument/util/AbstractPartition.java- *
micrometer-core/src/main/java/io/micrometer/core/instrument/util/AbstractPartition.java: * Modified from {@link com.google.common.collect.Lists#partition(List, int)}.
micrometer-core/src/main/java/io/micrometer/core/instrument/util/AbstractPartition.java- *
--
micrometer-core/src/main/java/io/micrometer/core/instrument/util/StringEscapeUtils.java-    /**
micrometer-core/src/main/java/io/micrometer/core/instrument/util/StringEscapeUtils.java:     * Modified from the quote method in:
micrometer-core/src/main/java/io/micrometer/core/instrument/util/StringEscapeUtils.java-     * https://github.com/codehaus/jettison/blob/master/src/main/java/org/codehaus/jettison/json/JSONObject.java
--
micrometer-core/src/main/java/io/micrometer/core/instrument/util/StringUtils.java-     *
micrometer-core/src/main/java/io/micrometer/core/instrument/util/StringUtils.java:     * Modified from {@link org.apache.commons.lang.StringUtils#isBlank(String)}.
micrometer-core/src/main/java/io/micrometer/core/instrument/util/StringUtils.java-     *
```

I could not find appropriate copyright attribution in the micrometer-core distribution, or in the project documentation.

tazle:
Note that in case of https://github.com/codehaus/jettison/blob/master/src/main/java/org/codehaus/jettison/json/JSONObject.java, looking at the copyright statement there, and the code itself, it seems likely that the code is actually a derivative work of https://github.com/stleary/JSON-java/blob/master/JSONObject.java which is notoriously *not* licensed under the permissive Apache 2.0 license, but rather the JSON.org license, which forbids using the software for Evil.

adriancole:
Experience from working in ASF space has that you are required to retain the copyright notice (NOTICE), if a project has one. The advisory copyright line in headers is a bit mushier. I've certainly seen projects do different things here, from nothing, to duplicating the header just to retain an old one, to always including a NOTICE file anytime ASL is used, to a combination of all the above.

[...]

PR #1788, opened 2020-01-12
Reduces code used to implement AbstractPartition and backfills tests

adriancole:
This accidentally used some fancy things not needed and triggered #1786


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmCompilationMetricsTest.java
```diff
@@ -23,7 +23,7 @@
 import java.lang.management.ManagementFactory;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
-import static org.junit.Assume.assumeTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
```
shakuzen: Thanks for catching this. I've pushed this fix to `master` now in 5876298bd66443ecefe4c064a9ea67f98026bf66. I'm confused how the build passed with this before... but nonetheless fixed now.

adriancole: thx for follow-up!

shakuzen:
Thanks Adrian. I've rebased this onto `1.3.x` so we can include this in our upcoming patch release for that line.

PR #1790, opened 2020-01-12
Changes StringEscapeUtils to be based on ASL code

adriancole:
This switches to code I wrote for Zipkin which was adapted from
Moshi, an ASL 2.0 project with no NOTICE file.

See #1786


shakuzen:
Thanks again. I've rebased this onto `1.1.x`.

PR #1793, opened 2020-01-13
Add support for Stackdriver resource labels

jmccance:
Add support for defining arbitrary resource labels to the monitored
resource when configuring creating a time series.

Stackdriver enables associating metrics with different "monitored
resources", such as `k8s_container`. Monitored resources can specify
certain required "resource labels" (distinct from metric labels), such
as `pod_name` for `k8s_container`. Any metric submitted to that resource
type without all of the required resource labels will be rejected. By
enabling arbitrary resource labels to be defined in StackdriverConfig,
we'll enable more flexible integration with Stackdriver.

### DIFF
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverConfig.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverConfig.java
@@ -18,6 +18,9 @@ package io.micrometer.stackdriver;
 import io.micrometer.core.instrument.config.MissingRequiredConfigurationException;
 import io.micrometer.core.instrument.step.StepRegistryConfig;
 
+import java.util.Collections;
+import java.util.Map;
+
 /**
  * {@link StepRegistryConfig} for Stackdriver.
  *
@@ -41,5 +44,9 @@ public interface StackdriverConfig extends StepRegistryConfig {
     default String resourceType() {
         String resourceType = get(prefix() + ".resourceType");
         return resourceType == null ? "global" : resourceType;
+    }
+
+    default Map<String, String> resourceLabels() {
+        return Collections.emptyMap();
     }
 }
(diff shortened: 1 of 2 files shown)


saturnism:
is micrometer tag translated to label?

saturnism:
regardless this is a nice addition :)

jmccance:
@saturnism No changes to how tags are currently handled. I believe tags are all added as labels to the metric (aka time series). The resource labels are separate, and don't change from moment to moment. It seems to be one way around Stackdriver's limit on 10 labels per time series.

jmccance:
Could someone with appropriate permissions retry my build in Circle? Ran both `./gradlew micrometer-spring-legacy:test` and `./gradlew check test` locally and both passed fine. Since there's no output to indicate what exactly failed, not sure I can do much additional debugging from my end. 😞

jmccance:
Pushed an empty commit to re-run the build and it came up green. ✅ Also green after rebasing to remove said empty commit.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverConfig.java
```diff
@@ -42,4 +45,8 @@ default String resourceType() {
         String resourceType = get(prefix() + ".resourceType");
         return resourceType == null ? "global" : resourceType;
     }
+
+    default Map<String, String> resourceLabels() {
+        return Collections.emptyMap();
```
shakuzen: To allow for flexibility in implementations of the Config interfaces, we typically retrieve the config property using the `get` method which returns a String and then convert to the desired return type. For example, an implementation of `StackdriverConfig` could be `StackdriverConfig config = System::getProperty;` which allows configuring all the properties via system properties, except for resourceLabels right now.

I'm not sure how we would convert a String to a Map, though. We may not have much of a choice in this case.

jmccance: Yeah, I'm not sure there's a good way to handle that, since we'd have to define our own string-encoding of what a Map looks like. If you do want to go that route, let me know and I'll 

FWIW, I found another case where a Config object expects a map in `HumioConfig`. Looks like a nearly identical situation. There we're returning `null`, but I'd much rather return an empty map since it doesn't require any special handling.

https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java#L48-L58

jmccance:
Moving to a new branch rebased off master. (I had committed straight to the tag in my fork. Live and learn.)

New PR here: https://github.com/micrometer-metrics/micrometer/pull/1818

PR #1794, opened 2020-01-14
Removes core shade dependency on pcollections

### LINKED ISSUE #1785: micrometer-core apparently distributes pcollections without proper attribution
While looking inside the micrometer-core JAR at https://repo1.maven.org/maven2/io/micrometer/micrometer-core/1.3.1/micrometer-core-1.3.1.jar I noticed that it apparently includes pcollections from https://github.com/hrldcpr/pcollections without including the copyright notice from https://github.com/hrldcpr/pcollections/blob/master/LICENSE in the distribution.

### LINKED ISSUE #1791: Remove core dependency on pcollections?
I was auditing how pcollections is used, as we were asked to cite it due to the side effects of shading in #1785.

I read through the benefits claimed by pcollections, notably how it allows lazy chaining without copying all values. However, in auditing use of the types in the core module, I notice collections are not folded.. all uses I could find were adding or removing single elements. It may have been more sophisticated in a version before, but currently, I'd question why we can't use SDK libraries instead of pcollections, especially considering all the ops happen under locks anyway. 

Adding a single element to a collection under a lock vs swapping references each time seems something we can prove out with benchmarks, and I'd prefer the simpler less dependencies way out vs using a third party library which isn't (as I perceive) being played to its strengths anyway.

adriancole:
This removes the shaded dependency on pcollections from the core
library. In doing so, this simplifies the build and eliminates the
need to do bookkeeping around the non-ASL dep.

Instead, common libraries take the task. Since formerly write ops
happened under locks anyway, there's not likely any performance
decrease expected.

NOTE: `MeterRegistry.getMeters()` previously used pcollections HashTreePMap which consistently ordered by the key's hashCode. We don't retain that behavior as hashCode algos are not stable anyway.

Fixes #1791

Partially fixes #1785, though it is still used in some implementations.
Follow-up change should occur, if this is merged, to yank from there
also.


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/JooqExecuteListener.java
```diff
@@ -69,40 +70,40 @@ public void end(ExecuteContext ctx) {
...
-            synchronized (sampleLock) {
-                sampleByExecuteContext = sampleByExecuteContext.minus(ctx);
-            }
+        Timer.Sample sample;
+        synchronized (sampleLock) {
+            sample = sampleByExecuteContext.remove(ctx);
+        }
+        if (sample == null) return;
 
-            String exceptionName = "none";
```
adriancole: everything below here is just indent related as the above early return removes the nesting

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -29,12 +29,9 @@
 import io.micrometer.core.instrument.search.Search;
 import io.micrometer.core.instrument.util.TimeUtils;
 import io.micrometer.core.lang.Nullable;
-import org.pcollections.HashTreePMap;
-import org.pcollections.HashTreePSet;
-import org.pcollections.PMap;
-import org.pcollections.PSet;
 
 import java.util.*;
```
anuraaga: Should probably clean this

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -300,6 +300,8 @@ Meter register(Meter.Id id, Meter.Type type, Iterable<Measurement> measurements)
      * @return The set of registered meters.
      */
     public List<Meter> getMeters() {
+        // NOTE: previously used pcollections HashTreePMap which consistently ordered by the key's
```
anuraaga: Maybe this is better on release notes instead of code

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -628,14 +631,14 @@ public Meter remove(Meter.Id id) {
...
         if (m != null) {
             synchronized (meterMapLock) {
-                m = meterMap.get(mappedId);
+                m = meterMap.remove(mappedId);
                 if (m != null) {
-                    meterMap = meterMap.minus(mappedId);
-
-                    for (Id synthetic : syntheticAssociations.getOrDefault(id, HashTreePSet.empty())) {
-                        remove(synthetic);
+                    Set<Id> synthetics = syntheticAssociations.remove(id);
```
anuraaga: Much nicer

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/JooqExecuteListener.java
```diff
@@ -33,7 +33,7 @@
     private final Supplier<Iterable<Tag>> queryTagsSupplier;
 
     private final Object sampleLock = new Object();
-    private volatile PMap<ExecuteContext, Timer.Sample> sampleByExecuteContext = HashTreePMap.empty();
+    private final Map<ExecuteContext, Timer.Sample> sampleByExecuteContext = new LinkedHashMap<>();
```
anuraaga: `HashMap` looks like enough don't see any ordered access.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -49,7 +49,7 @@
     private final Iterable<Tag> tags;
 
     private final Object connectionSamplesLock = new Object();
-    private volatile PMap<Connection, Timer.Sample> connectionSamples = HashTreePMap.empty();
+    private final Map<Connection, Timer.Sample> connectionSamples = new LinkedHashMap<>();
```
anuraaga: Ditto

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
adriancole: agreed

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/JooqExecuteListener.java
adriancole: I personally use LHM for the sake of consistency in tests, but I notice that isn't used here, so sgtm

adriancole:
thx for the review @anuraaga

adriancole:
https://github.com/micrometer-metrics/micrometer/pull/1795 finishes the job

jkschneider:
There may be some expected performance degradation with `CHM`. Similar results with several runs. To really compare we'd need to implement both and run the benchmark at the same time, but I think the result would be similar.

## With PCollections

> Result "io.micrometer.benchmark.core.MeterRegistrationBenchmark.sum":
>   N = 1285521
>   mean =     ≈ 10⁻⁴ ±(99.9%) 0.001 ms/op
> 
>   Histogram, ms/op:
>     [0.000, 0.013) = 1284361 
>     [0.013, 0.025) = 920 
>     [0.025, 0.038) = 221 
>     [0.038, 0.050) = 10 
>     [0.050, 0.063) = 3 
>     [0.063, 0.075) = 1 
>     [0.075, 0.088) = 2 
>     [0.088, 0.100) = 1 
>     [0.100, 0.113) = 0 
>     [0.113, 0.125) = 0 
>     [0.125, 0.138) = 1 
>     [0.138, 0.150) = 0 
>     [0.150, 0.163) = 0 
>     [0.163, 0.175) = 1 
>     [0.175, 0.188) = 0 
> 
>   Percentiles, ms/op:
>       p(0.0000) =     ≈ 10⁻⁶ ms/op
>      p(50.0000) =     ≈ 10⁻⁵ ms/op
>      p(90.0000) =     ≈ 10⁻⁴ ms/op
>      p(95.0000) =     ≈ 10⁻⁴ ms/op
>      p(99.0000) =     ≈ 10⁻⁴ ms/op
>      p(99.9000) =      0.001 ms/op
>      p(99.9900) =      0.028 ms/op
>      p(99.9990) =      0.045 ms/op
>      p(99.9999) =      0.154 ms/op
>     p(100.0000) =      0.163 ms/op

## With CHM

Result "io.micrometer.benchmark.core.MeterRegistrationBenchmark.sum":
  N = 1170573
  mean =     ≈ 10⁻⁴ ±(99.9%) 0.001 ms/op

  Histogram, ms/op:
    [0.000, 0.050) = 1170567 
    [0.050, 0.100) = 4 
    [0.100, 0.150) = 1 
    [0.150, 0.200) = 0 
    [0.200, 0.250) = 0 
    [0.250, 0.300) = 0 
    [0.300, 0.350) = 0 
    [0.350, 0.400) = 0 
    [0.400, 0.450) = 0 
    [0.450, 0.500) = 0 
    [0.500, 0.550) = 1 
    [0.550, 0.600) = 0 

  Percentiles, ms/op:
      p(0.0000) =     ≈ 10⁻⁶ ms/op
     p(50.0000) =     ≈ 10⁻⁴ ms/op
     p(90.0000) =     ≈ 10⁻⁴ ms/op
     p(95.0000) =     ≈ 10⁻⁴ ms/op
     p(99.0000) =     ≈ 10⁻⁴ ms/op
     p(99.9000) =      0.001 ms/op
     p(99.9900) =      0.027 ms/op
     p(99.9990) =      0.037 ms/op
     p(99.9999) =      0.436 ms/op
    p(100.0000) =      0.505 ms/op

jkschneider:
Also, there are other immutable collections libraries now with ASL licenses, if this is really the concern, e.g. [vavr](https://github.com/vavr-io/vavr).

adriancole:
I think the javadoc sums up the blocking concerns, paraphrased into blocking is a bug. Iteration implied to populate getMeters safe copy (CHM.values()) "seems fine" I'm not sure we need to reproduce JMH work on CHM.

Personally, not much interest trying to swap and then shade with another 3rd party thing unless there's a significant reason. Seems easier to grok using the most straightforward solution and with my current understanding, this is it.

@raphw I know you have some fancy maps and good reason for them. Do you see a reason for this meter registry to intentionally not use CHM? If so, what would you use, SDK or otherwise?

raphw:
I don't really see a point in external collection frameworks unless using primitive collections, weak references concurrently (weak-lock-free) or highly concurrent queues (JCTools). Everything else is mainly for syntax convenience. Rather wrap immutable collections via the Collections class and define a contract for it.

jkschneider:
I did a bunch of tests last night, and I think this seems fine. The key, as @adriancole pointed out, is that `getMeters()` makes a copy, so it doesn't really matter how slow publishing is.

adriancole:
thanks tons for the look, jon!

thx also raphael for the feedback

PR #1795, opened 2020-01-14
Removes pcollections from statsd implementations

### LINKED ISSUE #1785: micrometer-core apparently distributes pcollections without proper attribution

adriancole:
This continues from #1794 and completes removing pcollections and its
MIT license attribution concerns, from the build.

This approach is to use `ConcurrentMap.computeIfAbsent` instead of the
former pcollections types. A side effect of this is more readable code.

Fixes #1785

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/EtsyStatsdLineBuilder.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/EtsyStatsdLineBuilder.java
@@ -21,17 +21,17 @@ import io.micrometer.core.instrument.Statistic;
 import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.util.HierarchicalNameMapper;
 import io.micrometer.core.lang.Nullable;
-import org.pcollections.HashTreePMap;
-import org.pcollections.PMap;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 public class EtsyStatsdLineBuilder extends FlavorStatsdLineBuilder {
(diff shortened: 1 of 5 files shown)


adriancole:
Note: the first three commits are from #1794 and I'll rebase once that's merged

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/EtsyStatsdLineBuilder.java
```diff
@@ -21,17 +21,17 @@
...
 
 public class EtsyStatsdLineBuilder extends FlavorStatsdLineBuilder {
     private final HierarchicalNameMapper nameMapper;
-    private final Object namesLock = new Object();
     @SuppressWarnings({"NullableProblems", "unused"})
     private volatile NamingConvention namingConvention;
     @Nullable
     private volatile String nameNoStat;
-    private volatile PMap<Statistic, String> names = HashTreePMap.empty();
+    private final ConcurrentMap<Statistic, String> names = new ConcurrentHashMap<>();
```
jkschneider: One side effect of this I believe is that iteration (which may happen in publishing) limits throughput on operations that are potentially generating new tags in another thread and adding meters to the map. I think...

jkschneider:
@adriancole If you'd like to rebase, we can merge?

adriancole:
rebased, thanks!

PR #1798, opened 2020-01-15
Fix execution of ./gradlew :micrometer-benchmarks-core:jmh

anuraaga:
Main fix is to bump the `gradle-jmh-plugin` version (0.4.8 doesn't support Gradle 5.5+).  After that, found the need to exclude duplicates (due to java 9 dependencies, servlet) and enable big zips.

### DIFF
--- a/benchmarks/benchmarks-core/build.gradle
+++ b/benchmarks/benchmarks-core/build.gradle
@@ -16,4 +16,6 @@ jmh {
     fork = 1
     warmupIterations = 1
     iterations = 1
+    duplicateClassesStrategy = 'exclude'
+    zip64 = true
 }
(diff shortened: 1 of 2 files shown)


adriancole:
nice. will try later but I suspect end result is gradle jmh works.

while I had tried updating I didnt think through why the zip error was occurring. nice work

shakuzen:
@anuraaga could you rebase onto `1.3.x` so we can target this at there (`git rebase --onto 1.3.x master jmh-fixes`)? I tried to do it myself but got permission denied.

anuraaga:
Sure ran that command

adriancole:
verified this works. thanks tons!

PR #1699, opened 2019-11-10
Upgrade to Gradle 6.0

izeye:
I was trying to upgrade to Gradle 6.0, but I got the following failure:

```
FAILURE: Build failed with an exception.

* What went wrong:
Failed to notify build listener.
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
> org.gradle.api.publish.maven.internal.publication.MavenPublicationInternal.getPublishableFiles()Lorg/gradle/api/file/FileCollection;
[...]

PR #1801, opened 2020-01-16
Upgrade to Gradle 6.1

anuraaga:
When working on #1798, it was a bit annoying since my normal `JAVA_HOME` is Java 13. It'd be good to upgrade to Gradle 6.0 so it's easier for contributors that might be using it.

I noticed the nebula plugins used are very old - constraining some transitive dependencies allows things to work but updating or removing them might make sense.

Also
- Enable annotation processor for JMH. This means benchmark main methods will work when Gradle is configured to build / run using Gradle. I recommend using that setting instead of IntelliJ since it tends to have less gotchas and doesn't require additional configuration of IntelliJ for things to work
- Make Gradle evaluation debug message more precise (not necessarily building)


anuraaga:
Note I had a build failure which looked like OOM, running once more to see if flake or not.

anuraaga:
Green now - I was working on 1.3.x before mistakenly thinking it's a good branch to work on :)

One caveat is that I've only tested normal build type tasks, including `publishToMavenLocal`. It's not completely verified this works without running all tasks that you use, such as artifact deployment tasks, I can't guarantee those wouldn't throw a MethodNotFoundException without trying...

shakuzen:
This supersedes #1699

PR #1804, opened 2020-01-16
Add pcollections copyright and license to LICENSE file

shakuzen:
While #1794 and #1794 replaced our usage of pcollections for the upcoming 1.4.0 release, maintenance branches still use it, so we need to include the MIT license from pcollections. This will be added for the `1.1.x` and `1.3.x` branches but not for the `master` branch where it does not apply.

See #1785

---

We shade the pcollections classes under a different group ID at build time for us in the micrometer-core and micrometer-registry-statsd modules. Since its license is not the Apache v2 license, its license and copyright should be mentioned in the LICENSE file.

---

This follows the instructions in [Apache's licensing how-to for permissively-licensed dependencies](https://www.apache.org/dev/licensing-howto.html#permissive-deps) opting to copy the short MIT license in-line rather than use a "pointer".

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,3 +199,32 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+
+OTHER LICENSED CODE:
+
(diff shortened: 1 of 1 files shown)


adriancole:
looks legit

Issue #491, opened 2018-03-14
Support Graphite 1.1.x tags


ITman1:
Hi,
is there a plan to support exporting metrics into Graphite using tags http://graphite.readthedocs.io/en/latest/tags.html?
Thank you.

jkschneider:
There is now ;). Thanks for bringing it to our attention.

ibhi:
Hi, any recent updates on this issue?

shakuzen:
No update so far. If someone were to send a pull request that added this support, it would help move things along. To that end, I've labeled the issue with the `help wanted` tag.

fitzoh:
I'm suddenly interested in this...

Any thoughts on what the implementation looks like?
A switch in the graphite registry config that changes the namer?

PR #1806, opened 2020-01-18
Graphite tag support

fitzoh:
~~Super Janky~~ Quite reasonable (@jkschneider edit) support for #491.

Adds an alternative dimensional name mapper and naming convention for graphite that supports [tagging](https://graphite.readthedocs.io/en/latest/tags.html).

### DIFF
--- a/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteDimensionalNamingConvention.java
+++ b/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteDimensionalNamingConvention.java
@@ -1,0 +1,80 @@
+/**
+ * Copyright 2020 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.graphite;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.NamingConvention;
+import io.micrometer.core.lang.Nullable;
+
+import java.text.Normalizer;
+import java.util.regex.Pattern;
+
+/**
+ * {@link NamingConvention} for Graphite.
+ *
+ * @author Jon Schneider
+ * @author Johnny Lim
+ * @author Andrew Fitzgerald
+ */
+public class GraphiteDimensionalNamingConvention implements NamingConvention {
+    /**
+     * A list that probably is blacklisted: https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/grammar.py#L48-L55.
+     * Empirically, we have found others.
+     */
+    private static final Pattern PATTERN_NAME_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ ?:]");
+    private static final Pattern PATTERN_TAG_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ ?:.]");
+    private final NamingConvention delegate;
+
+    public GraphiteDimensionalNamingConvention() {
+        this(NamingConvention.camelCase);
+    }
+
+    public GraphiteDimensionalNamingConvention(NamingConvention delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public String name(String name, Meter.Type type, @Nullable String baseUnit) {
+        return sanitizeName(this.delegate.name(normalize(name), type, baseUnit));
+    }
+
+    @Override
+    public String tagKey(String key) {
+        return sanitizeTag(this.delegate.tagKey(normalize(key)));
+    }
+
+    @Override
+    public String tagValue(String value) {
+        return sanitizeTag(this.delegate.tagValue(normalize(value)));
+    }
+
+    /**
+     * Github Issue: https://github.com/graphite-project/graphite-web/issues/243
+     * Unicode is not OK. Some special chars are not OK.
+     */
+    private String normalize(String name) {
+        return Normalizer.normalize(name, Normalizer.Form.NFKD);
+    }
+
+    private String sanitizeName(String delegated) {
+        return PATTERN_NAME_BLACKLISTED_CHARS.matcher(delegated).replaceAll("_");
+    }
+
+    private String sanitizeTag(String delegated) {
+        return PATTERN_TAG_BLACKLISTED_CHARS.matcher(delegated).replaceAll("_");
+    }
+
+}
--- a/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteDimensionalNamingConvention.java
+++ b/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteDimensionalNamingConvention.java
@@ -34,12 +34,13 @@ public class GraphiteDimensionalNamingConvention implements NamingConvention {
      * A list that probably is blacklisted: https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/grammar.py#L48-L55.
      * Empirically, we have found others.
      */
-    private static final Pattern PATTERN_NAME_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ ?:]");
-    private static final Pattern PATTERN_TAG_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ ?:.]");
+    private static final Pattern PATTERN_NAME_BLACKLISTED_CHARS = Pattern.compile("[{}(),=\\[\\]/ ?:;]");
+    private static final Pattern PATTERN_TAG_KEY_BLACKLISTED_CHARS = Pattern.compile("[;!^=]");
+    private static final Pattern PATTERN_TAG_VALUE_BLACKLISTED_CHARS = Pattern.compile("[;~]");
     private final NamingConvention delegate;
 
     public GraphiteDimensionalNamingConvention() {
-        this(NamingConvention.camelCase);
+        this(NamingConvention.dot);
     }
 
     public GraphiteDimensionalNamingConvention(NamingConvention delegate) {
@@ -53,12 +54,12 @@ public class GraphiteDimensionalNamingConvention implements NamingConvention {
 
     @Override
     public String tagKey(String key) {
-        return sanitizeTag(this.delegate.tagKey(normalize(key)));
+        return sanitizeTagKey(this.delegate.tagKey(normalize(key)));
     }
 
     @Override
     public String tagValue(String value) {
-        return sanitizeTag(this.delegate.tagValue(normalize(value)));
+        return sanitizeTagValue(this.delegate.tagValue(normalize(value)));
     }
 
     /**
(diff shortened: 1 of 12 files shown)


implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteTagNameMapper.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Tag based {@link HierarchicalNameMapper} for Graphite.
+ *
+ * @author Jon Schneider
+ * @author Johnny Lim
+ * @author Andrew Fitzgerald
```
fitzoh: Copy-Paste-Modify from the `GraphiteHierarchicalNameMapper` (same story for the test)

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
```diff
@@ -0,0 +1,52 @@
...
+class GraphiteTagNameMapperTest {
+    private final GraphiteTagNameMapper nameMapper = new GraphiteTagNameMapper("stack", "app.name");
+    private final Meter.Id id = new SimpleMeterRegistry().counter("my.name",
+            "app.name", "MYAPP", "stack", "PROD", "other.tag", "value").getId();
+
+    @Issue("#561")
+    @Test
+    void tagsAsPrefix() {
+        assertThat(nameMapper.toHierarchicalName(id, NamingConvention.camelCase))
+                .isEqualTo("PROD.MYAPP.myName;otherTag=value");
```
fitzoh: Does it make sense to allow tag prefixing if actual tags are an option?

implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteTagNameMapper.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class GraphiteTagNameMapper implements HierarchicalNameMapper {
```
fitzoh: Once we have tags, is it really a hierarchical metric?

If we say it is, that makes it easier to reuse the existing graphite registry and just sub in the new name mapper....

If it doesn't there's a lot more work to do since we're also likely no longer using the dropwizard registry as our base (right?)

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link GraphiteTagNameMapper}.
+ *
+ * @author Jon Schneider
+ * @author Johnny Lim
+ * @author Andrew Fitzgerald
+ */
+class GraphiteTagNameMapperTest {
```
fitzoh: TODO: Add tests/logic for valid chars for graphite tags/values
https://graphite.readthedocs.io/en/latest/tags.html#carbon

>Tag names must have a length >= 1 and may contain any ascii characters except `;!^=`. Tag values must also have a length >= 1, and they may contain any ascii characters except `;~`.

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
jkschneider: I would think not? Can you think of a case where we would want it? We could adopt the view here too that it would be peculiar for someone to want some mixture of prefixing and tags support, that if they did they'd likely want to configure one tag as a prefix and the rest as Graphite tags, etc. And the API already supports them creating their own `HierarchicalNameMapper` and doing just that.

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
jkschneider: Will we have to update `GraphiteNamingConvention` to put a placeholder tag value like `unknown` in when the tag value is blank?

implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteTagNameMapper.java
jkschneider: First impression: If this tag name mapper is the default `HierarchicalNameMapper` (which I think makes sense), then most people would never even interact with the type because they'd use one of the other constructors like `GraphiteMeterRegistry(GraphiteConfig config, Clock clock)`.

And I think it makes sense for this to be a specialization of `HierarchicalNameMapper`, because unlike purely dimensional monitoring systems, Graphite doesn't _require_ the use of this tags support. Even folks using Graphite >1.1 may have their reasons for avoiding it (e.g. consistency with existing names/dashboards?).

So I think you have done the right thing.

implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteTagNameMapper.java
```diff
@@ -0,0 +1,61 @@
...
+    private final List<String> tagsAsPrefix;
+
+    public GraphiteTagNameMapper(String... tagsAsPrefix) {
+        this.tagsAsPrefix = Arrays.asList(tagsAsPrefix);
+    }
+
+    @Override
+    public String toHierarchicalName(Meter.Id id, NamingConvention convention) {
+        StringBuilder hierarchicalName = new StringBuilder();
+        for (String tagKey : tagsAsPrefix) {
```
jkschneider: I think we can strike this.

implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteTagNameMapper.java
jkschneider: No issue other than just please update the year on the license header at top.

fitzoh: 5b5a6ae78ca4e56213df8e222b1113adeb26aa3c

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
fitzoh: a1bb940f936c692f8c3411821e0ccbe7b096e775

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
fitzoh: >Will we have to update GraphiteNamingConvention to put a placeholder tag value like `unknown` in when the tag value is blank?

I believe so (although I think i'm leaning towards `unspecified`).


>Also, should update the naming convention to bar the use of ; and =.

👍

fitzoh: So I added a (failing) test suite in 612566a1ddf4740d8a6485083064740873e1c88c.

After we dropped the prefix tags, I think it makes sense to use a new naming convention for tag based Graphite metrics.

I don't think camel casing makes since any more, instead we should honor the metric name passed in by the user unless there are invalid characters.

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
```diff
@@ -0,0 +1,96 @@
...
+    private final GraphiteTagNameMapper nameMapper = new GraphiteTagNameMapper();
+    private final SimpleMeterRegistry registry = new SimpleMeterRegistry();
+    private final NamingConvention namingConvention = new GraphiteNamingConvention();
+
+    @Test
+    void simpleName() {
+        Meter meter = registry.counter("a.simple.counter");
+        assertThat(getName(meter)).isEqualTo("a.simple.counter");
```
fitzoh: Now that tags aren't part of the metric name, I think it makes sense to use them as is instead of camel casing

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
```diff
@@ -0,0 +1,96 @@
...
+     * > Tag values must also have a length >= 1, and they may contain any ascii characters except ;~
+     * > UTF-8 characters may work for names and values, but they are not well tested and it is not recommended to use non-ascii characters in metric names or tags.
+     * - https://graphite.readthedocs.io/en/latest/tags.html#carbon
+     *
+     */
+    @Test
+    void terribleTagNaming() {
+        Meter meter = registry.counter("a.simple.counter", "key;!^=~ why?", "value;!^=~ why?");
+        assertThat(getName(meter)).isEqualTo("a.simple.counter;key____~ why?=value_!^=_ why?");
+    }
```
fitzoh: Spaces are the biggest question mark for me, but they're not in the explicitly disallowed character list.

implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteTagNameMapper.java
fitzoh: a1bb940

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
fitzoh: I'm thinking I did that wrong, and those tests should really be for a new `GraphiteHierarchicalNamingConvention`

jkschneider: > I don't think camel casing makes since any more, instead we should honor the metric name passed in by the user unless there are invalid characters.

Agreed, the only reason for camel casing was so they could do wildcard queries for hierarchical names where the tags embedded in the name were separating by dots. So the dot syntax is fine now with the richer tags.

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
jkschneider: Should be easy to test. JVM memory metrics ship a tag value with a space in it by default.

implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteDimensionalNamingConvention.java
```diff
@@ -0,0 +1,92 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.graphite;
+
+import com.google.common.base.Strings;
```
jkschneider: Need to strike this unused import and a couple other little checkstyle problems.

fitzoh: 2e08e9c5e467524292e6faec6df35a7dd9e12c08

implementations/micrometer-registry-graphite/src/test/java/io/micrometer/graphite/GraphiteTagNameMapperTest.java
fitzoh: Yeah, verified with a team member/contributor that spaces are valid, and it's pretty clear the source code allows it: https://github.com/graphite-project/carbon/blob/c5d7cc8a4f2eb0429594c14e035869b38cc79cf3/lib/carbon/util.py#L342-L359

jkschneider:
Thanks for the cool feature @fitzoh!

workless:
This PR seems broken, it always generate broken names due to DropWizard behavior.

From DropWizard, GraphiteReporter [1]:
`graphite.send(prefix(name, type.getCode()), format(value), timestamp);`
where
name="my.metric;tag=value"
type.getCode() = "p99"

final string sent to graphite: "my.metric;tag=valuep99"

Unless we patch DropWizard side i can't see how this code may work

[1] https://github.com/dropwizard/metrics/blob/v4.1.7/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java#L351

shakuzen:
@workless could you please open a new issue so we can take a look at it? /cc @fitzoh

PR #1812, opened 2020-01-21
Use java-platform to configure platform instead of spring dependencies.

anuraaga:
I noticed that spring dependencies plugin is being used to define a `java-platform`, even though `java-platform` comes with syntax to do it natively. So switched to native syntax.

Also fixed a bug from #1801 where the `maven-publish` configuration needs to be outside `if(project != 'micrometer-bom')` because the bom is also published. Currently, if it were published Gradle 6+ users would have a problem since the generated Gradle metadata does not interface with nebula and has no entries.

For reference: https://docs.gradle.org/current/userguide/java_platform_plugin.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,21 +74,21 @@ subprojects {
             ext.year = Calendar.getInstance().get(Calendar.YEAR)
             skipExistingHeaders = true
         }
+    }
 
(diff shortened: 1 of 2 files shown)


adriancole:
drive-by: less special is more better

PR #1819, opened 2020-01-26
Move dependency version declarations to one location using dependency…

anuraaga:
… constraints.

When defining dependencies in every project, it's easy for versions to get misaligned, for example in PR's like #1816. Gradle's constraints are equivalent to Maven's dependency management and allow defining versions in one location which should reduce this sort of drift.

By stripping out the `dependencyManagement` from POMs, this results in no real change compared to the current POMs while still being more DRY.

I found a misaligned dependency, `reactor-netty` which used `latest.release` in one place and a fixed version in another. For now, I've stuck with the fixed version but let me know if one is preferred.

Also removed the redundant `commons-text` dependency we found in previous PR.


anuraaga:
Found using native Gradle features like this doesn't work well with spring plugin's `downloadDependencies` will investigate

anuraaga:
Yup it'd work - constraints apply to transitive dependencies and unversioned direct ones, but not to versioned direct dependencies.

shakuzen:
> I found a misaligned dependency, `reactor-netty` which used `latest.release` in one place and a fixed version in another. For now, I've stuck with the fixed version but let me know if one is preferred.

I upgraded to `latest.release` in #1782 which I merged after you opened this pull request. I've rebased and merged it into this PR.

PR #1824, opened 2020-01-28
Add Gradle Wrapper Validation GitHub Action

izeye:
This PR adds Gradle Wrapper Validation GitHub Action.

See https://github.com/gradle/wrapper-validation-action


shakuzen:
Can this be used with older versions of Gradle? If so, we could apply this to the maintenance branches also.

izeye:
@shakuzen Thanks for the feedback! I changed the base branch to 1.1.x and confirmed that [it's working in my fork](https://github.com/izeye/micrometer/pull/1).

shakuzen:
Thanks for checking.

PR #1827, opened 2020-01-29
Add common metrics name prefix

pszemus:
I was missing a possibility to set a common prefix that will be added to every metric's name.
That PR addresses that.
It's trivial and it's only purpose it to start a discussion about that. If it's a good idea to have such a functionality I'll be glad to take any comments and correct the code.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdConfig.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdConfig.java
@@ -162,4 +162,12 @@ public interface StatsdConfig extends MeterRegistryConfig {
         String v = get(prefix() + ".buffered");
         return v == null || Boolean.parseBoolean(v);
     }
+
+    /**
+     * @return Common prefix that will be added to every metric's name.
+     */
+    default String namePrefix() {
+        String v = get(prefix() + ".namePrefix");
+        return v == null ? "" : v;
+    }
 }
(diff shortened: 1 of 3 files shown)


pszemus:
@checketts 
I have and application that runs in cloud, and of course "region", "hostname" are candidates for tags but I'd like every metric (eg. "flights_number") to be prefixed with application's name, so it could be found in path: application_name.flights_number.region.X.hostname.Y.value=Z

checketts:
Thanks for adding those details. So it sounds like you don't think a mapping of tags to the hierarchy would work in that case? 

I'm actually not familiar at all with actually using the hierarchical naming stuff since I don't run any of those currently. So I'll defer to other contributors in that regard.

jkschneider:
@pszemus There are a couple ways this can be accomplished fairly easily today, so I'd rather avoid adding a new config option if it really isn't necessary.

## Option 1: A `MeterFilter` prepending the name

As @checketts says, really adding a common tag to all metrics is better, as it's more portable. But this is easy.

```java
registry.config().meterFilter(new MeterFilter() {
    @Override
        return id.withName("app." + id.getName());
    }
});
```

## Option 2: A common tag plus custom `HierarchicalNameMapper`

```java
StatsdMeterRegistry registry = StatsdMeterRegistry.builder(configWithFlavor(StatsdFlavor.ETSY))
        .clock(clock)
        .nameMapper((id, convention) -> id.getTag("app") + id.getName() + ...)
        .build();

registry.config().commonTags("app", "myapp");
```

PR #1835, opened 2020-01-31
Kafka binder without JMX

### LINKED ISSUE #1095: Add support for KafkaProducerMetrics (similar to KafkaConsumerMetrics)

### LINKED ISSUE #1096: Add support for KafkaStreamsMetrics (similar to KafkaConsumerMetrics)

jeqo:
Support for Clients (Consumer/Producer/Admin APIs) and Streams via `metrics` API, instead of JMX-based metrics.

There are 3 different interfaces to gather metrics from Kafka applications:
- JMX exported metrics
- Metrics Reporter
- Metrics API

JMX exported metrics have been used as a first attempt to gather metrics, but there some downsides to consider: 
- Instrumentation needs to know set of metrics to gather before hand, therefore maintenance become very cumbersome. Set of metrics changes over releases, making this even more difficult.

Metrics Reporter is another option, which require to load an implementation of MetricsReporter when configuring Kafka clients. Downsides (?):
- Requires adding class name to Kafka configurations, which differs from current micrometer's user experience (compared with Metrics API)

Metrics API (proposed by this PR) uses `KafkaClient#metrics()` method to gather current metrics available in a client instance.

Resolves #1095
Resolves #1096


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+                }
+            });
+        }
+    }
+
+    private TimeGauge registerTimeGauge(MeterRegistry registry, Metric metric, String metricName,
+                                        Iterable<Tag> extraTags) {
+        return TimeGauge.builder(
+                metricName, metric, TimeUnit.SECONDS, m -> {
+                    checkAndRegisterMetrics(registry);
```
ggonchar: This recursive logic on each meter register call is not so intuitive to me,  can you please add some reasoning comment

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+                    if (m.metricValue() instanceof Double) {
```
ggonchar: Minor but I find this if-statement more readable when extracted and reused

jeqo: abstracting this in a single method.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
jeqo: adding more comments trying to make this clear, let me know if it looks better now.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+        Map<MetricName, ? extends Metric> metrics = metricsSupplier.get();
+        //Only happens first time number of metrics change
+        if (currentSize.get() != metrics.size()) {
+            currentSize.set(metrics.size());
+            Map<String, Set<Meter>> boundMeters = new HashMap<>();
+            //Register meters
+            metrics.forEach((name, metric) -> {
+                //Filter out metrics from group "app-info", that includes metadata
+                if (METRIC_GROUP_APP_INFO.equals(name.group())) {
+                    currentSize.incrementAndGet();
```
shakuzen: Why is the `currentSize` incremented here? Isn't the metric already accounted for in the size of the Map (L177 `currentSize.set(metrics.size());`)?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+                }
+                Meter meter = bindMeter(registry, metric);
+                //Collect metrics with same name to validate number of labels
+                Set<Meter> meters = boundMeters.get(metric.metricName().name());
+                if (meters == null) meters = new HashSet<>();
+                meters.add(meter);
+                boundMeters.put(metric.metricName().name(), meters);
+            });
+
+            //Remove meters with lower number of tags
```
shakuzen: Could we prevent both from being registered at the same time? That is, remove the meters with the lower number of tags before adding the meter with more tags? As it is, there is a race condition where meters could be retrieved (published/polled/scraped) with more than one meter registered with the same name but different tags.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+            .description(metric.metricName().description())
+            .register(registry);
+    }
+
+    private ToDoubleFunction<Metric> toMetricValue(MeterRegistry registry) {
+        return metric -> {
+            //Double-check if new metrics are registered; if not (common scenario)
+            //it only adds metrics count validation
+            checkAndBindMetrics(registry);
+            if (metric.metricValue() instanceof Double) {
```
shakuzen: Do we know what other kinds of values are expected from Kafka metrics? If it were say a String, us returning `0.0` could be confusing. I wonder if `NaN` is better when the value isn't a double.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIT.java
```diff
@@ -0,0 +1,98 @@
...
+
+        int producerAndConsumerMetrics = registry.getMeters().size();
+        assertTrue(producerAndConsumerMetrics > producerMetrics);
+
+        String topic = "test";
+        producer.send(new ProducerRecord<>(topic, "key", "value"));
+        producer.flush();
+
+        registry.getMeters()
+                .forEach(meter -> System.out.println(meter.getId() + " => " + meter.measure()));
```
shakuzen: This is nice for checking registered metrics and values when running the test locally. 👍

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+import java.util.function.Supplier;
+import java.util.function.ToDoubleFunction;
+import java.util.stream.Collectors;
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.streams.KafkaStreams;
+import org.jetbrains.annotations.NotNull;
```
shakuzen: Should be `io.micrometer.core.lang.NonNull` instead. But actually, all of the return values are considered `@NonNull` due to the `@NonNullApi` annotation. We should add a checkstyle rule to ban the jetbrains package import because this isn't the first time things have gotten mixed up.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+        return metric.metricName().tags()
+            .entrySet()
+            .stream()
+            .map(entry -> Tag.of(entry.getKey(), entry.getValue()))
+            .collect(Collectors.toList());
+    }
+
+    private String metricName(Metric metric) {
+        String value =
+            METRIC_NAME_PREFIX + metric.metricName().group() + "." + metric.metricName().name();
```
shakuzen: I didn't check which part it is coming from but I noticed `metrics` is included in every metric name. This feels superfluous and just makes the name longer. What do you think about getting rid of the "metrics" part of the name. Unless of course there are some kafka metrics about metrics. Or if Kafka users would expect the name with the `metrics` part in it, then maybe it is better to keep it. What do you think?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+     * Gather metrics from Kafka metrics API and register Meters.
+     * <p>
+     * As this is a one-off execution when binding a Kafka client, Meters include a call to this
+     * validation to double-check new metrics when returning values. This should only add the cost of
+     * validating meters registered counter when no new meters are present.
+     */
+    void checkAndBindMetrics(MeterRegistry registry) {
+        Map<MetricName, ? extends Metric> metrics = metricsSupplier.get();
+        //Only happens first time number of metrics change
+        if (currentSize.get() != metrics.size()) {
```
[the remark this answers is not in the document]
jeqo: That's true. I was only considering metrics number is increasing over time, under the assumption that we have only one client (producer/consumer/streams) per `KafkaMetrics` instance, and clients will send/poll an increasing number of topics.

Another option would be to keep a set of `MetricName`s and compare both sets instead of just the number; maybe a bit more expensive but more correct as well.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+            currentSize.set(metrics.size());
+            Map<String, Set<Meter>> boundMeters = new HashMap<>();
+            //Register meters
+            metrics.forEach((name, metric) -> {
+                //Filter out metrics from group "app-info", that includes metadata
```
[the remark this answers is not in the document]
jeqo: Found this KIP related to adding version and commit-id as metrics: https://cwiki.apache.org/confluence/display/KAFKA/KIP-188+-+Add+new+metrics+to+support+health+checks (#Client-side metrics)

And the implementation side: 

https://github.com/apache/kafka/blob/191f2faae07b6608b0601dc2caf204b196a8fc47/clients/src/main/java/org/apache/kafka/common/utils/AppInfoParser.java#L89-L95

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
jeqo: Mistake. I messed up assuming that not adding this metrics, will affect comparison. Anyway, is fixed now :)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
jeqo: Agree. Fixed.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
jeqo: I was considering `NaN` before reading this <https://www.robustperception.io/get-thee-to-a-nannary> but it will depend on how Micrometer semantics work on this case.

After taking another look at Kafka source-code I found there are certain metrics that return generic values, where we can find integers. I'll add this case to this function.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
jeqo: I agree it looks much better without `-metrics` suffix in between. I don't think there is any real issue other than people used to or having dashboards that include metrics with this suffix included, as comes from JMX. If we're ok with prioritizing better naming than familiarity/similarity with JMX naming, I can update this.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
jeqo: You're right. It could be safer to remove all registered meters before starting a new round, and synchronize this block in order to avoid race conditions. Assuming there are no meters with different number of tags on the same call--my experience is that this diff happens _between_ calls, removing upfront meters will avoid the need for future filtering. I've added a refactor to experiment with this.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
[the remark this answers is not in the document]
shakuzen: Commit ID doesn't seem as useful unless snapshot versions are used, but maybe we could add the kafka client version as a tag to all the kafka metrics?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
shakuzen: I was also thinking about compatibility with the previous `KafkaConsumerMetrics` as Micrometer users may already have dashboards built around those and it would be best if the consumer metrics could continue to be used as is with the new implementation.

config/checkstyle/checkstyle.xml
```diff
@@ -13,7 +13,7 @@
 
         <!-- Imports -->
         <module name="IllegalImportCheck" >
-            <property name="illegalPkgs" value="com.google.common.(?!cache).*,org.apache.commons.text.*,org.slf4j.*"/>
+            <property name="illegalPkgs" value="com.google.common.(?!cache).*,org.apache.commons.text.*,org.slf4j.*,org.jetbrains.*"/>
```
shakuzen: Thanks for this. I'll commit this change to the maintenance branches and merge into master so it has a separate audit trail from this pull request.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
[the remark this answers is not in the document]
jeqo: Yes. I'll add `version` as tag, and `startTimeMs` as meter. Sounds good?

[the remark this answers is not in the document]
shakuzen: `version` as a tag name can be problematic because it is likely to be used by a user as a common tag (possibly referring to the version of the application). Maybe a separate metric for the kafka client version is better, after all.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
shakuzen: The advice in there seems to be do not register a metric if it does not have a numeric value. That seems like good advice. Let's not register metrics with non-numeric values. Then, assuming the value type cannot change for a given metric, we don't need to handle the non-numeric case in the `ToDoubleFunction`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,265 @@
...
+    public KafkaMetrics(KafkaStreams kafkaStreams, Iterable<Tag> tags) {
+        this(kafkaStreams::metrics, tags);
+    }
+
+    /**
+     * {@link KafkaStreams} metrics binder
+     *
+     * @param kafkaStreams instance to be instrumented
+     */
+    public KafkaMetrics(KafkaStreams kafkaStreams) {
```
shakuzen: I think we'll have to split this out to another class. Compilation will fail when using `KafkaMetrics` without `kafka-streams` on the compilation classpath.

jeqo:
I'd like to add that I do agree an approach like https://github.com/micrometer-metrics/micrometer/pull/1173 could be complementary to export metrics when there are off-the-shelf components like Kafka Connectors, and we can pass this instrumentation via config.

jeqo:
I've removed the `TimeGauge` for rates, as API doesn't allow to get TimeUnit.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaMetricsTest.java
```diff
@@ -0,0 +1,155 @@
...
+        assertThat(registry.getMeters()).hasSize(1);
+        assertThat(registry.getMeters().get(0).getId().getTags()).hasSize(2);
+    }
+
+    @Test void shouldRemoveMeterWithLessTags() {
+        //Given
+        Supplier<Map<MetricName, ? extends Metric>> supplier = () -> {
+            Map<String, String> tags = new LinkedHashMap<>();
+            Map<MetricName, KafkaMetric> metrics = new LinkedHashMap<>();
+            MetricName metricName = new MetricName("a", "b", "c", tags);
```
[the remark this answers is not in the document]
jeqo: thanks. I've rephrase the test to fix this.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,191 @@
...
+    private final Supplier<Map<MetricName, ? extends Metric>> metricsSupplier;
+    private final Iterable<Tag> extraTags;
+    private final int extraTagsSize;
+
+    /**
+     * Keeps track of current set of metrics. When this values change, metrics are bound again.
+     */
+    private volatile Map<MetricName, Meter> currentMeters = new HashMap<>();
+
+    private String kafkaVersion = "unknown";
```
jeqo: @shakuzen is this better than empty string?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -0,0 +1,191 @@
...
+     * validation to double-check new metrics when returning values. This should only add the cost of
+     * validating meters registered counter when no new meters are present.
+     */
+    void checkAndBindMetrics(MeterRegistry registry) {
+        Map<MetricName, ? extends Metric> metrics = metricsSupplier.get();
+        if (!currentMeters.keySet().equals(metrics.keySet())) {
+            synchronized (this) { //Enforce only happens once when metrics change
+                if (!currentMeters.keySet().equals(metrics.keySet())) {
+                    //Clean registry
+                    for (Meter meter : currentMeters.values()) registry.remove(meter);
```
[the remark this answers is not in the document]
jeqo: True. I've removed it as this was used when I wasn't using `registry` but `currentMeters` map. I've added a conditional to check that meters are not re-registered.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaMetricsTest.java
```diff
@@ -0,0 +1,155 @@
...
+            return metrics;
+        };
+        KafkaMetrics kafkaMetrics = new KafkaMetrics(supplier);
+        MeterRegistry registry = new SimpleMeterRegistry();
+        //When
+        kafkaMetrics.checkAndBindMetrics(registry);
+        //Then
+        assertThat(registry.getMeters()).hasSize(1);
+        //When
+        kafkaMetrics.checkAndBindMetrics(registry);
```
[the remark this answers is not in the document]
jeqo: True. I've fixed this but using a set instead of map and reference result from supplier directly instead of adding names at the end to compare complete results.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
shakuzen: Yes, I think so.

jeqo:
Sounds great! I've test it with this sample application <https://github.com/jeqo/tracing-kafka-apps/compare/metrics> and works as expected on all components (producer/consumer/streams).

jeqo:
e.g. producer metrics in Prometheus format:

```
# HELP kafka_producer_topic_record_send_rate The average number of records sent per second for a topic.
# TYPE kafka_producer_topic_record_send_rate gauge
kafka_producer_topic_record_send_rate{client_id="producer-1",kafka_version="2.4.0",topic="events-v1",} 0.0
# HELP kafka_producer_batch_split_rate The average number of batch splits per second
# TYPE kafka_producer_batch_split_rate gauge
kafka_producer_batch_split_rate{client_id="producer-1",kafka_version="2.4.0",} 0.0
# HELP kafka_producer_incoming_byte_rate The number of bytes read off all sockets per second
# TYPE kafka_producer_incoming_byte_rate gauge
kafka_producer_incoming_byte_rate{client_id="producer-1",kafka_version="2.4.0",} 2.696326255476913
# HELP kafka_producer_request_total The total number of requests sent
# TYPE kafka_producer_request_total counter
kafka_producer_request_total{client_id="producer-1",kafka_version="2.4.0",} 452.0
# HELP kafka_producer_batch_split_total The total number of batch splits
# TYPE kafka_producer_batch_split_total counter
kafka_producer_batch_split_total{client_id="producer-1",kafka_version="2.4.0",} 0.0
# HELP kafka_producer_request_latency_avg The average request latency in ms
# TYPE kafka_producer_request_latency_avg gauge
kafka_producer_request_latency_avg{client_id="producer-1",kafka_version="2.4.0",} NaN
# HELP kafka_producer_node_request_latency_max 
# TYPE kafka_producer_node_request_latency_max gauge
kafka_producer_node_request_latency_max{client_id="producer-1",kafka_version="2.4.0",node_id="node--1",} NaN
```

saltos:
Will you skip the group `kafka-metrics-count`? This group contains the only one counter - amount of the counters. I am sure `kafka-metrics-count.count` is the only counter which has the same name for a producer and consumer but different amount of tags. Maybe you can remove the double-check for new metrics completely. I've tested your code without double-checking and it works for me.

jeqo:
@saltos thanks for your review. I'm ok removing `kafka-metrics-count` as could be of little/none use (e.g. comparing number of metrics between versions maybe?) but I don't think this removes the need for double checking as metrics with different tags would happen between clients state transition (e.g. producer before/after sending messages to a topic will have a metric without/with topic tag depending on the state). wdyt?

saltos:
@jeqo thank you for the reply. I've add some extra test for my code... You are right.
My producer has no metrics with the `topic` tag before sending any message so there is no conflict with the existing metric with different amount of tags, but a consumer has several metrics with different amount of tags. There is a metric with the `client-id` tag and with the `client-id`, `topic`, `partition` tags at the same time... So you code will be truly useful.

jeqo:
@shakuzen do you have an opinion about removing `kafka-metrics-count` group? Currently meter name is `kafka.kafka.count.count` which seems useless. 

I'm between adding an special path to name this counter properly (e.g. `kafka.metrics.count`), or filtering out the group altogether (1 counter only)--easiest path.

Testing consumer/producer (IT) with the same registry lead to 2 meters with the same tags: `client-id` and `kafka-version`, which looks ok. If eventually there is a tags disparity between consumers/producers only the one with more tags will persist. @saltos could you confirm if this is the case on your side?--if it is then removing the group would be more consistent than exposing counter for only one client.

shakuzen:
> @shakuzen do you have an opinion about removing `kafka-metrics-count` group? Currently meter name is `kafka.kafka.count.count` which seems useless.

I would be inclined to go with filtering it out unless a Kafka user can make a case for why they would want that metric specifically. It would not break anyone to filter it out initially an undo that decision, but it might be breaking for someone if we decide to filter it later.

shakuzen:
@jeqo Thank you so much for your perseverance on getting this to the state it is now and testing. Supporting Kafka producer metrics and stream metrics have been highly popular feature requests, and with this we will finally have them. Great work!

saltos:
>@saltos could you confirm if this is the case on your side?
Yes, as a result I have only one metric with `client-id` and `listener` tags (from a consumer).

saltos:
I found some odd behaviour with a consumer metrics.
I have 3 listeners but the only one `bytes_consumed_total` metric.

> kafka_consumer_fetch_manager_bytes_consumed_total{client_id="prototype-0",group="kafka-spring",kafka_version="2.3.1",listener="foo-dlt-listener",topic="foo_DLT",} 0.0

Function `checkAndBindMetrics` has the code that can lead to a loss of some metrics when new metric has the same tags with different values.
```java
Collection<Meter> meters = registry.find(metricName(metric)).meters();
for (Meter meter : meters) {
    if (meter.getId().getTags().size() < (metricTags(metric).size() + extraTagsSize))
        registry.remove(meter);
    // Check if already exists
    else if (meter.getId().getTags().equals(metricTags(metric))) return;
    else hasLessTags = true;
}
```
The condition `meter.getId().getTags().equals(metricTags(metric))` will be `false` because `metricTags(metric)` doesn't contains the extra tags.
I changed this condition to:
```java
for (Meter meter : meters) {
    List<Tag> meterTags = meter.getId().getTags();
    List<Tag> metricTags = metricTags(metric);
    metricTags.addAll(tags);

    if (meterTags.size() < metricTags.size()) {
        registry.remove(meter);
    } else if (meterTags.size() == metricTags.size()) {
        if (meter.getId().getTags().equals(metricTags)) {
            // do not register the metric with the same tags
            return;
        } else {
            // create the new metric with different values of the tags
            break;
        }
    } else {
        hasLessTags = true;
    }
}
```
and got all 3 metrics.

> kafka_consumer_fetch_manager_bytes_consumed_total{client_id="prototype-0",group="kafka-spring",kafka_version="2.3.1",listener="bar-listener",topic="bar",} 1113.0
> kafka_consumer_fetch_manager_bytes_consumed_total{client_id="prototype-0",group="kafka-spring",kafka_version="2.3.1",listener="foo-listener",topic="foo",} 2268.0
> kafka_consumer_fetch_manager_bytes_consumed_total{client_id="prototype-0",group="kafka-spring",kafka_version="2.3.1",listener="foo-dlt-listener",topic="foo_DLT",} 0.0

shakuzen:
@saltos Thank you for testing this out and reporting back your findings - it's great to find these kinds of things before the release. Would you please make a new issue to track this? If you would be interested, also feel free to make a pull request with your mentioned changes. Ideally we would want to add a test case that covers the scenario you found. Thanks again for the help!

saltos:
@shakuzen Ok, I will make an issue and try to make a pull request with the test, that's my pleasure. You are doing a really remarkable job.

jeqo:
just realized I was creating the PR while this discussion was ongoing, sorry if I've interrupted your work on this @saltos 

#1896 should solve this, while removing some boilerplate code, take a look at it please.

saltos:
@jeqo Awesome, great job! This solves the problem.

PR #1841, opened 2020-02-05
 Fix Log4j2 metrics recording duplicates when non-additive loggers share configuration

rj-jemmic:
**Log4j2 metrics - Fix logger configuration containing duplicates of metrics filter when multiple non-additive loggers share logger configuration**

**Issue:**
When multiple non-additive loggers share the same logging configuration in the Log4j2 configuration and Log4j2 metrics are enabled, the filter that records these metrics is added multiple times to the shared logging configuration. This causes duplicate recordings of log events (the counters per log level will be incremented multiple times for a single log event).

**Changes:**
Changed binding in Log4j2Metrics to iterate over logger configurations instead of each logger, and added checks to not add the MetricsFilter when it is already present in the logging configuration.

**Tests:**
Added minimal unit test for this change

This is my first pull request so am open to suggestions & remarks


rj-jemmic:
Please close as posted from wrong account

PR #1847, opened 2020-02-11
Apply the accept filter to the mappedId instead of originId

dlemaigl:
This PR change the Id passed to accept filter to mapped Id (id after mapped filters are applied) instead of original id.


jkschneider:
Merged with d1a13b32427c274342fc2974e16f3d933cc67698

Issue #1852, opened 2020-02-13
Build fails because of JOOQ 3.13.0


schmidt-galen-heb:
[JOOQ](https://github.com/jOOQ/jOOQ) released version [3.13.0 on February 12th 2020](https://github.com/jOOQ/jOOQ/releases/tag/version-3.13.0).

Micrometer is incompatible with this release, and because it pins the version to [latest.release](https://github.com/micrometer-metrics/micrometer/blob/f6a2687f13f2b9fd7351ce9126a9f04238da79dd/dependencies.gradle#L55), this has broken the build:

```
$ ./gradlew clean compileJava
... snip ...
> Task :micrometer-core:compileJava FAILED
/Users/0000000/git/micrometer-metrics/micrometer/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/MetricsDSLContext.java:63: error: MetricsDSLContext is not abstract and does not override abstract method dropTemporaryTableIfExists(Table<?>) in DSLContext
public class MetricsDSLContext implements DSLContext {
       ^
... snip ...

20 errors
47 warnings

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':micrometer-core:compileJava'.
> Compilation failed; see the compiler error output for details.
```

schmidt-galen-heb:
I've opened https://github.com/micrometer-metrics/micrometer/pull/1853 to pin the version to 3.12.x, which fixes the build. I'd propose doing that as a breakfix, and then opening another issue to add support for JOOQ 3.13+

PR #1854, opened 2020-02-14
Enable NewRelicMeterRegistry to publish via APM in addition to API (resolves #1761)

### LINKED ISSUE #1761: Enhanced support for NewRelic APM
Howdy! I'm using NewRelic APM to monitor my Springboot applications, and am investigating reporting custom MicroMeter metrics to NewRelic Insights.

This was previously implemented in feature request https://github.com/micrometer-metrics/micrometer/issues/122.
Two potential integration paths were investigated:
1. Indirect integration via the [NewRelic Java Agent](https://docs.newrelic.com/docs/agents/java-agent/api-guides/guide-using-java-agent-api)
1. Direct integration via the [Insights REST API](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/introduction-event-api)

Ultimately, option (2) was chosen because:
1. Integration via the agent required the use of [APM custom metrics](https://docs.newrelic.com/docs/agents/manage-apm-agents/agent-data/collect-custom-metrics), which are less powerful than Insights metrics
1. The agent must be installed as a [Java agent](https://docs.newrelic.com/docs/agents/java-agent/installation/include-java-agent-jvm-argument) on the application JVM, which requires customizing the JVM startup arguments as well as providing an agent configuration file

In my case however, we've already integrated the APM agent, and option (1) requires use of an [Insights API key](https://docs.newrelic.com/docs/apis/get-started/intro-apis/types-new-relic-api-keys#event-insert-key), as well as [specifying the NewRelic account ID](https://github.com/micrometer-metrics/micrometer/blob/114d62be80c51768031a0672b021bb30a917c615/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicConfig.java#L66), neither of which are required by the agent (and thus we aren't currently providing them).

The agent API allows programmatic access to the Insights API via [NewRelic.getAgent().getInsights()](https://docs.newrelic.com/docs/agents/java-agent/api-guides/guide-using-java-agent-api#obtain-references), and metrics can then be recorded via the [Insights.recordCustomEvent(...)](https://newrelic.github.io/java-agent-api/javadoc/index.html?com/newrelic/api/agent/Insights.html) method. NewRelic's guide on this is [here](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/insert-custom-events-new-relic-apm-agents).

Note: the [current implementation](https://github.com/micrometer-metrics/micrometer/blob/114d62be80c51768031a0672b021bb30a917c615/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java#L111) also records metrics as a [custom event](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/introduction-event-api#submit-event) internally, but does so using the insights API directly.

In addition to the existing `NewRelicMeterRegistry`, I'd like to have the option of using a class (called something like `NewRelicApmMeterRegistry`) which utilized the APM agent's Java API to submit metrics. Since submitting custom events via the agent does not appear to require any additional configuration, this is a much simpler integration path for consumers who are already utilizing the NewRelic APM.

(additional note: I've looked briefly at [newrelic/micrometer-registry-newrelic](https://github.com/newrelic/micrometer-registry-newrelic), but it also appears to require an insights API key, which is what I'm trying to avoid for applications already integrated with the APM).

schmidt-galen-heb:
https://github.com/micrometer-metrics/micrometer/issues/1761

I've done extensive testing of this with some of my own applications, and everything appears to be working correctly.

The scope of the changes are fairly large, but I've been very careful to ensure that all of the changes should be backwards compatible for existing clients.

Specifically, all existing configurations will continue working as before, and none of the generated JSON payloads to the Insights API should change.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -17,6 +17,7 @@ def VERSIONS = [
         // <=1.5.11 or 1.5.18 doesn't break with Micrometer, but open metrics won't be correct necessarily.
         'com.netflix.hystrix:hystrix-core:1.5.12',
         'com.netflix.spectator:spectator-reg-atlas:latest.release',
+        'com.newrelic.agent.java:newrelic-api:latest.release',
         'com.signalfx.public:signalfx-java:latest.release',
         'com.squareup.okhttp3:okhttp:latest.release',
         'com.wavefront:wavefront-sdk-java:latest.release',
@@ -52,7 +53,8 @@ def VERSIONS = [
         'org.hdrhistogram:HdrHistogram:latest.release',
         'org.hibernate:hibernate-entitymanager:latest.release',
         'org.hsqldb:hsqldb:latest.release',
-        'org.jooq:jooq:latest.release',
+        // Pin version temporarily to restore builds. See gh-1852
+        'org.jooq:jooq:3.12.+',
         'org.jsr107.ri:cache-ri-impl:1.0.0',
         'org.junit.jupiter:junit-jupiter:latest.release',
         'org.junit.platform:junit-platform-launcher:latest.release',
--- a/implementations/micrometer-registry-new-relic/build.gradle
+++ b/implementations/micrometer-registry-new-relic/build.gradle
@@ -2,6 +2,11 @@ dependencies {
     api project(':micrometer-core')
 
     implementation 'org.slf4j:slf4j-api'
+    implementation 'com.newrelic.agent.java:newrelic-api'
 
     testImplementation project(':micrometer-test')
+
+    testImplementation 'ch.qos.logback:logback-classic'
+    testImplementation 'com.fasterxml.jackson.core:jackson-databind'
+    testImplementation 'com.google.guava:guava'
 }
--- a/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicConfig.java
+++ b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicConfig.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017 Pivotal Software, Inc.
+ * Copyright 2019 Pivotal Software, Inc.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,8 +15,8 @@
  */
 package io.micrometer.newrelic;
 
-import io.micrometer.core.instrument.config.MissingRequiredConfigurationException;
 import io.micrometer.core.instrument.step.StepRegistryConfig;
+import io.micrometer.core.lang.Nullable;
 
 /**
  * Configuration for {@link NewRelicMeterRegistry}.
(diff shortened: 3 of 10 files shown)


dependencies.gradle
```diff
@@ -52,7 +53,8 @@ def VERSIONS = [
         'org.hdrhistogram:HdrHistogram:latest.release',
         'org.hibernate:hibernate-entitymanager:latest.release',
         'org.hsqldb:hsqldb:latest.release',
-        'org.jooq:jooq:latest.release',
+        // Pin version temporarily to restore builds. See gh-1852
```
schmidt-galen-heb: Included so the build will pass, can be removed once https://github.com/micrometer-metrics/micrometer/issues/1852 is resolved

implementations/micrometer-registry-new-relic/build.gradle
```diff
@@ -2,6 +2,11 @@ dependencies {
     api project(':micrometer-core')
 
     implementation 'org.slf4j:slf4j-api'
+    implementation 'com.newrelic.agent.java:newrelic-api'
```
schmidt-galen-heb: Adding APM support requires the addition of a compile time dependency on the newrelic-api.
The API is fairly light, containing only annotations, interfaces, POJOs, and enums, so I don't expect this to cause any issues for consumers not using the APM integration

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicConfig.java
```diff
@@ -32,46 +32,75 @@ default String prefix() {
...
 
+    /**
+     * The {@link NewRelicIntegration} method to use for reporting metric data.
+     * <br><br>
+     * Defaults to {@link NewRelicIntegration#API}.
+     *
+     * @return The NewRelic integration method
+     */
+    default NewRelicIntegration integration() {
+        return NewRelicIntegration.fromString(get(prefix() + ".integration")).orElse(NewRelicIntegration.API);
```
schmidt-galen-heb: Default to API integration to preserve existing behavior

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
```diff
@@ -315,26 +412,78 @@ public Builder httpClient(HttpSender httpClient) {
...
     }
 
-    private class Attribute {
-        private final String name;
-        private final Object value;
+    /**
+     * POJO representation of an Insights event.
+     */
+    // VisibleForTesting
+    static class NewRelicEvent {
```
schmidt-galen-heb: To allow the APM/API integrations to share the majority of the code, a DTO to represent the event was created

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicNamingConvention.java
```diff
@@ -45,16 +47,16 @@ public NewRelicNamingConvention(NamingConvention delegate) {
 
     @Override
     public String name(String name, Meter.Type type, @Nullable String baseUnit) {
-        return StringEscapeUtils.escapeJson(toValidNewRelicString(delegate.name(name, type, baseUnit)));
+        return toValidNewRelicString(delegate.name(name, type, baseUnit));
```
schmidt-galen-heb: With the addition of the APM integration, there was a potential issue where attributes would either be double JSON escaped, or never JSON escaped.
Removing the JSON escaping from the `NamingConvention` and only performing it when the data is actually converted into JSON prevents this from happening.

implementations/micrometer-registry-new-relic/src/test/java/io/micrometer/newrelic/NewRelicMeterRegistryTest.java
```diff
@@ -15,372 +15,903 @@
...
+import static org.mockito.Mockito.verifyNoInteractions;
 
 /**
  * Tests for {@link NewRelicMeterRegistry}.
  *
  * @author Johnny Lim
+ * @author Galen Schmidt
  */
+@SuppressWarnings("ConstantConditions")
 class NewRelicMeterRegistryTest {
```
schmidt-galen-heb: There were previously several code paths which weren't tested by this class, including conversion of various types of Meters into JSON, as well as ensuring global attributes were included.

I've ensured there's now a test for each subclass of `Meter`, as well as tests to ensure the global tags are applied.

implementations/micrometer-registry-new-relic/src/test/java/io/micrometer/newrelic/NewRelicMeterRegistryTest.java
```diff
@@ -15,372 +15,903 @@
...
+
     private final NewRelicMeterRegistry meterNameEventTypeEnabledRegistry = new NewRelicMeterRegistry(meterNameEventTypeEnabledConfig, clock);
-    private final NewRelicMeterRegistry registry = new NewRelicMeterRegistry(config, clock);
+
+    private final NewRelicMeterRegistry meterNameEventTypeDisabledRegistry = new NewRelicMeterRegistry(meterNameEventTypeDisabledConfig, clock);
+
+    @Test
+    void eventFor_Gauge() {
+        eventFor_Gauge(this.meterNameEventTypeEnabledRegistry, ImmutableMap.<String, Object>builder()
+                .put("eventType", "myGauge")
```
schmidt-galen-heb: The previous test did a simple String comparison between the generated JSON and the expected value. This broke with the addition of `NewRelicEvent`, because the attributes were now stored in a Map, meaning the keys were written in an non-deterministic order in the JSON.

I've refactored the tests to use Jackson to read the JSON object into a Map, and then use AssertJ's map comparator instead. That way, the tests only care that the JSON has the correct keys, values, and types, making the tests easier to write + less prone to false failures.

PR #1858, opened 2020-02-18
Performance enhancement for logback's isDebugEnabled (et al) calls.

perhuss:
During profiling of an application running on Spring Boot with Hibernate, we noted that 4% of cpu was used in ThreadLocal.get(), where most of the calls came through micrometer on calls to isTraceEnabled() and isDebugEnabled(). Since the calls were only to test if logging would happen, (with format parameter set to null), counters would never be incremented for these calls. By moving the test on format parameter first, the much more expensive thread local call can be avoided, without putting a penalty on other usages. Fix was verified in our setup.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
@@ -130,7 +130,9 @@ class MetricsTurboFilter extends TurboFilter {
 
     @Override
     public FilterReply decide(Marker marker, Logger logger, Level level, String format, Object[] params, Throwable t) {
-        // If format is null we can short circuit
+        // When filter is asked for decision for an isDebugEnabled call or similar test, there is no message (ie format) 
+        // and no intention to log anything with this call. We will not increment counters and can return immediately and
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
```diff
@@ -130,13 +130,18 @@ public void close() {
 
     @Override
     public FilterReply decide(Marker marker, Logger logger, Level level, String format, Object[] params, Throwable t) {
+        // If format is null we can short circuit
```
checketts: Can you add to the comment _why_ the format is null? Based on your PR description it sounds like format is null when doing an `isDebugEnabled` check. That would be useful info here in the code

checketts:
Thanks for your contribution and investigation!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
perhuss: Thanks for feedback; updated with more elaborate comment!

Issue #1056, opened 2018-12-05
Support Datadog distributions via DogStatsD


jkschneider:
https://docs.datadoghq.com/developers/metrics/distributions/

shakuzen:
The page has a note at the top:

> This feature is in beta. Contact Datadog support to enable distribution metrics for your account.

I suppose we could still implement it as an incubating feature.

shakuzen:
#1859 is open to add support for Datadog distributions via DogStatsD. There's still no documented API to send distributions via the HTTP API. I'm not sure it is a good match anyways, since we would have to decide how to buffer/send the per event distribution data separate from the rest of the metrics for the DatadogMeterRegistry.

bxacsq:
Is there any update about this feature? Current our services send metrics to Datadog, and the lack of the datadog distribution support in micrometer makes it inconvenient to get the latency stats of the online services.

vsingal-p:
Is there any update on this? Distributions would lower the cost of custom metrics. It's very important for many teams

shakuzen:
Support has been merged for this with `DistributionSummary` type meters in #2745 and `Timer` type meters in #2775. If the summary is configured with the `percentileHistogram` enabled, the metric will be published to dogstatsd as a distribution type metric. This will be available in the upcoming 1.8.0-M3 milestone release. Please try it out and let us know if there is any feedback.

PR #1859, opened 2020-02-18
Support Datadog distributions via DogStatsD

gtrog:
First attempt at adding support for recording histogram values as DataDog distributions in StatsD.

Pretty straightforward, this will add an additional interface for configuring DataDog-specific StatsD options with the option to override the `publishDistributions()` option.  It will default to `false` if not overridden, using the default behavior.

More information about DataDog distributions:
https://docs.datadoghq.com/metrics/distributions/

Source code for DataDog StatsD client where the distribution type is used:
https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/NonBlockingStatsDClient.java#L1121

Related to #1056

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -299,7 +299,7 @@ public class StatsdMeterRegistry extends MeterRegistry {
             lineBuilderFunction = id2 -> {
                 switch (statsdConfig.flavor()) {
                     case DATADOG:
-                        return new DatadogStatsdLineBuilder(id2, config());
+                        return new DatadogStatsdLineBuilder(id2, statsdConfig, config());
                     case TELEGRAF:
                         return new TelegrafStatsdLineBuilder(id2, config());
                     case SYSDIG:
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/datadog/DataDogStatsdConfig.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/datadog/DataDogStatsdConfig.java
@@ -1,0 +1,33 @@
+/**
+ * Copyright 2017 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.statsd.datadog;
+
+import io.micrometer.statsd.StatsdConfig;
+
+/**
+ * StatsD configuration with specific DataDog options
+ */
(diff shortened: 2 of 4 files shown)


gtrog:
Tested against DataDog and verified that metrics are coming in and showing up in the "Distribution Metrics" page

jkschneider:
@gtrog Is there a higher cost in datadog associated with shipping as `Distribution` rather than `Histogram`? Also, is there a compatibility issue that would prevent us from wanting to ship distributions at all?

### If higher cost, no compatibility issue

I'd suggest we remove the `DatadogStatsdConfig` and just respond to whether the associated `Timer`, `DistributionSummary`, or `LongTaskTimer` has percentile histograms enabled. In this way we can focus at a per-metric level the decision of whether or not to accept this extra cost (as opposed to on a registry level).

### If no higher cost, no compatibility issue

Is there any reason we shouldn't just _always_ ship distributions then? It appears to me that distribution [can do everything](https://docs.datadoghq.com/developers/metrics/types/?tab=distribution#metric-types) that histogram can, better.

#### If compatibility issue

_Then_ I understand the `DatadogStatsdConfig` thing, maybe. I suppose it would only be a problem for folks using old agents? How likely is it do we think that agents are so old out there that they would break?

shakuzen:
In addition to Jon's feedback above, thoughts I had when I looked at this were if we have a configuration option, rather than a subclass `DatadogStatsdConfig` which adds some complexity for users (we don't do this anywhere else), we could put it in `StatsdConfig` and have it do nothing for flavors that don't support distributions. In the future other flavors may add support for distributions.

gtrog:
Sure, I chose the subclass approach so it was clear that the behavior was only relevant to DataDog as to not cause confusion, but I understand this is a different approach from how other configs are handled.

According to https://docs.datadoghq.com/account_management/billing/custom_metrics/?tab=histogram#counting-custom-metrics, by default, histograms and distributions will generate the same number of custom metrics (which is how pricing is determined).  Once you add percentile aggregations, then the number of custom metrics doubles and so does the cost potentially.  Distributions also require extra configuration on the DataDog side to make them recognized and usable, so, it's not a simple switch from histogram to distribution.

In terms of compatibility issues, it's possible that older agents would likely drop the value altogether, since distributions are a fairly new addition.

jkschneider:
> Once you add percentile aggregations

@gtrog It's not clear to me how you add them? Is this something done in DD after the fact or as part of the agent configuration?

jkschneider:
> we could put it in StatsdConfig and have it do nothing for flavors that don't support distributions

@shakuzen I've noticed other monitoring system vendors actually adopting DogstatsD as their flavor in the past. So likely any other vendor adding distributions to their StatsD agents would accept the Datadog convention.

bxacsq:
Is there any update about this feature? Current our services send metrics to Datadog, and the lack of the datadog distribution  support in micrometer makes it inconvenient to get the latency stats of the online services.

pivotal-issuemaster:
@gtrog Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=1859)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=1859) to manually synchronize the status of this Pull Request.

hkawa608:
> It's not clear to me how you add them? Is this something done in DD after the fact or as part of the agent configuration?

https://docs.datadoghq.com/metrics/distributions/
According to this doc, `something done in DD after the fact` is the answer.
```
After electing to apply percentile aggregations on a distribution metric, these aggregations are automatically available in the graphing UI:
```

vsingal-p:
Do we have an ETA on this? Our team wants to use distributions with DogStasD

vsingal-p:
Is there any update when would this be merged and released?

shakuzen:
This has been superseded by #2745

PR #1863, opened 2020-02-25
Make ProcessorMetrics compatible with more J9 versions

eyalkoren:
I just tested this with a J9 1.8.0 JVM (that comes with the `websphere-liberty:8.5.5` image) and it seems that it contains the `com.sun.management.OperatingSystemMXBean` interface, but the actual MBean implements the IBM interface, which means this won't work. Reversing the order fixes it though.


jkschneider:
Added to 1.1.x with 101a09405c8b4b579f0e39db5343dc9e927c1c34

jkschneider:
Merged forward to 1.3.x with c02136855dc860387888519788babb27411ac236.

jkschneider:
Merged forward to 1.5.x with 905341b25775e0fcd29caa10a5eeaae17c4bde84.

jkschneider:
Merged forward to master with b486e1261cade606eb0ff826fbae318f6d0c9e86.

jkschneider:
Thanks for the fix @eyalkoren!

PR #1864, opened 2020-02-25
Improve perf by reducing the number of requests to Datadog

gzussa:
This PR improves performance for the `micrometer-registry-datadog` implementation.
Instead of sending metrics `unit` and `type` attributes as part of the unbatched edit metric endpoint `/api/v1/metrics/<METRIC_NAME>`, we now  submit them as part of the batched metric intake endpoint `/api/v1/series`.

Note that metric description cannot be submitted as part of the metric intake endpoint at the moment.
However, this attribute is optional  and can be disabled. If disabled the edit metric endpoint won't be call at all. 

Related issue #1867.

### DIFF
--- a/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
+++ b/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
@@ -242,10 +242,9 @@ public class DatadogMeterRegistry extends StepMeterRegistry {
                 .findAny()
                 .map(t -> ",\"host\":\"" + escapeJson(t.getValue()) + "\"")
                 .orElse("");
-
         // Create type attribute
         String type = ",\"type\":\"" + DatadogMetricMetadata.sanitizeType(statistic) + "\"";
-        // Create type attribute
+        // Create unit attribute
         String unit = "";
         if (id.getBaseUnit() != null || overrideBaseUnit != null) {
             unit = ",\"unit\":\"" + DatadogMetricMetadata.sanitizeBaseUnit(id.getBaseUnit(), overrideBaseUnit) + "\"";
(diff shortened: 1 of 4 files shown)


implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -222,47 +224,56 @@ private void addToMetadataList(Map<String, DatadogMetricMetadata> metadata, Mete
...
+        // Create host attribute
         String host = config.hostTag() == null ? "" : stream(tags.spliterator(), false)
                 .filter(t -> requireNonNull(config.hostTag()).equals(t.getKey()))
                 .findAny()
                 .map(t -> ",\"host\":\"" + escapeJson(t.getValue()) + "\"")
                 .orElse("");
 
+        // Create type attribute
+        String type = ",\"type\":\"" + DatadogMetricMetadata.sanitizeType(statistic) + "\"";
+        // Create type attribute
```
zippolyte:
```suggestion
        // Create unit attribute
```

shakuzen:
Just so I'm sure I understand the changes here

**Before**
- Sending the type, unit, description info in the metadata once per meter per application instance lifetime

**After**
- (if sending descriptions is enabled - default is enabled) Sending the description metadata once per meter per application instance lifetime
- Sending the unit and type info every time with the meter value

If that is an accurate description of the changes, it sounds like the overall data sent will be significantly more with these changes, since the meter data points will generally be published more than once in an application instance's lifetime, and they will now contain the unit and type info each time. The number of calls is only reduced if sending descriptions is disabled, and it is only reduced by 1 call per meter once, which is substantially less than the number of calls typically expected over the lifetime of an application to publish meter data.

That all said, I'm fine with the changes if this is the more canonical way Datadog expects to receive the given info about meters.

gzussa:
If description is enable, performances are not going to be impacted. We will still make the same number of requests. For instance, we will make a small number of request to send metric data in batch using our metric intake endpoint and we will make one request per metric name (aka `meter`) to the metadata endpoint (at least, per application lifetime).

Our metadata endpoint (aka metric metadata edit endpoint) is not supposed to be hit as often as our metric intake endpoint (processing time is also a tiny bit slower).

Some users are sending a lot of of metrics (distinct meters) using this library. As a consequence, the metadata endpoint is hit a lot and it creates in there case, lot of lag. They don't really care about the description. Fortunately enough, our intake endpoint support all metadata but the description (without any perf impact). So i made changes to submit all possible metadata as part of our metric intake endpoint rather than through the edit metric metadata endpoint. Users who care about performance will disable description to take advantage of this.

Of course, this type of performance improvement only make sense for users that defined a lot of meters. It won't changes much for small usage of this library.

shakuzen:
Thanks for the response @gzussa. This is good information to have here for reference.

PR #1865, opened 2020-02-25
Replace deprecated dependency configurations

izeye:
This PR replaces deprecated dependency configurations.

### DIFF
--- a/benchmarks/benchmarks-core/build.gradle
+++ b/benchmarks/benchmarks-core/build.gradle
@@ -8,6 +8,9 @@ dependencies {
     jmh 'io.dropwizard.metrics5:metrics-core:latest.release'
     jmh 'io.prometheus:simpleclient_common:latest.release'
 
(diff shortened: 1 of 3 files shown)


benchmarks/benchmarks-core/build.gradle
```diff
@@ -8,6 +8,9 @@ dependencies {
     jmh 'io.dropwizard.metrics5:metrics-core:latest.release'
     jmh 'io.prometheus:simpleclient_common:latest.release'
 
+    jmh 'io.dropwizard.metrics:metrics-core'
+    jmh 'com.google.guava:guava'
```
checketts: I'm guessing these were transitive deps before and with the `implementation` change it exposed that?

izeye: @checketts Yes, that's the reason why I added them explicitly as they are not available with the changes any more.

PR #1868, opened 2020-02-26
Provided support for Hazelcast 4.0

### LINKED ISSUE #1697: Support Hazelcast 4
The [Hazelcast 4 BETA-1 has been released](https://hazelcast.com/blog/hazelcast-imdg-4-0-beta/), and our current support fails to even compile against it. See #1696. We will need to consider how to best support Hazelcast 3/4 going forward.

alparslanavci:
Fixes #1697.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetricsTest.java
@@ -16,11 +16,11 @@
 package io.micrometer.core.instrument.binder.cache;
 
 import com.hazelcast.core.Hazelcast;
-import com.hazelcast.core.IMap;
-import com.hazelcast.monitor.LocalMapStats;
-import com.hazelcast.monitor.NearCacheStats;
-import com.hazelcast.monitor.impl.LocalMapStatsImpl;
-import com.hazelcast.monitor.impl.NearCacheStatsImpl;
+import com.hazelcast.map.IMap;
+import com.hazelcast.map.LocalMapStats;
+import com.hazelcast.nearcache.NearCacheStats;
+import com.hazelcast.internal.monitor.impl.LocalMapStatsImpl;
+import com.hazelcast.internal.monitor.impl.NearCacheStatsImpl;
 
 import io.micrometer.core.instrument.FunctionCounter;
 import io.micrometer.core.instrument.FunctionTimer;
@@ -90,7 +90,7 @@ class HazelcastCacheMetricsTest extends AbstractCacheMetricsTest {
         assertThat(nearEvictions.value()).isEqualTo(nearCacheStats.getEvictions());
 
         // timings
-        TimeUnit timeUnit = TimeUnit.NANOSECONDS;
+        TimeUnit timeUnit = TimeUnit.MILLISECONDS;
         FunctionTimer getsLatency = fetch(meterRegistry, "cache.gets.latency").functionTimer();
         assertThat(getsLatency.count()).isEqualTo(localMapStats.getGetOperationCount());
         assertThat(getsLatency.totalTime(timeUnit)).isEqualTo(localMapStats.getTotalGetLatency());
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -9,7 +9,7 @@ def VERSIONS = [
         'com.google.dagger:dagger-compiler:2.11',
         'com.google.guava:guava:latest.release',
         'com.google.inject:guice:4.1.0',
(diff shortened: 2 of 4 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetricsTest.java
```diff
@@ -90,7 +90,7 @@ void reportMetrics() {
         assertThat(nearEvictions.value()).isEqualTo(nearCacheStats.getEvictions());
 
         // timings
-        TimeUnit timeUnit = TimeUnit.NANOSECONDS;
+        TimeUnit timeUnit = TimeUnit.MILLISECONDS;
```
checketts: Thanks for the contribution! Can you explain why this change is needed? (I'm not particularly familiar with this unit test either)

alparslanavci: I think the previous tests are also incorrect, I tried with the old version but it was returning zero. But I know that `localMapStats.getTotalGetLatency()` returns the latency in milliseconds, not nanoseconds.

mesutcelik:
Hi @checketts , Do you have any other comments before merging this?

checketts:
I'm a little worried about locking users into upgrading HazelCast if they upgrade Micrometer. I wish this support was in Hazelcast itself. 

It might be a better option to offer both a Hazelcast v3 and v4 option for users.

mesutcelik:
I see. 
I think one option would be to implement Hazelcast4CacheMetrics.java  and add hazelcast-4.jar as compile time dependency to micrometer. Then Users can can switch between v3 and v4 using a flag and based on their runtime dependency.

checketts:
> I think one option would be to implement Hazelcast4CacheMetrics.java and add hazelcast-4.jar as compile time dependency to micrometer.

Yes. That would be best. Also perhaps we could deprecate the v3 one, based on any end of life that Hazelcast v3 itself has.

leszko:
@checketts We had a chat inside Hazelcast about how to approach the issue and here are some thoughts.

- Supporting both Hazelcast versions at the same time can be challenging, because Hazelcast is added as a compile-time dependency, so usually you just update the dependency version. Everything else sounds like a hack.
- Technically, we could find some solution, but it would probably involve patching your release with some classes compiled with Hazelcast 4 + some dynamic class loading and class selection on your side. Or do you maybe have some simpler solution in mind?
- I see in some Spring Modules there were major dependencies update in the minor releases, for example, in [Spring Data Redis](https://docs.spring.io/spring-data/redis/docs/current/changelog.txt), `jedis` is updated `2.9.3 => 3.1.0` in the minor release

What are you thoughts on that? Do you consider updating to Hazelcast 4 in `1.5.0`? Or do you maybe have already plans for the next major release?

checketts:
> Technically, we could find some solution, but it would probably involve patching your release with some classes compiled with Hazelcast 4 + some dynamic class loading and class selection on your side. 

While dynamic class loading isn't fun, it offers the best compatibility and upgrade path. As long as Hazelcast v3 and v4 are both being maintained, I think it makes sense to do a best effort support. If it were as simple as offering both binders behind the scenes, and dynamically choosing the right one when possible, that would be ideal.

>    I see in some Spring Modules there were major dependencies update in the minor releases, for example, in Spring Data Redis, jedis is updated 2.9.3 => 3.1.0 in the minor release
>
> What are you thoughts on that? Do you consider updating to Hazelcast 4 in 1.5.0? 

It will be up to @shakuzen 's vote on the end result. I'm not opposed, I'm mainly trying to ask the questions that can encourage the best solution.

leszko:
@checketts ok, we'll send another PR on Monday, with the code compatible with both Hazelcast 3 and 4. Then we can decide which PR to merge.

Btw. thanks @checketts for being so responsive!

pkgonan:
Hi, when is the release date of this PR? :)

leszko:
@checketts I've sent an alternative PR #1952 which supports both Hazelcast 3 and Hazelcast 4. PTAL.

mesutcelik:
Hi @checketts @shakuzen ,

I see we have two options here. 

1. Upgrade to hazelcast 4 in micrometer-1.5 release
2. Give both hazelcast 3 and 4 support in micrometer-1.5 release by merging #1952, define hazelcast 3 deprecation policy and drop hazelcast 3 support in one of the future release.. 1.6,1.7 or 2.x

shakuzen:
Thanks for the pull request. Dropping support for Hazelcast 3 in a minor release is not something we're likely to do, unless a majority of users are requesting it. I will take a look at the other pull request.

PR #1877, opened 2020-03-04
Fix Javadoc error and warnings

izeye:
There are Javadoc error and warnings as follows:

```
$ ./gradlew clean javadoc
...
> Task :micrometer-core:javadoc
/Users/user/IdeaProjects/micrometer/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/MetricsDSLContext.java:54: error: block element not allowed within inline element <code>: pre
 * <code><pre>
         ^
1 error

> Task :micrometer-registry-statsd:javadoc
/Users/user/IdeaProjects/micrometer/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java:424: warning: no @return
    public int queueSize() {
               ^
/Users/user/IdeaProjects/micrometer/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java:432: warning: no @return
    public int queueCapacity() {
               ^
2 warnings
...
$
```

This PR fixes them.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -428,7 +428,7 @@ public class StatsdMeterRegistry extends MeterRegistry {
 
     /**
      * @deprecated queue capacity is no longer available since 1.4.0
-     * @return queue capacity
+     * @return constant {@literal -1}
      */
     @Deprecated
     public int queueCapacity() {
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -419,7 +419,7 @@ public class StatsdMeterRegistry extends MeterRegistry {
 
     /**
      * @deprecated queue size is no longer available since 1.4.0
-     * @return queue size
+     * @return constant {@literal -1}
      */
     @Deprecated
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -419,6 +419,7 @@ protected DistributionStatisticConfig defaultHistogramConfig() {
 
     /**
      * @deprecated queue size is no longer available since 1.4.0
+     * @return queue size
```
shakuzen: While the original intent of the method was to return the queue size, it no longer does that, so it would be better for the JavaDoc to reflect that.
```suggestion
     * @return constant {@literal -1}
```

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -427,6 +428,7 @@ public int queueSize() {
 
     /**
      * @deprecated queue capacity is no longer available since 1.4.0
+     * @return queue capacity
```
shakuzen: Same as the queueSize
```suggestion
     * @return constant {@literal -1}
```

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
izeye: @shakuzen Thanks for the feedback! I applied the suggested changes.

PR #1892, opened 2020-03-09
Make index type configurable for ES <7

### LINKED ISSUE #1891: ElasticMeterRegistry should allow indexType override
Currently, for ES versions < 7, indexType is hardcoded in TYPE_PATH_BEFORE_VERSION_7 static variable.
It would be good to have a way to override the indexType, so it can be something other than 'doc'

magaton:
Resolves #1891

### DIFF
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
@@ -286,8 +286,8 @@ public class ElasticMeterRegistry extends StepMeterRegistry {
     }
 
     /**
-	 * Return index type. default is 'doc'
-	 *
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
```diff
@@ -287,11 +286,21 @@ static int countCreatedItems(String responseBody) {
     }
 
     /**
-     * Return index name.
-     *
-     * @return index name.
-     * @since 1.2.0
-     */
+	 * Return index type. default is 'doc'
```
shakuzen: We should note here that this only applies to Elasticsearch versions before 7.

magaton: Thanks, done.

PR #1896, opened 2020-03-11
Fix kafka same tags

jeqo:
Based on @saltos report: https://github.com/micrometer-metrics/micrometer/pull/1835#issuecomment-597457115


saltos:
This solves the problem with the different values of the tag.

PR #1898, opened 2020-03-11
Fix rejected cloudwatch metric with empty tag

### LINKED ISSUE #1813: CloudWatchMeterRegistry sends tags with empty values which CloudWatch then rejects
Project reactor integration for instance defines an `exception` tag with a default value of "", for when collecting metrics for arbitrary sequences (`reactor.flow.duration`).

Similar to #1167

With the following workaround I can get the metrics to show up in CloudWatch:

```java
@Configuration
public class MetricsConfig {
  @Bean
  MeterRegistryCustomizer<CloudWatchMeterRegistry> sanitizeReactorTags() {
    return registry ->
        registry
            .config()
            .meterFilter(
                MeterFilter.replaceTagValues("exception", tag -> tag.isBlank() ? "None" : tag));
  }
}
```
^ I chose to turn the blank tag into "None", similar to how the Spring request latency metrics look like, but the tag could just as well be removed I guess.

carhartl:
CloudWatch rejects any metric with a dimension that has an empty value, which might cause unexpected loss of metrics. Reactor for instance produces metrics where the "exception" dimension may have an empty value, and thus those metrics do not show up in CloudWatch without a workaround.

Similar to how the InfluxMeterRegistry behaves (see #1166) we're dropping such a faulty dimension before sending off the data.

~Additionally dropping dimensions with values exceeding the maximum (255).~
EDIT: Additionally implementing a CloudWatch naming convention that ensures the maximum length for both key and value are adhered to (by truncating).

Reference:
https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_Dimension.html

Closes #1813

### DIFF
--- a/implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchMeterRegistry.java
+++ b/implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchMeterRegistry.java
@@ -27,6 +27,7 @@ import io.micrometer.core.instrument.util.NamedThreadFactory;
 import io.micrometer.core.instrument.util.StringUtils;
 import io.micrometer.core.instrument.util.TimeUtils;
 import io.micrometer.core.lang.Nullable;
+import io.micrometer.core.util.internal.logging.WarnThenDebugLogger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -150,6 +151,7 @@ public class CloudWatchMeterRegistry extends StepMeterRegistry {
     // VisibleForTesting
     class Batch {
         private long wallTime = clock.wallTime();
+        private WarnThenDebugLogger logger = new WarnThenDebugLogger(Batch.class);
 
         private Stream<MetricDatum> gaugeData(Gauge gauge) {
             MetricDatum metricDatum = metricDatum(gauge.getId(), "value", gauge.value());
@@ -271,9 +273,17 @@ public class CloudWatchMeterRegistry extends StepMeterRegistry {
 
         private List<Dimension> toDimensions(List<Tag> tags) {
             return tags.stream()
-                    .filter(t -> StringUtils.isNotBlank(t.getValue()))
+                    .filter(this::acceptableTag)
                     .map(tag -> new Dimension().withName(tag.getKey()).withValue(tag.getValue()))
                     .collect(toList());
+        }
+
+        private boolean acceptableTag(Tag tag) {
+            final boolean isAcceptable = StringUtils.isNotBlank(tag.getValue());
+            if (!isAcceptable) {
+                logger.log(String.format("Dropping tag with empty value: '%s'", tag.getKey()));
+            }
+            return isAcceptable;
         }
     }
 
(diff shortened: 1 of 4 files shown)


carhartl:
Open question: did this for the cloudwatch2 implementation only, should it also be done for the  other (older?) cloudwatch implementation?

shakuzen:
Thanks for the pull request. Would you be so kind as to rebase your changes on the `1.1.x` branch and target the pull request at that so we can get the fix into maintenance releases? I'll take care of merging it forward.

Related to your question, the cloudwatch2 implementation does not exist in the `1.1.x` branch. So the changes should be made for the older implementation, as we still support it in `1.1.x` even though it is deprecated in later releases. This is a good reminder to consider when we stop publishing the older implementation, though. So thank you for that.

implementations/micrometer-registry-cloudwatch2/src/main/java/io/micrometer/cloudwatch2/CloudWatchMeterRegistry.java
```diff
@@ -295,6 +296,10 @@ StandardUnit toStandardUnit(@Nullable String unit) {
 
         private List<Dimension> toDimensions(List<Tag> tags) {
             return tags.stream()
+                    // Take restrictions for dimensions in CloudWatch into account
+                    // https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_Dimension.html
+                    .filter(t -> StringUtils.isNotBlank(t.getValue()))
+                    .filter(t -> t.getValue().length() <= 255)
```
shakuzen: For maximum length constraints, we have historically truncated the tag value. We do this in a `NamingConvention` implementation that the MeterRegistry is configured to use.

implementations/micrometer-registry-cloudwatch2/src/main/java/io/micrometer/cloudwatch2/CloudWatchMeterRegistry.java
```diff
@@ -295,6 +296,10 @@ StandardUnit toStandardUnit(@Nullable String unit) {
 
         private List<Dimension> toDimensions(List<Tag> tags) {
             return tags.stream()
+                    // Take restrictions for dimensions in CloudWatch into account
+                    // https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_Dimension.html
+                    .filter(t -> StringUtils.isNotBlank(t.getValue()))
```
shakuzen: If we are dropping a tag, we could use our `WarnThenDebugLogger` to give some visibility to users that a tag is being dropped and for what reason. Otherwise, it may be too hard to notice and track down for users.

carhartl: The implementation feels a bit clumsy but since we're operating on a stream I did not have a better idea.

implementations/micrometer-registry-cloudwatch2/src/main/java/io/micrometer/cloudwatch2/CloudWatchMeterRegistry.java
carhartl: Done.

implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchMeterRegistry.java
```diff
@@ -149,6 +150,7 @@ public void onSuccess(PutMetricDataRequest request, PutMetricDataResult result)
     // VisibleForTesting
     class Batch {
         private long wallTime = clock.wallTime();
+        private WarnThenDebugLogger logger = new WarnThenDebugLogger(Batch.class);
```
shakuzen: I'd like to move this up to the CloudWatchMeterRegistry class, but I can take care of this in a follow-up polish commit.

PR #1909, opened 2020-03-17
Allowing CountAtBucket to store doubles

newffy:
Near all usages of `CountAtBucket.bucket()` are made from `double` context. This allows to make `HistogramSnapshot` objects for histograms with `double` bucket boundaries


jkschneider:
Merged with 3da05bbb8a66371f6e01ecb5963497e2b4850147.

jkschneider:
This change precedes additions for VictoriaMetrics sparsely populated histograms, and also better supports `DistributionSummary` whose base unit may not always be an integral value (e.g. temperature).

Further context around this change here: https://micrometer-metrics.slack.com/archives/C662HUJC9/p1584048617111800

PR #1913, opened 2020-03-18
Rename tags provider classes to indicate Servlet dependency

shakuzen:
This was added as part of the Jetty connector metrics, but I noticed that they are not generic to any HTTP implementation, so I thought best to include Servlet in the name before releasing 1.4. Let me know if this sounds good for now or if we want to make something more generic to encompass more than Servlet @jkschneider


jkschneider:
LGTM

PR #1917, opened 2020-03-19
Newrelic java agent polish

neiljpowell:
- Renamed NewRelicHttpClientProvider to NewRelicApiClientProvider.  
- Added agentClientProviderEnabled property to NewRelicConfig to make Client Provider selection property driven.
@shakuzen

### DIFF
--- a/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/ClientProviderType.java
+++ b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/ClientProviderType.java
@@ -16,5 +16,5 @@
 package io.micrometer.newrelic;
 
 public enum ClientProviderType {
-    insights_api, insights_agent
+    INSIGHTS_API, INSIGHTS_AGENT
 }
--- a/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/ClientProviderType.java
+++ b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/ClientProviderType.java
@@ -1,0 +1,5 @@
+package io.micrometer.newrelic;
+
+public enum ClientProviderType {
+    insights_api, java_agent 
+}
--- a/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/ClientProviderType.java
+++ b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/ClientProviderType.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright 2020 Pivotal Software, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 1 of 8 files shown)


implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicConfig.java
```diff
@@ -53,6 +53,16 @@ default String eventType() {
         return v;
     }
 
+    /**
+     * When this is {@code true}, the New Relic metrics will be published with the NewRelicAgentClientProvider.
+     * Defaults to {@code false}.
+     * @return whether to use the NewRelicAgentClientProvider
+     */
+    default boolean agentClientProviderEnabled() {
```
shakuzen: I'm wondering if an Enum wouldn't be better than a boolean. If we want to integrate the now [separate New Relic registry](https://github.com/newrelic/micrometer-registry-newrelic) into this repository, it might make sense as just another client provider implementation.

neiljpowell: Good point. That would be more flexible.  Let me update it.

neiljpowell: Given the potential to add support for New Relic's implementation I was explicit with the enum values.  The original integration is with their Insights API.  Their new registry would makes sense as a "metrics_api" value.

neiljpowell: Done with the change.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/ClientProviderType.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.newrelic;
+
+public enum ClientProviderType {
+    insights_api, insights_agent
```
shakuzen: Let's make these uppercase to match the naming convention for enums and let's leniently parse case insensitive when converting from String to enum

shakuzen: Given the names here, it seems the client providers should be renamed accordingly. Probably `NewRelicInsightsApiClientProvider` and `NewRelicInsightsAgentClientProvider`

neiljpowell: Yep! Done.

neiljpowell: Yep! Done.

PR #1918, opened 2020-03-19
New Relic Registry Builder Polish

neiljpowell:
Fixed Builder method names for api and agent client provider

### DIFF
--- a/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
+++ b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
@@ -44,10 +44,7 @@ public class NewRelicMeterRegistry extends StepMeterRegistry {
      * @param clock  The clock to use for timings.
      */
     public NewRelicMeterRegistry(NewRelicConfig config, Clock clock) {
-        this(config, ( (config.clientProviderType() != null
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java
```diff
@@ -108,20 +108,20 @@ NewRelicClientProvider getClientProvider() {
         }
 
         /**
-         * Use the New Relic Java Agent-based client provider to publish metrics.
+         * Use the New Relic Insights Java Agent-based client provider to publish metrics.
          * @return builder
          * @since 1.4.0
          */
-        public Builder agentClientProvider() {
+        public Builder insightsAgentClientProvider() {
```
shakuzen: Do we even need these methods anymore now that this is selected with the `NewRelicConfig` which is required to be passed to the Builder?

neiljpowell: I can actually remove them with a slight adjustment to where I put the constructor config check.  Will do.

neiljpowell: I left Builder and constructor support for providing a client provider.

PR #1920, opened 2020-03-20
Allow `ExecutorServiceMetrics` accepting metric name prefix

### LINKED ISSUE #1919: ExecutorServiceMetrics should allow customization of metric name
Unlike many JVM metrics, `ExecutorService` is actually instantiated by users and there may be many in an app. If the metric id can't be customised, multiple executors with different tags will fail to register due to Prometheus restrictions.

Is it ok to add an ID prefix to the factories for executor metrics? Happy to make the change if so but one thing that comes to mind is given there are already so many factories, if every one had a prefix version it would be a lot of new methods, wonder what a good practice in this codebase would be for that.

More context in https://github.com/line/armeria/issues/2605

ikhoon:
Motivation:

`ExecutorService` metrics are hardcoded with 'executor.' prefix.
Some users want to monitor an `ExecutorService` with their own namespace to avoid conflict
with other executors such as Reactor Scheduler.

Modifications:

- Add `metricPrefix` paremeter to `ExecutorServiceMetrics` factory methods
- Add `metricPrefix` parameter to `Timed*` class constructor

Result:
Fixes #1919

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
@@ -30,6 +30,8 @@ import java.util.concurrent.*;
 
 import static java.util.Arrays.asList;
 
+import com.google.common.annotations.VisibleForTesting;
+
 /**
  * Monitors the status of executor service pools. Does not record timings on operations executed in the {@link ExecutorService},
  * as this requires the instance to be wrapped. Timings are provided separately by wrapping the executor service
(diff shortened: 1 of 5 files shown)


anuraaga:
Thanks @ikhoon!

anuraaga:
@shakuzen If you get a chance to look at this, will be great. It seems like a bug to me since the metrics allow customization of tags without name, which isn't really allowed by Prometheus's data model.

adriancole:
@saturnism I wonder if this could help with some things we found in the past, like implicit executor usage being hard to track down. maybe that was only in tracing though..

checketts:
And just to be clear why a different prefix is needed instead of adding an extra tag, is because different executor metrics are using different tags and since Prometheus requires that they all have the same tag count, that causes trouble?

anuraaga:
Yup exactly. Various libraries will have different tags they're concerned with for their executors so with name being constant they can't run in the same app right now.

The other option would be to remove customization of tags but that's probably too limiting, not to mention backwards incompatible.

saturnism:
> I wonder if this could help with some things we found in the past, like implicit executor usage being hard to track down. maybe that was only in tracing though..

IIRC, it was that in trace, we were seeing many `async` spans and it wasn't unclear where it came from as there were no information abt the executor that executed the code.  I think we discussed a generic way to mark autoconfiguration-provided ExecutorService as not to trace.  Currently we need to explicitly list the bean names.

anuraaga:
Hello - just wondering if anyone could take a look at this change? Thanks!

shakuzen:
Thank you for the pull request and the patience on this. It's unfortunate we require a `name` tag that basically serves the same purpose as the prefix will, but I don't see a better alternative. Overall this looks like what we need to do. I would like to preserve `executor` in the name, opting for any prefix to be prepended onto that, e.g. `reactor.executor` / `armeria.executor`.
I will make those changes to this pull request before merging. Thanks again.

ikhoon:
>  I would like to preserve executor in the name, opting for any prefix to be prepended onto that, e.g. reactor.executor / armeria.executor.

That sounds good to me. And thanks for considering this PR. 😀

shakuzen:
Thanks everyone. Let me know if there's any feedback on the final changes or with trying snapshots. I'm planning to do the 1.5 release very soon, which will have these changes.

PR #1926, opened 2020-03-21
Disable Spring Integration graph endpoint

izeye:
I see the following error when running a sample like `SimpleSample`:

```
Error starting ApplicationContext. To display the conditions report re-run your application with 'debug' enabled.
:59:09 main ERROR o.s.b.d.LoggingFailureAnalysisReporter - 

***************************
APPLICATION FAILED TO START
***************************

Description:

A component required a bean named 'convert.input' that could not be found.


Action:

Consider defining a bean named 'convert.input' in your configuration.


Process finished with exit code 1
```

Disabling Spring Integration graph endpoint seems to resolve the error, so this PR disables it.

### DIFF
--- a/samples/micrometer-samples-boot2/src/main/java/io/micrometer/boot2/samples/integration/SpringIntegrationApplication.java
+++ b/samples/micrometer-samples-boot2/src/main/java/io/micrometer/boot2/samples/integration/SpringIntegrationApplication.java
@@ -15,30 +15,27 @@
  */
 package io.micrometer.boot2.samples.integration;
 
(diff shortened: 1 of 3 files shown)


jkschneider:
@izeye In 259f23b I moved the whole integration sample to another subproject to avoid this issue. Do you think it's worth doing this on the 1.4.x line? If you regularly use the samples to test fixes in older release branches perhaps?

izeye:
@jkschneider Thanks for the feedback! As the issue this PR was trying to resolve has been already resolved and there will be no upcoming releases in the 1.4.x line, we can close this now.

PR #1927, opened 2020-03-21
Use micrometer-core project dependency in micrometer-samples-boot2

izeye:
When I was looking at the build failure from https://github.com/micrometer-metrics/micrometer/pull/1921#issuecomment-601674179, I noticed that `micrometer-samples-boot2` uses Spring Boot managed version of `micrometer-core` dependency whereas it uses backend implementations from the project. This PR changes to use `micrometer-core` project dependency in `micrometer-samples-boot2`.

I don't know the root cause of the build failure, but this change seems to fix it.

### DIFF
--- a/samples/micrometer-samples-boot2/build.gradle
+++ b/samples/micrometer-samples-boot2/build.gradle
@@ -14,12 +14,15 @@ dependencyManagement {
 }
 
 dependencies {
+    implementation project(":micrometer-core")
     ['atlas', 'azure-monitor', 'prometheus', 'datadog', 'elastic', 'ganglia', 'graphite', 'jmx', 'influx', 'statsd', 'new-relic', 'cloudwatch', 'cloudwatch2', 'signalfx', 'wavefront', 'elastic', 'dynatrace', 'humio', 'appoptics', 'stackdriver'].each { sys ->
         implementation project(":micrometer-registry-$sys")
     }
 
     implementation 'org.springframework.boot:spring-boot-starter-web'
-    implementation 'org.springframework.boot:spring-boot-starter-actuator'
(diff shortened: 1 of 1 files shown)


izeye:
The build in my CircleCI has already finished before creating this PR, and its results didn't show up here. So I force-pushed without changes. The results seem okay now.

samples/micrometer-samples-boot2/build.gradle
```diff
@@ -14,12 +14,15 @@ dependencyManagement {
...
 dependencies {
+    implementation project(":micrometer-core")
     ['atlas', 'azure-monitor', 'prometheus', 'datadog', 'elastic', 'ganglia', 'graphite', 'jmx', 'influx', 'statsd', 'new-relic', 'cloudwatch', 'cloudwatch2', 'signalfx', 'wavefront', 'elastic', 'dynatrace', 'humio', 'appoptics', 'stackdriver'].each { sys ->
         implementation project(":micrometer-registry-$sys")
     }
 
     implementation 'org.springframework.boot:spring-boot-starter-web'
-    implementation 'org.springframework.boot:spring-boot-starter-actuator'
+    implementation('org.springframework.boot:spring-boot-starter-actuator') {
+        exclude group: 'io.micrometer', module: 'micrometer-core'
```
shakuzen: Do we need to explicitly exclude this here or does the direct declaration of `implementation project(":micrometer-core")` override the transitive dependency? Just curious.

izeye: @shakuzen I thought the same thing, but it seems to be necessary based on the build results [before](https://circleci.com/gh/izeye/micrometer/680) and [after](https://circleci.com/gh/izeye/micrometer/682) the exclusion although I didn't look into it.

PR #1930, opened 2020-03-21
Fix NewRelicMeterRegistry - No enum constant java.util.concurrent.TimeUnit.seconds

### LINKED ISSUE #1931: v1.4 NewRelicMeterRegistry - java.lang.IllegalArgumentException
java.lang.IllegalArgumentException: No enum constant java.util.concurrent.TimeUnit.seconds occurs in the writerTimer methods of the client provider implementations. 

See fix PR for more detail: https://github.com/micrometer-metrics/micrometer/pull/1930
@shakuzen

neiljpowell:
@shakuzen  I found a critical issue that exists in both of the Client Provider implementations when processing Timers.  The previous version of Micrometer used the NewRelicMeterRegistry’s BaseTimeUnit to compute mean, totalTime, etc.
v1.3.x Ex:
timer.mean(getBaseTimeUnit())

Whereas the Client Providers are coded to use the meter.Id BaseUnit (lack of registry and allows TImeUnit to be specified at meter creation vs. tied to registry). The issue is that the Meter.Id BaseUnit is lowercase and fails to convert to TimeUnit.
v1.4 Ex:
TimeUnit timeUnit = TimeUnit.valueOf(timer.getId().getBaseUnit());
timer.mean(timeUnit)

The underlying cause is in the base MeterRegistry (line 254) when a Timer is being created, it uses getBaseTimeUnitStr() which applies toLowercase() to the original TimeUnit provided.

This PR fixes with (timer.getId().getBaseUnit().toUpperCase(), to continue allowing  Timer Meter's to be created with varying TimeUnits.  It can be changed to use the registry's baseTimeUnit if preferred.  Or should the underlying MeterRegistry.getBaseTimeUnitStr() be changed? It would be more consistent/elegant if the Meter.Id's baseUnit could easily convert back to TimeUnit.

Resolves #1931

### DIFF
--- a/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicInsightsAgentClientProvider.java
+++ b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicInsightsAgentClientProvider.java
@@ -167,7 +167,7 @@ public class NewRelicInsightsAgentClientProvider implements NewRelicClientProvid
     @Override
     public Map<String, Object> writeTimer(Timer timer) {
         Map<String, Object> attributes = new HashMap<>();
-        TimeUnit timeUnit = TimeUnit.valueOf(timer.getId().getBaseUnit().toUpperCase());
+        TimeUnit timeUnit = timer.baseTimeUnit();
         addAttribute(COUNT, timer.count(), attributes);
         addAttribute(AVG, timer.mean(timeUnit), attributes);
         addAttribute(TOTAL_TIME, timer.totalTime(timeUnit), attributes);		
@@ -181,7 +181,7 @@ public class NewRelicInsightsAgentClientProvider implements NewRelicClientProvid
     @Override
     public Map<String, Object> writeFunctionTimer(FunctionTimer timer) {
         Map<String, Object> attributes = new HashMap<>();
-        TimeUnit timeUnit = TimeUnit.valueOf(timer.getId().getBaseUnit().toUpperCase());
+        TimeUnit timeUnit = timer.baseTimeUnit();
         addAttribute(COUNT, timer.count(), attributes);
         addAttribute(AVG, timer.mean(timeUnit), attributes);
         addAttribute(TOTAL_TIME, timer.totalTime(timeUnit), attributes);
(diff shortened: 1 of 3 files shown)


neiljpowell:
@shakuzen  Actually if you look at the NewRelicMeterRegistryTest.writeFunctionTimer, you'll see I'm creating it with TimeUnit.MILLISECONDS but the expected results have seconds.  It's quite misleading to offer TimeUnit on the API then it be ignored.

shakuzen:
Just for some background: Registering Timers to a MeterRegistry with different base time units isn't a use case we currently support. Using the MeterRegistry's base time unit is intentional to have a consistent time unit when reporting metrics to a backend, which may have preferences about time unit (e.g. Prometheus conventionally uses seconds). Note that whatever the base time unit is, a timer can record events in any time unit.

The concept of base unit is generic to all meters, but Timers specifically have a concept of base time unit because we know they are time-based. The generic base unit gets populated with a String value of its base time unit.

>  Actually if you look at the NewRelicMeterRegistryTest.writeFunctionTimer, you'll see that the FunctionTimer is creating with TimeUnit.MILLISECONDS but the expected results have seconds. It's quite misleading to offer TimeUnit on the API then it be ignored.

A FunctionTimer tracks timing via a function, which you may not control. As long as we know the time unit and specify it, we know how to convert the value returned by the function to other time units (e.g. the base time unit). The value isn't ignored, it is used in converting.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicInsightsAgentClientProvider.java
```diff
@@ -181,7 +181,7 @@ public void publish(NewRelicMeterRegistry meterRegistry) {
     @Override
     public Map<String, Object> writeFunctionTimer(FunctionTimer timer) {
         Map<String, Object> attributes = new HashMap<>();
-        TimeUnit timeUnit = TimeUnit.valueOf(timer.getId().getBaseUnit());
+        TimeUnit timeUnit = TimeUnit.valueOf(timer.getId().getBaseUnit().toUpperCase());
```
shakuzen: For `Timer` and `FunctionTimer`, there is a `baseTimeUnit` method that will return the base TimeUnit directly without having to do this. `LongTaskTimer` does not have such a method, which may have just been an oversight.

neiljpowell: I've updated those two.  Would the addition of LongTaskTimer getBaseTimeUnit(), AbstractLongTaskTimer and simple updates to newLongTaskTimer() methods in some meter registries be too much for this?  I have a branch with all tests working.

neiljpowell: @shakuzen  See PR https://github.com/micrometer-metrics/micrometer/pull/1944 for adding baseTimeUnit to LongTaskTimer.

shakuzen: Thank you for making it as a separate pull request. That would have been my preference anyways.

Issue #1123, opened 2019-01-04
Check NaN and infinity values validity across all meter registries


izeye:
See https://github.com/micrometer-metrics/micrometer/pull/1116#discussion_r245373370

# Meter registry checklist
- [x] AppOptics: #1109
- [ ] Atlas: N/A
- [x] CloudWatch: #1138
- [ ] Datadog: N/A
- [x] Dynatrace: #1116
- [x] Elasticsearch: #1135
- [ ] Ganglia: N/A
- [ ] Graphite: N/A
- [x] Humio: #1174
- [x] InfluxDB: #1138
- [ ] JMX: N/A
- [x] KairosDB: #1138
- [x] New Relic Insights: #1150
- [ ] Prometheus: N/A
- [ ] SignalFx: N/A
- [ ] Stackdriver: N/A
- [x] StatsD: #778
- [x] Wavefront: #778

jkschneider:
We need to handle it correctly everywhere, but some monitoring systems do indeed handle NaN and Infinity.

izeye:
@shakuzen I created one more PR (https://github.com/micrometer-metrics/micrometer/pull/1174) for this but didn't check all yet 😅

izeye:
I added a checklist for meter registries and checked all the items which already handle NaN and infinities for gauges, time gauges and function counters.

Any help for unchecked items would be appreciated.

ramunask:
For Elastic I am still getting errors:
```
[2020-03-23T11:22:19,308][DEBUG][o.e.a.b.TransportShardBulkAction] [AUDIT] [metrics-2020-03][0] failed to execute bulk item (index) index {[metrics-2020-03][_doc][wumzBnEBbzVK4unYMuEs], source[{"@timestamp":"2020-03-23T09:22:19.300Z","name":"tomcat_servlet_request","type":"timer","host":"SERVER1","name_tag":"default","count":0.0,"sum" :NaN,"mean":0.0}]}
org.elasticsearch.index.mapper.MapperParsingException: failed to parse
	at org.elasticsearch.index.mapper.DocumentParser.wrapInMapperParsingException(DocumentParser.java:191) ~[elasticsearch-7.6.1.jar:7.6.1]
	at org.elasticsearch.index.mapper.DocumentParser.parseDocument(DocumentParser.java:74) ~[elasticsearch-7.6.1.jar:7.6.1]
[...]

PR #1932, opened 2020-03-23
Handle NaN for FunctionTimer in ElasticMeterRegistry

izeye:
This PR changes to handle NaN for `FunctionTimer` in `ElasticMeterRegistry`.

See https://github.com/micrometer-metrics/micrometer/issues/1123#issuecomment-602488229


izeye:
Other meter registries that can't handle NaN need to be checked for the same concern.

PR #1938, opened 2020-03-24
Implement callEnd method in OkHttpMetricsEventListener

### LINKED ISSUE #1937: OkHttpMetricsEventListener leaks memory when used with OkHttpClient caching
I have recently been using Retrofit to make API calls, and two of the features we have are request metrics and caching requests. For Caching we use [okhttp3.Cache](https://github.com/square/okhttp/blob/okhttp_3.14.x/okhttp/src/main/java/okhttp3/Cache.java). Since we heavily use Spring Boot 2.x, micrometer was an easy choice for metrics and implementation was easy with the `OkHttpMetricsEventListener`. 

## The Problem

Over time we have seen instances of our applications run out of memory and die / stop responding. We narrowed this down to the combination of metrics and caching when making API calls. 

## The Cause

The `OkHttpMetricsEventListener` maintains a map of all `Call`s, and it seems that calls are removed in two ways:

1. Call ends in Failure - which happens if an IOException is thrown
2. The response headers are read

When caching is used, neither of these conditions happen - there is no error thrown, and the response headers aren't read because a network call is never made.

We tracked this down to `OkHttpMetricsEventListener` not implementing the `EventListener.callEnd()` method, which is what would get called when a cache hit happens in OkHttp. Because this isn't implemented, the `Call`s end up living in the `callState` map and don't become eligible for garbage collection. Over time this fills up the available memory and the JVM dies.

## The Solution

Implementing the `callEnd()` method in the `OkHttpMetricsEventListener` to remove the `Call` from the `callState` map for requests that don't fall into one of the scenarios listed above.

Fixed by #1938 

### JVM Heap size

image
before fix: **yellow** | after fix: **green**

image
before fix: **yellow** | after fix: **green**

### Response Times

image
before fix: **yellow** | after fix: **green**

### Heap Analysis

image

image

ahatzz11:
Fixes #1937

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
@@ -147,7 +147,7 @@ class OkHttpMetricsEventListenerTest {
     }
 
     @Test
-    void cachedRequestsDoNotLeakMemory(@WiremockResolver.Wiremock WireMockServer server, @TempDir Path tempDir) throws IOException {
+    void cachedResponsesDoNotLeakMemory(@WiremockResolver.Wiremock WireMockServer server, @TempDir Path tempDir) throws IOException {
         OkHttpMetricsEventListener okHttpMetricsEventListener = OkHttpMetricsEventListener.builder(registry, "okhttp.requests").build();
         OkHttpClient clientWithCache = new OkHttpClient.Builder()
                 .eventListener(okHttpMetricsEventListener)
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
@@ -21,13 +21,17 @@ import io.micrometer.core.instrument.MockClock;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.simple.SimpleConfig;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import okhttp3.Cache;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
+import okhttp3.Response;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
 import ru.lanwen.wiremock.ext.WiremockResolver;
 
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.concurrent.TimeUnit;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.*;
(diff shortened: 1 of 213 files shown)


johnrengelman:
In addition to this change, it may be worth considering using a `WeakHashMap` instead of `ConcurrentHashMap` to hold the state. In this way, if the `Call` instance no longer has a strong reference it can be garbage collected from the `callState` map and released.

To ensure Thread safety, the `WeakHashMap` can be wrapped with `Collections.synchronizedMap(new WeakHashMap())`

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
```diff
@@ -141,4 +145,25 @@ void uriTagWorksWithUriMapper(@WiremockResolver.Wiremock WireMockServer server)
...
+                .url(server.baseUrl())
+                .build();
+
+        clientWithCache.newCall(request).execute().close();
+        assertThat(okHttpMetricsEventListener.callState).isEmpty();
+        try (Response response = clientWithCache.newCall(request).execute()) {
+            assertThat(response.cacheResponse()).isNotNull();
+        }
+
+        assertThat(okHttpMetricsEventListener.callState).isEmpty();
```
shakuzen: I added this test which I think captures the scenario described. Unfortunately, this assertion is failing for me. Any ideas why @ahatzz11? I'm not seeing the `callEnd` callback called when closing the cached response.

ahatzz11: @shakuzen Looking at the test my first guess was that the second request that is made is saved to `Response response` on line 163, but that response never has a `response.close()`. 

I pulled your test down and it's passing for me locally without any changes, not sure why that is yet. 

I'm curious - why is there a `try` block here?

shakuzen: The try block is what closes the Response when the block closes.

ahatzz11: I've done some debugging locally on our project and all of our cached calls end up hitting the `callEnd()` method and the `callState` becomes empty. 

To me this test is accurately testing this scenario. It seems like there is a call that is still in the `callState` despite both calls being closed out which would is partly what this PR is trying to resolve. 

I'll continue digging more tomorrow.

johnrengelman: The test is failing on the 1.1.x branch because it's using OkHttp 3.11 which has a bug in which `callEnd` was not executed. This was fixed around the `3.14.2` version of OkHttp:
https://github.com/square/okhttp/commit/59e68cd2bcba789e7faca70f8910afd2edeed2be
https://github.com/square/okhttp/issues/5143

@shakuzen how do you want to address this? Is upgrading okhttp in a maintenance branch acceptable? Alternatively, we could use the `WeakHashMap` change I described above and then include a release note that says when using the metrics integration, to be sure to upgrade to okhttp 3.14.2 or newer.

shakuzen: @johnrengelman thank you for tracking this down. Since we only have an optional dependency on OkHttp, the version is for checking compatibility and that the integration works as expected back to that version. We typically don't upgrade minor versions of dependencies in maintenance branches, so we can ensure the code works with versions of the library from that generation. That said, we could make an exception and upgrade given a fix isn't available in `3.11.x`.

I wanted to consider the change to a `WeakHashMap` separate from this fix, and probably only in `master`. I worry a bit that it would mask this kind problem where we should be explicitly cleaning up something but are not.

johnrengelman: @shakuzen dilemma understood! Do you want us to leave this in your hands to figure out or is there something we can do to help move this forward?

johnrengelman: I also verified that no other methods on `EventListener` are invoked for a cached response in v3.11 of OkHttp.

johnrengelman: I think I found a solution, but I think it still runs into a compatibility issue. We could change `OkHttpMetricsEventListener` to extend `EventListener.Factory` instead of `EventListener`. Then OkHttp would create a new listener per `Call` instead of a single global  listener which would get GC'd with each Call.

shakuzen: I've gone ahead and upgraded the okhttp dependency to 3.14.7 and the test is passing now. We can include a note in the release note about this. From the information shared here, it's only critical (though obviously recommended for everyone) for users using a `Cache` with okhttp to upgrade for this fix to work, correct?

johnrengelman: That seems to be the case, yes.

ahatzz11: Since `callEnd()` wasn't implemented before wouldn't it also be possible to have a call get stuck if anything other than an `IOException` was thrown?

Issue #1928, opened 2020-03-21
Build with JDK 14


shakuzen:
Now that Java 14 is officially released, we should upgrade our build to use that.

PR #1939, opened 2020-03-25
Upgrade to Gradle Wrapper 6.3

izeye:
This PR upgrades to Gradle Wrapper 6.3.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -205,7 +205,7 @@ subprojects {
 }
 
 wrapper {
-    gradleVersion = '6.2.2'
+    gradleVersion = '6.3'
 }
 
 defaultTasks 'build'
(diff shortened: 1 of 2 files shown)


shakuzen:
Thank you. I have targeted this at `1.4.x` as I'd like to keep less differences for build where possible between the maintenance branches. Gradle releases fairly often and we've got quite a gap in versions between our oldest supported maintenance branch and our newest. I didn't go older than 1.4.x just because with the changes made there it will be easy to apply this there but not so much to older branches. We'll see what we can do going forward. Build changes should hopefully be transparent to users.

shakuzen:
Also of note, this will make #1928 easier, as Gradle 6.3 supports running with JDK 14.

PR #1944, opened 2020-03-26
Long task timer add base time unit

neiljpowell:
Add baseTimeUnit to LongTaskTimer for consistency with other timers.  This adds AbstractLongTaskTimer which DefaultLongTaskTimer and SpectatorLongTaskTimer extends. Updates to the other implementations.  Other changes are to use a registry's getBaseTimeUnit() in newLongTaskTimer() to construct the timer.

See related comments: https://github.com/micrometer-metrics/micrometer/pull/1930

### DIFF
--- a/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/AtlasMeterRegistry.java
+++ b/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/AtlasMeterRegistry.java
@@ -174,7 +174,7 @@ public class AtlasMeterRegistry extends MeterRegistry {
 
     @Override
     protected LongTaskTimer newLongTaskTimer(Meter.Id id) {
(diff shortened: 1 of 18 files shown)


neiljpowell:
@shakuzen Would you prefer this PR to target master vs. 1.4.x?

shakuzen:
Checking with @jkschneider if there was any historical reason why `LongTaskTimer`s didn't include a method to get the base time unit like other Timers.
As for the target, either is fine. I almost wanted to go all the way back to 1.1.x, but it's only really come up as a need for the New Relic changes in 1.4.x, so adding it from there is probably fine.

neiljpowell:
latest 1.4.x updates merged, if you do want to accept for 1.4.x.

shakuzen:
Sorry for the delay and change of direction on this. This is now superseded by a redesign of LongTaskTimer done with #2005. We'll live with the lack of base time unit formally in 1.4. I've updated the New Relic portion in `master` already with dda07438395c37b04f1e7abe1845f944506d7a78

PR #1946, opened 2020-03-26
Use retryWhen() in StatsdMeterRegistry

izeye:
This PR changes to `retryWhen()` in `StatsdMeterRegistry` as `retry()` has been deprecated since https://github.com/reactor/reactor-core/pull/1979.


izeye:
The initial build has failed, but the subsequent retried build has succeeded. I'm not sure if it's flakiness that has existed before this change or is newly introduced due to this change.

See https://circleci.com/gh/izeye/micrometer/705

shakuzen:
> The initial build has failed, but the subsequent retried build has succeeded. I'm not sure if it's flakiness that has existed before this change or is newly introduced due to this change.

That's probably pre-existing flakiness. We probably need to take another look at anything we can do to make those tests less flaky. It was already quite a challenge to get them in the state they are in though.

PR #1952, opened 2020-03-30
Provide support for Hazelcast 4 (with compatibility to Hazelcast 3)

### LINKED ISSUE #1697: Support Hazelcast 4

leszko:
Fixes #1697

Alternative to #1868, uses reflection to keep the compatibility to Hazelcast 3


shakuzen:
Thanks for this. It'd be best if we have a sample project with Hazelcast 3 and a sample project with Hazelcast 4 that gets built and tests run on each to ensure we don't break anything going forward. They could go in the `samples` folder and added to the build.

leszko:
@shakuzen Good, I'll it to this PR.

leszko:
Added two sample projects:  'hazelcast', 'hazelcast-3-compatibility'. @shakuzen, could you please check?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetrics.java
```diff
@@ -45,7 +44,7 @@
      * @param <V>      The cache value type.
      * @return The instrumented cache, unchanged. The original cache is not wrapped or proxied in any way.
      */
-    public static <K, V, C extends IMap<K, V>> C monitor(MeterRegistry registry, C cache, String... tags) {
+    public static <K, V, C> C monitor(MeterRegistry registry, C cache, String... tags) {
```
snicoll, line 47: All those generics are useless now aren't they?

shakuzen: Indeed they are (unfortunately). I've polished this in 8f1bff3a94dce5ca105b288ddf7def8da109f36d

PR #1953, opened 2020-03-30
Drop zero-count infinity bucket in StackdriverMeterRegistry

izeye:
This PR changes to drop zero-count infinity bucket in `StackdriverMeterRegistry`.


shakuzen:
To be explicit, this is because of this behavior documented for Stackdriver bucket counts, I assume?

> If present, `bucket_counts` should contain N values, where N is the number
>        of buckets specified in `bucket_options`. If you supply fewer than N
>        values, the remaining values are assumed to be 0.

izeye:
@shakuzen I just did so for consistency as we already do so above the code:

https://github.com/micrometer-metrics/micrometer/blob/a8c6cdb567a6e74af37af657830a8eb9cf8d74e1/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java#L411-L424

Issue #1900, opened 2020-03-12
Usage of `numThreads` parameter 


poznachowski:
What I see in `PushRegistryConfig` interface is a `numThreads` parameters (that defaults to 2):
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushRegistryConfig.java#L50

However, it seems that it's not used at all (and never was - as far as I see):
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java#L67

The ScheduledExecutorService is using a single thread. You could you please clarify that?

I found it, while seeking the best approach to provide business metrics basing on SQL queries, that may take some time to execute. I hoped that I could use this parameter to parallelize retrieving Gauge values (SQL executions). I know that's out of scope (of the submitted issue), but could you provide any recommendations for my case? It seems that managing scheduled executor myself and create Gauges around calculated values is the best option.

jkschneider:
It was a holdover originally from `AtlasConfig`, which is maintained in the Spectator project. Micrometer doesn't use this anywhere AFAIK.

Concerning parallelizing SQL queries... 

> It seems that managing scheduled executor myself and create Gauges around calculated values is the best option.

This is definitely the best option in your case. In general, Micrometer's view of gauges is that their values are updated precisely when the gauge value is retrieved (at metrics publishing). But since you may conclude that your SQL queries are expensive and should run less frequently than the metrics publishing interval, then updating an intermediate value tied to the gauge on an interval that you determine is a good idea. You're effectively trading off mean-time-to-resolution for query impact.

PR #1955, opened 2020-03-31
Deprecate PushRegistryConfig.numThreads()

izeye:
This PR adds no support note to `PushRegistryConfig.numThreads()`'s Javadoc.

See gh-1900

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushRegistryConfig.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushRegistryConfig.java
@@ -44,9 +44,15 @@ public interface PushRegistryConfig extends MeterRegistryConfig {
     }
 
     /**
+     * Return the number of threads to use with the scheduler.
+     *
+     * Note that this configuration is NOT supported.
+     *
      * @return The number of threads to use with the scheduler. The default is
      * 2 threads.
+     * @deprecated since 1.1.13
      */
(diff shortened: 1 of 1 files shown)


shakuzen:
I think we're better off deprecating it. I assume the desired behavior is publishing the metrics (or batch of metrics) using `numThreads` threads. I'm not sure how we would generically implement this across push-based registries, especially because some use their own SDK that we do not control behavior like this.

izeye:
@shakuzen Thanks for the feedback! I changed to deprecate it.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushRegistryConfig.java
```diff
@@ -44,9 +44,15 @@ default boolean enabled() {
     }
 
     /**
+     * Return the number of threads to use with the scheduler.
+     *
+     * Note that this configuration is NOT supported.
+     *
      * @return The number of threads to use with the scheduler. The default is
      * 2 threads.
+     * @deprecated since 1.1.13
```
shakuzen, line 53:
```suggestion
     * @deprecated since 1.1.13 because this configuration is not used
```

izeye: @shakuzen Thanks! I applied the suggested change in 66cfe68.

Issue #1958, opened 2020-04-01
NewRelicClientProvider implementations not using NamingConvention configured via Builder


shakuzen:
The configured `NamingConvention` from the Builder is not passed to the NewRelicClientProvider implementations, so the default `NewRelicNamingConvention` is always used.

neiljpowell:
@shakuzen The implementations use the default just like the NewRelicMeterRegistry.  NamingConvention hasn't been exposed as a construction/builder option in previous versions, so the effect is the same.

shakuzen:
> NamingConvention hasn't been exposed as a construction/builder option in previous versions, so the effect is the same.

I'm wondering whether exposing it at all was a mistake (see https://github.com/micrometer-metrics/micrometer/pull/1957#issuecomment-607058490) without a request from users to be able to customize it. We're ending up with multiple places to pass things which requires understanding about which will eventually be used. If we are exposing it, though, users would expect it to have the intended effect.

neiljpowell:
Yep, understood. I was just pointing out that it's not actually a bug but I'm happy to help with the update to expose it now or for 1.5.  If I recall, there are other MeterRegistry implementations that also don't expose NamingConvention, which could be due to a strict preference of their target backends.  

Speaking of another possible oversight, it would be good to expose the underlying HttpUrlConnectionSender proxyHost and proxyPort options all the way to the NewRelicConfig.  It's needed for anyone using the API implementation in an environment without open Internet access.  We could do this and exposing NamingConvention for 1.5.

shakuzen:
The `namingConvention` is already a method on the Builder from 1.4.0, which is what caused me to notice and open this issue https://github.com/micrometer-metrics/micrometer/commit/5d5617ded1975298f9ed9ea4972f07fe31a4c313#diff-d8e6a06ddbf45f53edb233b7b4729cfcR127-R130

shakuzen:
Release notes should mention the removal of the constructor with NamingConvention from public API - it was a mistake this constructor was ever included in public API.

PR #1962, opened 2020-04-01
NewRelicClientProvider Use NamingConvention From Registry

neiljpowell:
This is to resolve https://github.com/micrometer-metrics/micrometer/issues/1958
Allow construction of client provider implementations with a naming convention, so that it can be supplied by the NewRelicMeterRegistry. This covers normal construction/builder use cases, except if a client provider is supplied to registry.  If the client provider is custom, then it's reasonable to expect the creator to explicitly use the NewRelicNamingConvention or not.

Alternatively, setNamingConvetion() could be added to the client provider implementations and used in registry construction.  Having NewRelicClientProvider.setNamingConvention() would make the expectation more clear/formal for other implementations.

@shakuzen

### DIFF
--- a/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicInsightsAgentClientProvider.java
+++ b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicInsightsAgentClientProvider.java
@@ -54,17 +54,18 @@ public class NewRelicInsightsAgentClientProvider implements NewRelicClientProvid
     private final NewRelicConfig config;
     private final Agent newRelicAgent;
     // VisibleForTesting
-    final NamingConvention namingConvention;
+    NamingConvention namingConvention;
     
     public NewRelicInsightsAgentClientProvider(NewRelicConfig config) {
-        this(config, new NewRelicNamingConvention());
-    }
-    
-    public NewRelicInsightsAgentClientProvider(NewRelicConfig config, NamingConvention namingConvention) {
-        this(config, NewRelic.getAgent(), namingConvention);
+        this(config, NewRelic.getAgent(), new NewRelicNamingConvention());
     }
 
-    public NewRelicInsightsAgentClientProvider(NewRelicConfig config, Agent newRelicAgent, NamingConvention namingConvention) {
+    public NewRelicInsightsAgentClientProvider(NewRelicConfig config, Agent newRelicAgent) {
+        this(config, newRelicAgent, new NewRelicNamingConvention());
+    }
+
+    // VisibleForTesting
+    NewRelicInsightsAgentClientProvider(NewRelicConfig config, Agent newRelicAgent, NamingConvention namingConvention) {
 
         if (!config.meterNameEventTypeEnabled() && StringUtils.isEmpty(config.eventType())) {
             throw new MissingRequiredConfigurationException("eventType must be set to report metrics to New Relic");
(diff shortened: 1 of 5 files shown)


shakuzen:
> This covers normal construction/builder use cases, except if a client provider is supplied to registry. If the client provider is custom, then it's reasonable to expect the creator to explicitly use the NewRelicNamingConvention or not.

I think we'll want to add a JavaDoc comment to the builder's `namingConvention` method to explain when it will or won't be used, similar to the `httpClient` method. It would probably be most clear if there weren't a public constructor with naming convention in the client providers, but that's already been released. We could still deprecate them if it brings clarity.

> Alternatively, setNamingConvetion() could be added to the client provider implementations and used in registry construction. Having NewRelicClientProvider.setNamingConvention() would make the expectation more clear/formal for other implementations.

I was hesitant about going this route, but it mirrors what we do typically (and still currently in the NewRelicMeterRegistry constructor):

https://github.com/micrometer-metrics/micrometer/blob/2b3fffea3e2c07d9e8575364f16293582528273b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicMeterRegistry.java#L77

So a mutable naming convention seems the route to go. In fact, we want to make sure that use of the above overrides the current naming convention, though I don't think dynamic changes to naming convention is a general use case.

neiljpowell:
Actually, the client providers haven't had a constructor with naming convention until this PR.  I planned to remove them going the mutable naming convention setter route.

shakuzen:
> Actually, the client providers haven't had a constructor with naming convention until this PR.

The following public constructors exist for both already. If them being public was a mistake in the first place, we could reduce its visibility and add a note to the release notes (as it is a breaking change).

https://github.com/micrometer-metrics/micrometer/blob/d23b8f3153b2f65d87ab323960b218d22d53231b/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicInsightsAgentClientProvider.java#L62

https://github.com/micrometer-metrics/micrometer/blob/f3748f4bbb045a910e7ab2b07efad2a5e588ee8e/implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicInsightsApiClientProvider.java#L81

neiljpowell:
Oh right. I'll reduce visibility, if you think it won't be too impactful.  They shouldn't be there if the registry doesn't even offer a public constructor with naming convention and it can only be customized via the builder.

implementations/micrometer-registry-new-relic/src/main/java/io/micrometer/newrelic/NewRelicInsightsAgentClientProvider.java
```diff
@@ -252,4 +258,8 @@ void sendEvents(Meter.Id id, Map<String, Object> attributes) {
             }
         }
     }
+    
+    public void setNamingConvention(NamingConvention namingConvention) {
```
shakuzen, line 262: Actually, I'm wondering if this should even be public. To simplify things, we want it to all go through the builder/registry, but this allows things to be changed outside of those.

shakuzen: Of course methods inherited from an interface can't be package-private, so that complicates things...

shakuzen:
> Looks good but using `MeterRegistry#config().namingConvetion()` still doesn't have the effect it should. I will follow-up with a commit fixing that and adding a test for it.

See 4374ae1

PR #1970, opened 2020-04-03
Fix MeterRegistry.gauge @return javadoc

padyx:
The Javadoc for the function `MeterRegistry.gauge(String, Iterable<Tag>, @Nullable T, ToDoubleFunction<T>)` is not quite correct.

It states that it returns "The number that was passed in..." while it actually takes and returns an arbitrary state object.

I suggest to change it with this PR changes this to "The state object that was passed in"


padyx:
Great, thank you! 
And thanks for the great work 👍

shakuzen:
I followed up with https://github.com/micrometer-metrics/micrometer/commit/2ca38de83f655cbdee7dbbcdf2711d3113a9e397 in a similar vein to these changes

PR #1971, opened 2020-04-04
Polish HibernateQueryMetrics changes

izeye:
This PR polishes `HibernateQueryMetrics` changes as follows:

- Removing `queryStatistics` `null` check as it doesn't seem to be necessary based on its implementation:

```java
	public QueryStatisticsImpl getQueryStatistics(String queryString) {
		return queryStatsMap.getOrCompute(
				queryString,
				s -> new QueryStatisticsImpl( s )
		);
	}
```

- Removing duplicate query check as it doesn't seem to be necessary based on its implementation:

```java
	public String[] getQueries() {
		return queryStatsMap.keysAsArray();
	}
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019 Pivotal Software, Inc.
+ * Copyright 2020 Pivotal Software, Inc.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateQueryMetrics.java
```diff
@@ -113,49 +112,46 @@ public void onPostLoad(PostLoadEvent event) {
         void registerQueryMetric(Statistics statistics) {
             for (String query : statistics.getQueries()) {
                 QueryStatistics queryStatistics = statistics.getQueryStatistics(query);
-                if (queryStatistics == null) continue;
-                String queryName = meterRegistry.config().namingConvention().tagValue(query);
```
shakuzen, line 117: Removing this removes sanitization of the query tag value, no?

izeye: @shakuzen I thought that it's okay as we already sanitize tag values in meter registries as we do for other meters.

shakuzen: That's true. We usually register names/values in a canonical form also, but you're right it doesn't make sense to do double work here.

Issue #1942, opened 2020-03-25
New kafka metrics is really slow  


tchiotludo:
I've an application with a simple Kafka Stream. 
I've added latest implementation of [KafkaStreamsMetrics](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaStreamsMetrics.java). 

The loading time of my `/prometheus` endpoint is very long and consuming a lot of cpu.
I try to troubleshot this and I found this to be really slow : 

https://github.com/micrometer-metrics/micrometer/blob/d9655b1f89944a7d14460168b68fcd2bb05b6ddf/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java#L154-L160

If I removed the double check, loading time is fine again : 
```java
//Double-check if new metrics are registered
checkAndBindMetrics(registry);
```

With `checkAndBindMetrics` : 
```bash
time curl http://kestra:2553/prometheus
real	0m6,209s
user	0m0,015s
sys	0m0,001s
```

Without `checkAndBindMetrics` : 
```bash
time curl http://kestra:2553/prometheus
real	0m0,079s
user	0m0,013s
sys	0m0,013s
```

Since in Kafka metrics, there is a lot of gauge with a lot of tags (metrics x topic x partition x thread, 300 for my application), I think we must not do this check or used a cache if necessary (didn't see any impact removing this check), this check will fetch all the metrics and use a synchronized methods and will be really slow, since call for every metrics.

shakuzen:
Thank you for trying out the new KafkaStreamsMetrics, and sorry for the performance issue. We'll take a look at improving this. /cc @jeqo

jeqo:
Thanks @tchiotludo. Yes, keeping Kafka metrics registry updated has been one of the most challenging parts when implementing this instrumentation. This double check is in place to keep the registry up to date when metrics change (e.g. when sending data to/consuming data from a new topic).
This check is under the assumption that the number of metrics will stabilize at some point (e.g. an application sending data to some topics will register all metrics when sending at least one message to all topics it's suppose to, same for consumer). Although this is not optimal, with current API this is the strategy to keep metrics updated.

Could you share a bit more about your use-case? e.g. how many stream-processors/producers/consumers are you instrumenting? how many topics are been used by clients and often this changes? are you getting this performance penalty on every call or only once?

[...]

PR #1973, opened 2020-04-06
Refresh Kafka metrics periodically instead of on each metric read

tchiotludo:
relate to #1942

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
@@ -24,6 +24,8 @@ import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.MeterBinder;
 import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
+
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
@@ -59,6 +61,7 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
 
     private final Supplier<Map<MetricName, ? extends Metric>> metricsSupplier;
     private final Iterable<Tag> extraTags;
+    private final Duration refreshDuration;
     private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
 
     /**
@@ -69,12 +72,19 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
     private String kafkaVersion = "unknown";
 
     KafkaMetrics(Supplier<Map<MetricName, ? extends Metric>> metricsSupplier) {
-        this(metricsSupplier, emptyList());
+        this(metricsSupplier, emptyList(), Duration.ofSeconds(60));
     }
 
     KafkaMetrics(Supplier<Map<MetricName, ? extends Metric>> metricsSupplier, Iterable<Tag> extraTags) {
         this.metricsSupplier = metricsSupplier;
         this.extraTags = extraTags;
+        this.refreshDuration = Duration.ofSeconds(60);
+    }
+
+    KafkaMetrics(Supplier<Map<MetricName, ? extends Metric>> metricsSupplier, Iterable<Tag> extraTags, Duration refreshDuration) {
+        this.metricsSupplier = metricsSupplier;
+        this.extraTags = extraTags;
+        this.refreshDuration = refreshDuration;
     }
 
     @Override public void bindTo(MeterRegistry registry) {
@@ -91,7 +101,7 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
         }
         if (startTime != null) bindMeter(registry, startTime, meterName(startTime), meterTags(startTime));
         // Collect dynamic metrics
-        scheduler.scheduleAtFixedRate(() -> checkAndBindMetrics(registry), 0, 5, SECONDS);
+        scheduler.scheduleAtFixedRate(() -> checkAndBindMetrics(registry), 0, refreshDuration.getSeconds(), SECONDS);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -87,7 +91,7 @@ else if (START_TIME_METRIC_NAME.equals(name.name()))
         }
         if (startTime != null) bindMeter(registry, startTime, meterName(startTime), meterTags(startTime));
         // Collect dynamic metrics
-        checkAndBindMetrics(registry);
+        scheduler.scheduleAtFixedRate(() -> checkAndBindMetrics(registry), 0, 5, SECONDS);
```
jeqo, line 94: This looks good to start with!

I was wondering two additional things to consider here:
- If number of metrics eventually stabilize, we could consider to have a back-off here to query less frequently if we get all metrics registered. Although this might not fit well with applications that have a more dynamic behavior (e.g. subscribing/sending to topics by pattern)
- We could customize how frequent this rate should be by adding a input parameter maybe? This way if the application has multiple topics, then we can increase this.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIntegrationTest.java
```diff
@@ -90,6 +90,8 @@ void shouldManageProducerAndConsumerMetrics() {
         out.println("Meters from producer after sending and consumer before poll:");
         printMeters(registry);
 
+        Thread.sleep(5 * 1000);
```
jeqo, line 93: We could make a check first and then schedule to avoid the initial sleep?

tchiotludo:
Just found an issue, we need to properly close the scheduler on system exit. 
On my case, the scheduler blocked main thread on proper exit

jeqo:
> Just found an issue, we need to properly close the scheduler on system exit.
> On my case, the scheduler blocked main thread on proper exit

We could add `AutoCloseable` interface to support closing the thread. Similar to `TomcatMetrics`

tchiotludo:
I added the `AutoCloseable` but seems to be a big warning that user need to use it like this : 
```java
        AdminClient client = AdminClient.create(properties);
        try (KafkaClientMetrics metrics = new KafkaClientMetrics(client)) {
            metrics.bindTo(meterRegistry);
        }
```
Need a proper documentation on that !

jeqo:
ofc. @shakuzen is this a common way to handle closing resources on binders? I've seen it in Tomcat, but not sure if there is another (better) way been used in other instrumentations.

tchiotludo:
Personally, I try to look and don't find other way. 
I try with `registry.isClosed()` but it's not try properly on exit. 
Seems to be the better with the low knowledge I have in micrometer :smile:

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
shakuzen: A backoff strategy is an interesting idea. I'm not sure it is worth the complexity until we have more reason for it. If the registry is a step registry, I would use the step as the interval. Otherwise, we can default to 1 minute (our general default step).

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIntegrationTest.java
shakuzen: Yes, I think either doing the check as part of the logic or triggering the check in the test so we don't have to wait is better.

shakuzen:
> ofc. @shakuzen is this a common way to handle closing resources on binders? I've seen it in Tomcat, but not sure if there is another (better) way been used in other instrumentations.

The `MeterBinder` should be closed when the application shuts down. Not immediately after registering.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIntegrationTest.java
jeqo: If frequency were to be 1 min, then I’d prefer to do a first check before scheduling to have an initial set of metrics available

tchiotludo: Done :)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
tchiotludo: Made a configuration parameter + change default value

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIntegrationTest.java
tchiotludo: For information, there is an initial fetch with the `scheduleAtFixedRate(Runnable command,
                                                  long initialDelay,
                                                  long period,
                                                  TimeUnit unit)`, I set an `initialDelay` to `0`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -65,12 +72,19 @@
     private String kafkaVersion = "unknown";
 
     KafkaMetrics(Supplier<Map<MetricName, ? extends Metric>> metricsSupplier) {
-        this(metricsSupplier, emptyList());
+        this(metricsSupplier, emptyList(), Duration.ofSeconds(60));
```
jeqo, line 75: nit: maybe turn duration into a constant?

tchiotludo: done :)

shakuzen:
Polished with 8d817f2f6324f33a1490487c951d1423f6425581. I wanted to use the registry's step interval as the default refresh interval, if possible, but I realized we don't have a way to get that from a MeterBinder and didn't think it was worth making those changes at this time. Also we don't actually expose a way to customize the refresh interval through public API now, but I think that's fine until we better understand such a use case.

tchiotludo:
This make sense, to be honest, I already use the binder on a production app and I don't change the default refresh duration since a lot a scraper have a default to 60 sec.

PR #1978, opened 2020-04-08
fix (kafka): only consider same client meters when filtering out

### LINKED ISSUE #1968: Bug in new KafkaMetrics (1.4.0+) when using multiple clients and common tags
I was attempting to use the new Kafka meter binders introduced in 1.4.0 and ran into a problem. The application has multiple Kafka producers and consumers, and uses a `MeterFilter` to apply common tags to all meters. The behavior I'm seeing is that meters only get registered for _one_ consumer and _one_ producer.

I traced the problem to `checkAndBindMetrics`, specifically where the code removes/ignores meters with "less tags". The first Kafka client works fine, but all meters for any subsequent client will not be registered because they won't yet have the common tags applied by the `MeterFilter`, thus will have "less tags".

When I comment out this code, all client meters are registered as expected:
```
                        //Kafka has metrics with lower number of tags (e.g. with/without topic or partition tag)
                        //Remove meters with lower number of tags
//                        boolean hasLessTags = false;
//                        for (Meter other : registry.find(meterName).meters()) {
//                            List<Tag> tags = other.getId().getTags();
//                            if (tags.size() < meterTags.size()) registry.remove(other);
//                            // Check if already exists
//                            else if (tags.size() == meterTags.size())
//                                if (tags.equals(meterTags)) return;
//                                else break;
//                            else hasLessTags = true;
//                        }
//                        if (hasLessTags) return;
```
I'm sure this code was added for good reason, I'm just not clear on why meters with less tags need to be removed, so I'm hesitant to submit this in a PR. 

/cc @jeqo for his thoughts.

Slight tangent: in the course of this investigation I read through several issues & PRs and just wanted to say that I recognize that Kafka client metrics are complicated, and that I appreciate the effort being put forth in this project to try to get it as right as possible -- adding support for producer and streams clients, and getting away from JMX. I'm also following PR #1173 re: Kafka client binders using a Kafka metric reporter.

jeqo:
* Handle multiple clients
* Handle common tags

Fix #1968

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
@@ -59,6 +59,8 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
     static final String VERSION_METRIC_NAME = "version";
     static final String START_TIME_METRIC_NAME = "start-time-ms";
     static final Duration DEFAULT_REFRESH_INTERVAL = Duration.ofSeconds(60);
+    static final String KAFKA_VERSION_TAG_NAME = "kafka-version";
+    static final String CLIENT_ID_TAG_NAME = "client-id";
 
     private final Supplier<Map<MetricName, ? extends Metric>> metricsSupplier;
     private final Iterable<Tag> extraTags;
@@ -71,6 +73,7 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
     private volatile Set<MetricName> currentMeters = new HashSet<>();
 
     private String kafkaVersion = "unknown";
+    private String clientId = "unknown";
 
     KafkaMetrics(Supplier<Map<MetricName, ? extends Metric>> metricsSupplier) {
         this(metricsSupplier, emptyList());
@@ -92,6 +95,7 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
         Metric startTime = null;
         for (Map.Entry<MetricName, ? extends Metric> entry : metrics.entrySet()) {
             MetricName name = entry.getKey();
+            if (clientId.equals("unknown")) clientId = name.tags().get(CLIENT_ID_TAG_NAME);
             if (METRIC_GROUP_APP_INFO.equals(name.group()))
                 if (VERSION_METRIC_NAME.equals(name.name()))
                     kafkaVersion = (String) entry.getValue().metricValue();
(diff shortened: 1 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -132,6 +139,17 @@ else if (tags.size() == meterTags.size())
         }
     }
 
+    private boolean differentClient(List<Tag> tags) {
+        boolean anotherClient = false, anotherVersion = false;
+        for (Tag tag : tags) {
+            if (tag.getKey().equals(CLIENT_ID_TAG_NAME))
+                if (!clientId.equals(tag.getValue())) anotherClient = true;
+            if (tag.getKey().equals(KAFKA_VERSION_TAG_NAME))
+                if (!kafkaVersion.equals(tag.getValue())) anotherVersion = true;
```
shakuzen, line 148: I'm wondering how there would be multiple versions of the kafka client at the same time without some classloader trickery. That said, we can still guard against this - just wondering if it is a case you are expecting.

jeqo: True, I think I went to far with this hehe. I was considering two apps same client id with diff version but that would happen on different jvms. I will keep it simple without version and add an IT.

jeqo:
@shakuzen , added an IT case for it.

Im still wondering how to handle the case when common tags are applied via registry. any thoughts on that?

shakuzen:
@jeqo I've added a commit with the hack I mentioned in my comment on the issue. Please take a look and let me know what you think. Particularly if I've capture the scenario correctly in the test code. If it looks right, would you and/or @snagafritz try out the code in this pull request to see if it now solves both of the issues raised?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -87,20 +88,34 @@
     }
 
     @Override public void bindTo(MeterRegistry registry) {
+        commonTags = getCommonTags(registry);
+        prepareToBindMetrics(registry);
+        checkAndBindMetrics(registry);
```
shakuzen, line 93: I should have changed this in a separate commit, but note I've made sure we're calling `checkAndBindMetrics` synchronously on binding to ensure that has been done once by the return of the binding. This may allow us to simplify some of the test code that otherwise explicitly calls it.

Issue #2002, opened 2020-04-17
1.3.x and 1.4.x not compatible


anuraaga:
I noticed https://github.com/micrometer-metrics/micrometer/commit/3da05bbb8a66371f6e01ecb5963497e2b4850147 has a backwards incompatible change of the return type from `Long getMaximumExpectedValue()` to `Double getMaximumExpectedValue()`. Since micrometer uses semantic versioning, as documented at https://micrometer.io/docs/support, I guess this wasn't intended?

Because of the change, all Armeria users are locked into 1.3.x, and if a library targeted 1.4.x and used a new API, it would be impossible to use this library with Armeria without asking it to downgrade, possibly losing some monitoring features. For reference, this change downgraded Armeria's dependency from 1.4.x to 1.3.x, requiring code change since they are not API or ABI compatible.

https://github.com/line/armeria/commit/890631e15dff9b7030f20b18649d461e3efb5b34#diff-c966b7c3ea79ab16c67724170ff57f47R115

I understand that 1.4.x is not LTS, but I guess that doesn't preclude it from actually being used, it's more about whether it gets patch updates or not. But with this API break, more and more users will be unable to use it due to the API mismatch.

Unfortunately 1.4.x is released, but do you think we can add a method to both 1.3.x and 1.4.x, `getMaximumExpectedValueAsDouble` to provide libraries a way to work with both versions?

shakuzen:
Just something that came to mind, here's a hack I think should work with 1.3.x and 1.4.x:

```java
Long someLong = (distributionStatisticConfig.getMaximumExpectedValue() == null) ? null : ((Number) distributionStatisticConfig.getMaximumExpectedValue()).longValue();
Double someDouble = distributionStatisticConfig.getMinimumExpectedValue() == null ? null : ((Number) distributionStatisticConfig.getMinimumExpectedValue()).doubleValue();
```

anuraaga:
I would have to try to make sure, but pretty sure it doesn't work since `Long` vs `Double` are not ABI compatible so the method call itself will crash with `NoSuchMethodError`

```
'java.lang.Long io.micrometer.core.instrument.distribution.DistributionStatisticConfig.getMaximumExpectedValue()'
java.lang.NoSuchMethodError: 'java.lang.Long io.micrometer.core.instrument.distribution.DistributionStatisticConfig.getMaximumExpectedValue()'
[...]

PR #2004, opened 2020-04-17
Add get*ValueAsDouble and getServiceLevelObjectiveBoundaries methods …

anuraaga:
…to allow libraries to use these methods regardless of whether 1.3.x or 1.4.x is on the classpath.

This is for #2002 but for the 1.3.x branch. Because 1.4.x has methods with the same name but different return types, a library compiled against 1.4.x won't be usable with 1.3.x due to ABI change.

See #2003

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
@@ -154,7 +154,7 @@ public class DistributionStatisticConfig implements Mergeable<DistributionStatis
 
     /**
      * @deprecated Use {@link #getMinimumExpectedValueAsDouble}. If you use this method, your code
-     * will not be compatible with code that uses Micrometer 1.4.x.
+     * will not be compatible with code that uses Micrometer 1.4.x and later.
      */
     @Deprecated
     @Nullable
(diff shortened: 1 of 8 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
```diff
@@ -148,15 +152,35 @@ public Integer getPercentilePrecision() {
...
     @Nullable
-    public Long getMinimumExpectedValue() {
-        return minimumExpectedValue;
+    public Double getMinimumExpectedValueAsDouble() {
+        return minimumExpectedValue != null ? (double) minimumExpectedValue : null;
+    }
+
+    /**
+     * @deprecated Use {@link #getMaximumExpectedValueAsDouble}. If you use this method, your code
+     * will not be compatible with code that uses Micrometer 1.3.x.
```
shakuzen, line 178:
```suggestion
     * will not be compatible with code that uses Micrometer 1.4.x.
```

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
```diff
@@ -148,15 +152,35 @@ public Integer getPercentilePrecision() {
         return percentilePrecision;
     }
 
+    /**
+     * @deprecated Use {@link #getMinimumExpectedValueAsDouble}. If you use this method, your code
+     * will not be compatible with code that uses Micrometer 1.4.x.
```
shakuzen, line 157: 1.4.x will be end of life as soon as 1.5 is released (soon™), so probably worth mentioning "1.4.x and later"

PR #2007, opened 2020-04-18
Update default value logic for graphite tags

fitzoh:
When Graphite tag support was initially added to micrometer in #1806, tag support was configured to be on by default.

In an ensuing discussion (https://github.com/micrometer-metrics/micrometer-docs/issues/123), @snicoll suggested updating the logic to default to `true` *unless* the user has specified prefix tags.

This PR implements those changes, as well as adding a configuration property for `graphiteTagsEnabled`.

Also related: https://github.com/spring-projects/spring-boot/issues/20884

### DIFF
--- a/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteConfig.java
+++ b/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteConfig.java
@@ -49,13 +49,13 @@ public interface GraphiteConfig extends DropwizardConfig {
 
     /**
      * @return Whether Graphite tags should be used, as opposed to a hierarchical naming convention.
-     * Defaults to true if a value is present for {@link GraphiteConfig#tagsAsPrefix}.
+     * Defaults to true if no values are present for {@link GraphiteConfig#tagsAsPrefix}.
      * @see <a href="https://graphite.readthedocs.io/en/latest/tags.html">Graphite Tag Support</a>
      * @since 1.4.0
      */
     default boolean graphiteTagsEnabled() {
         String v = get(prefix() + ".graphiteTagsEnabled");
-        return v == null ? tagsAsPrefix().length > 0 : Boolean.parseBoolean(v);
+        return v == null ? tagsAsPrefix().length == 0 : Boolean.parseBoolean(v);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteConfig.java
```diff
@@ -49,12 +49,13 @@ default String prefix() {
 
     /**
      * @return Whether Graphite tags should be used, as opposed to a hierarchical naming convention.
-     * Defaults to {@code true}.
+     * Defaults to true if a value is present for {@link GraphiteConfig#tagsAsPrefix}.
```
shakuzen, line 52: Don't we want this to be `true` if a value is NOT present for `tagsAsPrefix`? This enables dimensional tags, `tagsAsPrefix` is for hierarchical naming. /cc @snicoll

snicoll: Yes, we want the opposite. 

The way I understood this is that if `tagsAsPrefix` has a value, you'd want the "non-native" way of tagging. If you are upgrading to a new version of micrometer that provides a native version of tagging, I think you should not be affected by this change.

So, a user upgrading:

* If they have `tagsAsPrefix` set we should use that (native tagging not enabled since it's going to ignore those anyway)
* If they don't have `tagsAsPrefix` set, then offer the native tagging (native tagging enabled since the user didn't express a use of the previous mechanism).

Makes sense? I think that's a fair summary of our discussion so far.

shakuzen: Sounds right to me. I wanted to make sure I wasn't mistaken.

fitzoh:
yeah, I'm running on very little sleep with quarantine toddler, almost
definitely flipped it.
should be able to update shortly

implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteConfig.java
fitzoh: Yeah, I'm running on very little sleep with quarantine toddler.

Updated in 03ea12adfc7666beb4b8e38e593d3d43c1473781

shakuzen: No problem. Thanks for updating and take care.

PR #2008, opened 2020-04-19
Completable future support for @Timed

JoseAlavez:
When utilizing reactive programming libraries that rely on Futures, `@Timed` will report the execution time for the `CompletionStage` creation, but not then it's completed. 

This PR is solves this issue, by intercepting `CompletionStage`s return types and stopping the timer once completed normally or exceptionally.


jkschneider:
Wow, thanks for the improvement @JoseAlavez. Significant first contribution!

PR #2009, opened 2020-04-19
Completable future support for @Counted

JoseAlavez:


jkschneider:
Thanks again @JoseAlavez. Glad this now mirrors what `@Timed` does.

PR #2020, opened 2020-04-22
Use FunctionTimer for cache.load.duration (fixes #1880)

### LINKED ISSUE #1880: Caffeine binder uses Gauge/TimeGauge for cumulative measures.
A couple of the metrics exposed by Caffeine are cumulative, so they should be wrapped accordingly, instead of using Gauge/TimeGauge.

* `cache.eviction.weight`/`evictionWeight()` should use FunctionCounter instead of Gauge.
* `cache.load.duration`/`totalLoadTime()` should use FunctionTimer instead of TimeGauge.

When the binder was first merged, FunctionCounter/FunctionTimer weren't yet available:
https://github.com/micrometer-metrics/micrometer/pull/78#issuecomment-321640355

john-karp:
Completes conversion of cumulative meters for Caffeine - see https://github.com/micrometer-metrics/micrometer/issues/1880#issuecomment-601889864

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
@@ -151,9 +151,11 @@ public class CaffeineCacheMetrics extends CacheMeterBinder {
 
         if (cache instanceof LoadingCache) {
             // dividing these gives you a measure of load latency
-            TimeGauge.builder("cache.load.duration", cache, TimeUnit.NANOSECONDS, c -> c.stats().totalLoadTime())
+            FunctionTimer.builder("cache.load.duration", cache,
+                                  c -> c.stats().loadCount(),
+                                  c -> c.stats().totalLoadTime(), TimeUnit.NANOSECONDS)
                     .tags(getTagsWithCacheName())
-                    .description("The time the cache has spent loading new values")
+                    .description("Cache loads, including both successes and failures")
                     .register(registry);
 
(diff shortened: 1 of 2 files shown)


shakuzen:
I'm not necessarily opposed to this change, but are we gaining anything? This effectively adds a counter for total loads, but that's already available by summing the existing counters for successful + failed loads.

john-karp:
My motivation is to handle the conversion from a cumulative cache loading time for registries that are stepped, not cumulative. The current implementation doesn't produce useful metrics with a stepped registry. AFAIK the only two meter types that can do the conversion are FunctionCounter and FunctionTimer.

I guess it would be possible to use FunctionCounter, and tweak the time units as necessary, but that seems hacky and goes against "Never count something you can time with a Timer or summarize with a DistributionSummary!"

So then if you go with a FunctionTimer, then it is necessary to plug in `c.stats().loadCount()`.

Ideally caffeine would provide `successLoadTime()` and `failureLoadTime()`, but I'm not sure how much work would be required Caffeine-side for that, or even if they would want such a change.

ben-manes:
Can you please clarify based on the response in https://github.com/ben-manes/caffeine/issues/409? A `failureLoadTime` seems at first a reasonable ask wrt symmetry, but operationally the metric doesn't appear to provide much insight. The time it takes to fail seems much less actionable than fixing the cause of the failure occurring.

jkschneider:
@ben-manes Please see my comment on the caffeeine issue [here](https://github.com/ben-manes/caffeine/issues/409#issuecomment-630312311).

ben-manes:
Thanks @jkschneider!

jonatan-ivanov:
Is this still needed/an issue? Do you still want us to merge this PR?

PR #2029, opened 2020-04-25
Add logback logger context listener that re-adds MetricTurboFilters after logger context reset

### LINKED ISSUE #2028: LogbackMetrics stop to count logback events after LoggerContext reset
Hi!

I use logback configuration with `scan="true"` that reconfigures logback whenever `logback.xml` changes. However, the first reload causes that no more logback events are recorded by the micrometer `LogbackMetrics`.

The reason this happens is that logback `LoggerContext` is reset when the configuration is reloaded and the reset clears the turbo filter list. Registering a `LoggerContextListener` that re-adds the MetricsTurboFilter solves this issue.

I'm going to attach a patch to LogbackMetrics that fixes this.

Thanks,
M.B.

bendiscz:
The `LogbackMetrics` stops to record logback events after logback LoggerContext reset. The reset can happen for instance when a configuration is reloaded via the "scan" feature. The reset listener solves this issue by re-adding all metrics turbo filters after every reset.

Note that the metricsTurboFilters map was changed from a `ConcurrentHashMap` to a `HashMap` with `synchronized` blocks. This is to prevent a race-condition between a call to `bindTo` method and logback reset event.

Fixes #2028

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
@@ -18,6 +18,7 @@ package io.micrometer.core.instrument.binder.logging;
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.LoggerContext;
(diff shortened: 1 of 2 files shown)


bendiscz:
You are right, this should be covered by a unit test, working on it right now.

Issue #1616, opened 2019-10-01
Apache HttpClient per route metrics


shakuzen:
See https://github.com/micrometer-metrics/micrometer/pull/1223#pullrequestreview-295350576 and prior comments.

sahilkamboj334:
Any update on this as well @shakuzen  ? currently i have to write my own metrics binder that extends existed pooling binder and add route specific metrics. However i also need to schedule a task which monitors every 5 sec and report it to meter registry.

shakuzen:
> However i also need to schedule a task which monitors every 5 sec and report it to meter registry.

It wouldn't be unprecedented to do that in a MeterBinder implementation, but it would be nice to avoid that if possible. I wonder if there's any callback available now where we can be notified of changes to routes. We have had #2031 open for a while, but I think there are some issues to work out such as handling this changing of routes. I've asked if there is interest in updating that PR. If there isn't, we'd try to take a look at another PR that addresses this.

shakuzen:
We have deprecated our Apache HttpClient instrumentation in #7158 in favor of the module maintained along with the project now. See [the documentation](https://hc.apache.org/httpcomponents-client-5.6.x/observation.html). Further enhancements to the instrumentation should be made there. For that reason, I'll close this issue and the related PR. Let us know if we can help review anything from the Micrometer perspective to be added to the httpclient5-observation module.

PR #2031, opened 2020-04-26
Apache HttpClient per route connection metrics 

mYakimchenko:
This PR adds per route connection metrics according to #1616

I'm not sure if it's good decision to introduce abstraction over `ConnPoolControl` as I found only one usage across _Apache HttpComponents_ of `HttpRoute` as generic type. It will break compatibility as well 🤔

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/AbstractPoolingHttpClientConnectionManagerMetricsBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/AbstractPoolingHttpClientConnectionManagerMetricsBinder.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2020 VMware, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.httpcomponents;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.lang.NonNull;
+import org.apache.http.conn.routing.HttpRoute;
+import org.apache.http.pool.ConnPoolControl;
+import org.apache.http.pool.PoolStats;
+
+import java.util.Collection;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Base class which collects total metrics including {@link HttpRoute} state from {@link ConnPoolControl}, for example {@link org.apache.http.impl.conn.PoolingHttpClientConnectionManager}
+ * for synchronous HTTP clients or {@link org.apache.http.impl.nio.conn.PoolingNHttpClientConnectionManager} for asynchronous HTTP clients.
+ * <p>
+ * It monitors the overall and per route connection pool state.
+ *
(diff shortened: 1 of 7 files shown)


jkschneider:
@mihanjk First of all, thanks for all the work that went into this. Seems like maybe you're first major Github PR to an OSS project? Looks like you have a bright future.

Give me a bit to examine this closer, please.

mihanjk:
@jkschneider Thanks. I opened pull request to open source project for the first time. 
It will be great if you can point out contributor guidelines for the project. I didn't find it

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/AbstractPoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,103 @@
...
+    public void bindTo(@NonNull MeterRegistry registry) {
+        super.bindTo(registry);
+        updateRoutesMetrics(registry);
+    }
+
+    /**
+     * Adds metrics for newly created {@link HttpRoute} to the given meter registry.
+     */
+    public void updateRoutesMetrics(MeterRegistry registry) {
+        for (HttpRoute route : getRoutes(connectionManager)) {
```
alan-kan-dragon, line 67: 👋 I come across this PR trying to resolve #1616 as I have similar requirements as well. Great initiative!

However, I wonder how this piece will work as the gauge per route is registered at "init" time and won't be updated going forward. If I understand the problem correctly, we need to add a gauge on a route created event but it's buried deep into the apache httpclient implementation...

mYakimchenko: Hi, Alan! Your comprehension of the problem is absolutely correct. Thanks for having a look 👍 

I didn't find out another way of exposing the routes by using contract provided by`PoolingHttpClientConnectionManager` class. So I've decided to move the responsibility of scheduling exposed method `updateRoutesMetrics` to the client side. That's not obvious, so I need to expand docs at least.

Potentially the problem can be solved by registering another gauge for that case. I'm not sure if it's a good idea because each call of `getRoutes` method cause acquiring the lock and creation of another `HashSet` object.

BobDu:
A year has passed. Is there any update for this feature? Thanks。

jchambers:
> A year has passed. Is there any update for this feature? Thanks。

@BobDu I'm not sure why you're tagging me on this issue, and suspect you meant to tag somebody else.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/AbstractPoolingHttpClientConnectionManagerMetricsBinder.java
```diff
@@ -0,0 +1,103 @@
...
+                poolStatsByHost.put(hostName, stats);
+                registerRouteMetrics(hostName, stats, registry);
+            }
+        }
+    }
+
+    private void registerRouteMetrics(String host, PoolStats poolStats, MeterRegistry registry) {
+        Gauge.builder("httpcomponents.httpclient.pool.route.max",
+                poolStats,
+                PoolStats::getMax)
```
tippl, line 80: I know this is year old, but i've come across this trying to implement it in my app.
The information below applies at least to `PoolingHttpClientConnectionManager`.

You can't use/reuse the PoolStats object here, because the object is created when `connectionManager.getStats(route);` is called and the primitive values are filled into it. It is an immutable object, so your statistics will permanently display the state they were in the moment registerRouteMetrics was first called for the host.

I've worked around this by passing the HttpRoute object into the registerRouteMetrics and calling getStats in the Gauge lambda. This seems to work correctly even if the connection pool completely empties and the Route is no longer returned in getStats. However it will need further testing to see if it works correctly in all cases.

shakuzen:
Sorry this PR has sat for so long without review from the team. Is there interest in updating it - that is, rebasing on the latest changes, which include an upgrade to Apache HttpComponents 5? I wonder if any of the review comments may be addressed by new features available in the latest Apache HC 5.

shakuzen:
Closing this due to inactivity and since our Apache HttpClient instrumentation is now deprecated in favor of the instrumentation maintained along with the project in a new module httpclient5-observation - see [the documentation](https://hc.apache.org/httpcomponents-client-5.6.x/observation.html). If someone wants to contribute an enhancement to that and the Micrometer team can be of help in reviewing from the Micrometer perspective, please let us know.

PR #2033, opened 2020-04-27
Introducing TimeWindowStats

massamany:
Hi,

This PR is not ready to be merged, but I wanted to open a discussion on metrics available in DistributionSummaries. May be it could be applicable to other metrics, for the moment I'm not experienced enough in micrometer code to be exhaustive, so this is why this discussion is needed.

First, when I started to use DistributionSummaries, I noticed that it provides Max value, but not Min value.
Also, I wanted to retrieve some more information, like mean and frequency.

So I started to create my own components. But doing so, I noticed that some of the metrics provided by DistributionSummary are not limited to time window (sum, count, mean) while others are (max).

I did not mange to find out why, but I wanted to have these metrics and the new one inside time window.
Typically, it's interesting to have average frequency (events / sec) on the last 2 hours for instance, but it has no meaning on the full uptime.

So I started to develop TimeWindowStats.
1 - It provides metrics, all in time window : [min, max, sum, mean, count, frequency, age].
2 - It can replace TimeWindowMax and TimeWindowSum.
3 - The "record" method does not use a system like the "updateMax" method in TimeWindowMax, but it relies on "updateAndGet" exposed by AtomicLong, which makes code safer and more maintainable.
4 - It will be easier to make it evolve with the "Snapshot" interface.
5 - It's tested :-)

I used PrometheusDistributionSummary as an example to how it can be used.

But I did not change default metrics exposed by Summaries. May-be it could be opportunate ?

Thanks in advance for your feedback. I think this could help improve metrics exposed by micrometer but I need some advices to make it clean and coherent with the whole code base :-)

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusDistributionSummary.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusDistributionSummary.java
@@ -15,15 +15,22 @@
  */
 package io.micrometer.prometheus;
 
-import io.micrometer.core.instrument.AbstractDistributionSummary;
-import io.micrometer.core.instrument.Clock;
-import io.micrometer.core.instrument.distribution.*;
-import io.micrometer.core.instrument.util.MeterEquivalence;
-import io.micrometer.core.lang.Nullable;
-
 import java.time.Duration;
 import java.util.concurrent.atomic.DoubleAdder;
 import java.util.concurrent.atomic.LongAdder;
+
+import io.micrometer.core.instrument.AbstractDistributionSummary;
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.distribution.CountAtBucket;
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
+import io.micrometer.core.instrument.distribution.FixedBoundaryVictoriaMetricsHistogram;
+import io.micrometer.core.instrument.distribution.Histogram;
+import io.micrometer.core.instrument.distribution.HistogramSnapshot;
+import io.micrometer.core.instrument.distribution.TimeWindowFixedBoundaryHistogram;
+import io.micrometer.core.instrument.distribution.TimeWindowStats;
+import io.micrometer.core.instrument.distribution.TimeWindowStats.TimeWindowStatsSnapshot;
+import io.micrometer.core.instrument.util.MeterEquivalence;
+import io.micrometer.core.lang.Nullable;
 
 public class PrometheusDistributionSummary extends AbstractDistributionSummary {
     private static final CountAtBucket[] EMPTY_HISTOGRAM = new CountAtBucket[0];
@@ -31,7 +38,7 @@ public class PrometheusDistributionSummary extends AbstractDistributionSummary {
     private final Histogram histogram;
     private final LongAdder count = new LongAdder();
     private final DoubleAdder amount = new DoubleAdder();
-    private final TimeWindowMax max;
+    private final TimeWindowStats stats;
 
     private final HistogramFlavor histogramFlavor;
 
@@ -45,7 +52,7 @@ public class PrometheusDistributionSummary extends AbstractDistributionSummary {
                 scale, false);
 
         this.histogramFlavor = histogramFlavor;
-        this.max = new TimeWindowMax(clock, distributionStatisticConfig);
+        this.stats = new TimeWindowStats(clock, distributionStatisticConfig);
 
         if (distributionStatisticConfig.isPublishingHistogram()) {
             switch (histogramFlavor) {
@@ -72,7 +79,7 @@ public class PrometheusDistributionSummary extends AbstractDistributionSummary {
     protected void recordNonNegative(double amount) {
         count.increment();
         this.amount.add(amount);
-        max.record(amount);
+        stats.record(amount);
 
         if (histogram != null)
             histogram.recordDouble(amount);
@@ -88,9 +95,17 @@ public class PrometheusDistributionSummary extends AbstractDistributionSummary {
         return amount.doubleValue();
     }
 
+    /**
+     * Shortcut to {@link #getTimeWindowStatsSnapshot()}.getMax()
+     */
     @Override
     public double max() {
-        return max.poll();
+        return stats.max();
+    }
+
+    // TODO : Add this method to interface.
+    public TimeWindowStatsSnapshot getTimeWindowStatsSnapshot() {
+        return stats.getSnapshot();
     }
 
     public HistogramFlavor histogramFlavor() {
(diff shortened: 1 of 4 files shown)


jkschneider:
@massamany Just to give some quick feedback on why max is treated differently than other distribution statistics in terms of its alignment to the publishing interval...

image

massamany:
Hi. Thanks for your comment.
I understand the reasons why Max is on time window, because it's the same reason why I wanted to have Min, Frequency and Mean also in the time window.
I understand the total count in the total uptime, even if I don't think it's useful in many cases. But the Mean computed in the full uptime, I have to say that I'm not convinced. So, to compute Mean in the time window, the Count and Sum must also be computed in this TW.
This is the why of the TimeWindowStats.
To be retro compatible, I did not propose to modify current behaviours. Also note that encapsulating the stats in a "Snapshot" class makes it more flexible for the future.

jkschneider:
@massamany Let's discuss these individually perhaps:

- **Min** I can see wanting this to have the same behavior as max _maybe_, but then again, it's hard to imagine how an instance under the kinds of pressure leading to an abnormally high max correlated to a failed attempt to push metrics would also have a lower min? Regardless of what we decide, I do want to carefully circumscribe even the tracking of min to those few monitoring systems that expect/require it. Min has proven over time to just not be that useful as an availability signal.
- **Count** (a.k.a. Frequency, a.k.a. Throughput) There is a tradeoff in windowing max like we are, and that's that the signal takes a while to decay on charts (shown below). The current notion is that any adverse relationship between throughput and response time is detectable in the max, but _not_ windowing throughput gives us a truer measure of it in any successfully reported interval. Decaying throughput too would unnecessarily give us two signals of this adverse relationship and dilute the value of throughput independently?
- **Mean** Partially I don't care because mean is such a generally worthless measure anyway. "Average: a random number that falls somewhere between the maximum and 1/2 the median. Most often used to ignore reality." Regardless, the windowing or not of mean follows from how we treat Count and Sum from which it is derived I think.

image

massamany:
Hi !

Thanks again for your comment. This is interesting to have this discussion, as I'm beginning in all these supervision aspects, so I have a lot to learn.

So, I understand the fact that the Min is not as interesting as other indicators. But, as I saw thing when I began this development, is that if Max increases, it's bad signal. But if Min does not, may-be it's just not as bad as if Min also increases. I don't know.

And the other one you did not speak about, is the frequency.
If the number of request / minutes increases from 100 to 500 for instance, it's may-be not surprising to see the Max and Min increase accordingly.
And it could help to detect moments in the day when applications are more or less used.

In fact, it's for the frequency that I put in place a count in time window. I think this one can be a real help for troubleshooting and sizing adjustments.

Well, that's all I can tell for the moment. Again, as I'm currently putting them in place, I don't have feedbacks for the moment on if they are really useful or just noise.

Cheers!

jkschneider:
> And the other one you did not speak about, is the frequency.

I tried to address this in the "Count" paragraph, indicating that frequency is the same thing as count is the same thing as throughput (with respect to timers anyway). Search for the paragraph starting with:

> Count (a.k.a. Frequency, a.k.a. Throughput)

Continuing the conversation though:

> If the number of request / minutes increases from 100 to 500 for instance, it's may-be not surprising to see the Max and Min increase accordingly.

For sure, throughput and latency _can_ be correlated, though there are other reasons latency may increase that are unrelated to throughput, for example:

* A blocking downstream service call is taking longer to respond now.
* Response time is proportional to the amount of time taken to fetch data which, due to an application design choice, may vary substantially depending on input.

In many of these cases, throughput (or whatever you want to call it, like frequency) will not increase but max latency will spike. And we want a finer grained throughput signal then, not one that unnecessarily smooths over a longer time window.

I think you'll find that metrics publishing succeeds in all but the most extreme situations. So the vast majority of the time you'll have your nice throughput signal and a response time signal from max.

jkschneider:
> is that if Max increases, it's bad signal. But if Min does not, may-be it's just not as bad as if Min also increases. I don't know.

Frankly, I think you should _only_ care about when max increases. When the max spikes, real users are impacted. Consider the implications of the ["birthday problem"](https://en.wikipedia.org/wiki/Birthday_problem). Even if the max only applies to 1% of your user interactions, supposing no coordination, for 100 requests there is a 63% chance that _some_ request will land in the top 1% latency.

massamany:
Hi,

First, sorry, I did not see the "a.k.a. Frequency". 

Then, thanks for all these answers and details about how supervision data might be used and interpreted. Think I learned things and it changed my vision on some points.

But still, I have doubt on the the throughput utility. I'll have to test it myself and make my own opinion I suppose.

And I think that even if the data is not directly exposed in the Summary (like count and max) well it could be offered and easily accessible (like mean).

All this, of course, if it does not introduce performance issues ! So, ultimate point, I made some bench of my code, just to see.

I saw that AtomicLong#updateAndGet has horrible performances. So I replaced it with a technique similar to the one used in TimeWindowMax and I got back to acceptable numbers.

Also, the internal structure of TimeWindowMax and TimeWindowStats are different : the current implementation performs more computation at the "recording" moment, while my implementations tries to limit operations at that moment and lets the "polling" compute a little more.
So I benched :
 1 - the original TimeWindowMax
 2 - my TimeWindowStats implementation
 3 - a test TimeWindowMax implementation based on my internal structure 
 4 - the class TimeWindowSum, which I saw recently

The bench is : 8 threads making 10 millions of recording (meaning **80 million** records in total), one thread polling every 10 ms. I tried different buffer lengthes and rotating intervals.
The figures speak clearly :
 1 - mainly between 500 - 550 ms. It gets worse with higher buffer length. 
 2 - mainly between 600 - 650 ms. It stays constant with higher buffer length.
 3 - mainly between 400 - 450 ms. It stays constant with higher buffer length.
 4 - around 7 / 8 seconds.

To conclude :
 - there is a overhead with the stats. but not so high.
 - the current TimeWindowMax implementation _could_ be improved in terms of performances.
 - the current TimeWindowSum implementation _should_ be improved in terms of performances.

Hoping these figures can be helpful to find leads for optimizations,

Cheers.

jkschneider:
> the current TimeWindowMax implementation could be improved in terms of performances.
the current TimeWindowSum implementation should be improved in terms of performances.

These are optimizations I think that would be nice to have regardless of what we do with `TimeWindowStats`. Thanks for this study!

> I have doubt on the the throughput utility.

I'd suggest trying it out a bit and seeing what you think.

So we're maybe in agreement on the goal: in the end, the decision of whether or not to window stats over a longer interval is not really about performance, it's about utility.

massamany:
> These are optimizations I think that would be nice to have regardless of what we do with `TimeWindowStats`. Thanks for this study!

I agree : this has nothing to do with the current PR. But I wanted to share this information with you.


> I'd suggest trying it out a bit and seeing what you think.

Will do, but trying to isolate what is useful or not, knowing everything we discussed.


> So we're maybe in agreement on the goal: in the end, the decision of whether or not to window stats over a longer interval is not really about performance, it's about utility.

Yep. Thanks for all this conversation. I'll close the PR. Please tell me if you want help for the perf thing.

jkschneider:
> Yep. Thanks for all this conversation. I'll close the PR. Please tell me if you want help for the perf thing.

Absolutely, would welcome a PR demonstrating this improvement!

Issue #457, opened 2018-02-21
min function on HistogramsSnapshot


randysecrist:
Quick question ...

Why doesn't https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/HistogramSnapshot.java have a min function?

jkschneider:
Just wanted to be clear that there was a use for it before introducing. What are you thinking of doing?

jkschneider:
BTW, workaround is a P0 percentile:

```
@Test
void minimum() {
    DistributionSummary summary = DistributionSummary.builder("my.summary")
            .publishPercentiles(0)
            .register(new SimpleMeterRegistry());

    // dummy data
    for (int i = 1; i < 10; i++) {
        summary.record(i);
    }

    assertThat(summary.percentile(0)).isEqualTo(1);
}
```

randysecrist:
Historical monitoring of snapshots (but not necessarily alert) a distribution summary for say ... minimum values for something like say ... request size.  All the other relevant stats are there; just seems to be missing this one.

jkschneider:
Reason I'm seeking concrete use cases is we should determine whether this is an always-on thing or an option to select. Leaning towards the latter unless there are an overwhelming number of use cases.

Request size has an interesting characteristic that it has a natural lower bound (request header size). Interestingly, it means request size essentially is never normally distributed and standard deviation is useless, but I digress...

Would you alert on min size?

jkschneider:
Sounds like default off with an option to turn it on would be OK then?

randysecrist:
Ya, I think that would work.  Thanks!

jkschneider:
This would be useful for the StackDriver [Range](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TypedValue#Range) aggregate as well.

jkschneider:
When implemented, fix min statistic on Azure App Insights implementation as well. #441.

martel:
This would also be useful while reporting to Cloudwatch where you could use StatisticSets: 
https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_StatisticSet.html

Instead of emitting each value separately in the reporter: 
https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-cloudwatch/src/main/java/io/micrometer/cloudwatch/CloudWatchMeterRegistry.java#L165-L168

[...]

PR #2038, opened 2020-04-28
add min statistic for distribution summaries and timers (#457)

thrxde:
to use StatisticSet for CloudwatchMetrics, minimum and maximum statistic value is required

### DIFF
--- a/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/SpectatorDistributionSummary.java
+++ b/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/SpectatorDistributionSummary.java
@@ -69,6 +69,11 @@ public class SpectatorDistributionSummary extends AbstractDistributionSummary {
         return Double.NaN;
     }
 
+    @Override
+    public double min() {
+        return Double.NaN;
+    }
+
     @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")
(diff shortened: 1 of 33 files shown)


massamany:
The TimeWindowMin would be obsolete with the proposition in #2033. But I did not go as far with the integration in the summaries. Merging the 2 PR could give something good, by adding not only min, but other stats (freq for instance seems very interesting to me) :-)

thrxde:
> The TimeWindowMin would be obsolete with the proposition in #2033. But I did not go as far with the integration in the summaries. Merging the 2 PR could give something good, by adding not only min, but other stats (freq for instance seems very interesting to me) :-)

#2033 Interesting approach. But didn't try to implement a concret "CloudwatchDistributionSummary". I tryed to extend the interface to provide the min value to all the regestry implementations. 
I don't know if this is nesessary for the other fields like "freq" and "age".  I think we need some help from micrometer experts.

jkschneider:
@thrxde I think I'd state a preference for limiting min to just the Cloudwatch implementation, because there is some history behind instrumentation containing min and it never really being used anywhere in testing for availability, so it's just extra overhead with no benefit?

But there is clearly a benefit to meeting the expectations of Cloudwatch.

thrxde:
@jkschneider thanks for the response. So I will close this PR and try to limit the "min" feature to Cloudwatch.

PR #2056, opened 2020-05-05
Remove implementation dependency on logback in stackdriver module.

IRus:
For libraries best practise is to include only slf4j-api as dependency.

### DIFF
--- a/implementations/micrometer-registry-stackdriver/build.gradle
+++ b/implementations/micrometer-registry-stackdriver/build.gradle
@@ -3,7 +3,7 @@ dependencies {
 
     api 'com.google.cloud:google-cloud-monitoring'
     implementation 'org.slf4j:slf4j-api'
-    implementation 'ch.qos.logback:logback-classic'
+    compileOnly 'ch.qos.logback:logback-classic'
 
     testImplementation project(':micrometer-test')
 }


implementations/micrometer-registry-stackdriver/build.gradle
```diff
@@ -3,7 +3,7 @@ dependencies {
 
     api 'com.google.cloud:google-cloud-monitoring'
     implementation 'org.slf4j:slf4j-api'
-    implementation 'ch.qos.logback:logback-classic'
+    compileOnly 'ch.qos.logback:logback-classic'
```
checketts, line 6: Should it perhaps be a test dependency?

IRus: IMHO, It should be another module and compile dependency. If it's intended to be shipped to end users. If it's just for internal use - yes, test module also works.

IRus: Or just delete manual level change, idk.

jkschneider: It's used in the little utility class `ClearCustomMetricDescriptors`, but OK to move to `compileOnly` as you suggest.

jkschneider:
Thanks for catching this @IRus!

PR #2057, opened 2020-05-06
Upgrade to Gradle Wrapper 6.4

izeye:
This PR upgrades to Gradle Wrapper 6.4.


jkschneider:
Thanks again @izeye!

shakuzen:
I'd like to apply this to `1.5.x` to keep unnecessary differences in the build between maintenance branches to a minimum given it should not affect users consuming Micrometer but can make maintenance more burdensome to have such differences between branches. Done with 94f0b35

PR #2062, opened 2020-05-06
Fix tag mismatch due to target tags in OkHttp metrics

izeye:
This PR fixes tag mismatch due to target tags in OkHttp metrics as it won't work with the Prometheus meter registry.

I also noticed that request tags feature won't work with the Prometheus meter registry, but I didn't fix it here as it's an optional feature and I couldn't find a nice way to fix it at the moment. So it'd be better to be handled separately.


izeye:
The request tags feature has been introduced in 1.2.x branch, so if we fix it, target branch will be 1.3.x anyway.

jkschneider:
Merged to master with c9efdafda24f3ff2199f6c614d00b3a55b9d7151.

PR #2065, opened 2020-05-06
Fix tag mismatch due to request tags in OkHttp metrics

izeye:
This PR is an attempt to fix the tag mismatch due to request tags mentioned in https://github.com/micrometer-metrics/micrometer/pull/2062.

I'm not sure if there's a better way with the current implementation. I'm also not sure if this is a bug or limitation with Prometheus as it's an optional feature. I just went with 1.5.x branch, but if this solution is acceptable and we want to fix it in 1.3.x branch, I can retarget its branch.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
@@ -26,7 +26,10 @@ import okhttp3.*;
 
 import java.io.IOException;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


jkschneider:
@izeye I think you've done the right thing. It's an optional feature that when used with Prometheus results in a bug? So perhaps we can retarget to 1.3.x and merge forward from there as you suggest.

izeye:
@jkschneider Thanks for the quick feedback! I changed the base branch to 1.3.x.

jkschneider:
Merged forward to 1.5.x with 131cf3731816b335d9bbbcc2acf237b64181dab8.

jkschneider:
Merged forward to master with ad74c2d70e0efe4cacff309b5a11965dadd9dc78.

Issue #1875, opened 2020-03-03
Add metrics for OkHttp ConnectionPool


benhubert:
Hi! I think it would be helpful to have metrics for OkHttp's [ConnectionPool](https://square.github.io/okhttp/3.x/okhttp/okhttp3/ConnectionPool.html) usage.

I already implemented this for us and created a PR. I'm uncertain about some implementation details, see my comments in the PR for details.

jkschneider:
@benhubert Would you like to open a PR with this implementation? Seems like you've got it working...

benhubert:
Well, that was my intention, but seems that I forgot to open the PR. Thanks for your hint.
Now it's ready for review: https://github.com/micrometer-metrics/micrometer/pull/2066

PR #2066, opened 2020-05-07
added metrics for OkHttp connection pool

benhubert:


jkschneider:
> What I don't like about it, that the idle value is a subset of total, so accessing okhttp.pool.connection.count without any tag won't give you the overall sum. Instead it will give you the overall sum + the idle connections, which is not a useful value.

I think there is a solution to this... Let's just subtract total from idle and tag it as `state=active`. Then the sum of the tags `active+idle = total`.

jkschneider:
Fixes #1875

jkschneider:
See if [this](https://github.com/micrometer-metrics/micrometer/commit/56ae455f1f3ec48f661824962bef69776f33fa64#diff-19f65912d4c22fc246aa20cbeb68bc9eR141-R165) makes sense @benhubert. We just hold total and idle together in a wrapper and use a `CountDownLatch` to refresh the stats whenever both gauges have been published and one tries to access the value on the next publish interval.

PR #2071, opened 2020-05-08
Polish KafkaMetrics

izeye:
This PR reintroduces part of ae045ebb8731e5572d82d0bf037786f3821535e8 that was reverted in d8b101e421bd4db8e3ad4cb38e6c7c1adb61973c.

It's trivial, and I'm okay with either way, but for future reference, I'd like to understand the reasons why those changes have been reverted.

Here are the reasons why I have proposed the changes in the first place:

- For the `includeCommonTags` flag, it is used not to include common tags when binding meters although they'll be deduplicated in the end. It's for consistency with others.
- For the change in `differentClient()`, it returns early if it's already confirmed that a metric is for a different client as the remaining checks aren't necessary. It's for efficiency.

This PR also changes to make tags be created as lazy as possible for efficiency.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
@@ -157,7 +157,6 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
                 }
 
                 String meterName = meterName(metric);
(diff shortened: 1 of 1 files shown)


izeye:
@jkschneider Could you please explain the reasons why those changes have been reverted?

jkschneider:
@izeye Purely merge hell, and not intentional! Sorry about that.

izeye:
@jkschneider Thanks for the quick feedback!

That's okay. I was trying to understand the reasons 😅

PR #2075, opened 2020-05-09
Fix ClassCastException in OpenTSDBMeterRegistry.writeLongTaskTimer()

izeye:
This PR fixes `ClassCastException` in `OpenTSDBMeterRegistry.writeLongTaskTimer()`.

This PR also changes as follows:

- Use snapshot for histogram counts in `writeTimer()`
- Use Prometheus variant for `LongTaskTimer` implementation

### DIFF
--- a/implementations/micrometer-registry-opentsdb/src/main/java/io/micrometer/opentsdb/OpenTSDBLongTaskTimer.java
+++ b/implementations/micrometer-registry-opentsdb/src/main/java/io/micrometer/opentsdb/OpenTSDBLongTaskTimer.java
@@ -1,0 +1,68 @@
+/**
+ * Copyright 2020 VMware, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-opentsdb/src/main/java/io/micrometer/opentsdb/OpenTSDBLongTaskTimer.java
```diff
@@ -0,0 +1,68 @@
...
+                snapshot.count(),
+                snapshot.total(),
+                snapshot.max(),
+                snapshot.percentileValues(),
+                Arrays.stream(snapshot.histogramCounts())
+                    .map(countAtBucket -> lastSnapshot == null ?
+                            countAtBucket :
+                            new CountAtBucket(
+                                    countAtBucket.bucket(),
+                                    countAtBucket.count() + lastSnapshot[i.getAndIncrement()].count()
```
shakuzen, line 59: I'm noticing behavior that doesn't look right related to this. If I scrape the prometheus endpoint while there is an active task tracked with a LongTaskTimer and histogram enabled, the bucket counts increase with each scrape. All other stats related to the LongTaskTimer (`+Inf` bucket, active count, duration sum, max) do not change over scrapes, and they go to zero when there are no active tasks. The histogram bucket counts (except for `+Inf`) do not go to zero.

PR #2076, opened 2020-05-09
Share cumulative histogram counts variant for LongTaskTimer

izeye:
This PR changes to share cumulative histogram counts variant for `LongTaskTimer`.


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusLongTaskTimer.java
```diff
@@ -16,53 +16,21 @@
...
  *
  * @author Jon Schneider
  * @since 1.5.0
+ * @deprecated since 1.5.2, use {@link CumulativeHistogramLongTaskTimer}.
  */
-public class PrometheusLongTaskTimer extends DefaultLongTaskTimer {
-    @Nullable
-    private CountAtBucket[] lastSnapshot;
-
+@Deprecated
```
jkschneider, line 31: Do you think these can be considered to be internal implementation details not subject to the same compatibility guarantees? Is there any scenario in which someone could break unless they explicitly casted the `LongTaskTimer` returned by `PrometheusMeterRegistry` to this type?

izeye: @jkschneider I just did this as it is `public`. Actually, I was wondering if registry-specifc meter implementations need to be `public` as some of them are `public`, but the others are package-private. If it's safe to assume that it's internal, I'll drop it completely. Let me know what you think.

jkschneider: @izeye Yes, I think they should have all been package private really.

izeye: @jkschneider Thanks for the feedback! I removed it.

izeye:
@jkschneider Thanks!

I was hesitating to apply the same changes to the `Timer` implementations due to some public APIs, but I'll give it a try.

PR #2082, opened 2020-05-10
Share cumulative histogram counts variant for Timer

izeye:
This PR is similar to https://github.com/micrometer-metrics/micrometer/pull/2076, but for the `Timer` implementations.


micrometer-core/src/main/java/io/micrometer/core/instrument/internal/HistogramFlavor.java
```diff
@@ -0,0 +1,27 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.internal;
```
jkschneider, line 16: `HistogramFlavor` is referred to in `PrometheusConfig`, so probably can't be internal.

izeye: @jkschneider Actually, I wanted to make this be referred in `PrometheusConfig`, but I couldn't do so due to backward compatibility. I created this for internal use instead. That's why I was hesitating to do this change as mentioned in https://github.com/micrometer-metrics/micrometer/pull/2076#issuecomment-626336733.

Alternatively, we could rename it to something like `HistogramType` or `HistogramStyle` and move it to `io.micrometer.core.instrument` or `io.micrometer.core.instrument.config` for deprecation and ultimate removal of `io.micrometer.prometheus.HistogramFlavor` and `io.micrometer.opentsdb.OpenTSDBFlavor`. Let me know what you think.

jkschneider: `OpenTSDBFlavor` is kind of reserved for a different purpose, more expansive than just histogram formatting. Since OpenTSDB is so wide-open on format (kind of like Influx), there are other conventions that may eventually become known to us, and for which we would add an option to `OpenTSDBFlavor`.

`HistogramFlavor` on the other I think will always apply only to Prometheus. It's because `VictoriaMetrics` reads a variant of Prometheus' exposition format with a particular expectation around histogram formatting that this option exists.

izeye: @jkschneider Thanks for the explanation! I didn't know the intention. So are you okay with as-is? Or replacing `io.micrometer.prometheus.HistogramFlavor` by renaming? Or any other better way you have in mind?

jkschneider: I'm not sure. Here's what I'm thinking: the commonality between these two timers isn't as much about the cumulative nature of the stats as it is about support for VictoriaMetrics format. And VictoriaMetrics isn't probably going to support some other cumulative monitoring system. 

For example, we could be adding a cumulative mode to the Datadog implementation now that it has rate functions (and distributions which conceptually map to Micrometer histograms). So it would seem awkward if Datadog couldn't use the `CumulativeHistogramTimer`?

jkschneider: Lastly, I'm a bit embarrassed about the capitalization miss on `HistogramFlavor` in 1.4.0. Sorry about that.

izeye: @jkschneider Thanks for the quick feedback! How about `PrometheusTimer` instead of `CumulativeHistogramTimer`? If it is still inappropriate, feel free to close this.

jkschneider: I'm on the fence on this. What do you think @shakuzen? This comes down to a tradeoff of duplication vs. implementation-specific types in core.

I did add `FixedBoundaryVictoriaMetricsHistogram` to core, judging that its inherent complexity tipped the scales in favor of deduplication.

shakuzen:
A couple things have changed since this was last reviewed that might make a different direction more favorable. `PrometheusMeterRegistry` now supports exemplars, and `CumulativeTimer` has been updated so it has a protected constructor that takes a `Histogram`. Extending `CumulativeTimer` could probably reduce the duplication in PrometheusTimer and OpenTSDBTimer enough that we don't need something else. `CumulativeTimer` is now used in the OtlpMeterRegistry. What do you think @izeye?

Issue #2069, opened 2020-05-08
Broken graphite metrics names with tags


workless:
Graphite tags always generate broken names due to DropWizard behavior.

From DropWizard, GraphiteReporter [1]:
`graphite.send(prefix(name, type.getCode()), format(value), timestamp);`
where
name="my.metric;tag=value"
type.getCode() = "p99"

final string sent to graphite: "my.metric;tag=valuep99"

From graphite logs
`08/05/2020 08:21:00 :: [tagdb] Tagging component.api.setStatus;host=mail.example.com.mean, component.api.sendTextMessage;host=mail.example.com.p95, component.api.join_meeting;host=mail.example.com.m1_rate, component.api.getConversations;host=mail.example.com.mean`

Unless we patch DropWizard side i can't see how this code may work

[1] https://github.com/dropwizard/metrics/blob/v4.1.7/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java#L351

workless:
@fitzoh @shakuzen 
What do you think it's the best approach?

fitzoh:
Yeah, saw your comment yesterday, should be able to dig in this weekend.
Hoping I can find an easy fix, if not the easy first step is probably updating defaults

fitzoh:
I opened a PR (https://github.com/dropwizard/metrics/pull/1576) to dropwizard metrics that should allow us to work around this.
No idea what kind of time frame we will be looking at for that to be merged or for a new release to be cut.


I agree with @workless that there isn't a clean way to fix this without modifying the dropwizard code.

The options I can think of:

* Wait for the dropwizard metrics maintainers to merge/cut a new release, then modify on the micrometer side
* fork/publish a modified version of the graphite reporter and work off that
* vendor the modified version of the graphite reporter in this repo and use that until an official release is out
* revert the default and/or add a note and/or remove the feature from micrometer until the dropwizard metrics release is available


Either way, the Spring Boot release team should probably be made aware, as I believe this feature is present in v2.3.0.RC1 (cc @snicoll)

jkschneider:
@fitzoh It may be the reason we need to cut ties with Dropwizard as a passthrough for Graphite. We had a similar situation with the Ganglia registry a while back.

fitzoh:
[...]

PR #2083, opened 2020-05-10
Prevent Graphite reporter from modifying user tags 

fitzoh:
# Context
As noted in #2069, the recently added Graphite tag support doesn't work for most metric types due to the way that the Dropwizard reporter appends metric attributes (min, max, p95, etc).

A metric that is reported this way using the traditional format:
`my.timer.host.foo.max` 
Is currently reported like this using the tag format:
`my.timer;host=foo.max`

https://github.com/dropwizard/metrics/pull/1576 was recently merged on the reporter side, adding a new flag to the reporter which would cause it to report the example used above as `my.timer;host=foo;metricattribute=max`.

# What this PR does now
This PR adds a failing test to the end to end `GraphiteMeterRegistryTest` showing what the modified graphite output will eventually be. 

It may also serve as a discussion starter around if the upstream changes to the Dropwizard metrics library are appropriate, or if more changes are required/desirable (is a `metricattribute` tag? Does it need to be more configurable?).

# What this PR should do before merging
Pull in a Dropwizard metrics release containing https://github.com/dropwizard/metrics/pull/1576  and ensure that the Graphite reporter is configured to use the tag format when appropriate.

### DIFF
--- a/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteMeterRegistry.java
+++ b/implementations/micrometer-registry-graphite/src/main/java/io/micrometer/graphite/GraphiteMeterRegistry.java
@@ -59,6 +59,7 @@ public class GraphiteMeterRegistry extends DropwizardMeterRegistry {
                 .withClock(new DropwizardClock(clock))
                 .convertRatesTo(config.rateUnits())
                 .convertDurationsTo(config.durationUnits())
+                .addMetricAttributesAsTags(config.graphiteTagsEnabled())
                 .build(getGraphiteSender(config));
     }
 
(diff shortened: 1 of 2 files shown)


fitzoh:
Force pushed a minor modification to the test and updated the production code to work with `io.dropwizard.metrics:metrics-graphite:4.1.8-SNAPSHOT` (built locally, it doesn't appear that SNAPSHOTS are being published).

fitzoh:
[Dropwizard Metrics release](https://github.com/dropwizard/metrics/releases/tag/v4.1.8) is out and the build is green after forcing a rebuild.

shakuzen:
On further consideration, due to the hard requirement on a recent patch version, it would likely cause more support issues to merge this into `1.5.x`. I will make a release note that lets users know how they can work around the issue by providing their own reporter with this property set rather than deferring to the default one we create.

PR #2094, opened 2020-05-13
Fix FunctionCounter for AtlasMeterRegistry

brharrington:
Before it was reporting the values as a gauge and did not
match other counters. This changes it to the Spectator
helper to update a counter so the measured rate of change
will be consistent with a normal counter.


jkschneider:
Merged to 1.3.x with f62f2125435b152528462f892722f6bd2d798c0d.

jkschneider:
Merged forward to 1.5.x with c047343f616ec55f074b3f18335a9455e2bc4b7e.

jkschneider:
Merged forward to master with 00c2771419a6fce29d9285136b9a1e9ab00582a8.

jkschneider:
Thanks @brharrington. This should go out with LTS patches in the 1.3.x and 1.5.x lines as well as the next minor release.

PR #2101, opened 2020-05-15
Update naming convention consistently in StatsdLineBuilders

izeye:
This PR changes to update naming convention consistently in `StatsdLineBuilder` implementations. For example, with the current implementations, `name` could be `null` if `line()` is accessed concurrently.


jkschneider:
Merged forward to 1.3.x with b425d2221e3c24ddc28181ffd4fd023ffbc0d99f.

Issue #2087, opened 2020-05-11
Deadlock due to ConcurrentHashMap.compute in PrometheusMeterRegistry


krm1312:
We just tried to upgrade a relatively large application to spring boot 2.3.0.RC1 which includes micrometer 1.5.0.  The application relies on redis and we have included various metrics for lettuce (redis).

PrometheusMeterRegistry has been modified to use ConcurrentHashMap.compute() at https://github.com/micrometer-metrics/micrometer/blob/93375782923415f1c947f477557ec8b49354f8eb/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L415.

I believe this causing a deadlock for us now in that getting the value of a particular gauge (spring boot health - https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#howto-map-health-indicators-to-metrics) ends up (in different threads) trying to register other timers in the registry which deadlocks.

This doesn't seem like too uncommon of a use-case where getting the value of one meter might trigger initialization of another meter.  It does seem a bit strange to me that registering a meter would trigger getting its value, but, I'm not sure if that changed.

```
"main@1" prio=5 tid=0x1 nid=NA waiting
  java.lang.Thread.State: WAITING
	 blocks lettuce-kqueueEventLoop-16-1@19158
	 blocks lettuce-kqueueEventLoop-17-1@19751
	  at jdk.internal.misc.Unsafe.park(Unsafe.java:-1)
	  at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:234)
	  at java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1798)
	  at java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3128)
	  at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1868)
	  at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2021)
	  at io.lettuce.core.protocol.AsyncCommand.await(AsyncCommand.java:83)
	  at io.lettuce.core.LettuceFutures.awaitOrCancel(LettuceFutures.java:118)
	  at io.lettuce.core.FutureSyncInvocationHandler.handleInvocation(FutureSyncInvocationHandler.java:69)
	  at io.lettuce.core.internal.AbstractInvocationHandler.invoke(AbstractInvocationHandler.java:80)
[...]

PR #2103, opened 2020-05-16
Make MicrometerCollector implement Collector.Describable

izeye:
This PR changes to make `MicrometerCollector` implement `Collector.Describable`.

Closes gh-2087

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
@@ -27,20 +27,23 @@ import java.util.stream.Stream;
 import static java.util.stream.Collectors.toList;
 
 /**
+ * {@link Collector} for Micrometer.
+ *
  * @author Jon Schneider
+ * @author Johnny Lim
  */
-class MicrometerCollector extends Collector {
+class MicrometerCollector extends Collector implements Collector.Describable {
     private final Meter.Id id;
     private final Map<List<String>, Child> children = new ConcurrentHashMap<>();
     private final String conventionName;
     private final List<String> tagKeys;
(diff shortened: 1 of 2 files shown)


izeye:
For the record, I chose to use `Type.HISTOGRAM` unconditionally as it includes all the names of `Type.SUMMARY`.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
```diff
@@ -77,6 +78,37 @@ public boolean isEmpty() {
...
+                return Collections.singletonList(
+                        new MetricFamilySamples(conventionName, Type.COUNTER, help, Collections.emptyList()));
+
+            case GAUGE:
+                return Collections.singletonList(
+                        new MetricFamilySamples(conventionName, Type.GAUGE, help, Collections.emptyList()));
+
+            case TIMER:
+                return Collections.singletonList(
+                        new MetricFamilySamples(conventionName, Type.HISTOGRAM, help, Collections.emptyList()));
```
shakuzen, line 94: Also need the max GAUGE like the distribution summary

shakuzen: We should increase test cases for filtering by time series name to ensure we're not missing any like this. For example, with a Timer named `my.timer` and attempting to filter by `my_timer_duration_seconds_max` (`PrometheusMeterRegistryTest` uses the `PrometheusDurationNamingConvention` by default) does not work as is right now.

izeye: @shakuzen Thanks for pointing it out! I changed as suggested.

PR #2107, opened 2020-05-17
Polish build scripts for Spring Boot samples

izeye:
This PR upgrades to Spring Boot 2.3.0.RELEASE for samples.

This PR also polishes build scripts for Spring Boot samples.

### DIFF
--- a/samples/micrometer-samples-boot2-reactive/build.gradle
+++ b/samples/micrometer-samples-boot2-reactive/build.gradle
@@ -1,13 +1,9 @@
 plugins {
     id 'java'
-    id "io.spring.dependency-management" version "1.0.9.RELEASE"
+    id 'org.springframework.boot' version '2.2.7.RELEASE'
 }
 
-dependencyManagement {
-    imports {
-        mavenBom 'org.springframework.boot:spring-boot-dependencies:2.2.7.RELEASE'
-    }
-}
+apply plugin: 'io.spring.dependency-management'
 
 dependencies {
     implementation project(":micrometer-core")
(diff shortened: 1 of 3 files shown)


izeye:
The build for this PR has failed with the following compile error:

```
> Task :micrometer-samples-boot2:compileJava
/home/circleci/micrometer/samples/micrometer-samples-boot2/src/main/java/io/micrometer/boot2/samples/components/ServiceLevelObjectiveConfiguration.java:62: error: incompatible types: String is not a functional interface
                                        s -> s.name("http.server.requests").tag("uri", uri -> uri.startsWith("/api")),
                                                                           ^
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error
```

Based on the error message, it seemed to reference Spring Boot-managed `micrometer-core`, not project one.

I couldn't reproduce this locally.

I repurposed this PR to just polishing build scripts only as I have no clue for how this happens only on CircleCI at the moment, and I couldn't find an easy way to debug this.

checketts:
What JVM version was Circle running? I recall hitting something like this with a Java 8 (specific update) version a long time ago.

izeye:
@checketts Thanks for the feedback!

It's using JDK 14 (`cimg/openjdk:14.0`), and I was trying to reproduce it with JDK 14 (`14.0.0.hs-adpt`) on macOS.

See https://app.circleci.com/pipelines/github/izeye/micrometer/513/workflows/f5bfc896-3d5d-47e0-b883-588dfa37c68c/jobs/871/parallel-runs/0/steps/0-0

shakuzen:
So there's an issue with building the samples against Spring Boot 2.3 still (at least on CI)? If so, let's create an issue so we can track that. I'll get this merged in the meantime.

izeye:
@shakuzen I created #2129 to demonstrate how it fails.

PR #2108, opened 2020-05-17
Add level2 cache region name as a tag in HibernateMetrics

### LINKED ISSUE #2097: Expose per-region statistics for Hibernate L2 cache
As of Micrometer 1.5, the `HibernateMetrics` binder exposes L2 cache statistics from the `org.hibernate.stat.Statistics` interface. This is useful, but it would be better if this were done on a per-cache-region basis, giving us more fine-grained per-entity and per-collection information.

This can be done by using `getSecondLevelCacheRegionNames` and `getSecondLevelCacheStatistics`, giving you a `SecondLevelCacheStatistics` object.

As well as providing the same hit/miss/put stats you get from the higher-level `Statistics` interface, you also get cache size and memory usage stats. These can then be tagged with the region name.

kennymacleod:
As suggested in #2097, we can change `HibernateMetrics` so that rather than just exposing counters for the hits/misses/puts of the Level2 cache as a whole, we can drill down into individual cache regions.

This allows us to add the cache region as a tag to each meter, and also allows us to additionally expose the region element count and memory usage.

Noe that in this PR I've removed the existing counters for top-level hits/misses/puts, and added new ones on a per-region basis. This assumes that the top-level count is equal to the sum of the individual regions. I've checked the Hibernate source, and this is indeed the case, e.g. when the hit count is incremented for a region, the top-level counter is also incremented.

Therefore, existing users of these meters will see no change if they ignore the new region tags.

Note that I've targetted the master branch here. I could retarget at 1.5.x if required.

Resolves #2097

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
@@ -16,7 +16,6 @@
 package io.micrometer.core.instrument.binder.jpa;
 
 import io.micrometer.core.instrument.FunctionCounter;
-import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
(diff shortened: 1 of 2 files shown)


kennymacleod:
I've reverted the addition of the new gauges for getting the region size and element count, as those entail invoking methods on the `Region` objct, and depending on the region implementation (e.g. a distributed cache) that can be very expensive and entail a network call.

shakuzen (approved):
Looks good to me. Thank you.

PR #2113, opened 2020-05-20
feat(influx): Add support for InfluxDB v2

### LINKED ISSUE #1974: InfluxDB 2.0 API support
Looks like the 'direct' api for 1.x will not work with 2.0.
Thankfully the changes seem pretty small, for comparison [1.x write api](https://docs.influxdata.com/influxdb/v1.7/tools/api/#write-http-endpoint) and [2.x write api](https://v2.docs.influxdata.com/v2.0/api/#operation/PostWrite)

The body remains the same (line protocol), only major change I believe is for auth (though user + password is probably also possible), instead of `db` query param there is `org` and `bucket`.
There is also an official [java client library](https://github.com/influxdata/influxdb-client-java), though I am not sure if micrometer needs all the functionality it provides.

Sorry if there is already something in place to configure micrometer so it works with influxdb 2.0 (directly), but I did not find it.

bednar:
Closes #1974
Closes #1423

Added support for InfluxDB v2. 

This change introduces new parameters for `Influx` registry:

- `org` - Specifies the destination organization for writes.
  - only for InfluxDB v2
- `bucket` - Specifies the destination bucket for writes.
  - only for InfluxDB v2
  - `bucket` is alias for `db` => user can specify either `db` nor `bucket`
- `token` - Authentication token to authorize API requests.
  - support both version of InfluxDB v2

The changes of existing parameters:

- `retentionDuration` - For the InfluxDB v2 the value should be an amount of seconds (e.g. 3600) and it's used when `autoCreateDb` is enabled
- `autoCreateDb` - This also support the InfluxDB v2.


bednar:
Hi @jkschneider,

What is an ideal way to parse JSON response in a registry? Could I use `json-path` or try implement it without additionally dependencies?

https://github.com/micrometer-metrics/micrometer/blob/c68525346b9d6f3b8758b1f43b3b4120b199ca95/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java#L347

https://github.com/micrometer-metrics/micrometer/blob/c68525346b9d6f3b8758b1f43b3b4120b199ca95/implementations/micrometer-registry-influx/build.gradle#L5

Regards

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -283,4 +414,40 @@ public String toString() {
...
+                if (config.token() != null) {
+                    requestBuilder.withHeader("Authorization", "Bearer " + config.token());
+                }
+            }
+        },
+        V2 {
+            @Override
+            String writeEndpoint(final InfluxMeterRegistry registry) throws UnsupportedEncodingException {
+                InfluxConfig config = registry.config;
+                return config.uri() + "/api/v2/write?&precision=ms&bucket=" + URLEncoder.encode(config.bucket(), "UTF-8") + "&org=" + URLEncoder.encode(registry.org, "UTF-8");
```
DangerousDarlow, line 437: InfluxDB supports variable timestamp precision. Looking at the Micrometer documentation this doesn't seem to be an option?

bednar: Yeah. Currently the Micrometer only supports millisecond precision for timestamp. See - https://github.com/micrometer-metrics/micrometer/blob/07701bd311af40419964a54add6ae60b3cfdf29b/micrometer-core/src/main/java/io/micrometer/core/instrument/Clock.java#L47

bednar:
Hi @shakuzen, @jkschneider from my POV the PR is prepared to review. Is there anything else what can I do?

Thanks for your review.

timhallinflux:
any progress on a review here?

timhallinflux:
bump

bednar:
Sources are rebased with master. 

Could we do something to review this PR?

PiotrNowicki:
Guys, any updates on this or plans when Influx DB v2 will be supported?

ivankudibal:
Hi Micrometer, is there any plan to merge this PR?

fmmatthewzeemann:
Hi Micrometer, any updates on merging this PR? Really need  InfluxDB v2 support ...

vertex-github:
+1 on this - Influxdb2 is up to v2.0.2 - Id really like to port our data storage to the new engine from our SpringBoot apps

murphytalk:
+1 
hope to see the InfluxDbv2 solution gets merged sooner.

timhallinflux:
bump

pivotal-issuemaster:
@bednar Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2113)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2113) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@bednar Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2113)!

overshareware:
Any timeline on this being merged? Desperate for v2 support.

LimmaPaulus:
I have been using this PR with manual build for a long time on production. It has worked flawlessly.

vertex-github:
Hey folks, is there something missing from the PR or an issue with the approach taken or is this just pending due to lack of resources for review / merge?

shakuzen:
I've added the issue #1974 to the 1.7 milestone so we make sure to review this in that time frame. There's clearly a lot of interest in InfluxDB v2 support, so this will be prioritized.

bednar:
@shakuzen thanks for awesome news. The PR is rebased and ready to review.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxConfig.java
```diff
@@ -119,18 +124,65 @@ default boolean compressed() {
...
+     */
+    @Nullable
+    default String org() {
+        return getString(this, "org").orElse(null);
+    }
+
+    /**
+     * Specifies the destination bucket for writes. Takes either the ID or Name interchangeably.
+     * See detail info: <a href="https://v2.docs.influxdata.com/v2.0/organizations/buckets/view-buckets/">How to retrieve the <i>bucket</i> parameter in the InfluxDB UI.</a>
+     * @return The destination organization for writes.
```
shakuzen, line 148:
```suggestion
     * @return The destination bucket (or db) for writes.
```

bednar: Fixed

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -75,43 +85,86 @@ public static Builder builder(InfluxConfig config) {
...
 
+    private void createBucketIfNecessary() {
+
+        try {
+            String createBucketJSON = new CreateBucketJSONBuilder(config.bucket(), org)
+                    .setEverySeconds(config.retentionDuration())
+                    .build();
+
+            HttpSender.Request.Builder requestBuilder = httpClient
+                    .post(config.uri() + "/api/v2/bucket");
```
shakuzen, line 135:
```suggestion
                    .post(config.uri() + "/api/v2/buckets");
```

Unfortunately, it seems the Influx v2 API returns a 200 response code even though the URI is wrong. We should add some tests with an InfluxDB instance running, using testcontainers, to ensure things are working as expected now and in the future. That doesn't need to block this pull request; I can follow-up with that work after merging this.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/CreateBucketJSONBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+import io.micrometer.core.lang.Nullable;
+
+/**
+ * Builds a JSON Bucket representation for InfluxDB 2.
+ *
+ * @author Jakub Bednar
+ */
+class CreateBucketJSONBuilder {
+
+    private static final String BUCKET_TEMPLATE = "{\"name\":\"%s\",\"orgID\":\"%s\",\"retentionRules\":[%s]}";
```
shakuzen, line 27: It seems we need to generate a valid orgID, which is different from the org name. Passing an org name here results in the following error being returned:

```
{
	"code": "invalid",
	"message": "failed to unmarshal json: id must have a length of 16 bytes"
}
```

bednar: We have to check if the `org` parameter is `ID` otherwise retrieve `ID` by something like: [retrieveOrg](https://github.com/micrometer-metrics/micrometer/blob/229dcc3f4f2ab79e04fc85e1c22379323ee589cf/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java#L338)
or It is a little bit out of scope of and we can remove the support for creating Bucket.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
bednar: Fixed `bucket` -> `buckets`. Thanks.

I will prepare an integration test over https://www.testcontainers.org/modules/databases/influxdb/.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/CreateBucketJSONBuilder.java
shakuzen: > It is a little bit out of scope of and we can remove the support for creating Bucket.

I'm leaning towards this direction. Perhaps we can revisit that option later. For initial v2 support, I think it is fine to not have it. Most companies will want to configure a retention policy on the bucket when creating it, and they will make tokens with specific access, so it seems highly likely that the token configured in the app won't have permission to create buckets anyways.

bednar: Ok, I will remove support for creating Bucket.

bednar: The support for creating `Bucket` is removed.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
bednar: The support for creating `Bucket` is removed. 

Do you want to create integration tests over `testcontainers` for publishing metrics?

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -232,6 +249,84 @@ private String influxLineProtocol(Meter.Id id, String metricType, Stream<Field>
...
+                    c -> StepRegistryConfig.validate(c),
+                    checkRequired("token", InfluxConfig::token).andThen(Validated::nonBlank),
+                    checkRequired("org", (Function<InfluxConfig, String>) it -> org)
+                            .andThen(Validated::nonBlank)
+            ).orThrow();
+        }
+    }
+
+    private void retrieveOrg() {
+        String uri = config.uri() + "/api/v2/orgs";
```
shakuzen, line 303: As far as I can tell, only an admin or all access token can read the orgs. A read/write token (which is what I would expect to be provided for applications to write metrics) will not get a response with any orgs, even the one it belongs to. Here is what I get back with a read/write token:

```
{
    "links": {
        "self": "/api/v2/orgs"
    },
    "orgs": []
}
```

While it would be nice for users to have one less thing to configure, I am not sure it is worth having this attempt to retrieve the org. It also adds the JSON parser dependency. Users will have to configure the bucket and token anyways, so it isn't like we can provide something that works with defaults alone.

What do you think about removing the retrieve org logic?

micrometer-core/src/main/java/io/micrometer/core/ipc/http/HttpSender.java
```diff
@@ -288,12 +291,19 @@ public Response send() throws Throwable {
 
     class Response {
         public static final String NO_RESPONSE_BODY = "<no response body>";
+        private static final Map<String, List<String>> NO_RESPONSE_HEADERS = Collections.unmodifiableMap(new HashMap<>());
```
shakuzen, line 294: Why not `Collections.emptyMap()`?

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -232,6 +249,84 @@ private String influxLineProtocol(Meter.Id id, String metricType, Stream<Field>
...
+                                .orElse(InfluxDBVersion.V2);
+                        logger.debug("InfluxDB version configured to: '{}'", influxDBVersion);
+                    })
+                    .onError(response -> {
+                        logger.error("unable to ping InfluxDB: '{}'. Use v2 API.: {}", uri, response.body());
+                        // Unable ping http://server/ping => use v2 API
+                        influxDBVersion = InfluxDBVersion.V2;
+                    });
+        } catch (Throwable e) {
+            logger.error("unable to ping InfluxDB: '{}', Use v1 API.", uri, e);
```
shakuzen, line 283: What exception are we anticipating that would end up in this catch block? We are defaulting to v2 above if there is an error calling the ping endpoint, so why is this defaulting to v1?

shakuzen (changes requested):
We should note that userName and password are only applicable to InfluxDB 1.x. I've left some other comments in the code.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
shakuzen: I've created #2504 to track adding the integration tests. I don't want to block merging this enhancement on that. We want to include support for InfluxDB 2 in the 1.7 milestone release tomorrow if possible.

bednar: Ok

micrometer-core/src/main/java/io/micrometer/core/ipc/http/HttpSender.java
bednar: Fixed

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
bednar: > What exception are we anticipating that would end up in this catch block? 

Defaulting version to `v1`.

> We are defaulting to v2 above if there is an error calling the ping endpoint, so why is this defaulting to v1?

This implementation was prepared for early version of InfluxDB v2 which didn't supports the `/ping` endpoint. For the current version it doesn't sense => I changed the implementation. Now the the default version is set to `v1`.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
bednar: > What do you think about removing the retrieve org logic?

I agree. 

It is same case as a creating a `bucket` => the support for retrieve `org` is removed.

bednar:
> We should note that userName and password are only applicable to InfluxDB 1.x. I've left some other comments in the code.

I updated docs for `userName` and `password` with: _The authenticating by 'userName' and 'password' is not supported for the InfluxDB v2._

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
shakuzen: A problem with only determining the version from the results of a ping endpoint is that if the ping request happens to fail on the first publish, then the version will be set to v1 for the application, even if you wanted v2 and configured things for v2 and the influxdb backend is available after the ping request.

Since we aren't trying to retrieve the org anymore, perhaps we should add a config for `apiVersion` and have it default to v2 if `org` is configured, v1 otherwise. This way users can explicitly set the influx API version they intend to target.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -101,17 +111,21 @@ private void createDatabaseIfNecessary() {
...
     @Override
     protected void publish() {
+        recognizeInfluxDBVersion();
         createDatabaseIfNecessary();
 
         try {
-            String influxEndpoint = config.uri() + "/write?consistency=" + config.consistency().toString().toLowerCase() + "&precision=ms&db=" + config.db();
+            String influxEndpoint = influxDBVersion.writeEndpoint(this);
             if (StringUtils.isNotBlank(config.retentionPolicy())) {
                 influxEndpoint += "&rp=" + config.retentionPolicy();
```
shakuzen, line 120: While not harmful (the write request is still successful with this), `rp` is not a parameter that the v2 write endpoint recognizes.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
bednar: Changed. The determining of InfluxDB version is based on the config.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
bednar: Fixed

micrometer-core/src/main/java/io/micrometer/core/ipc/http/HttpSender.java
```diff
@@ -304,6 +313,10 @@ public String body() {
             return body;
         }
 
+        public Map<String, List<String>> headers() {
```
shakuzen, line 316: Since this isn't used anymore, we can get rid of these changes until we have a use for them, I think. If there is some use outside of supporting InfluxDB v2, we can consider them in a separate pull request.

shakuzen (approved):
Thank you very much @bednar for taking care of all the review so quickly and for the work in the first place. It is very much appreciated. I think this is almost ready to be merged. I will run a few more tests with it after I get some sleep, and then I can do any polishing along with merging it.

micrometer-core/src/main/java/io/micrometer/core/ipc/http/HttpSender.java
bednar: Fixed

implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryVersionsTest.java
```diff
@@ -0,0 +1,395 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for check compatibility of {@link InfluxMeterRegistry} with InfluxDB v1 and InfluxDB v2.
+ *
+ * @author Jakub Bednar (19/05/2020 09:00)
+ */
+@ExtendWith(WiremockResolver.class)
+public class InfluxMeterRegistryVersionsTest {
```
jonatan-ivanov, line 49: We should add tests that test `Basic` and `Token`, right now only `Bearer` auth is tested.

implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryVersionsTest.java
```diff
@@ -0,0 +1,395 @@
...
+                .withHeader("Authorization", equalTo("Bearer my-token")));
+    }
+
+    @Test
+    void writeToV2TokenRequired(@WiremockResolver.Wiremock WireMockServer server) {
+
+        server.stubFor(any(urlPathEqualTo("/api/v2/write"))
+                .willReturn(aResponse().withStatus(204)));
+
+        InfluxConfig config = new InfluxConfig() {
```
jonatan-ivanov, line 97: I would use a mock instead of implementing it.

jonatan-ivanov:
I talked to @shakuzen, let's merge this now and polish in a follow-up PR (see my comments).

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -85,9 +85,12 @@ private void createDatabaseIfNecessary() {
                     .setRetentionReplicationFactor(config.retentionReplicationFactor())
                     .setRetentionShardDuration(config.retentionShardDuration()).build();
 
-            httpClient
+            HttpSender.Request.Builder requestBuilder = httpClient
                     .post(config.uri() + "/query?q=" + URLEncoder.encode(createDatabaseQuery, "UTF-8"))
-                    .withBasicAuthentication(config.userName(), config.password())
+                    .withBasicAuthentication(config.userName(), config.password());
+            config.apiVersion().addHeaderToken(config, requestBuilder);
```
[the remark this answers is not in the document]
shakuzen: I was curious about this too in my testing. It will overwrite the `Authorization` header, which is the behavior we want, but we should make sure we are testing for it, to prevent any bugs from that behavior potentially changing in the future. I will add a note to the JavaDoc for the `userName` and `password` config methods to mention they won't be used if a token is configured.

> I would do this login this class not in the config.

Could you clarify what you mean?

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxApiVersion.java
```diff
@@ -0,0 +1,65 @@
...
+        String writeEndpoint(final InfluxConfig config) {
+            String influxEndpoint = config.uri() + "/write?consistency=" + config.consistency().toString().toLowerCase() + "&precision=ms&db=" + config.db();
+            if (StringUtils.isNotBlank(config.retentionPolicy())) {
+                influxEndpoint += "&rp=" + config.retentionPolicy();
+            }
+            return influxEndpoint;
+        }
+
+        @Override
+        void addHeaderToken(final InfluxConfig config, final HttpSender.Request.Builder requestBuilder) {
```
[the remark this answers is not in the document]
shakuzen: I suppose the problem is that if the `token` isn't configured, then a header shouldn't be added. That logic would need to be implemented at the callsite instead of here in that case.

implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryVersionsTest.java
shakuzen: I don't think we mock config interfaces in other tests, so this is more in line with the code base.

implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxMeterRegistryVersionsTest.java
shakuzen: There's a test for `Token` already called `writeToV2Token`. I'll add tests for basic auth.

implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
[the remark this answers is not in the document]
jonatan-ivanov: typo, I wanted to write this: I would do this logic in this class not in the config.

PR #2114, opened 2020-05-20
Add reason parameter to meter registration failure listener

izeye:
This PR adds a reason parameter to meter registration failure listener as failures can vary theoretically, and it's useful to know the reason for listener writers.


jkschneider:
Good idea!

PR #2115, opened 2020-05-21
Polish timer builder changes

izeye:
This PR polishes timer builder changes.

> - Is the name `Timer.ResourceSample` as bad as I feel it is?

While doing this, I was wondering if `Timer.AutoCloseableSample` or `Timer.AutoCloseableBuilder` might be easier to understand. I didn't do so as I wasn't sure.


izeye:
Oops, I missed de72730c41b34e122e3634856fdf7c25db624940, and I didn't know that the overrides were intentional for binary backward compatibility. I restored them in d84489a.

Issue #1758, opened 2019-12-17
Add possibility to filter out metrics by name in prometheus scrape


dellekappa:
The CollectorRegistry provided by Prometheus offers the `filteredMetricFamilySamples` method that gives the option to scrape out only some specific metric samples.

This option could be usefull in order to expose a parametric scrape endpoint that can will return only the values of interests instead of the whole registry.

The implementation is quite easy. It's enough to overload the `scrape` methods in `PrometheusMeterRegistry` with an additional parameter `Set<String> includedNames`

shakuzen:
If the intent is to filter metrics exposed to Prometheus, you can already do this by configuring a `MeterFilter` on the `PrometheusMeterRegistry` that filters by meter name. Does that work for your use case?

PR #2119, opened 2020-05-23
Add convenient methods for filtered scraping

izeye:
This PR adds convenient methods for filtered scraping.

This PR also adds a sample for filtered scraping.

See gh-1758


jkschneider:
Nice feature!

PR #2120, opened 2020-05-23
Prevent StringIndexOutOfBoundsException in Datadog statsD on empty tag value

agarbayo:
I'm unsure if this is the best approach. For example I noticed that on empty tag value InfluxDB seems to drop the tag.
Given guidance on the right approach, I'm happy to modify the PR.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/DatadogStatsdLineBuilder.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/DatadogStatsdLineBuilder.java
@@ -75,6 +75,9 @@ public class DatadogStatsdLineBuilder extends FlavorStatsdLineBuilder {
     }
 
     private String sanitizeTagValue(String value) {
+        if (value.isEmpty()) {
+            return value;
+        }
         return (value.charAt(value.length() - 1) == ':') ? value.substring(0, value.length() - 1) + '_' : value;
     }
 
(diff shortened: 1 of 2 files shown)


shakuzen:
Thanks for reporting this and providing a pull request. I don't think Datadog accepts empty/blank tag values like that. From https://docs.datadoghq.com/tagging/

> **Note**: A tag cannot end with a colon, for example `tag:`

If I'm understanding the [example here](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics#send-metrics-using-dogstatsd-and-the-shell), a tag without a value (`shell` in the example) should be sent without a colon at all. I don't think we've ever supported this since Micrometer tags have values, even if they're blank. But it seems like the right thing to do here is interpret an empty tag value as being a valueless tag in dogstatsd.

We'll want to fix this in `1.1.x` so we can fix it in upcoming patch versions. Would you be willing to update the pull request to interpret empty tag values as valueless tags in DogstatsdLineBuidler?

agarbayo:
Thank you for the information. 

Yes, I will update shortly the PR to interpret empty tag values as valueless tags.

mathurn:
Hi, apologies if this is not the right place to ask but is this change part of the `1.1.15` release? If so, when will that release be going live?

shakuzen:
> is this change part of the `1.1.15` release? If so, when will that release be going live?

@mathurn Yes, as indicated by this issue being assigned to the `1.1.15` milestone, it will be included in that release. We will be working on getting that release out very soon (as early as today). Please feel free to try snapshots of the release `1.1.15-SNAPSHOT` to ensure the fix you're waiting on works as expected.

PR #2136, opened 2020-06-03
Add support for strong reference to TimeGauge

izeye:
This PR adds support for strong reference to `TimeGauge` similar to `Gauge`.

This PR also changes to use it in `NewRelicMeterRegistryTest` as it seems to be useful for the changes made in #2135.


izeye:
I rebased this onto the current `master` branch to resolve the conflicts and updated Javadoc `@since` tag and `@Incubating` `since` attribute to be 1.7.0 as @jkschneider has assigned it to the 1.7.0 milestone.

Issue #2138, opened 2020-06-04
StatsdMeterRegistry sends multiple duplicated count measurements


chengzp:
Micrometer version: 1.5.1

As time progresses, `StatsdMeterRegistry` seems to be creating more and more udp connections. A single counter increment is then sent to all those connections, leading to duplicate counts. For example, in one particular case, the ngrep for a specific counter increment shows 7 identical messages (the message contents are omitted):
```
U 172.30.23.125:40280 -> 172.31.23.196:8125 #12277
U 172.30.23.125:55138 -> 172.31.23.196:8125 #12278
U 172.30.23.125:50911 -> 172.31.19.45:8125 #12279
U 172.30.23.125:58785 -> 172.31.25.42:8125 #12280
U 172.30.23.125:46418 -> 172.31.25.42:8125 #12281
U 172.30.23.125:45846 -> 172.31.24.89:8125 #12282
U 172.30.23.125:47359 -> 172.31.24.89:8125 #12283
```
(The `172.31.*.*` IP addresses are what our statsd server's hostname resolves to.)

We also saw this:
```
$ jcmd 168 GC.class_histogram | grep StatsdMeterRegistry
1036:            13            208  io.micrometer.statsd.StatsdMeterRegistry$$Lambda$152/0x00000001002e8040
1105:             7            168  io.micrometer.statsd.StatsdMeterRegistry$$Lambda$113/0x0000000100250440
1316:             7            112  io.micrometer.statsd.StatsdMeterRegistry$$Lambda$122/0x0000000100255c40
1334:             1            104  io.micrometer.statsd.StatsdMeterRegistry
2583:             1             24  <redacted>.StatsdMeterRegistryFactory$build$config$1
2696:             1             24  io.micrometer.statsd.StatsdMeterRegistry$$Lambda$112/0x0000000100250040
2989:             1             16  <redacted>.StatsdMeterRegistryFactory
3368:             1             16  io.micrometer.statsd.StatsdMeterRegistry$$Lambda$110/0x0000000100229c40
3369:             1             16  io.micrometer.statsd.StatsdMeterRegistry$$Lambda$164/0x00000001002ea840
3370:             1             16  io.micrometer.statsd.StatsdMeterRegistry$$Lambda$201/0x000000010032e840
3371:             1             16  io.micrometer.statsd.StatsdMeterRegistry$$Lambda$77/0x0000000100213c40
```
[...]

PR #2142, opened 2020-06-06
Dispose old client when new one is created

chengzp:
Attempt at fixing https://github.com/micrometer-metrics/micrometer/issues/2138


jkschneider:
Thanks, we'll merge #2151 forward to master separately.

Issue #2095, opened 2020-05-14
New Kafka Metrics duplicate metrics


tchiotludo:
With the last kafka metrics on `implementation "io.micrometer:micrometer-core:1.4.2"`, I've just got a new error that crash my application : 
```
2020-05-14 08:23:11,061 ERROR akfa-queue-1 tionHandlers Caught an exception in Thread[kakfa-queue-1,5,main]. Shutting down.
java.lang.IllegalArgumentException: Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'kafka_consumer_fetch_manager_bytes_consumed_total' containing tag keys [client_id, kafka_version, topic]. The meter you are attempting to register has keys [client_id, kafka_version].
```

<details><summary>stack trace</summary><p>

	at io.micrometer.prometheus.PrometheusMeterRegistry.lambda$collectorByName$9(PrometheusMeterRegistry.java:382)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1932)
	at io.micrometer.prometheus.PrometheusMeterRegistry.collectorByName(PrometheusMeterRegistry.java:369)
	at io.micrometer.prometheus.PrometheusMeterRegistry.newFunctionCounter(PrometheusMeterRegistry.java:280)
	at io.micrometer.core.instrument.MeterRegistry$More.lambda$counter$1(MeterRegistry.java:866)
	at io.micrometer.core.instrument.MeterRegistry.lambda$registerMeterIfNecessary$5(MeterRegistry.java:541)
	at io.micrometer.core.instrument.MeterRegistry.getOrCreateMeter(MeterRegistry.java:594)
	at io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:548)
	at io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:541)
	at io.micrometer.core.instrument.MeterRegistry.access$500(MeterRegistry.java:76)
	at io.micrometer.core.instrument.MeterRegistry$More.counter(MeterRegistry.java:866)
	at io.micrometer.core.instrument.FunctionCounter$Builder.register(FunctionCounter.java:122)
	at io.micrometer.core.instrument.composite.CompositeFunctionCounter.registerNewMeter(CompositeFunctionCounter.java:58)
[...]

PR #2143, opened 2020-06-08
Add option to include client ID in meter names for Kafka

izeye:
This PR adds option to include client ID in meter names for Kafka.

Closes gh-2095


shakuzen:
I'm not clear on the case when this would be desirable. I'm going to close this for now in light of other recent changes to KafkaMetrics, which have hopefully alleviated some of the issues faced. We can revisit this if the use case becomes more clear.

PR #2155, opened 2020-06-17
Newline character not allowed in tag values

matheusmartinvb:
Influxdb's line protocol explicitly states it does not support newline character in tag values (see beginning of https://docs.influxdata.com/influxdb/v1.8/write_protocols/line_protocol_reference) however they are not handled and if present cause the batch to failed and continuously retry.  The proposed change simply replace newlines with whitespaces.


shakuzen:
Thank you for reporting this issue and including a change request. The logic you've proposed to introduce is something we typically do in a `NamingConvention` implementation and we already have one for Influx (`InfluxNamingConvention`). Also, the logic as is replaces newlines with a space, which needs to be escaped [as mentioned later in the documentation](https://docs.influxdata.com/influxdb/v1.8/write_protocols/line_protocol_reference#special-characters) you shared.
We want to fix this in all upcoming patch releases, so we would target the change at the latest supported maintenance branch (currently `1.1.x`) and merge forward from there.
Given the above, I've opted to commit a fix for this issue in 1bacd2d538f0774accfdf38b5105cad2753324d1 and will merge this forward. Thanks.

Issue #2157, opened 2020-06-18
MetricsDSLContext discards the user's ExecuteListeners


ben-manes:
The mutations of the jOOQ configuration replaces, rather than amends, the user supplied ExecuteListeners. The `derive(ExecuteListeners)` and `set(ExecuteListeners)` should build a combined array when setting.

https://github.com/micrometer-metrics/micrometer/blob/06e895f2a8eced07eae69c449ab74f969a7b6a2b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/MetricsDSLContext.java#L83-L87

https://github.com/micrometer-metrics/micrometer/blob/06e895f2a8eced07eae69c449ab74f969a7b6a2b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/MetricsDSLContext.java#L97-L101

izeye:
@ben-manes Thanks for the issue!

I created https://github.com/micrometer-metrics/micrometer/pull/2158 to try to resolve this.

PR #2158, opened 2020-06-18
Preserve user's ExecutorListeners

izeye:
This PR changes to preserve user's `ExecutorListener`s.

Closes gh-2157

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/MetricsDSLContext.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/MetricsDSLContext.java
@@ -35,6 +35,7 @@ import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


ben-manes:
It looks like `time(configuration)` is called on every query which oddly redoes adding the listener. Doesn't that seem excessive since it was added when initially configured? This would create a lot more object churn and add the listener multiple times if nested usages. Should all of the time methods really delegate to this?

ben-manes:
That said, this snapshot passes my unit tests. :)

Confirmed about duplicates, as I can see it added once on `withMetrics` and then on each query, which should skew the stats.

```
[
  ...ExceptionTranslator@1302b347, 
  ...TransactionSupport@4a7bb0a6, 
  ...CountingExecuteListener@4776fcd5, 
  io.micrometer.core.instrument.binder.db.MetricsDSLContext$$Lambda$398/0x0000000801162040@221577de, 
  io.micrometer.core.instrument.binder.db.MetricsDSLContext$$Lambda$398/0x0000000801162040@7d55ab04
]

izeye:
@ben-manes Thanks for catching it! I created https://github.com/micrometer-metrics/micrometer/pull/2159 to try to resolve it.

jkschneider:
#2159 is merged.

PR #2159, opened 2020-06-19
Ensure to use single JooqExecuteListener

izeye:
This PR changes to ensure to use single `JooqExecuteListener`.

See gh-2158


ben-manes:
Thanks for the quick fix! All looks healthy on my side 😄

PR #2168, opened 2020-06-25
Attach missed @Nullable to ElasticConfig

noveler17:
As its Javadoc, `io.micrometer.elastic.ElasticConfig#get(String)` could return `null`.
But, `@Nonnull` was attached to it in 3e1b71d62256f45fcf01f56ad5ab310586277358.


jkschneider:
These little details make all the difference, thanks for taking the time @noveler17.

PR #2173, opened 2020-07-02
Honor configured step duration when building Wavefront senders

jchambers:
This fixes an issue with the Wavefront registry where, no matter what is set for a `step` duration in `WavefrontConfig`, the Wavefront sender will always emit metrics once per second. I'm fairly confident the once-per-second behavior is not intended because it seems to go against the documentation for `WavefrontConfig#step()` and also against Wavefront's own [example code](https://docs.wavefront.com/micrometer.html#example), which shows providing a `step` of `Duration.ofSeconds(10)` with an explanation that:

> The following code fragment creates a Wavefront registry that emits data every 10 seconds…

Cheers!


shakuzen:
Thank you for reporting this and providing a fix. Sorry for the issue - it looks like we missed this when switching from using our own `HttpSender` abstraction to Wavefront's SDK for sending metrics. Unfortunately I don't see a good way to test that this is having the affect we want, but we'll get the fix merged and I'll add a (brittle) test asserting the internal value is what we expect on the default builder. Thanks again!

jchambers:
Yeah—I had the same struggle with testing, but think what you've come up with is reasonable. Cheers, and thanks!

PR #2183, opened 2020-07-09
Add log4j2 async test (#2176)

### LINKED ISSUE #2176: Async log4j2 log events are not counted accurately
Micrometer provides the `Log4j2Metrics` meter binder to add metrics for Log4j2 log events. Unfortunately the implementation is flawed. It is implemented as a filter, but filters can be invoked multiple times and must therefore not have any side effects.

In practice the filter _is_ invoked multiple times by Log4j2 in many circumstances (e.g. in the presence of multiple and/or async loggers). The result is that the counter is increased multiple times for each single log event and the counter values become too high.

dennysfredericci:
Hi all, hello from Belgium! :)

I just create a test for this issue and indeed the filter have been called multiple times for same event.

Fixes #2176

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
@@ -144,4 +144,20 @@ class Log4j2MetricsTest {
         logger1.info("Hello, world!");
         assertThat(registry.get("log4j2.events").tags("level", "info").counter().count()).isEqualTo(1);
     }
+
+    @Issue("#2176")
+    @Test
+    void asyncLogShouldNotBeDuplicated() throws IOException {
+        ConfigurationSource source = new ConfigurationSource(getClass().getResourceAsStream("/binder/logging/log4j2-async-logger.xml"));
+        Configurator.initialize(null, source);
+
+        Logger logger = LogManager.getLogger(Log4j2MetricsTest.class);
+
+        new Log4j2Metrics().bindTo(registry);
+
+        assertThat(registry.get("log4j2.events").tags("level", "info").counter().count()).isEqualTo(0);
+        logger.info("Hello, world!");
+        assertThat(registry.get("log4j2.events").tags("level", "info").counter().count()).isEqualTo(1);
+    }
+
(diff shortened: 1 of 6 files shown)


dennysfredericci:
Hi Guys,

When we have an async logger configured a call to log.info or any other level triggers 2 to the filter, i found a way to use the isEndOfBatch attribute from LogEvent to check if this is the last call and then metrics counter will increment only once instead of multiple times.

I also create a simple spring-boot project to test it locally. You can see it here: https://github.com/dennysfredericci/springboot-micrometer-issue-2183

shakuzen:
> Just noticed now, I did the PR for master branch should I change it to 1.1.X branch?

Yes, could you please retarget this at the `1.1.x` branch so we can include the fix for all upcoming patch releases?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
```diff
@@ -144,4 +144,20 @@ void noDuplicateLoggingCountWhenMultipleNonAdditiveLoggersShareConfig() {
...
+        ConfigurationSource source = new ConfigurationSource(getClass().getResourceAsStream("/binder/logging/log4j2-async-logger.xml"));
+        Configurator.initialize(null, source);
+
+        Logger logger = LogManager.getLogger(Log4j2MetricsTest.class);
+
+        new Log4j2Metrics().bindTo(registry);
+
+        assertThat(registry.get("log4j2.events").tags("level", "info").counter().count()).isEqualTo(0);
+        logger.info("Hello, world!");
+        assertThat(registry.get("log4j2.events").tags("level", "info").counter().count()).isEqualTo(1);
```
shakuzen, line 160: I think we have a race here since the logger is asynchronous, there's no guarantee that the counter will be incremented by this point.

shakuzen: I still wonder about the theoretical race, but after much debugging, it seems like the issue was probably due to test execution order and global state from before the execution of the newly added test. Changing the cleanup method to `@BeforeEach` seems to have fixed it consistently for me locally.

Issue #1513, opened 2019-07-12
InfluxMeterRegistry - More Granular Configuration Options


austinarbor:
I'm mostly coming from a spring-boot actuator use case perspective here, but it'd be nice to get more configuration options for using the `micrometer-registry-influx` package to customize how the metrics are sent to InfluxDB.  For example, if I use the InfluxDB client directly, I can instantiate a `Point` and  specify the exact values I want to use for the  measurement, field, tags, etc.  I dug through the micrometer registry code briefly, and I really can't see a way to specify what I want to use as the `field` when using this meter registry.  From what I can tell, whatever you pass into `Metrics.gauge(...)` or `Metrics.counter(...)` etc becomes the measurement, and then depending on the type of metric you are creating, the field name gets set to one of the fixed values from `Statistic`.  Adding more customization puts the control of the metrics into the developer's hands, rather than having to fallback to the statsd line protocol and having the team responsible for managing the metrics infrastructure setup/maintain naming convention translations for each application that sends data.

shakuzen:
Could you share more about what kind of customizations you're wanting to make and the motivation for that? All registries make opinionated decisions about what to call things like "sum", "count", etc. and we try to use what's most common for the given backend. If there is some reason the current choices are not ideal, we'd like to know and understand why.

That said, you can always make your own `MeterRegistry` implementation, or extend the `InfluxMeterRegistry` so it writes meters how you want, if you have some specific requirements.

austinarbor:
@shakuzen I'm mainly just talking about being able to specify the field names.  For example, like in the [Getting Started docs](https://docs.influxdata.com/influxdb/v1.7/introduction/getting-started/)f I had a `temperature` measurement, I'd want to be able to have an `internal` field and an `external` field.  As far as I know, this is not possible with the current registry.  On the InfluxDB side this would help keep cardinality down, since in the current implementation, sending gauges for `temperature.internal` and `temperature.external` would be two separate measurements.

I guess I'm almost thinking of an application-level InfluxDB line-protococol template that would be able to translate `Metrics.<type>(<name>,...)` into the line protocol

pmioulet:
It would also be nice to be able to add default tags to all measurements, such as an application or environment name.

LukasVyhlidka:
Hi, I just found this issue when investigating why our influx DB has very very huge tag cardinality :) 

The problem here is that tags in micrometer metrics does not semantically equal to tags in influx.

[...]

PR #2184, opened 2020-07-09
Influx Tag mapper (#1513)

LukasVyhlidka:
Pull request for issue #1513


PyvesB (commented):
From what I can tell, this would do the job, though I believe there's a potential problem with escaping of field values. If `t.getValue()` itself contains quotes, those would need to be escaped when using the line protocol. Indeed, according to the [InfluxDB documentation](https://docs.influxdata.com/influxdb/v1.3/write_protocols/line_protocol_reference/#special-characters):

> For string field values use a backslash character \ to escape double quotes "

tommcdo:
Influx also supports [fields of other types](https://docs.influxdata.com/influxdb/v1.8/concepts/key_concepts/#discussion), so I wonder if it some specialized serialization would be useful here.

In my use case, I need to write multiple fields with numerical values.

leonardehrenfried:
Hello all, I was also hit with the "high series cardinality" problem in influxdb and this seems like the correct solution.

If I resolved the merge conflicts, fixed the escaping and resubmitted it, would there be a chance that it will be merged?

Issue #2174, opened 2020-07-02
Illegal reflective access by ExecutorServiceMetrics


danishnawab:
Ran into the following warning at runtime when trying to monitor an `ExecutorService` using the micrometer binder `ExecutorServiceMetrics`:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics (file:.../micrometer-core-1.5.1.jar) to field java.util.concurrent.Executors$DelegatedScheduledExecutorService.e
WARNING: Please consider reporting this to the maintainers of io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

The following test reproduces the issue:
```java
import io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics;
import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
import org.junit.Test;

import java.util.concurrent.Executors;

public class MicrometerTest {
    @Test
    public void illegalReflectiveAccess() {
        ExecutorServiceMetrics.monitor(new SimpleMeterRegistry(), Executors.newSingleThreadScheduledExecutor(), "monitoredExecutor");
    }
}
```
The issue is observed on both of the following runtimes:
```
openjdk version "14.0.1" 2020-04-14
OpenJDK Runtime Environment Zulu14.28+21-CA (build 14.0.1+8)
OpenJDK 64-Bit Server VM Zulu14.28+21-CA (build 14.0.1+8, mixed mode, sharing)
```
```
openjdk version "14" 2020-03-17
OpenJDK Runtime Environment AdoptOpenJDK (build 14+36)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 14+36, mixed mode, sharing)
```

The aforementioned reflective access can be seen at [line 290 of the ExecutorServiceMetrics.java](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java#L290)

izeye:
Some metrics from the binder depend on reflective access to `Executors`'s internal implementation details, but Java 9+ warns illegal reflective accesses.

There are some options, for example:

1. Remove the reflection-based feature.
[...]

PR #2185, opened 2020-07-11
Add option to disable illegal access in ExecutorServiceMetrics

izeye:
This PR adds an option to disable illegal reflective accesses in `ExecutorServiceMetrics`.

Closes gh-2174


shakuzen (approved):
I like the pragmatism of this. Let's go with this, and we'll continue to explore if we can achieve the same without reflective access. If we sort that out, we can deprecate the method then.

PR #2187, opened 2020-07-12
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


sullis:
@jkschneider

jkschneider:
It's a little trickier here I think. In Micrometer, the master branch uses `latest.release` pretty much everywhere so every build pulls latest. On release branches we use locked dependencies. So if dependabot were to discover a vulnerability in a release branch, it would have to update the lock.

shakuzen:
I have mixed feelings about this and would need to look into it more. We sometimes get requests to upgrade dependencies due to published CVEs - usually for one or more of our shaded dependencies. So far I think these have always been false positives for Micrometer because we do not use the component of the dependency that has the vulnerability (e.g. HTTP server in Netty). Other than our shaded dependencies, users of Micrometer can generally use later versions of dependencies than those specified in Micrometer - we are not forcing old versions on users in most cases.

As Jon mentioned, Dependabot would need to update Gradle dependency lock files consistently so as to not break the build. We would probably need to change our current approach to specifying versions to take advantage of automated dependency upgrades. I also worry about the added toil for little benefit - Micrometer integrates/instruments a lot of things which result in a lot of optional dependencies. So to be transparent, this isn't very high priority to investigate for me unless a case can be made for its importance in spite of the points made here.

jkschneider:
I don't think it is possible to integrate dependabot's workflow with the locking/release branch scheme going on in Micrometer. Thanks for the effort still @sullis!

PR #2192, opened 2020-07-15
coordinated omission MemoryPoolThresholdPauseDetector

mgulla:
This feature allows a new PauseDetector to be configured with a polling interval. So if memory threshold is at a peak, the notification handler should hault future operations and poll the memory beans until the threshold is no longer met:

registry.config().pauseDetector(new MemoryPoolThresholdPauseDetector(3000));


shakuzen (commented):
Thanks for the pull request. It's an interesting idea to use the built-in mechanism to detect a memory usage threshold being exceeded. I haven't heard about much usage of this before, so it was neat to learn about it. There are a number of caveats to using it mentioned in [the JavaDoc](https://docs.oracle.com/en/java/javase/21/docs/api/java.management/java/lang/management/MemoryPoolMXBean.html#UsageThreshold), like which memory pools it should be set on depending on which garbage collector is being used and who sets the usage threshold.

More broadly, I did not understand how the behavior described in the MemoryPollingThread JavaDoc and sysout lines would happen; it claims it will block instruments from running but I don't see how. I'm also not sure that's the behavior we want, if I understand the intention correctly. It would have the effect of the high memory usage issue not showing up in the published metrics because metrics were blocked until the memory usage wasn't high, right?

Given the above, I'm closing the pull request for now. If you still think this would be a valuable enhancement, please give us feedback on the above concerns.

PR #2198, opened 2020-07-21
Adding total JVM heap/non-heap memory usage metrics

veera-chenna:
This PR adds total JVM heap/non-heap memory usage metrics to JvmMemoryMetrics meter binder

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
@@ -101,7 +101,7 @@ public class JvmMemoryMetrics implements MeterBinder {
 
         //The used and committed size of the returned memory usage is the sum of those values of all heap/non-heap memory pools
         // whereas the max size of the returned memory usage represents the setting of the non-heap/non-heap memory which may
-        // not be the sum of those of all non-heap memory pools. If the setting is missing for non-heap memory pools these values come as -1.
+        // not be the sum of those of all non-heap memory pools. If the setting is missing for heap/non-heap memory pools these values come as -1.
 
         MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
@@ -125,7 +125,7 @@ public class JvmMemoryMetrics implements MeterBinder {
                 .baseUnit(BaseUnits.BYTES)
                 .register(registry);
 
-        Gauge.builder("jvm.memory.heap_used_percent", memoryMXBean, (mem) -> getHeapUsagePercent(mem))
+        Gauge.builder("jvm.memory.heap.used.percent", memoryMXBean, (mem) -> getHeapUsagePercent(mem))
                 .tags(tags)
                 .description("The percentage of used memory with respect to maximum amount of memory that can be used for memory management of heap")
                 .register(registry);
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
@@ -24,13 +24,11 @@ import io.micrometer.core.instrument.binder.MeterBinder;
 import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
 
-import java.lang.management.BufferPoolMXBean;
-import java.lang.management.ManagementFactory;
-import java.lang.management.MemoryPoolMXBean;
-import java.lang.management.MemoryType;
-import java.lang.management.MemoryUsage;
+import java.lang.management.*;
 
 import static io.micrometer.core.instrument.binder.jvm.JvmMemory.getUsageValue;
+import static io.micrometer.core.instrument.binder.jvm.JvmMemory.getTotalAreaUsageValue;
+import static io.micrometer.core.instrument.binder.jvm.JvmMemory.getHeapUsagePercent;
 import static java.util.Collections.emptyList;
 
 /**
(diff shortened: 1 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
```diff
@@ -100,6 +98,58 @@ public void bindTo(MeterRegistry registry) {
...
+                .baseUnit(BaseUnits.BYTES)
+                .register(registry);
+
+        Gauge.builder("jvm.memory.committed", memoryMXBean, (mem) -> getTotalAreaUsageValue(mem, MemoryUsage::getCommitted, "heap"))
+                .tags(tagsWithId)
+                .description("The amount of memory in bytes that is committed for the Java virtual machine to use")
+                .baseUnit(BaseUnits.BYTES)
+                .register(registry);
+
+        Gauge.builder("jvm.memory.heap_used_percent", memoryMXBean, (mem) -> getHeapUsagePercent(mem))
```
jkschneider, line 128: we should only use `.` and not `_` in meter names and tag keys.

jkschneider, line 128: is there a reason we need to precompute the percent when we are shipping its constituent parts?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
```diff
@@ -100,6 +98,58 @@ public void bindTo(MeterRegistry registry) {
...
+
+        //The used and committed size of the returned memory usage is the sum of those values of all heap/non-heap memory pools
+        // whereas the max size of the returned memory usage represents the setting of the non-heap/non-heap memory which may
+        // not be the sum of those of all non-heap memory pools. If the setting is missing for heap/non-heap memory pools these values come as -1.
+
+        MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
+
+        Iterable<Tag> tagsWithId = Tags.concat(tags,"id", "total", "area", "heap");
+
+        Gauge.builder("jvm.memory.used", memoryMXBean, (mem) -> getTotalAreaUsageValue(mem, MemoryUsage::getUsed, "heap"))
```
jkschneider, line 110: why sum these up? dimensional monitoring systems will allow you to select all individual heap areas and sum over them as part of the query?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
veera-chenna: We are using Statsd meter registry to send metrics to InfluxDB via Telegraf agent and we use Grafana for visualising these metrics. As cross measurement calculations are not directly supported in InfluxDB, this percentage calculation is not possible when we get the "jvm_memory_used" and "jvm_memory_max" in different measurements. 

Also when we use Kapacitor for alerts on such calculations, we need to perform joins for the the sums of "jvm_memory_used" and "jvm_memory_max" for individual pools. When there are even few missing values for any of the pools in either "jvm_memory_used" or "jvm_memory_max", it is resulting in either wrong alerts being raised or already raised alerts being reset in Kapacitor. 

To avoid all these calculations at different levels, it is better to precompute the percent.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
veera-chenna: We are using Statsd meter registry to send metrics to InfluxDB and we use Grafana for visualising these metrics. As you can see in the screenshot attached, we are seeing some missing points in the Grafana panels when we are running queries to sum over the individual heap areas. 

Grafana_JVM_Heap_Committed_Used

Same issue is seen in our alerting mechanism, Kapacitor also. We are seeing wrong alerts due to any missing points of individual heap areas. To avoid these issues, it would be better to take these values directly from MemoryMXBean without any calculations.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
veera-chenna: Changed meter name to use `.` instead of `_` .

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
jkschneider: Can you shed some light on why there are missing points? Telegraf dropping them? What's going on with this?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
jkschneider: I want to try to come up with a solution for this for you. But pre-computing doesn't seem to be the right way, or we'd be adding a ton of this throughout instrumentation everywhere. You're adding pre-computed stats for things you care about now, but there's no logical end to this progression. Can you help me understand why Influx division doesn't work to compute percentiles? https://docs.influxdata.com/influxdb/v1.8/query_language/math_operators/#division

jkschneider: i.e. is it because they aren't from the same `FROM xxx`? Is this a limitation of Influx or of Grafana's UI implementation of Influx queries?

veera-chenna: Yes, it is because they are not from same `FROM xxx`. This is a limitation of Influx, not the Grafana.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
veera-chenna: We are investigating on whether missing points are at the Telegraf side or at statsd push. Allow me sometime to get back on this.

shakuzen:
As Jon mentioned, summing should be possible in the metrics backend. Making a separate metric for it is redundant and wasteful.

Issue #2130, opened 2020-05-28
Classloading issue in JvmGcMetrics


Chewbizz:
On this line:
https://github.com/micrometer-metrics/micrometer/blob/63771c1c166729ef4bcb280272cff2dc386272ea/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java#L207

The ClassLoader used is the one from JvmGcMetrics' class itself.
In some environments, the GarbageCollectionNotificationInfo class can be perfectly accessible, but not from the same ClassLoader as the one used for JvmGcMetrics itself.

For instance, I'm using a J2EE container, and the ClassLoader from which the JvmGcMetrics class is loaded is a dedicated ClassLoader for a WAR file (where I use JvmGcMetrics).
But the GarbageCollectionNotificationInfo class is loaded before the WAR is started, by the Web Application Container, using another ClassLoader.

So the method returns false here, although the GarbageCollectionNotificationInfo class is available.

In the same method, the ManagementFactory.getMemoryPoolMXBeans() is used some lines above.
And these beans are way more guaranteed to be loaded with a ClassLoader that is able to access the GarbageCollectionNotificationInfo class.
So a possible fix would be to use the ClassLoader from one of these beans' class.
Or simply something like the following should work:

Class.forName("com.sun.management.GarbageCollectionNotificationInfo", false,
                    ManagementFactory.getMemoryMXBean().getClass().getClassLoader());

Another way to deal with it is to rely on the ClassLoader returned by Thread.currentThread().getContextClassLoader()
This would lead to the same issue in my environment, but allows the caller to specify another ClassLoader to use (and would allow me to workaround the issue).

PR #2200, opened 2020-07-22
Fix classloading issue when using a Java EE container

vferries:
CLOSES : https://github.com/micrometer-metrics/micrometer/issues/2130

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -195,7 +195,7 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
     private static boolean isManagementExtensionsPresent() {
         try {
             Class.forName("com.sun.management.GarbageCollectionNotificationInfo", false,
-                    JvmGcMetrics.class.getClassLoader());
+                    MemoryPoolMXBean.class.getClassLoader());
             return true;
         } catch (Throwable e) {
             // We are operating in a JVM without access to this level of detail


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -204,7 +204,7 @@ private static boolean isManagementExtensionsPresent() {
 
         try {
             Class.forName("com.sun.management.GarbageCollectionNotificationInfo", false,
-                    JvmGcMetrics.class.getClassLoader());
+                    ManagementFactory.getMemoryMXBean().getClass().getClassLoader());
```
shakuzen, line 207: While this looks like it should work, the `MemoryMXBean` is not related to what this class is doing. Would something like `MemoryPoolMXBean.class.getClassLoader()` or even `ManagementFactory.class.getClassLoader()` work? Are those also loaded by the application container classloader?

shakuzen: I tested this with Wildfly and it looks like `MemoryPoolMXBean.class.getClassLoader()` resolves the classloader issue. Let me know if that's not the case for your application server.

vferries: I just checked it too and it works.

I will update the PR. You want it from 1.1.x branch to 1.1.x branch ?

vferries:
PR rebased to 1.1.x, see here :
https://github.com/micrometer-metrics/micrometer/pull/2205

PR #2202, opened 2020-07-22
Add "host" tag to JettyClientMetrics

### LINKED ISSUE #2054: Add "host" tag to JettyClientMetrics.
Having bult in support for "host" tag found in other metric clients such as OkHttp3 will be very helpful.

P.S- I can work on this if it is approved.

dennysfredericci:
Resolves #2054

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyClientTags.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyClientTags.java
@@ -45,6 +45,8 @@ public final class JettyClientTags {
 
     private static final Tag METHOD_UNKNOWN = Tag.of("method", "UNKNOWN");
 
+    private static final Tag HOST_UNKNOWN = Tag.of("host", "UNKNOWN");
+
     private static final Pattern TRAILING_SLASH_PATTERN = Pattern.compile("/$");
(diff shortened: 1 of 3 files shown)


dennysfredericci:
IMHO is not needed, but let's see what @izeye, @jkschneider or  @shakuzen, thinks.

shakuzen:
> Should we add a MeterFilter for the "host" tag? As it's based on user input, if it is unbound it could cause unwanted heap pressure.

That's a good point to consider. Since this is the client metrics, the client decides what the host is, unless I'm mistaken. It would be a problem if the server-side metrics included this since the server can't know if the client is trusted to not abuse the host header.
This also aligns with the tags on the [`http.client.requests` metrics in Spring Boot](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-http-clients) (though it is named `clientName` there instead of `host`). There could still be a cardinality concern depending on the architecture and whether the Jetty client is calling individual instances by their unique host name or a domain/load balancer. For that, capping the unique host tags by default could make sense, though I see Spring Boot for instance does not do this with their `clientName` tag currently - only the `uri` tax has a max tags filter. We already configure the meter filter for max `uri` tags in JettyClientMetrics. Let's consider the `host` tag capping in a separate issue.

PR #2210, opened 2020-07-24
Replace deprecated Testcontainers usages

izeye:
This PR replaces deprecated Testcontainers usages.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIntegrationTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIntegrationTest.java
@@ -30,6 +30,7 @@ import org.junit.jupiter.api.Test;
 import org.testcontainers.containers.KafkaContainer;
 import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.junit.jupiter.Testcontainers;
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIntegrationTest.java
```diff
@@ -42,7 +43,7 @@
 @Tag("docker")
 class KafkaClientMetricsIntegrationTest {
     @Container
-    private KafkaContainer kafkaContainer = new KafkaContainer();
+    private KafkaContainer kafkaContainer = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:latest"));
```
bsideup, line 46: FYI it is not recommended to use `:latest`, consider using a fixed tag instead

izeye: @bsideup Thanks for the feedback! I created https://github.com/micrometer-metrics/micrometer/pull/2211 to change to use a fixed version.

For the record, I used `latest` tag intentionally to keep the image up-to-date automatically like other depenedencies on `master`, but there's no similar mechanism to dependency lock, so I changed it as suggested.

PR #2211, opened 2020-07-25
Use a fixed version for Kafka Docker image

izeye:
This PR changes to use a fixed version for the Kafka Docker image.

See https://github.com/micrometer-metrics/micrometer/pull/2210#discussion_r460374459


izeye:
@shakuzen 1.5.x is already using a fix version (5.2.1) via the default constructor. Do you mean to upgrade the image to the latest (5.5.1)?

shakuzen:
> 1.5.x is already using a fix version (5.2.1) via the default constructor. Do you mean to upgrade the image to the latest (5.5.1)?

Good question. I wasn't looking closely enough to realize all of that. I suppose the answer depends on what versions of Kafka we intend to support in our 1.5.x line, which we aren't very explicit about currently. Let's try upgrading to the latest version in 1.5.x assuming that doesn't break anything

izeye:
@shakuzen Thanks for the quick feedback! I updated as suggested.

PR #2217, opened 2020-07-30
Add Neo4j Java driver metrics.

meistermeier:
This PR might come more or less out of the sudden for this project but it had a little bit of history that spawned in a Spring Boot pull request (https://github.com/spring-projects/spring-boot/pull/22302#issuecomment-664940792).

The idea is to expose the available Neo4j Java Driver metrics as (core) micrometer meters instead of keeping it only Spring Boot scoped.

If you have any questions, please feel free to ask me or @michael-simons (mentioning him here to pull him in).

Update: I can see that the build is failing due the missing licence header. I can go ahead and add the VMWare header (to be honest, I do not have any opinions on this), but is a signed Pivotal CLA enough or do I have to also sign a VMWare CLA somewhere? Cannot find the contribution information.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -70,10 +70,12 @@ def VERSIONS = [
         'org.mockito:mockito-core:latest.release',
         'org.mockito:mockito-inline:latest.release',
         'org.mongodb:mongodb-driver-sync:latest.release',
+        'org.neo4j.driver:neo4j-java-driver:4.1.+',
         'org.slf4j:slf4j-api:1.7.+',
         'org.springframework:spring-context:latest.release',
         'org.testcontainers:junit-jupiter:latest.release',
         'org.testcontainers:kafka:latest.release',
+        'org.testcontainers:neo4j:latest.release',
         'org.testcontainers:testcontainers:latest.release',
         'ru.lanwen.wiremock:wiremock-junit5:latest.release',
         'software.amazon.awssdk:cloudwatch:latest.release'
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -70,7 +70,7 @@ def VERSIONS = [
         'org.mockito:mockito-core:latest.release',
         'org.mockito:mockito-inline:latest.release',
         'org.mongodb:mongodb-driver-sync:latest.release',
-        'org.neo4j.driver:neo4j-java-driver:4.1.+',
+        'org.neo4j.driver:neo4j-java-driver:latest.release',
         'org.slf4j:slf4j-api:1.7.+',
         'org.springframework:spring-context:latest.release',
         'org.testcontainers:junit-jupiter:latest.release',
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
@@ -68,50 +68,43 @@ public class Neo4jMetrics implements MeterBinder {
 
             // acquisition attempts
             FunctionCounter.builder(PREFIX + ".acquisition", poolMetrics, ConnectionPoolMetrics::acquired)
-                    .tags(Tags.concat(poolTags, "result", "successful"))
+                    .tags(Tags.concat(poolTags, "outcome", "success"))
                     .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections that have been acquired.")
                     .register(meterRegistry);
 
             FunctionCounter
                     .builder(PREFIX + ".acquisition", poolMetrics, ConnectionPoolMetrics::timedOutToAcquire)
-                    .tags(Tags.concat(poolTags, "result", "timedOutToAcquire"))
+                    .tags(Tags.concat(poolTags, "outcome", "timeout"))
                     .baseUnit(BASE_UNIT_CONNECTIONS)
                     .description("The amount of failures to acquire a connection from a pool within maximum connection "
                                     + "acquisition timeout.")
                     .register(meterRegistry);
 
-            // connections successfully created and closed
-            FunctionCounter.builder(PREFIX, poolMetrics, ConnectionPoolMetrics::created)
-                    .tags(Tags.concat(poolTags, "state", "created"))
-                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections have ever been created.")
-                    .register(meterRegistry);
-
-            FunctionCounter.builder(PREFIX, poolMetrics, ConnectionPoolMetrics::closed)
-                    .tags(Tags.concat(poolTags, "state", "closed"))
-                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections have been closed.")
-                    .register(meterRegistry);
-
             // creation attempts
             FunctionCounter.builder(PREFIX + ".creation", poolMetrics, ConnectionPoolMetrics::created)
-                    .tags(Tags.concat(poolTags, "state", "created"))
+                    .tags(Tags.concat(poolTags, "outcome", "success"))
                     .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections have ever been created.")
                     .register(meterRegistry);
 
             FunctionCounter.builder(PREFIX + ".creation", poolMetrics, ConnectionPoolMetrics::failedToCreate)
-                    .tags(Tags.concat(poolTags, "state", "failedToCreate"))
+                    .tags(Tags.concat(poolTags, "outcome", "failure"))
                     .baseUnit(BASE_UNIT_CONNECTIONS)
                     .description("The amount of connections have been failed to create.").register(meterRegistry);
 
             // active pool size
-            Gauge.builder(PREFIX + ".active", poolMetrics, ConnectionPoolMetrics::idle)
+            Gauge.builder(PREFIX + ".current", poolMetrics, ConnectionPoolMetrics::idle)
                     .tags(Tags.concat(poolTags, "state", "idle"))
                     .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections that are currently idle.")
                     .register(meterRegistry);
 
-            Gauge.builder(PREFIX + ".active", poolMetrics, ConnectionPoolMetrics::inUse)
+            Gauge.builder(PREFIX + ".current", poolMetrics, ConnectionPoolMetrics::inUse)
                     .tags(Tags.concat(poolTags, "state", "inUse"))
                     .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections that are currently in-use.")
                     .register(meterRegistry);
+
+            FunctionCounter.builder(PREFIX + ".closed", poolMetrics, ConnectionPoolMetrics::failedToCreate)
+                    .baseUnit(BASE_UNIT_CONNECTIONS)
+                    .description("The amount of connections have been closed.").register(meterRegistry);
         };
     }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright 2020 VMware, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package io.micrometer.core.instrument.binder.neo4j;
 
 import io.micrometer.core.instrument.FunctionCounter;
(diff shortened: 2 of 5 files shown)


dependencies.gradle
```diff
@@ -65,10 +65,12 @@ def VERSIONS = [
         'org.mockito:mockito-core:latest.release',
         'org.mockito:mockito-inline:latest.release',
         'org.mongodb:mongo-java-driver:latest.release',
+        'org.neo4j.driver:neo4j-java-driver:4.1.+',
```
jkschneider, line 68: Can we keep this to `latest.release` and allow dependency locking to do it's thing?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+
+            FunctionCounter.builder(PREFIX + ".created", poolMetrics, ConnectionPoolMetrics::created).tags(poolTags)
+                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections have ever been created.")
+                    .register(meterRegistry);
+
+            FunctionCounter.builder(PREFIX + ".failedToCreate", poolMetrics, ConnectionPoolMetrics::failedToCreate)
+                    .tags(poolTags).baseUnit(BASE_UNIT_CONNECTIONS)
+                    .description("The amount of connections have been failed to create.").register(meterRegistry);
+
+            Gauge.builder(PREFIX + ".idle", poolMetrics, ConnectionPoolMetrics::idle).tags(poolTags)
```
jkschneider, line 70: Let's combine `.idle` and `.inUse` into one metric: maybe just the `PREFIX` or `PREFIX + ".active"`. Together they represent the "live" connections, whether they are in use or not. Move idle and inUse to something like a `status` tag.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+
+            FunctionCounter.builder(PREFIX + ".acquired", poolMetrics, ConnectionPoolMetrics::acquired).tags(poolTags)
+                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections that have been acquired.")
+                    .register(meterRegistry);
+
+            FunctionCounter.builder(PREFIX + ".closed", poolMetrics, ConnectionPoolMetrics::closed).tags(poolTags)
+                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections have been closed.")
+                    .register(meterRegistry);
+
+            FunctionCounter.builder(PREFIX + ".created", poolMetrics, ConnectionPoolMetrics::created).tags(poolTags)
```
jkschneider, line 62: What is the difference between created and acquired? Suggest combining `.created` and `.failedToCreate` into one metric, maybe `PREFIX + ".creations"` with an `outcome` tag of created or failed. This makes it easy to create an error ratio signal on the division of two tags.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+            Gauge.builder(PREFIX + ".idle", poolMetrics, ConnectionPoolMetrics::idle).tags(poolTags)
+                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections that are currently idle.")
+                    .register(meterRegistry);
+
+            Gauge.builder(PREFIX + ".inUse", poolMetrics, ConnectionPoolMetrics::inUse).tags(poolTags)
+                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections that are currently in-use.")
+                    .register(meterRegistry);
+
+            FunctionCounter
+                    .builder(PREFIX + ".timedOutToAcquire", poolMetrics, ConnectionPoolMetrics::timedOutToAcquire)
```
jkschneider, line 79: Similar, combine acquired and timedOutToAcquire if this is indeed different from creation semantics.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
meistermeier: `Created` are created connection (idle or inUse) in the connection pool. (Incrementing from 0 - max. pool size)
`Acquired` are the used connections in general (0-Long.MAX).

meistermeier:
Thanks for the feedback. I updated my PR with the suggested changes. I tried to group the meters in suitable groups but somehow torn between duplicating the `created` meter (https://github.com/micrometer-metrics/micrometer/pull/2217/files#diff-5c6bcf970da03aa18f5978a59df41f44R84) or leaving the `closed` meter alone as a top level meter.
Personally I would prefer the duplication to put the closed connections in a relationship to the created connection.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+            FunctionCounter
+                    .builder(PREFIX + ".acquisition", poolMetrics, ConnectionPoolMetrics::timedOutToAcquire)
+                    .tags(Tags.concat(poolTags, "result", "timedOutToAcquire"))
+                    .baseUnit(BASE_UNIT_CONNECTIONS)
+                    .description("The amount of failures to acquire a connection from a pool within maximum connection "
+                                    + "acquisition timeout.")
+                    .register(meterRegistry);
+
+            // connections successfully created and closed
+            FunctionCounter.builder(PREFIX, poolMetrics, ConnectionPoolMetrics::created)
```
jkschneider, line 84: Does using "PREFIX" only break elastic? I think so... We'd need a unique suffix to this. I wonder as well whether it makes sense to combine created and closed. I can see it both ways...

michael-simons: Hi Jon. I'm chiming in for Gerrit.

Tbh, I don't understand your question. What's the thing about Elastic? 

If we need a third vote, than I'd go for the duplication so that we can see both metrics related to each other.

michael-simons:
Please let me know if there's something else I can do.

michael-simons:
Ping…

michael-simons:
Is there anything we (Neo4j) can do to get this merged? This is really important for several people and we don't want them to run unsupported workarounds.

shakuzen (commented):
How about the following organization of metrics? Tags in brackets `{}`.
```
neo4j.driver.connections
    .current{state=idle,inUse}
    .creation{outcome=success,failure}
    .acquisition{outcome=success,timeout}
    .closed
```

meistermeier:
Thanks for your reply and input.
I agree on the naming problematic. I tried to recall how we ended with this constellation but because this is now slightly older than one year, I could not find the discussions :)
Since you are the maintainer, I am happy to arrange it the way, you say "sounds" best.

meistermeier:
I updated the PR branch but kept the variable naming for the metrics.
This is due to the fact that Spring Boot will create those instances and yes, it could be that there are more than one driver instance.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
```diff
@@ -0,0 +1,111 @@
...
+    private final Iterable<Tag> tags;
+
+    public Neo4jMetrics(String name, Driver driver, Iterable<Tag> tags) {
+        this.driver = driver;
+        this.tags = Tags.concat(tags, "name", name);
+    }
+
+    @Override
+    public void bindTo(MeterRegistry meterRegistry) {
+        Metrics metrics = this.driver.metrics();
```
shakuzen, line 61: Shouldn't we check `driver.isMetricsEnabled()` before calling `driver.metrics()`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private static final String BASE_UNIT_CONNECTIONS = "connections";
+
+    private final Driver driver;
+
+    private final Iterable<Tag> tags;
+
+    public Neo4jMetrics(String name, Driver driver, Iterable<Tag> tags) {
+        this.driver = driver;
+        this.tags = Tags.concat(tags, "name", name);
```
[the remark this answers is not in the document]
shakuzen: As I think about this, `name` is very generic. What would be the use case of having multiple Driver instances? What is contextually significant? Is each instance connecting to different databases or clusters? If so, maybe `database` or `cluster` is a more specific, descriptive tag, then.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
```diff
@@ -0,0 +1,111 @@
...
+                    .tags(Tags.concat(poolTags, "state", "idle"))
+                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections that are currently idle.")
+                    .register(meterRegistry);
+
+            Gauge.builder(PREFIX + ".current", poolMetrics, ConnectionPoolMetrics::inUse)
+                    .tags(Tags.concat(poolTags, "state", "inUse"))
+                    .baseUnit(BASE_UNIT_CONNECTIONS).description("The amount of connections that are currently in-use.")
+                    .register(meterRegistry);
+
+            FunctionCounter.builder(PREFIX + ".closed", poolMetrics, ConnectionPoolMetrics::failedToCreate)
```
shakuzen, line 105:
```suggestion
            FunctionCounter.builder(PREFIX + ".closed", poolMetrics, ConnectionPoolMetrics::closed)
```

meistermeier: For the first one: I really don't know what would be a good fit here.
Multiple instances could be created because of another database connection (within the same cluster)
but also there are other credentials that act as another user.

For the second one: Good catch, thanks.

shakuzen (approved):
On the `name` tag, I guess this isn't different than the DBCP `name` tag in Spring Boot instrumentation now, and we can't know more specific context on the Driver as far as I understand. So we'll go ahead as is.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/neo4j/Neo4jMetrics.java
```diff
@@ -0,0 +1,114 @@
...
+        this.tags = Tags.concat(tags, "name", name);
+    }
+
+    @Override
+    public void bindTo(MeterRegistry meterRegistry) {
+        if (!this.driver.isMetricsEnabled()) {
+            return;
+        }
+        Metrics metrics = this.driver.metrics();
+        metrics.connectionPoolMetrics().forEach(this.getPoolMetricsBinder(meterRegistry));
```
shakuzen, line 65: I was reading the JavaDoc for `connectionPoolMetrics()` and it mentions that connection pools are dynamically added and removed. It seems like we will only get metrics for whatever pools are initially available at binding time. Won't this be an issue when a pool is removed or added?

meistermeier: This _could_ be the case but on the other hand it could also be a collection that does not change at runtime in practice.
I am waiting for our driver team to explain it to me ;) Will return if I get any insights.

shakuzen: Thank you for checking on it. This could be important enough to reconsider including the metrics as is in our upcoming release.

If pools do dynamically change, we might need to periodically check, or better yet, have some callback for when it changes. Or even better yet, move the micrometer instrumentation to the driver where it should be easier to handle pool changes than from outside the driver.

meistermeier: I talked with the team and they explained that the `connectionPoolMetrics` could change if the topology changes in the cluster. Since we transferred this bit in this form from Spring Boot, I would say: let's keep it as it is right now but I can foresee a change for more dynamic pool support incoming. I will investigate more in the upcoming weeks.

shakuzen: > Since we transferred this bit in this form from Spring Boot

I might be missing some context, but I assume this is referencing the previous pull request to Spring Boot, which didn't accept the metrics changes into Spring Boot. So there is no prior released version of this that users are already relying on, if I'm not mistaken. I would rather not release a version we already know will have issues for users when topology changes happen.

We can get better metrics with a Micrometer-based implementation of `ConnectionPoolMetrics` in the neo4j-java-driver. Is that something the driver team would consider with an optional Micrometer dependency? We could look at contributing that over there instead of this here. We could then get proper Timer metrics on the connection pools also since we would have access to individual timing recordings, which we can't do with the current approach here.

meistermeier:
Sorry for the partial resurrection of this PR :wink: 
Since there is no other place (we are not in "your" Slack anymore), I can find to have a conversation on for this case, I wanted to ask, if there were any initiatives to bring the micrometer support to the Neo4j Java driver?
Any news regarding the offered PR for the driver, @shakuzen ?

shakuzen:
Thanks for the reminder. I've started a work-in-progress branch [here](https://github.com/neo4j/neo4j-java-driver/compare/5.0...shakuzen:micrometer-metrics?expand=1). Things compile but I haven't got a chance to test it yet.

Issue #2226, opened 2020-08-06
NaN or question mark "?" as value of mean field in InfluxLine protocol


admoca60:
Dear all

I have a system with InfluxMeterRegistry configured as default MeterRegistry.

Sometimes, I could see some strange values for some fields in TomcatMetrics, but I suspect that this is not strictly related to any MeterBinder but the core for StepMeterRegistry model.

In my example, we can see how any specific Tomcat metric for influx has the following behaviour:

```
tomcat.global.request,tagkey1=value1,tagkey2=value2 sum=955,count=23,mean=? 1596267900005000000
```

Other times, in other custom metrics, I can see the "NaN" randomly as "mean" field. For example:

```
grpc.server,tagkey1=value1,tagkey2=value2 rpc.completed.responseTime.sum=290.577193,rpc.completed.responseTime.count=9,rpc.completed.responseTime.mean=NaN,rpc.completed.responseTime.upper=48.271632 1596352020000000000
```

Note: I cannot reproduce them easily

izeye:
@admoca60 Thanks for the issue!

This seems to be similar to #1932, but for `InfluxMeterRegistry`. I created https://github.com/micrometer-metrics/micrometer/pull/2227 to try to resolve this.

PR #2227, opened 2020-08-06
Handle NaN for FunctionTimer in InfluxMeterRegistry

izeye:
This PR changes to handle NaN for `FunctionTimer` in `InfluxMeterRegistry`.

Closes gh-2226

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
@@ -189,14 +189,20 @@ public class InfluxMeterRegistry extends StepMeterRegistry {
         return Stream.empty();
     }
 
-    private Stream<String> writeFunctionTimer(FunctionTimer timer) {
-        Stream<Field> fields = Stream.of(
-                new Field("sum", timer.totalTime(getBaseTimeUnit())),
-                new Field("count", timer.count()),
-                new Field("mean", timer.mean(getBaseTimeUnit()))
-        );
-
-        return Stream.of(influxLineProtocol(timer.getId(), "histogram", fields));
+    // VisibleForTesting
+    Stream<String> writeFunctionTimer(FunctionTimer timer) {
+        double sum = timer.totalTime(getBaseTimeUnit());
+        if (Double.isFinite(sum)) {
+            Stream.Builder<Field> builder = Stream.builder();
+            builder.add(new Field("sum", sum));
+            builder.add(new Field("count", timer.count()));
+            double mean = timer.mean(getBaseTimeUnit());
+            if (Double.isFinite(mean)) {
+                builder.add(new Field("mean", mean));
+            }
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java
```diff
@@ -189,14 +189,19 @@ protected void publish() {
...
-                new Field("count", timer.count()),
-                new Field("mean", timer.mean(getBaseTimeUnit()))
-        );
-
-        return Stream.of(influxLineProtocol(timer.getId(), "histogram", fields));
+    // VisibleForTesting
+    Stream<String> writeFunctionTimer(FunctionTimer timer) {
+        double sum = timer.totalTime(getBaseTimeUnit());
+        double mean = timer.mean(getBaseTimeUnit());
+        if (Double.isFinite(sum) && Double.isFinite(mean)) {
```
shakuzen, line 196: In the examples given in the issue, the `sum` had a finite value but the `mean` didn't. I'm wonder if it would make sense to exclude only the `mean` field in that case. I'm not sure what implications this might have in InfluxDB for the `mean` field to be there sometimes but not others.

izeye: @shakuzen Thanks for the feedback!

I think that `mean` should have used the same snapshot as the `sum` for consistency but uses its own snapshot due to the current implementation. I just chose to drop them all as a whole, but if retaining part of it is better, I could split the two. I don't know the implications of doing so, either as I'm not an InfluxDB user.

shakuzen: I think keeping the sum and count even if the mean isn't available is better. I did some ad hoc testing locally and I don't think it is a problem for queries - no data will be returned if querying for the `mean` field during that time, but other fields can be queried during that time.

izeye: @shakuzen Does it make sense to do the same to `ElasticMeterRegistry` for consistency?

See https://github.com/micrometer-metrics/micrometer/pull/2228

shakuzen: It may, but I'd like to setup a test with that to ensure it doesn't break anything for Elasticsearch users on querying, so we can probably follow-up on that. I want to get the next set of patch releases out ASAP.

izeye: @shakuzen Thanks for the quick feedback! I updated as suggested.

shakuzen: Thank you for the quick turnaround. Always appreciated!

PR #2246, opened 2020-08-27
Remove io.projectreactor:reactor-core dependencies

izeye:
This PR removes `io.projectreactor:reactor-core` dependencies as it's a transitive dependency from `io.projectreactor.netty:reactor-netty`.


shakuzen:
While reactor-core will be transitively pulled in by reactor-netty, I would rather the explicit dependency since we are directly using reactor-core classes in our code. If we only used reactor-netty classes, I would probably agree with this change. It is also possible we would want to upgrade the reactor-core version to a more recent patch release than the one reactor-netty pulls in. Let me know if you think we should still make this change for some reason, though.

PR #2248, opened 2020-09-02
Add tomcat connection metrics.

puppylpg:
Monitor the tomcat connection metrics to know the situation about server connections with client.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
@@ -156,6 +156,24 @@ public class TomcatMetrics implements MeterBinder, AutoCloseable {
                     .tags(allTags)
                     .baseUnit(BaseUnits.THREADS)
                     .register(registry);
+
+            Gauge.builder("tomcat.connections.current", mBeanServer,
+                    s -> safeDouble(() -> s.getAttribute(name, "connectionCount")))
+                    .tags(allTags)
+                    .baseUnit(BaseUnits.CONNECTIONS)
+                    .register(registry);
+
+            Gauge.builder("tomcat.connections.keepalive.current", mBeanServer,
+                    s -> safeDouble(() -> s.getAttribute(name, "keepAliveCount")))
+                    .tags(allTags)
+                    .baseUnit(BaseUnits.CONNECTIONS)
+                    .register(registry);
+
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
```diff
@@ -153,6 +153,24 @@ private void registerThreadPoolMetrics(MeterRegistry registry) {
...
                     .baseUnit(BaseUnits.THREADS)
                     .register(registry);
+
+            Gauge.builder("tomcat.connections.current", mBeanServer,
+                    s -> safeDouble(() -> s.getAttribute(name, "connectionCount")))
+                    .tags(allTags)
+                    .baseUnit(BaseUnits.CONNECTIONS)
+                    .register(registry);
+
+            Gauge.builder("tomcat.connections.keepalive.count", mBeanServer,
```
shakuzen, line 163: For consistency, I'm thinking `tomcat.connections.keepalive.current` is a better name

shakuzen, line 163: Would it also make sense to have the `maxKeepAliveRequests` so people can monitor how close to the configured maximum this metric is?

shakuzen (approved):
Looks good. Thank you for the contribution.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
puppylpg: As [the doc](https://tomcat.apache.org/tomcat-8.5-doc/config/http.html) states, `maxKeepAliveRequests` is:
> The maximum number of HTTP requests which can be pipelined until the connection is closed by the server. Setting this attribute to 1 will disable HTTP/1.0 keep-alive, as well as HTTP/1.1 keep-alive and pipelining. Setting this to -1 will allow an unlimited amount of pipelined or keep-alive HTTP requests. If not specified, this attribute is set to 100.

So it's different from count of current keep alive connections.

shakuzen: Thank you for the quick response and the reference. In that case, we probably don't need that one.

puppylpg: :smile_cat:

PR #2262, opened 2020-09-16
Support metrics for multiple Tomcat MBeans of same type

### LINKED ISSUE #1989: TomcatMetrics: Servlet metrics are missing with multiple servlets
TomcatMetrics only provides servlet metrics for one servlet (and not for all registered servlets). This happens due to the fix of issue #1792 which causes `TomcatMetrics.registerMetricsEventually` to only consired the last servlet (see pull request #1828).

Tested/verified with version: 1.3.7

Expectation is that the class TomcatMetrics provides metrics for all registered servlets.

As stated in the pull request the current solution is not the best. Therefore and due to the fact that I need JMX based metrics in my project, I might have a better/more reliable solution:

1. Switch back to `forEach`

2. The JMX query pattern must be provided by the caller (metrics class)
This solves the problem mentioned in #1792 with multiple results, where only one result is expected (`"Tomcat:type=ThreadPool,name=*,*"` (current pattern) vs. `"Tomcat:type=ThreadPool,name=*"` (proposed pattern))

3. Switch the `NotificationFilter` logic to use `ObjectName.apply(ObjectName)`

4. Only remove the `NotificationListener` if the JMX query is not a pattern

5. Extract the JMX query functionality to a helper class
So it can be used by others.

Below you find the final helper class used in my project (5), see also the references to the mentioned items as comment (like `// (1)`). You are welcome to use/adopt this helper or I can help integrating it.
```java
public class JmxBasedMetricsRegistrationHelper {

  private final MBeanServer m_mBeanServer;
  private final Set<NotificationListener> m_notificationListeners = ConcurrentHashMap.newKeySet();

  public JmxBasedMetricsRegistrationHelper() {
    this(getMBeanServer());
  }

  public JmxBasedMetricsRegistrationHelper(MBeanServer mBeanServer) {
    m_mBeanServer = mBeanServer;
  }

  /**
   * If the MBean already exists, register metrics immediately. Otherwise register an MBean registration listener with
   * the MBeanServer and register metrics when/if the MBean becomes available.
   */
  public void registerMetrics(String name, BiConsumer<MBeanServer, ObjectName> consumer) { // (2)
    ObjectName namePattern = toObjectName(name);
    registerMetrics(namePattern, consumer);
  }

  /**
   * If the MBean already exists, register metrics immediately. Otherwise register an MBean registration listener with
   * the MBeanServer and register metrics when/if the MBean becomes available.
   */
  public void registerMetrics(ObjectName namePattern, BiConsumer<MBeanServer, ObjectName> consumer) { // (2)
    Set<ObjectName> objectNames = m_mBeanServer.queryNames(namePattern, null);
    if (!objectNames.isEmpty()) {
      // MBean is present, so we can register metrics now.
      objectNames.forEach(objectName -> consumer.accept(m_mBeanServer, objectName)); // (1)
      return;
    }

    // MBean isn't yet registered, so we'll set up a notification to wait for them to be present and register
    // metrics later.
    registerMetricsWithNotificationListener(namePattern, consumer);
  }

  protected void registerMetricsWithNotificationListener(ObjectName namePattern, BiConsumer<MBeanServer, ObjectName> consumer) {
    NotificationListener notificationListener = new NotificationListener() {

      @Override
      public void handleNotification(Notification notification, Object handback) {
        MBeanServerNotification mBeanServerNotification = (MBeanServerNotification) notification;
        ObjectName objectName = mBeanServerNotification.getMBeanName();
        consumer.accept(m_mBeanServer, objectName);
        if (!namePattern.isPattern()) { // (4)
          // only remove the listener if there is only one query result expected,
          // otherwise some metrics might be missing
          try {
            m_mBeanServer.removeNotificationListener(MBeanServerDelegate.DELEGATE_NAME, this);
            m_notificationListeners.remove(this);
          }
          catch (InstanceNotFoundException | ListenerNotFoundException ex) {
            throw new RuntimeException(ex);
          }
        }
      }
    };
    m_notificationListeners.add(notificationListener);

    NotificationFilter notificationFilter = notification -> {
      if (!MBeanServerNotification.REGISTRATION_NOTIFICATION.equals(notification.getType())) {
        return false;
      }

      // we can safely downcast now
      ObjectName objectName = ((MBeanServerNotification) notification).getMBeanName();
      return namePattern.apply(objectName); // (3)
    };

    try {
      m_mBeanServer.addNotificationListener(MBeanServerDelegate.DELEGATE_NAME, notificationListener, notificationFilter, null);
    }
    catch (InstanceNotFoundException e) {
      // should never happen
      throw new RuntimeException("Error registering MBean listener", e);
    }
  }

  protected ObjectName toObjectName(String name) {
    try {
      return new ObjectName(name);
    }
    catch (MalformedObjectNameException e) {
      throw new RuntimeException("Error registering JMX based metrics", e);
    }
  }

  public Iterable<Tag> nameTag(ObjectName name) {
    String nameTagValue = name.getKeyProperty("name");
    if (nameTagValue != null) {
      return Tags.of("name", nameTagValue.replaceAll("\"", ""));
    }
    return Collections.emptyList();
  }

  /**
   * Removes resources (like the MBean registration listeners) from the MBeanServer.
   */
  public void cleanup() {
    for (NotificationListener notificationListener : m_notificationListeners) {
      try {
        m_mBeanServer.removeNotificationListener(MBeanServerDelegate.DELEGATE_NAME, notificationListener);
      }
      catch (InstanceNotFoundException | ListenerNotFoundException ex) {
        throw new RuntimeException(ex);
      }
    }
  }

  public static MBeanServer getMBeanServer() {
    List<MBeanServer> mBeanServers = MBeanServerFactory.findMBeanServer(null);
    if (!mBeanServers.isEmpty()) {
      return mBeanServers.get(0);
    }
    return ManagementFactory.getPlatformMBeanServer();
  }
}
```

Sample usage/implementation for Eclipse Scout UI metrics:
```java
public class ScoutUiMetrics implements MeterBinder, AutoCloseable {

  private final JmxBasedMetricsRegistrationHelper m_helper;
  private final Iterable<Tag> m_tags;

  public ScoutUiMetrics() {
    this(Collections.emptyList());
  }

  public ScoutUiMetrics(Iterable<Tag> tags) {
    m_helper = new JmxBasedMetricsRegistrationHelper();
    m_tags = tags;
  }

  @Override
    registerSessionMetrics(registry);
  }

  /**
   * Registers Scout session metrics based on {@link org.eclipse.scout.rt.ui.html.management.ISessionMonitorMBean}.
   */
  protected void registerSessionMetrics(MeterRegistry registry) {
    ObjectName namePattern = MBeanUtility.toJmxName("org.eclipse.scout.rt.ui.html", PlatformIdentifier.get(), "Sessions");
    m_helper.registerMetrics(namePattern, this::registerSessionMetrics);
  }

  protected void registerSessionMetrics(MBeanServer mBeanServer, ObjectName name) {
    final MeterRegistry registry = BEANS.get(MeterRegistry.class);
    final Tags tags = Tags.concat(m_tags, m_helper.nameTag(name));

    registry.gauge("scout.ui.session.clientsessions", tags, mBeanServer, mBeanAttributeFunction(name, "NumClientSessions"));
    registry.gauge("scout.ui.session.httpsessions", tags, mBeanServer, mBeanAttributeFunction(name, "NumHttpSessions"));
    registry.gauge("scout.ui.session.uisessions", tags, mBeanServer, mBeanAttributeFunction(name, "NumUiSessions"));
  }

  protected ToDoubleFunction<MBeanServer> mBeanAttributeFunction(ObjectName name, String attribute) {
    return mBeanServer -> safeDouble(() -> mBeanServer.getAttribute(name, attribute));
  }

  protected double safeDouble(Callable<Object> callable) {
    try {
      return Double.parseDouble(callable.call().toString());
    }
    catch (Exception e) {
      return Double.NaN;
    }
  }

  @Override
  public void close() {
    m_helper.cleanup();
  }
}
```

shakuzen:
This works with multiple MBeans for any of the metric types now, which would be expected when there are multiple servlets registered, for example. To support the case when TomcatMetrics is bound before Tomcat MBeans are available, the NotificationListener is no longer unregistered when multiple MBeans could be returned in the query since we have no way to know when all have been registered. All NotificationListeners are still unregistered when `close` is called.

Resolves #1989

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
@@ -26,7 +26,6 @@ import org.apache.catalina.Manager;
 import javax.management.*;
 import java.lang.management.ManagementFactory;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Callable;
@@ -135,7 +134,7 @@ public class TomcatMetrics implements MeterBinder, AutoCloseable {
     }
 
     private void registerThreadPoolMetrics(MeterRegistry registry) {
-        registerMetricsEventually("type", "ThreadPool", (name, allTags) -> {
+        registerMetricsEventually(":type=ThreadPool,name=*", (name, allTags) -> {
             Gauge.builder("tomcat.threads.config.max", mBeanServer,
                     s -> safeDouble(() -> s.getAttribute(name, "maxThreads")))
                     .tags(allTags)
@@ -157,7 +156,7 @@ public class TomcatMetrics implements MeterBinder, AutoCloseable {
     }
 
     private void registerCacheMetrics(MeterRegistry registry) {
-        registerMetricsEventually("type", "StringCache", (name, allTags) -> {
+        registerMetricsEventually(":type=StringCache", (name, allTags) -> {
             FunctionCounter.builder("tomcat.cache.access", mBeanServer,
                     s -> safeDouble(() -> s.getAttribute(name, "accessCount")))
                     .tags(allTags)
@@ -167,11 +166,11 @@ public class TomcatMetrics implements MeterBinder, AutoCloseable {
                     s -> safeDouble(() -> s.getAttribute(name, "hitCount")))
                     .tags(allTags)
                     .register(registry);
-        }, false);
+        });
     }
 
     private void registerServletMetrics(MeterRegistry registry) {
-        registerMetricsEventually("j2eeType", "Servlet", (name, allTags) -> {
+        registerMetricsEventually(":j2eeType=Servlet,name=*,*", (name, allTags) -> {
             FunctionCounter.builder("tomcat.servlet.error", mBeanServer,
                     s -> safeDouble(() -> s.getAttribute(name, "errorCount")))
                     .tags(allTags)
@@ -191,7 +190,7 @@ public class TomcatMetrics implements MeterBinder, AutoCloseable {
     }
 
     private void registerGlobalRequestMetrics(MeterRegistry registry) {
-        registerMetricsEventually("type", "GlobalRequestProcessor", (name, allTags) -> {
+        registerMetricsEventually(":type=GlobalRequestProcessor,name=*", (name, allTags) -> {
             FunctionCounter.builder("tomcat.global.sent", mBeanServer,
                 s -> safeDouble(() -> s.getAttribute(name, "bytesSent")))
                 .tags(allTags)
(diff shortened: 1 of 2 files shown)


shakuzen:
@brunokoeferli feel free to review. I have incorporated most of your suggestions.

Issue #2270, opened 2020-09-23
LogbackMetrics ignoreMetrics will ignore forever if exception is thrown in Runnable


tekener:
In case that the Runnables run() methods throws an exception the ThreadLocal `ignoreMetrics`  stays true forever.
 
LogbackMetrics.java
```   java
public static void ignoreMetrics(Runnable r) {
        ignoreMetrics.set(true);
        r.run();
        ignoreMetrics.remove();
}
```

wuguanyu:
I want know, When should we need invoke LogbackMetrics ignoreMetrics

shakuzen:
> When should we need invoke LogbackMetrics ignoreMetrics ?

See [the JavaDoc](https://www.javadoc.io/doc/io.micrometer/micrometer-core/latest/io/micrometer/core/instrument/binder/logging/LogbackMetrics.html). Copied below:

> Used by Counter.increment() implementations that may cause a logback logging event to occur. Attempting to instrument that implementation would cause a StackOverflowError.

Unless you are making a custom `Counter` implementation, I would not expect users to need to use it. Because the Logback metrics are counters, care needs to be taken when incrementing a counter if it may cause logging, since it could lead to infinite recursion of incrementing.

PR #2271, opened 2020-09-23
Fix LogbackMetrics ignoreMetrics will ignore forever

tekener:
In case that the Runnable throws an exception the ThreadLocal variable ignoreMetrics will stay set.

Fixes https://github.com/micrometer-metrics/micrometer/issues/2270


tekener:
See https://github.com/micrometer-metrics/micrometer/pull/2271

shakuzen:
Thank you for finding the issue and sending a fix. Would you be willing to add a test for this case as well?

tekener:
Hi @shakuzen added the test.

shakuzen (approved):
Thank you for the fix. I've rebased onto `1.1.x` so we can include this in the next round of patch releases.

PR #2272, opened 2020-09-24
Add Mongo collection tag to MongoMetricsCommandListener

philleonard:
## What is this?
A suggested method to extract the `collection` from the `CommandStartedEvent` and propagating to `CommandListener#commandSucceeded` and `CommandListener#commandFailed`. Unfortunately, the collection name is not attainable from the `CommandSucceededEvent` and `CommandFailedEvent` and must be fished out by linking the `request_id` from the `CommandStartedEvent`. Hence the introduction of the Caffeine cache.

## Backstory
We introduced a `MongoMetricsCommandListener` for our organisation's Java services to track collection based usage statistics, as command alone did not provide us with the desired granularity of insights. We have used this across our organisation's services, and have noticed no issues, performance or otherwise. The cache size and expire after write duration were picked to suit our needs, this is unlikely suitable for instances which handle more than 1000 concurrent Mongo queries.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
@@ -15,15 +15,25 @@
  */
 package io.micrometer.core.instrument.binder.mongodb;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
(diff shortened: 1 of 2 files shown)


shakuzen:
> Hence the introduction of the Caffeine cache.

Forcing a Caffeine dependency on users of the Mongo metrics is not something we would do, so we would have to think of another way to achieve things.

jonatan-ivanov:
We did some improvements around this and we now have a `collection` tag, should we close this PR?

shakuzen:
Superseded by #2337

PR #2282, opened 2020-10-08
DRAFT: First pass micrometer registry for opentelemetry

ebullient:


implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryTimer.java
```diff
@@ -0,0 +1,106 @@
...
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.distribution.Histogram;
+import io.micrometer.core.instrument.distribution.HistogramSnapshot;
+import io.micrometer.core.instrument.distribution.NoopHistogram;
+import io.opentelemetry.common.Labels;
+import io.opentelemetry.metrics.DoubleValueRecorder;
+
+public class OpenTelemetryTimer extends AbstractMeter implements Timer {
+    final DoubleValueRecorder recorder;
```
jkschneider, line 32: Should there be one of these for "sum", "count", "max", etc rather than just one for what appears to be effectively just "sum"?

ebullient: This is a question I asked yesterday. The Java API doesn't carry over the spec descriptions in the Javadoc fully. According to the spec: 

> The default aggregation for ValueRecorder computes the minimum and maximum values, the sum of event values, and the count of events, allowing the rate, the mean, and range of input values to be monitored.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#valuerecorder

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryCounter.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public void increment(double amount) {
+        counter.add(amount);
+    }
+
+    @Override
+    public double count() {
+        return 0;
```
kenfinnigan, line 42: Should this return `counter`?

ebullient: No. it is supposed to return the double count value.. BUT, in opentelemetry, the interface is read only (so the value isn't kept where it can be returned). There are some options here that I intend to play with today.

kenfinnigan: Gotcha, thanks

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryCounter.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package io.micrometer.opentelemetry;
+
+import io.micrometer.core.instrument.AbstractMeter;
+import io.micrometer.core.instrument.Counter;
+import io.opentelemetry.metrics.DoubleCounter;
+
+/**
+ * Use an Asynchronous Callback to observe the value of the Micrometer counter.
```
jkwatson, line 23: this comment doesn't seem to be still true, unless i'm missing something.

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionCounter.java
```diff
@@ -0,0 +1,37 @@
...
+    final DoubleSumObserver observer;
+
+    public OpenTelemetryFunctionCounter(Id id, T obj, ToDoubleFunction<T> value, DoubleSumObserver observer) {
+        super(id, obj, value);
+        this.observer = observer;
+        this.observer.setCallback(this::update);
+    }
+
+    public void update(AsynchronousInstrument.DoubleResult result) {
+        result.observe(count(), Labels.empty());
```
jkwatson, line 35: does the value of `count()` get reset ever, or is it monotonically increasing?

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionTimer.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.function.ToLongFunction;
+
+import io.micrometer.core.instrument.cumulative.CumulativeFunctionTimer;
+import io.opentelemetry.common.Labels;
+import io.opentelemetry.metrics.AsynchronousInstrument;
+import io.opentelemetry.metrics.DoubleSumObserver;
+import io.opentelemetry.metrics.DoubleValueObserver;
+
+public class OpenTelemetryFunctionTimer<T> extends CumulativeFunctionTimer<T> {
+    final DoubleSumObserver countObserver;
```
jkwatson, line 29: private?

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionTimer.java
```diff
@@ -0,0 +1,53 @@
...
+    final DoubleValueObserver valueObserver;
+
+    public OpenTelemetryFunctionTimer(Id id, T obj,
+                                      ToLongFunction<T> countFunction,
+                                      ToDoubleFunction<T> totalTimeFunction,
+                                      TimeUnit totalTimeFunctionUnit,
+                                      TimeUnit baseTimeUnit,
+                                      DoubleSumObserver countObserver,
+                                      DoubleValueObserver valueObserver) {
```
jkwatson, line 38: Just thinking-out-loud here, but could you inject a `Meter`/`MeterProvider` here and let the constructor create the two instruments, so their names, etc, would be kept in synch automatically?

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionTimer.java
```diff
@@ -0,0 +1,53 @@
...
+        super(id, obj, countFunction, totalTimeFunction, totalTimeFunctionUnit, baseTimeUnit);
+        this.countObserver = countObserver;
```
jkwatson, line 40: do we need to keep a reference to these observers? I think the SDK will keep a handle to them, either way.

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryGauge.java
```diff
@@ -0,0 +1,44 @@
...
+    final DoubleValueObserver observer;
+
+    public OpenTelemetryGauge(Id id, T obj, ToDoubleFunction<T> value, DoubleValueObserver observer) {
+        super(id, obj, value);
+        this.observer = observer;
+        this.observer.setCallback(this::update);
+    }
+
+    public void update(AsynchronousInstrument.DoubleResult result) {
+        result.observe(value(), Labels.empty());
```
jkwatson, line 42: I assume the plan would be to map tags onto these labels eventually?

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryMeter.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.opentelemetry;
+
+import io.micrometer.core.instrument.Measurement;
+import io.micrometer.core.instrument.internal.DefaultMeter;
+
+public class OpenTelemetryMeter extends DefaultMeter {
```
jkwatson, line 21: What is the DefaultMeter for, and is it necessary to do anything with this?

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+    protected Timer newTimer(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, PauseDetector pauseDetector) {
+        DoubleValueRecorder.Builder builder = otelMeter.doubleValueRecorderBuilder(id.getName())
+                .setUnit("s");
+        copyAttributes(id, builder);
+        return new OpenTelemetryTimer(id, clock, builder.build(), TimeUnit.SECONDS);
+    }
+
+    @Override
+    protected DistributionSummary newDistributionSummary(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, double scale) {
+        return null;
```
jkwatson, line 93: I think this would end up being mapped to a ValueRecorder. At the moment, we'd need to ignore the config, but hopefully when we get a Views API, that could be used to configure a View that would match.

Just FYI, there's has been chatter (maybe just in my head?) about the possibility of views  that are backed by micrometer implementations. In that case, you'd use the OTel APIs, but provide micrometer-implemented views to provide functionality that might be missing from OTel.

Speaking of which, did we ever end up doing the experiment that @kenfinnigan was working on, where the OTel Meter API would be implemented with a micrometer-backed implementation?

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private void copyAttributes(Meter.Id id, AsynchronousInstrument.Builder builder) {
+        builder.setConstantLabels(labelsFrom(id));
+        if ( id.getDescription() != null ) {
+            builder.setDescription(id.getDescription());
+        }
+    }
+
+    private void copyAttributes(Meter.Id id, SynchronousInstrument.Builder builder) {
+        builder.setConstantLabels(labelsFrom(id));
```
jkwatson, line 144: just FYI, constant labels have been removed from the API, and will be gone with the 0.10.0 release. They may come back in the future, but we removed them for now.

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryTimer.java
jkwatson: There is still ongoing discussion about what the default aggregation should be for ValueRecorders (it's MMSC today). And, there is allegedly a views API coming which would allow any sort of customization of the aggregation for a given instrument type. I did a prototype of a proto-views API in this (very out of date, stale) PR:  https://github.com/open-telemetry/opentelemetry-java/pull/1412

ebullient: Point stands.. I only need the one instrument in this case. ;)

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryCounter.java
ebullient: correct. I'll do a pass-through of comments when things settle. (won't resolve this comment as reminder)

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionTimer.java
ebullient: Sure.

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionTimer.java
ebullient: Yes! I didn't think of that. I'll try it

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionTimer.java
ebullient: hmm. maybe we don't. I'll fiddle.

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryGauge.java
ebullient: The labels are all added earlier, right? This is just the last-minute ones..

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryMeter.java
ebullient: Open Question... which is why this one is sitting here. ;)

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryRegistry.java
ebullient: He has it started. We're doing a little in parallel.

IMO: The two APIs will both continue to exist and be used.. so it's all about what happens behind them to make everything smooth.

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryRegistry.java
ebullient: ok. so that means I have to hold/defer applying the labels here.. and do them where you noticed the emptyLabels thing was happening instead. thanks

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionCounter.java
ebullient: count as defined in the CumulativeFunctionCounter: 
```java
    @Override
    public double count() {
        T obj2 = ref.get();
        return obj2 != null ? (last = f.applyAsDouble(obj2)) : last;
    }
```
The FunctionCounter approach is not doing any aggregation itself: it is observing the value observed by applying the specified function on the specified object. To tie back to the docs: 

> Micrometer also provides a more infrequently used counter pattern that tracks a monotonically increasing function (a function that stays the same or increases over time, but never decreases). Some monitoring systems, like Prometheus, push cumulative values for counters to the backend, but others publish the rate at which a counter is incrementing over the push interval. By employing this pattern, you allow the Micrometer implementation for your monitoring system to choose whether to rate normalize the counter or not and your counter remains portable across different types of monitoring systems. 

So, it's up to this registry to decide how to work w/ the OTel SDK to do the right thing

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryGauge.java
ebullient: Ah-ha! I didn't realize your comment below re: removing the mechanism for adding tags earlier.

So yes, you're right, will fix.

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryFunctionCounter.java
jkwatson: ok, then maybe an UpDownSumObserver might be better, if we can't guarantee monotonicity of the `count()` value.

implementations/micrometer-registry-opentelemetry/src/main/java/io/micrometer/opentelemetry/OpenTelemetryRegistry.java
kenfinnigan: @jkwatson it's in progress, getting back into it after a break. Putting together an application to exercise all the aspects to validate it all works

jkwatson: It would be cool if we can use the same app with this registry, to compare and contrast the outputs.

kenfinnigan: I think that would work, as I was planning on having a mix of usage with Micrometer and OTeL APIs to ensure the two work seamlessly

anuraaga:
@ebullient @jkwatson I was thinking it could be nice to look into an OTel registry soon - any thoughts on reviving this PR or should we start from scratch? The approach of implementing a step registry with OTel API types seems like a reasonable one to me.

ebullient:
> @ebullient @jkwatson I was thinking it could be nice to look into an OTel registry soon - any thoughts on reviving this PR or should we start from scratch? The approach of implementing a step registry with OTel API types seems like a reasonable one to me.

I was talking w/ @jonatan-ivanov about this last night. I am happy to have anyone checkout/revisit this PR branch and stomp all over it to try out the revised metrics API to see if there are any gaps..

ebullient:
@kittylyst is also taking a pass at this. ;)

Issue #2223, opened 2020-08-04
support of dynamic tags where the value can change during runtime


jackhammer2k:
I have a feature request for dynamic tag value evaluation. Currently the values of tags must be specified at creation time. However there are cases where they change during runtime. Specifically in my case I want to identify the leader of several instances with a tag. I want to add that tag as a common tag, because it applies to all meters. Because the tag value can change during runtime I cannot specify it as common tag at creation time.

I'm thinking about something like this:

```
public interface Tag extends Comparable<Tag> {
[...]
    static Tag of(String key, Supplier<String> valueSupplier) {
        return new ImmutableTag(key, valueSupplier);
    }
[...]
}
```
```
public class ImmutableTag implements Tag {
[...]
    @Override
    public String getValue() {
        return valueSupplier.get();
    }
[...]
}
```

I would also try to implement that feature, but before I want to be sure it has good chances to be integrated.

SimonScholz:
I'd love to have such a feature as well. Because unfortunately the `WebMvcTagsProvider` only applies for timer instances and not for regular `Counter` instances.

jkschneider:
@jackhammer2k Where this gets really complicated is Micrometer maintains an internal map of which meters have already been registered, so if you repeatedly call `registry.timer("my.timer", "key", "value")` you get the same `Timer` instance back. 

With dynamic tagging, there is this problem that what was previously two meters suddenly becomes one (e.g. a dynamically determined tag key `key` now collides with a statically determined key `key` on another meter).

So the implementation isn't so straightforward. Open to ideas.

SimonScholz:
@jkschneider Well the key itself will be static in @jackhammer2k 's proposal. It's just the value that changes.
We have some common tags we'd like to apply for each meter, but the value of the common tag would change during runtime.
Passing these common tag values all the time is really inconvenient!

shakuzen:
> We have some common tags we'd like to apply for each meter, but the value of the common tag would change during runtime.

I'm curious to understand more of these use cases. What kind of dynamic value common tags are you using?

SimonScholz:
> > We have some common tags we'd like to apply for each meter, but the value of the common tag would change during runtime.
> 
> I'm curious to understand more of these use cases. What kind of dynamic value common tags are you using?

Thanks for having a look, Tommy.

We have some meta data for each and every request as request headers, which we want to add as tags to each and every metric.
[...]

PR #2285, opened 2020-10-09
Introduced a DynamicValueTag class to determine tag values at runtime

SimonScholz:
Hi,

In our application we'd love to have common tags for each and every metric, but these common tags might change during runtime. Passing these dynamic common tag values to each and every counter or timer we create is really inconvenient!

This PR is related to this issue: https://github.com/micrometer-metrics/micrometer/issues/2223

I was also already able to test this new feature, by adding the following code to the `io.micrometer.boot2.samples.components.ServiceLevelObjectiveConfiguration`:

```
    @Bean
    MeterRegistryCustomizer<MeterRegistry> MeterRegistryCommonTagsCustomizer() {
        Tag staticTag = Tag.of("staticTagKey", "static data");
        Tag dynamicTag = Tag.of("dynamicTagKey", () -> "Millis:" + System.currentTimeMillis());
        return registry -> registry.config().commonTags(Arrays.asList(staticTag, dynamicTag));
    }
```
Then when I started the `io.micrometer.boot2.samples.LoggingRegistrySample` application together with the `MeterRegistryCustomizer` mentioned above, I was able to see the different `currentTimeMillis` in the system out.

Having this feature would save us a lot of effort, because otherwise we would need to pass these dynamic tag values to each and every counter or timer we create in our code.


SimonScholz:
@jackhammer2k and @jkschneider do you have any objections about this PR?

micrometer-core/src/main/java/io/micrometer/core/instrument/Tag.java
```diff
@@ -15,20 +15,47 @@
...
+     * @return an instance of {@link Tag}
+     */
     static Tag of(String key, String value) {
         return new ImmutableTag(key, value);
     }
 
+    /**
+     * Create a {@link Tag} instance with dynamic values, which may change during runtime.
+     * <p>
+     * NOTE: If the given valueSupplier returns null, an empty {@link String} will be returned as tag value.
```
jackhammer2k, line 45: I would add the comment about the null-behavior also to DynamicValueTag and Tags.of(...).

micrometer-core/src/main/java/io/micrometer/core/instrument/Tag.java
```diff
@@ -15,20 +15,47 @@
...
+     * @param key of the tag
+     * @param value of the tag
+     * @return an instance of {@link Tag}
+     */
     static Tag of(String key, String value) {
         return new ImmutableTag(key, value);
     }
 
+    /**
+     * Create a {@link Tag} instance with dynamic values, which may change during runtime.
```
jackhammer2k, line 43: I think to speak about "value" is better than "values" here, because although it can take on several values over time, its always just a single value at evaluation time.

jackhammer2k:
I really like the PR and could only find little things in the Javadocs.

micrometer-core/src/main/java/io/micrometer/core/instrument/DynamicValueTag.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public String getValue() {
+        String value = valueSupplier.get();
+        return value != null ? value : "";
+    }
+
+    @Override
+    public boolean equals(@Nullable Object o) {
```
jackhammer2k, line 52: `MeterRegistry` stores all meters in a map, where `Meter.Id` is used as key. But this means that `Meter.Id.hashcode()` or `Meter.Id.equals()` will be called and must not change over time. But because
`Meter.Id.hashcode()` calls `Tags.hashcode()` which calls `DynamicValueTag.hashcode()`, your implementation must only take the key into account to show the required stable bahavior. However I'm not completely sure if its an issue to not take the value into account in equality checks. Maybe thats the issue jkschneider was talking about.

micrometer-core/src/main/java/io/micrometer/core/instrument/Tag.java
SimonScholz: Thanks, I've also added this information to `DynamicValueTag` and `Tags.of(...)`

micrometer-core/src/main/java/io/micrometer/core/instrument/DynamicValueTag.java
SimonScholz: Let's see what @jkschneider or maybe @izeye think about the `equals` check.

SimonScholz:
@jonatan-ivanov Thanks for asking. Meanwhile I left the company, where we've had the need for this feature, but you can find my description of the need here: https://github.com/micrometer-metrics/micrometer/issues/2223#issuecomment-715310456

PR #2287, opened 2020-10-10
Add ch.qos.logback:logback-classic to micrometer-registry-elastic test scope

izeye:
When running `ElasticsearchMeterRegistryElasticsearch7IntegrationTest.indexTemplateShouldApply()`, the following logging failures occur:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
SLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
SLF4J: Defaulting to no-operation MDCAdapter implementation.
SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
```

This PR adds `ch.qos.logback:logback-classic` to `micrometer-test` and removes it from `micrometer-registry-jmx` and `micrometer-registry-statsd` to make projects depending `micrometer-test` benefit from it.

### DIFF
--- a/implementations/micrometer-registry-elastic/build.gradle
+++ b/implementations/micrometer-registry-elastic/build.gradle
@@ -6,4 +6,5 @@ dependencies {
     testImplementation project(':micrometer-test')
     testImplementation 'org.testcontainers:elasticsearch:latest.release'
     testImplementation 'com.jayway.jsonpath:json-path:latest.release'
+    testImplementation 'ch.qos.logback:logback-classic'
 }


shakuzen (commented):
While micrometer-test is mostly used internally, it is something we publish and would expect external registry implementations, for example, to use. Therefore, we don't want to prescribe unnecessary dependencies for everyone, even if only in test code, like logback. I think I'd rather keep the dependency only in our modules that need it for now.

In that sense, I'm wondering if we shouldn't review the other dependencies currently. For example, it looks like testcontainers isn't a necessary dependency for micrometer-test.

izeye:
@shakuzen Thanks for the feedback! I added it to `micrometer-registry-elastic` directly.

> In that sense, I'm wondering if we shouldn't review the other dependencies currently. For example, it looks like testcontainers isn't a necessary dependency for micrometer-test.

I can work on this either here or separately if it's okay.

shakuzen:
> I can work on this either here or separately if it's okay.

Thanks, Johnny. Let's do it separately.

Issue #2255, opened 2020-09-07
Default index pattern not compatible with Elasticsearch 7.9 xpack metrics template


danielbonelli:
Hi Team,

https://github.com/micrometer-metrics/micrometer/blob/9b6d7e16fa15cfb5dd2cd28b711a2a6358d43f2d/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java#L148

indexLine is always { "index" : {} } or with pipeline but according to elasticsearch (https://www.elastic.co/guide/en/elasticsearch/reference/7.9/use-a-data-stream.html#data-streams-bulk-indexing-requests) data stream documentation it only supports "create"

micrometer-registry-elastic-1.5.4

Thanks

shakuzen:
Could we take a step back and describe what you are trying to achieve and the problem you are having before we get into the cause?
The ElasticMeterRegistry does not currently use data streams, which are a very recently added feature as far as I can tell. Is this an enhancement request to support shipping metrics to a data stream instead of as documents?

lnorregaard:
Hi @shakuzen 

I ran into this today that from Elasticsearch 7.9 on Elastic cloud they are using an xpack default feature with a managed index called metrics-*-*. This is a data stream index. 

So what I needed to do was to call the management.metrics.export.elastic.index=metrics_something so it is not hitting the managed template. 

Maybe an idea is to have this as an option if it is a data stream index or a normal index?

danielbonelli:
Hi @shakuzen 

I tried to use out of the box micrometer-registry-elastic-1.5.4 in a spring boot application (2.1.3.RELEASE)
With this yaml configuration
`metrics:
        export:
            elastic:
                enabled: true
                host: http://localhost:9200
                auto-create-index: true
                index: metrics
                step: 10s
        distribution:
            percentiles-histogram:
                http:
                    server:
                        requests: true
            sla:
                http:
                    server:
                        requests: 100ms, 400ms, 500ms, 2000ms
            percentiles:
                http:
                    server:
                        requests: 0.5, 0.75, 0.95, 0.99`

And was getting this error: "only write ops with an op_type of create are allowed in data streams"

What is happening is that calling POST http://localhost:9200/metrics-2020-09/_bulk which gets called from ElasticMeterRegistry publish method is creating a Data Stream and it does not support "index" action only "create"

Hope this clears the issue a bit

izeye:
[...]

PR #2288, opened 2020-10-10
Change default index name prefix to "micrometer-metrics"

izeye:
This PR upgrades Elasticsearch in integration tests to the latest.

### DIFF
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java
@@ -68,12 +68,12 @@ public interface ElasticConfig extends StepRegistryConfig {
 
     /**
      * The index name to write metrics to.
-     * Default is: "metrics"
+     * Default is: "micrometer-metrics"
      *
      * @return index name
      */
     default String index() {
-        return getString(this, "index").orElse("metrics");
+        return getString(this, "index").orElse("micrometer-metrics");
     }
 
     /**
--- a/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/AbstractElasticsearchMeterRegistryIntegrationTest.java
+++ b/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/AbstractElasticsearchMeterRegistryIntegrationTest.java
@@ -113,7 +113,7 @@ abstract class AbstractElasticsearchMeterRegistryIntegrationTest {
     }
 
     private static String getDockerImageName(String version) {
-        return "docker.elastic.co/elasticsearch/elasticsearch-oss:" + version;
+        return "docker.elastic.co/elasticsearch/elasticsearch:" + version;
     }
 
 }
--- a/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticsearchMeterRegistryElasticsearch6IntegrationTest.java
+++ b/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticsearchMeterRegistryElasticsearch6IntegrationTest.java
@@ -25,7 +25,7 @@ class ElasticsearchMeterRegistryElasticsearch6IntegrationTest
 
     @Override
     protected String getVersion() {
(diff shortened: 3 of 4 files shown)


izeye:
OSS version of Elasticsearch doesn't have the build-in index template for the `metrics-*-*` index pattern that triggers #2255, so I changed the integration tests for Elasticsearch to use non-OSS version of Elasticsearch to reproduce it. I'm not sure if it's worth having both versions.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html

izeye:
Based on the doc, it seems that there are two options to work out of the box since Elasticsearch 7.8 as follows:

1. Changing the default index name prefix.
2. Making the index template have a higher priority than the built-in one.

Option 1 will be a breaking change. Option 2 seems to cause a problem to a user using Elastic Agent. Option 1 seems to be better for a long-term perspective. I updated this PR to use a new default index name prefix, `micrometer-metrics`. I'm not sure if there's any better option at the moment.

danielbonelli:
Hey izeye,
I think micrometer-registry-elastic-1.5.5 fixes this problem. 
Although i would like _source on the mapping of the template to be enabled (using a configuration property)
`
"mappings": {
   "_source": {
    "enabled": true/false
  }  
}
`

izeye:
@danielbonelli Thanks for the feedback.

> I think micrometer-registry-elastic-1.5.5 fixes this problem.

Given that Elasticsearch 7.8 contains a breaking change for Micrometer's current implementation, I'm not sure how it could be achieved at the moment.

You can use Elasticsearch 7.8 or above by either:

- Changing the index name prefix.
- Changing the priority of the index template if you don't use Elastic Agent.

> Although i would like _source on the mapping of the template to be enabled (using a configuration property)
> `
> "mappings": {
>    "_source": {
>     "enabled": true/false
>   }  
> }
> `

It doesn't seem relevant to this PR, so it'd be better to discuss separately if you want to pursue it.

implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java
```diff
@@ -68,12 +68,12 @@ default String host() {
...
     /**
      * The index name to write metrics to.
-     * Default is: "metrics"
+     * Default is: "micrometer-metrics"
      *
      * @return index name
      */
     default String index() {
-        return getString(this, "index").orElse("metrics");
+        return getString(this, "index").orElse("micrometer-metrics");
```
shakuzen, line 76: I wonder what % of users are not affected by the issue, which with my current understanding would be those using the OSS version of Elasticsearch. This also seems like it will be a pain for all users upgrading because old data will be in one index pattern and new data will be in another.

I like the simplicity of this solution but I wonder if we shouldn't do something more complicated to avoid a painful upgrade unless necessary. We could use a different index prefix only if we detect a version of Elasticsearch affected by the problem. Either way we will need to write good release notes for this.

izeye: > We could use a different index prefix only if we detect a version of Elasticsearch affected by the problem.

@shakuzen Thanks for the feedback! I thought the same at first, but I inclined to have the same index name prefix in a single Micrometer release for consistency later. I'm not sure which one is better for users though.

implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/AbstractElasticsearchMeterRegistryIntegrationTest.java
```diff
@@ -113,7 +113,7 @@ private String sendHttpGet(String uri) throws Throwable {
     }
 
     private static String getDockerImageName(String version) {
-        return "docker.elastic.co/elasticsearch/elasticsearch-oss:" + version;
+        return "docker.elastic.co/elasticsearch/elasticsearch:" + version;
```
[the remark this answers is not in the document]
shakuzen: Made #2315 to track this

shakuzen (approved):
Unfortunate as the breaking change is, it's probably the best path forward given the circumstances.

PR #2289, opened 2020-10-12
Remove Testcontainers from micrometer-test

izeye:
This PR removes Testcontainers from `micrometer-test`.

WireMock and Mockito also can be removed, but I'm not sure if there's any clear-cut rule for this. If they are good candidates to be removed, I'll update this PR accordingly.

See https://github.com/micrometer-metrics/micrometer/pull/2287#pullrequestreview-506283872


izeye:
> WireMock and Mockito also can be removed,

WireMock is used internally in `micrometer-test`, so it can't be removed, but its scope can be reduced from `api` to `implementation`.

PR #2292, opened 2020-10-13
Use default constructor for atomic objects with default initial value

izeye:
This PR changes to use default constructor for atomic objects with default initial value.

See https://github.com/spring-projects/spring-framework/pull/25846


shakuzen (approved):
Thanks for sharing the performance tip. I was not aware of this. I might have preferred being explicit in some places over saving a few nanoseconds. In tests, for example, it might be less effort to read what the test is doing with the explicit value there. I'm sure most people reading the tests know the default value but it's just a slight extra effort in reading. No need to revert them back, though, unless others feel strongly enough about it.

PR #2298, opened 2020-10-16
Use JMX domain from Manager by default in TomcatMetrics

izeye:
This PR changes to use JMX domain from `Manager` by default in `TomcatMetrics`.

See https://github.com/spring-projects/spring-boot/issues/23717


shakuzen (approved):
Thanks for the quick action on this. I think we should make the change in the 1.1.x line. It fixes the bug previously addressed in 1.0.11 but without any API changes and without requiring users to change their code but it still honors any customization users might be doing with `setJmxDomain`.

izeye:
@shakuzen Thanks for the feedback! I changed the base branch to 1.1.x.

PR #2306, opened 2020-10-26
Support non-generational garbage collector metrics

### LINKED ISSUE #1861: Support Shenandoah GC in JvmGcMetrics
Micrometer: 1.3.5
JDK: 11.0.6

Right now all metrics exported from `JvmGcMetrics` are 0.

### LINKED ISSUE #2305: Support ZGC in JvmGcMetrics
Similar to #1861 for ZGC.

shakuzen:
There were some assumptions in JvmGcMetrics that the garbage collector was generational, with a designated young and old region. ZGC and Shenandoah are examples of OpenJDK garbage collectors that are not generational. GC metrics should be properly recorded with these changes.

Resolves #1861
Resolves #2305

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -47,8 +47,11 @@ import static java.util.Collections.emptyList;
 /**
  * Record metrics that report a number of statistics related to garbage
  * collection emanating from the MXBean and also adds information about GC causes.
+ * <p>
+ * This provides metrics for OpenJDK garbage collectors: serial, parallel, G1, Shenandoah, ZGC.
  *
  * @author Jon Schneider
+ * @author Tommy Ludwig
  * @see GarbageCollectorMXBean
  */
 @NonNullApi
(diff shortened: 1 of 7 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -91,44 +99,41 @@ public void bindTo(MeterRegistry registry) {
...
-            .description("Count of positive increases in the size of the old generation memory pool before GC to after GC")
-            .register(registry);
-
         Counter allocatedBytes = Counter.builder("jvm.gc.memory.allocated").tags(tags)
             .baseUnit(BaseUnits.BYTES)
-            .description("Incremented for an increase in the size of the young generation memory pool after one GC to before the next")
+            .description("Incremented for an increase in the size of the (young) heap memory pool after one GC to before the next")
             .register(registry);
 
+        Counter promotedBytes = (oldGenPoolName == null) ? null : Counter.builder("jvm.gc.memory.promoted").tags(tags)
```
shakuzen, line 124: Promoted only makes sense with generational regions, so we only register the metric when using a collector with an old region.

shakuzen (commented):
This adds an annotation to mark tests related to GC, which can then be run with different garbage collectors configured. Only 2 minimal tests right now, though.

PR #2311, opened 2020-10-28
Spam less with KafkaMetrics and Prometheus

shakuzen:
Logging each occurrence of this at INFO level has been too verbose. Switch to logging the first occurrence at WARN and subsequent occurrences at DEBUG to create less spam for most common logging configurations. We have had reports of this log being spammy in various places (issues, Slack, Stackoverflow). While I do want to improve the situation so we avoid this happening so much, and part of that requires we get reports of when it happens, there is a balance to be had.


shakuzen:
While I think logging each occurrence at INFO is still too noisy, I'm hopeful that #2312 fixes the currently known cases when this was regularly happening.

PR #2321, opened 2020-11-01
Fixes #2320

### LINKED ISSUE #2320: PrometheusNamingConvention duplicates "_seconds" suffix with default configuration for Timer metrics
*Version*: Latest

By default in [PrometheusMeterRegistry](https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L72) configured PrometheusNamingConvention with default constructor which configures "" value in `timerSuffix`.

And when the method `name` called for the Timer metric, it will append the `_seconds` suffix even when it is already present.
see: [this switch case](https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java#L69-L74)

At the same time, that implementation won't add _total suffix for counter metrics.
see:  [this switch case](https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java#L64-L67). 

Am I missing some intentional reason why it is happening in this way or it is a bug?

Fetsivalen:
fix duplication of the _seconds suffix in PrometheusNamingConvention


Fetsivalen:
@shakuzen  is anything wrong with this PR?
Or I just need to wait until You'll have time for it?

shakuzen (approved):
Looks good. Thank you for the pull request. We'll want to rebase on the `1.3.x` branch so that we can merge into there and include this in upcoming patch releases, but I can handle that when merging.

Fetsivalen:
@shakuzen  I can rebase it to 1.3x branch.
But I'm unsure will it help you, or make things more complicated.
so should I rebase?

PR #2322, opened 2020-11-01
Fix javadocs error

ST-DDT:
My IDE displays the following javadoc errors.
Ideally, these would also appear in the build log, but they do not, nevertheless we should fix them anyway.


shakuzen (approved):
Thank you.

PR #2332, opened 2020-11-02
Avoid using GcInfo in method signature

izeye:
This PR changes to avoid using `GcInfo` in the method signature.

Closes gh-2331


shakuzen (approved):
Thanks for the fix. I should have realized this would cause problems when introducing the change.

PR #2335, opened 2020-11-02
Emit second level cache statistics per-region for valid regions

### LINKED ISSUE #2334: IllegalArgumentException in HibernateMetrics with query cache enabled
Related to #2097, it seems this region value isn't always present/enabled (or it isn't resolved).
```
}
2020-11-02 14:01:09,281 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (vert.x-eventloop-thread-9) HTTP Request to /metrics failed, error id: b0b2170e-6d87-4338-9b21-ec10464c8d47-2: java.lang.IllegalArgumentException: Unknown cache region : default-update-timestamps-region
        at org.hibernate.stat.internal.StatisticsImpl.lambda$getDomainDataRegionStatistics$4(StatisticsImpl.java:570)
        at org.hibernate.stat.internal.StatsNamedContainer.getOrCompute(StatsNamedContainer.java:71)
        at org.hibernate.stat.internal.StatisticsImpl.getDomainDataRegionStatistics(StatisticsImpl.java:564)
        at org.hibernate.stat.internal.StatisticsImpl.getDomainDataRegionStatistics(StatisticsImpl.java:36)
        at io.micrometer.core.instrument.binder.jpa.HibernateMetrics.lambda$bindTo$3(HibernateMetrics.java:188)
        at io.micrometer.core.instrument.cumulative.CumulativeFunctionCounter.count(CumulativeFunctionCounter.java:39)
        at io.micrometer.prometheus.PrometheusMeterRegistry.lambda$newFunctionCounter$10(PrometheusMeterRegistry.java:243)
        at io.micrometer.prometheus.MicrometerCollector.collect(MicrometerCollector.java:70)
        at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.findNextElement(CollectorRegistry.java:190)
        at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.nextElement(CollectorRegistry.java:223)
        at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.nextElement(CollectorRegistry.java:144)
        at io.prometheus.client.exporter.common.TextFormat.write004(TextFormat.java:22)
```
https://hibernate.atlassian.net/browse/HHH-12529 
In master, getDomainDataRegionStatistics throws IllegalArgumentException if the region can't be resolved or the region is a QueryResultsRegion.

Fails in Quarkus 1.9.0 (moving to 1.6.0 from 1.5.x).

ebullient:
Resolves #2334 

A second-level cache is optional. 

Pre-5.3, the now-deprecated methods would return null.
In 5.3, the new `getDomainDataRegionStatistics` will throw an IllegalArgumentException if the region can't be resolved.

If we're going to use `getDomainDataRegionStatistics`, the minimum version for hibernate needs to  be set (to require 5.3). 

Given it was deliberately set to 5.+, I've downgraded to the (yes, deprecated) method that works across 5.x versions

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
@@ -179,27 +179,16 @@ public class HibernateMetrics implements MeterBinder {
 
         // Second Level Caching
 
-        long numRegions = Arrays.stream(statistics.getSecondLevelCacheRegionNames())
+        Arrays.stream(statistics.getSecondLevelCacheRegionNames())
                 .filter(this::hasDomainDataRegionStatistics)
-                .peek(regionName -> {
+                .forEach(regionName -> {
                     counter(registry, "hibernate.second.level.cache.requests", "The number of cacheable entities/collections successfully retrieved from the cache",
-                            stats -> stats.getSecondLevelCacheStatistics(regionName).getHitCount(), "region", regionName, "result", "hit");
+                            stats -> stats.getDomainDataRegionStatistics(regionName).getHitCount(), "region", regionName, "result", "hit");
                     counter(registry, "hibernate.second.level.cache.requests", "The number of cacheable entities/collections not found in the cache and loaded from the database",
-                            stats -> stats.getSecondLevelCacheStatistics(regionName).getMissCount(), "region", regionName, "result", "miss");
+                            stats -> stats.getDomainDataRegionStatistics(regionName).getMissCount(), "region", regionName, "result", "miss");
                     counter(registry, "hibernate.second.level.cache.puts", "The number of cacheable entities/collections put in the cache",
-                            stats -> stats.getSecondLevelCacheStatistics(regionName).getPutCount(), "region", regionName);
-                })
-                .count();
-
-        if ( numRegions == 0 ) {
-            // Fallback to previous behavior, with the "all" region tag
-            counter(registry, "hibernate.second.level.cache.requests", "The number of cacheable entities/collections successfully retrieved from the cache",
-                    Statistics::getSecondLevelCacheHitCount, "region", "all", "result", "hit");
-            counter(registry, "hibernate.second.level.cache.requests", "The number of cacheable entities/collections not found in the cache and loaded from the database",
-                    Statistics::getSecondLevelCacheMissCount, "region", "all", "result", "miss");
-            counter(registry, "hibernate.second.level.cache.puts", "The number of cacheable entities/collections put in the cache",
-                    Statistics::getSecondLevelCachePutCount, "region", "all");
-        }
+                            stats -> stats.getDomainDataRegionStatistics(regionName).getPutCount(), "region", regionName);
+                });
 
         // Entity information
         counter(registry, "hibernate.entities.deletes", "The number of entity deletes", Statistics::getEntityDeleteCount);
@@ -261,11 +250,10 @@ public class HibernateMetrics implements MeterBinder {
     }
 
     private boolean hasDomainDataRegionStatistics(String regionName) {
-        // In 5.1/5.2, getSecondLevelCacheStatistics returns null if the region can't be resolved.
         // In 5.3, getDomainDataRegionStatistics (a new method) will throw an IllegalArgumentException
         // if the region can't be resolved.
         try {
-            return statistics.getSecondLevelCacheStatistics(regionName) != null;
+            return statistics.getDomainDataRegionStatistics(regionName) != null;
         } catch (IllegalArgumentException e) {
             return false;
         }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
@@ -178,15 +178,28 @@ public class HibernateMetrics implements MeterBinder {
             Statistics::getCloseStatementCount, "status", "closed");
 
         // Second Level Caching
-        Arrays.stream(statistics.getSecondLevelCacheRegionNames())
-                .forEach(regionName ->  {
+
+        long numRegions = Arrays.stream(statistics.getSecondLevelCacheRegionNames())
+                .filter(this::hasDomainDataRegionStatistics)
+                .peek(regionName -> {
                     counter(registry, "hibernate.second.level.cache.requests", "The number of cacheable entities/collections successfully retrieved from the cache",
-                            stats -> stats.getDomainDataRegionStatistics(regionName).getHitCount(), "region", regionName, "result", "hit");
+                            stats -> stats.getSecondLevelCacheStatistics(regionName).getHitCount(), "region", regionName, "result", "hit");
                     counter(registry, "hibernate.second.level.cache.requests", "The number of cacheable entities/collections not found in the cache and loaded from the database",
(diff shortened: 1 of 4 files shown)


jkschneider:
LGTM

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -177,16 +177,29 @@ public void bindTo(MeterRegistry registry) {
...
+                            stats -> stats.getSecondLevelCacheStatistics(regionName).getMissCount(), "region", regionName, "result", "miss");
                     counter(registry, "hibernate.second.level.cache.puts", "The number of cacheable entities/collections put in the cache",
-                            stats -> stats.getDomainDataRegionStatistics(regionName).getPutCount(), "region", regionName);
-                });
+                            stats -> stats.getSecondLevelCacheStatistics(regionName).getPutCount(), "region", regionName);
+                    return regionName;
+                })
+                .count();
+
+        if ( numRegions <= 0 ) {
```
jonatan-ivanov, line 194: Should not this be `==`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -177,16 +177,29 @@ public void bindTo(MeterRegistry registry) {
         counter(registry, "hibernate.statements", "The number of prepared statements that were released",
             Statistics::getCloseStatementCount, "status", "closed");
 
-        // Second Level Caching
-        Arrays.stream(statistics.getSecondLevelCacheRegionNames())
-                .forEach(regionName ->  {
+        // Second Level Caching:
+        long numRegions = Arrays.stream(statistics.getSecondLevelCacheRegionNames())
+                .filter(regionName -> hasDomainDataRegionStatistics(regionName))
```
jonatan-ivanov, line 182: Can't this be replaced by a method reference?

```suggestion
                .filter(this::hasDomainDataRegionStatistics)
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -177,16 +177,29 @@ public void bindTo(MeterRegistry registry) {
...
+                .map(regionName -> {
```
jonatan-ivanov, line 183: I believe what you are looking for is `peek`. It is like `forEach` but it returns a stream consisting of the elements of the original stream that you can count. Also, since `peek` uses a `Consumer` (instead of a `Function`), you don't need to return anything. Also, map should not have any side-effects.

jonatan-ivanov: I have another suggestion around this: to me a lambda is rather glue-code, I think would move this block into a separate method.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
ebullient: should be. I guess old-school paranoia. ;)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
ebullient: I can try a few things.. this is the result of a few iterations trying to figure out how to handle the fallback case.

ebullient: I agree that peek is better. Should have thought of it. +1 for review. ;)
I would like the two blocks near each other (register these things for a region name if there are data region statistics for it, or register these other things if there were no data regions discovered)... so we could move both to a method...

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
jonatan-ivanov: I know what you are talking about. :D

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
jonatan-ivanov: I think having a fallback is a great idea. I also think that what you did is the simplest thing that you can do on Java 8 to handle the fallback. I wanted to suggest to do a `.findAny()` and some `Optional` magic but then I realized that Java 8 does not have those magic words. :)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -178,15 +178,28 @@ public void bindTo(MeterRegistry registry) {
             Statistics::getCloseStatementCount, "status", "closed");
 
         // Second Level Caching
-        Arrays.stream(statistics.getSecondLevelCacheRegionNames())
-                .forEach(regionName ->  {
+
+        long numRegions = Arrays.stream(statistics.getSecondLevelCacheRegionNames())
+                .filter(this::hasDomainDataRegionStatistics)
```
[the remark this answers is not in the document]
shakuzen: I was able to reproduce the issue so indeed this filter is needed, which then makes me question if the JavaDoc is accurate.

shakuzen (changes requested):
> If we're going to use getDomainDataRegionStatistics, the minimum version for hibernate needs to be set (to require 5.3).

Thank you for pointing this out. I'd rather go in the direction of explicitly making the minimum Hibernate version 5.3 for Micrometer 1.6. I can update the release notes.
In line with that, I think I would rather use the non-deprecated methods. I think the filter you've added here will ensure that meters aren't registered that will throw later.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jpa/HibernateMetrics.java
```diff
@@ -178,15 +178,28 @@ public void bindTo(MeterRegistry registry) {
...
+                            stats -> stats.getSecondLevelCacheStatistics(regionName).getMissCount(), "region", regionName, "result", "miss");
                     counter(registry, "hibernate.second.level.cache.puts", "The number of cacheable entities/collections put in the cache",
-                            stats -> stats.getDomainDataRegionStatistics(regionName).getPutCount(), "region", regionName);
-                });
+                            stats -> stats.getSecondLevelCacheStatistics(regionName).getPutCount(), "region", regionName);
+                })
+                .count();
+
+        if ( numRegions == 0 ) {
+            // Fallback to previous behavior, with the "all" region tag
```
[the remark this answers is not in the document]
ebullient: I was on the fence, so I went to the "let's fall back to whatever less fancy thing we were doing before". ;)
I am happy to ditch it entirely.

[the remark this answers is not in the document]
ebullient: But I'll ask Sanne about it.. because it looks like the path that is incrementing should tip over, too.. so it should all work, or should all be bad..

shakuzen (approved):
Thank you.

shakuzen:
I've rebased onto and merged into 1.6.x. I've also updated the Micrometer 1.6 release notes to mention requiring Hibernate 5.3 or later now.

PR #2333, opened 2020-11-02
Handle null longLivedPoolName in JvmHeapPressureMetrics.bindTo()

izeye:
This PR changes to handle `null` `longLivedPoolName` in `JvmHeapPressureMetrics.bindTo()`.

Closes gh-2330

PR #2336, opened 2020-11-03
Check for null long-lived pool name

### LINKED ISSUE #2330: NullPointerException in JvmMemory.isOldGenPool in 1.6.0 with unknown gc pool name
```
Caused by: java.lang.NullPointerException
        at io.micrometer.core.instrument.binder.jvm.JvmMemory.isOldGenPool(JvmMemory.java:50)
        at io.micrometer.core.instrument.binder.jvm.JvmHeapPressureMetrics.bindTo(JvmHeapPressureMetrics.java:81)
        at io.quarkus.micrometer.runtime.MicrometerRecorder.configureRegistries(MicrometerRecorder.java:100)
        at io.quarkus.deployment.steps.MicrometerProcessor$configureRegistry-2125003310.deploy_0(MicrometerProcessor$configureRegistry-2125003310.zig:144)
        at io.quarkus.deployment.steps.MicrometerProcessor$configureRegistry-2125003310.deploy(MicrometerProcessor$configureRegistry-2125003310.zig:40)
        at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:477)
```

ebullient:
OpenJ9 uses a different naming scheme for memory pools, which means certain existing checks will fail (return a null value).

Let's make sure we don't throw NullPointerExceptions when registering heap metrics.

Resolves #2330 
Supersedes #2333

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmHeapPressureMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmHeapPressureMetrics.java
@@ -73,20 +73,21 @@ public class JvmHeapPressureMetrics implements MeterBinder, AutoCloseable {
 
     @Override
     public void bindTo(@NonNull MeterRegistry registry) {
-        Gauge.Builder<AtomicReference<Double>> builder = Gauge.builder("jvm.memory.usage.after.gc", lastOldGenUsageAfterGc, AtomicReference::get)
-                .tags(tags)
-                .tag("area", "heap")
-                .description("The percentage of old gen heap used after the last GC event, in the range [0..1]")
-                .baseUnit(BaseUnits.PERCENT);
 
         if ( longLivedPoolName != null ) {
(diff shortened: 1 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmHeapPressureMetrics.java
```diff
@@ -78,10 +78,14 @@ public void bindTo(@NonNull MeterRegistry registry) {
                 .tag("area", "heap")
                 .description("The percentage of old gen heap used after the last GC event, in the range [0..1]")
                 .baseUnit(BaseUnits.PERCENT);
-        if (JvmMemory.isOldGenPool(longLivedPoolName))
-            builder.tag("generation", "old");
-        else
-            builder.tag("pool", longLivedPoolName);
+
+        if ( longLivedPoolName != null ) {
```
shakuzen, line 82: If we can't get the long-lived pool name, then we can't calculate the `jvm.memory.usage.after.gc` and might as well not register it. As is, it will always be zero since it is never updated (the update is guarded by a null-check on `longLivedPoolName`).

shakuzen (approved):
Looks good, but I'll move the `jvm.memory.usage.after.gc` meter registration inside the null-check as a polish.

Issue #1799, opened 2020-01-15
Support "collection" tag in "mongodb.driver.commands" metric


netme:
Currently, `mongodb.driver.commands` metric exposes the following tags:
- command
- cluster.id
- server.address
- status

https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java#L62

It would be great to make it more verbose and additionally expose MongoDB collection names.

chrgue:
We were very happy when we found out there is a way to get such metrics.

But for services which are interacting with more then one collection the "collection" tag seams to be mandatory to generate useful insights.

netme:
Thanks, @chrgue for taking a look at this ticket 

I have tried to mock a possible solution:

```java
public class MongoMetricsCommandListener implements CommandListener {

    private final ConcurrentHashMap<Integer, String> collectionsCache;
    
    // ... 

    @Override
    public void commandStarted(final CommandStartedEvent event) {
        extractCommandName(event)
                .ifPresent(it -> collectionsCache.put(event.getRequestId(), it));
    }

    @Override
    public void commandSucceeded(final CommandSucceededEvent event) {
        timeCommand(event, getCollectionNameOrUnknown(event), "SUCCESS", event.getElapsedTime(TimeUnit.NANOSECONDS));
    }

    @Override
    public void commandFailed(final CommandFailedEvent event) {
        timeCommand(event, getCollectionNameOrUnknown(event), "FAILED", event.getElapsedTime(TimeUnit.NANOSECONDS));
    }
    // ...
}
```

MongoDB driver does expose the collection name, however, it is only available in `commandStarted`. We can try to use some cache to get this data accessible in both `commandSucceeded` and `commandFailed`. What do you think about this approach?

PR #2337, opened 2020-11-03
Adds collection name tag to MongoMetricsCommandListener

bono007:
Fixes gh-1799


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Attempts to determine the name of the collection a command is operating on.
+     *
+     * <p>Because some commands either do not have collection info or it is problematic to determine the collection info,
+     * there is an allow list of command names {@link #COMMANDS_WITH_COLLECTION_NAME} used. If {@code commandName} is
+     * not in the allow list or there is no collection info in {@code command}, it will use the content of the
+     * {@code 'collection'} field on {@code command}, if it exists.
+     *
+     * <p>Taken from <a href="https://github.com/openzipkin/brave/blob/master/instrumentation/mongodb/src/main/java/brave/mongodb/TraceMongoCommandListener.java#L115>TraceMongoCommandListener.java</a>
```
bono007, line 47: Its unfortunate that there is no convenient way to share this code between Brave and Micrometer.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -15,59 +15,118 @@
...
+import io.micrometer.core.util.internal.logging.InternalLogger;
+import io.micrometer.core.util.internal.logging.InternalLoggerFactory;
 
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
 /**
- * {@link CommandListener} for collecting command metrics from {@link MongoClient}.
+ * {@link CommandListener} for collecting command metrics from {@code MongoClient}.
```
bono007, line 37: This seemingly innocent link tags requires the `import com.mongodb.MongoClient` which then requires Mongo 3 driver, which then prevents from being used in Spring Boot 2.3+ w/o specifying the legacy driver.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -15,59 +15,118 @@
...
     }
 
     @Override
     public void commandFailed(CommandFailedEvent event) {
-        timeCommand(event, "FAILED", event.getElapsedTime(TimeUnit.NANOSECONDS));
+        timeCommand(event, removeCollectionNameFromCache(event), "FAILED", event.getElapsedTime(TimeUnit.NANOSECONDS));
     }
 
-    private void timeCommand(CommandEvent event, String status, long elapsedTimeInNanoseconds) {
-        timerBuilder
```
bono007, line 63: As this was written before, it was actually not thread-safe and tags from different threads could overwrite tags from other threads as `Timer.Builder.tag()` mutates the tags collection on the builder. Instead, here we reconstruct the instance on each command. I dug in a bit and it did not look expensive to recreate the builder. Let me know if that is not the case.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -15,59 +15,118 @@
...
+    private void addCollectionNameToCache(CommandEvent event, String collectionName) {
+        if (collectionNamesCache.size() < collectionNamesCacheMaxSize) {
+            collectionNamesCache.put(event.getRequestId(), collectionName);
+            return;
+        }
+        if (collectionNamesCacheOverflowErrorTracker == null) {
+            return;
+        }
+        // Cache over capacity - log every ~Nth for a good balance of warning and not spamming the logs.
+        collectionNamesCacheOverflowErrorTracker.maybeInvoke(() -> LOG.warn("Collection names cache is full - Mongo is not calling listeners properly"));
```
bono007, line 123: In practice, we have found that logging this on the very 1st occurrence only, is easy for it to "get lost" in the sea of logs. Every 100th (or Nth) occurrence is a middle-ground between spam and seeing that there is something weird going on in the system. Ideally, this should never happen to begin with.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -15,59 +15,118 @@
...
                 .tag("cluster.id", event.getConnectionDescription().getConnectionId().getServerId().getClusterId().getValue())
                 .tag("server.address", event.getConnectionDescription().getServerAddress().toString())
+                .tag("collection", collection)
                 .tag("status", status)
                 .register(registry)
                 .record(elapsedTimeInNanoseconds, TimeUnit.NANOSECONDS);
     }
 
+    private void addCollectionNameToCache(CommandEvent event, String collectionName) {
+        if (collectionNamesCache.size() < collectionNamesCacheMaxSize) {
```
bono007, line 115: A safety-net in case for some reason Mongo stops calling properly on command end and the cache is not getting cleaned (popped).

chrgue:
Hi there,
does this has a chance to be merged into master?!

Cheers

bono007:
Hi @shakuzen , I know you all are super busy and I did not want to bug either, if you could drop some status on this that would be much appreciated. And if you don't have time to do that right now, I understand that as well. Thanks.

bono007:
@jonatan-ivanov not sure if you have seen this proposal or not. IF you could take a review that would be great. Thank you. 

I am updating conflicts now.

bono007:
@jonatan-ivanov @shakuzen would it be possible to get this in w/ the other round of mongo metrics changes? This, couples w/ those other Mongo metrics MRs that were recently merged will allow us to remove our customizations in our applications. 
Thank you.

bono007:
Just checking in on this one. Could we get some eyes on it? Thanks.

bono007:
@shakuzen - sorry to bug again. Just trying to manage expectations on this one as we are hoping to unroll our local changes once this feature is available in Micrometer.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -56,32 +82,123 @@ public MongoMetricsCommandListener(MeterRegistry registry) {
...
-        Timer.builder("mongodb.driver.commands")
-                .description("Timer of mongodb commands")
-                .tags(tagsProvider.commandTags(event))
-                .register(registry)
+    private void timeCommand(CommandEvent event, String collection, long elapsedTimeInNanoseconds) {
+        Timer.Builder timerBuilder = Timer.builder(this.timerMetricName)
+                .description(this.timerMetricDesc)
+                .tags(this.tagsProvider.commandTags(event));
+        if (this.includeCollectionNameTag) {
+            timerBuilder.tag("collection", collection);
```
jonatan-ivanov, line 129: Can we move most of the new code introduced in this file into the `tagsProvider`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -56,32 +82,123 @@ public MongoMetricsCommandListener(MeterRegistry registry) {
...
+        if (collectionNamesCacheOverflowErrorTracker == null) {
+            return;
+        }
+        // Cache over capacity - log every ~Nth for a good balance of warning and not spamming the logs.
+        collectionNamesCacheOverflowErrorTracker.maybeInvoke(() -> LOG.warn("Collection names cache is full - Mongo is not calling listeners properly"));
+    }
+
+    private String removeCollectionNameFromCache(CommandEvent event) {
+        return Optional.ofNullable(collectionNamesCache.remove(event.getRequestId()))
+                .orElse("unknown");
```
jonatan-ivanov, line 150: Usually not using an `Optional` instead of an if block is a good practice, since, here, the Optional is not returned, what do you think about:
```suggestion
        String  collectionName = collectionNamesCache.remove(event.getRequestId());
        return collectionName != null ? collectionName != null : "unknown";
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -56,32 +82,123 @@ public MongoMetricsCommandListener(MeterRegistry registry) {
...
+    }
+
+    public static Builder builder(MeterRegistry registry) {
+        return new Builder(registry);
+    }
+
+    public static class Builder {
+        private final MeterRegistry registry;
+        private String timerMetricName = "mongodb.driver.commands";
```
jonatan-ivanov, line 159: I would keep these hard-coded this to ensure consistency: if people start renaming these with a lot of services, it is guaranteed that they will end-up having different for the same thing. If they really want to rename them, they can do it using a `MeterFilter`.

jonatan-ivanov (changes requested):

jonatan-ivanov:
@bono007 Thank you for the PR and sorry we did not get back to you earlier.
This looks good and seems very useful to me, we can definitely put this into 1.8.

bono007:
> @bono007 Thank you for the PR and sorry we did not get back to you earlier.
> This looks good and seems very useful to me, we can definitely put this into 1.8.

Your welcome @jonatan-ivanov and no worries on the timeline - I know you are all busy on other amazing things. Thanks for hopping back on this though.  I will make the updates w/in the next 12 hrs.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
bono007: Oh yeh, if there is already a mechanism for this I will gladly replace.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/EveryNthOperation.java
```diff
@@ -0,0 +1,51 @@
...
+ * <p>A common use is to log a message every N times a situation occurs to avoid spamming the application log.
+ *
+ * <p>Uses non-blocking concurrency and is thread safe.
+ */
+class EveryNthOperation {
+
+    private final AtomicInteger callCount = new AtomicInteger(0);
+    private final int nth;
+
+    EveryNthOperation(final int nth) {
```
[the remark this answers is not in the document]
bono007: Lol - yeh I know. My company does so its "Old habits die hard" context switch always. I will remove :)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
```diff
@@ -0,0 +1,72 @@
...
+
+import io.micrometer.core.instrument.util.StringUtils;
+import org.bson.BsonDocument;
+import org.bson.BsonString;
+import org.bson.BsonValue;
+
+class MongoCommandUtil {
+
+    // See https://docs.mongodb.com/manual/reference/command for the command reference
+    private static final Set<String> COMMANDS_WITH_COLLECTION_NAME = new LinkedHashSet<>(Arrays.asList(
```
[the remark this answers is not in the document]
bono007: Its possible I was using it for something in the initial impl and forgot to switch it. I too like to reach for the simple things and then use the special ones when warranted.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param commandName name of the mongo command
+     * @param command mongo command object
+     * @return optional collection name or empty if could not be determined or not in the allow list of command names
+     */
+    Optional<String> determineCollectionName(final String commandName, final BsonDocument command) {
+        if (COMMANDS_WITH_COLLECTION_NAME.contains(commandName)) {
+            final Optional<String> collectionName = getNonEmptyBsonString(command.get(commandName));
+            if (collectionName.isPresent()) {
+                return collectionName;
+            }
```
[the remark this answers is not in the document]
bono007: Agreed. At the very least I typically use `orElse` or `orElseGet` in this scenario. At this point I am more than certain the code was chopped/refactored at some point and I did not do another pass over it. 

Good suggestions.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -56,32 +82,123 @@ public MongoMetricsCommandListener(MeterRegistry registry) {
...
+            throw new IllegalArgumentException("collectionNamesCacheMaxSize must be set to a positive value");
+        }
+        this.collectionNamesCacheMaxSize = builder.collectionNamesCacheMaxSize;
+        this.collectionNamesCacheOverflowErrorTracker = builder.collectionNamesCacheOverflowErrorLogInterval < 1 ? null : new EveryNthOperation(builder.collectionNamesCacheOverflowErrorLogInterval);
     }
 
     @Override
-    public void commandStarted(CommandStartedEvent commandStartedEvent) {
+    public void commandStarted(CommandStartedEvent event) {
+        if (!includeCollectionNameTag) {
```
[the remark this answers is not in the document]
bono007: I will revisit.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
bono007: Ahh yeh. Now I remember. This code (the collectionName tag) was originally included in my contribution for the tags provider MR we did. Because it complicated the code, I wanted to push the simpler pieces through first and then follow on in a subsequent MR (this current one). I will go back and refactor this to fit into the TagsProvider model we added. 👍🏻 


```suggestion
            timerBuilder.tag("collection", collection);
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/EveryNthOperation.java
```diff
@@ -0,0 +1,51 @@
...
+        if (nth < 1) {
+            throw new IllegalArgumentException("nth must be > 0");
```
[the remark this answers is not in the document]
bono007: This one is `> 0`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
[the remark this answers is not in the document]
bono007: I think I was using this previously to maintain order w/o incurring cost of TreeSet. But nonetheless, its not needed now. Good catch.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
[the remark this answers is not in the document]
bono007: Ok, I don't feel so crazy now. I have the code like this for a reason...

We only want to go through the `getNonEmptyBsonString(command.get(commandName))` when the command is in the allow list (other commands do not hold the collection info under the name of the command in the bson map. If that is not present then we want to do the default 'collection' lookup. 


```java
COMMANDS_WITH_COLLECTION_NAME.contains(commandName))

File: micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java [hunk 14]
```diff
@@ -0,0 +1,72 @@
...
+        }
+        // Some other commands, like getMore, have a field like {"collection": collectionName}.
+        return getNonEmptyBsonString(command.get("collection"));
+    }
+
+    /**
+     * @return trimmed string from {@code bsonValue} or empty if value was not a non-empty string
```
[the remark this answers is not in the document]
bono007: I feel the return type implies this but I am ok clarifying in the comment. It is a private method as well.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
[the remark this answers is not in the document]
bono007: I put this in originally as a way for a consumer to basically disable the collection name aspect in case there were issues related to the cache etc.. Somewhat of an emergency lever. I am ok w/ the impl as it is now though - we have been running a clone of this in PROD now for 6 months. I will remove the safeguard. If there is a bug - we will instead fix it quickly.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
bono007: > Usually not using an Optional instead of an if block is a good practice

I am assuming you meant "is not a good practice" or "is a bad practice".  Personally, I like using Optional but I know we don't use it on Spring Boot in this way and its not a surprise that we would not want to use it here. Adjusting.

jonatan-ivanov: I guess I overcomplicated it 😄 : I wrote _not_ using an `Optional` is a good practice in these cases (i.e.: doing something else like `if` or `ternary` op.) but, what I simply meant is using an `Optional` is not a good practice in these cases. 🙂

I like `Optional` a lot though in this case I think it is not necessary (it is more useful for return values). There is a talk about [Optional by Stuart Marks](https://www.youtube.com/watch?v=fBYhtvY19xA) where he explains this in greater details.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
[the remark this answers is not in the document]
jonatan-ivanov: I might be missing something but that is still the case with the suggested change. Let me try to walk trough this.
The original code:
```java
if (COMMANDS_WITH_COLLECTION_NAME.contains(commandName)) {
    Optional<String> collectionName = getNonEmptyBsonString(command.get(commandName));
    if (collectionName.isPresent()) {
        return collectionName;
    }
}

return getNonEmptyBsonString(command.get("collection"));
```
This has three branches:
1. Command is in the allow list and non-empty optional -> `getNonEmptyBsonString(command.get(commandName))`
2. Command is in the allow list and empty optional -> `getNonEmptyBsonString(command.get("collection"))`
3. Command is not in the allow list -> `getNonEmptyBsonString(command.get("collection"))` `//same as #2`

So if we do this:
```java
if (COMMANDS_WITH_COLLECTION_NAME.contains(commandName)) {
    return getNonEmptyBsonString(command.get(commandName))
            .orElseGet(() -> getNonEmptyBsonString(command.get("collection")));
}

return getNonEmptyBsonString(command.get("collection"));
```

We should get the same behavior, right?
(Please notice that I only replaced the inner `if`, the outer allow list check is still there as-is).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
```diff
@@ -0,0 +1,78 @@
...
+     * @param commandName name of the mongo command
+     * @param command mongo command object
+     * @return optional collection name or empty if could not be determined or not in the allow list of command names
+     */
+    Optional<String> determineCollectionName(String commandName, BsonDocument command) {
+        if (COMMANDS_WITH_COLLECTION_NAME.contains(commandName)) {
+            Optional<String> collectionName = getNonEmptyBsonString(command.get(commandName));
+            if (collectionName.isPresent()) {
+                return collectionName;
+            }
```
jonatan-ivanov, line 61: Please check my comment here: https://github.com/micrometer-metrics/micrometer/pull/2337#discussion_r660961391

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019 VMware, Inc.
+ * Copyright 2021 VMware, Inc.
```
jonatan-ivanov, line 2: Since this is not a new file, I guess we can keep it at 2019.

jonatan-ivanov (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
```diff
@@ -0,0 +1,78 @@
...
+    // See https://docs.mongodb.com/manual/reference/command for the command reference
+    private static final Set<String> COMMANDS_WITH_COLLECTION_NAME = new HashSet<>(Arrays.asList(
+            "aggregate", "count", "distinct", "mapReduce", "geoSearch", "delete", "find", "findAndModify",
+            "insert", "update", "collMod", "compact", "convertToCapped", "create", "createIndexes", "drop",
+            "dropIndexes", "killCursors", "listIndexes", "reIndex"));
+
+    /**
+     * Attempts to determine the name of the collection a command is operating on.
+     *
+     * <p>Because some commands either do not have collection info or it is problematic to determine the collection info,
```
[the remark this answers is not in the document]
bono007: At some point either IDEA or checkstyle was complaining about them not being there. I can remove.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProvider.java
```diff
@@ -28,12 +32,40 @@
  */
 public class DefaultMongoCommandTagsProvider implements MongoCommandTagsProvider {
 
+    private static final WarnThenDebugLogger WARN_THEN_DEBUG_LOGGER = new WarnThenDebugLogger(MongoMetricsCommandListener.class);
+
+    // The following is all related to extracting collection name
+    private final MongoCommandUtil mongoCommandUtil = new MongoCommandUtil();
+
+    private final ConcurrentHashMap<Integer, String> collectionNamesCache = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
bono007: Its not a "cache" from a purist standpoint - it holds collection name state across the listener `start -> stop|failed` calls. 

I would prefer something to capture this fact, maybe `inFlightCommandCollectionNames` or `startedCommandCollectionNames`. Alas.... NAMING :)

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
bono007: Yep. I am used to the ranges in the other Spring projects.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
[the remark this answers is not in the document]
bono007: Yep, I understood your suggestion. What I was trying to convey is that the suggested change is more complex as there is now repeated call to `getNonEmptyBsonString(command.get("collection")))` in an `orElseGet()`. This will currenlty get handled when it falls through to the default `return getNonEmptyBsonString(command.get("collection"));`. 

In this case, IMO the `Optional.isPresent` is the cleaner of the two options as it does not violate DRY and it easier to read. 

All of that being said, I don't have a super strong opinion and will go w/ your suggestion if need be. Just lmk.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProvider.java
```diff
@@ -28,12 +32,40 @@
...
                 Tag.of("command", event.getCommandName()),
+                Tag.of("collection", removeCollectionNameFromCache(event)),
                 Tag.of("cluster.id", event.getConnectionDescription().getConnectionId().getServerId().getClusterId().getValue()),
                 Tag.of("server.address", event.getConnectionDescription().getServerAddress().toString()),
                 Tag.of("status", (event instanceof CommandSucceededEvent) ? "SUCCESS" : "FAILED"));
     }
+
+    @Override
+    public void commandStarted(CommandStartedEvent event) {
+        mongoCommandUtil.determineCollectionName(event.getCommandName(), event.getCommand())
```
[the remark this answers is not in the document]
bono007: I did look at collapsing into the DefaultMongoCommandTagsProvider and it clutters it up a bit and also does the same for the test. 

What is the benefit you see by inlining it?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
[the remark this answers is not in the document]
jonatan-ivanov: I just checked the html output, it seems javadoc adds `</p>`.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProvider.java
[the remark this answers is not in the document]
jonatan-ivanov: Basically not having a non-shared utility class and DefaultMongoCommandTagsProvider would encapsulate the logic that way.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProvider.java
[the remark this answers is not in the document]
jonatan-ivanov: 👍🏼

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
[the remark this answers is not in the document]
jonatan-ivanov: To me, calling `ifPresent` (and `get`) on an `Optional` is usually a bad smell. I think having `getNonEmptyBsonString` once in a supplier once in a normal method call is not a big sacrifice but I get your point,  let's keep it as is (we can also put hat call behind a supplier but that would unnecessarily complicate it).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProvider.java
```diff
@@ -68,4 +79,41 @@ private String getAndRemoveCollectionNameForCommand(CommandEvent event) {
...
+     * not in the allow list or there is no collection info in {@code command}, it will use the content of the
+     * {@code 'collection'} field on {@code command}, if it exists.
+     *
+     * <p>Taken from <a href="https://github.com/openzipkin/brave/blob/master/instrumentation/mongodb/src/main/java/brave/mongodb/TraceMongoCommandListener.java#L115>TraceMongoCommandListener.java</a>
+     *
+     * @param commandName name of the mongo command
+     * @param command mongo command object
+     * @return optional collection name or empty if could not be determined or not in the allow list of command names
+     */
+    protected Optional<String> determineCollectionName(String commandName, BsonDocument command) {
```
bono007, line 97: This is not private for 2 reasons.

1. VisibleForTesting - I still want to test this piece of the tags provider solo (and re-use the existing MongoCommandUtil tests).  Typically I would do that w/ package protected modifier. I went w/ `protected` though for the following reason. 

2. If a consumer needs another way to determine the collection name from a command event they could extend the DTP and override this method. I don't want to encourage that but it would give a consumer the ability to adjust the behavior if need be. @jonatan-ivanov what are your stance on "consumer-friendly and leaving some key methods extendable" vs. "lock it all down as any public or protected API is an increase in the dependency blast radius that we have to support". ?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProviderTest.java
```diff
@@ -52,25 +57,25 @@ void defaultCommandTags() {
     }
 
     @Test
-    void handlesCacheFullGracefully() {
+    void handlesCommandsOverLimitGracefully() {
```
bono007, line 60: More removal of the "cache" terminology.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProviderTest.java
```diff
@@ -97,4 +102,57 @@ private CommandSucceededEvent commandSucceededEvent(final int requestId) {
                 new BsonDocument(),
                 1200L);
     }
+
+    @Nested
```
bono007, line 106: The choices were:

1. Just inline these tests and keep the longer name of "determineCollectionNameWith...`
2. Put these tests in a separate test class named something like `DefaultMongoTagsProviderDetermineCollectionNameTest`
3. Inline them but separate them clearly w/ a `@Nested` test class.

Wdyt?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandUtil.java
bono007: Do you mind resolving this one now?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProviderTest.java
jonatan-ivanov: 👍🏼

jonatan-ivanov (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProvider.java
jonatan-ivanov: 👍🏼 
I think it is great! It provides an escape route, also one of the original goal of the tag provider was making it flexible enough so that users can override the original/provide their own, this matches with that nicely.

jonatan-ivanov:
@bono007 Thank you very much for fixing this! 🙂

bono007:
Thanks for the thorough and fun code review and getting this merged @jonatan-ivanov . 

Question: By looking at the [release calendar](https://calendar.spring.io/) it looks like this (in Micrometer 1.8) will make it in Spring Boot 2.6.0. Do you think that is correct?

jonatan-ivanov:
👍🏼 

Yes, it is correct. We are trying to follow the Boot release cadence with our LTS releases (i.e.: 2.6.0 -> 1.8.0).
With Spring Boot 2.6.0, you should get Micrometer 1.8.0; our current plan is to release 1.8.0 M1, M2, M3, RC1, GA a week before Boot releases the same versions for 2.6.0. This means that 1.8.0-M1 should be out on July 14. The rest is not in the calendar for Micrometer but you can calculate it from the Spring Boot releases.

PR #2356, opened 2020-11-21
Add (optional) extra tags to mongo connection pool

### LINKED ISSUE #2355: Support for Connection pool tags provider
This is especially important in applications which have multiple mongo connection pools, for example.

Thanks in advance, and great job/project!

gustavomonarin:
This is extremely useful for example to differentiate the connection pool in an application with multiple different connections to different databases which could be in the same mongo server.

resolves #2355


bono007:
Hi @gustavomonarin , I had a similar pull request queued up (#2360) as it was initially built on top of another pending pull request. I saw this PR and since its in a close area I went ahead and submitted mine for the command listener so we can make sure they end up using similar mechanisms. I went w/ the tags provider approach for the command listener. You went w/ the extra tags approach. I asked [this](https://github.com/micrometer-metrics/micrometer/pull/2360#issuecomment-733747575). 

Thanks

gustavomonarin:
Closing this PR in favor of #2437, as discussed with @bono007  that should make it more flexible and in line with more complex cases which could not fit by a static list of extra tags without contextual  information (ie. data from the event itself).

PR #2360, opened 2020-11-25
Adds pluggable tags provider to MongoMetricsCommandListener

bono007:
Allows future customization of tags used by MongoMetricsCommandListener


bono007:
There is a similar pull request #2356 for the connection pool listener that uses extra tags approach whereas this one for the. command listener goes w/ the tags provider approach 

@shakuzen and/or team, please advise which approach is preferred, tags provider or extra tags. I think either is fine but we should probably land on a single consistent approach for these two listeners. 

Thanks

gustavomonarin:
Hey bono, thanks for the details.

The tags provider approach looks really good, specially in the case of the command listener, where the command event has more contextual data.

Unfortunately, this extra information was not present as of the connection pool event as of the mongo3 driver, so i went to a simpler approach.

However in order to keep the code consistent, i would be happy to change my open PR to use the tags provider approach  as well ;)

bono007:
@gustavomonarin, exactly right on the `more contextual data` - I have an app that has the need to pass tags differently at runtime on the commands based on the "context" in the application so that is why I did the tags provider approach.

I think it would be good to keep them consistent.  If you are ok moving to the tags provider I think that would be great. 

I would keep the PRs separate still though. I understand the desire to get it reviewed/merged but I am sure the team has limited resources and many priorities :).  Since the listeners are configured in our application, I simply created my own listener locally until this gets merged. Then I will follow back and get rid of the custom listener. That removes the sense of urgency for me.

gustavomonarin:
Makes sense about the separate prs and understand about the maintainers priorities.

I also have done the same with the listener in my code and have no urgency.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -61,9 +83,7 @@ public void commandFailed(CommandFailedEvent event) {
 
     private void timeCommand(CommandEvent event, String status, long elapsedTimeInNanoseconds) {
         timerBuilder
-                .tag("command", event.getCommandName())
-                .tag("cluster.id", event.getConnectionDescription().getConnectionId().getServerId().getClusterId().getValue())
-                .tag("server.address", event.getConnectionDescription().getServerAddress().toString())
+                .tags(tagsProvider.commandTags(event))
                 .tag("status", status)
```
deejay1, line 87: the `status` tag is provided by tagsProvider already, so probably we shouldn't need to duplicate it here

bono007: True. I will update w/in the next 24 hrs.

deejay1: You could also change the timeCommand parameter list to drop the `status` one

bono007: Ok, I finally got back around to updating this. Removing the `status` did clean up the code a bit. I originally left it in thinking that status is something that is core to the listener and *should* always be provided. However, that is rather presumptuous of me and the next consumer may not want status, may want to call it something else - this is the whole purpose of the tags provider :) 

I rebased and squashed as it was out of date. I was not going to squash but my keyboard nad muscle memory had a different plan.

deejay1 (approved):

deejay1:
@bono007 can you fix the conflicts or should?

bono007:
I can get it @deejay1 - I did not notice them or I would have done it w/o you having to ask. :)

bono007:
@deejay1 that should do it

deejay1 (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoMetricsCommandTagsProvider.java
```diff
@@ -0,0 +1,38 @@
+/**
+ * Copyright 2019 VMware, Inc.
```
jonatan-ivanov, line 2: Could you please fix the year (2020)?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoMetricsCommandTagsProvider.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class DefaultMongoMetricsCommandTagsProvider implements MongoMetricsCommandTagsProvider {
+
+    @Override
+    public Iterable<Tag> commandTags(CommandEvent event) {
+        return Tags.of(
+                Tag.of("command", event.getCommandName()),
+                Tag.of("cluster.id", event.getConnectionDescription().getConnectionId().getServerId().getClusterId().getValue()),
+                Tag.of("server.address", event.getConnectionDescription().getServerAddress().toString()),
+                Tag.of("status", (event instanceof CommandFailedEvent) ? "FAILED" : "SUCCESS"));
```
jonatan-ivanov, line 36: What do you think about checking `CommandSucceededEvent` and fall back to `"FAILED"` instead?
So if for any reason a third type of event will be passed in the future, it will mark this as `"FAILED"` which will draw better user attention than `"SUCCESS"` so there is a higher chance that the bug will be reported.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
```diff
@@ -48,21 +70,18 @@ public void commandStarted(CommandStartedEvent commandStartedEvent) {
...
     @Override
     public void commandSucceeded(CommandSucceededEvent event) {
-        timeCommand(event, "SUCCESS", event.getElapsedTime(TimeUnit.NANOSECONDS));
+        timeCommand(event, event.getElapsedTime(TimeUnit.NANOSECONDS));
     }
 
     @Override
     public void commandFailed(CommandFailedEvent event) {
-        timeCommand(event, "FAILED", event.getElapsedTime(TimeUnit.NANOSECONDS));
+        timeCommand(event,event.getElapsedTime(TimeUnit.NANOSECONDS));
```
jonatan-ivanov, line 78: Space after comma (`,`).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandTagsProvider.java
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright 2019 VMware, Inc.
```
jonatan-ivanov, line 2: Could you please fix the year?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandTagsProvider.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.mongodb.event.CommandEvent;
+import io.micrometer.core.instrument.Tag;
+
+/**
+ * Provides {@link Tag Tags} for Mongo command metrics.
+ *
+ * @author Chris Bono
+ */
+public interface MongoMetricsCommandTagsProvider {
```
jonatan-ivanov, line 26: What do you think about putting `@FunctionalInterface` on this class?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
```diff
@@ -92,6 +97,38 @@ void shouldCreateFailedCommandMetric() {
         assertThat(registry.get("mongodb.driver.commands").tags(tags).timer().count()).isEqualTo(1);
     }
 
+    @Test
+    void shouldCreateSuccessCommandMetricWithCustomSettings() {
```
jonatan-ivanov, line 101: Could you please add a test for the failed scenario too?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
```diff
@@ -149,4 +186,4 @@ void destroy() {
         }
     }
 
-}
+}
```
jonatan-ivanov, line 189: Could you please add final newline (see `.editorconfig` file)?
(Most IDEs are supporting it OOB or through a plugin.)

jonatan-ivanov (changes requested):

jonatan-ivanov:
# tags provider vs. extra tags

Seemingly, this PR kind of eliminates the need of #2356 since users can extend or wrap the `DefaultMongoMetricsCommandTagsProvider` and do something like this in their own implementation of `commandTags`:
```java
return Tags.of(super/delegate.commandTags(event)).and(Tag.of(...)));
```

On the other hand, I like the simplicity of the extra tags, it is more direct and provides good API discoverability.
Since the tag provider gives you access to `CommandEvent`, I think I would definitely keep that behavior. With this I think there are at least two ways to merge this:
1. Merging this PR and closing #2356
2. Merging this PR and applying #2356 on top of this
This basically means adding `extraTags` to `DefaultMongoMetricsCommandTagsProvider` so users can easily add extra tags that are not dependent on the `CommandEvent` without implementing any logic on their side

@bono007 @gustavomonarin WDYT?
/cc @shakuzen

bono007:
@jonatan-ivanov thanks for all of the feedback. I will get back w/ adjustments w/in 24 hrs.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandTagsProvider.java
bono007: I am ok w/ that and was initially going to but I noticed that it was not used anywhere else in the repo so I left it out for consistency w/ the intention to follow up and ask if it was something not used in MM on purpose or just not used yet. Thanks for the reminder.  

@shakuzen - thoughts?

bono007:
@jonatan-ivanov I incorporated your feedback w/ the one exception of the `@FunctionalInterface` but am happy to add that as well when we hear back from Tommy.

bono007:
> # tags provider vs. extra tags


So the 2 Mongo listeners are named: 

- `MongoMetricsCommand`Listener
- `MongoMetricsConnectionPool`Listener

and the naming in this code proposal is targeted only to the `MongoMetricsCommand` variant (`MongoMetricsCommandTagsProvider`) as their base set of tags differs. 

IIRC, in the linked discussion me and @gustavomonarin landed on both using the tags provider approach, one variant for each listener. However, I agree w/ @jonatan-ivanov that having the "extra tags" provides a nice API and also has the added benefit of allowing the extra tags to vary across MongoClient instances. In the "ask" on #2356 the extra tag is `mongoTemplate=foo` which would vary across MongoClients in the case where the consuming app is using multiple MongoClients. 

My vote would be to have both "tags provider" + "optional extra tags" in **both** listeners. Default tags and extra tags are complimentary and not necessarily the same thing.

If we did the above, a possible order of execution could be: 

1. Update this MR to support optional extra tags (me)
2. Update #2356 to support a `MongoMetricsCommandPoolListener` (@gustavomonarin )

Thoughts?

gustavomonarin:
great to see all the work and interaction on this subject @bono007 , @deejay1  and @jonatan-ivanov .

I have closed the  #2356  and opened a new #2437 which is inline with the implementation solution in this PR. 
This approach  is more flexible and in line with more complex cases which could not fit by a static list of extra tags without contextual information (ie. data from the event itself).

Happy to receive any feedback.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandTagsProvider.java
bono007: I went ahead and added it as it should not be an issue - if it is I will remove it.

jonatan-ivanov:
@bono007 I think it is a good idea, please ping me once you added the extra tags and I will take a look.
@gustavomonarin Great, once this is settled and merged, I'm going to jump onto your PR.

bono007:
Sounds good @jonatan-ivanov . I should be able to look at this w/in the next 12hrs.

bono007:
@jonatan-ivanov - I was unable to get to this today - the day job got in my way hehe. I will get to it w/in the next 12 hrs (but for real this time).

bono007:
So after going through the motions and adding the extra tags concept I am now of the opinion that I think we can do without it. 

My initial thinking was to create a tags customizer concept in Spring Boot that would allow consumers to only define that (if they wanted to add tags). This would prevent them from re-defining the entire `DefaultMongoMetricsCommandTagsProvider` bean. However, I don't think that ends up buying us too much. 

**Not adding tag customizer**
(+) Simpler, less moving pieces
(-) Consumers have to redefine the entire tags provider bean and add the tags to it via clunky `super/delegate.tags` pattern

**Adding tag customizer**
(-) Adding another new concept of tags customizer  into Spring Boot may be confusing w/ the concept of tags provider
(-) Also customizers typically operate on builders which we do not have here. 

I think we should go w/ what we have for now. If consumers of Boot want to add tags they would re-define the following bean:

```java
@Bean
MongoMetricsCommandTagsProvider mongoMetricsCommandListenerTagsProvider() {
    return new DefaultMongoMetricsCommandTagsProvider() {
        @Override
        public Iterable<Tag> commandTags(CommandEvent event) {
            return Tags.of(super.commandTags(event))
                    .and(Tag.of("mongoTemplate", "t1"))
                    .and(Tag.of("extraStuff", "foo"));
        }
    };
}
```
The auto-configuration for Spring Boot would then look like this:

```java
@Bean
@ConditionalOnMissingBean
MongoMetricsCommandTagsProvider mongoMetricsCommandListenerTagsProvider() {
    return new DefaultMongoMetricsCommandTagsProvider();
}

@Bean
@ConditionalOnMissingBean
MongoMetricsCommandListener mongoMetricsCommandListener(MeterRegistry meterRegistry,
                                                          MongoMetricsCommandTagsProvider mongoMetricsCommandTagsProvider) {
return new MongoMetricsCommandListener(meterRegistry, mongoMetricsCommandTagsProvider);
}

@Bean
MongoClientSettingsBuilderCustomizer mongoMetricsCommandListenerClientSettingsBuilderCustomizer(
  MongoMetricsCommandListener mongoMetricsCommandListener) {
return (builder) -> builder.addCommandListener(mongoMetricsCommandListener);
}

```

If we feel strongly about the extra tags I can add them here - I just changed thinking and was wanting to run by you all.

bono007:
Hi @jonatan-ivanov - I realized I did not tag you in my [follow up](https://github.com/micrometer-metrics/micrometer/pull/2360#issuecomment-776874991). More than likely you are just busy w/ OTel (we are super excited about GA of that btw) and other stuff - but just wanted to ping you in case this got missed in between other notifications etc..

jonatan-ivanov:
@bono007 Sorry for not getting back to you earlier about this. This brings us back to this: https://github.com/micrometer-metrics/micrometer/pull/2360#issuecomment-774316463

I still have the same opinion: `DefaultMongoMetricsCommandTagsProvider` eliminates the need of extra tags since as you wrote it too, you can extend it and add whatever you want. Though, I think if you need to add the extra tags, it is better from usability/API discoverability if they are there. If you don't it is just added complexity.

Since this is just an extra nice shortcut, can I make this up to you and @gustavomonarin? :)
If you would not use it, let's merge the PR's without it (we can also add it later), if it would make your life easier, we can add the possibility to provide extra tags in the ctor of `DefaultMongoMetricsCommandTagsProvider`.

bono007:
> Sorry for not getting back to you earlier about this.

No worries @jonatan-ivanov . We are on the same page - I am fine w/ merging this w/o the extra tags. @gustavomonarin do you agree?

Thanks for the follow up @jonatan-ivanov

gustavomonarin:
Sure, I agree that the provider eliminates the need of the extraTags.

The extra tags pr was closed as soon as the pr #2437 was open. The issue itself was still named extraTags, which could cause some confusion - sorry for that, i have change adjusted there as well.

jonatan-ivanov (approved):

jonatan-ivanov:
@bono007 I modified your PR:
- merged master (mongo4.x) and resolved merge conflicts (only imports were conflicting)
- added the `tagsProvider` as a javadoc `@param`
- migrated the tests to support mongo 4.x (based on the previous mongo 4.x migration, this was straightforward)

jonatan-ivanov:
@bono007 Thank you very much putting this together! :)

Issue #2354, opened 2020-11-18
CompositeMeterRegistry fails to propagate Meter removals to a MeterRegistry that has Filters


ldesgoui:
Hello,

Consider the simple set up:
Metrics.globalRegistry has no Filters applied
A DatadogMeterRegistry instance has a CommonTags Filter applied and is parented by Metrics.globalRegistry
When Meters are written to Metrics.globalRegistry, they are propagated to the DatadogMeterRegister instance and the Filters are applied
When Meters are removed from Metrics.globalRegistry, they are propagated to the DatadogMeterRegister instance but the Filters are not applied, causing the lookup to fail and them to not get properly deleted

This is caused by [this event handler](https://github.com/micrometer-metrics/micrometer/blob/4009d7baad00799079130e18a29fb9ed2f105e4c/micrometer-core/src/main/java/io/micrometer/core/instrument/composite/CompositeMeterRegistry.java#L70) which calls [this method](https://github.com/micrometer-metrics/micrometer/blob/4009d7baad00799079130e18a29fb9ed2f105e4c/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L653-L655), in which the filters are not applied.

This causes the equivalent of a memory leak and aggravates heavily with high-cardinality Meters (such as metrics from reactor executors while using "elastic" schedulers)

I don't have enough Java expertise to offer an accompanying fix and tests, apologies.

shakuzen:
Thanks for the detailed report. The behavior of the remove method has changed since it was introduced - it used to apply `MeterFilter`s, now it doesn't. We've run into the issue of one behavior being desired in some cases in not others. 

While I think we should be able to fix this regardless, can share more details about your use case for removing meters? This is just so we have a more complete picture of the background behind the use case.

shakuzen:
Here is a unit test that I believe captures the scenario (written in `CompositeMeterRegistryTest`):

```java
@Test
@Issue("#2354")
void meterRemovedInChildRegistryWithModifyingFilter() {
    this.simple.config().commonTags("host", "localhost");
    this.composite.add(this.simple);

    Counter counter = this.composite.counter("my.counter");
    this.composite.remove(counter);

    assertThat(this.composite.getMeters()).isEmpty();
    assertThat(this.simple.getMeters()).isEmpty(); // this assertion fails
}
```

ldesgoui:
[...]

PR #2362, opened 2020-11-27
Remove meters from child registries using mapped Id

shakuzen:
Child registries may have `MeterFilter` configured that alter the Id of meters when registered. In this case, calling `remove` with the Id from the `CompositeMeterRegistry` will not match unless it also has the same `MeterFilter` configured. The behavior of the `remove` method has fluctuated as these cases come to light. This adds public API for removing a meter with its pre-mapped Id and clarifies the behavior in the JavaDoc of all `remove` methods. `CompositeMeterRegistry` can now use this new API to have the expected behavior.

Resolves gh-2354

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -604,6 +604,9 @@ public abstract class MeterRegistry {
     }
 
     /**
+     * Remove a {@link Meter} from this {@link MeterRegistry registry}. This is expected to be a {@link Meter} with
+     * the same {@link Id} returned when registering a meter - which will have {@link MeterFilter}s applied to it.
+     *
      * @param meter The meter to remove
      * @return The removed meter, or null if the provided meter is not currently registered.
      * @since 1.1.0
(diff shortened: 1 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -614,15 +617,29 @@ public Meter remove(Meter meter) {
...
-    Meter remove(Meter.Id id, boolean applyFilters) {
+    public Meter remove(Meter.Id id, boolean applyFilters) {
         if (applyFilters) {
             return remove(getMappedId(id));
         }
         return remove(id);
     }
 
     /**
+     * Remove a {@link Meter} from this {@link MeterRegistry registry} based on its {@link Id} after any
```
shakuzen, line 639: Re-reading this, it's not as clear as I would like... let me try again.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -614,15 +617,29 @@ public Meter remove(Meter meter) {
...
+     *
+     * @param id the id of the meter to remove
+     * @param applyFilters whether to apply this registry's filters to the id before attempting to remove
+     * @return The removed meter, or null if the meter is not found
+     * @since 1.3.16
+     */
+    @Incubating(since = "1.3.16")
     @Nullable
-    Meter remove(Meter.Id id, boolean applyFilters) {
+    public Meter remove(Meter.Id id, boolean applyFilters) {
```
shakuzen, line 631: Since this will add public API, I was thinking about it more, and I'm not sure I like this signature, given passing `false` is the same behavior as not passing the parameter at all (existing methods). I'm wondering if it is better to make a different method name. Therein lies the naming bikeshed, though.
* `removeByPreFilterId(Meter.Id preFilterId)`
Something along the lines of this is what I'm leaning towards.

admoca60: Ok, it could be Ok, but in that case, don't forget to modify the "CompositeMeterRegistry.java" file to replace `r -> r.remove(m.getId(), true))` per `r -> r.removeByPreFilterId(m.getId()))`

Issue #1906, opened 2020-03-16
Remove configuring Elasticsearch index type


shakuzen:
As a note for future planning, once we no longer support Elasticsearch versions less than 7, we should remove this configuration option.

Based on Elasticsearch's published EOL schedule, it looks like that will be sometime around the end of this year: https://www.elastic.co/support/eol

> Elasticsearch | EOL date | Maintained until
> -- | -- | --
> 6.8.x | 2020-11-20 | 8.0.0

_Originally posted by @shakuzen in https://github.com/micrometer-metrics/micrometer/issues/1891#issuecomment-597086693_

izeye:
I created https://github.com/micrometer-metrics/micrometer/pull/2361 to try to resolve this.

PR #2363, opened 2020-11-27
Make _source.enabled configurable for ElasticMeterRegistry

izeye:
This PR changes to make `_source.enabled` configurable for `ElasticMeterRegistry`. While working on this, I noticed that there's no `_source.enabled` setting for Elasticsearch 6. So `_source` field does exist by default in Elasticsearch 6. I didn't touch the code related to Elasticsearch 6 as it seems to be going to be dropped soon in #1906.

Closes gh-1629

### DIFF
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java
@@ -181,6 +181,17 @@ public interface ElasticConfig extends StepRegistryConfig {
         return getSecret(this, "apiKeyCredentials").orElse(null);
     }
 
+    /**
+     * Enable {@literal _source} in the index template.
+     * Default is: {@code false}
+     *
+     * @return whether {@literal _source} will be enabled in the index template
+     * @since 2.0.0
+     */
+    default boolean enableSource() {
+        return getBoolean(this, "enableSource").orElse(false);
+    }
+
     @Override
     default Validated<?> validate() {
         return checkAll(this,
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
@@ -37,8 +37,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiFunction;
 import java.util.function.Consumer;
-import java.util.function.Function;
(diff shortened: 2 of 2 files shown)


izeye:
I rebased this on the current master to resolve conflicts.

shakuzen:
I think this could be useful to people during a proof-of-concept phase where they are exploring things in Kibana and setting up a new dashboard or something, but it's almost certainly not something that should ever be enabled in production. I have concern about how easy it would be for this to be mistakenly enabled in production by making it configurable. I'm not sure how to best mitigate that.

jonatan-ivanov:
@izeye Would you please add a `WARN` message to every publish that warns the users that source is enabled? This way people would be able to turn this on but will be warned so that they have a chance to catch this in prod.

izeye:
@shakuzen @jonatan-ivanov Thanks for the feedback!

I updated as @jonatan-ivanov suggested.

jonatan-ivanov (approved):

OleksandrOnopriienko:
Can anybody explain me please, in which version this functionality was included?

jonatan-ivanov:
@kubickrock Se on the side:
```
Milestone
1.10.0-M1
```

So first this was released in `1.10.0-M1`, if you want to know the first stable/GA release that's without the `-M1` so `1.10.0`.

davidfigueroaMLP:
@jonatan-ivanov please note that this change is not in 1.10 nor in 1.11. added reference to micrometer-registry-elastic 1.11 and it is not there. it is only in the 2.0 branch which has not been released. can this change be included in the next 1.x release (1.12)

https://github.com/micrometer-metrics/micrometer/blob/2.0.x/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java#L191

shakuzen:
@davidfigueroaMLP thank you for bringing that to our attention. I don't think it was intentional it got dropped. I've re-opened https://github.com/micrometer-metrics/micrometer/issues/1629 and assigned it to the 1.12 backlog.

davidfigueroaMLP:
@shakuzen would it make sense to assign it to the [1.11.x backlog](https://github.com/micrometer-metrics/micrometer/milestone/206) in hope that it will be released sooner?

shakuzen:
@davidfigueroaMLP In general we don't consider enhancements for patch releases, but we could consider making an exception. That said, I don't see this as very critical. The configuration only controls the index template we create if missing. Once the index template is created in Elasticsearch, we will not overwrite an existing one. This means for an existing Elasticsearch cluster you've been using with Micrometer, you will need to update the index template on Elasticsearch outside of your application, anyway. So this would only help if publishing your metrics to a new cluster or if you delete the index template. Am I missing anything? Is there some reason this is more important for you that I don't realize?

denistsyplakov:
Is it released?

PR #2364, opened 2020-11-27
Set index to false for metric values in ElasticMeterRegistry

izeye:
This PR changes to set `index` to `false` for metric values in `ElasticMeterRegistry`.

See https://github.com/micrometer-metrics/micrometer/issues/1629#issuecomment-565184964


shakuzen (approved):
Thank you for the pull request @izeye, and thank you for making the suggestion @xeraa.

Issue #1019, opened 2018-11-19
Common tags duplicates meter tags.


ArtyomGabeev:
Configuring common tags on registry level with combination of meter metrics leads to tags duplication.
Steps to reproduce:
1. Configure common metrics though meter registry config
2. Register metric with additional tags
3. Check registered metric id

The issue occurs in MeterFilter.commonTags which concatenates metric tags twice:
**meter tags + meter tags + common tags**

Issue found when I've tried to remove such metric from registry.
Removal logic applies all MeterFilters again, which leads to tags duplication again and break hashCode contract of Tags.class.

Example:
`
       
        SimpleMeterRegistry registry = new SimpleMeterRegistry();

        registry.config().commonTags("common-key", "common-value");

        Gauge gauge = Gauge.builder("test-gauge", () -> 0L)
                .tags("key", "value")
                .register(registry);

        System.out.print(gauge.getId());

        registry.remove(gauge);

        Assert.isTrue(registry.getMeters().isEmpty(), () -> "Registry is not empty");

`
Output:
> MeterId{name='test-gauge', tags=[tag(common-key=common-value),tag(key=value),tag(key=value)]}
> Exception in thread "main" java.lang.IllegalArgumentException: Registry is not empty

izeye:
@ArtyomGabeev Thanks for the report. This is a duplicate of #978.

tommyk-gears:
I believe this issue is still valid and not really a duplicate of #978, `MeterFilter#commonTags` looks like this;

```java
    static MeterFilter commonTags(Iterable<Tag> tags) {
        return new MeterFilter() {
            @Override
            public Meter.Id map(Meter.Id id) {
                return id.withTags(Tags.concat(tags, id.getTagsAsIterable()));
            }
        };
    }
```

But, `Meter.Id#withTags` **concatenates** the provided tags to the existing tags. The code in `MeterFilter#commonTags` seems to assume that `Meter.Id#withTags` **replaces** the existing tags with the provided tags, hence doing an external concatenation.

Now, the tag deduplication works and hides the duplicate tags - but this thing just generates lots of duplicate tags (one duplicate per common tag, for each `Meter.Id` basically). Maybe not a big deal, but the solution is really simple, just remove the `Tags.concat` call from `MeterFilter#commonTags`;

```java
    static MeterFilter commonTags(Iterable<Tag> tags) {
        return new MeterFilter() {
            @Override
            public Meter.Id map(Meter.Id id) {
[...]

PR #2370, opened 2020-12-05
Use replaceTags() in MeterFilter.commonTags()

izeye:
This PR changes to use `replaceTags()` in `MeterFilter.commonTags()`.

See gh-1019

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
@@ -45,6 +45,9 @@ public interface MeterFilter {
     /**
      * Add common tags that are applied to every meter created afterward.
      *
+     * The common tags will not override tag values from a meter ID. They will also not override previously configured
+     * common tag MeterFilters that have the same tag key.
+     *
      * @param tags Common tags.
      * @return A common tag filter.
      */
@@ -52,7 +55,7 @@ public interface MeterFilter {
         return new MeterFilter() {
             @Override
             public Meter.Id map(Meter.Id id) {
-                return id.withTags(Tags.concat(tags, id.getTagsAsIterable()));
+                return id.replaceTags(Tags.concat(tags, id.getTagsAsIterable()));
             }
         };
     }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
@@ -227,10 +227,10 @@ public interface Meter {
         }
 
         /**
-         * Generate a new id with an additional tag. If the key of the provided tag already exists, this overwrites
+         * Generate a new id with additional tags. If the key of the provided tag already exists, this overwrites
          * the tag value.
          *
-         * @param tags The tag to add.
+         * @param tags The tags to add.
          * @return A new id with the provided tags added. The source id remains unchanged.
          * @since 1.1.0
          */
@@ -241,7 +241,7 @@ public interface Meter {
         /**
          * Generate a new id replacing all tags with new ones.
          *
-         * @param tags The tag to add.
+         * @param tags The tags to add.
          * @return A new id with the only the provided tags. The source id remains unchanged.
          * @since 1.1.0
          */
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/MeterFilterTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/MeterFilterTest.java
@@ -61,6 +61,15 @@ class MeterFilterTest {
     }
 
     @Test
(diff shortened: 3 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
```diff
@@ -52,7 +52,7 @@ static MeterFilter commonTags(Iterable<Tag> tags) {
         return new MeterFilter() {
             @Override
             public Meter.Id map(Meter.Id id) {
-                return id.withTags(Tags.concat(tags, id.getTagsAsIterable()));
+                return id.replaceTags(Tags.concat(tags, id.getTagsAsIterable()));
```
shakuzen, line 55: If I'm understanding correctly, the motivation is to save one `Tags#concat()` call but get the same result. Is that correct? This is making me think we should have more tests and more clear JavaDocs about the behavior around common tags and maybe these withTags/replaceTags methods.

izeye: @shakuzen Correct. I updated this to add some tests and update their Javadoc as suggested. Let me know what you think.

izeye: For the record, I didn't update Javadoc for `withTags()` with its current behavior as I wasn't sure if it was intentional.

izeye: We might also want to update Javadoc for sorting and deduplication logic in `Tags`.

micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
```diff
@@ -45,14 +45,16 @@
     /**
      * Add common tags that are applied to every meter created afterward.
      *
+     * The common tags will not override tag values from a meter ID.
```
shakuzen, line 48: They will also not override previously configured common tag MeterFilters that have the same tag key.

izeye: @shakuzen Thanks for the feedback! I didn't include it as I thought the statement implied it. However, being explicit might be better. I added it in c971d65 as suggested.

shakuzen (approved):
Thanks! Let's apply this to 1.3.x since the behavior hasn't changed and we're testing/documenting that better now. I made a quick JMH benchmark to test the performance difference with the following results:

```
Benchmark                                                         Mode  Cnt     Score     Error   Units
CommonTagsBenchmark.replaceTags                                   avgt    9    96.878 ±   4.341   ns/op
CommonTagsBenchmark.replaceTags:·gc.alloc.rate                    avgt    9  1500.667 ±  67.458  MB/sec
CommonTagsBenchmark.replaceTags:·gc.alloc.rate.norm               avgt    9   160.013 ±   0.001    B/op
CommonTagsBenchmark.replaceTags:·gc.churn.G1_Eden_Space           avgt    9  1503.793 ±  67.637  MB/sec
CommonTagsBenchmark.replaceTags:·gc.churn.G1_Eden_Space.norm      avgt    9   160.353 ±   2.634    B/op
CommonTagsBenchmark.replaceTags:·gc.churn.G1_Survivor_Space       avgt    9     0.006 ±   0.003  MB/sec
CommonTagsBenchmark.replaceTags:·gc.churn.G1_Survivor_Space.norm  avgt    9     0.001 ±   0.001    B/op
CommonTagsBenchmark.replaceTags:·gc.count                         avgt    9   474.000            counts
CommonTagsBenchmark.replaceTags:·gc.time                          avgt    9   331.000                ms
CommonTagsBenchmark.withTags                                      avgt    9   226.051 ±  15.164   ns/op
CommonTagsBenchmark.withTags:·gc.alloc.rate                       avgt    9  1992.846 ± 116.868  MB/sec
CommonTagsBenchmark.withTags:·gc.alloc.rate.norm                  avgt    9   496.037 ±  40.333    B/op
CommonTagsBenchmark.withTags:·gc.churn.G1_Eden_Space              avgt    9  1995.505 ± 118.433  MB/sec
CommonTagsBenchmark.withTags:·gc.churn.G1_Eden_Space.norm         avgt    9   496.692 ±  40.389    B/op
CommonTagsBenchmark.withTags:·gc.churn.G1_Survivor_Space          avgt    9     0.007 ±   0.006  MB/sec
CommonTagsBenchmark.withTags:·gc.churn.G1_Survivor_Space.norm     avgt    9     0.002 ±   0.002    B/op
CommonTagsBenchmark.withTags:·gc.count                            avgt    9   629.000            counts
CommonTagsBenchmark.withTags:·gc.time                             avgt    9   433.000                ms
```

Faster and less allocation with these changes.

izeye:
@shakuzen Thanks for the benchmark! I changed the base branch to 1.3.x now.

PR #2392, opened 2020-12-29
Mecrometer-Logzio

fadi-khatib:
Integrating Micrometer with Logz.io.
Will allow Micrometer users to send metrics to logz.io.


Doron-Bargo:
Hi @shakuzen , can you help here?

yyyogev:
Hi @shakuzen, any ETA for review this PR?

shakuzen:
Thank you for the pull request. It looks quite complete. Sorry for the delay on responding here. I thought we had an open issue about Logz.io support, but I can't find it so I must be misremembering. I was looking for documentation on sending metrics to Logz.io, and I came across [this page about Micrometer integration](https://docs.logz.io/shipping/prometheus-sources/java-custom-metrics-micrometer.html), which led me to the existing GitHub repository: https://github.com/logzio/micrometer-registry-logzio.

I think that repository is probably a better place for this integration, at least for now, since you all will be able to provide support specific to Logz.io better than the Micrometer team can, and we are already maintaining many backends. What do you think about adding documentation for it to our [Micrometer docs site](https://github.com/micrometer-metrics/micrometer-docs) instead of moving the implementation here?

PR #2396, opened 2021-01-09
Pin jackson-databind to 2.12.0 temporarily

izeye:
Builds on master currently fail due to https://github.com/FasterXML/jackson-bom/issues/37. This PR pins `jackson-databind` to 2.12.0 temporarily to restore builds.


izeye:
https://github.com/FasterXML/jackson-bom/issues/37 seems to be fixed now.

PR #2402, opened 2021-01-12
Make `MongoMetricsCommandListener` thread safe by not sharing `Timer.Builder`

### LINKED ISSUE #2401: `MongoMetricsCommandListener` is not thread safe
`MongoMetricsCommandListener` is not thread safe because it uses a shared instance of `Timer.Builder`. This means that concurrent calls to `commandSucceeded()` or `commandFailed()` can generate incorrect metrics.

I've created a unit test here that reproduces the issue: https://github.com/jshields-squarespace/micrometer/commit/80d0fd6733ba3567a0275e231a93f97ed0dcc1bb

This test creates 4 threads that execute different Mongo operations, then checks that the metric corresponding to each command has been incremented by 1. The test fails when the threads are run concurrently (`concurrent = true`), but passes when the threads are run in serial (`concurrent = false`).

Example concurrent output:

```
[mongod output] iterationsCompleted=1, insertCommandCount=1, updateCommandCount=2, deleteCommandCount=0, dropIndexesCommandCount=1

Expecting:
 <2L>
to be equal to:
 <1L>
but was not.
```

Example serial output:

```
[mongod output] iterationsCompleted=100, insertCommandCount=100, updateCommandCount=100, deleteCommandCount=100, dropIndexesCommandCount=100
```

jshields-squarespace:
As detailed in issue #2401, `MongoMetricsCommandListener` is not thread safe because it uses a shared instance of `Timer.Builder`. This means that concurrent calls to `commandSucceeded()` or `commandFailed()` can generate incorrect metrics.

This PR makes `MongoMetricsCommandListener` thread safe by generating a new `Timer.Builder` for each call instead of reusing a shared instance.

As far as I can tell, instantiating a new `Timer.Builder` is not an expensive operation.

Fixes #2401

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java
@@ -36,9 +36,6 @@ import java.util.concurrent.TimeUnit;
 @Incubating(since = "1.2.0")
 public class MongoMetricsCommandListener implements CommandListener {
 
-    private final Timer.Builder timerBuilder = Timer.builder("mongodb.driver.commands")
-            .description("Timer of mongodb commands");
-
     private final MeterRegistry registry;
 
     public MongoMetricsCommandListener(MeterRegistry registry) {
@@ -60,7 +57,8 @@ public class MongoMetricsCommandListener implements CommandListener {
     }
 
     private void timeCommand(CommandEvent event, String status, long elapsedTimeInNanoseconds) {
-        timerBuilder
+        Timer.builder("mongodb.driver.commands")
+                .description("Timer of mongodb commands")
                 .tag("command", event.getCommandName())
                 .tag("cluster.id", event.getConnectionDescription().getConnectionId().getServerId().getClusterId().getValue())
                 .tag("server.address", event.getConnectionDescription().getServerAddress().toString())
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
@@ -22,6 +22,7 @@ import com.mongodb.event.ClusterListenerAdapter;
 import com.mongodb.event.ClusterOpeningEvent;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tags;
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@jshields-squarespace Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2402)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2402) to manually synchronize the status of this Pull Request.

shakuzen (approved):
The changes look good to me. Thank you for the thorough investigation and the pull request. Would you also be able to contribute a version of the test you shared in the issue? It will be good to have such a test to ensure this doesn't regress through any future changes.

shakuzen:
Also, if not too much trouble, could you rebase on and change the target branch to `1.3.x` so we can easily merge forward the changes from there for inclusion in upcoming patch releases? I can also do this when merging, so not a big deal either way.

jshields-squarespace:
Sure, I can rebase and change the target branch.

Regarding the test case: the reason I didn't add it to this PR is because from my understanding it's impossible to write concurrency / race condition tests that always fail in the event of a race condition. Maybe it fails on one machine, but passes on another. Or maybe it fails for some race conditions, but not all of them. So the test can be a little misleading.

That said, it could still provide some value, and I'm OK adding it to this PR if that's what you'd prefer.

shakuzen:
> That said, it could still provide some value, and I'm OK adding it to this PR if that's what you'd prefer.

I agree, it would not be a deterministic proof of lack of concurrency issue, but it could provide some possibility of catching this regressing. There's the possibility the false sense of security is a detriment, but as long as we're aware of the limitation, I think it is better to have than not. If we wanted to go hardcore on it, we could write a [jcstress test](https://github.com/openjdk/jcstress), but I don't want to block getting this fix in on that. I have been meaning to add some jcstress tests for other areas of the code but have not done so yet. We can probably revisit testing this concurrency correctness with jcstress once we have another example in this codebase.

jshields-squarespace:
Sounds good. I'll put up another commit that adds a version of the unit test to this PR.

pivotal-issuemaster:
@jshields-squarespace Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2402)!

jshields-squarespace:
@shakuzen I've added a unit test for thread safety. Let me know what you think.

shakuzen (approved):
Looks good to me. Thanks!

PR #2406, opened 2021-01-13
Add dynatrace meter registry for metrics api v2

### LINKED ISSUE #2295: Support for Dynatrace metrics API v2
Dynatrace has released a new metric ingestion API much flexible than previous one.
https://www.dynatrace.com/support/help/dynatrace-api/environment-api/metric-v2/post-ingest-metrics/

Ingestion flow has changed significantly:
* no need to preregister metrics
* metrics are not constrained to a custom device anymore
* more flexible naming conventions (eg. no `custom:` prefix)
* payload type as [line protocol](https://www.dynatrace.com/support/help/how-to-use-dynatrace/metrics/metric-ingestion/metric-ingestion-protocol/)
* possibility to ingest gauges and counters

Since there are quite changes and new api is still in EAP, I'd propose to create a new shinny meter registry, let's say `micrometer-registry-dynatrace2` and deprecate the first version`micrometer-registry-dynatrace` so first version would be still available to be used to do not lose the history of the custom metrics registered with v1.

If is ok from your side, I'd try to contribute in this new registry in the upcoming days.

damass:
Fix #2295  - Adding new implementation (dynatrace2)  to make use of Dynatrace Metrics API v2

### DIFF
--- a/implementations/micrometer-registry-dynatrace2/src/main/java/io/micrometer/dynatrace2/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace2/src/main/java/io/micrometer/dynatrace2/DynatraceConfig.java
@@ -45,6 +45,9 @@ public interface DynatraceConfig extends StepRegistryConfig {
     default String entityId() { return getString(this,"entityId").orElse(""); }
 
     default String apiToken() {
+        if (this.uri().contains("localhost")){
+            return getSecret(this, "apiToken").orElse("");
+        }
         return getSecret(this, "apiToken").required().get();
     }
 
@@ -64,7 +67,7 @@ public interface DynatraceConfig extends StepRegistryConfig {
                 check("deviceName", DynatraceConfig::deviceName),
                 check("groupName", DynatraceConfig::groupName),
                 check("entityId", DynatraceConfig::entityId),
-                checkRequired("apiToken", DynatraceConfig::apiToken),
+                check("apiToken", DynatraceConfig::apiToken),
                 checkRequired("uri", DynatraceConfig::uri),
                 check("batchSize", DynatraceConfig::batchSize)
                         .andThen(invalidateWhenGreaterThan(MAX_METRIC_LINES_PER_REQUEST))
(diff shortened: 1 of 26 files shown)


implementations/micrometer-registry-dynatrace2/src/main/java/io/micrometer/dynatrace2/DynatraceConfig.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    default String deviceName() { return getString(this,"deviceName").orElse(""); }
+
+    default String groupName() { return getString(this,"groupName").orElse(""); }
+
+    default String entityId() { return getString(this,"entityId").orElse(""); }
+
+    default String apiToken() {
+        if (this.uri().contains("localhost")) {
```
jonatan-ivanov, line 48: This is always required for V1, did V1 mandate auth while V2 does not?

implementations/micrometer-registry-dynatrace2/src/main/java/io/micrometer/dynatrace2/Streams.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.dynatrace2;
+
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+class Streams {
```
jonatan-ivanov, line 21: Why is this needed, would not calling `StreamSupport.stream` be ok (the rest of the codebase does that)?
I think if we can agree that dropping the parallel parameter has some value that worth the change, we can do it everywhere but I would do it in a separate PR.

jonatan-ivanov (changes requested):
Thank you for the PR! :)

I haven't finished reviewing this but would you please merge master back to your branch?
There is a trivial merge conflict and there were some checkstyle changes so this will fail after the merge.

jonatan-ivanov (changes requested):

jonatan-ivanov:
Closing in favor of #2581

Issue #2415, opened 2021-01-25
Improve performance while initializing KafkaMetrics


kptfh:
This code in `KafkaMetrics` leads to quadratic complexity. 
```
metrics.forEach((name, metric) -> {
    for (Meter other : registryMetersByNames.getOrDefault(meterName, Collections.emptyList())) {
    }
}
```
Faced very poor performance while running  a lot of Kafka consumers and producers (over 400)

jonatan-ivanov:
Fixed in https://github.com/micrometer-metrics/micrometer/pull/2416

PR #2416, opened 2021-01-25
Improve performance while initializing KafkaMetrics

kptfh:
https://github.com/micrometer-metrics/micrometer/issues/2415

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
@@ -28,6 +28,7 @@ import io.micrometer.core.lang.NonNullFields;
 
 import java.time.Duration;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


shakuzen (commented):
I don't see how this changes the iterations required. Grouping by meter name would need to be done outside of the `metrics.forEach` loop to reduce iterations, unless I'm missing something.

kptfh:
Hi. Sure. It should be outside the loop. Inaccurately ported this change. Will fix.
Thank you for review.

kptfh:
fixed

jonatan-ivanov:
@kptfh It seems the compiler does not like it, could you please take a look?

siim-romanov:
Probably just needs to catch up with this change in master
https://github.com/micrometer-metrics/micrometer/pull/2418

kptfh:
@jonatan-ivanov @siim-romanov fixed

kptfh:
Hi guys! Do you have any comments regarding this pr?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -172,12 +177,15 @@ void checkAndBindMetrics(MeterRegistry registry) {
                 // Kafka has metrics with lower number of tags (e.g. with/without topic or partition tag)
                 // Remove meters with lower number of tags
                 boolean hasLessTags = false;
-                for (Meter other : registry.find(meterName).meters()) {
+                List<Meter> metersForName = registryMetersByNames.getOrDefault(meterName, emptyList());
+                List<Meter> metersWithLessTags = new ArrayList<>(metersForName.size());
+                for (Meter other : metersForName) {
```
jonatan-ivanov, line 182: Is `metersWithLessTags` and `metersForName` really used here?
How about doing this:
```suggestion
                for (Meter other : registryMetersByNames.getOrDefault(meterName, emptyList())) {
```

And also removing `metersWithLessTags.add(other);` and `metersForName.removeAll(metersWithLessTags);`

jonatan-ivanov (changes requested):

jonatan-ivanov:
@kptfh Thanks for the fix, could you please rebase this to `1.6.x`?

jonatan-ivanov (approved):

PR #2430, opened 2021-02-05
Suppress deprecation warnings from MetricsDSLContext

izeye:
This PR suppresses deprecation warnings from `MetricsDSLContext`.


jonatan-ivanov:
This was heroic. :)

PR #2431, opened 2021-02-05
Making Tags::concat more flexible by passing a covariant type

### LINKED ISSUE #2092: `Iterable<Tag>` vs `Iterable<? extends Tag>`
I noticed that Micrometer methods sometimes accept `Iterable<? extends Tag>`:
```sh
$ git grep 'Iterable<? extends Tag>'
micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java:    public Tags and(@Nullable Iterable<? extends Tag> tags) {
micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java:    public static Tags concat(@Nullable Iterable<? extends Tag> tags, @Nullable Iterable<Tag> otherTags) {
micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java:    public static Tags concat(@Nullable Iterable<? extends Tag> tags, @Nullable String... keyValues) {
micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java:    public static Tags of(@Nullable Iterable<? extends Tag> tags) {
```
... but mostly require `Iterable<Tag>`:
```sh
$ git grep -l 'Iterable<Tag>' | wc -l
88
```

Two questions:
- For consistency, shouldn't the second argument to `Tags.concat` also accept `Iterable<? extends Tag>`?
- Since `Tag` is an interface and even has a public subtype provided by Micrometer itself (`ImmutableTag`), shouldn't most/all other methods also accept an `Iterable<? extends Tag>`?

(I'm raising this question mostly because it came up while writing a [Refaster](https://errorprone.info/docs/refaster) template. I can understand if the suggested changes aren't worth potentially breaking client code, even though I expect that the suggested changes are mostly safe. If there's interest in this change I'm willing to prepare a PR.)

jonatan-ivanov:
Fixes #2092

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java
@@ -191,7 +191,7 @@ public final class Tags implements Iterable<Tag> {
      * @param otherTags the second set of tags
      * @return the merged tags
      */
-    public static Tags concat(@Nullable Iterable<? extends Tag> tags, @Nullable Iterable<Tag> otherTags) {
+    public static Tags concat(@Nullable Iterable<? extends Tag> tags, @Nullable Iterable<? extends Tag> otherTags) {
         return Tags.of(tags).and(otherTags);
     }
 


shakuzen:
My only concern would be the one already mentioned https://github.com/micrometer-metrics/micrometer/issues/2092#issuecomment-774541470. If you've tested this and can confirm it is abi compatible, then we're good to merge this.

PR #2439, opened 2021-02-07
Fix MultipleIndicator.combine()

izeye:
`MultipleIndicator.combine()` has a parameter named `objectives` somehow, so it uses the parameter instead of `this.objectives`.

This PR changes to use `this.objectives` simply by removing the parameter that seems to be added accidentally.


jonatan-ivanov (changes requested):
Should not this go to 1.7?

izeye:

I thought it needs to be fixed on the 1.6.x somehow. If the broken compatibility is unacceptable, how about deprecating the current one in favor of the new one?

izeye:
I changed to preserve the previous method with deprecation in 7b7c8d0.

jonatan-ivanov:
Thanks! :)

izeye:
@jonatan-ivanov Thanks for the review!

Micrometer 1.6.4 has been released already, so the Javadoc `@since` tag has been outdated now. I created #2473 to resolve it.

PR #2442, opened 2021-02-08
Inject required meter binders into predefined SLOs

izeye:
There are some `MeterBinder`s like `JvmHeapPressureMetrics` that implement `AutoCloseable`, and it's not possible to manage their lifecycles in the predefined SLOs.

One possible solution is to inject required meter binders into the predefined SLOs. This PR changes to do so to try to resolve it.


izeye:
FYI: https://github.com/micrometer-metrics/micrometer/pull/2442/files?w=1 is better to review.

Issue #2384, opened 2020-12-10
NPE in `MongoMetricsConnectionPoolListener` on shutdown


jshields-squarespace:
With Micrometer 1.3.14, during application shutdown we get NPEs from `MongoMetricsConnectionPoolListener`.

This looks related to #2117. However that was supposedly fixed in 1.3.10, and we're experiencing this issue with 1.3.14.

It seems like `checkedOutCount` no longer has an entry for the server ID, presumably because `connectionPoolClosed` was already called during the application shutdown.

Stack trace:

```
java.lang.NullPointerException: null
	at io.micrometer.core.instrument.binder.mongodb.MongoMetricsConnectionPoolListener.connectionCheckedIn(MongoMetricsConnectionPoolListener.java:91) ~[micrometer-core-1.3.14.jar:1.3.14]
	at com.mongodb.internal.connection.DefaultConnectionPool$PooledConnection.close(DefaultConnectionPool.java:411) ~[mongo-java-driver-3.11.2.jar:?]
	at com.mongodb.internal.connection.DefaultServerConnection.release(DefaultServerConnection.java:69) ~[mongo-java-driver-3.11.2.jar:?]
	at com.mongodb.operation.CommandOperationHelper$13.call(CommandOperationHelper.java:721) ~[mongo-java-driver-3.11.2.jar:?]
	at com.mongodb.operation.OperationHelper.withReleasableConnection(OperationHelper.java:501) ~[mongo-java-driver-3.11.2.jar:?]
	at com.mongodb.operation.CommandOperationHelper.executeRetryableCommand(CommandOperationHelper.java:703) ~[mongo-java-driver-3.11.2.jar:?]
	at com.mongodb.operation.CommandOperationHelper.executeRetryableCommand(CommandOperationHelper.java:695) ~[mongo-java-driver-3.11.2.jar:?]
	at com.mongodb.operation.BaseFindAndModifyOperation.execute(BaseFindAndModifyOperation.java:70) ~[mongo-java-driver-3.11.2.jar:?]
	at com.mongodb.client.internal.MongoClientDelegate$DelegateOperationExecutor.execute(MongoClientDelegate.java:206) ~[mongo-java-driver-3.11.2.jar:?]
	at com.mongodb.client.internal.MongoCollectionImpl.executeFindOneAndUpdate(MongoCollectionImpl.java:792) ~[mongo-java-driver-3.11.2.jar:?]
[...]

PR #2454, opened 2021-02-16
Guard against NPE in MongoMetricsConnectionPoolListener

shakuzen:
Logically, the other events should not happen after the connection pool has been closed, but it seems they sometimes do. When they do (e.g. on shutdown), a NPE was being thrown. This change guards against such NPEs.

Resolves gh-2384


jshields-squarespace:
Thanks for this!

Issue #2461, opened 2021-02-20
Align TimedAspect and CountedAspect precedence of extraTags and tagsBasedOnJoinPoint


kinddevil:
When overwriting the `extraTags` in Aspect, the behavior is different - `extraTags` can be replaced in the `TimedAspect` but in `CountedAspect` as below.

I have a function with extra tags like
```
@Timed(extraTags = {"extra", "A"})
@Counted(extraTags = {"extra", "B"})
public void recordMetrics() {
...
}
```

When I overwrite the `extra` tag in `TimedAspect` 
```
@Bean
  public TimedAspect timedAspect(MeterRegistry registry) {
    return new TimedAspect(registry, p -> Arrays.asList(Tag.of("extra", "AA"));
  }
```
The `extra` changed to `AA` successfully.
But when I use some similar code in `CountedAspect` it failed.
```
  @Bean
  public CountedAspect countedAspect(MeterRegistry registry) {
    return new CountedAspect(registry,  p -> Arrays.asList(Tag.of("extra", "BB"));
  }
```
The value is still `B`.

It is because of the different orders when applying the `ProceedingJoinPoint` tags https://github.com/micrometer-metrics/micrometer/blob/939e0e4a36cc3bf05785c6436f1f7a72e710c9d9/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java#L148 and https://github.com/micrometer-metrics/micrometer/blob/939e0e4a36cc3bf05785c6436f1f7a72e710c9d9/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java#L192.

Can we make them in the same order?

kinddevil:
There is a fix for the issue https://github.com/micrometer-metrics/micrometer/pull/2462, can someone take a look? Thank you!

kinddevil:
Hi @shakuzen, do you have any suggestions for this issue?

PR #2462, opened 2021-02-20
Align precedence of applying extra tags and tagsBasedOnJoinPoint between TimedAspect and CountedAspect

kinddevil:
## What problem does this PR solve?
The bahaviors between `TimedAspect` and `CountdAspect` for `extraTags` are different -
- `TimedAspect` applies `ProceedingJoinPoint` tags after `extraTags`.
- `CountdAspect` applies `ProceedingJoinPoint` tags before `extraTags`.

It causes issues when overwriting `extraTags` with customized `JoinPoint`.

## What is changed and how it works?
Move `extraTags` before applying `ProceedingJoinPoint` tags in `CountdAspect`.

Related issue
Resolves https://github.com/micrometer-metrics/micrometer/issues/2461

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
@@ -256,23 +256,16 @@ public class CountedAspect {
     }
 
     private void record(ProceedingJoinPoint pjp, Counted counted, String exception, String result) {
-        counter(pjp, counted).tag(EXCEPTION_TAG, exception)
-            .tag(RESULT_TAG, result)
+        Counter.Builder builder = Counter.builder(counted.value())
+            .description(counted.description().isEmpty() ? null : counted.description())
             .tags(counted.extraTags())
-            .register(registry)
-            .increment();
-    }
-
-    private Counter.Builder counter(ProceedingJoinPoint pjp, Counted counted) {
-        Counter.Builder builder = Counter.builder(counted.value()).tags(tagsBasedOnJoinPoint.apply(pjp));
-        String description = counted.description();
-        if (!description.isEmpty()) {
-            builder.description(description);
-        }
+            .tag(EXCEPTION_TAG, exception)
+            .tag(RESULT_TAG, result)
+            .tags(tagsBasedOnJoinPoint.apply(pjp));
         if (meterTagAnnotationHandler != null) {
             meterTagAnnotationHandler.addAnnotatedParameters(builder, pjp);
         }
-        return builder;
+        builder.register(registry).increment();
     }
 
     /**
--- a/samples/micrometer-samples-spring-framework6/build.gradle
+++ b/samples/micrometer-samples-spring-framework6/build.gradle
@@ -13,6 +13,7 @@ dependencies {
     implementation project(":micrometer-observation")
 
     testImplementation project(":micrometer-observation-test")
+    testImplementation project(":micrometer-test")
     testImplementation(libs.aspectjweaver)
     testImplementation libs.awaitility
     testImplementation(libs.contextPropagation)
--- a/samples/micrometer-samples-spring-framework6/src/test/java/io/micrometer/samples/spring6/aop/CountedAspectTest.java
+++ b/samples/micrometer-samples-spring-framework6/src/test/java/io/micrometer/samples/spring6/aop/CountedAspectTest.java
@@ -17,12 +17,15 @@ package io.micrometer.samples.spring6.aop;
 
 import io.micrometer.common.annotation.ValueExpressionResolver;
 import io.micrometer.common.annotation.ValueResolver;
+import io.micrometer.core.Issue;
 import io.micrometer.core.annotation.Counted;
 import io.micrometer.core.aop.CountedAspect;
 import io.micrometer.core.aop.CountedMeterTagAnnotationHandler;
 import io.micrometer.core.aop.MeterTag;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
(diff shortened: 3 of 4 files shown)


pivotal-issuemaster:
@kinddevil Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2462)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2462) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@kinddevil Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2462)!

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -145,20 +145,17 @@ private void recordCompletionResult(ProceedingJoinPoint pjp, Counted counted, Th
...
-        String description = counted.description();
-        if (!description.isEmpty()) {
-            builder.description(description);
+        try {
+            Counter.builder(counted.value())
+                    .description(counted.description().isEmpty() ? null : counted.description())
+                    .tags(counted.extraTags())
+                    .tag(EXCEPTION_TAG, exception)
+                    .tag(RESULT_TAG, result)
+                    .tags(tagsBasedOnJoinPoint.apply(pjp))
```
elma0, line 154: Can we rely here on order of tags methods? Could tags with the same key be swapped after sort() and decouple() in Tags.of()?

jonatan-ivanov:
@kinddevil Thank you for the issue and the PR and sorry for not getting back to you earlier.
What do you think about making it possible to override jointpoint tags with annotation tags?
That way the more specific thing (annotation tags can be different for every method) can override the broader thing (jointpoint tags are for every annotated method)?

kinddevil:
> @kinddevil Thank you for the issue and the PR and sorry for not getting back to you earlier. What do you think about making it possible to override jointpoint tags with annotation tags? That way the more specific thing (annotation tags can be different for every method) can override the broader thing (jointpoint tags are for every annotated method)?
> 
> There is a weird scenario though: if a method is already annotated and the user can’t change it, they won’t be able to programmatically override it using jointpoint tags. Have you bumped into this issue?

Hi @jonatan-ivanov, thanks for the reply!
Our requirement is to have a dynamic tag according to the input parameter. For example
```
@Counted(extraTags = {"status", "$status"})
public void recordMetrics(Enum status) {
...
}
```
What we want is to change the `status` tag based on the input parameter like `SUCCESS` `ERROR` etc.

Besides this requirement, the current behaviors/orders between `timed` and `counted` are different as described in https://github.com/micrometer-metrics/micrometer/issues/2461.
If we don't want to override the annotation tags by design, instead of this change, do you think we need to change `timed` to make the annotation unchangeable? Since the current behaviors in `timed` and `counted` are not aligned.
More details:
https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java#L301
https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java#L232

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
kinddevil: Yes the order is deterministic w/ `tags`. 
There is no place to sort or decouple tags, can you give more context for this case?

shakuzen: I believe they're referring to the implementation of `Tags` which sorts and deduplicates tags with the same key, but it is designed so that later tags will take precedence over earlier tags.

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -145,20 +145,17 @@ private void recordCompletionResult(ProceedingJoinPoint pjp, Counted counted, Th
...
-                .register(meterRegistry)
-                .increment();
-    }
-
-    private Counter.Builder counter(ProceedingJoinPoint pjp, Counted counted) {
-        Counter.Builder builder = Counter.builder(counted.value()).tags(tagsBasedOnJoinPoint.apply(pjp));
-        String description = counted.description();
-        if (!description.isEmpty()) {
-            builder.description(description);
+        try {
```
shakuzen, line 148: Why add the try-catch? Is it because the user-provided `tagsBasedOnJoinPoint` function might throw an exception?

shakuzen (approved):
Thanks for the pull request and reporting the issue. Sorry it sat unmerged for such a very long time. I think the proposal makes sense. I'll work on rebasing on top of the latest changes and get it merged.

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
shakuzen: I opened a separate issue for this: https://github.com/micrometer-metrics/micrometer/issues/5584

kinddevil: Yes `tagsBasedOnJoinPoint` may throw an exception and we may want to handle the expectation explicitly.
Do you prefer a fix in the current PR or a separate one based on https://github.com/micrometer-metrics/micrometer/issues/5584?

shakuzen: I think it will be better to handle that separately from this PR. I opened #5585 for it. Thanks for bringing it to our attention.

jonatan-ivanov (approved):
I polished the tests a bit and rebased the branch on `main`.

Issue #2455, opened 2021-02-17
Remove Bintray from release process


shakuzen:
We currently promote our release artifacts to Maven Central through Bintray. Given the [sunsetting of Bintray/Jcenter](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/), we will need to adjust our release process.

PR #2467, opened 2021-02-23
Replace jcenter() with mavenCentral()

izeye:
This PR replaces `jcenter()` with `mavenCentral()`.

See gh-2455

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
     repositories {
-        jcenter()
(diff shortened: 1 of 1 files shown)


shakuzen:
I did wonder if everything needed for the build is available in central, but I suppose the passing build answers that question.

jonatan-ivanov:
@shakuzen
>the passing build answers that question

On 1.3 :) I'm going to do the merge and let's see.

PR #2468, opened 2021-02-23
Correct PostgreSQL metrics for dead tuples

easimon:
1. `pg_stat_user_tables` has a row for each table in the database, so selecting the first row of it results in "dead rows for a single random table". Calculating the sum over `n_dead_tup` results in all dead rows instead.

2. `pg_stat_database` has only one row for each database, so the sum() over `numbackends` was harmless but unnecessary, so I removed that one.

3. At least for PostgreSQL, there's a difference between `getObject(n, Long.class)` and `getLong(n)`. The former can only convert INT4 and INT8 (but not NUMERIC and DECIMAL) to Long, the latter converts all numeric types. `sum(n_dead_tup)` is such a case where this matters, since its type is NUMERIC.

See https://github.com/micrometer-metrics/micrometer/pull/2474 as well, which is a larger refactoring, but also has integration tests for this change.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
@@ -44,7 +44,7 @@ public class PostgreSQLDatabaseMetrics implements MeterBinder {
 
     private static final String SELECT = "SELECT ";
 
-    private static final String QUERY_DEAD_TUPLE_COUNT = getUserTableQuery("n_dead_tup");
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@easimon Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2468)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2468) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@easimon Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2468)!

shakuzen:
Superseded by #2474

PR #2470, opened 2021-02-23
Polish

izeye:
This PR polishes 8f117bacd70e8f6900cf66072b4bb0c25b28ff66.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -45,7 +45,7 @@ modified copy of the PCollections library by Harold Cooper.
   * Homepage: https://github.com/hrldcpr/pcollections
 
 This product contains a modified portion of the 'org.springframework.lang'
-package in the Spring Framework library, distributed by Pivotal, Inc:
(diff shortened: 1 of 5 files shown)


jonatan-ivanov:
Thanks! :)

NOTICE
```diff
@@ -45,7 +45,7 @@ modified copy of the PCollections library by Harold Cooper.
   * Homepage: https://github.com/hrldcpr/pcollections
 
 This product contains a modified portion of the 'org.springframework.lang'
-package in the Spring Framework library, distributed by Pivotal, Inc:
+package in the Spring Framework library, distributed by VMware, Inc:
```
shakuzen, line 48: This was intentional to match Spring Framework's [notice.txt](https://github.com/spring-projects/spring-framework/blob/8c9c59eba0bf1d8efd01d4130e46f158c5bf9df0/src/docs/dist/notice.txt#L2).
Either is probably fine, but I'm not a lawyer.

PR #2474, opened 2021-02-24
Correct PostgreSQL metrics for dead tuples, with integration tests and refactoring

easimon:
Larger version of https://github.com/micrometer-metrics/micrometer/pull/2468

- Corrected PostgreSQL metrics for dead tuples
- Refactored PostgreSQL metrics, extracted String constants for less repetitions.
- Added integration tests to verify (esp.) the dead rows gauge.

1. `pg_stat_user_tables` has a row for each table in the database, so selecting the first row of it results in "dead rows for a single random table". Calculating the sum over `n_dead_tup` results in all dead rows instead.

2. `pg_stat_database` has only one row for each database, so the sum() over `numbackends` was harmless but unnecessary, so I removed that one.

3. At least for PostgreSQL, there's a difference between `getObject(n, Long.class)` and `getLong(n)`. The former can only convert INT4 and INT8 (but not NUMERIC and DECIMAL) to Long, the latter converts all numeric types. `sum(n_dead_tup)` is such a case where this matters, since its type is NUMERIC.

I could understand if someone finds this too much of a change, so I created both MRs.


shakuzen (approved):
Looks good to me. Thank you for finding the issue, sending a fix, and adding integration tests for us. Much appreciated. I will rebase the changes on our `1.7.x` branch so that we can include it in our next round of patch releases.

shakuzen:
Sorry for the long wait to get this in. Thank you.

PR #2475, opened 2021-02-25
Add tag "loggerName" to LogbackMetrics

michalkowol:


pivotal-issuemaster:
@michalkowol Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2475)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2475) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@michalkowol Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2475)!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
```diff
@@ -182,27 +158,98 @@ public FilterReply decide(Marker marker, Logger logger, Level level, String form
...
         }
 
+        String loggerName = logger.getName();
+
         // cannot use logger.isEnabledFor(level), as it would cause a StackOverflowError by calling this filter again!
         if (level.isGreaterOrEqual(logger.getEffectiveLevel())) {
             switch (level.toInt()) {
                 case Level.ERROR_INT:
-                    errorCounter.increment();
+                    getOrCreateErrorCounter(loggerName).increment();
```
shakuzen, line 167: "get or create" is already the behavior when registering a meter, so I don't think the Maps add value here unless I'm missing something.

shakuzen (commented):
Thanks for the pull request. This is interesting and I can see how it would be useful. I do worry if it is a good default, however, given the potential for high cardinality. Perhaps it would be better to make this possible to configure without it being the default. We would also want to do something similar for our Log4j2Metrics.

Issue #2465, opened 2021-02-23
Update Caffeine


jonatan-ivanov:
Caffeine `3.0.0` has been released recently (https://search.maven.org/artifact/com.github.ben-manes.caffeine/caffeine) with some breaking changes, micrometer supports 2.x (https://github.com/micrometer-metrics/micrometer/commit/766fa07804c819a86c4cb7269e4cd752e8a99eee).

ben-manes:
I believe the only change that impacts this project is `StatsCounter`. This can be made backwards compatible by retaining the deprecated `recordEviction()` method and removing the `@Override`. It should cross compile.

izeye:
@ben-manes Thanks! I created https://github.com/micrometer-metrics/micrometer/pull/2476 to try to resolve this.

jonatan-ivanov:
Fixed in #2478

PR #2476, opened 2021-02-25
Use latest.release for Caffeine

izeye:
This PR changes to use `latest.release` for Caffeine.

This PR also updates `CaffeineStatsCounter` to support Caffeine 3+ while maintaining Caffeine 2+ support.

Closes gh-2465

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -3,7 +3,7 @@ def VERSIONS = [
         'colt:colt:1.2.0',
         'com.amazonaws:aws-java-sdk-cloudwatch:latest.release',
         'com.fasterxml.jackson.core:jackson-databind:latest.release',
-        'com.github.ben-manes.caffeine:caffeine:2.+',
+        'com.github.ben-manes.caffeine:caffeine:latest.release',
         'com.github.charithe:kafka-junit:latest.release',
         'com.github.tomakehurst:wiremock-jre8-standalone:latest.release',
         'com.google.cloud:google-cloud-monitoring:1.+',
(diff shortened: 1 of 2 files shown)


izeye:
The `docker-tests` job has failed as follows:

```
> Task :micrometer-core:compileJava
/home/circleci/micrometer/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineStatsCounter.java:21: error: cannot access Cache
import com.github.benmanes.caffeine.cache.Cache;
                                         ^
  bad class file: /home/circleci/.gradle/caches/modules-2/files-2.1/com.github.ben-manes.caffeine/caffeine/3.0.0/8b66ca2498c63c7bb8cde37fe25fe52cb503fde1/caffeine-3.0.0-all.jar(com/github/benmanes/caffeine/cache/Cache.class)
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

The `docker-tests` job seems to use JDK 8.

If we merge this, we can't use JDK8 for building Micrometer.

izeye:
Is it okay to raise the minimum version of JDK from 8 to 11 for building Micrometer?

ben-manes:
Thanks @izeye. You can stay fixed on 2.x and be compatible going forward. Consumers can declare 3.x in their build and use Micrometer on JDK11+, which seems like the right balance in your case?

If you want to test more thoroughly then you could have dual 8/11 runs and have the version conditionally set by the jvm detection. That might not be worth the trouble, as this is a pretty minor compatibility change.

izeye:
@ben-manes Thanks for the suggestion! I created https://github.com/micrometer-metrics/micrometer/pull/2478 in case we want to stay to be able to build Micrometer against JDK 8.

jonatan-ivanov:
@izeye 
>Is it okay to raise the minimum version of JDK from 8 to 11 for building Micrometer?
I wish we could do that. :)

@ben-manes Thanks for moving the Java community forward and moving away from a very old Java version. :)

jonatan-ivanov:
I'm closing this in favor of: #2478

PR #1837, opened 2020-02-03
Support running docker in tests on CircleCI build

shakuzen:
We want to use Testcontainers for testing various functionality with real backends running. We need to run such tests on a `machine` type executor on CircleCI. This makes a separate CircleCI job to run those, using a newly added Gradle task `dockerTest` that only runs the tests tagged (with JUnit Jupiter's `@Tag("docker")`) with the `docker` label (see [sample test](https://github.com/micrometer-metrics/micrometer/pull/1837/files/d24244d4f8238de72acc6c3b6d98df891aba4631#diff-4d99805b0f2aec647c408ab8c1a1f0e5)). The default `test` task will not run the tests tagged with `docker`.

Unblocks things like #1429

Also upgrades to CircleCI's version 2.1 configuration which allows some deduplication of the executor, which was in order as different build images were being used in different jobs (by mistake). Also dedups the almost identical steps used in both the `build` and `docker-test` jobs.

PR #2477, opened 2021-02-26
Update machine-executor.image to ubuntu-2004:202101-01

izeye:
This PR updates `machine-executor.image` to `ubuntu-2004:202101-01` as the current image seems to be going to be EOLed soon as follows:

> Note: Ubuntu 16.04 reaches the end of its LTS window at the end of April 2021 and will no longer be supported by Canonical. As a result, the final 16.04 CircleCI machine image release by us will take place in April to include the most recent security patches. We suggest upgrading to the Ubuntu 20.04 image for continued releases past April. 2021.

See https://circleci.com/docs/2.0/configuration-reference/#machine

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
   machine-executor:
     working_directory: ~/micrometer
     machine:
-      image: ubuntu-1604:202007-01
+      image: ubuntu-2004:202101-01
 
 commands:
   gradlew-build:


jonatan-ivanov:
For the record:
https://discuss.circleci.com/t/linux-machine-executor-image-july-update-new-image/36600
https://discuss.circleci.com/t/linux-machine-executor-images-2021-january-q1-update/38976

jonatan-ivanov:
Does anyone know (@shakuzen?) why the circleci config is different for the different versions?
I thought this should go to 1.3 but when I checked it, the config version is different there.

izeye:
@jonatan-ivanov `machine-executor` has been introduced since Micrometer 1.4.0.

See https://github.com/micrometer-metrics/micrometer/pull/1837

shakuzen:
Yes, as @izeye pointed out, the machine executor was only introduced as a means of running docker tests. Ideally we would have been able to do that on a docker executor, but I was never able to get that working and it seems to be a limitation of CircleCI's docker executor setup.

PR #2478, opened 2021-02-26
Support Caffeine 3

izeye:
See gh-2465
See gh-2476


jonatan-ivanov (approved):

jonatan-ivanov:
Thanks! :)

shakuzen:
It would be nice to have a sample that uses Caffeine 3 just so we cover our bases (micrometer-core is built/tested with Caffeine 2.x), but to do this, we can't say the whole project can be built with JDK 8 anymore. I don't know that that is such a bad thing these days, but it is something to consider. I've left a related comment at https://github.com/micrometer-metrics/micrometer/pull/2480#issuecomment-787859958

PR #2479, opened 2021-02-26
Update circle-jdk-executor.image to cimg/openjdk:15.0.1

izeye:
This PR updates `circle-jdk-executor.image` to `cimg/openjdk:15.0.1`.


shakuzen (approved):
The `15.0` tag should be an alias for the latest patch release of JDK 15, which was the original intention. It is better build hygiene to use a specific (patch) version for repeatable builds, though. JDK 15.0.2 is already out, but it looks like the corresponding CircleCI image is not available yet.

PR #2480, opened 2021-02-26
Add a build job on JDK 8

izeye:
This PR adds a build job on JDK 8 as [CONTRIBUTING.md](https://github.com/micrometer-metrics/micrometer/blob/master/CONTRIBUTING.md#building) claims that Micrometer can be built against JDK 8, but there's no explicit check for it except the side effect from the `docker-tests` job.

See https://github.com/micrometer-metrics/micrometer/pull/2476#issuecomment-786301470

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,12 @@ executors:
       GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
     docker:
       - image: cimg/openjdk:16.0.2
+  circle-jdk8-executor:
+    working_directory: ~/micrometer
+    environment:
+      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
+    docker:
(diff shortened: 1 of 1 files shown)


shakuzen:
I agree in general we should test we support what we say (building with JDK 8 in this case) or change what we say we support.
Our continuing to support building Micrometer with JDK 8 has been a convenience to contributors since JDK 8 is still the most common JDK. Also we haven't had a substantial reason to use Java 9+ features in our tests. It may be time to reconsider that decision. It does not seem overly burdensome to ask that contributors have JDK 11 (or even the latest JDK, probably) to build the project. The most important thing for users is that JDK 8 compatibility is maintained, which should be covered by the compiler `release` flag.

Of course this just shifts this issue from running a CI build with JDK 8 to JDK 11 if we bump the minimum JDK to build. I'm not sure the value is high enough to run such a job on every build - though the harm might be considered low also (more throttling from concurrent job limits). I wonder if a daily build to check the project can still be built with older JDK versions is sufficient, assuming we would notice a build failure of that daily job.

shakuzen:
@izeye would you mind rebasing this now that the jdk11 build job is merged?

izeye:
@shakuzen Sure, I've done that in 873e274.

PR #2485, opened 2021-03-04
Fix duplicate +Inf bucket in Prometheus histograms

fstab:
Histograms without an upper bound already have an `+Inf` bucket. The current implementation added an `+Inf` bucket regardless, resulting in two `+Inf` buckets.

This fix makes sure that the `+Inf` bucket is only added if the histogram is configured with a finite upper bound.


shakuzen:
Thanks for reporting this and providing a fix. This seems to have surfaced as a consequence of our switching from `long` to `double` for buckets. Previously, `Long.MAX_VALUE` was used as the last bucket. Now, with `double` we can use `Double.POSITIVE_INFINITY`. Due to that, `Long.MAX_VALUE` and the `+Inf` bucket are present in the 1.3.x branch and I wonder if we should keep MAX_VALUE or not.

shakuzen (approved):
Thanks. I have rebased on and targeted the 1.5.x branch so we can merge the fix forward from there.
FYI: I have added a polish commit to simplify the tests as 3e06732

Issue #2419, opened 2021-01-27
OpenMetrics support for PrometheusMeterRegistry


shakuzen:
The [OpenMetrics specification](https://github.com/OpenObservability/OpenMetrics/blob/master/specification/OpenMetrics.md) has been released (finalized?) and now the [Prometheus Java client uses the OpenMetrics data model since 0.10.0](https://github.com/prometheus/client_java/releases/tag/parent-0.10.0). We are already on that version in the `master` branch now, which required #2418 to compile. This issue is to track other changes that should/could be made in support of OpenMetrics.

- ~[Units](https://github.com/OpenObservability/OpenMetrics/blob/master/specification/OpenMetrics.md#unit)~ (deferred to #2482)
- [x] Support multiple scraping formats (Prom and OpenMetrics text format) #2486 
- [ ] `_created` timestamp (does [Prometheus server even do anything with these currently?](https://github.com/prometheus/prometheus/issues/6541))

Note: Experimental support for OpenMetrics scrape format was added to [Prometheus 2.5.0](https://github.com/prometheus/prometheus/releases/tag/v2.5.0).

shakuzen:
Support for units has been spun off to a separate issue #2482 due to the problems encountered adding it. Units are not strictly required in the OpenMetrics format, so this is not blocking.

shakuzen:
Closing to make the current state of things hopefully less confusing. OpenMetrics format scrape support was already delivered with #2486. There are some additional parts of OpenMetrics that we could potentially add support for in the future, which are linked to this issue.

PR #2486, opened 2021-03-05
OpenMetrics scrape support

shakuzen:
Adds a `contentType` parameter to scrape methods so the scrape can be returned in OpenMetrics format or Prometheus format.

See #2419


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java
```diff
@@ -95,28 +107,46 @@ public String scrape() {
...
      *
+     * @param contentType the Content-Type of the scrape.
      * @param includedNames Sample names to be included. All samples will be included if {@code null}.
      * @return Content that should be included in the response body for an endpoint designated for
      * Prometheus to scrape from.
-     * @since 1.6.0
+     * @since 1.7.0
      */
-    public String scrape(Set<String> includedNames) {
+    public String scrape(String contentType, @Nullable Set<String> includedNames) {
```
shakuzen, line 146: I am considering removing these convenience methods since they haven't been released yet and they haven't been requested by any user so far and we have quite a few `scrape` variants now. Spring Boot won't use them since it makes the scrape endpoint from the underlying `CollectorRegistry` instead of our `PrometheusMeterRegistry`. This leaves the only known consumer our sample, which is essentially the Spring Boot Actuator Prometheus endpoint if it used `PrometheusMeterRegistry` (but as stated, it doesn't).

shakuzen: I suppose it would be nice to offer the methods for other frameworks to achieve something similar to Spring Boot, but if they also use the underlying CollectorRegistry, then these methods still won't be used.

PR #2487, opened 2021-03-06
Bintray Sunset

jonatan-ivanov:
Utilizing the `signing` and the `maven-publish` plugins to remove Bintray from the release process. 
Fixes gh-2455

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,6 +86,14 @@ subprojects {
             }
             repositories {
                 maven {
+                    name = 'Milestone'
+                    url = 'https://repo.spring.io/milestone'
+                    credentials {
+                        username findProperty('MILESTONE_REPO_USER')
+                        password findProperty('MILESTONE_REPO_PASSWORD')
+                    }
+                }
+                maven {
                     name = 'MavenCentral'
                     url = 'https://s01.oss.sonatype.org/service/local/staging/deploy/maven2'
                     credentials {
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,8 @@ allprojects {
 }
 
 subprojects {
+    apply plugin: 'signing'
+
     if (project.name != 'micrometer-bom') {
         apply plugin: 'java'
         apply plugin: 'checkstyle'
@@ -74,6 +76,29 @@ subprojects {
         plugins.withId('maven-publish') {
             sourceCompatibility = JavaVersion.VERSION_1_8
             targetCompatibility = JavaVersion.VERSION_1_8
+        }
+    }
+
+    plugins.withId('maven-publish') {
+        publishing {
+            publications {
+                nebula(MavenPublication)
+            }
+            repositories {
+                maven {
+                    name = 'MavenCentral'
+                    url = 'https://s01.oss.sonatype.org/service/local/staging/deploy/maven2'
+                    credentials {
+                        username findProperty('MAVEN_CENTRAL_USER')
+                        password findProperty('MAVEN_CENTRAL_PASSWORD')
+                    }
+                }
+            }
+        }
+
+        signing {
+            useInMemoryPgpKeys(findProperty('SIGNING_KEY'), findProperty('SIGNING_PASSWORD'))
+            sign publishing.publications.nebula
         }
     }
 
--- a/gradle/deploy.sh
+++ b/gradle/deploy.sh
@@ -15,12 +15,10 @@ elif [ $CIRCLE_TAG ]; then
   openssl aes-256-cbc -d -in gradle.properties.enc -out gradle.properties -k "$KEY" -md sha256
   case "$CIRCLE_TAG" in
   *-rc\.*)
-    ./gradlew -Prelease.disableGitChecks=true -Prelease.useLastTag=true candidate bintrayUpload $SWITCHES -x release -x bintrayPublish
-    ./gradlew -Prelease.disableGitChecks=true -Prelease.useLastTag=true :micrometer-core:bintrayPublish $SWITCHES -x bintrayUpload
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -77,6 +79,37 @@ subprojects {
...
 
+    plugins.withId('maven-publish') {
+        publishing {
+            publications {
+                nebula(MavenPublication)
+            }
+            repositories {
+                maven {
+                    name = 'Milestone'
+                    url = 'https://repo.spring.io/milestone'
```
shakuzen, line 90: I don't know the difference between using the maven-publish plugin directly versus using the artifactory plugin to publish to this repository. The snapshot publish uses the artifactory plugin as far as I understand. Could we check and compare a snapshot vs candidate release with these changes?

gradle/deploy.sh
```diff
@@ -15,12 +15,10 @@ elif [ $CIRCLE_TAG ]; then
   openssl aes-256-cbc -d -in gradle.properties.enc -out gradle.properties -k "$KEY" -md sha256
   case "$CIRCLE_TAG" in
   *-rc\.*)
-    ./gradlew -Prelease.disableGitChecks=true -Prelease.useLastTag=true candidate bintrayUpload $SWITCHES -x release -x bintrayPublish
-    ./gradlew -Prelease.disableGitChecks=true -Prelease.useLastTag=true :micrometer-core:bintrayPublish $SWITCHES -x bintrayUpload
+    ./gradlew -Prelease.disableGitChecks=true -Prelease.useLastTag=true candidate publishAllPublicationsToMilestoneRepository $SWITCHES -x release -x artifactoryPublish -x bintrayUpload -x bintrayPublish
```
shakuzen, line 18: Nit: We could make `candidate` depend on the publish task in the build so the command here is simpler. I know it sounds a bit superfluous with all the skipped tasks here... We could follow-up getting rid of those with an update to the spring-release-plugin we're using.

shakuzen, line 18: publishAll is probably fine, but I wonder if it wouldn't be better to be more specific just in case there is more than one publication registered. Would we want to publish all of them? Can't we publish the nebula publication specifically since that's the one we expect to be there?

shakuzen (approved):
Looks good.
I think releasing the staged repository is still required for final releases and not covered yet. Perhaps we can consider the gradle nexus publish plugin for that. https://github.com/gradle-nexus/publish-plugin/

build.gradle
```diff
@@ -77,6 +79,37 @@ subprojects {
...
+                        username findProperty('MAVEN_CENTRAL_USER')
+                        password findProperty('MAVEN_CENTRAL_PASSWORD')
+                    }
+                }
+            }
+        }
+
+        signing {
+            useInMemoryPgpKeys(findProperty('SIGNING_KEY'), findProperty('SIGNING_PASSWORD'))
+            sign publishing.publications.nebula
```
shakuzen, line 109: Will our snapshot releases be signed too? I wasn't sure given the setup.

shakuzen: I tried to do a `pTML` (publishToMavenLocal) which is what I do to test changes in snapshots locally, but it failed because the signing properties aren't configured. We want contributors to be able to publish local snapshots without needing to configure signing keys. That said, we do want to ensure signing is done for remote publishing. I'm not sure the best way to go about that. I added
```
if (hasProperty('SIGNING_KEY') && hasProperty('SIGNING_PASSWORD')) {
```
but that makes it possible for remote publishing to be attempted without signing artifacts, which I think we want to avoid - it will fail Maven Central's requirements, for example, and is just good practice if we easily can.

build.gradle
jonatan-ivanov: Since it is just a maven repo eventually, both will work, the artifactory plugin was a little more convenient since it was already set-up. Though, using the same process for all of our repos would make more sense to me too, also it will enable us to remove the artifactory plugin. I added the snapshot repo, the artifactory plugin is not used from now on and everything is published in the same way.

gradle/deploy.sh
jonatan-ivanov: I was thinking on it too but eventually I did not do it since this way it is easier to test things locally and gives more information about what is happening. What do you think about cleaning this-up with the release plugin cleanup (that should also eliminate the excluded tasks)?

build.gradle
jonatan-ivanov: Oops, I messed this up. This signs snapshots too as well as local publications.
I fixed it this way:
```groovy
signing {
            required = hasProperty('SIGNING_KEY')
}
```

This tells the signing plugin that signature is required when the key is present otherwise the task will be skipped so if you have the key it will sign local builds, if you don't it won't. I did not include the password since the gpg key can be passwordless.

jonatan-ivanov:
@shakuzen  Changes since the last review:
- `closeAndReleaseMavenCentralStagingRepository` task (https://github.com/gradle-nexus/publish-plugin)
- snapshot is published the same way as everything else (maven-publish instead of artifactory)
- `publishAllPublicationToMavenCentralRepository` -> `publishNebulaPublicationToMavenCentralRepository`
- optional signature: `publishToMavenLocal` works without a key
- `deploy.sh` matches on milestone now instead of rc (I tested it, it works)

Also, we can decide if we want to release manually or automatically: `closeMavenCentralStagingRepository` vs. `closeAndReleaseMavenCentralStagingRepository`. Right now, the latter is used.

build.gradle
shakuzen: That fixes the issue of contributors being able to publish a snapshot to their local maven repository. We should probably still think about checking that signing is configured if we're publishing a release to a remote repository.

gradle/deploy.sh
shakuzen: Yes, I think it's fine to defer to when we change the release plugin. I think all of this is vetting the process before we bake it into a released version of the plugin.

shakuzen (approved):

PR #2501, opened 2021-03-13
Remove PrometheusFilteredSample

izeye:
`PrometheusFilteredSample` seems to be redundant now as the filtered scrape for Prometheus is supported since Spring Boot 2.4.0.

See https://github.com/spring-projects/spring-boot/pull/21545


jonatan-ivanov:
Thanks! :)

PR #2502, opened 2021-03-14
Remove maven-central-sync job

izeye:
It seems to be missed in 16805ce79cf5d8a9bb6b2bb8a7d1cacdb5ecdf14.

We will no longer sync from Bintray to Maven central, so remove all related scripts/jobs.


shakuzen (changes requested):
Good catch. Actually, the whole `maven-central-sync` job and references to it can be removed. It uses the `mavenCentralSync` task, which is for Bintray.

shakuzen (approved):
I've updated the pull request

PR #2505, opened 2021-03-15
Support Datadog UDP origin detection

### LINKED ISSUE #2417: Support Datadog UDP origin detection
Datadog origin detection allows to detect where the container metrics come from, and tag metrics automatically [1]

Datadog clients use `DD_ENTITY_ID` environment variable to append `dd.internal.entity_id` tag
to the metric (Example: [2]). The agent uses this tag to detect the origin (e.g. a Kubernetes pod) and append its tags (various k8s tags like pod name, labels, annotations).

However, `io.micrometer.statsd` implementation does not support this.
Since it already has different flavors, perhaps this datadog-specific functionality could be implemented in the `DATADOG` flavor. Then the datadog UDP origin detection would work out of the box.

What do you think?

[1] https://docs.datadoghq.com/developers/dogstatsd/?tab=kubernetes#origin-detection-over-udp
[2] Example implementation in java-dogstatsd-client: https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/NonBlockingStatsDClient.java#L1759-L1770

s12v:
Resolves #2417 

Adds datadog internal entity id (from `DD_ENTITY_ID` environment variable) as a tag, so that UDP origin detection can work.

Similar to the Datadog official client implementation: https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/NonBlockingStatsDClient.java#L1759-L1770

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/DatadogStatsdLineBuilder.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/DatadogStatsdLineBuilder.java
@@ -27,6 +27,9 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.stream.Collectors;
 
 public class DatadogStatsdLineBuilder extends FlavorStatsdLineBuilder {
+
+    private static final String ENTITY_ID_TAG_NAME = "dd.internal.entity_id";
+
     private final Object conventionTagsLock = new Object();
     @SuppressWarnings({"NullableProblems", "unused"})
     private volatile NamingConvention namingConvention;
@@ -37,6 +40,8 @@ public class DatadogStatsdLineBuilder extends FlavorStatsdLineBuilder {
     @SuppressWarnings("NullableProblems")
     private volatile String tagsNoStat;
     private final ConcurrentMap<Statistic, String> tags = new ConcurrentHashMap<>();
+    @Nullable
+    private String ddEntityId;
 
     public DatadogStatsdLineBuilder(Meter.Id id, MeterRegistry.Config config) {
         super(id, config);
@@ -48,6 +53,11 @@ public class DatadogStatsdLineBuilder extends FlavorStatsdLineBuilder {
         return name + amount + "|" + type + tagsByStatistic(stat);
     }
 
+    // VisibleForTesting
+    void setDdEntityId(@Nullable String ddEntityId) {
+        this.ddEntityId = ddEntityId;
+    }
+
     private void updateIfNamingConventionChanged() {
         NamingConvention next = config.namingConvention();
         if (this.namingConvention != next) {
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@s12v Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2505)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2505) to manually synchronize the status of this Pull Request.

shakuzen:
@s12v thanks for the pull request. Did you have a chance to test this out? I just tried it, and I wasn't able to get the origin detection tags on the metrics. It's possible I'm missing some piece of configuration, but my metrics are getting to Datadog, just without any added origin detection tags. I added the `DD_ENTITY_ID` environment variable to my application [as documented](https://docs.datadoghq.com/developers/dogstatsd/?tab=kubernetes#origin-detection-over-udp), and I set `useHostPID` and `originDetection` to `true` on the agent configuration.

s12v:
hey @shakuzen, we use UDP origin detection by manually setting `dd.internal.entity_id: ${DD_ENTITY_ID:}` tag in Spring configuration, I can confirm it works (in general). However, I did not run this PR in a real environment, I will try to set it up.
Did you run it in Kubernetes?

shakuzen:
> Did you run it in Kubernetes?

Yes, I ran it in a minikube environment on my machine. Perhaps I just don't know what I should be looking for in Datadog to confirm it is doing what it is supposed to. What should I be checking for and expecting to be different when this is working? I expected there to be additional identifying tags on the metrics sent from the application.

s12v:
> What should I be checking for and expecting to be different when this is working? I expected there to be additional identifying tags on the metrics sent from the application.

yes, the metrics should have additional tags. You can set `DD_DOGSTATSD_TAG_CARDINALITY` to `high` in order to have more tags . Also, can use `DD_KUBERNETES_POD_ANNOTATIONS_AS_TAGS` and `DD_KUBERNETES_POD_LABELS_AS_TAGS` options (we use them to append tags like service name, stage, team, etc.)

See https://docs.datadoghq.com/getting_started/tagging/assigning_tags/?tab=containerizedenvironments#environment-variables, https://docs.datadoghq.com/agent/kubernetes/tag/?tab=containerizedagent#pod-annotations-as-tags, and https://docs.datadoghq.com/agent/kubernetes/tag/?tab=containerizedagent#pod-annotations-as-tags

shakuzen:
> yes, the metrics should have additional tags. You can set `DD_DOGSTATSD_TAG_CARDINALITY` to `high` in order to have more tags .

I wasn't seeing any additional tags, which is why I could not confirm it was working with these changes.

s12v:
@shakuzen, thanks for looking into this. I was able to test it locally in minikube:

Screenshot 2021-04-08 at 20 09 45

it did not work out of the box. Turns out, by default datadog-agent has some issues with minikube (https://github.com/DataDog/datadog-agent/issues/4752#issuecomment-582599217)

I temporary uploaded my setup to https://github.com/s12v/sbpg (sent an invitation so you could check)

shakuzen:
Thanks for the investigation @s12v. I was just using minikube as an easy way to try it out with Kubernetes locally. It's fine if the feature doesn't work on minikube as long as it works on Kubernetes setups users will actually be deploying apps to and monitoring with Datadog. Though it is good to know of this limitation with minikube in case other users run into it in the future. I will try it out with your repository and perhaps I'll try it out on another non-minikube Kubernetes. Thanks again. Hopefully we should be able to get this merged soon.

shakuzen (approved):
I was able to get this working, thanks to your investigation. Thank you for the feature. We should document this feature being available in the [Micrometer docs](https://github.com/micrometer-metrics/micrometer-docs/).

PR #2517, opened 2021-03-18
Call awaitTermination when stopping a PushMeterRegistry

tburch:
Currently any in-flight publishes from a `PushMeterRegistry ` are discarded when the `PushMeterRegistry` is stopped. 

This PR adds a call to `ScheduledExecutorService#awaitTermination()` when stopping the `PushMeterRegistry ` to let those calls finish (up to the timeout). The timeout used in the `awaitTermination()` call is configurable.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -82,7 +82,14 @@ public abstract class PushMeterRegistry extends MeterRegistry {
     public void stop() {
         if (scheduledExecutorService != null) {
             scheduledExecutorService.shutdown();
-            scheduledExecutorService = null;
+            try {
+                scheduledExecutorService.awaitTermination(config.shutdownTimeout().toMillis(), TimeUnit.MILLISECONDS);
+            } catch (InterruptedException ie) {
+                scheduledExecutorService.shutdownNow();
+                Thread.currentThread().interrupt();
+            } finally {
+                scheduledExecutorService = null;
+            }
         }
     }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushRegistryConfig.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushRegistryConfig.java
@@ -91,6 +91,12 @@ public interface PushRegistryConfig extends MeterRegistryConfig {
         return getInteger(this, "batchSize").orElse(10000);
     }
 
+    /**
+     * @return The timeout to use when calling awaitTermination while shutting down the ScheduledExecutorService. The
+     * default is 5 seconds.
+     */
+    default Duration shutdownTimeout() { return getDuration(this, "shutdownTimeout").orElse(Duration.ofSeconds(5)); }
+
     @Override
     default Validated<?> validate() {
         return validate(this);
@@ -109,7 +115,8 @@ public interface PushRegistryConfig extends MeterRegistryConfig {
                 check("connectTimeout", PushRegistryConfig::connectTimeout),
                 check("readTimeout", PushRegistryConfig::readTimeout),
                 check("batchSize", PushRegistryConfig::batchSize),
-                check("numThreads", PushRegistryConfig::numThreads)
+                check("numThreads", PushRegistryConfig::numThreads),
+                check("shutdownTimeout", PushRegistryConfig::shutdownTimeout)
         );
     }
 }
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/push/PushRegistryConfigTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/push/PushRegistryConfigTest.java
@@ -49,6 +49,7 @@ class PushRegistryConfigTest {
         props.put("push.readTimeout", "1w");
         props.put("push.batchSize", "Z");
         props.put("push.step", "up");
+        props.put("push.shutdownTimeout", "down");
 
         // overall not valid
         assertThat(config.validate().isValid()).isFalse();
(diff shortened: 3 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -82,7 +82,14 @@ public void start(ThreadFactory threadFactory) {
     public void stop() {
         if (scheduledExecutorService != null) {
             scheduledExecutorService.shutdown();
-            scheduledExecutorService = null;
+            try {
+                scheduledExecutorService.awaitTermination(config.shutdownTimeout().toMillis(), TimeUnit.MILLISECONDS);
+            } catch (InterruptedException ie) {
+                scheduledExecutorService.shutdownNow();
+                Thread.currentThread().interrupt();
```
jonatan-ivanov, line 89: I'm thinking if doing this is a good idea rather than just log a `WARN` event and carry on (assuming this happens at JVM shutdown).
What do you think?

tburch: @jonatan-ivanov The backstory on how I ran into this was that I have an AWS Lambda container running on Micronaut with Micrometer for metrics. Since the container is paused between Lambda invocations, you can't get metrics pushed at a reliable interval because the push interval goes from 60 seconds of wall clock time to 60 seconds of invocation time, which can be awhile for infrequently invoked lambdas.
In order to push the metrics, I'm actually invoking [`PushMeterRegistry#publish`](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java#L45 )after every Lambda invocation (via reflection since it's not public - and maybe that's another PR). 
My other idea was to start/stop the `PushMeterRegistry` around every Lambda invocation, but `PushMeterRegistry #stop()` didn't block because of the missing `awaitTermination()` call, this didn't behave as I expected, thus this PR.

tburch: Also, calling `shutdown` and then `awaitTermination` is the proper way to nicely shutdown an `ExecutorService`...

jonatan-ivanov: I was only talking about these two lines:
```java
scheduledExecutorService.shutdownNow();
Thread.currentThread().interrupt();
```
As far as I remember, calling just `shutdown` is a proper way to nicely shutdown an `ExecutorService` **if** you don't want to block on it or you don't care about the results.

But if you do, then this is the proper way of nicely shutting down an `ExecutorService`, please see the [javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html):

```java
 void shutdownAndAwaitTermination(ExecutorService pool) {
   pool.shutdown(); // Disable new tasks from being submitted
   try {
     // Wait a while for existing tasks to terminate
     if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
       pool.shutdownNow(); // Cancel currently executing tasks
       // Wait a while for tasks to respond to being cancelled
       if (!pool.awaitTermination(60, TimeUnit.SECONDS))
           System.err.println("Pool did not terminate");
     }
   } catch (InterruptedException ie) {
     // (Re-)Cancel if current thread also interrupted
     pool.shutdownNow();
     // Preserve interrupt status
     Thread.currentThread().interrupt();
   }
 }
```

What I meant to say with my original comment is instead of calling `shutdownNow` and `interrupt` the `Thread` again, wouldn't it be simpler just to log a `WARN` event that says `"Waiting for the threadpool to shut down timed out."`? Assuming this happens at JVM shutdown, this may (or may not) give your tasks a little bit of an extra time to finish their job instead of cancelling them immediately.

jonatan-ivanov: @shakuzen Do you happen to know if `stop` should be fire-and-forget or can we block on it for a while?

tburch: @jonatan-ivanov Sorry, I didn't fully understand your prior comment, but do now.

shakuzen:
Thanks for making the pull request and providing the context on your use case. We have some other issues previously raised around metrics in short-lived processes. I've added an item to our Micrometer 1.8 planning to take a look at this overall and make sure we're providing a good solution. While we could make the kind of changes proposed here, they make sense for your use case because you are calling publish via reflection. That signals we need to take a step back and ensure we're properly supporting the use case to begin with.

shakuzen:
On further thought, if you're calling `publish` yourself, the `scheduledExecutorService` isn't involved, so I'm not sure how this change would even help your use case. Only scheduled publishes would be affected, but you're saying you're forcing a publish outside of the scheduled ones.

tburch:
@shakuzen Sorry, I guess that wasn't clear. I was hoping to call start/stop instead of publish via reflection, but because stop didn't block, I didn't have a way to guarantee that the request got out. The lambda doesn't always close between requests, so I couldn't count on close either.

Micronaut will properly call close once [this issue](https://github.com/micronaut-projects/micronaut-micrometer/pull/219) is released. However, I still think it's important to block on close via `awaitTermination` so that the final publish gets executed.

shakuzen:
When I said the final `publish` will happen on the thread calling the `close` method, I meant that it will be blocking. Calling `awaitTermination` is for the scheduler. The final publish on close is not done via the scheduler so a change to the scheduler is unrelated. You can check the code in the `close` method.

Stopping and starting the registry after each request is not expected usage, and so long as the final publish happens, you should have the metrics through the lambda shutting down. You shouldn't need to control the publishing manually. Or am I missing something?

tburch:
Understood. My concern with `close` in AWS Lambda is that I'd observed that I didn't get metrics published when the container was terminated, which could be a Micronaut bug or limitation that AWS Lambda imposes on the lambda (like the container is never unfrozen).

shakuzen:
Since we are publishing on close outside of the scheduler (and blocking), I don't see how the proposed changes do what they seem to imply. We'll still take a look at what we can do to improve things for publishing metrics in short-lived processes, but unless I'm missing something, this doesn't look like a change we should make. If you're still having some issue not covered by an already open issue, please let us know and we can look into it more.

PR #2520, opened 2021-03-22
Use cimg/openjdk:16.0.1 for CircleCI

izeye:
This PR changes to use `cimg/openjdk:16.0.0` for CircleCI.


izeye:
Its build has failed as follows:

```
Execution failed for task ':micrometer-core:compileJava'.
> java.lang.IllegalAccessError: class org.gradle.internal.compiler.java.ClassNameCollector (in unnamed module @0xe7b777f) cannot access class com.sun.tools.javac.code.Symbol$TypeSymbol (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.code to unnamed module @0xe7b777f
```

Gradle 6.8.3 doesn't seem to work on JDK 16.

shakuzen:
> Gradle 6.8.3 doesn't seem to work on JDK 16.

Not without some workarounds, unfortunately. See https://github.com/gradle/gradle/issues/13481. I think Gradle 7.0 should be out "soon", so perhaps we should just wait for that. It's a shame support without workarounds hasn't been released in a 6.x version of Gradle.

shakuzen:
Now that the Gradle 7 upgrade is in, this should be unblocked. Would you rebase and target it at the `1.5.x` branch, please?

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ executors:
     environment:
       GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
     docker:
-      - image: cimg/openjdk:15.0.1
+      - image: cimg/openjdk:16.0.0
```
shakuzen, line 9: OpenJDK 16.0.1 is available already but I don't see a corresponding version for cimg/openjdk yet, unfortunately. We can follow-up with a patch version bump when that is available.

shakuzen:
Backing this out of the 1.5.14 release since we're back to being blocked on upgrading to Gradle 7. While Gradle 6.9 can compile with JDK 16, it cannot run with it, which would require us to have two JDK versions in our build environment. We'll unfortunately have to come back to this when we can manage to upgrade to Gradle 7.

jonatan-ivanov:
@izeye Could you please try this with the latest 1.5.x?
See: https://github.com/micrometer-metrics/micrometer/pull/2693

(We also need to `@Disable` tests for the `ExecutorService` instrumentation if we want this now)

.circleci/config.yml
shakuzen: It's available now. I'm updating the pull request

micrometer-core/gradle.lockfile
```diff
@@ -104,8 +104,8 @@ org.apache.kafka:kafka-clients:2.5.0=compileClasspath,runtimeClasspath,testCompi
 org.apache.kafka:kafka-streams:2.5.0=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
 org.apache.logging.log4j:log4j-api:2.13.2=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
 org.apache.logging.log4j:log4j-core:2.13.2=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
-org.apache.tomcat.embed:tomcat-embed-core:8.5.54=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
-org.apache.tomcat:tomcat-annotations-api:8.5.54=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+org.apache.tomcat.embed:tomcat-embed-core:8.5.69=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
```
shakuzen, line 107: Our Tomcat tests fail on Java 16 without this upgrade due to an InaccessibleObjectException.
https://tomcat.apache.org/tomcat-8.5-doc/changelog.html#Tomcat_8.5.62_(markt)

PR #2522, opened 2021-03-23
Upgrade to JMH Gradle plugin 0.6.2

izeye:
This PR upgrades to JMH Gradle plugin 0.6.2.

### DIFF
--- a/benchmarks/benchmarks-core/build.gradle
+++ b/benchmarks/benchmarks-core/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id "me.champeau.gradle.jmh" version "0.5.2"
+    id "me.champeau.jmh" version "0.6.2"
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


shakuzen (approved):
Thanks, as always. Can we easily apply this to the 1.3.x branch? If so, I'd like to keep this configuration up-to-date and consistent across maintenance branches.

izeye:
@shakuzen JMH Gradle plugin 0.6.2 seems to [require Gradle 6.8 or later](https://github.com/melix/jmh-gradle-plugin#what-plugin-version-to-use), so it doesn't seem to be able to apply it to the 1.3.x branch. Is it okay with applying it to the 1.5.x branch and later?

shakuzen:
> Is it okay with applying it to the 1.5.x branch and later?

Yes, that's fine. I've updated the milestone.

izeye:
@shakuzen Thanks for the quick feedback! I changed the base branch to the 1.5.x.

For the record, I updated dependency locks as follows:

```
./gradlew clean check downloadDependencies --update-locks org.openjdk.jmh:jmh-core,org.openjdk.jmh:jmh-generator-annprocess
```

Issue #2525, opened 2021-03-24
CVE-2021-21290 in shaded netty in micrometer-statsd-registry


atennapel:
There is a new CVE in the shaded netty in micrometer-statsd-registry: https://nvd.nist.gov/vuln/detail/CVE-2021-21290 .
Can netty be upgraded?

See detailed information below:

Dependency | Vulnerability IDs | Package | Highest Severity | CVE Count | Confidence | Evidence Count
-- | -- | -- | -- | -- | -- | --
micrometer-registry-statsd-1.6.5.jar (shaded: io.netty:netty-codec-http2:4.1.56.Final) | cpe:2.3:a:netty:netty:4.1.56:*:*:*:*:*:*:* | pkg:maven/io.netty/netty-codec-http2@4.1.56.Final | MEDIUM | 1 | Highest | 9
micrometer-registry-statsd-1.6.5.jar (shaded: io.netty:netty-handler-proxy:4.1.56.Final) | cpe:2.3:a:netty:netty:4.1.56:*:*:*:*:*:*:* | pkg:maven/io.netty/netty-handler-proxy@4.1.56.Final | MEDIUM | 1 | Highest | 9
micrometer-registry-statsd-1.6.5.jar (shaded: io.netty:netty-codec-http:4.1.56.Final) | cpe:2.3:a:netty:netty:4.1.56:*:*:*:*:*:*:* | pkg:maven/io.netty/netty-codec-http@4.1.56.Final | MEDIUM | 1 | Highest | 9
micrometer-registry-statsd-1.6.5.jar (shaded: io.netty:netty-handler:4.1.56.Final) | cpe:2.3:a:netty:netty:4.1.56:*:*:*:*:*:*:* | pkg:maven/io.netty/netty-handler@4.1.56.Final | MEDIUM | 1 | Highest | 9
micrometer-registry-statsd-1.6.5.jar (shaded: io.netty:netty-codec-socks:4.1.56.Final) | cpe:2.3:a:netty:netty:4.1.56:*:*:*:*:*:*:* | pkg:maven/io.netty/netty-codec-socks@4.1.56.Final | MEDIUM | 1 | Highest | 9
[...]

PR #2531, opened 2021-03-24
Exclude netty http modules

shakuzen:
We do not use HTTP in the StatsD registry implementation, so we would avoid some of the pain of false positive CVE detection on our shaded netty dependencies by excluding HTTP-related modules we don't use. This will also reduce the size of the statsd registry artifact.

See #2525

### DIFF
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -6,7 +6,10 @@ dependencies {
     api project(':micrometer-core')
 
     implementation 'io.projectreactor:reactor-core'
-    implementation 'io.projectreactor.netty:reactor-netty'
+    implementation('io.projectreactor.netty:reactor-netty') {
+        exclude module: 'netty-codec-http2'
+        exclude module: 'netty-codec-http'
+    }
 
     testImplementation project(':micrometer-test')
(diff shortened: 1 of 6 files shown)


shakuzen:
@violetagg would you mind letting me know if this seems reasonable? This branch is on reactor-netty 0.9.x and we don't use the HttpClient or any servers; only UdpClient/TcpClient. On our main branch, we plan to use reactor-netty-core when we upgrade to 1.0.x to avoid the HTTP stuff we don't use.

violetagg:
> @violetagg would you mind letting me know if this seems reasonable? This branch is on reactor-netty 0.9.x and we don't use the HttpClient or any servers; only UdpClient/TcpClient. On our main branch, we plan to use reactor-netty-core when we upgrade to 1.0.x to avoid the HTTP stuff we don't use.

That's exactly what was the reason for separating Reactor Netty on modules.

stevefranchak:
Did #2588 undo excluding netty http codec modules, and, if so, was that intentional?

I updated micrometer-registry-statsd from 1.7.6 to 1.8.1 and noticed that io/micrometer/shaded/io/netty/handler/codec/http is being included in the JAR file. Previously, I was able to write off CVEs like CVE-2021-43797 as a false positive because the http modules weren't being included, but that's not possible for me to justify anymore.

Issue #780, opened 2018-08-09
@Timed on a controller being processed twice when using TimedAspect


izeye:
When I was creating [a working sample](https://github.com/izeye/sample-micrometer-spring-boot/tree/timed-annotation) for [an SO question](https://stackoverflow.com/questions/51749391/timed-annotation-in-spring-metrics), I noticed `@Timed` on a controller being processed twice when using `TimedAspect`. One in `TimedAspect` and the other in `WebMvcMetricsFilter`.

See the above working sample for reproducing the issue.

mweirauch:
It's the same when the Jersey instrumentation and the `TimedAspect` are used together.

Additionally, I had `PrometheusMeterRegistry` choke on different tag cardinality on identically named metrics in some constellation which I can't remember offhand.

michaelmcfadyen:
I don't believe this is a bug. The spring boot documentation states that if you want to use `@Timed` you should disable the metrics filter

> Auto-configuration enables the instrumentation of requests handled by Spring MVC. When management.metrics.web.server.auto-time-requests is true, this instrumentation occurs for all requests. Alternatively, when set to false, you can enable instrumentation by adding @Timed to a request-handling method

Source: https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html#production-ready-metrics-spring-mvc

izeye:
@michaelmcfadyen I'm not sure how you came to the conclusion but IIUC, the part of the documentation you quoted is just saying that how `@Timed` works with `WebMvcMetricsFilter`. So if you simply set `management.metrics.web.server.auto-time-requests=false`, you just lose `http.server.requests` metrics but `@Timed` will be processed twice. Try the above "a working sample" I linked with the property you mentioned.

jkschneider:
Trying to think of how we could prevent it without having `TimedAspect` and `WebMvcMetricsFilter` have specific knowledge of each other. Seems most likely to me that we will do away with the `TimedAspect` altogether in Spring Boot 2.1 in favor of a non-AOP approach.

kwtan:
We extends TimedAspect to workaround this issue
`
    @Override
    public Object timedMethod(ProceedingJoinPoint pjp) throws Throwable {
        Class targetClass = pjp.getTarget().getClass();
        if (targetClass.isAnnotationPresent(RestController.class) || targetClass.isAnnotationPresent(Controller.class)) {
            return pjp.proceed();
        }

[...]

PR #2551, opened 2021-04-08
Introduce a skip predicate to the TimedAspect and CountedAspect

jonatan-ivanov:
So that users can skip Timer creation based on the JoinPoint, fixes gh-780


shakuzen (commented):
We should add more explanation on why someone would want to use the skip predicate so it can be more organically discovered. Other than that, seems like a good addition.

shakuzen (approved):
Looks good to me.

PR #2553, opened 2021-04-08
Move Dynatrace metrics exporter for API v1 to a separate package

pirgeo:
This is a PR in preparation for the Dynatrace API v2 exporter (#2295). It moves the current v1 implementation into the v1 package while keeping the interface unchanged, making this a compatible change.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -79,17 +79,13 @@ public interface DynatraceConfig extends StepRegistryConfig {
 
     @Override
     default Validated<?> validate() {
-        // Currently only v1 is implemented. This check will be extended once more versions are added.
-        if (apiVersion() == DynatraceApiVersion.V1) {
-            return checkAll(this,
-                    c -> StepRegistryConfig.validate(c),
-                    checkRequired("apiToken", DynatraceConfig::apiToken),
-                    checkRequired("uri", DynatraceConfig::uri),
-                    checkRequired("deviceId", DynatraceConfig::deviceId),
-                    check("technologyType", DynatraceConfig::technologyType).andThen(Validated::nonBlank)
-            );
-        }
-
-        return Validated.invalid("apiVersion", apiVersion(), "API version could not be validated.", InvalidReason.MALFORMED);
+        return checkAll(this,
+                c -> StepRegistryConfig.validate(c),
+                checkRequired("apiToken", DynatraceConfig::apiToken),
+                checkRequired("uri", DynatraceConfig::uri),
+                checkRequired("deviceId", DynatraceConfig::deviceId),
+                check("technologyType", DynatraceConfig::technologyType).andThen(Validated::nonBlank),
+                check("apiVersion", DynatraceConfig::apiVersion)
+        );
     }
 }
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -15,12 +15,11 @@
  */
 package io.micrometer.dynatrace;
 
+import io.micrometer.core.instrument.config.validate.InvalidReason;
 import io.micrometer.core.instrument.config.validate.Validated;
 import io.micrometer.core.instrument.step.StepRegistryConfig;
 import io.micrometer.core.instrument.util.StringUtils;
 import io.micrometer.core.lang.Nullable;
-
-import java.util.function.Function;
 
 import static io.micrometer.core.instrument.config.MeterRegistryConfigValidator.*;
 import static io.micrometer.core.instrument.config.validate.PropertyValidator.*;
@@ -70,21 +69,19 @@ public interface DynatraceConfig extends StepRegistryConfig {
     /**
      * Return the version of the target Dynatrace API.
      *
-     * @return a {@link String} containing the version of the targeted Dynatrace API.
+     * @return a {@link DynatraceApiVersion} containing the version of the targeted Dynatrace API.
      */
-    default String apiVersion() {
-        // if not specified, defaults to v1 for backwards compatibility.
-        return getString(this, "apiVersion")
-                .map(val -> StringUtils.isEmpty(val) ? "v1" : val)
-                .get();
+    default DynatraceApiVersion apiVersion() {
+        // if not specified or invalid, defaults to v1 for backwards compatibility.
+        return getEnum(this, DynatraceApiVersion.class, "apiVersion")
+                .orElse(DynatraceApiVersion.v1);
     }
 
     @Override
     default Validated<?> validate() {
         // Currently only v1 is implemented. This check will be extended once more versions are added.
-        Function<DynatraceConfig, Validated<String>> versionFunc = checkRequired("apiVersion", DynatraceConfig::apiVersion);
-        String versionStr = versionFunc.apply(this).get();
-        if (versionStr.equals("v1")) {
+
+        if (apiVersion() == DynatraceApiVersion.v1) {
             return checkAll(this,
                     c -> StepRegistryConfig.validate(c),
                     checkRequired("apiToken", DynatraceConfig::apiToken),
@@ -94,6 +91,6 @@ public interface DynatraceConfig extends StepRegistryConfig {
             );
         }
 
-        throw new IllegalArgumentException(String.format("The provided API version is not valid: %s.", versionStr));
+        return Validated.invalid("apiVersion", apiVersion(), "API version could not be validated.", InvalidReason.MALFORMED);
     }
 }
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -78,6 +78,11 @@ public interface DynatraceConfig extends StepRegistryConfig {
 
     @Override
     default Validated<?> validate() {
+        Validated<DynatraceApiVersion> apiVersionValidation = checkRequired("apiVersion", DynatraceConfig::apiVersion).apply(this);
+        if (apiVersionValidation.isInvalid()) {
+            return apiVersionValidation;
+        }
+
         return validateV1();
     }
 
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -71,7 +71,7 @@ public interface DynatraceConfig extends StepRegistryConfig {
      * @return a {@link DynatraceApiVersion} containing the version of the targeted Dynatrace API.
      */
     default DynatraceApiVersion apiVersion() {
-        // if not specified or invalid, defaults to v1 for backwards compatibility.
+        // if not specified, defaults to v1 for backwards compatibility.
         return getEnum(this, DynatraceApiVersion.class, "apiVersion")
                 .orElse(DynatraceApiVersion.V1);
     }
@@ -89,7 +89,6 @@ public interface DynatraceConfig extends StepRegistryConfig {
     default Validated<?> validateV1() {
         return checkAll(this,
                 c -> StepRegistryConfig.validate(c),
-                checkRequired("apiVersion", DynatraceConfig::apiVersion),
                 checkRequired("apiToken", DynatraceConfig::apiToken),
                 checkRequired("uri", DynatraceConfig::uri),
                 checkRequired("deviceId", DynatraceConfig::deviceId),
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -74,13 +74,13 @@ public interface DynatraceConfig extends StepRegistryConfig {
     default DynatraceApiVersion apiVersion() {
         // if not specified or invalid, defaults to v1 for backwards compatibility.
         return getEnum(this, DynatraceApiVersion.class, "apiVersion")
-                .orElse(DynatraceApiVersion.v1);
+                .orElse(DynatraceApiVersion.V1);
     }
 
     @Override
     default Validated<?> validate() {
         // Currently only v1 is implemented. This check will be extended once more versions are added.
-        if (apiVersion() == DynatraceApiVersion.v1) {
+        if (apiVersion() == DynatraceApiVersion.V1) {
             return checkAll(this,
                     c -> StepRegistryConfig.validate(c),
                     checkRequired("apiToken", DynatraceConfig::apiToken),
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -20,6 +20,8 @@ import io.micrometer.core.instrument.step.StepRegistryConfig;
 import io.micrometer.core.instrument.util.StringUtils;
 import io.micrometer.core.lang.Nullable;
 
+import java.util.function.Function;
+
 import static io.micrometer.core.instrument.config.MeterRegistryConfigValidator.*;
 import static io.micrometer.core.instrument.config.validate.PropertyValidator.*;
 
@@ -27,6 +29,7 @@ import static io.micrometer.core.instrument.config.validate.PropertyValidator.*;
  * Configuration for {@link DynatraceMeterRegistry}
  *
  * @author Oriol Barcelona
+ * @author Georg Pirklbauer
  */
 public interface DynatraceConfig extends StepRegistryConfig {
 
@@ -49,7 +52,7 @@ public interface DynatraceConfig extends StepRegistryConfig {
 
     default String technologyType() {
         return getSecret(this, "technologyType")
-                .map(v -> StringUtils.isEmpty(v) ? "java" : v)
+                .map(val -> StringUtils.isEmpty(val) ? "java" : val)
                 .get();
     }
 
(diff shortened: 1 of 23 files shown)


pivotal-issuemaster:
@pirgeo Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2553)!

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceCustomMetric.java
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.micrometer.dynatrace.v1;
+
+class DynatraceCustomMetric {
```
pirgeo, line 19: This was moved from the parent package.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceMeterRegistryImplV1.java
```diff
@@ -0,0 +1,300 @@
...
+
+/**
+ * Implementation for Dynatrace v1 metrics API export.
+ *
+ * @author Oriol Barcelona
+ * @author Jon Schneider
+ * @author Johnny Lim
+ * @author PJ Fanning
+ */
+public class DynatraceMeterRegistryImplV1 extends DynatraceMeterRegistryImplBase {
```
pirgeo, line 52: This was moved from the parent package. Most of the code is the same as in the original implementation in the `DynatraceMeterRegistry`. Changes represent the fact that this class does not extend the `StepMeterRegistry` and therefore has no access to the fields of the base class.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
```diff
@@ -64,14 +67,33 @@ default String group() {
...
-        return checkAll(this,
-                c -> StepRegistryConfig.validate(c),
-                checkRequired("apiToken", DynatraceConfig::apiToken),
-                checkRequired("uri", DynatraceConfig::uri),
-                checkRequired("deviceId", DynatraceConfig::deviceId),
-                check("technologyType", DynatraceConfig::technologyType).andThen(Validated::nonBlank)
-        );
+        // Currently only v1 is implemented. This check will be extended once more versions are added.
+        Function<DynatraceConfig, Validated<String>> versionFunc = checkRequired("apiVersion", DynatraceConfig::apiVersion);
+        String versionStr = versionFunc.apply(this).get();
```
jonatan-ivanov, line 86: You can do something like this instead so that you only change one line of this whole method:
```
checkRequired("apiVersion", DynatraceConfig::apiVersion).andThen(v -> v.invalidateWhen(<check-version-here>)),
```

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistryImplBase.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Base class for implementations of the {@link io.micrometer.dynatrace.DynatraceMeterRegistry} delegate.
+ *
+ * @author Georg Pirklbauer
+ */
+public abstract class DynatraceMeterRegistryImplBase {
```
jonatan-ivanov, line 31: To me it's not clear what is the purpose of this class.
Is this a `MeterRegistry` that will be used as a delegate or is it a client?

In the former case Could you please rename this class to something better, e.g.: `AbstractDynatraceMeterRegistry` and extend `StepMeterRegistry`?
Impl and Base seems weird to me, e.g.: if you want to implement cars your interface would be a `Car`, that you can specialize further by having an `AbstractCar implements Car` maybe a `SimpleCar extends AbstractCar` and a `SpecialCar extends AbstractCar`.

In the latter, could you please rename this class to something better, e.g.: `DynatraceV1Client` and clean it up a little so that it will look like a client not a registry?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceNamingConvention.java
```diff
@@ -30,7 +30,7 @@
  * @author Johnny Lim
  * @since 1.1.0
  */
-public class DynatraceNamingConvention implements NamingConvention {
+class DynatraceNamingConvention implements NamingConvention {
```
jonatan-ivanov, line 33: The other naming conventions are public (also, this is a breaking change but I'm not sure if it is really used).
@shakuzen Do you happen to know why are they public and if moving this is fine?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceCustomMetric.java
jonatan-ivanov: This seems like a new addition. Where did you move this from?

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/DynatraceConfigTest.java
```diff
@@ -62,4 +63,32 @@ void valid() {
...
+        Map<String, String> properties = new HashMap<String, String>() {{
+            put("dynatrace.apiToken", "secret");
+            put("dynatrace.uri", "https://uri.dynatrace.com");
+            put("dynatrace.deviceId", "device");
+            put("dynatrace.apiVersion", "v-INVALID");
+        }};
+
+        DynatraceConfig config = properties::get;
+
+        assertThrows(IllegalArgumentException.class, config::validate);
```
jonatan-ivanov, line 78: Please always check the error message and the cause too.

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/v1/DynatraceMeterRegistryImplV1Test.java
```diff
@@ -40,9 +42,10 @@
  *
  * @author Johnny Lim
  */
-class DynatraceMeterRegistryTest {
+class DynatraceMeterRegistryImplV1Test {
 
     private final DynatraceMeterRegistry meterRegistry = createMeterRegistry();
+    private final DynatraceMeterRegistryImplV1 meterRegistryImpl = createMeterRegistryImpl();
```
jonatan-ivanov, line 48: `meterRegistryImpl` is a weird term, it should be named like `someSpecialMeterRegistry`

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceMeterRegistryImplV1.java
jonatan-ivanov: Please check my comment I made on the abstract class. Could you please fix this too accordingly?

jonatan-ivanov (changes requested):

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceNamingConvention.java
shakuzen: They are public because users may want to extend the naming convention that we provide and use by default with some extra logic of their own. We would not want to break this in a minor release.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceCustomMetric.java
pirgeo: it was originally in `DynatraceMeterRegistry`.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistryImplBase.java
pirgeo: Renamed the abstract base class to `AbstractDynatraceExporter` and the implementation to `ApiV1DynatraceExporter`. I also renamed the function for exporting data to `export` so it's distinguishable from the `publish` function on the registry.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
```diff
@@ -64,14 +67,33 @@ default String group() {
         return get(prefix() + ".group");
     }
 
+    /**
+     * Return the version of the target Dynatrace API.
+     *
+     * @return a {@link String} containing the version of the targeted Dynatrace API.
+     */
+    default String apiVersion() {
```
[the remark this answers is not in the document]
pirgeo: Moved from a `String` to an `enum` and updated the tests accordingly.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
pirgeo: I simplified this a bit, also thanks to the suggestion of using an enum.

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/DynatraceConfigTest.java
pirgeo: The cause seems to be `null` for this case. As far as I understand that means that the Exception is already the root cause, does that seem right to you?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistryImplBase.java
pirgeo: Moved it again, now to `DynatraceExporterV1`, which seems more readable to me.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceNamingConvention.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micrometer.dynatrace;
+package io.micrometer.dynatrace.v1;
```
jonatan-ivanov, line 16: This is a breaking change, and based on @shakuzen's comment  (users can extend it and implement their own logic) we should not break this in a minor release.
What we can do instead is something similar that we did with the meter registry and hide the v1/v2 logic behind this class: having a v1 NamingConvention and a v2 NamingConvention and based on a ctor param set the right one (delegate), the default is v1.

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/v1/DynatraceExporterV1Test.java
```diff
@@ -40,9 +43,10 @@
  *
  * @author Johnny Lim
  */
-class DynatraceMeterRegistryTest {
+class DynatraceExporterV1Test {
 
     private final DynatraceMeterRegistry meterRegistry = createMeterRegistry();
+    private final DynatraceExporterV1 meterRegistryImpl = createMeterRegistryImpl();
```
jonatan-ivanov, line 49: Could you please remove "impl" from the names (methods, variables)? The exporter is not a registry anymore.

jonatan-ivanov (changes requested):

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/DynatraceConfigTest.java
```diff
@@ -62,4 +64,32 @@ void valid() {
...
+        Map<String, String> properties = new HashMap<String, String>() {{
+            put("dynatrace.apiToken", "secret");
+            put("dynatrace.uri", "https://uri.dynatrace.com");
+            put("dynatrace.deviceId", "device");
+            put("dynatrace.apiVersion", "v-INVALID");
+        }};
+        DynatraceConfig config = properties::get;
+
+        ValidationException ve = assertThrows(ValidationException.class, config::validate);
+        assertThat(ve.getMessage()).startsWith("dynatrace.apiVersion was 'v-INVALID' but it should be one of ");
```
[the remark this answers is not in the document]
pirgeo: Ah, that makes sense. I am not familiar with the validation system, so this was very helpful! Thanks.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
```diff
@@ -64,14 +66,30 @@ default String group() {
...
     default Validated<?> validate() {
-        return checkAll(this,
-                c -> StepRegistryConfig.validate(c),
-                checkRequired("apiToken", DynatraceConfig::apiToken),
-                checkRequired("uri", DynatraceConfig::uri),
-                checkRequired("deviceId", DynatraceConfig::deviceId),
-                check("technologyType", DynatraceConfig::technologyType).andThen(Validated::nonBlank)
-        );
+        // Currently only v1 is implemented. This check will be extended once more versions are added.
+        if (apiVersion() == DynatraceApiVersion.V1) {
```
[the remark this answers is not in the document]
pirgeo: This is much easier, and I adopted it for the v1 exporter. However, for the v2 exporter, we will only require the API version and checking these four required parameters with a `.andThen(v -> v.invalidateWhen(<check-version-here>)` construct for v1 seems a bit hard to read. Is there another way to make this more concise?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
```diff
@@ -64,14 +66,30 @@ default String group() {
...
+            return checkAll(this,
+                    c -> StepRegistryConfig.validate(c),
+                    checkRequired("apiToken", DynatraceConfig::apiToken),
+                    checkRequired("uri", DynatraceConfig::uri),
+                    checkRequired("deviceId", DynatraceConfig::deviceId),
+                    check("technologyType", DynatraceConfig::technologyType).andThen(Validated::nonBlank)
+            );
+        }
+
+        return Validated.invalid("apiVersion", apiVersion(), "API version could not be validated.", InvalidReason.MALFORMED);
```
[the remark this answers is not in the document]
pirgeo: I removed it after changing according to the code example above.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceNamingConvention.java
pirgeo: Moved it back up a level and added a delegate as suggested.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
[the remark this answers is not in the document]
jonatan-ivanov: InfluxConfix does something similar though, in this case, I would create a `validateV1` and a `validateV2` method (the current `validate` becomes `validateV1`). With them you can do:
```
@Override
default Validated<?> validate() {
    return apiVersion() == V1 ? validateV1() : validateV2();
}
```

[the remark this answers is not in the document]
pirgeo: I split it out to a separate function, and also removed the invalidVersion test, as the test data is invalid and can never happen in a non-test environment. If the spring boot app is started with an application property of `v-INVALID`, for example, it will just crash on startup and provide all valid values (which is just `V1`) in this case.
Another way to get around this would be to explicitly validate the `apiVersion` before accessing it, something like the following: 
```
@Override
default Validated<?> validate() {
    Validated<DynatraceApiVersion> apiVersionValidation = checkRequired("apiVersion", DynatraceConfig::apiVersion).apply(this);
    if (apiVersionValidation.isInvalid()) {
        return apiVersionValidation;
    }
    return apiVersion() == V1 ? validateV1() : validateV2();
}
```
which seems a bit unnecessary if this case never actually appears during use.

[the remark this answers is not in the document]
pirgeo: I just realized that that is only the case when using micrometer with Spring Boot. I think the solution I posted above (with the additional step of validating the apiVersion) should work, if i just access the `apiVersion()` method without validation first I recieve a `ValidationException`. I added the additional validation step.

pirgeo:
Hi @jonatan-ivanov, I think I addressed all the comments above. Please let me know if there's anything missing so we can move on with this. Thanks!

jonatan-ivanov:
@pirgeo Sorry for not getting back to you earlier. I merged your PR but not yet into master.
We are only merging smaller fixes to master between the RC1 and the GA version of 1.7.0. We can branch off 1.7.x but that means one more branch to maintain for us (we will do it after the release).

I'm going to try/fix a few things (validation*) on the `move-dynatrace-registry-v1` branch but in the meantime, you can handle this branch as master and build the v2 feature on top of it.

validation*: I totally forgot that Java only supports private methods in interfaces from Java 9. I'm not sure we should have a public validateV1 method on the config interface.

jonatan-ivanov:
@pirgeo please let me know what do you think: https://github.com/micrometer-metrics/micrometer/commit/4f534928b246cf2e5be219248ddf2594484d152b

jonatan-ivanov:
@pirgeo I also broke the circular dependency between the registry and the exporter: https://github.com/micrometer-metrics/micrometer/commit/1115208083bbfd302b075c29433775fbb27d0d8c

pirgeo:
@jonatan-ivanov Thanks a lot, these changes look good to me!

PR #2560, opened 2021-04-12
Refactor JvmGcMetrics with less specialized logic for non-gen collectors

shakuzen:
When we introduced support for non-generational collectors (Shenandoah, ZGC), the logic for their metrics was special case in the code. This refactor makes more code shared regardless of whether the collector is generational or not. The main difference with non-generational collectors is they do not have a promoted counter since there is no old generation. This refactoring should make it easier to support additional collectors (e.g. OpenJ9 collectors).

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -61,6 +61,8 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
     private static final InternalLogger log = InternalLoggerFactory.getInstance(JvmGcMetrics.class);
 
     private final boolean managementExtensionsPresent = isManagementExtensionsPresent();
+    // VisibleForTesting
+    final boolean isGenerationalGc = isGenerationalGcConfigured();
 
     private final Iterable<Tag> tags;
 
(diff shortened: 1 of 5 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -204,6 +187,13 @@ public void bindTo(MeterRegistry registry) {
         }
     }
 
+    private boolean isGenerationalGcConfigured() {
+        return ManagementFactory.getMemoryPoolMXBeans().stream()
+                .map(MemoryPoolMXBean::getName)
```
shakuzen, line 192: This should also filter for heap type pools. Will fix in a follow-up commit.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
```diff
@@ -47,12 +47,12 @@ static boolean isAllocationPool(String name) {
...
-    static boolean isOldGenPool(String name) {
-        return name != null && (name.endsWith("Old Gen") || name.endsWith("Tenured Gen"));
-    }
-
-    static boolean isNonGenerationalHeapPool(String name) {
-        return "Shenandoah".equals(name) || "ZHeap".equals(name);
+    static boolean isLongLivedPool(String name) {
+        return name != null && (name.endsWith("Old Gen")
+                || name.endsWith("Tenured Gen")
+                || "Shenandoah".equals(name)
```
shakuzen, line 53: Shenandoah and ZGC should also be added to the `isAllocationPool` method logic since they each are the allocation and long-lived pool for those non-generational collectors. Fixing in a follow-up commit.

shakuzen:
Addressed the above two issues in 5e40dc9

PR #2567, opened 2021-04-13
Pin version for org.hsqldb:hsqldb to 2.5.+

izeye:
This PR pins the version for `org.hsqldb:hsqldb` to `2.5.+` to allow this repo to be built on JDK 8 again as 2.6.0 requires JDK 11.


shakuzen (approved):
Thanks for catching this.

PR #2549, opened 2021-04-06
Isolate polling of each pollable meter in StatsdMeterRegistry poll()

matthewlowry:
Relates to #2543.

This change prevents a misbehaving pollable meter (a gauge with a custom value function that might throw an NPE, for example) from preventing ongoing polling of other pollable meters.

PR #2573, opened 2021-04-16
Log meter polling failure in `StatsdMeterRegistry`

izeye:
This PR changes to log meter polling failures once at warn and then at debug in `StatsdMeterRegistry`.

See gh-2549


jonatan-ivanov (approved):

jonatan-ivanov:
Thank you! :)

PR #2581, opened 2021-04-27
Add an exporter for the Dynatace metrics API v2

### LINKED ISSUE #2295: Support for Dynatrace metrics API v2

pirgeo:
This is the implementation for the Dynatrace metrics API v2.

Resolves #2295 
Closes #2406

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
@@ -44,11 +44,7 @@ import java.util.stream.StreamSupport;
  * @author Georg Pirklbauer
  */
 public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
-    private static final String DEFAULT_ONEAGENT_ENDPOINT = "http://127.0.0.1:14499/metrics/ingest";
-    private static final int MAX_BATCH_SIZE = 1000;
-
     private static final String METRIC_EXCEPTION_FORMATTER = "Could not serialize metric with name %s: %s";
-    private static final String ILLEGAL_ARGUMENT_EXCEPTION_FORMATTER = "Illegal value for metric with name %s: %s Dropping...";
 
     private static final Pattern EXTRACT_LINES_OK = Pattern.compile("\"linesOk\":\\s?(\\d+)");
     private static final Pattern EXTRACT_LINES_INVALID = Pattern.compile("\"linesInvalid\":\\s?(\\d+)");
@@ -60,13 +56,11 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
     private static final Logger logger = LoggerFactory.getLogger(DynatraceExporterV2.class.getName());
     private static final Map<String, String> staticDimensions = Collections.singletonMap("dt.metrics.source", "micrometer");
 
-    private static final int METRIC_LINE_MAX_LENGTH = 2000;
-
     public DynatraceExporterV2(DynatraceConfig config, Clock clock, HttpSender httpClient) {
         super(config, clock, httpClient);
 
         if (config.uri().isEmpty()) {
-            this.endpoint = DEFAULT_ONEAGENT_ENDPOINT;
+            this.endpoint = DynatraceMetricApiConstants.getDefaultOneAgentEndpoint();
         } else {
             this.endpoint = config.uri();
         }
@@ -95,19 +89,25 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
         return DimensionList.fromCollection(dimensions);
     }
 
+    /**
+     * Export to the Dynatrace v2 endpoint. Measurements that contain NaN or Infinite values, as
+     * well as serialized data points that exceed length limits imposed by the API will be dropped
+     * and not exported. If the number of serialized data points exceeds the maximum number of
+     * allowed data points per request they will be sent in chunks.
+     *
+     * @param meters A list of {@link Meter Meters} that are serialized as one or more metric lines.
+     */
     @Override
     public void export(@Nonnull List<Meter> meters) {
-        Map<Boolean, List<String>> metricLines =
+        // Lines that are too long to be ingested into Dynatrace, as well as lines that contain NaN
+        // or Inf values are dropped and not returned from "toMetricLines", and are therefore
+        // dropped.
+        List<String> metricLines =
                 meters.stream()                            // turn List<Meter> into Stream<Meter>
                         .flatMap(this::toMetricLines)      // turn Stream<Meter> into Stream<String>
-                        .collect(Collectors.partitioningBy(DynatraceExporterV2::lineLengthBelowLimit));
+                        .collect(Collectors.toList());     // Collect to List.
 
-        // both keys will be present, even if empty.
-        if (!metricLines.get(false).isEmpty()) {
-            logger.info("dropping {} lines that exceed the line length limit of {}", metricLines.get(false).size(), METRIC_LINE_MAX_LENGTH);
-        }
-
-        sendInBatches(metricLines.get(true));
+        sendInBatches(metricLines);
     }
 
     private void showErrorIfEndpointIsInvalid(String uri) {
@@ -170,18 +170,12 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
     private Stream<String> makeSummaryLine(Meter meter, double min, double max, double total, long count) {
         List<String> serializedLine = new ArrayList<>(1);
         try {
-            throwIfValueIsInvalid(max);
-            throwIfValueIsInvalid(total);
-
             serializedLine.add(
                     createMetricBuilder(meter)
                             .setDoubleSummaryValue(min, max, total, count)
                             .serialize());
         } catch (MetricException e) {
             logger.warn(String.format(METRIC_EXCEPTION_FORMATTER, meter.getId().getName(), e.getMessage()));
-        } catch (IllegalArgumentException iae) {
-            // drop lines containing NaN or Infinity silently.
-            logger.debug(String.format(ILLEGAL_ARGUMENT_EXCEPTION_FORMATTER, meter.getId().getName(), iae.getMessage()));
         }
 
         return streamOf(serializedLine);
@@ -240,15 +234,6 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
         return toGauge(meter);
     }
 
-    private void throwIfValueIsInvalid(Double value) {
-        if (value.isNaN()) {
-            throw new IllegalArgumentException("Value cannot be NaN.");
-        }
-        if (value.isInfinite()) {
-            throw new IllegalArgumentException("Value cannot be infinite.");
-        }
-    }
-
     Stream<String> toGauge(Meter meter) {
         return streamOf(meter.measure()).map(
                 measurement -> createGaugeLine(meter, measurement))
@@ -257,15 +242,11 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
 
     private String createGaugeLine(Meter meter, Measurement measurement) {
         try {
-            throwIfValueIsInvalid(measurement.getValue());
             return createMetricBuilder(meter)
                     .setDoubleGaugeValue(measurement.getValue())
                     .serialize();
         } catch (MetricException e) {
             logger.warn(String.format(METRIC_EXCEPTION_FORMATTER, meter.getId().getName(), e.getMessage()));
-        } catch (IllegalArgumentException iae) {
-            // drop lines containing NaN or Infinity silently.
-            logger.debug(String.format(ILLEGAL_ARGUMENT_EXCEPTION_FORMATTER, meter.getId().getName(), iae.getMessage()));
         }
         return null;
     }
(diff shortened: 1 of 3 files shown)


pirgeo:
Would it be possible to rename the target branch from `move-dynatrace-registry-v1` to something like `update-dynatrace-registries`? This might become quite confusing otherwise.

pirgeo:
We also opened PRs for the updated properties in Spring Boot and for the Micrometer documentation:
- Spring Boot: https://github.com/spring-projects/spring-boot/pull/26258
- Spring Boot issue: https://github.com/spring-projects/spring-boot/issues/24978
- Micrometer documentation: https://github.com/micrometer-metrics/micrometer-docs/pull/160

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -0,0 +1,383 @@
...
+        try {
+            String body = String.join("\n", metricLines);
+            if (logger.isDebugEnabled()) {
+                logger.debug("sending lines:\n" + body);
+            }
+
+            httpClient.post(endpoint)
+                    .withHeader("Authorization", "Api-Token " + config.apiToken())
+                    .withHeader("Content-Type", "text/plain")
+                    .withHeader("User-Agent", "micrometer")
```
jonatan-ivanov, line 333: Would it make sense to send the user-agent to the V1 api too?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -0,0 +1,383 @@
...
+    private void send(List<String> metricLines) {
+        try {
+            String body = String.join("\n", metricLines);
+            if (logger.isDebugEnabled()) {
+                logger.debug("sending lines:\n" + body);
+            }
+
+            httpClient.post(endpoint)
+                    .withHeader("Authorization", "Api-Token " + config.apiToken())
+                    .withHeader("Content-Type", "text/plain")
```
jonatan-ivanov, line 332: As far as I remember `withPlainText ` already sets the content type to  `text/plain`.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -0,0 +1,383 @@
...
+                    .withHeader("User-Agent", "micrometer")
+                    .withPlainText(body)
+                    .send()
+                    .onSuccess((r) -> handleSuccess(metricLines.size(), r))
+                    .onError((r) -> logger.error("Failed metric ingestion. code={} response.body={}", r.code(), r.body()));
```
jonatan-ivanov, line 337: extra space
```suggestion
                    .onError((r) -> logger.error("Failed metric ingestion.code={} response.body={}", r.code(), r.body()));
```

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -0,0 +1,383 @@
...
+        } catch (Throwable throwable) {
+            logger.error("Failed metric ingestion: {}", throwable.getMessage());
+        }
+    }
+
+    private void handleSuccess(int totalSent, HttpSender.Response r) {
```
jonatan-ivanov, line 343: How about `response`?
```suggestion
    private void handleSuccess(int totalSent, HttpSender.Response response) {
```

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -0,0 +1,383 @@
...
+                            totalSent, linesOkMatchResult.group(1), linesInvalidMatchResult.group(1));
+                } else {
+                    logger.warn("could not parse response: {}", r.body());
+                }
+            } else {
+                logger.warn("could not parse response: {}", r.body());
+            }
+        } else {
+            // common pitfall if URI is supplied in v1 format (without endpoint path)
+            logger.error("Expected status code 202, got {}. Did you specify the ingest path (e. g. /api/v2/metrics/ingest)?", r.code());
```
jonatan-ivanov, line 359: Should be `e.g.:`
```suggestion
            logger.error("Expected status code 202, got {}. Did you specify the ingest path (e.g.: /api/v2/metrics/ingest)?", r.code());
```

jonatan-ivanov: As far as I understand, this happens if `error` is not null in the response, right?
I'm wondering if logging `error` out would give more info to the users.

jonatan-ivanov (changes requested):

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
```diff
@@ -86,15 +110,22 @@ default DynatraceApiVersion apiVersion() {
...
                                                         checkRequired("apiToken", DynatraceConfig::apiToken),
                                                         checkRequired("uri", DynatraceConfig::uri),
                                                         checkRequired("deviceId", DynatraceConfig::deviceId),
                                                         check("technologyType", DynatraceConfig::technologyType).andThen(Validated::nonBlank)
                                                 );
                                             } else {
-                                                return apiVersionValidation; // V2 validation comes here
+                                                return checkAll(this,
+                                                        c -> StepRegistryConfig.validate(c),
+                                                        check("apiToken", DynatraceConfig::apiToken)
```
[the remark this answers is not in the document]
pirgeo: When an API token is present, we require both API token and URL. If there is no API token, we require only the URL, but i am not sure how to validate the URL in that case.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -0,0 +1,383 @@
...
+                (kv) -> Dimension.create(kv.getKey(), kv.getValue())
+        ).collect(Collectors.toList());
+        // construct a dimensionlist from the combined diemensions.
+        return DimensionList.fromCollection(dimensions);
+    }
+
+    @Override
+    public void export(@Nonnull List<List<Meter>> partitions) {
+        Map<Boolean, List<String>> metricLines =
+                partitions.stream()
```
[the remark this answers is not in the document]
pirgeo: The `MeterPartition` is part of `io.micrometer.core.instrument.util`, which I didn't necessarily want to change. Anyway, I think in our case it does not make a difference if we use the partition or not, unless is it somehow required by Micrometer?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
pirgeo: This happens if the HTTP status code is not 202, which is the signal for successfully ingested metrics. This can happen if the passed URL is a valid URL, but not a valid ingestion endpoint.

implementations/micrometer-registry-dynatrace/build.gradle
```diff
@@ -2,6 +2,8 @@ dependencies {
     api project(':micrometer-core')
 
     implementation 'org.slf4j:slf4j-api'
+    
+    implementation('com.dynatrace.metric.util:dynatrace-metric-utils-java:0.1.1')
```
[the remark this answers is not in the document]
pirgeo: Will do!

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
pirgeo: The user-agent header is not taken into account in the (legacy) V1 API, so we will leave it out.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
pirgeo: If the success code is not a 202 there is no error message. In that case, the response body will contain a whole webpage, not just a json object. Printing that is probably not helpful to the user, and it will clutter the logs.

pirgeo:
Hi @jonatan-ivanov, I think I have addressed all of the comments, please take another look :)

jonatan-ivanov:
@pirgeo Thank you very much. I back-merged master to the v1 refactor commit, I'm going to do some testing then merge it to master tomorrow. After that, I'm going to modify this PR and point to master and review your changes.

pirgeo:
@jonatan-ivanov We have released a new version of the Dynatrace exporter library which enforces some of the restrictions that we checked in this code before. We included the updated version in the latest commit (https://github.com/micrometer-metrics/micrometer/pull/2581/commits/bcd543d2c54dda62ce4319c4cc6247a613dfec6d). The checks that were removed in this code are now handled in the library.

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/v2/DynatraceExporterV2Test.java
```diff
@@ -354,6 +355,8 @@ void toCounterInvalidCases() {
         Counter counter = meterRegistry.find("my.counter1").counter();
         assertNotNull(counter);
 
+        // NaN and infinity are ignored. The counter value stays at 0.0 when adding one of NaN,
```
pirgeo, line 358: Please verify that the behavior we are experiencing here is expected.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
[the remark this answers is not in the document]
jonatan-ivanov: If I interpret and simplify this, this means that the url must always be non-empty but the token can be empty, right?

Basically this is the truth table for the validation if I understand this correctly:
token | uri | valid
------------ | ------------- | -------------
0 | 0 | 0
0 | 1 | 1
1 | 0 | 0
1 | 1 | 1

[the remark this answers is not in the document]
jonatan-ivanov: As I implemented this, I got confused since the exporter handles the case when the URL is empty, can we put this logic here?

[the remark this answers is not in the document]
jonatan-ivanov: What do you think?
https://github.com/jonatan-ivanov/micrometer/commit/46f2b77b747b017ba37a0ef585b98d0dc65c3c02

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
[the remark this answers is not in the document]
jonatan-ivanov: The point of partitioning is to balance between sending a lot of requests of one giant one.
One big request might not go through while many small ones can overwhelm the network.

So from the user perspective it is pretty desirable. I think adding a ctor to the `MeterPartition` is ok.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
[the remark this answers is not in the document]
pirgeo: > if the token is present the url is required but if there is no token we fall back to a default url?

I think it is slightly different from defaulting to the OneAgent endpoint and making the token optional, since it will not produce an error in the case where only a token is passed, but no endpoint. I like your proposal a lot, but I would like to take care of the case where only a token is passed and inform the user. Might I propose something like this: https://github.com/dynatrace-oss-contrib/micrometer/commit/2c43bddc5e260d44dbc0f8f0b479e08fd18c3771 ? When implementing this I also realized that we had sent an empty token if none is set, so I added a fix while I was at it.

jonatan-ivanov:
I merged the v1 refactor branch to `main` and deleted it. Since this was pointing to the v1 refactor, I pushed this to `add-dynatrace-registry-v2` so that I can work on it too (which marked this PR as merged).

PR #2587, opened 2021-04-29
More accurate metrics for 404 responses from Jersey2

bentmann:
I'm aware of the issue of tag explosion that motivates the mapping of 404 responses to a single URI such as "not_found". However, there are two distinct cases where the JAX-RS response can be 404:
- no REST resource could be matched to the incoming request
- a REST resource was matched to the request and the application code (via exception mapping or otherwise) created a response with 404 as status

I argue only the first case applies to the tag explosion concern. The second case however is not different than matched resources yielding HTTP 400.

With the proposed change, the HTTP 404 responses emitted by Jersey apps can be more appropriately attributed to the originating resource.

### DIFF
--- a/micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/JerseyTags.java
+++ b/micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/JerseyTags.java
@@ -95,7 +95,7 @@ public final class JerseyTags {
             if (isRedirection(status)) {
                 return URI_REDIRECTION;
             }
-            if (status == 404) {
+            if (status == 404 && event.getUriInfo().getMatchedResourceMethod() == null) {
                 return URI_NOT_FOUND;
             }
         }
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@bentmann Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2587)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2587) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@bentmann Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=2587)!

shakuzen:
Ported to micrometer-core in def88d1

Issue #2378, opened 2020-12-10
StatsdMeterRegistry infinite reconnect loop with Reactor 2020.0


shakuzen:
In some ad hoc testing, I noticed that behavior like described in #2177 is back again in 1.5.8, which contains #2359 upgrading reactor, reactor-netty, and netty. Since there were no other related changes in 1.5.8, it seems very likely that change is causing this issue to resurface. I will have to investigate more to understand why previously applied measures are no longer working.

Reverting the upgrade until a fix can be worked out may be the best course of action.

When the issue happens, which seems to be some variable amount of time after startup, with the statsd server down the whole time, DEBUG logs like the following are output rapidly and until the application is stopped:

```
2020-12-10 10:41:46.375 DEBUG 81279 --- [      udp-nio-2] i.m.s.r.n.r.NewConnectionProvider        : [id: 0xa0f006f1, L:/127.0.0.1:55452 - R:localhost/127.0.0.1:8125] Connected new channel
2020-12-10 10:41:46.375 DEBUG 81279 --- [      udp-nio-2] i.m.s.r.n.r.NewConnectionProvider        : [id: 0xa0f006f1, L:/127.0.0.1:55452 - R:localhost/127.0.0.1:8125] onStateChange([connected], SimpleConnection{channel=[id: 0xa0f006f1, L:/127.0.0.1:55452 - R:localhost/127.0.0.1:8125]})
2020-12-10 10:41:46.375 DEBUG 81279 --- [      udp-nio-2] i.m.s.r.n.r.NewConnectionProvider        : [id: 0xa0f006f1, L:/127.0.0.1:55452 - R:localhost/127.0.0.1:8125] onStateChange([configured], ChannelOperations{SimpleConnection{channel=[id: 0xa0f006f1, L:/127.0.0.1:55452 - R:localhost/127.0.0.1:8125]}})
2020-12-10 10:41:46.375 DEBUG 81279 --- [      udp-nio-2] i.m.shaded.reactor.netty.udp.UdpClient   : [id: 0xa0f006f1, L:/127.0.0.1:55452 - R:localhost/127.0.0.1:8125] Handler is being applied: io.micrometer.statsd.StatsdMeterRegistry$$Lambda$649/0x0000000800ea8f80@6b45cdee
[...]

Issue #792, opened 2018-08-17
Unix Domain Socket support (DataDog/StatsD)


kgoralski:
java-dogstatsd-client is going to support Unix Domain Socket (see https://github.com/DataDog/java-dogstatsd-client/pull/42)

is Micrometer is going to support it too? 

https://docs-staging.datadoghq.com/xvello/dsd_uds/developers/dogstatsd/unix_socket/
https://github.com/DataDog/datadog-agent/wiki/Unix-Domain-Sockets-support
https://github.com/DataDog/java-dogstatsd-client#unix-domain-socket-support

jonnywray:
This would be really useful in a Kubernetes environment for adding various deployment details, container, and k8s tags to the metrics.

We've experimented with different approaches for the DD reporter for Dropwizard metrics and UDS was the easiest to get up and running. Support here (and so for Spring Boot 2.x) would be great.

jonnywray:
For sure. In k8s the application knows nothing about the deployment environment, but of course it's useful for custom metrics coming from an app to have those, e.g. k8s namespace, host name, container details, etc. Using the DD API is out as that means the app needs the values.

There are two recommended ways of dealing with it outlined in the DD docs (https://docs.datadoghq.com/agent/kubernetes/dogstatsd/) and we've found the unix domain socket approach much easier to get going. And I think is the recommended approach now. Once this is working the various docker and k8s tags are added automatically to your custom metrics.

I tried the other approach, using a k8s hostPort and DogStatsD instance, but could never get it to work. 

There is a work around using DD API. You can use the k8s downward api to add environment variable to your application deployment and use them as custom tags in a MeterRegistryCustomizer. But, you don't get all available tags this way (e.g. docker image details) and automated tagging would be great.

jkschneider:
Cool, thanks! Sounds like a very useful feature.

PandaXass:
This feature could potentially work around https://github.com/containernetworking/plugins/issues/123 for Datadog agent?

Simwar:
@PandaXass I believe so, as the connection will be through the socket, not via the port mapped via hostPort.
@jkschneider is there any news about adding UDS support to micrometer?

aantono:
Something like this would need to be added: https://github.com/Frankqsy/netty-sample/blob/master/netty-client/src/main/java/org/daisy/stevin/netty/sample/client/uds/UdsClient.java

liu0013:
[...]

PR #2588, opened 2021-04-30
Upgrade to Reactor 2020.0

shakuzen:
Re-attempts #2359. This is using the `tryEmitNext` method instead of `emitNext` to avoid any failure handling - we want to drop elements on failure.
This implementation hopefully obviates #2378.
~Opened as a draft until we branch 1.7.x so `main` is ready for 1.8 development.~
This will make the implementation of #792 straightforward since Reactor Netty added support for UDS in 1.0.0.


implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/BufferingFluxTest.java
```diff
@@ -66,7 +66,7 @@ void bufferMultipleStrings() {
                 "fourteen bytes"
         );
 
-        Flux<String> buffered = BufferingFlux.create(source, "\n", 27, Long.MAX_VALUE);
+        Flux<String> buffered = BufferingFlux.create(source, "\n", 27, 1000);
```
shakuzen, line 69: See https://github.com/micrometer-metrics/micrometer/pull/2359/files#r529735243

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -79,8 +80,7 @@
     private final HierarchicalNameMapper nameMapper;
     private final Map<Meter.Id, StatsdPollable> pollableMeters = new ConcurrentHashMap<>();
     private final AtomicBoolean started = new AtomicBoolean();
-    DirectProcessor<String> processor = DirectProcessor.create();
-    FluxSink<String> fluxSink = new NoopFluxSink();
+    Sinks.Many<String> sink = new NoopManySink();
```
shakuzen, line 83: As I was working on this, I wished there was a built-in no-op implementation, but then I wondered if we really need one. If we don't subscribe, then it is almost a no-op. Maybe our `NoopManySink` isn't really needed?

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/LogbackMetricsSuppressingManySink.java
```diff
@@ -0,0 +1,81 @@
...
+    private final Sinks.Many<String> delegate;
+
+    public LogbackMetricsSuppressingManySink(Sinks.Many<String> delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public Sinks.EmitResult tryEmitNext(String s) {
+        LogbackMetrics.ignoreMetrics(() -> delegate.tryEmitNext(s));
+        return Sinks.EmitResult.OK;
```
shakuzen, line 37: We should document why we don't care about the EmitResult.

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
shakuzen: Let's explore this in a separate change.

PR #2603, opened 2021-05-12
Add connector name tag for `JettyConnectionMetrics` and `JettySslHandshakeMetricsBinder`

izeye:
This PR adds connector name tag for `JettyConnectionMetrics`.

See https://github.com/spring-projects/spring-boot/pull/26418#discussion_r631308899

/cc @wilkinsona

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
@@ -114,6 +114,36 @@ public class JettyConnectionMetrics extends AbstractLifeCycle implements Connect
                 .register(registry);
     }
 
+    /**
+     * Create a {@code JettyConnectionMetrics} instance. {@link Connector#getName()} will be used for
+     * {@literal connector.name} tag.
+     *
+     * @param registry registry to use
+     * @param connector connector to instrument
+     * @since 1.8.0
+     */
+    public JettyConnectionMetrics(MeterRegistry registry, Connector connector) {
+        this(registry, connector, Tags.empty());
+    }
+
+    /**
+     * Create a {@code JettyConnectionMetrics} instance. {@link Connector#getName()} will be used for
+     * {@literal connector.name} tag.
+     *
+     * @param registry registry to use
+     * @param connector connector to instrument
+     * @param tags tags to add to metrics
+     * @since 1.8.0
+     */
+    public JettyConnectionMetrics(MeterRegistry registry, Connector connector, Iterable<Tag> tags) {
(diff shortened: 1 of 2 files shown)


shakuzen (commented):
I'm generally in favor of this change, but it makes an asymmetry between the outcome of the two ways the `JettyConnectionMetrics` can be applied. If you do `connector.addBean(new JettyConnectionMetrics(registry))` then the `connector.name` tag won't be there. Since we don't have a reference to the `Connector`, we can't do much about that given the current API. I wonder if we shouldn't move the API in the direction of passing the `Connector` to the constructor so we can add the `connector.name` tag in either case.

wilkinsona:
I also wonder if it should be possible not to use the connector’s name as a tag? In the

izeye:
@wilkinsona I'm not sure if I fully understand your comment. Do you mean a flag to control addition of the tag for connector's name? Maybe, you might mean to provide more context in the following incomplete sentence that seems to have been removed accidentally.

wilkinsona:
Oops. Sorry, not sure how I managed that. I intended to say the following:

In the single connector case, if things change as suggested above by @shakuzen, that could be done by only setting the `name` tag if the user just passes in the `Connector`. If they call a `Connector, Iterable<Tag>` method then the default `name` tag could perhaps be skipped. That doesn't work so well for the multi-connector case as there's no way to provide connector-specific tags. Maybe that doesn't matter so much and the user could just call the new single connector method multiple times?

izeye:
@wilkinsona Thanks for clarification! I can see how it could work, but I'm not sure if it's natural to use them as a signal for adding the tag. Maybe, I might miss something.

izeye:
I updated this PR to try to apply the suggestion from @shakuzen. I didn't deprecate the previous constructors as they provide a way not to use the connector's name as a tag. I wasn't able to apply the suggestion from @wilkinsona as I don't think I fully understand it.

Let me know what you think.

If it's okay, I'll update this PR to apply the same changes to `JettySslHandshakeMetricsBinder` as well.

See https://github.com/spring-projects/spring-boot/pull/26418#discussion_r631309055

wilkinsona:
Sorry for the confusion, @izeye. Please ignore my suggested way of achieving the goal. It's really just the goal itself that's important and that is for a user to be able to register metrics for a `Connector` without the `connector.name` tag being added automatically. If this isn't a pattern that's followed elsewhere in Micrometer then please disregard this. I'd rather than the `connector.name` tag was always added than it never was.

izeye:
@wilkinsona Thanks for the feedback!

I updated this PR to apply the same changes to `JettySslHandshakeMetricsBinder`.

Please let me know if anything else needs to be done.

wilkinsona:
Looks good to me, @izeye.

shakuzen (approved):
Looks good to me.

PR #2604, opened 2021-05-17
Mark 1.3.20 as a final patch for 1.3.x

izeye:
This PR marks 1.3.20 as a final patch for 1.3.x as looking at https://github.com/micrometer-metrics/micrometer-docs/commit/42a8ba3a127851f6d8a64d34131250a993d4562b, 1.3.x seems to have been EOLed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ See [Micrometer's support policy](https://micrometer.io/docs/support) for more d
 | `1.0.x`            | Yes | `1.0.11`    |
 | `1.1.x`            | Yes | `1.1.19`    |
 | `1.2.x`            | No  | `1.2.2`     |
-| `1.3.x`            | Yes |  |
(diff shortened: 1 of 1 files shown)


shakuzen:
There is a possibility that a commercial support request will require us to do another release of the 1.3.x line still. I generally expect 1.3.20 will be the final release, but I don't want to make it official until it's finalized. Effectively, OSS support has ended for 1.3.x, though.

izeye:
@shakuzen Thanks for the feedback! I wasn't aware of the commercial support. Is there any scheduled date for the end of its commercial support?

izeye:
This has been superseded by 8ef39b44ddf3b23e404d01cc70244159aa10e62d and 5dd4ffc6e324915715655f779354f98871e275f9.

PR #2607, opened 2021-05-19
Dynatrace v1 refactor

jonatan-ivanov:
Refactoring v1 so that we can add support for v2, see: #2295


jonatan-ivanov:
@pirgeo I was testing this today and bumped into an intermittent error coming back from the Dynatrace backend:

```
:08:01 dynatrace-metrics-publisher ERROR i.m.d.v.DynatraceExporterV1 - failed to send metrics to dynatrace: {"error":{"code":400,"message":"Constraints violated.","constraintViolations":[{"path":"series[29].datapoints[0]","message":"Data point timestamp is lower than creation timestamp of the timeseries configuration.Creation timestamp is: 1621890481076","parameterLocation":"PAYLOAD_BODY","location":null}]}}
```

I'm using the `DynatraceSample`. Atm I don't know if this is happening because of a change on the backend, the refactor or something else (the branch is up-to-date with `main`). My clock seems to be in-sync.
Have you seen this issue before? Do you happen to know what could cause this?

pirgeo:
@jonatan-ivanov Is this only happening in the refactored version? I can't reproduce the problem, but I assume that it has something to do with the fact that in the old v1 API, the timeseries receivers first have to be explicitly created before data is ingested. This happens on the first export. If the data points are recorded before the first export (even though they are only exported after) they might still hold timestamps that are lower than the creation time. This should only happen on the first export, after that the receivers should exist and the error should no longer appear.

How long is the export interval that you are testing here?

jonatan-ivanov:
@pirgeo That makes perfect sense and explains the behavior I saw. With this I was able to repro the issue reliably by just creating a new meter manually that Dynatrace did not know before. I was also able to repro this on `main` so it is not specific to the refactor. I don't know if we can anything about it, but created an issue for the sake of tracking this: https://github.com/micrometer-metrics/micrometer/issues/2618

PR #1027, opened 2018-11-23
Relocate META-INF/native/libnetty_transport_native_epoll_x86_64.so

izeye:
This PR relocates `META-INF/native/libnetty_transport_native_epoll_x86_64.so` to `META-INF/native/libio_micrometer_shaded_netty_transport_native_epoll_x86_64.so`.

Fixes gh-976

PR #2611, opened 2021-05-20
Rename netty native files so they can be loaded

shakuzen:
Without this, I was getting a failure using the statsd module on macOS because our shaded netty could not load the shaded macOS native DNS library. See similar error in https://github.com/netty/netty/pull/11260. The macos native DNS library is being used from reactor-netty 1.x, which is why we don't see this issue in prior release lines.

This change is a redo of #1027 which was previously reverted in #1058 due to #1054. Do we have a way to reproduce the scenario that was happening in #1054? I tried running an example app on linux with this change and didn't immediately see the error, but I don't know the exact conditions for it either.

/cc @izeye in case you know more about the previous issue and how to check it doesn't get reintroduced by this change again.

### DIFF
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -26,7 +26,7 @@ shadowJar {
     relocate 'reactor', 'io.micrometer.shaded.reactor'
     relocate 'org.reactivestreams', 'io.micrometer.shaded.org.reactorstreams'
     relocate 'io.netty', 'io.micrometer.shaded.io.netty'
-    exclude 'META-INF/native/libnetty_transport_native_epoll_x86_64.so'
+    relocate 'META-INF/native/libnetty', 'META-INF/native/libio_micrometer_shaded_netty'
     metaInf {
         from "$rootDir/LICENSE"
         from "$rootDir/NOTICE"


violetagg (approved):

jonatan-ivanov (approved):

shakuzen:
Testing on Linux, I can see we get logs like the following with this change (and therefore epoll enabled) when the statsd daemon isn't running:
```
2021-05-28 03:11:45.544  WARN 6373 --- [    udp-epoll-1] i.m.s.reactor.netty.channel.FluxReceive  : [id:35f314ec, L:/127.0.0.1:60434 - R:localhost/127.0.0.1:8125] An exception has been observed post termination, use DEBUG level to see the full stack: java.net.PortUnreachableException: readAddress(..) failed: Connection refused
2021-05-28 03:11:45.554  WARN 6373 --- [    udp-epoll-1] i.m.s.reactor.netty.channel.FluxReceive  : [id:35f314ec, L:/127.0.0.1:60434 - R:localhost/127.0.0.1:8125] An exception has been observed post termination, use DEBUG level to see the full stack: java.net.PortUnreachableException: readAddress(..) failed: Connection refused
```
This is a bit verbose compared to without epoll enabled, but we can improve that in a separate change, I think.

PR #2619, opened 2021-05-26
Support for Dynatrace V2 API

jonatan-ivanov:


jonatan-ivanov:
@pirgeo I made a few improvements and it seems I also found a bug:

If I let this run:
```java
public static void main(String[] args) throws InterruptedException {
    DynatraceConfig config = new DynatraceConfig() {
        @Override
        public DynatraceApiVersion apiVersion() {
            return V2;
        }

        @Override
        public String uri() {
            return "<uri>";
        }

        @Override
        public String apiToken() {
            return "<api-key>";
        }

        @Override
        public String get(String key) {
            return null;
        }
    };

    DynatraceMeterRegistry registry = DynatraceMeterRegistry.builder(config).build();
    registry.gauge("test-gauge-v2", 1.0, v -> (int) (Math.random() * 10 + 10));
    registry.gauge("test-gauge-v2-small", 1.0, v -> Math.random());

    Thread.sleep(Long.MAX_VALUE);
}
```

After a couple of iterations, I bump into a serialization issue with the gauges:
```
:42:00 dynatrace-metrics-publisher WARN i.m.d.v.DynatraceExporterV2 - Could not serialize metric with name test-gauge-v2-small: Value was NaN.
:42:00 dynatrace-metrics-publisher WARN i.m.d.v.DynatraceExporterV2 - Could not serialize metric with name test-gauge-v2: Value was NaN.
```

The error is coming from the `Metric.Builder` (`.serialize()`). As your comments says this should skip `NaN` but once this gets `NaN`, it will be `NaN` next time and no metrics are published.

I need to take a look at this deeper but as far as I remember I was able to repro the issue before my changes so they should not effect it.

jonatan-ivanov:
@pirgeo I tried to improve this a little, I haven't checked the validation logic and the tests (including the Nan/Inf behavior) but I'm going to do it soon.

shank9918:
Hi @jonatan-ivanov, just wanted to check on the availability of these changes. We are using Dynatrace V1 APIs with micrometer-dynatrace-registry at the moment and looking forward to using the V2 APIs. So, when do you think these will changes will be released?

jonatan-ivanov:
@shank9918 After this PR get merged, you can use it as a snapshot release (should happen soon, see my previous comment about what needs to be done).
As for non-snapshot releases, we need sync-up with the Dynatrace OSS Team and figure out how/when to release this.

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/v2/DynatraceExporterV2Test.java
```diff
@@ -0,0 +1,404 @@
...
+        assertThat(actual).isEmpty();
+    }
+
+    @Test
+    void toCounterInvalidCases() {
+        meterRegistry.counter("my.counter1");
+        Counter counter = meterRegistry.find("my.counter1").counter();
+        assertNotNull(counter);
+
+        // NaN and infinity are ignored. The counter value stays at 0.0 when adding one of NaN,
```
jonatan-ivanov, line 358: @pirgeo You see this behavior because there is a step counter under the hood that returns you the previous value (`0.0`) until the next step.

jonatan-ivanov:
@pirgeo I think this is close to be merged, I fixed a couple of things:
- Added your proposal for the validation
- Fixed tests by using `MockClock` instead of `Thread.sleep`
- Registering 0th percentile (for min) overwritten the user-defined ones (if any) so I implemented a [fix](https://github.com/micrometer-metrics/micrometer/pull/2619/commits/499a4d81bdd32eca87548f6a0fc6b83d438ef320)

I also have a few notes, just to double check if these are intended and you are fine with them:
- Registering 0th percentile (for min) will also mean that the 0th percentile gauge will be reported to Dynatrace, this happens if I register a timer without any user-defined tags or percentiles
```
test-timer-v2,dt.metrics.source=micrometer gauge,min=125.82912,max=190.0,sum=954.0,count=6 1622746380008
test-timer-v2.percentile,phi=0,dt.metrics.source=micrometer gauge,125.82912 1622746380011
```
- The Dynatrace java lib does not filter out empty tags, so this will be sent on the wire:
```
test-timer-v2,tag1=value1,dt.metrics.source=micrometer,tag2= gauge,min=142.606336,max=191.0,sum=866.0,count=5 1622745540019
```
Please notice `tag2=`. This does not seem to cause any issues, the tag is filtered out on the server-side.

Could you please check if these look ok for you?

pirgeo (commented):
@jonatan-ivanov Thanks, the changes are looking very good to me.
- Regarding the 0th percentile export: We would rather not export the extra percentile line. Do you have any suggestions as to how we could drop just that line, preferably without having to look through the created lines and dropping it after creation?
- Regarding the empty tags, this is something we would like to keep, for now. As you saw, empty tags are currently dropped upon ingestion, but might be used in the future. I have added suggestions to change the test method names.

jonatan-ivanov:
@pirgeo Great, I'm happy you liked them.

**Empty tags:** Makes sense, thanks for catching the test names, I've merged your changes in.

**0th percentile export:** I agree. I think we should not export the extra percentile line, I wasn't 100% sure if it was intentional or not. I think there are two ways to solve this:
1. Hardcoding a `0` value if there is no good use-case for `min` (we are doing this for other meter registries)
2. Using the same `MeterFilter` to deny the artificially created 0th percentile Meters. This is somewhat tricky/hacky since `accept` does not know about the `DistributionConfig`. Because of this, we need to record which meters can be the subject of removal in the `config` phase, see: https://github.com/micrometer-metrics/micrometer/pull/2619/commits/21da805c0f7aa6df9339af9a9de6c079ec42ad88.
This fortunately happens once per registered meter and before the registry or the exporter would receive it. Also, the exporter does not need the `Meter` itself, it only needs the `HistogramSnapshot`, please see the tests.

What do you think?

jonatan-ivanov:
I merged it based on the email exchange we had about this.
@pirgeo Thank you very much for your epic contribution, we really appreciate your help!

pirgeo:
@jonatan-ivanov Thanks a lot! Appreciate your commitment and your help :)

PR #2623, opened 2021-05-27
Include the number of active instances in CommonsObjectPool2Metrics

wengertj:
This PR adds the number of currently active instances in the pool as a gauge, very much like the exisiting `num.idle` gauge.

(...I had to refactor the test a little bit. The "dangling" pool instances made the test very flaky. If you have a cleaner/nicer solution than try-with-resources I'm open for it 👍 )


jonatan-ivanov:
@wengertj Thank you very much for the PR! :)

PR #2631, opened 2021-06-02
missing @Nullable added to Timer.record

### LINKED ISSUE #2630: @Nullable annotation missing for Timer.record() methods
The following two methods of `Timer` interface are missing `@Nullable` annotation.

```java
    <T> T record(Supplier<T> f);
    <T> T recordCallable(Callable<T> f) throws Exception;
```

---

It's not guaranteed that `Supplier` (or `Callable`) provided by a user will return a nonnull value.

As `Timer` interface inherits `@Nonnull` annotation from [package-info file](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/package-info.java) this causes Kotlin compiler to treat the above methods as never returning null, which is not true.

A simple example why `@Nullable` should be applied:
image

jkrauze:
fixes #2630

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
@@ -119,6 +119,7 @@ public interface Timer extends Meter, HistogramSupport {
      * @param <T> The return type of the {@link Supplier}.
      * @return The return value of {@code f}.
      */
+    @Nullable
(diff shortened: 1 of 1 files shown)


shakuzen (approved):
Thank you for finding, reporting, and fixing this. I'll retarget it to the `1.5.x` branch so we can include it in upcoming patch releases.

Issue #2637, opened 2021-06-04
Native image support investigation


shakuzen:
The purpose of this task is to investigate and identify what the gaps are between Micrometer metrics available on the JVM vs running as a native image (i.e. GraalVM native images).

Most things already work and we are not aware of any outright failures. We have already added some code to guard against unavailable things in native images that throw exceptions, e.g. #1948.

The result of this investigation should be tasks in Micrometer and in native image tools like GraalVM.

shakuzen:
### What's the diff
Not an exhaustive list, but from comparing one sample application on the JVM vs compiled as a GraalVM native image I found the following.

#### Missing
The GraalVM native image was missing the following metrics:
* FileDescriptorMetrics
* ProcessorMetrics
* JvmMemoryMetrics
* JvmGcMetrics

#### Needs attention
* Classloader metrics are present (though always have a 0 value). I don't think it makes sense to register classloader metrics when running as a native image.
* `jvm.*` prefix is on various metrics. Does it make sense to keep this for native images?
* JvmThreadMetrics mostly works already. Only the metrics about thread state are missing.

### Platform MBeans
All of the things mentioned above are based on JMX MBeans.
While the JVM platform MBeans cannot be used in native images, GraalVM provides substitutions that implement some of the functionality in them where it makes sense. See work like https://github.com/oracle/graal/pull/2547.

shakuzen:
### Micrometer tasks
- [ ] FileDescriptorMetrics reflection config
- [ ] ProcessorMetrics reflection config

### GraalVM
* [BufferPoolMXBeans and MemoryPoolMXBeans are unsupported for now](https://github.com/oracle/graal/blob/89e4cfc7aeea69970b60c64cd075ceb2a104e864/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/management/ManagementSupport.java#L139-L144)

shakuzen:
I think this investigation is done for now. We may find additional gaps to work on in the future. I have shared my findings with the GraalVM team on what we aren't able to fix with changes in Micrometer. If there are relevant updates to GraalVM, we can take advantage of that accordingly.

Haarolean:
@shakuzen hi, considering https://github.com/oracle/graal/issues/4510 is implemented, is there anything else that has to be done on micrometer side to make this work? Currently I don't see the required metrics within native build.

shakuzen:
[...]

PR #2638, opened 2021-06-04
Add reflection config for ProcessorMetrics and FileDescriptorMetrics

shakuzen:
Without this configuration, ProcessorMetrics and FileDescriptorMetrics are not available in a Micrometer application run as a native image. This explicitly lets the native image compiler know what class and methods will have reflection used on them. With this configuration, the ProcessorMetrics and FileDescriptorMetrics are the same in applications run on the JVM or as native images.

Related to #2637

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,6 +117,7 @@ subprojects {
         license {
             ext.year = Calendar.getInstance().get(Calendar.YEAR)
             skipExistingHeaders = true
+            exclude '**/*.json' // comments not supported
         }
 
         // Publish resolved versions.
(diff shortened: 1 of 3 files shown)


shakuzen:
Since I don't think comments are supported in JSON, we'll need to make the license check skip JSON files like reflect-config.json.

jonatan-ivanov (approved):

shakuzen:
Considering this as low risk (famous last words) and arguably fixing a bug in our native image support (albeit we've never officially claimed everything works in native images), I'm going to target it to 1.5.x. Code changes, on the other hand, I would be much more hesitant to backport to improve native image support.

PR #2643, opened 2021-06-09
Avoid keeping a strong reference to caches being monitored

### LINKED ISSUE #2642: Monitored Cache is not elligible for garbage collection
The[ `io.micrometer.core.instrument.binder.cache.CaffeineCacheMetrics`](https://github.com/micrometer-metrics/micrometer/blob/258a2f8ad2342244d950ba48d45792f2c373d8b8/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java) class keeps a strong reference to the `com.github.benmanes.caffeine.cache.Cache` instance that `CaffeineCacheMetrics` was created with. The `CacheMeterBinder` super-class on the other hand has a reference wrapped in `WeakReference`. Is it intentional that `CaffeineCacheMetrics` keeps a strong reference?

Combining this with the fact that the gauges created in `CacheMeterBinder.bindTo` reference the `size()`, `missCount()`, `hitCount()`, `putCount()` and `evictionCount()` methods in `CacheMeterBinder`, as far as I can tell, it means that monitored `Cache` instances are never eligible for garbage collection.

slovdahl:
Fixes #2642.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
@@ -177,7 +177,7 @@ public class CaffeineCacheMetrics extends CacheMeterBinder {
                 .description("The sum of weights of evicted entries. This total does not include manual invalidations.")
                 .register(registry);
 
-        if (cache instanceof LoadingCache) {
+        if (cache.get() instanceof LoadingCache) {
             // dividing these gives you a measure of load latency
             TimeGauge.builder("cache.load.duration", cache.get(), TimeUnit.NANOSECONDS, c -> c.stats().totalLoadTime())
                     .tags(getTagsWithCacheName())
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
@@ -20,9 +20,11 @@ import com.github.benmanes.caffeine.cache.stats.CacheStats;
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
+import io.micrometer.core.lang.Nullable;
 
 import java.lang.ref.WeakReference;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 /**
  * Collect metrics from Caffeine's {@link com.github.benmanes.caffeine.cache.Cache}. {@link CaffeineStatsCounter} is an
@@ -122,52 +124,27 @@ public class CaffeineCacheMetrics extends CacheMeterBinder {
 
     @Override
     protected Long size() {
-        Cache<?, ?> ref = cache.get();
-        if (ref != null) {
-            return ref.estimatedSize();
-        }
-
-        return null;
+        return getOrDefault(Cache::estimatedSize, null);
     }
 
     @Override
     protected long hitCount() {
-        Cache<?, ?> ref = cache.get();
-        if (ref != null) {
-            return ref.stats().hitCount();
-        }
-
-        return 0L;
+        return getOrDefault(c -> c.stats().hitCount(), 0L);
     }
 
     @Override
     protected Long missCount() {
-        Cache<?, ?> ref = cache.get();
-        if (ref != null) {
-            return ref.stats().missCount();
-        }
-
-        return null;
+        return getOrDefault(c -> c.stats().missCount(), null);
     }
 
     @Override
     protected Long evictionCount() {
-        Cache<?, ?> ref = cache.get();
-        if (ref != null) {
-            return ref.stats().evictionCount();
-        }
-
-        return null;
+        return getOrDefault(c -> c.stats().evictionCount(), null);
     }
 
     @Override
     protected long putCount() {
-        Cache<?, ?> ref = cache.get();
-        if (ref != null) {
-            return ref.stats().loadCount();
-        }
-
-        return 0L;
+        return getOrDefault(c -> c.stats().loadCount(), 0L);
     }
 
     @Override
@@ -197,5 +174,25 @@ public class CaffeineCacheMetrics extends CacheMeterBinder {
                             "because no value was found or an exception was thrown while loading")
                     .register(registry);
         }
+    }
+
+    @Nullable
+    private Long getOrDefault(Function<Cache<?, ?>, Long> function, @Nullable Long defaultValue) {
+        Cache<?, ?> ref = cache.get();
+        if (ref != null) {
+            return function.apply(ref);
+        }
+
+        return defaultValue;
+    }
+
+    @Nullable
+    private long getOrDefault(Function<Cache<?, ?>, Long> function, @Nullable long defaultValue) {
+        Cache<?, ?> ref = cache.get();
+        if (ref != null) {
+            return function.apply(ref);
+        }
+
+        return defaultValue;
     }
 }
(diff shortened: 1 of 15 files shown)


slovdahl:
I'm getting test failures locally in `CaffeineCacheMetricsTest.reportExpectedGeneralMetrics`, and I can't figure why that is. Works fine locally on upstream main.

shakuzen (commented):
The implementations of Gauge and FunctionCounter do not call the value function if the underlying obj is `null`. For example, in `CacheMeterBinder`, where it has
```java
FunctionCounter.builder("cache.gets", cache.get(), c -> hitCount())
```
if the weak reference's `get()` returns `null` then the `hitCount()` method will not be called. This is why with the current implementation, we don't really care what the abstract methods return if the cache is GC'd because they won't be called. This behavior is not very apparent and took me a bit of looking at it to realize.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
```diff
@@ -121,50 +122,75 @@ public CaffeineCacheMetrics(Cache<?, ?> cache, String cacheName, Iterable<Tag> t
...
 
     @Override
     protected void bindImplementationSpecificMetrics(MeterRegistry registry) {
-        FunctionCounter.builder("cache.eviction.weight", cache, c -> c.stats().evictionWeight())
+        FunctionCounter.builder("cache.eviction.weight", cache.get(), c -> c.stats().evictionWeight())
                 .tags(getTagsWithCacheName())
                 .description("The sum of weights of evicted entries. This total does not include manual invalidations.")
                 .register(registry);
 
         if (cache instanceof LoadingCache) {
```
shakuzen, line 180: This is why the test is failing. The type needs to be checked on `cache.get()` now

slovdahl: Ah, it's so obvious once you see it. Thanks!

slovdahl:
@shakuzen do you want me to fix all the caches in this same PR, or should I do it in separate PRs?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
slovdahl: Fixed, and the tests are green :tada:

slovdahl:
@shakuzen do you want me to add some tests for validating that no strong reference is kept? It's probably doable by doing one or mor e `System.gc()` calls after clearing the reference to the `Cache` instance.

shakuzen:
Yes, I think the cause and fix is the same for all so fixing them all in the same pull request sounds reasonable to me. And yes, a test would be good. I have one locally I can work into the pull request if you don't feel like making one.

slovdahl:
> Yes, I think the cause and fix is the same for all so fixing them all in the same pull request sounds reasonable to me

I'll update the PR with the same kind of fix for all them then :+1: 

> And yes, a test would be good. I have one locally I can work into the pull request if you don't feel like making one.

Sounds great!

slovdahl:
Updated the PR accordingly. AFAICT, the `JCacheMetrics` class does not need any `WeakReference` wrapping. Do you agree with that?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
```diff
@@ -121,50 +122,75 @@ public CaffeineCacheMetrics(Cache<?, ?> cache, String cacheName, Iterable<Tag> t
 
     @Override
     protected Long size() {
-        return cache.estimatedSize();
+        Cache<?, ?> ref = cache.get();
```
jonatan-ivanov, line 125: An improvement for these could be a common method that receives:
- a lambda that returns a value based on the cache
- a default value in case the reference is null

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetrics.java
```diff
@@ -138,23 +153,23 @@ protected void bindImplementationSpecificMetrics(MeterRegistry registry) {
     }
 
     private void nearCacheMetrics(MeterRegistry registry) {
-        if (cache.getLocalMapStats().getNearCacheStats() != null) {
-            Gauge.builder("cache.near.requests", cache, cache -> cache.getLocalMapStats().getNearCacheStats().getHits())
+        if (cache.getLocalMapStats() != null && cache.getLocalMapStats().getNearCacheStats() != null) {
+            Gauge.builder("cache.near.requests", cache, cache -> cache.getLocalMapStats() != null ? cache.getLocalMapStats().getNearCacheStats().getHits() : Double.NaN)
```
jonatan-ivanov, line 157: At this point, `cache.getLocalMapStats()` can't be null, right?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
slovdahl: Good idea :+1: 

Two different common methods, one for `Long` and one for `long`, or what's your preference?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetrics.java
slovdahl: Hmm, good question. I think you may be right, but it's a bit confusing with the subtle dependencies between the two classes now (`HazelcastCacheMetrics` and `HazelcastIMapAdapter`).

The method is annotated with `@Nullable` though, so removing the `null` check will cause IntelliJ to complain. :unamused: But I can definitely remove the explicit `null` checks if you want me to.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
slovdahl: Pushed a version with two methods now for `CaffeineCacheMetrics`, `GuavaCacheMetrics` and `EhCache2Metrics`.

jonatan-ivanov: I think what you did is great. It's not a must at the moment but I think the method with long can be more generic by doing something like this: `private <T> T getOrDefault(Function<Cache<?, ?>, T> function, @Nullable T defaultValue)`.

I don't think this is very pressing since we only have one non-primitive type now but could be a nice touch for the future.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetrics.java
jonatan-ivanov: I dug deeper on this, I don't think anymore that my previous comment is valid. :)
This seemed a little weird for first sight but it makes perfect sense for the second, thanks.

jonatan-ivanov (approved):

jonatan-ivanov:
This looks good to me the only thing I miss is testing this, I guess we have two issues here:
1. Forcing weak references to return null
2. Preventing weak references to return null (if GC will clean things up during tests, our current tests could break since we will get the default value)

@shakuzen Could you please share what do you have locally to test this?
I'm thinking about having another ctor that accepts a reference so that we can control it and simulate "GC" events.

shakuzen (approved):

shakuzen:
I was going to merge this to the maintenance branches originally, but with the addition of generics, it is more appropriate to include this in the next feature release. We can separately consider a back-port to maintenance branches without the generics if there is demand.

PR #2646, opened 2021-06-13
Improve escape method speed and memory allocation in InfluxNamingConvention

tonytamwk:
```
Benchmark                                                                     Mode  Cnt     Score   Error   Units
ReplaceEscapeCharactersBenchmark.regex                                        avgt    2   514.968           ns/op
ReplaceEscapeCharactersBenchmark.regex:·gc.alloc.rate                         avgt    2  1128.722          MB/sec
ReplaceEscapeCharactersBenchmark.regex:·gc.alloc.rate.norm                    avgt    2   640.000            B/op
ReplaceEscapeCharactersBenchmark.regex:·gc.churn.G1_Eden_Space                avgt    2  1142.778          MB/sec
ReplaceEscapeCharactersBenchmark.regex:·gc.churn.G1_Eden_Space.norm           avgt    2   647.971            B/op
ReplaceEscapeCharactersBenchmark.regex:·gc.count                              avgt    2    40.000          counts
ReplaceEscapeCharactersBenchmark.regex:·gc.time                               avgt    2    35.000              ms
ReplaceEscapeCharactersBenchmark.replaceFromMap                               avgt    2   115.703           ns/op
ReplaceEscapeCharactersBenchmark.replaceFromMap:·gc.alloc.rate                avgt    2   690.757          MB/sec
ReplaceEscapeCharactersBenchmark.replaceFromMap:·gc.alloc.rate.norm           avgt    2    88.000            B/op
ReplaceEscapeCharactersBenchmark.replaceFromMap:·gc.churn.G1_Eden_Space       avgt    2   714.227          MB/sec
ReplaceEscapeCharactersBenchmark.replaceFromMap:·gc.churn.G1_Eden_Space.norm  avgt    2    90.976            B/op
ReplaceEscapeCharactersBenchmark.replaceFromMap:·gc.count                     avgt    2    25.000          counts
ReplaceEscapeCharactersBenchmark.replaceFromMap:·gc.time                      avgt    2    32.000              ms
```

### DIFF
--- a/benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/ReplaceEscapeCharactersBenchmark.java
+++ b/benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/ReplaceEscapeCharactersBenchmark.java
@@ -1,0 +1,96 @@
+/**
+ * Copyright 2020 VMware, Inc.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.benchmark.core;
+
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.profile.GCProfiler;
+import org.openjdk.jmh.runner.Runner;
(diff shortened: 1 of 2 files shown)


shakuzen (commented):
Thank you for the pull request and JMH benchmark. While we're generally interested in making things faster with less memory allocation, this is not a particularly hot code path in general. Also importantly, this changes from a thread safe implementation to one with data race issues. That is maybe fine with the current usage in `InfluxMeterRegistry`, but `NamingConvention` isn't explicitly for single-threaded usage. We would need to carefully consider if we want to go the route of abandoning thread safety for performance gains and how we enforce that in our code and at least communicate it to any potential third-party usage. If we do end up using this, there are probably several other places with similar usage too.

shakuzen:
Thank you again for the pull request. I'm going to close this for now for the reason I stated before. While we want to improve speed and efficiency, I think we want to keep `InfluxNamingConvention` thread safe. Especially since so far we haven't had a reported issue from users that they've identified the current implementation causing overhead issues for them. Let us know if you have other suggestions on how we can improve this. We would be happy to consider them.

PR #2648, opened 2021-06-14
Fix TimeWindowMax for case of long period of inactivity

### LINKED ISSUE #2647: TimeWindowMax produces wrong values after long period of inactivity
**Code to reproduce:**
```java
        AtomicLong managedTime = new AtomicLong();
        Clock clock = new Clock() {
            @Override
            public long wallTime() {
                return managedTime.get();
            }

            @Override
            public long monotonicTime() {
                return managedTime.get();
            }
        };

        TimeWindowMax timeWindowMax = new TimeWindowMax(clock, 60_000, 3);
        timeWindowMax.record(32);
        System.out.println(timeWindowMax.poll()); // prints 32

        // emulate 12 hours of inactivity
        managedTime.set(TimeUnit.HOURS.toMillis(12));
        System.out.println(timeWindowMax.poll()); // prints 0 as expected

        timeWindowMax.record(666);
        System.out.println(timeWindowMax.poll()); // prints 0. 666 is missed because of bug

        timeWindowMax.record(13);
        System.out.println(timeWindowMax.poll()); // prints 0. both 666 and 13 are missed because of bug

        timeWindowMax.record(100500);
        System.out.println(timeWindowMax.poll()); // prints 0. All 666, 13 and 100500 are erased because of bug
    }
```

**Impact:**
The impact is not very critical because usually, a monitoring agent (like Telegraph or something else) touches your metrics on a regular basis, as a result, rotation happens often. 

**scenario to reproduce**

1. Imagine the network had been broken on the host,  as a result, nobody was touching Timer because both incoming requests from clients and incoming requests from the monitoring agent were not able to reach the application.
2. Network was fixed after several hours. User requests as well as requests from the monitoring agent now successfully reach the application.
3. We have zero values instead of maximums, and it is hard to understand that is not a bug of the monitoring agent or monitoring database. 

**Cause:**
image
After long period of inactivity ```rotate``` leaves ```lastRotateTimestampMillis``` in the far in the past.

vladimir-bukhtoyarov:
Fixes #2647

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowMax.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowMax.java
@@ -116,7 +116,8 @@ public class TimeWindowMax {
     }
 
     private void rotate() {
-        long timeSinceLastRotateMillis = clock.wallTime() - lastRotateTimestampMillis;
+        long wallTime = clock.wallTime();
+        long timeSinceLastRotateMillis = wallTime - lastRotateTimestampMillis;
         if (timeSinceLastRotateMillis < durationBetweenRotatesMillis) {
             // Need to wait more for next rotation.
             return;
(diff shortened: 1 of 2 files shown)


shakuzen (approved):
Thank you for the fix.

PR #2650, opened 2021-06-15
Fix DistributionSummary for Dynatrace

izeye:
This PR fixes `DistributionSummary` for Dynatrace.


shakuzen (approved):
Thanks for catching this. Could we add a test for this so it stays fixed going forward? Particularly I'm thinking about the v1/v2 split in main and how we want this to stay fixed in those and in the future.

izeye:
@shakuzen Thanks for the feedback! I updated to add a test in f2b6c65.

jonatan-ivanov (approved):

jonatan-ivanov:
Woah. :o Thank you very much for fixing this.

PR #2651, opened 2021-06-16
Fix error log for custom metric put failures

izeye:
This PR fixes an error log for custom metric put failures as the `Throwable` parameter was ignored prior to this change.

This PR also removes an unnecessary log level guard.


shakuzen (commented):
Looks good to me. Thanks.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you for fixing these. I deleted the curly braces since after your change the lambda become a one-liner.

Issue #877, opened 2018-09-26
Support meters with the same name but different set of tag keys in PrometheusMeterRegistry


j0xaf:
In connection with `PrometheusMeterRegistry` and the new `KafkaConsumerMetrics` (thanks to @jkschneider) I have a problem that metrics won´t register correctly because of `PrometheusMeterRegistry` complaining about metrics with same name but different set of key tags. It claims that metrics of same name are required in prometheus to have same set of tags.

I believe the error thrown at https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L357 is wrong. Prometheus perfectly accepts metrics of same name with disjunct set of tags (the metrics name itself is just a tag `__name__`). We have in our prometheus production instance metrics like `up{application="xyz", job="abc"}` and `up{application="abc", job="123", host="prometheus"}` coexisting without any issue whatsoever.

Example of exception text, stacktrace below:

> Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter containing tag keys [application, client_id]. The meter you are attempting to register has keys [application, client_id, topic].

Context of variables

```
id = MeterId{name='kafka.consumer.records.lag.max', tags=[ImmutableTag{key='application', value='heimdall'}, ImmutableTag{key='client.id', value='consumer-1'}, ImmutableTag{key='topic', value='probe_t'}]}
existingCollector = MeterId{name='kafka.consumer.records.lag.max', tags=[ImmutableTag{key='application', value='heimdall'}, ImmutableTag{key='client.id', value='consumer-1'}]}
```

Stacktrace:

```
collectorByName:348, PrometheusMeterRegistry (io.micrometer.prometheus)
newGauge:225, PrometheusMeterRegistry (io.micrometer.prometheus)
lambda$gauge$1:244, MeterRegistry (io.micrometer.core.instrument)
apply:-1, 4552198 (io.micrometer.core.instrument.MeterRegistry$$Lambda$373)
lambda$registerMeterIfNecessary$5:514, MeterRegistry (io.micrometer.core.instrument)
apply:-1, 491928400 (io.micrometer.core.instrument.MeterRegistry$$Lambda$338)
getOrCreateMeter:563, MeterRegistry (io.micrometer.core.instrument)
registerMeterIfNecessary:525, MeterRegistry (io.micrometer.core.instrument)
[...]

PR #2653, opened 2021-06-17
Issue 877: Add support for meters with the same name but different tag keys in prometheus registry

angry-cellophane:
I have an idea how to resolve this issue https://github.com/micrometer-metrics/micrometer/issues/877
and raising this PR to discuss it. 
The PR has a basic implementation + some simple tests to verify it works and makes sense in general.
I wanted to make sure that the solution makes sense first before going the full path and spending hours or days on making it bullet-proof and efficient. So any input is welcome.

So what this fix does:
1. Current version of micrometer-prometheus creates a collector for each pair of <metric name, list of tag keys>. Collector stores key values dynamically in a map.
2. Prometheus doesn't allow to register two collectors with the same metric name, but it doesn't limit the collector to one particular set of label names (or tag keys in micrometer). The main job of the collector is to return metric samples, which is a list of <tag key, tag value, value> basically. Prometheus supports metrics samples with the same metric name but different tag keys. 
3. So if the `MicrometerCollector` is changed to store tag keys dynamically like it does for tag values now, it can return tag keys and values in samples without any issue and prometheus will be able to parse it correctly.

What I did:
1. Added a new holder for tags to keep a list of keys and values separately for convenience.
2. Changed the map in `MicrometerCollector` to store the holders instead of tag values.
3. Changed some other dependent methods in `MicrometerCollector` to get tag keys from the holders.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
@@ -33,41 +33,87 @@ import static java.util.stream.Collectors.toList;
  * @author Johnny Lim
  */
 class MicrometerCollector extends Collector implements Collector.Describable {
+
+    static final class TagsHolder {
+        final List<String> keys;
+        final List<String> values;
+
+        private TagsHolder(List<String> keys, List<String> values) {
+            this.keys = keys;
+            this.values = values;
+        }
+
+        static TagsHolder from(List<Tag> tags) {
+            Objects.requireNonNull(tags, "tags");
+
+            List<String> keys = new ArrayList<>(tags.size());
+            List<String> values = new ArrayList<>(tags.size());
+
+            for (Tag tag : tags) {
+                keys.add(tag.getKey());
+                values.add(tag.getValue());
+            }
+
+            return new TagsHolder(
+                    Collections.unmodifiableList(keys),
+                    Collections.unmodifiableList(values)
+            );
+        }
+
+        public List<String> getKeys() {
+            return keys;
+        }
+
+        public List<String> getValues() {
+            return values;
+        }
+
+        @Override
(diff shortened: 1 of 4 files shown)


nhartner:
It would be great to see this merged. Is there any reason it hasn't been? Not sure who to tag on this but how about it @jonatan-ivanov?

crolopez:
Any updates? The solution seems simple and would provide great value.

jonatan-ivanov:
Please see this issue for more details: https://github.com/prometheus/client_java/issues/696
TL;DR: it possible but strongly discouraged.

zorba128:
Please take a look at [/prometheus/client_java/issues/901].
While I might agree that having two metrics with different sets of labels is somehow problematic, I think this all is deeper problem that its disallowed to have same metrics published by two different collectors (same set of labels, just different label values).
And this is core problem that makes new api really hard to use anywhere but on simple code as on code snippets...

tereigo:
Same problem here:
I've got a container-like app which can instantiate multiple instances of the business logic classes inside one jvm/process

Each instance exposes the same set of metrics

If I could set a label for each metric using the instance name and such metric objects are treated individually then it would work for this setup.

Otherwise I don't see any good solution apart from adding instance name into the metric name which is super ugly

shakuzen:
@angry-cellophane thank you for this pull request. The code has shifted since this was opened because of the Prometheus Java client releasing a new major version and our Prometheus modules being split between the 0.x and 1.x clients. Finally in the recently released Prometheus Java client 1.5.0 the restriction in the client itself around metrics with the same name but a different set of label names has been lifted. In #7168 I applied changes similar to the ones here but a bit different. Let us know if there's any feedback. I adapted the MicrometerCollectorTest code you added here to the new Prometheus client API and committed it as 9188fe37be7ae0ed8c21da0cf68f865569458c2f. With that, I think we can close this PR and the linked issue is now resolved; the changes will be released in Micrometer 1.17.0-M3.

PR #2656, opened 2021-06-17
Update dynatrace-metric-utils-java dependency

arminru:


implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/v2/DynatraceExporterV2Test.java
```diff
@@ -332,8 +332,8 @@ void toMeterLine() {
 
     @Test
     void gaugeWithInvalidNameShouldBeDropped() {
-        meterRegistry.gauge("~~~", 1.23);
-        Gauge gauge = meterRegistry.find("~~~").gauge();
+        meterRegistry.gauge("", 1.23);
+        Gauge gauge = meterRegistry.find("").gauge();
```
arminru, line 336: These tests had to be adapted because with the new normalization/sanitization rules followed by the library, this will be replaced with a valid metric identifier. An empty string will still provoke an invalid one.

dependencies.gradle
```diff
@@ -2,7 +2,7 @@ def VERSIONS = [
         'ch.qos.logback:logback-classic:1.2.+',
         'colt:colt:1.2.0',
         'com.amazonaws:aws-java-sdk-cloudwatch:latest.release',
-        'com.dynatrace.metric.util:dynatrace-metric-utils-java:latest.release',
+        'com.dynatrace.metric.util:dynatrace-metric-utils-java:1.0.+',
```
jonatan-ivanov, line 5: I think we should keep this at `latest.release` so new minor versions will pick-up the latest version of the library automatically, what do you think?
(Patch versions are dependency locked so that releasing a new version of the lib won't break them.)

jonatan-ivanov: Btw originally this was `0.3.+`, I just merged a PR yesterday that changed it to `latest.release`.

jonatan-ivanov:
Thank you very much!

dependencies.gradle
arminru: @jonatan-ivanov In my understanding `latest.release` would also accept a breaking `2.0.0` release, if any, so maybe `1.+` would be a better choice if you'd like to pick up minor patches automatically.

jonatan-ivanov: @arminru Yes, it will, but it will only effect the non-released `main` branch, everything else is dependency locked.
If this happens we can either move to 2.0.0 or use 1.+. I guess this could be one reason why most of the dependencies in that file are using `latest.release`

PR #2658, opened 2021-06-18
Update Spring Boot example config for Dynatrace

arminru:


arminru:
Pending on https://github.com/spring-projects/spring-boot/pull/26258.

jonatan-ivanov (commented):
I'm thinking if we should keep the original V1 samples too:
- Rename `DynatraceSample` to `DynatraceV1Sample`
- Reaname `application-dynatrace.yml` to `application-dynatrace-v1.yml`
- Modify the profile in `DynatraceV1Sample` to `dynatrace-v1`
- Copy `DynatraceV1Sample` to `DynatraceV2Sample` and add profile `dynatrace-v2`
- Use the config in this PR and create `application-dynatrace-v2.yml`

WDYT?

arminru:
@jonatan-ivanov These examples are intended as a starting point for new users, right? We should encourage them to use the new v2 registry/API right away rather than starting with the old one at this point. If that would be fine with you, I'd prefer changing the example rather than having both in there. Otherwise your suggested changes would sound alright to me. Let me know what you think. Also, @pirgeo will take over this PR as I'm on vacation.

jonatan-ivanov:
@arminru Have a great vacation! :)

@arminru @pirgeo
I think the main intention of the examples is to show a working demo and to help the users.
Since the V1 registry is still available and it is the default, I think it would be a good idea to keep the V1 example around. 

I totally get the intention of encouraging V2. What do you think about putting a comment both to the v1 `.properties` and `.java` files that is saying something like the Dynatrace Team recommends using the V2 API?

We can also put a `WARN` log into the V1 example app.

pirgeo:
@jonatan-ivanov I played around with that for a bit, but hit a bump: It seems to be forbidden by the `checkstyle` command to import `org.slf4j.*` in the example classes - Is there a way of printing a warning from the example app?

jonatan-ivanov:
Ouch, I see:
```
[ant:checkstyle] [ERROR] .../samples/DynatraceSampleV1.java:19:1: Illegal import - org.slf4j.LoggerFactory. [IllegalImport]
```

A couple of alternatives:
1. `InternalLoggerFactory.getInstance(DynatraceSampleV1.class)`
The problem with this is that users should not use this, so I would not advertise in that sample
2. `System.out.println();` or `System.err.println();`
This actually is not as bad as it sounds since you don't need any features that a logger has and IDEs show the error stream in different color so it will be more apparent (you can also put it into some frame: `---`, `===`)
3. `//CHECKSTYLE:OFF` and `//CHECKSTYLE:ON` 😈 , see: [`SuppressionCommentFilter`](https://checkstyle.sourceforge.io/config_filters.html#SuppressionCommentFilter_Properties)

Here's an example for the third option (I [reverted](https://github.com/jonatan-ivanov/micrometer/commits/dynatrace-example-review) your log removal and merged back main): https://github.com/jonatan-ivanov/micrometer/commit/64e588a4a7ad4d8c12d03e5624f6c56f2c8a4754

I think I would just go with `System.err.println();` because that is the most apparent and the simplest.

arminru:
@jonatan-ivanov Your suggestions sound good but for now we decided to be less explicit and just added comments indicating that v2 is preferred rather than printing warnings right away. We can always make it more explicit in future.

jonatan-ivanov:
Sounds good, thank you for adding the example.

PR #2673, opened 2021-06-30
Add response code to log message

### LINKED ISSUE #2669: Add response status to error logs in DynatraceMeterRegistry
It would be useful for troubleshooting, see: https://github.com/micrometer-metrics/micrometer/issues/2667

This should be added from 1.5.x. The V2 implementation in 1.8 has already have this information, we can do the same in V1, see: https://github.com/micrometer-metrics/micrometer/blob/7afca0af04e0f1c136b4419d4a8b12cbd9aea2c7/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java#L271

pirgeo:
Resolves #2669


jonatan-ivanov (changes requested):
Could you please make this change against the 1.5.x branch?

jonatan-ivanov (approved):

jonatan-ivanov:
@pirgeo Thank you very much! :)

PR #2677, opened 2021-06-30
Log verification

jonatan-ivanov:


micrometer-test/src/main/java/io/micrometer/core/util/internal/logging/InternalMockLogger.java
```diff
@@ -0,0 +1,277 @@
...
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.util.internal.logging;
```
jonatan-ivanov, line 16: Is `internal` a good place for these (they are in the same package as the interfaces they implement)?

shakuzen: Sounds like a good place to me.

micrometer-test/src/test/java/io/micrometer/core/util/internal/logging/InternalMockLoggerFactoryTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Jonatan Ivanov
+ */
+class InternalMockLoggerFactoryTest {
+    @BeforeAll
+    static void setUpClass() {
+        InternalLoggerFactory.setDefaultFactory(new InternalMockLoggerFactory());
```
shakuzen, line 29: Do we not need to clean up setting this global state? It seems like this could affect other tests.

shakuzen (approved):
Looks good besides the potential leaking of global state change commented on in the review.

micrometer-test/src/test/java/io/micrometer/core/util/internal/logging/InternalMockLoggerFactoryTest.java
jonatan-ivanov: Instead of restoring the original one, I eliminated this completely from the tests by getting the logger directly from the `InternalMockLoggerFactory` through a convenience method. Also, I realized that this can cause issues during parallel test execution so I added support to inject the mock logger in a safe way.

Issue #2683, opened 2021-07-06
Dropwizard Timer instances always created with system clock


slovdahl:
**Describe the bug**

Whenever a Dropwizard `com.codahale.metrics.Timer` instance is created in `io.micrometer.core.instrument.dropwizard.DropwizardMeterRegistry`, the underlying Dropwizard instance is always created with `com.codahale.metrics.Clock#defaultClock()` as the backing clock instance, even if the clock is overridden with a custom implementation in Micrometer.

**Environment**

 - Micrometer version 1.5.15 (the relevant code looks the same in 1.7.1)
 - Micrometer registry Prometheus + JMX
 - OS: Linux
 - Java version: 1.8.0_292

**To Reproduce**
How to reproduce the bug:

Failing test: #2684 

**Expected behavior**

The overridden clock is used in the underlying `com.codahale.metrics.Timer` which makes it easier to write tests.

**Additional context**

We have tests that assert on the metrics exported via JMX, and currently we have to insert 5 seconds of sleep after every timer update because internally `com.codahale.metrics.Timer` (or actually `com.codahale.metrics.Meter`) uses a hard-coded tick interval of 5 seconds. This means that changes to timers are delayed with 5 seconds in JMX output.

![image](https://user-images.githubusercontent.com/1417619/124602465-8e007f00-de71-11eb-86e6-ee769cc21290.png)

![image](https://user-images.githubusercontent.com/1417619/124602492-948ef680-de71-11eb-967c-b36e8ee465a1.png)

PR #2684, opened 2021-07-06
Use Micrometer clock with Dropwizard Timer

slovdahl:
Related to #2683.


slovdahl:
Pushed a second commit that fixes the issue. Tests passing locally at least.

PR #2685, opened 2021-07-06
Pin jOOQ to 3.14.+

izeye:
This PR pins jOOQ to 3.14.+ as jOOQ 3.15.0 isn't compatible with the current `main`.


jonatan-ivanov:
Error log: https://ge.micrometer.io/s/2anfkkrdbi7bg/console-log?task=:micrometer-core:compileJava

jonatan-ivanov:
Thanks for fixing this! :)

PR #2687, opened 2021-07-07
Replace okhttp3.internal.http.HttpMethod.requiresRequestBody()

izeye:
This PR replaces `okhttp3.internal.http.HttpMethod.requiresRequestBody()` as it's an internal API of OkHttp.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/ipc/http/OkHttpSender.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/ipc/http/OkHttpSender.java
@@ -49,24 +49,36 @@ public class OkHttpSender implements HttpSender {
         }
 
         byte[] entity = request.getEntity();
-        String requestMethod = request.getMethod().toString();
+        Method method = request.getMethod();
+        String methodValue = method.toString();
         if (entity.length > 0) {
             String contentType = request.getRequestHeaders().get("Content-Type");
             MediaType mediaType = contentType != null
                     ? MediaType.get(contentType + "; charset=utf-8")
                     : MEDIA_TYPE_APPLICATION_JSON;
             RequestBody body = RequestBody.create(entity, mediaType);
-            requestBuilder.method(requestMethod, body);
+            requestBuilder.method(methodValue, body);
         } else {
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/ipc/http/OkHttpSender.java
```diff
@@ -49,24 +49,36 @@ public Response send(Request request) throws Throwable {
...
         }
 
         okhttp3.Response response = client.newCall(requestBuilder.build()).execute();
         return new Response(response.code(), response.body() == null ? null : response.body().string());
     }
+
+    private static boolean requiresRequestBody(Method method) {
+        switch (method) {
+            case POST:
+            case PUT:
```
jonatan-ivanov, line 77: Can we add `PATCH` too?
Maybe also `PROPPATCH` and `REPORT`?
see: [`HttpMethod`](https://github.com/RunningTheSnail/Okhttp/blob/84cbd2dae1708967c3b1bd71ccecdecefd6354f5/okhttp/src/main/java/okhttp3/internal/http/HttpMethod.java#L27-L33)

jonatan-ivanov (changes requested):

micrometer-core/src/main/java/io/micrometer/core/ipc/http/OkHttpSender.java
izeye: @jonatan-ivanov `HttpSender` doesn't support `PATCH` or the other methods:

https://github.com/micrometer-metrics/micrometer/blob/202132323a2a684acff0eb81dcbe7f4a51643e6f/micrometer-core/src/main/java/io/micrometer/core/ipc/http/HttpSender.java#L336-L338

jonatan-ivanov: Yepp, you are right. We can future-proof this but I think we are ok.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you! :)

PR #2689, opened 2021-07-07
Polish logging test support

izeye:
This PR polishes logging test support a bit. I tried to shorten their names by dropping the "Internal" prefix for brevity, but let me know if the previous names are preferred.

### DIFF
--- a/micrometer-test/src/main/java/io/micrometer/core/util/internal/logging/MockLogger.java
+++ b/micrometer-test/src/main/java/io/micrometer/core/util/internal/logging/MockLogger.java
@@ -15,7 +15,7 @@
  */
 package io.micrometer.core.util.internal.logging;
 
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Queue;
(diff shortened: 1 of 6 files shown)


micrometer-test/src/main/java/io/micrometer/core/util/internal/logging/MockLogger.java
```diff
@@ -28,25 +28,25 @@
...
+    MockLogger(String name) {
         this.name = name;
     }
 
     /**
      * @return The recorded {@link LogEvent} instances, in descending order by age (the oldest is the first one).
      */
     public List<LogEvent> getLogEvents() {
-        return Arrays.asList(logEvents.toArray(new LogEvent[0]));
+        return new ArrayList<>(logEvents);
```
jonatan-ivanov, line 49: The difference between the two is that as far as I remember the first one is immutable but I think we are good anyways.

jonatan-ivanov (approved):

jonatan-ivanov:
Tank you! :)

PR #2692, opened 2021-07-08
Rename OneAgentMetadataEnrichment to DynatraceMetadataEnrichment

pirgeo:
This renames the config property from OneAgent metadata to Dynatrace metadata since the metadata retrieved is not restricted to only the OneAgent but might come from other components in the future.

I will update both https://github.com/spring-projects/spring-boot/pull/26258 and https://github.com/micrometer-metrics/micrometer-docs/pull/160 as soon as this is merged.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
@@ -71,7 +71,7 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
                 .withPrefix(config.metricKeyPrefix())
                 .withDefaultDimensions(parseDefaultDimensions(config.defaultDimensions()));
 
-        if (config.enrichWithOneAgentMetadata()) {
+        if (config.enrichWithDynatraceMetadata()) {
             factoryBuilder.withOneAgentMetadata();
         }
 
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -71,7 +71,7 @@ public DynatraceExporterV2(DynatraceConfig config, Clock clock, HttpSender httpC
                 .withPrefix(config.metricKeyPrefix())
                 .withDefaultDimensions(parseDefaultDimensions(config.defaultDimensions()));
 
-        if (config.enrichWithOneAgentMetadata()) {
+        if (config.enrichWithDynatraceMetadata()) {
```
jonatan-ivanov, line 74: Do you also want to replace this (line#94)?
>Potential misconfiguration detected: Token is provided, but the endpoint is set to the local OneAgent endpoint,

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/DynatraceConfigTest.java
```diff
@@ -117,7 +117,7 @@ void testV2Defaults() {
         assertThat(config.uri()).isSameAs(DynatraceMetricApiConstants.getDefaultOneAgentEndpoint());
         assertThat(config.metricKeyPrefix()).isEmpty();
         assertThat(config.defaultDimensions()).isEmpty();
-        assertThat(config.enrichWithOneAgentMetadata()).isTrue();
+        assertThat(config.enrichWithDynatraceMetadata()).isTrue();
```
jonatan-ivanov, line 120: Do you also want to rename `testOneAgentEndpointWithDifferentPort`?

pirgeo:
Thanks for taking a look! There are two different features here which happen to have similar names:
- The metadata enrichment, which we renamed in this PR, does not only work with the OneAgent but also with other metadata sources (hence the renaming).
- The local OneAgent endpoint requires a OneAgent, therefore the wording stayed the same for this component.

jonatan-ivanov (approved):

jonatan-ivanov:
@pirgeo Makes sense, thank you for the fix! :)

PR #2696, opened 2021-07-13
Suppress `IllegalImport` check for samples

izeye:
This PR changes to suppress the `IllegalImport` check for the samples.

See https://github.com/micrometer-metrics/micrometer/pull/2658#issuecomment-872848875


jonatan-ivanov:
I was thinking if it's valid or not but actually, why would not it be? :)
Thanks for the fix!

PR #2713, opened 2021-07-22
Update Dynatrace samples for Spring Boot autoconfiguration

pirgeo:
Due to the changes introduced in https://github.com/spring-projects/spring-boot/pull/26258, we updated the samples for the Dynatrace meter registry. We also updated the docs in https://github.com/micrometer-metrics/micrometer-docs/pull/160.


jonatan-ivanov:
Thank you for adding these! :)

PR #2727, opened 2021-08-03
Use Dynatrace v2 export as default unless deviceId is set

pirgeo:
This PR introduces minor code changes to mimic the behavior of the Spring Boot integration for Dynatrace Micrometer Metrics. There, we changed the logic to fall back to the v2 exporter as a default, unless the `deviceId` property is set (https://github.com/spring-projects/spring-boot/pull/26258#issuecomment-880636873). This property is required for the v1 export, therefore all integrations currently using v1 will continue to work (since they specify the `deviceId` explicitly).

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -17,6 +17,7 @@ package io.micrometer.dynatrace;
 
 import com.dynatrace.metric.util.DynatraceMetricApiConstants;
 import io.micrometer.core.instrument.config.validate.Validated;
+import io.micrometer.core.instrument.config.validate.ValidationException;
 import io.micrometer.core.instrument.step.StepRegistryConfig;
 import io.micrometer.core.lang.Nullable;
 
@@ -34,7 +35,6 @@ import static io.micrometer.core.instrument.config.validate.PropertyValidator.*;
  * @since 1.1.0
  */
 public interface DynatraceConfig extends StepRegistryConfig {
-
     @Override
     default String prefix() {
         return "dynatrace";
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
```diff
@@ -129,7 +137,7 @@ default boolean enrichWithDynatraceMetadata() {
                                                 return checkAll(this,
                                                         checkRequired("apiToken", DynatraceConfig::apiToken),
                                                         checkRequired("uri", DynatraceConfig::uri),
-                                                        checkRequired("deviceId", DynatraceConfig::deviceId),
+                                                        check("deviceId", DynatraceConfig::deviceId).andThen(Validated::nonBlank),
```
shakuzen, line 140: We can make more complex validation rules to ensure we have required configuration set for v1 or v2. See, for example, the InfluxConfig validation where it also supports v1/v2 and defaults to one or the other depending on other configuration being set.
https://github.com/micrometer-metrics/micrometer/blob/83b7b8c323e741f9e469354afb9843b5044a9ed6/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxConfig.java#L219-L221

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
```diff
@@ -82,8 +81,17 @@ default String group() {
...
+        if (!deviceId().isEmpty()) {
+            // if deviceId is not empty, use v1 as default.
+            defaultVersion = DynatraceApiVersion.V1;
+        }
+
+        // If a device id is specified, use v1 as default. If it is not, use v2.
+        // The version can be overwritten explicitly when creating a MM config
+        // For Spring Boot, v1 is automatically chosen when the device id is set.
+        return getEnum(this, DynatraceApiVersion.class, "apiVersion")
+                .orElse(defaultVersion);
```
jonatan-ivanov, line 94: What do you think about this (static imports `V1` and `V2`):
`.orElse(deviceId().isEmpty() ? V2 : V1);`

jonatan-ivanov (approved):

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
shakuzen: Oh. Perils of only looking at the diff in the GitHub UI. Ignore me, then.

jonatan-ivanov (approved):

Issue #2721, opened 2021-07-29
Upgrade to reactor-netty 1.0.10


shakuzen:
We will need to switch to snapshots to resolve #792. This issue is to ensure we switch to the release version once available, before the 1.8.0-M2 release.

PR #2732, opened 2021-08-11
Upgrade from snapshots to reactor 2020.0.10

shakuzen:
Note that `2020.0.+` was resolving 2020.0.9 for me locally rather than the newer 2020.0.10. Surely Gradle isn't doing alphanumeric sorting rather than standard version number sorting?

Resolves gh-2721

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -91,7 +91,7 @@ subprojects {
           runtimeOnly version
         }
       }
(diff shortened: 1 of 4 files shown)


shakuzen:
I'm looking into a failure of one of the statsd tests on my macbook before merging.

jonatan-ivanov (approved):

PR #2745, opened 2021-08-12
Add support for Datadog distributions

cyberdelia:
This uses Datadog distributions when `DistributionStatisticConfig` is set to publish percentile histograms for a metric, which allows to use them only for specific metrics.

Related to #1056, #1859.


jkschneider (approved):
LGTM. Thank you much @cyberdelia

shakuzen:
This only adds support for Datadog's distribution type with our DistributionSummary. Am I missing any reason we wouldn't want this to apply to our Timer type also?

PR #2747, opened 2021-08-14
Add support for multiple paths to DiskSpaceMetrics

bono007:
### What
This code proposal will allow consumers to add disk space metrics to multiple paths.

### Why
SpringBoot auto-configures a `DiskSpaceMetrics` and also has a `DiskSpaceHealthContributor`. There is a desire to keep their [configurability in parity](https://github.com/spring-projects/spring-boot/issues/27306).

The health contributor has a [request to support multiple paths](https://github.com/spring-projects/spring-boot/issues/18359). The first attempted proposal was closed quite some time ago due to concern of breaking actuator output compatibility (root disk threshold values vs. child path thresholds etc.). I intend to get a 2nd attempt going now that a breaking change should be ok in a minor version bump (2.6).

This change to `DiskSpaceMetrics` will be followed with an update to [JvmMetricsAutoConfiguration](https://github.com/spring-projects/spring-boot/blob/18349dc556f4c05fe68b36dc79a6be0e0b303a21/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java#L75) to allow configuration of multiple paths.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/DiskSpaceMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/DiskSpaceMetrics.java
@@ -15,6 +15,7 @@
  */
 package io.micrometer.core.instrument.binder.system;
 
+import io.micrometer.core.annotation.Incubating;
 import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
@@ -25,6 +26,8 @@ import io.micrometer.core.lang.NonNullApi;
 import io.micrometer.core.lang.NonNullFields;
 
 import java.io.File;
+import java.util.Collections;
+import java.util.List;
 
 import static java.util.Collections.emptyList;
 
@@ -33,28 +36,40 @@ import static java.util.Collections.emptyList;
  *
  * @author jmcshane
  * @author Johnny Lim
+ * @author Chris Bono
  * @since 1.8.0
  */
 @NonNullApi
 @NonNullFields
 public class DiskSpaceMetrics implements MeterBinder {
+
     private final Iterable<Tag> tags;
-    private final File path;
-    private final String absolutePath;
+    private final List<File> paths;
 
     public DiskSpaceMetrics(File path) {
         this(path, emptyList());
     }
 
+    public DiskSpaceMetrics(List<File> paths) {
+        this(paths, emptyList());
+    }
+
     public DiskSpaceMetrics(File path, Iterable<Tag> tags) {
-        this.path = path;
-        this.absolutePath = path.getAbsolutePath();
+        this(Collections.singletonList(path), tags);
+    }
+
+    public DiskSpaceMetrics(List<File> paths, Iterable<Tag> tags) {
+        this.paths = paths;
         this.tags = tags;
     }
 
     @Override
     public void bindTo(MeterRegistry registry) {
-        Iterable<Tag> tagsWithPath = Tags.concat(tags, "path", absolutePath);
+        paths.forEach(path -> bindWithPath(registry, path));
+    }
+
+    private void bindWithPath(MeterRegistry registry, File path) {
+        Iterable<Tag> tagsWithPath = Tags.concat(tags, "path", path.getAbsolutePath());
         Gauge.builder("disk.free", path, File::getUsableSpace)
                 .tags(tagsWithPath)
                 .description("Usable space for path")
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/DiskSpaceMetrics.java
```diff
@@ -34,28 +36,40 @@
...
 
     public DiskSpaceMetrics(File path) {
         this(path, emptyList());
     }
 
+    public DiskSpaceMetrics(List<File> paths) {
+        this(paths, emptyList());
+    }
+
     public DiskSpaceMetrics(File path, Iterable<Tag> tags) {
```
bono007, line 57: I was hoping to not add 2 new constructors by editing the existing ones and changing their single `path` param to a varargs path param. That approach came up short w/ this particular `<init>` as varargs has to be the last in param list.  

I am open to suggestions. I think a builder would be overkill. What are thoughts on deprecation of the single path `<init>` variants?

shakuzen (commented):
If a user wants disk space metrics for multiple paths, wouldn't binding multiple `DiskSpaceMetrics` make sense? What is the benefit of the proposed changes over that approach?
Not that it necessarily matters, but I would think a single path is sufficient for most users.

bono007:
> If a user wants disk space metrics for multiple paths, wouldn't binding multiple `DiskSpaceMetrics` make sense? What is the benefit of the proposed changes over that approach?
> Not that it necessarily matters, but I would think a single path is sufficient for most users.

The main reason of going w/ this approach is to keep the auto-config simple by not having to dynamically register N beans as the paths will be config property driven. 

IOW, doing this
```java
	@Bean
	@ConditionalOnMissingBean
	public DiskSpaceMetrics diskSpaceMetrics() {
		return new DiskSpaceMetrics(props.getListOfPaths());
	}
```
is much simpler than creating an ApplicationContextInitializer (or other means) and registering N of the metrics dynamically via something like 
```java
props.getListOfPaths().forEach(path -> {
   context.registerBean("diskSpaceMetrics" + i, DiskSpaceMetrics.class,
                () -> new DiskSpaceMetrics(new File(path)));
}

```

bono007:
@shakuzen I rebased to pickup the move to the `system` package.

wilkinsona:
@bono007 Are you interested in reworking this to support multiple paths?  In light of https://github.com/spring-projects/spring-boot/issues/27306#issuecomment-903907394, that can be done without considering the disk space health indicator.  I think we'd probably end up with a `MetricBinder` implementation that consumes the paths. When called, it would create one `DiskSpaceMetrics` instance per path and binds it to the registry.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/DiskSpaceMetrics.java
```diff
@@ -33,28 +35,40 @@
...
  * @since 1.8.0
  */
 @NonNullApi
 @NonNullFields
 public class DiskSpaceMetrics implements MeterBinder {
+
     private final Iterable<Tag> tags;
-    private final File path;
-    private final String absolutePath;
+    private final List<File> paths;
```
shakuzen, line 46: We could be less specific than `List` since we don't care about order. How about `Collection`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/DiskSpaceMetrics.java
```diff
@@ -33,28 +35,40 @@
...
 
     public DiskSpaceMetrics(File path) {
         this(path, emptyList());
     }
 
+    public DiskSpaceMetrics(List<File> paths) {
```
shakuzen, line 52: While mostly intuitive, a JavaDoc would be nice. I know we're missing one on the existing constructors - I can follow-up with adding that separately. In the case of this constructor, the paths should be to different partitions, since there's no point getting the same information about the same partition for different paths on that partition.

shakuzen (approved):
I think we're okay to accept this. I saw in the Spring Boot pull request another proposal was made that wouldn't need these changes. Let us know.

bono007:
> I think we're okay to accept this. I saw in the Spring Boot pull request another proposal was made that wouldn't need these changes. Let us know.

Thanks @shakuzen. Yeh, we are going to handle the multi-paths on the Spring Boot side in a MeterBinder impl that will bind the N DiskSpaceMetrics (your original thought about the impl).

bono007:
@shakuzen if you think this could be valuable for others then I don't mind making the suggested changes and re-opening. Just let me know.

jonatan-ivanov:
For the sake of discoverability, the involved Spring Boot PR that will create `DiskSpaceMetrics` instances: https://github.com/spring-projects/spring-boot/pull/27660

shakuzen:
> if you think this could be valuable for others then I don't mind making the suggested changes and re-opening. Just let me know.

We haven't had a request for it until this. That could partially be due to DiskSpaceMetrics not being auto-configured in Spring Boot until recently. If other users are interested, we can re-visit this. I'd rather see requests from users for this than me assume it's something many users want.

PR #2749, opened 2021-08-15
Add support for Jersey3 which is based on Jakarta EE 9 (#2748)

### LINKED ISSUE #2748: Support for Jersey 3 server
Jersey 3 has moved to using Jakarta EE 9. Basically most things under javax.* moved to jakarta.*.
Any plans to support this in micrometer? Maybe micrometer-jersey3?

alexsatori:
Basically copy pasted files from jersey2, changed javax.ws.rs.* to jakarta.ws.rs*
And some voodoo for the dependencies as jersey2 and jersey3 have different incompatible deps. 

Resolves #2748

### DIFF
--- a/config/checkstyle/checkstyle-suppressions.xml
+++ b/config/checkstyle/checkstyle-suppressions.xml
@@ -9,4 +9,5 @@
 	<suppress files="samples[\\/].+" checks="IllegalImport" />
 	<suppress files="LogbackMetricsAutoConfiguration" checks="IllegalImport" />
 	<suppress files="io[\\/]micrometer[\\/]jersey2[\\/]server[\\/].+" checks="SpringJUnit5" />
+	<suppress files="io[\\/]micrometer[\\/]jersey3[\\/]server[\\/].+" checks="SpringJUnit5" />
 </suppressions>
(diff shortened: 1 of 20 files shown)


shakuzen (commented):
Thank you for the pull request. I suspect this approach of a micrometer-jersey3 module is probably the best route to go. It unfortunately means we have to make sure bugs common to both jersey modules get fixed in both. I'm wondering if it will make sense to drop the jersey2 module for Micrometer 2.0.
@mweirauch would you mind reviewing this if you have time?

mweirauch:
@shakuzen I am currently on vacation and I am not sure I will be able to do a thorough review afterwards right away. I don't have any - neither testing nor production - projects which are using the Jersey3 framework.

shakuzen:
Thanks for the quick response anyways @mweirauch and no worries.

stefanbirkner:
> I'm wondering if it will make sense to drop the jersey2 module for Micrometer 2.0.
Micrometer should continue to support Jersey 2. Upgrading from Jersey 2 to 3 is not a simple step because javax.* classes are replaced by jakarta.* classes. Unfortunately some libraries like Guice don't work with jakarta.* classes which makes it harder to upgrade applications from Jersey 2 to Jersey 3.

shakuzen:
Superseded by #2766

shakuzen:
> Micrometer should continue to support Jersey 2.

My comment was in the context of moving the jersey instrumentation to micrometer-core that supports both Jersey 2 and Jersey 3. Not publishing the micrometer-jersey2 module would be because it becomes redundant with micrometer-core; not because we stop supporting Jersey 2.

PR #2755, opened 2021-08-18
Revert "Upgrade to Gradle Wrapper 7.2"

izeye:
Reverts micrometer-metrics/micrometer#2753

Unfortunately, with the change and OpenJDK 11.0.11, the `:micrometer-core:compileJava` task fails as follows:

```
➜  micrometer git:(1.5.x) ./gradlew clean :micrometer-core:compileJava --no-build-cache --stacktrace
...
> Task :micrometer-core:compileJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':micrometer-core:compileJava'.
> error: release version 8 not supported

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':micrometer-core:compileJava'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:188)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:186)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:174)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
[...]

PR #2756, opened 2021-08-18
Add a build job on JDK 11

izeye:
This is similar to #2480, but for JDK 11.

See gh-2480
See gh-2755


izeye:
The build on JDK 11 is expected to fail, but it has passed due to the build cache as follows:

```
...
> Task :micrometer-core:compileJava FROM-CACHE
...
```

See https://app.circleci.com/pipelines/github/izeye/micrometer/739/workflows/69ef6975-c18a-4ac8-aee3-ec490f8d24c2/jobs/1316

jonatan-ivanov:
This is weird, that task must have missed the cache because of the different JDK version.

PR #2759, opened 2021-08-20
Update samples to use Spring Boot 2.6.0-M2

izeye:
This PR updates samples to use Spring Boot 2.6.0-M2.


jonatan-ivanov:
Thank you! :)

PR #2764, opened 2021-08-27
Use Tag#compareTo to detect duplicates

datibbaw:
Addresses #1432 by using the `compareTo()` method from the `Tag` interface, defined [here](https://github.com/micrometer-metrics/micrometer/blob/913f84e41f315c07ad7381001c9596358aa1e656/micrometer-core/src/main/java/io/micrometer/core/instrument/Tag.java#L33-L35)
```java
default int compareTo(Tag o) {
    return getKey().compareTo(o.getKey());
}
```
The above is already marked as overridable and implements the existing dedup logic almost exactly:
```java
for (int i = 0; i < n - 1; i++)
    if (!tags[i].getKey().equals(tags[i + 1].getKey()))
        tags[j++] = tags[i];
```
In this way, developers can extend `ImmutableTag`, which implements `Tag`, with their own class to override the `compareTo()` method like so.

```java
@Override
public int compareTo(Tag o) {
  return Comparator.comparing(Tag::getKey).thenComparing(Tag::getValue).compare(this, o);
}
```
To be clear, this isn't meant to be made available by default; because of its limited applicability, developers take on the risk of passing the appropriate values to the corresponding backend.


datibbaw:
@shakuzen Could you have another look at this? The code change itself doesn't do anything new, rather it uses the existing `Tag` interface to accomplish the same functionality while opening the implementation up for extension where necessary. I've added an example of this in the tests section.

datibbaw:
@shakuzen Hoping to get an answer to my earlier question; even though we use only Datadog in our organisation, it's the abstraction of metrics that we like to keep using.

shakuzen:
Sorry for the lack of response. I'll take a closer look at this again next week.

PR #2766, opened 2021-08-30
Support Jersey 3

stefanbirkner:
Jersey 3 is using classes from the jakarta.* packages instead of
javax.*. The Jersey 2 code can be used with Jersey 3 if we don't import
these classes.

This commit avoids importing javax.* classes by inlining code, not
storing variables and checking for NotFoundException by it's class name.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -18,7 +18,7 @@
 
         <!-- Imports -->
         <module name="IllegalImportCheck" >
-            <property name="illegalPkgs" value="com.google.common.(?!cache).*,org.apache.commons.text.*,org.slf4j.*,org.jetbrains.*,jdk.internal.jline.internal.*,reactor.util.annotation.*,org.checkerframework.checker.*"/>
+            <property name="illegalPkgs" value="com.google.common.(?!cache).*,org.apache.commons.text.*,org.slf4j.*,org.jetbrains.*,jdk.internal.jline.internal.*,reactor.util.annotation.*,org.checkerframework.checker.*,javax.ws.*"/>
             <property name="illegalClasses" value="org\.assertj\.core\.api\.Java6Assertions\..*"/>
             <property name="regexp" value="true"/>
         </module>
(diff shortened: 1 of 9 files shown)


mweirauch:
See also #2748 and #2749

shakuzen:
Thanks for the alternate solution to supporting Jersey 3. @alexsatori what do you think about this?

alexsatori:
It should work, the downside is that it leaves all the jersey2 dependencies. It seems though that they can be moved to "test" only.

mweirauch:
Oh, forgot to mention: I think it might also have been the case there is a separate jersey2 module because at that time the Junit5 support from Jersey2 was lacking.

Making Jersey3 support part of micrometer-core is a bit overkill, or? (Then we'd likely need to duplicate code, though.)

shakuzen:
> because at that time the Junit5 support from Jersey2 was lacking

Unfortunately it is still lacking https://github.com/eclipse-ee4j/jersey/issues/3662. That's not necessarily a blocker, though.

> Making Jersey3 support part of micrometer-core is a bit overkill, or? (Then we'd likely need to duplicate code, though.)

I'm not opposed to the idea of migrating the jersey instrumentation to core except for the duplication, as long as there isn't any other conflict. It'd be easier to do in Micrometer 2.0, where we can more reasonably stop publishing the micrometer-jersey2 module.

config/checkstyle/checkstyle.xml
```diff
@@ -18,7 +18,7 @@
 
         <!-- Imports -->
         <module name="IllegalImportCheck" >
-            <property name="illegalPkgs" value="com.google.common.(?!cache).*,org.apache.commons.text.*,org.slf4j.*,org.jetbrains.*,jdk.internal.jline.internal.*,reactor.util.annotation.*,org.checkerframework.checker.*"/>
+            <property name="illegalPkgs" value="com.google.common.(?!cache).*,org.apache.commons.text.*,org.slf4j.*,org.jetbrains.*,jdk.internal.jline.internal.*,reactor.util.annotation.*,org.checkerframework.checker.*,javax.ws.*"/>
```
shakuzen, line 21: This should make sure we don't mistakenly add back imports on these classes we got rid of from main code.

PR #2770, opened 2021-09-07
Kafka: Always get the current Metric instance from the metricsSupplier

### LINKED ISSUE #2726: Kafka Metrics become NaN after a while (possibly after rebalance)
**Describe the bug**
I noticed that in my application, the `kafka_consumer_fetch_manager_records_lag` metric is displayed as `NaN` after a while for some partitions. A rebalance seems to be one possible trigger for this to happen. After some debugging I think I have found the cause:

When the `KafkaMetrics` binder binds metrics, it checks if there already is a metric of the same name with the same tags. If the metric exists, it is not bound again. Kafka, on the other hand, sometimes provides a new `Metric` instance for the same name and tags. I could reproduce this locally by restarting the Kafka server, but in our real-life environment it also seems to happen after a rebalance.

This alone would probably cause outdated metrics to be displayed, because the `Gauge` instance references an old `Metric` that is probably never updated. But because the `DefaultGauge` uses a `WeakReference` for the metric, it loses this reference when the metric is garbage-collected. Then it starts returning `NaN` (see `DefaultGauge#value()`).

image

**Environment**
 - Micrometer version [e.g. 1.7.1]: 1.7.2
 - Micrometer registry [e.g. prometheus]: prometheus
 - OS: [e.g. macOS]: macOS
 - Java version: [e.g. output of `java -version`]:
    ```
    openjdk 13.0.2 2020-01-14
    OpenJDK Runtime Environment (build 13.0.2+8)
    OpenJDK 64-Bit Server VM (build 13.0.2+8, mixed mode, sharing)
   ```

**To Reproduce**
How to reproduce the bug:

It's a bit hard to reproduce because it depends on garbage collection and I'm not sure under what circumstances exactly Kafka changes the metric instances, but I could confirm it using a new Spring Boot 2.5.3 application with Actuator, Prometheus and Spring Cloud Stream Kafka by restarting the Kafka server and forcing GC a few times. If you need it, I could try to share a project that shows the problem.

**Expected behavior**
Kafka Metrics should stay current and most of them (e.g. `kafka_consumer_fetch_manager_records_lag`) should never become `NaN`.

Probably, when the binder encounters a metric that is already registered, it should replace that metric with the new one instead of leaving it.

lbilger:
Fixes #2726

See #2726 for details.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
@@ -139,7 +139,7 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
         }
 
         if (startTime != null) {
-            bindMeter(registry, startTime, meterName(startTime), meterTags(startTime));
+            bindMeter(registry, metricsSupplier, startTime.metricName(), meterName(startTime), meterTags(startTime));
         }
     }
 
@@ -194,7 +194,7 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
 
                 List<Tag> tags = meterTags(metric);
                 try {
-                    Meter meter = bindMeter(registry, metric, meterName, tags);
+                    Meter meter = bindMeter(registry, metricsSupplier, metric.metricName(), meterName, tags);
                     List<Meter> meters = registryMetersByNames.computeIfAbsent(meterName, k -> new ArrayList<>());
                     meters.add(meter);
                 }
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -212,36 +212,42 @@ else if (tags.size() == meterTagsWithCommonTags.size())
         }
     }
 
-    private Meter bindMeter(MeterRegistry registry, Metric metric, String name, Iterable<Tag> tags) {
-        Meter meter = registerMeter(registry, metric, name, tags);
+    private Meter bindMeter(MeterRegistry registry, Supplier<Map<MetricName, ? extends Metric>> metrics, MetricName metricName, String name, Iterable<Tag> tags) {
```
jonatan-ivanov, line 215: I think improving the variable names a little here and in the other methods would improve readability:
- `metrics` to `metricsSupplier`
- `name` to `meterName`

jonatan-ivanov (changes requested):
Can we rebase this to 1.6.x so that other LTS versions could be fixed too?

jonatan-ivanov:
@lbilger Thanks for the PR, I left a few comments: I think something behaves weirdly (though I tested it after I rebased your changes to `1.6.x`).

lbilger:
Hi Jonatan, thanks for having a look! I have rebased to 1.6.x, renamed the variables as requested and added a unit test. 

Can you share your test that showed the weird behavior? `metrics.get().get(metricName)` can return `null` if the metric is no longer present in the Consumer/Producer but `checkAndBindMetrics` has not yet been called since the metric was removed. If that is the case here, I would say it is correct and expected behavior to return NaN. If it is something else, I would have to investigate how it happens.

jonatan-ivanov (approved):

jonatan-ivanov:
The weird behavior was caused by me, in my sample (I used the kafka consumer + producer against a kafka instance), I closed the producer after sending the data so the kafka client removed a bunch of `KafkaMetric` instances so your fix did what it was expected to do, I somehow missed that my producer is closed.

Issue #880, opened 2018-09-28
Support to get http response size by micrometer


dipindas:
Please add support  to get http response size as a metric value

jkschneider:
Thanks. Surprised we didn't think of that earlier. Will do it with a `DistributionSummary` so you can best understand the distribution of response sizes over time.

mweirauch:
Just out of curiosity, would this include body-only-size or include the headers as well?

shakuzen:
@mweirauch I think we should do both body and headers (with a tag?) as much as possible.  Also, a separate metric for request and response sizes would be potentially useful.
![both?](https://media.giphy.com/media/3o7aCRloybJlXpNjSU/giphy.gif)

mweirauch:
Sounds good.
Would we be differentiating between server and client side instrumentation? Guess so.
So in analogy to `http.server.requests` and `http.client.requests`:
`http.server.request.size` and `http.client.request.size`? (Don't pin me on the naming.)

izeye:
> `http.server.request.size` and `http.client.request.size`? (Don't pin me on the naming.)

@mweirauch I guess you meant `http.server.response.size` and `http.client.response.size`?

fzyzcjy:
Hi is there any updates (after 2 years)?

drriguz:
> Hi is there any updates (after 2 years)?

Hi is there any updates (after 3 years)?

MayankManas1:
hi, is there any updates (after 4 years)?

jonatan-ivanov:
There was:
>[shakuzen](https://github.com/shakuzen) modified the milestones: [1.x](https://github.com/micrometer-metrics/micrometer/milestone/37), [1.11 backlog](https://github.com/micrometer-metrics/micrometer/milestone/189) [on Jan 31](https://github.com/micrometer-metrics/micrometer/issues/880#event-8397950650)

It's not a guarantee that it will make it to 1.11 but issues like this (`help-wanted` label) are good for contributions is anyone wants to file a PR.

babbarimam:
Hi @shakuzen ,
can you please confirm when and in which version we can expect the http response size metrics? 
Thanks

bclozel:
From a Framework instrumentation perspective:

[...]

PR #2772, opened 2021-09-09
Request and response size for Jersey

cyberdelia:
Emit request and response size if available, under the `http.server.request.size` and `http.server.response.size` metrics (the prefix is configurable like the other http metrics).

Fix: #880

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListener.java
@@ -20,6 +20,7 @@ import io.micrometer.core.instrument.LongTaskTimer;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer;
 import org.glassfish.jersey.server.ContainerRequest;
+import org.glassfish.jersey.server.ContainerResponse;
 import org.glassfish.jersey.server.model.ResourceMethod;
 import org.glassfish.jersey.server.monitoring.RequestEvent;
 import org.glassfish.jersey.server.monitoring.RequestEventListener;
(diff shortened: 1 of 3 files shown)


micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/MetricsRequestEventListener.java
```diff
@@ -93,6 +95,9 @@ public void onEvent(RequestEvent event) {
                     }
                 }
 
+                registry.summary(metricName + ".request.size", tagsProvider.httpRequestTags(event)).record(containerRequest.getLength());
+                registry.summary(metricName + ".response.size", tagsProvider.httpRequestTags(event)).record(containerResponse.getLength());
```
shakuzen, line 99: There was a comment on the original issue about whether to record the size of the body and/or headers. If we tag this as the body, then an instrumentation can add the values for headers if that's useful for someone. I'm not sure what the best tag name would be though... maybe `http.part` or even just `part`, with the value in this case being `body`

shakuzen (commented):
It's a bit unfortunate that the request count will be duplicated three times this way: once in the timer, and once in each DistributionSummary. Not a big problem, though.

micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/MetricsRequestEventListener.java
cyberdelia: I believe `http.headers.size` could work.

Not sure about actually measure the size (in bytes) of headers though, at it could lead to some security/performance issues, since there is no good way to obtain it.

Unrelated to headers size, should response size collection be optional maybe something like `autoSizeRequests` and `@Sized`?

cyberdelia: Any opinions?

cyberdelia:
Rebased against main branch.

Issue #2372, opened 2020-12-08
zgc pause time is not right


linking12:
![image](https://user-images.githubusercontent.com/4882172/101458029-42d0cd00-3971-11eb-9821-835638723f8c.png)

formula：rate(jvm_gc_pause_seconds_sum{application="$application", instance="$instance"}[1m])/rate(jvm_gc_pause_seconds_count{application="$application", instance="$instance"}[1m])

question：
The STW time of zgc is less 10ms, but why jvm_gc_pause_time is 822ms?  can not understand this phenomenon

linking12:
I think isConcurrentPhase is wrong for zgc

```
private boolean isConcurrentPhase(GarbageCollectionNotificationInfo info) {
    // So far the only indicator known is that the cause will be reported as "No GC"
    // when using CMS.
    //
    // For ZGC, the allocation stall seems to indicate that some thread or threads are
    // blocked trying to allocate. Even though it is not a true STW pause, counting it
    // as such seems to be less confusing.
    return "No GC".equals(info.getGcCause())            // CMS
        || "Shenandoah Cycles".equals(info.getGcName()) // Shenandoah
        || ("ZGC".equals(info.getGcName()) && !"Allocation Stall".equals(info.getGcCause()));
  }
```

shakuzen:
Thank you for the report. Things do seem probably out of sorts. I'd like to check on the ZGC devs mailing list what the correct way is to separate out concurrent and pause time from gc notifications.

shakuzen:
I also just came across [JDK-8255289](https://bugs.openjdk.java.net/browse/JDK-8255289). The comments seem to indicate it's no longer possible in the latest implementations to distinguish pause time from concurrent time using the gc notifications for ZGC. This is unfortunate, especially since there doesn't seem to be an alternative available currently either.

linking12:
> I also just came across [JDK-8255289](https://bugs.openjdk.java.net/browse/JDK-8255289). The comments seem to indicate it's no longer possible in the latest implementations to distinguish pause time from concurrent time using the gc notifications for ZGC. This is unfortunate, especially since there doesn't seem to be an alternative available currently either.

Allocation Stall is similar with pause, for ZGC, I suggest use this value to replace pause time.
For ZGC, the allocation stall seems to indicate that some thread or threads are blocked trying to allocate. Even though it is not a true STW pause, counting it as such seems to be less confusing.

shakuzen:
[...]

Issue #2561, opened 2021-04-12
Count the JVM GC events that happen before subscribing to notifications


shakuzen:
I noticed as part of investigating and testing #2514 that GC events can happen before we subscribe to notifications and those will not be counted. This will result in less GC count/time as well as lower allocated/promoted bytes than actual. We have aggregate information about GCs up until binding in the `GarbageCollectorMXBean`s. Since that is aggregate info, we can't accurately record it in timer since we lack the individual timings. It would be questionable to record the average time, but I'm not sure it is worse than missing these GC events entirely. Over time, either way becomes inconsequential, but for short-lived applications, the metrics not including events before binding could be significant.

marcingrzejszczak:
There was no interest in the issue since 2021. We’ll close it for now, and if there still will be any interest, please comment on this issue and we’ll consider re-opening it.

PR #2795, opened 2021-09-27
Support ZGC Cycles MXBean

shakuzen:
Since Java 17, the GarbageCollectorMXBean for ZGC has been split into `ZGC Pauses` and `ZGC Cycles`, which allows us to properly measure the pause time and the concurrent collection time. Before we only had access to the overall time.

Resolves gh-2372


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
```diff
@@ -61,6 +73,49 @@ void metersAreBound() {
         assertThat(registry.find("jvm.gc.memory.promoted").counter()).isNotNull();
     }
 
+    @Test
+    @Disabled("Garbage collection can happen before JvmGcMetrics are registered, making our metrics not match overall counts/timings")
```
shakuzen, line 77: While the test worked consistently for me on IntelliJ with ZGC and Shenandoah, it failed on the command line because of GC happening before JvmGcMetrics is registered, so the information in the GarbageCollectorMXBeans and our metrics do not add up. This is the issue described in #2561.

jonatan-ivanov (approved):

PR #2799, opened 2021-09-29
Ensure distribution can be configured per-metric

### LINKED ISSUE #2798: Can't configure DISTRIBUTION on per-metric basis for Datadog
**Describe the bug**
publishPercentileHistogram for all Timers/Summaries is set by the very first Timer/Summary and publishPercentileHistogram flag in timers/summaries that are created after that is ignored.

**Environment**

 - Micrometer version [1.8.0-M3]
 - Micrometer registry [StatsD with Datadog flavor]
 - OS: [macOS]
 - Java version: ❯ java --version 
openjdk 11.0.2 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

**To Reproduce**
How to reproduce the bug:
```
    @Test
    void timerSentAsDatadogDistribution_whenPercentileHistogramDisabled() {
        final Sinks.Many<String> lines = sink();
        final StatsdConfig config = configWithFlavor(StatsdFlavor.DATADOG);
        registry = StatsdMeterRegistry.builder(config)
                .clock(clock)
                .lineSink(toLineSink(lines, 2))
                .build();

        StepVerifier.create(lines.asFlux())
                .then(() -> {
                    Timer.builder("my.timer2").publishPercentileHistogram(false).register(registry).record(20, TimeUnit.SECONDS);
                    Timer.builder("my.timer").publishPercentileHistogram(true).register(registry).record(2, TimeUnit.SECONDS);
                })
                .expectNext("my.timer2:20000|ms")
                .expectNext("my.timer:2000|d")
                .verifyComplete();
    }

    @Test
    void timerSentAsDatadogDistribution_whenPercentileHistogramEnabled() {
        final Sinks.Many<String> lines = sink();
        final StatsdConfig config = configWithFlavor(StatsdFlavor.DATADOG);
        registry = StatsdMeterRegistry.builder(config)
                .clock(clock)
                .lineSink(toLineSink(lines, 2))
                .build();

        StepVerifier.create(lines.asFlux())
                .then(() -> {
                    Timer.builder("my.timer").publishPercentileHistogram(true).register(registry).record(2, TimeUnit.SECONDS);
                    Timer.builder("my.timer2").publishPercentileHistogram(false).register(registry).record(20, TimeUnit.SECONDS);
                })
                .expectNext("my.timer:2000|d")
                .expectNext("my.timer2:20000|ms")
                .verifyComplete();
    }
```

**Expected behavior**
See the failing tests above.

**Additional context**
Issue was introduced with this PR: https://github.com/micrometer-metrics/micrometer/pull/2745/files

Problematic change seems to be introduction of the second `StatsdLineBuilder lineBuilder(Meter.Id id, @Nullable DistributionStatisticConfig distributionStatisticConfig)' method and the fact that initial overload without DSC parameter was kept in place and being used.

The second problem that builds on top of the first one is the fact that `lineBuilderFunction` is cached in `StatsdMeterRegistry.lineBuilder(Meter.Id, DSC)` method when it must be created every single time now as we must be able to configure 2 different DatadogStatsdLineBuilder's - one that generates percentileHistogram and reports as DISTRIBUTION and another one that reports as regular TIMER or HISTOGRAM.

cyberdelia:
Ensure emitting distributions can be configured per-metric, as reported by @hdv.

Fix #2798

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -502,7 +502,8 @@ public class StatsdMeterRegistry extends MeterRegistry {
          * Used for completely customizing the StatsD line format. Intended for use by custom, proprietary
          * StatsD flavors.
          *
-         * @param lineBuilderFunction A mapping from a meter ID to a StatsD line generator that knows how to write counts, gauges
+         * @param lineBuilderFunction A mapping from a meter ID and a Distribution statistic configuration
+         *                            to a StatsD line generator that knows how to write counts, gauges
          *                            timers, and histograms in the proprietary format.
          * @return This builder.
          */
@@ -510,6 +511,24 @@ public class StatsdMeterRegistry extends MeterRegistry {
             this.lineBuilderFunction = lineBuilderFunction;
             return this;
         }
+
+
+        /**
+         * Used for completely customizing the StatsD line format. Intended for use by custom, proprietary
+         * StatsD flavors.
+         *
+         * @param lineBuilderFunction A mapping from a meter ID to a StatsD line generator that knows how to write counts, gauges
+         *                            timers, and histograms in the proprietary format.
+         *
+         * @return This builder.
+         * @deprecated Use {@link #lineBuilder(BiFunction)} instead since 1.8.0.
+         */
+        @Deprecated
+        public Builder lineBuilder(Function<Meter.Id, StatsdLineBuilder> lineBuilderFunction) {
+            this.lineBuilderFunction = (id, dsc) -> lineBuilderFunction.apply(id);
+            return this;
+        }
+
 
         public Builder nameMapper(HierarchicalNameMapper nameMapper) {
             this.nameMapper = nameMapper;
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -85,7 +85,7 @@ public class StatsdMeterRegistry extends MeterRegistry {
     private Disposable.Swap meterPoller = Disposables.swap();
 
     @Nullable
-    private Function<Meter.Id, StatsdLineBuilder> lineBuilderFunction;
+    private BiFunction<Meter.Id, DistributionStatisticConfig, StatsdLineBuilder> lineBuilderFunction;
 
     @Nullable
     private Consumer<String> lineSink;
@@ -112,7 +112,7 @@ public class StatsdMeterRegistry extends MeterRegistry {
                                 HierarchicalNameMapper nameMapper,
                                 NamingConvention namingConvention,
                                 Clock clock,
-                                @Nullable Function<Meter.Id, StatsdLineBuilder> lineBuilderFunction,
+                                @Nullable BiFunction<Meter.Id, DistributionStatisticConfig, StatsdLineBuilder> lineBuilderFunction,
                                 @Nullable Consumer<String> lineSink) {
         super(clock);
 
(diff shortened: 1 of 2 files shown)


hdv:
Cool! Thanks for the fix @cyberdelia!

I have a comment on tests. If you look at the test examples I provided in the issue, you can see that they both are for the Timer and they test different order of registration.
I'd imagine you might want to create 2 sets of tests: one for Timer and another one for DistributionSummary:
set for Timers will contain exactly what I wrote in the issue and set for DSummaries will be almost the same except for the type of the Metric.
Hope this makes sense.

I really like how you fixed it and kept the caching in place. I hope this idea didn't cross my mind because I was too tired debugging this issue for a few hours by that time and dropping the cache altogether was the only thing I was able to think of :-)

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -506,7 +506,7 @@ public Builder clock(Clock clock) {
          *                            timers, and histograms in the proprietary format.
          * @return This builder.
          */
-        public Builder lineBuilder(Function<Meter.Id, StatsdLineBuilder> lineBuilderFunction) {
```
jkschneider, line 509: This is a breaking API change. Is there a way you can add an overload to `lineBuilder` that uses the `BiFunction` but still preserves the `Function` variant as well, possibly in deprecated form?

jkschneider (changes requested):

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
cyberdelia: Fixed!

cyberdelia:
Build is failing due to external dependencies issues.

jonatan-ivanov (approved):

jonatan-ivanov:
@hdv Thank you very much catching, investigating and writing a test!
@cyberdelia Thank you very much fixing it!
I fixed `main`, if you back-merge `main` to your branch the build should be successful again.
I guess what is going on is the jackson release process takes a lot of time and we are depending on `latest.release` with `main` so if they are in the middle of a release, some modules are already published and some others haven't (not found error).

jkschneider (approved):

cyberdelia:
@jonatan-ivanov I rebased already, but no luck.

jkschneider:
@jonatan-ivanov I think we're good to merge this. Jackson will work itself out eventually. Happens periodically.

jkschneider:
BTW - Thank you @cyberdelia for the good work on this :)

jonatan-ivanov:
@cyberdelia Yeah if I would have pushed into upstream not into my fork. :D 
You need this commit: https://github.com/micrometer-metrics/micrometer/commit/4ba3146a8d90c1bae9aa6aaa352b5d12e087e462

@jkschneider I think we bump into this with almost every jackson release, they publish the bom before the other modules and the release process is long. I opened an issue for this though it is not the end of the world.

jonatan-ivanov:
@cyberdelia Thank you again!

jkschneider:
> @jkschneider I think we bump into this with almost every jackson release, they publish the bom before the other modules and the release process is long. I opened an issue for this though it is not the end of the world.

@jonatan-ivanov I've never found a better idea. If you pin, then you've just got another problem. I think there's technically a way to do it with a complicated bit of gradle, but haven't found it worthwhile for the occasional trouble ;)

jonatan-ivanov:
@jkschneider I think it's good to keep it as `latest.release`, I have two ideas that could help in this:
1. `latest.release.older-than-1d` (I think I'll ask Gradle folks if they want to add something like this)
2. jackson can improve this by publishing in a more optimal way (e.g.: lastly the bom) but I have no idea how hard this is, probably does not worth the hassle

PR #2808, opened 2021-10-08
Avoid reliance on fragile RegistryResolver implementation

### LINKED ISSUE #2806: Fragile reference to private synthetic `this$` field
The following code is using reflection to access a field with a name starting with `this$`:

https://github.com/micrometer-metrics/micrometer/blob/55892ea27afb24fd60bea3ca42e214dddebc2f7f/micrometer-test/src/main/java/io/micrometer/core/tck/RegistryResolver.java#L51

The OpenJDK javac generates private synthetic fields with names starting with `this$` as an implementation detail of inner classes. In the future that implementation detail may be changing, and the `this$` field will no longer be generated for all inner classes. For more information about the proposed change, see: https://bugs.openjdk.java.net/browse/JDK-8271717

Please consider alternatives to accessing the private synthetic `this$` field to ensure this code continues to work after the change. For example, consider passing an explicit copy of the enclosing instance to code that needs access to it, or adding an explicit getter to the inner class.

For example, given:

```java
class Outer {
  int x;
  class Inner1 {
    int f() {
      return x;
    }
  }
  class Inner2 {
    void g() {
      System.err.println("hello");
    }
  }
}
```

Currently `Inner1` and `Inner2` both have a synthetic field named `this$0` that stores a reference to `Outer`.

In the future the implementation detail might be changing to omit the field from classes that don't reference the enclosing instance. So in the example, `Inner1` would still have the synthetic field because it accesses the field `x` in the enclosing instance `Outer`. However `Inner2` would no longer have a synthetic field, because it doesn't access any state from its enclosing instance.

shakuzen:
Rather than use a `ParameterResolver`, this changes the implementations to use a protected class field instead. In order to achieve this, test cases in various interfaces were moved to inner classes in the `MeterRegistryCompatibilityKit` and all `MeterRegistry` parameters were removed. Everything still runs and passes as before.

`MeterRegistry` implementations are tested against our TCK by making a test class that extends `MeterRegistryCompatibilityKit` and implements its two abstract methods. All the TCK tests are inherited and run against the provided `MeterRegistry` implementation.

Resolves #2806

### DIFF
--- a/micrometer-test/src/main/java/io/micrometer/core/tck/RegistryResolver.java
+++ b/micrometer-test/src/main/java/io/micrometer/core/tck/RegistryResolver.java
@@ -27,6 +27,11 @@ import java.lang.reflect.Method;
 import static org.assertj.core.api.Assertions.fail;
 
 /**
+ * This is for internal use by the Micrometer MeterRegistry TCK: {@link MeterRegistryCompatibilityKit}.
+ * It allows resolving the {@link MeterRegistry} under test as a parameter to test methods.
+ * It relies on reflection and implementation details of the java compiler to resolve and invoke
+ * the top-level parent class method {@code registry()}.
+ *
  * @deprecated use {@link MeterRegistryCompatibilityKit#registry} instead.
  */
 @Deprecated


shakuzen:
I think we can apply this to maintenance branches, since we do want to compile with Java 18+ (still targeting `--release 8`) when available, and this change shouldn't break third-party implementations using the TCK either. We'll plan to delete `RegistryResolver` in 1.9 since it is public API, though it really has no utility besides with the TCK and probably should have been package private to begin with.

shakuzen:
Thanks for the review and the pointer to the `TestInstances` API.

Issue #2805, opened 2021-10-07
Allow customization of tags set by the Grpc client/server interceptors


taer:
**Please describe the feature request.**
The AbstractMetricCollectingInterceptor for the Grpc metrics currently hardcodes the tags for service, method, etc. It would be nice to allow this to be overridden on construction

**Rationale**
The `service` tag used to tag the grpc service is a fairly common tag. Some metric sinks even call it "reserved": https://docs.datadoghq.com/getting_started/tagging/

**Additional context**
It's possible to do this with a MeterFilter.renameTag call, but the metric code lives in a library. This means all consumers of the library would have to remember and perform this same rename.

shakuzen:
Thank you for letting us know. We should probably take the `TagsProvider` approach used elsewhere for the default tags and allowing customization.

I've opened #2807 about the Datadog reserved tag keys.

marcingrzejszczak:
Isn't customization possible by using the `counterCustomizer` and `timerCustomizer` ?

Issue #2622, opened 2021-05-26
Support extra tags for each gRPC call


ivansenic:
Currently the tags in the gRPC `MetricCollectingServerInterceptor` are fixed to `service, method, methodType & statusCode`. 

It would be great if there would be a support for a custom tags that depend on the `ServerCall` and `Metadata` as well, not only the method being invoked. This way the user would have freedom to add extra tags with values that depend on the caller.

jonatan-ivanov:
We have this functionality for other instrumentations, [search](https://github.com/micrometer-metrics/micrometer/find/main) for classes that contains TagsProvider in their names, e.g.: [`MongoCommandTagsProvider`](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoCommandTagsProvider.java) and [`DefaultMongoCommandTagsProvider`](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/DefaultMongoCommandTagsProvider.java) (see: [usage in `MongoMetricsCommandListener`](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java)).


I think if there is enough interest in it, we can do something similar. Would you be up to creating a PR for it?

jonatan-ivanov:
@ivansenic Would the components in my previous comment help?
Do you have any proposal?

ivansenic:
@jonatan-ivanov Creating the tag provider interface and default implementation is easy. I am more interested in how can this be plugged in the current implementation.

[...]

PR #2810, opened 2021-10-08
Add TagProvider to customize the tag names in Grpc metric interceptors

taer:
For https://github.com/micrometer-metrics/micrometer/issues/2805

I went with a simple tag provider. I left the existing tag customizers as is for now.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcTagProvider.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcTagProvider.java
@@ -1,0 +1,16 @@
+package io.micrometer.core.instrument.binder.grpc;
+
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import io.micrometer.core.instrument.Tag;
+
+public interface GrpcTagProvider {
+    /**
+     * Provides tags to be associated with metrics for the given {@code mothod}
+     * @param method The method the tags will be created for.
+     * @return tags to associate with metrics for the request and response exchange
+     */
+    Iterable<Tag> getBaseTags(final MethodDescriptor<?, ?> method);
+    Iterable<Tag> getTagsForResult(final MethodDescriptor<?, ?> method, Status.Code code);
+
+}
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcTagProvider.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcTagProvider.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2021 VMware, Inc.
+ * <p>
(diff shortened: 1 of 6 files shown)


taer:
Added some tests. 

@jonatan-ivanov See my prior comment about #2622    Would be pretty simple, but would change the implementation of the taggers pretty drastically.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcTagProvider.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+package io.micrometer.core.instrument.binder.grpc;
+
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import io.micrometer.core.instrument.Tag;
+
+public interface GrpcTagProvider {
```
ivansenic, line 23: This is not solving what was described in https://github.com/micrometer-metrics/micrometer/issues/2622.. There should be a way to add additional tags for each call.. This should work for both client and server side..

taer: Correct. See my comment here: https://github.com/micrometer-metrics/micrometer/pull/2810#issuecomment-944392883

The current implementation computes and caches all the tags for a `MethodDescriptor`. Adding the ability to tag based on `Metadata` as well would pretty much remove that optimization, resulting in new tags created per call instead of the current cache based solely on `MethodDescriptor`

I was awaiting feedback on that decision.

taer: I actually just thought of something. Instead of a single tagprovider, perhaps there could be 2 different interfaces. One w/ just MethodDescriptor and is cached, the other with both and is computer per call. It's been a while since I wrote this, I'll take a peek and see if we could accomplish both.

shakuzen:
> If we tossed in ServerCall and Metadata, it would necessitate a removal of the cache and a fresh call to `registerMeterIfNecessary` for each call. Or some sort of complex implementation that could maintain the cache for tags based on `methodDescriptor` only.
> 
> Thoughts?

I think that's the price we'll have to pay for the flexibility of tagging based on those.

ivansenic:
I added here in our project a way to add a tag provider that returns extra tags for each call -> https://github.com/stargate/stargate/blob/3030c23f4f1b82e904c713ca41e8d4557e62391d/grpc/src/main/java/io/micrometer/core/instrument/binder/grpc/TaggingMetricCollectingServerInterceptor.java..

As you can see if the tag provider returns empty tags, I am going back to the cached version. This is possible because tag provider must maintain the same tags for each call.. 

Such extension would be even simpler here directly in the micrometer code base, as I had to copy bunch of the code from the super class. Not sure if something like this would be acceptable, use `NoopTagProvider` and you can still have the cached version..

Would be happy to contribute myself..

Issue #2811, opened 2021-10-12
Exemplars support for Prometheus Counter


jonatan-ivanov:
This issue belongs to https://github.com/micrometer-metrics/micrometer/issues/2672 (Support for Exemplars), see the details there.

izeye:
I'm not sure if it's intentional, but this doesn't have any milestone assigned.

Issue #2812, opened 2021-10-12
Exemplars support for Prometheus Histogram


jonatan-ivanov:
This issue belongs to https://github.com/micrometer-metrics/micrometer/issues/2672 (Support for Exemplars), see the details there.

jonatan-ivanov:
@neiser I am (was? 😃), but we are always happy to see contributions so if you want to work on it, please feel free to do so.
I took a look at the Histogram support at the time I wrote support for Counters but it is a little more complicated case and I needed to focus on other things. We are planning to add this in the next minor release of Micrometer (1.9.0) so your timing is perfect.

neiser:
@jonatan-ivanov Alright, then let me see if I can provide a PR in the coming days.

neiser:
@jonatan-ivanov I've looked into the code and figured that I'd need to start at `PrometheusDistributionSummary#recordNonNegative` to trigger recording an exemplar for `count`/`amount`/`max` and all `histogram` buckets. I didn't quite get why `PrometheusDistributionSummary` manages its own `histogram`, while the base class `AbstractDistributionSummary` already has such a field (though configuration of it seems to be complex, in particular with the weirdly integrated case of `VictoriaMetrics` which also surprisingly appear in core module as `io.micrometer.core.instrument.distribution.FixedBoundaryVictoriaMetricsHistogram` (my impression was that micrometer tries to abstract away any particular backend system for metrics collection... which is kinda broken by this class IMHO).

Anyway, my question is now is: Should I attempt a refactoring which would make getting exemplars into histogram buckets easier? Or should I just hack a lot of stuff around the "prometheus" part of Micrometer to achieve whatever resolving this issue takes?

fscellos:
Hello i did some test to integrate exemplars in histogram. 
For test i implemented it by define a custom TimeWindowFixedBoundaryHistogram (some private attributes (bucket) does not allow a direct inheritance from AbstractTimeWindowHistogram) and try to inspire from code of @jonatan-ivanov in its counter.

ReadMe is in french but you can take a look at : [github / micrometer-prometheus-examplar ](https://github.com/fscellos/micrometer-prometheus-examplar ). If you want to check it, i'll be happy to try to find an implementation with you.

It seems to work from a spring boot application (by defining PrometheusMeterRegistry Bean to override default actuator behaviour) but i always have a little problem....
[...]

PR #2813, opened 2021-10-12
Exemplars support for Prometheus Counter

jonatan-ivanov:
fixes gh-2811


fscellos:
Hello @jonatan-ivanov , as i comment in #2812, it is necessary to take account sampled trace flags (https://www.w3.org/TR/trace-context/#sampled-flag) to be sure that the trace_id set in exemplars will be effectively recorded in backend storage.

fscellos:
Hello, this is doable with OpenTelemetry API (Span.current().getSpanContext().getTraceFlags().isSampled()). 
To use it, i think we must implement and use Samplers (ExemplarSampler) that take this information into account no new Examplar is this data is not set to true).
I will try.

jonatan-ivanov:
@fscellos That would be the responsibility of the implementation of the `SpanContextSupplier` that we are not providing here.

fscellos:
I create an issue on prometheus client for this case : https://github.com/prometheus/client_java/issues/765

shakuzen (approved):

Issue #2744, opened 2021-08-12
CumulativeHistogramLongTaskTimer histogram counts increase unexpectedly


shakuzen:
I'm noticing behavior that doesn't look right related to this. If I scrape the prometheus endpoint while there is an active task tracked with a LongTaskTimer and histogram enabled, the bucket counts increase with each scrape. All other stats related to the LongTaskTimer (`+Inf` bucket, active count, duration sum, max) do not change over scrapes, and they go to zero when there are no active tasks. The histogram bucket counts (except for `+Inf`) do not go to zero.

_Originally posted by @shakuzen in https://github.com/micrometer-metrics/micrometer/pull/2075#r687514745_

shakuzen:
The code was originally added with https://github.com/micrometer-metrics/micrometer/commit/9fa83e3d3638edf725aa5d803ae0b52b9017456f. I am not sure but I think the intention is to make the histogram buckets monotonic across Prometheus scrapes so that when a task completes, buckets do not decrease in value. @jkschneider can you confirm?

Unfortunately, this doesn't work because multiple Prometheus instances can scrape in the same interval, which causes bucket counts to double each scrape when the number of active tasks is the same. The difference will not be consistent between scrapes for the same Prometheus instance, either. Here are two tests. The first demonstrating the behavior we want, which passes. The second fails, which shows the problematic behavior with multiple scrapes (calls to `takeSnapshot`).

```java
@Test
void bucketCountsAreMonotonic_whenTasksComplete() {
    LongTaskTimer ltt = LongTaskTimer.builder("my.ltt").publishPercentileHistogram().register(registry);
    LongTaskTimer.Sample sample = ltt.start();
    clock.add(1, TimeUnit.MINUTES);
    CountAtBucket[] countAtBuckets = ltt.takeSnapshot().histogramCounts();

    assertThat(countAtBuckets[0].bucket(TimeUnit.MINUTES)).isEqualTo(2);
    assertThat(countAtBuckets[0].count()).isOne();

    sample.stop();
    // bucket counts after task completed should not decrease
    countAtBuckets = ltt.takeSnapshot().histogramCounts();
    assertThat(countAtBuckets[0].count()).isOne();
}

@Test
@Issue("#2744")
void bucketCountsStableWhenMultipleSnapshotsTaken() {
    LongTaskTimer ltt = LongTaskTimer.builder("my.ltt").publishPercentileHistogram().register(registry);
[...]

PR #2851, opened 2021-11-08
`DefaultLongTaskTimer` histogram bucket counting logic fix

shakuzen:
The histogram was not correct in all cases before. The added test demonstrates a case that would fail with the previous logic. Buckets that should have been zero were not.

I came across this while looking into another bug with LongTaskTimer histograms: #2744

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
@@ -30,6 +30,7 @@ import java.util.*;
 import java.util.concurrent.*;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 public class DefaultLongTaskTimer extends AbstractMeter implements LongTaskTimer {
     /**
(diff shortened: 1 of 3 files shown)


shakuzen:
Note: this is a separate issue from #2744. I have another test locally for that issue and will work on a fix for it separate from this change.

micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
```diff
@@ -156,25 +156,23 @@ public HistogramSnapshot takeSnapshot() {
...
             Iterator<SampleImpl> youngestToOldest = activeTasks.descendingIterator();
             while (youngestToOldest.hasNext()) {
                 SampleImpl activeTask = youngestToOldest.next();
-                i++;
-                if (bucket != null) {
-                    if (activeTask.duration(TimeUnit.NANOSECONDS) > bucket) {
-                        countAtBuckets.add(new CountAtBucket(bucket, i - 1));
-                        bucket = buckets.pollFirst();
-                    }
+                while (bucket != null && activeTask.duration(TimeUnit.NANOSECONDS) > bucket) {
```
jonatan-ivanov, line 164: Is `activeTask.duration(TimeUnit.NANOSECONDS)` recalculated in every iteration of this loop (as time passes while we iterate so as the duration)? Is it ok if it does?

shakuzen: It's calculated based on the configured `Clock`, which for this test is the `MockClock`, so it isn't changing unless explicitly updated.

shakuzen: Yes, that could result in some inconsistency. Let me update.

shakuzen: On thinking about it more, I don't think it would result in an inconsistent histogram. Though to keep in the spirit of a "snapshot" we should probably get the duration of all tasks early and use those rather than always using the latest duration.

shakuzen: Updated. Please check again.

jonatan-ivanov (approved):

Issue #2861, opened 2021-11-12
`MetricsRequestEventListener` with `@Timed(longTask=true)` makes two meters


shakuzen:
I'm not sure this behavior was intentional. If you annotate a Resource with `@Timed(longTask = true)` then you will end up with a `LongTaskTimer` and a `Timer` with the same name. I would have expected just a `LongTaskTimer`.
/cc @mweirauch

shakuzen:
Note: ported to the Jersey instrumentation in micrometer-core with 44f8a60803d6a333f7ad985c9c547d401682a0b8

PR #2862, opened 2021-11-12
Only make a LongTaskTimer with `@Timed(longTask=true)`

shakuzen:
Before both a `LongTaskTimer` and `Timer` were being made because the logic to make regular timers from the `@Timed` annotation did not filter out annotations where `longTask` was `true`.

Resolves gh-2861

### DIFF
--- a/micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/MetricsRequestEventListener.java
+++ b/micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/MetricsRequestEventListener.java
@@ -118,8 +118,9 @@ public class MetricsRequestEventListener implements RequestEventListener {
         }
 
         return timed.stream()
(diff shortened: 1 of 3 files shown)


micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/MetricsRequestEventListener.java
```diff
@@ -118,8 +118,9 @@ public void onEvent(RequestEvent event) {
         }
 
         return timed.stream()
-            .map(t -> Timer.builder(t, metricName).tags(tagsProvider.httpRequestTags(event)).register(registry))
-            .collect(Collectors.toSet());
+                .filter(annotation -> !annotation.longTask())
```
shakuzen, line 121: This is the filter logic that was missing.

PR #2863, opened 2021-11-15
Adding support to define a description by MeterRegistry

rhuan080:
Signed-off-by: Rhuan Rocha <rhuan080@gmail.com>

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -476,6 +476,44 @@ public abstract class MeterRegistry {
     }
 
     /**
+     * Register a gauge that reports the value of the object after the function
+     * {@code valueFunction} is applied. The registration will keep a weak reference to the object so it will
+     * not prevent garbage collection. Applying {@code valueFunction} on the object should be thread safe.
+     *
+     * @param name          Name of the gauge being registered.
+     * @param description          Description of the gauge being registered.
+     * @param tags          Sequence of dimensions for breaking down the name.
+     * @param stateObject   State object used to compute a value.
+     * @param valueFunction Function that produces an instantaneous gauge value from the state object.
+     * @param <T>           The type of the state object from which the gauge value is extracted.
+     * @return The state object that was passed in so the registration can be done as part of an assignment
+     * statement.
+     */
+    @Nullable
+    public <T> T gauge(String name,String description, Iterable<Tag> tags, @Nullable T stateObject, ToDoubleFunction<T> valueFunction) {
+        Gauge.builder(name, stateObject, valueFunction)
+                .description(description)
+                .tags(tags).register(this);
+        return stateObject;
+    }
+
+    /**
+     * Register a gauge that reports the value of the {@link Number}.
+     *
+     * @param name   Name of the gauge being registered.
+     * @param description   Description of the gauge being registered.
+     * @param tags   Sequence of dimensions for breaking down the name.
+     * @param number Thread-safe implementation of {@link Number} used to access the value.
+     * @param <T>    The type of the number from which the gauge value is extracted.
+     * @return The number that was passed in so the registration can be done as part of an assignment
+     * statement.
+     */
+    @Nullable
+    public <T extends Number> T gauge(String name, String description, Iterable<Tag> tags, T number) {
+        return gauge(name, description, tags, number, Number::doubleValue);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


rhuan080:
Hi,

Let me know if I need to apply some action to approve this PR.

jonatan-ivanov:
Hey @rhuan080,

There are two main ways to create a gauge (or any type of `Meter`s):
1. Using the registry
```java
AtomicInteger myGauge = registry.gauge("numberGauge", new AtomicInteger(0));
```
2. Using a builder:
```java
Gauge gauge = Gauge
    .builder("gauge", myObj, myObj::gaugeValue)
    .description("a description of what this gauge does") // optional
    .tags("region", "test") // optional
    .register(registry);
```

The builder supposed to give you all the possible options for configuration, using the registry is just a shortcut for the simplest scenarios. Since the description is optional and less important (than the name, tags, state) I don't think we should add it, it was designed this way intentionally.

What do you think, would the builder work for you?

rhuan080:
Hi @jonatan-ivanov, thank you for answering me!

Yes, but tag is optional as well. I agree the tag is more important than the description, but the description is used in many scenarios. I think creating a shortcut including the description is a good feature as well, as the description is used in many scenarios. In my scenario, all metrics have a description, thus I cannot use the shortcut. I used method overloading, thus it is just an option of shortcut to the scenario using a description. In my opinion, it just provides a good option without side effects. 

The builder works for me, but the shortcut turns the code clearer. Why not provide it to these scenarios as well?

jonatan-ivanov:
The tags are way more important than the description. Tags are vital part of the `Meter.Id` (see it `equals`, `hashCode`) of the meter and every new tag-value means a new time series.
The description does not have any of these properties, for most of our registries, the description is not even used (only used for datadog, dynatrace, prometheus, stackdriver).

Although, I completely agree that from the point of view of the output (for those 4 registries) the description is very useful from UX perspective but from the API usability and discoverability perspective, I think we should keep these shortcuts as minimal as possible.

Right now, we have 4 `gauge` methods, this PR would add 4 new ones; with this logic, we should duplicate every other shortcut methods not just `gauge`. Also, there are properties "more important" than the description, e.g.: in case of a `Timer` we have `minimumExpectedValue`, `maximumExpectedValue`, `publishPercentileHistogram`, `serviceLevelObjectives`, `publishPercentiles` and a couple of others. These do have an effect on the output data that the `Timer` produces (the description does not) so they might be considered more important. Should we add every possible combination of these too? I think we should not because that would make these shortcuts useless (it would be very hard to find them). We need to set these boundaries but everything that did not made into these shortcuts can be configured with the builder.

jonatan-ivanov:
@shakuzen @jkschneider Do you remember anyone asking for overloading these shortcuts by adding the description?

rhuan080:
I understood your question. However, the developer uses the IDE that in general has a autocomplete and suggests the methods to the developer. Thus, if the developer type ".gauge" the timer and count is not listed. Furthermore, the methods are well-documented by JavaDoc. In case of an increase of methods, we can rename it to help developers.

checketts:
I agree with @jonatan-ivanov that API consistency is important. If the description option were added to `gauge`s, then it really needs to be added to all types. That becomes a lot of extra noise with little benefit, and a lot of potential for problems: for example counter has a signature of `counter(String name, String... tags)`, adding a `description` parameter there could break a lot of code.

I can see the benefit of making some Kotlin extension functions available to help in this case (due to named and default parameters), but I don't want to assume you are using Kotlin.

jonatan-ivanov:
@rhuan080 I wrote my opinion above considering the capabilities of the IDEs: from API usability perspective, I think it is still important if users see a few methods (`.gauge` overloads) or a lot of them when their IDEs show their options. This is why we have just a few shortcuts so people won't get overwhelmed by the lot of different options (see my Timer example above). (This is also why we have `.more()` on the MeterRegistry.)

Also, as @checketts pointed out, this can be a breaking change in some cases. As he also mentioned you can works this around in Kotlin (extension functions), also in Groovy (metaprogramming) and Scala ("Pimp My Library pattern") too. In Java, you can have a `MeterFactory` class that exactly what you din in this PR without breaking other users or adding a bunch of new methods that they probably won't use.

rhuan080:
Hi @jonatan-ivanov, 

I`m sorry for my delay to answer. Thank you for explaining the reason. I think it makes sense!

jonatan-ivanov:
@rhuan080 I'm closing this but thank you for the PR, if you have other improvement ideas that you want to discuss, you can reach the community on https://slack.micrometer.io/

PR #2867, opened 2021-11-17
Add capability to export fixed buckets histogram to SignalFx

bogdandrutu:
This respects the same format as "Prometheus" histograms, see
https://docs.signalfx.com/en/latest/integrations/agent/monitors/prometheus-exporter.html

There is still a caveat for the "count" and "sum" which do not respect the same format
but that can be fixed in a separate PR since it may break some users.

### DIFF
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
@@ -25,10 +25,18 @@ import com.signalfx.metrics.flush.AggregateMetricSender;
 import com.signalfx.metrics.protobuf.SignalFxProtocolBuffers;
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.config.NamingConvention;
+import io.micrometer.core.instrument.distribution.CountAtBucket;
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
+import io.micrometer.core.instrument.distribution.HistogramSnapshot;
+import io.micrometer.core.instrument.distribution.pause.PauseDetector;
 import io.micrometer.core.instrument.step.StepMeterRegistry;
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -118,6 +129,22 @@ protected void publish() {
...
+        // TODO: Investigate if the default HistogramGauges metrics should be removed:
+        //  * Adds a new gauge ".histogram" with "le" tag when slo buckets defined.
+        //  * Adds a new gauge ".percentile" with "phi" tag when percentiles buckets are defined.
+        return super.newTimer(id, updateConfig(distributionStatisticConfig), pauseDetector);
+    }
+
+    @Override
+    protected DistributionSummary newDistributionSummary(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, double scale) {
+        // TODO: Investigate if the default HistogramGauges metrics should be removed:
+        //  * Adds a new gauge ".histogram" with "le" tag when slo buckets defined.
```
mateuszrzeszutek, line 143: The `.histogram` gauge seems to be almost equivalent to what you've implemented here, aside from the naming (`_bucket` meter name, `upper_bound` vs `le` tag) and the additional SLO bucket.
Have you considered using `NamingConvention` & `MeterFilter` to modify the configuration of the default `HistogramGauges`-generated meters?

bogdandrutu: Thanks for pointing to that (I looked into it before, but tried one more time):
* The upper_bound is as simple as asking the customer to add it manually which is not ideal but ok.
* The biggest problem is the fact that the reported metrics are Gauges not CumulativeCounters. I can do a small hack on our side to rewrite the type to be Counter.

bogdandrutu:
Dropping this in favor of a simpler approach https://github.com/micrometer-metrics/micrometer/pull/2977

PR #2871, opened 2021-11-23
Current sample and Scope API

shakuzen:


micrometer-core/src/main/java/io/micrometer/core/instrument/internal/TimedRunnable.java
```diff
@@ -27,20 +27,23 @@
...
     private final Timer.Sample idleSample;
+    private final Timer.Scope idleScope;
 
     TimedRunnable(MeterRegistry registry, Timer executionTimer, Timer idleTimer, Runnable command) {
         this.registry = registry;
         this.executionTimer = executionTimer;
         this.idleTimer = idleTimer;
         this.command = command;
         this.idleSample = Timer.start(registry);
+        this.idleScope = this.idleSample.makeCurrent();
```
shakuzen, line 38: I think probably we don't want to make the idle timer current. The idle timer starts when a `Runnable` is submitted and stops when it actually gets `run`. In between then, another `Timer` could be started on the thread where the `Runnable` was submitted, and I don't think it should know about the idle timer. What do others think?

PR #2874, opened 2021-11-26
`jvm.gc.live.data.size` and `max` not updating for optavgpause/optthruput collectors

shakuzen:
With the OpenJ9 non-generational collectors `optavgpause` and `optthruput`, a NPE was happening in the notification listener which caused some code in it to not be executed. Namely, setting the live and max data size was being effectively skipped for these collectors.

Now we check for `null` and only the allocated bytes counter is skipped when no allocation pool name is set.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
@@ -75,7 +75,8 @@ class JvmGcMetricsTest {
         if (!binder.isGenerationalGc) {
             return;
         }
-        assertThat(registry.find("jvm.gc.memory.promoted").counter().count()).isPositive();
+        // cannot guarantee an object was promoted, so cannot check for positive count
+        assertThat(registry.find("jvm.gc.memory.promoted").counter()).isNotNull();
     }
 
     @Test
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
@@ -34,7 +34,6 @@ import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.methods;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.within;
 import static org.awaitility.Awaitility.await;
-import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 /**
  * Tests for {@link JvmGcMetrics}.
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
```diff
@@ -65,13 +65,18 @@ void noJvmImplementationSpecificApiSignatures() {
...
     @Test
-    void metersAreBound() {
-        assertThat(registry.find("jvm.gc.live.data.size").gauge()).isNotNull();
-        assertThat(registry.find("jvm.gc.memory.allocated").counter()).isNotNull();
-        assertThat(registry.find("jvm.gc.max.data.size").gauge().value()).isGreaterThan(0);
+    void gcMetricsAvailableAfterGc() {
+        System.gc();
+        await().timeout(200, TimeUnit.MILLISECONDS).alias("NotificationListener takes time after GC")
+                .untilAsserted(() ->
+                        assertThat(registry.find("jvm.gc.live.data.size").gauge().value()).isPositive());
```
shakuzen, line 72: By forcing a GC and waiting for the notification listener to update metrics, this test does a much better job of ensuring expected metrics are available for the configured collector.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
```diff
@@ -65,13 +65,18 @@ void noJvmImplementationSpecificApiSignatures() {
...
+        assertThat(registry.find("jvm.gc.memory.allocated").counter().count()).isPositive();
```
shakuzen, line 73: This metric is still missing for `optavgpause` and `optthruput` due to the allocation pool name not being set. We don't have a CI job run with OpenJ9 because we're still waiting for CI images (possibly https://github.com/CircleCI-Public/cimg-openjdk/issues/78). I think we can fix the missing allocated counter separately from this pull request because it is still an improvement to get rid of the NPE which allows updating of the live and max data size gauges.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
```diff
@@ -65,13 +64,18 @@ void noJvmImplementationSpecificApiSignatures() {
...
+        assertThat(registry.find("jvm.gc.max.data.size").gauge().value()).isPositive();
 
-        assumeTrue(binder.isGenerationalGc);
```
shakuzen, line 73: When the condition failed, the whole test was marked as skipped, but if we want to use this same test for generational and non-generational collectors, we should run the part of this test before here for both and report it as passed (not skipped) for non-generational collectors if all assertions passed before here.

PR #2875, opened 2021-11-26
Do not update live/max gc size when unavailable

shakuzen:
Previously, we were updating the live/max gc memory pool(s) size for non-generational GCs on every notification. Some notifications (`Shenandoah Pauses` and `ZGC Pauses`) do not have memory pool size information included. By updating the metrics with those notifications, we were setting the live/max gauges to zero. These notifications are followed by a `Cycles` notification that contain the memory pool data. This made for a race condition between reading the live/max gc data size metrics and updating them. There is a small window during which the metrics were set to zero.

This fix avoids the race by skipping updating the live/max metrics for notifications known to not have memory pool info included.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -28,6 +28,7 @@ import io.micrometer.core.util.internal.logging.InternalLogger;
 import io.micrometer.core.util.internal.logging.InternalLoggerFactory;
 
 import javax.management.ListenerNotFoundException;
+import javax.management.Notification;
 import javax.management.NotificationEmitter;
 import javax.management.NotificationListener;
 import javax.management.openmbean.CompositeData;
@@ -73,6 +74,13 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
 
     private final List<Runnable> notificationListenerCleanUpRunnables = new CopyOnWriteArrayList<>();
 
+    private Counter allocatedBytes;
+    @Nullable
+    private Counter promotedBytes;
+    private AtomicLong allocationPoolSizeAfter;
+    private AtomicLong liveDataSize;
+    private AtomicLong maxDataSize;
+
     public JvmGcMetrics() {
         this(emptyList());
     }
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -208,8 +219,27 @@ private void countPoolSizeDelta(Map<String, MemoryUsage> before, Map<String, Mem
...
+        }
+
+        private boolean isMajorGenerationalGc(String gcName) {
+            return GcGenerationAge.fromGcName(gcName) == GcGenerationAge.OLD;
+        }
+
+        private boolean nonGenerationalGcShouldUpdateDataSize(String gcName) {
+            return !isGenerationalGc
+                    // Skip Shenandoah and ZGC gc notifications with the name Pauses due to missing memory pool size info
+                    && !gcName.endsWith("Pauses");
```
shakuzen, line 233: Hard to see in the diff (sorry for refactoring while bug fixing), but this is the key change.

shakuzen: Though as I'm thinking about it more, I'm not sure why we even need to consider non-generational collectors like this at all. I don't think there will be a GC notification for a non-generational collector where the `longLivedAfter < longLivedBefore` condition won't be met.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
```diff
@@ -103,6 +106,11 @@ void gcTimingIsCorrect() {
         checkCollectionTime(pauseTimeMs, concurrentTimeMs);
     }
 
+    @Test
+    @Issue("gh-2872")
+    void nonGenerationalSizeMetricsNotSetToZero() {
+    }
```
shakuzen, line 112: Oops. I intended to write a unit test for this issue, and I thought it would be easier to do now that the listener is not an anonymous class. I struggled to figure out how to instantiate some of the management classes involved with the gc notifications to check the behavior fixed. Let me take another whack at it or remove this before merging.

shakuzen:
With the test case finally added, this is ready for review. I had to give up on hand-crafting `Notification` instances to pass to the `NotificationListener` because it's just too painful. Instead I took the approach of capturing the real `Notification` used when a GC happens and replaying them to the listener but checking after each that the size metrics don't get set to zero, which was the issue.

jonatan-ivanov (approved):

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
```diff
@@ -103,6 +116,57 @@ void gcTimingIsCorrect() {
...
+                    notificationEmitter.removeNotificationListener(capturingListener);
+                } catch (ListenerNotFoundException ignore) {
+                }
+            });
+        }
+
+        try {
+            System.gc(); // capture real gc notifications
+            List<Notification> notifications = capturingListener.getNotifications();
+            assertThat(notifications).isNotEmpty();
```
shakuzen, line 144: This is flaky because while the GC will be finished once `System.gc()` finishes, the corresponding notifications are asynchronous and may take arbitrary time to be processed by the configured `NotificationListener` instances. Unfortunately, using awaitility to wait until we have a notification is probably not the best approach since we know, for example, the issue fixed in this pull request happens with Shenandoah which sends 4 notifications (we want to capture all), but other collectors will have a different number of notifications per GC. While still potentially flaky and time consuming, perhaps a 100ms or so sleep is the best way to improve the flakiness here. Though again, if we only capture a portion of the notifications for the gc, the test isn't necessary checking that the issue is fixed.

shakuzen: 100 ms sleep added in 13fa32548e909f32e08f4b5f29d264093bd7612b

PR #2877, opened 2021-11-26
Separate ZGC-dependent test from JvmGcMetricsTest

izeye:
This PR separates ZGC-dependent test from `JvmGcMetricsTest`.

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -115,7 +115,7 @@ task zgcTest(type: Test) {
     maxHeapSize = "1500m"
 
     useJUnitPlatform {
-        includeTags 'gc'
+        includeTags 'gc', 'zgc'
     }
 
     jvmArgs '-XX:+UseZGC'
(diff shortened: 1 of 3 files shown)


shakuzen (commented):
There's nothing specific to ZGC in that test case. It's demonstrating the issue described in https://github.com/micrometer-metrics/micrometer/issues/2561, which I could have made more clear in a comment or something.

izeye:
@shakuzen "Pauses" and "Cycles" in `GarbageCollectorMXBean.getName()` seem to be specific to ZGC as G1 GC doesn't seem to have them, for example. Am I missing something?

shakuzen:
I should have looked at the test more closely before commenting. Still, Shenandoah and ZGC use similar names for the GCMXBean names, so as mentioned [here](https://github.com/micrometer-metrics/micrometer/pull/2795/files#r716861120) the test demonstrates the issue for those two collectors. The underlying issue of collections happening before binding not being counted is not limited to those two collectors, though, so the test could probably be written better. I suspect I didn't polish it because it is disabled until we can get a fix for #2561

izeye:
@shakuzen Thanks for the quick feedback! Sounds better to be resolved in #2561, then. Closing.

PR #2892, opened 2021-12-06
rewroks impl to use unsafe unicast sink with MpScQueue

### LINKED ISSUE #2880: Statsd meter recordings are dropped when submitted from parallel threads
**Describe the bug**
The following test fails when added to `StatsdMeterRegistryPublishTest`:

```
    @ParameterizedTest
    @EnumSource(StatsdProtocol.class)
    void receiveParallelMetricsSuccessfully(StatsdProtocol protocol) throws InterruptedException {
        final int N = 10;

        skipUdsTestOnWindows(protocol);
        serverLatch = new CountDownLatch(N);
        server = startServer(protocol, 0);
        final int port = getPort(protocol);

        meterRegistry = new StatsdMeterRegistry(getUnbufferedConfig(protocol, port), Clock.SYSTEM);
        startRegistryAndWaitForClient();
        Counter counter = Counter.builder("my.counter").register(meterRegistry);

        IntStream.range(0, N)
                .parallel()
                .forEach(ignored -> counter.increment());

        assertThat(serverLatch.await(3, TimeUnit.SECONDS)).isTrue();
    }
```

It succeeds when `.parallel()` line is commented out.

**Environment**

 - Micrometer version [e.g. 1.7.1]: latest Micrometer from `main` branch
 - Micrometer registry [e.g. prometheus]: statsd
 - OS: [e.g. macOS] macOS
 ~/projects/micrometer #v1.8.0 *3 !1 ❯ java --version                                                                                                    base
openjdk 11.0.2 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

**To Reproduce**
How to reproduce the bug:
See the test above

**Expected behavior**
Test should succeed

**Additional context**
- Similar test succeeds for 1.7.6 release
- Looks like this change might be related: https://github.com/micrometer-metrics/micrometer/pull/2588
- increasing N to 1000 makes the test fail even without `.parallel()`

OlegDokuka:
closes #2880

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -28,6 +28,7 @@ import io.micrometer.core.util.internal.logging.WarnThenDebugLogger;
 import io.micrometer.statsd.internal.*;
 import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.util.AttributeKey;
+import io.netty.util.internal.shaded.org.jctools.queues.MpscUnboundedArrayQueue;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
@@ -39,6 +40,7 @@ import reactor.core.publisher.Sinks;
 import reactor.netty.Connection;
 import reactor.netty.tcp.TcpClient;
 import reactor.netty.udp.UdpClient;
+import reactor.util.concurrent.Queues;
 import reactor.util.retry.Retry;
 
 import java.net.InetSocketAddress;
@@ -76,6 +78,22 @@ import java.util.stream.DoubleStream;
 public class StatsdMeterRegistry extends MeterRegistry {
     private static final WarnThenDebugLogger warnThenDebugLogger = new WarnThenDebugLogger(StatsdMeterRegistry.class);
 
+    private static final boolean HAS_TURBO_FILTER;
+
+    static {
+        ClassNotFoundException e = null;
+        try {
+            Class.forName("ch.qos.logback.classic.turbo.TurboFilter",
+                    false,
+                    StatsdMeterRegistry.class.getClassLoader());
+        }
+        catch (ClassNotFoundException ex) {
+            e = ex;
+        }
+
+        HAS_TURBO_FILTER = e == null;
+    }
+
     private final StatsdConfig statsdConfig;
     private final HierarchicalNameMapper nameMapper;
     private final Map<Meter.Id, StatsdPollable> pollableMeters = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -185,6 +198,20 @@ void poll() {
 
     public void start() {
         if (started.compareAndSet(false, true)) {
+            if (statsdConfig.enabled()) {
+                Sinks.Many<String> sink = Sinks.unsafe()
+                                               .many()
+                                               .unicast()
+                                               .onBackpressureBuffer(new MpscUnboundedArrayQueue<>(256));
```
OlegDokuka, line 205: this should resolve problems with high CPU contention

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -213,10 +240,11 @@ public void onComplete() {
             } else {
                 final Publisher<String> publisher;
                 if (statsdConfig.buffered()) {
-                    publisher = BufferingFlux.create(this.sink.asFlux(), "\n", statsdConfig.maxPacketLength(), statsdConfig.pollingFrequency().toMillis())
+                    publisher = BufferingFlux.create(this.sink.asFlux().publish().autoConnect(), "\n",
+                                                     statsdConfig.maxPacketLength(), statsdConfig.pollingFrequency().toMillis())
                             .onBackpressureLatest();
                 } else {
-                    publisher = this.sink.asFlux();
+                    publisher = this.sink.asFlux().publish().autoConnect();
```
OlegDokuka, line 247: actually -  this may be optional. I don't know if we have tests for that but if we do I will try to remove that and put only in cases where we have retry In place

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
OlegDokuka: Alternatively, we can have a bounded Multiproducer queue here since as far as I understand we can drop part of the values

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -185,6 +194,17 @@ void poll() {
...
     public void start() {
         if (started.compareAndSet(false, true)) {
+            if (statsdConfig.enabled()) {
+                Sinks.Many<String> sink = Sinks.unsafe().many().unicast().onBackpressureBuffer(new MpscArrayQueue<>(Queues.SMALL_BUFFER_SIZE));
+
+                if (HAS_TURBO_FILTER) {
+                    sink = new LogbackMetricsSuppressingManySink(sink);
+                }
+
+                this.sink = sink;
```
shakuzen, line 204: Updating `this.sink` here does not update the reference that already registered meters have, which is why some of the `StatsdMeterRegistryPublishTest` tests are failing. For example, when the registry is stopped and then started, `this.sink` will be a new sink, but an already created `StatsdCounter` will still reference the old sink.

PR #2915, opened 2021-12-13
Remove CumulativeHistogramLongTaskTimerTest

izeye:
This PR removes `CumulativeHistogramLongTaskTimerTest` as it seems to be added accidentally in 42af250a6ecfac5bbab4ba41e3d840ef7a38df2d.


shakuzen (approved):
Yeah, we can remove it for now. See https://github.com/micrometer-metrics/micrometer/issues/2744#issuecomment-963885295 for the reason it was added in the corresponding PR. I ended up moving the test for that merged fix to `DefaultLongTaskTimerTest` but didn't remove `CumulativeHistogramLongTaskTimerTest` because I was expecting to follow-up soon with a fix that would have tests added there. That fix hasn't come yet, so we can remove this unused test class for now.

PR #2938, opened 2021-12-31
Use daemon thread in CommonsObjectPool2Metrics

izeye:
This PR changes to use a daemon thread in `CommonsObjectPool2Metrics`.


jonatan-ivanov:
Thanks! :)

Issue #2412, opened 2021-01-20
License header is a JavaDoc comment but probably shouldn't be


shakuzen:
I noticed when browsing [the JavaDocs site for Micrometer](https://www.javadoc.io/doc/io.micrometer/micrometer-core) that package descriptions were populated with the license header. Also, license header in `.java` files is a JavaDoc syntax comment, which the IDE warns about. We should probably change them to all be non-JavaDoc multi-line comments instead. Then we can make meaningful package descriptions for reasons like https://github.com/micrometer-metrics/micrometer-docs/issues/147#issuecomment-763594273 and avoid the IDE warning about JavaDoc comments above imports.

shakuzen:
Resolved by #2940

PR #2940, opened 2022-01-04
Use non-Javadoc multi-line comments for license headers

izeye:
This PR changes to use non-Javadoc multi-line comments for license headers.

See gh-2412


shakuzen (approved):
Thanks for taking care of this. It is something I kept putting off.

PR #2945, opened 2022-01-06
Introduce HttpTagsProvider with default implementation

shakuzen:
This allows us to provide a consistent default set of HTTP tags derived from the abstract request/response API that instrumentation will wrap using their concrete HTTP request/response objects.
It also allows end users (not instrumentors) to provide their own implementation if they want a different set of tags without rewriting the instrumentation provided.

TODO:
- [ ] Tests
- [ ] Replace existing instrumentation in micrometer that could use this.

It's unfortunate it shares the name and purpose of the `TagsProvider` interface but does not extend it because it has parameters specific to it. Perhaps this is something to consider evolving in a more clear way later on.

Example usage of this in instrumentation is like the following. Pass an `HttpTagsProvider` instance to the `HttpHandlerContext` being used with your `Timer.start` calls. For convenience, a constructor without `HttpTagsProvider` is available that will use the default implementation.

```java
HttpClientRequest wrappedRequest = new SpringHttpClientRequest(request);
HttpClientHandlerContext handlerContext = new HttpClientHandlerContext(wrappedRequest, this.tagsProvider);
Timer.Sample sample = Timer.start(this.meterRegistry, handlerContext);

// do an HTTP operation

HttpClientResponse wrappedResponse = new SpringHttpClientResponse(response, wrappedRequest);
handlerContext.setResponse(wrappedResponse);
sample.stop(getTimeBuilder(wrappedRequest, wrappedResponse, null)); // a caught exception can be passed here instead of null
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/HttpTags.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/HttpTags.java
@@ -1,0 +1,90 @@
+/**
+ * Copyright 2021 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.http;
+
+import io.micrometer.core.instrument.Tag;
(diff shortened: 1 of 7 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/HttpTags.java
```diff
@@ -0,0 +1,90 @@
...
+        return (response != null) ? Tag.of("status", Integer.toString(response.statusCode())) : STATUS_UNKNOWN;
+    }
+
+    /**
+     * Creates a {@code exception} tag based on the {@link Class#getSimpleName() simple
+     * name} of the class of the given {@code exception}.
+     * @param exception the exception, may be {@code null}
+     * @return the exception tag derived from the exception
+     */
+    public static Tag exception(Throwable exception) {
```
marcingrzejszczak, line 69: Just fyi In tracing we'll need to remap this. Remove it from tags and add exception to do the span.

jonatan-ivanov (approved):

jonatan-ivanov:
Checkstyle is not happy. :)

jonatan-ivanov (approved):

Issue #2947, opened 2022-01-07
LongTaskTimer always converts durations of measurements to nanos


jonatan-ivanov:
If you call the `measure` method of `LongTaskTimer` and check the duration, it's always in nanoseconds instead of the `baseUnit`.

PR #2948, opened 2022-01-07
Fix duration conversion of LongTaskTimer

jonatan-ivanov:
If you call the measure method of LongTaskTimer and check the duration, it's always in nanoseconds instead of the baseUnit.
It seems the original intention was using the base unit: https://github.com/micrometer-metrics/micrometer/commit/70925b355a6f795f3f10477023fab0b2cdafa72f#diff-551e65260957b90931b406fcb647b39367c83d11aa8f07d3f388c667d60f8971R124
fixes gh-2947

### DIFF
--- a/micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryCompatibilityKit.java
+++ b/micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryCompatibilityKit.java
@@ -40,6 +40,8 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 
 import static io.micrometer.core.instrument.MockClock.clock;
+import static io.micrometer.core.instrument.Statistic.ACTIVE_TASKS;
+import static io.micrometer.core.instrument.Statistic.DURATION;
 import static io.micrometer.core.instrument.util.TimeUtils.millisToUnit;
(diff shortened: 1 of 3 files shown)


micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryCompatibilityKit.java
```diff
@@ -379,6 +381,18 @@ void record() {
...
+                    measurement -> assertThat(measurement).satisfies(m -> {
+                        assertThat(m.getValue()).isEqualTo(1.0);
+                        assertThat(m.getStatistic()).isSameAs(ACTIVE_TASKS);
+                    }),
+                    measurement -> assertThat(measurement).satisfies(m -> {
+                        assertThat(m.getValue()).isEqualTo(TimeUtils.convert(10, TimeUnit.NANOSECONDS, t.baseTimeUnit()));
+                        assertThat(m.getStatistic()).isSameAs(DURATION);
+                    })
+            );
+            sample.stop();
```
shakuzen, line 394: Is this needed? Looks like the sample is stopped 3 lines below.

jonatan-ivanov: 😨

Issue #2944, opened 2022-01-05
"system.cpu.usage" metric changed meaning after JDK upgrade


crmky:
**Describe the bug**
Previously "system.cpu.usage" measures the system level CPU utilization even runs in a containerized environment. JDK changed its behavior in https://bugs.openjdk.java.net/browse/JDK-8265836 (https://github.com/openjdk/jdk8u/blob/master/jdk/src/solaris/classes/sun/management/OperatingSystemImpl.java#L82). Now behavior is very strange:
* If runs in a containerized environment with either quota or share applied (not the default 1024, refer to [this](https://github.com/openjdk/jdk8u/blob/master/jdk/src/linux/classes/jdk/internal/platform/cgroupv1/Metrics.java#L350)), it means container level CPU utilization.
* In other environment, it means system level CPU utilization.

**Environment**

 - Micrometer version: 1.3.20
 - Micrometer registry: prometheus
 - OS: Linux
 - Java version: openjdk version "1.8.0_312"
OpenJDK Runtime Environment (build 1.8.0_312-b07)
OpenJDK 64-Bit Server VM (build 25.312-b07, mixed mode)

**To Reproduce**
How to reproduce the bug:
Runs application in a containerized environment with either quota or share set with:
* JDK 1.8.0_312+
* JDK 11.0.13+

**Expected behavior**
Since this is a JDK behavior change, I think Micrometer can only change metric description to match the behavior change.

**Additional context**
Add any other context about the problem here, e.g. related issues.

jonatan-ivanov:
System in `system.cpu.usage` always mean the "system" where your JVM is running in.
It can be a "native" Docker container, a Docker container inside a VM (Windows, MacOS), a VM (or a VM in a VM) or bare metal. Micrometer does not know about this and uses what the JVM tells it. The behavior of the JVM sounds pretty reasonable to me, I don't find it strange, in fact I found it strange before the fix and the JVM was not container-aware.

Right now the description of `system.cpu.usage` is
>The "recent cpu usage" for the whole system

What would you recommend instead of this?
Something like this?
>The "recent cpu usage" of the system the application is running in

Would you be open to create a PR?

crmky:
[...]

PR #2969, opened 2022-01-11
Update "system.cpu.usage" description to match the updated JVM behavior

crmky:
Updated `system.cpu.usage` metric description to match the changed JVM behavior introduced by https://bugs.openjdk.java.net/browse/JDK-8265836.

The current metrics behavior:
* If application runs in a containerized environment with either quota or share applied (not the default 1024, refer to [this](https://github.com/openjdk/jdk8u/blob/master/jdk/src/linux/classes/jdk/internal/platform/cgroupv1/Metrics.java#L350)), it reports container level CPU utilization.
* In other environment, it reports system level CPU utilization.

Refer to https://github.com/micrometer-metrics/micrometer/issues/2944. cc @jonatan-ivanov


jonatan-ivanov (approved):

jonatan-ivanov:
@crmky Thank you for the PR, could you please rebase it against `1.7.x`?

crmky:
@jonatan-ivanov Rebased against `1.7.x`. Pls have a check. Thanks!

PR #2970, opened 2022-01-11
Drop summary and timer lines with a count of 0 in the Dynatrace exporter

pirgeo:
With this change, Dynatrace metrics lines (summaries and timers) with a count of 0 are dropped on the exporter side.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
@@ -174,6 +174,10 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
 
     private Stream<String> toSummaryLine(Meter meter, HistogramSnapshot histogramSnapshot, TimeUnit timeUnit) {
         long count = histogramSnapshot.count();
+        if (count < 1) {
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -174,6 +174,10 @@ private String createCounterLine(Meter meter, Measurement measurement) {
 
     private Stream<String> toSummaryLine(Meter meter, HistogramSnapshot histogramSnapshot, TimeUnit timeUnit) {
         long count = histogramSnapshot.count();
+        if (count < 1) {
```
jonatan-ivanov, line 177: Is there a difference on the backend between sending a zero measurement and not sending a measurement at all?
E.g.: If the service is down (or the meter got removed from Micrometer) vs. it is up but it is not receiving any traffic.

pirgeo: As far as I am aware, there currently is no difference for summary metrics.

jonatan-ivanov:
I'm not able to rebase this against 1.8.x so I opened a new PR with the rebased commit: https://github.com/micrometer-metrics/micrometer/pull/3030

PR #2972, opened 2022-01-12
Refine SimpleMeterRegistry.getMetersAsString()

izeye:
It prints as follows:

```
answers.count(COUNTER)[correct='true'] 2.0
answers.count(COUNTER)[correct='false'] 1.0
cache.latency.count.seconds(TIMER)[] 5.0
cache.latency.total_time.seconds(TIMER)[] 0.1
cache.miss.count(COUNTER)[] 42.0
custom.meter.value(OTHER)[] 42.0
custom.meter.unknown(OTHER)[] 21.0
handler.active_tasks.seconds(LONG_TASK_TIMER)[] 1.0
handler.duration.seconds(LONG_TASK_TIMER)[] 3.0
latency.count.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 10.0
latency.total_time.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.29
latency.max.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.038
processing.time.value.seconds(GAUGE)[] 0.3
request.size.count.bytes(DISTRIBUTION_SUMMARY)[] 10.0
request.size.total.bytes(DISTRIBUTION_SUMMARY)[] 1450.0
request.size.max.bytes(DISTRIBUTION_SUMMARY)[] 190.0
temperature.value.celsius(GAUGE)[] 24.0
```

"count" or "active_tasks" with "seconds" seems to be a bit confusing.

To try to avoid the confusion, this PR changes it to print as follows:

```
answers(COUNTER)[correct='true']
	count: 2.0
answers(COUNTER)[correct='false']
	count: 1.0
cache.latency.seconds(TIMER)[]
	count: 5.0
	total_time: 0.1
cache.miss(COUNTER)[]
	count: 42.0
custom.meter(OTHER)[]
	value: 42.0
	unknown: 21.0
handler.seconds(LONG_TASK_TIMER)[]
	active_tasks: 1.0
	duration: 3.0
latency.seconds(TIMER)[method='GET', service='test', uri='/api/people']
	count: 10.0
	total_time: 0.29
	max: 0.038
processing.time.seconds(GAUGE)[]
	value: 0.3
request.size.bytes(DISTRIBUTION_SUMMARY)[]
	count: 10.0
	total: 1450.0
	max: 190.0
temperature.celsius(GAUGE)[]
	value: 24.0
```

This PR also adds Javadoc `@since` tag and changes `@Incubating` `since` attribute for it to align with the other places in the codebase.

Please let me know what you think.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/simple/SimpleMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/simple/SimpleMeterRegistry.java
@@ -169,8 +169,9 @@ public class SimpleMeterRegistry extends MeterRegistry {
      * The output is meant to be readable by humans, please do not parse it programmatically because the format can change.
      *
      * @return text representation of the meters in the registry
+     * @since 1.9.0
      */
-    @Incubating(since = "2022-01-06")
+    @Incubating(since = "1.9.0")
     public String getMetersAsString() {
         return this.getMeters().stream()
                 .sorted(Comparator.comparing(meter -> meter.getId().getName()))
@@ -179,25 +180,41 @@ public class SimpleMeterRegistry extends MeterRegistry {
     }
 
     private String toString(Meter meter) {
-        String name = meter.getId().getName();
-        Meter.Type type = meter.getId().getType();
-        String baseUnit = meter.getId().getBaseUnit();
-        String tags = meter.getId().getTags().stream()
+        Meter.Id id = meter.getId();
+        String tags = id.getTags().stream()
                 .map(this::toString)
                 .collect(Collectors.joining(", "));
-
-        return StreamSupport.stream(meter.measure().spliterator(), false)
-                .map(measurement -> toString(name, type, baseUnit, tags, measurement))
-                .collect(Collectors.joining("\n"));
+        String baseUnit = id.getBaseUnit();
+        String meterUnitSuffix = baseUnit != null ? " " + baseUnit : "";
+        String measurements = StreamSupport.stream(meter.measure().spliterator(), false)
+                .map((measurement) -> toString(measurement, meterUnitSuffix))
+                .collect(Collectors.joining(", "));
+        return String.format("%s(%s)[%s]; %s", id.getName(), id.getType(), tags, measurements);
     }
 
     private String toString(Tag tag) {
         return String.format("%s='%s'", tag.getKey(), tag.getValue());
     }
 
-    private String toString(String name, Meter.Type type, @Nullable String baseUnit, String tags, Measurement measurement) {
-        String statistic = measurement.getStatistic().toString().toLowerCase();
-        String unit = baseUnit != null ? "." + baseUnit : "";
-        return String.format("%s.%s%s(%s)[%s] %s", name, statistic, unit, type, tags, measurement.getValue());
+    private String toString(Measurement measurement, String meterUnitSuffix) {
+        Statistic statistic = measurement.getStatistic();
+        return String.format("%s=%s%s",
+                statistic.toString().toLowerCase(),
+                measurement.getValue(),
+                getUnitSuffix(statistic, meterUnitSuffix));
+    }
+
+    private String getUnitSuffix(Statistic statistic, String meterUnitSuffix) {
+        switch (statistic) {
+            case DURATION:
+            case TOTAL_TIME:
+            case TOTAL:
+            case MAX:
+            case VALUE:
+                return meterUnitSuffix;
+
+            default:
+                return "";
+        }
     }
 }
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/simple/SimpleMeterRegistryTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/simple/SimpleMeterRegistryTest.java
@@ -171,24 +171,16 @@ class SimpleMeterRegistryTest {
                 )
         ).register(registry);
 
-        assertThat(registry.getMetersAsString()).isEqualTo(""
-                + "answers.count(COUNTER)[correct='true'] 2.0\n"
-                + "answers.count(COUNTER)[correct='false'] 1.0\n"
-                + "cache.latency.count.seconds(TIMER)[] 5.0\n"
-                + "cache.latency.total_time.seconds(TIMER)[] 0.1\n"
-                + "cache.miss.count(COUNTER)[] 42.0\n"
-                + "custom.meter.value(OTHER)[] 42.0\n"
-                + "custom.meter.unknown(OTHER)[] 21.0\n"
-                + "handler.active_tasks.seconds(LONG_TASK_TIMER)[] 1.0\n"
-                + "handler.duration.seconds(LONG_TASK_TIMER)[] 3.0\n"
-                + "latency.count.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 10.0\n"
-                + "latency.total_time.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.29\n"
-                + "latency.max.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.038\n"
-                + "processing.time.value.seconds(GAUGE)[] 0.3\n"
-                + "request.size.count.bytes(DISTRIBUTION_SUMMARY)[] 10.0\n"
-                + "request.size.total.bytes(DISTRIBUTION_SUMMARY)[] 1450.0\n"
-                + "request.size.max.bytes(DISTRIBUTION_SUMMARY)[] 190.0\n"
-                + "temperature.value.celsius(GAUGE)[] 24.0");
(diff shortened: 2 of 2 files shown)


jonatan-ivanov:
>"count" or "active_tasks" with "seconds" seems to be a bit confusing.

You are totally right about this, in fact I had a version that separated the Measurements and the Meters in a similar way you did. Let me give you some background why I ended-up with this.

We are going to write/rewrite instrumentation for a lot of things in the future and I wanted to have an easy and on-demand way to quickly see what is in the registry so that we can check what the instrumentation does under the hood. Right now you can use the Prometheus registry (or the debugger) for this but these are not as convenient as printing them out to stdout.

I wanted to keep the output as close to the memory representation as possible, also, I wanted to have one `Measurement`/time-series per line so that it is easy to manually diff these and we get 1:1 mapping about the cardinality. So there are 3 lines per Timer since it will produce 3 measurements as well as 3 time-series.

I left the unit on the measurement to be close to the memory representation. This is unfortunate since the Meter has the unit not the Measurement.

I have two ideas to resolve this while keeping an easy-to diff, one-line per Meter/Measurement representation:

Original:
```
latency.count.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 10.0
latency.total_time.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.29
latency.max.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.038
```

We can maintain a set for the unitless measurements and only add it if it makes sense:
```
latency.count(TIMER)[method='GET', service='test', uri='/api/people'] 10.0
latency.total_time.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.29
latency.max.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.038
```

Or we can have one line per Meter:
latency.seconds(TIMER)[method='GET', service='test', uri='/api/people']; count=10.0, total_time=0.29, max=0.038

I think I like this even more, since it is easier to see through what you registered though percentiles, sla and automatic histogram buckets can make this a little crowded but histogram buckets are not supported right now and the point of this is providing a "debug" output so I guess it should be fine. What do you think?

shakuzen:
I'm wondering if the unit for the meter should be concatenated to the name to begin with. I think it would be nice to have a clear distinction in the output between the meter name and the unit. In that sense, I guess I like the clarity of the one line per meter in that the measurement name isn't concatenated to the meter name which can confuse people to think a meter with such a name is registered to the registry.

Perhaps something like the following which reads nicely:
```
latency(TIMER)[method='GET', service='test', uri='/api/people']; count=10.0, total_time=0.29 seconds, max=0.038 seconds
```

izeye:
@jonatan-ivanov @shakuzen Thanks for the feedback!

The representation suggested by @shakuzen looks good to me. I can try to update this PR in that direction.

I'm wondering if you might want to share the `String` representation with the `LoggingMeterRegistry` which seems to have a good representation already.

jonatan-ivanov:
This is what I get if I run this through the `LoggingMeterRegistry`:
```
answers{correct=true} throughput=1/s
answers{correct=false} throughput=0.5/s
cache.miss{} throughput=21/s
processing.time{} value=0.3s
temperature{} value=24 celsius
handler{} active=1 milliseconds duration=1.323893618s
cache.latency{} throughput=2.5 milliseconds/s mean=0.02s
latency{method=GET,service=test,uri=/api/people} throughput=5/s mean=0.029s max=0.038s
request.size{} throughput=5/s mean=145 B max=190 B
custom.meter{} value=42, unknown=21
```

I like the idea of consolidating the two, however there are some differences that might make this harder (at least in 1.x):
- `LoggingMeterRegistry` does not contain the type of the meter, for debugging this would be great to have
- `LoggingMeterRegistry`does not show the "raw" values e.g.: it seems it has a rate other than a count, there is no sum but there is an avg which makes sense if you want to ship this data somewhere but it is less useful for debugging
- I like how the logging registry handles units, I think `SimpleMeterRegistry` could borrow that though `handler{} active=1 milliseconds duration=1.323893618s` is weird (active is the number of active tasks)

Based on this I guess the two are different enough to keep their output different. I like Tommy's representation, do you want to implement that?

izeye:
@jonatan-ivanov Thanks for the feedback! I gave it a try as suggested. Please let me know if there's anything that needs to be updated.

micrometer-core/src/test/java/io/micrometer/core/instrument/simple/SimpleMeterRegistryTest.java
```diff
@@ -171,24 +171,16 @@ void stringRepresentationOfMetersShouldBeOk() {
...
-                + "latency.total_time.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.29\n"
-                + "latency.max.seconds(TIMER)[method='GET', service='test', uri='/api/people'] 0.038\n"
-                + "processing.time.value.seconds(GAUGE)[] 0.3\n"
-                + "request.size.count.bytes(DISTRIBUTION_SUMMARY)[] 10.0\n"
-                + "request.size.total.bytes(DISTRIBUTION_SUMMARY)[] 1450.0\n"
-                + "request.size.max.bytes(DISTRIBUTION_SUMMARY)[] 190.0\n"
-                + "temperature.value.celsius(GAUGE)[] 24.0");
+        assertThat(registry.getMetersAsString()).isEqualTo("answers(COUNTER)[correct='true']; count=2.0\n" +
+                "answers(COUNTER)[correct='false']; count=1.0\n" +
+                "cache.latency(TIMER)[]; count=5.0, total_time=0.1 seconds\n" +
```
shakuzen, line 176: I was wondering why this didn't have a max, and it's because it is a `FunctionTimer`. This brings up the question of do we want that level of information available in the output or not. We don't have to answer that for this pull request, but it's something to consider.

shakuzen (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/simple/SimpleMeterRegistry.java
```diff
@@ -179,25 +180,51 @@ public String getMetersAsString() {
...
+        return String.format("%s=%s%s",
+                statistic.toString().toLowerCase(),
+                measurement.getValue(),
+                getUnitSuffix(statistic, meter, meterUnitSuffix));
+    }
+
+    private String getUnitSuffix(Statistic statistic, Meter meter, String meterUnitSuffix) {
+        switch (statistic) {
+            case DURATION:
+            case TOTAL_TIME:
```
jonatan-ivanov, line 210: Should we add `TOTAL` too so that `DistributionSummary` can have the same functionality?

micrometer-core/src/main/java/io/micrometer/core/instrument/simple/SimpleMeterRegistry.java
```diff
@@ -179,25 +180,51 @@ public String getMetersAsString() {
...
+                return meterUnitSuffix;
+
+            case MAX:
+                if (meter instanceof Timer) {
```
jonatan-ivanov, line 214: Should we add `DistributionSummary` here too?
Or do we need the `instaceof` check at all since if there is a max of something that has a unit we probably what to show what that unit was?

micrometer-core/src/main/java/io/micrometer/core/instrument/simple/SimpleMeterRegistry.java
```diff
@@ -179,25 +180,51 @@ public String getMetersAsString() {
...
+                    return meterUnitSuffix;
+                }
+                break;
+
+            case VALUE:
+                if (meter instanceof TimeGauge) {
```
jonatan-ivanov, line 220: Should we add `Gauge` here too?
Or do we need the `instaceof` check at all since if there is a value of something that has a unit we probably what to show what that unit was?

micrometer-core/src/main/java/io/micrometer/core/instrument/simple/SimpleMeterRegistry.java
```diff
@@ -179,25 +180,51 @@ public String getMetersAsString() {
...
+                    return meterUnitSuffix;
+                }
+                break;
+
+            default:
+                break;
+        }
+        return "";
     }
 }
```
jonatan-ivanov, line 230: With the comments above, how about something like this?

```java
    private String getUnitSuffix(Statistic statistic, String meterUnitSuffix) {
        switch (statistic) {
            case DURATION:
            case TOTAL_TIME:
            case TOTAL:
            case MAX:
            case VALUE:
                return meterUnitSuffix;
            default:
                return "";
        }
    }
}
```

izeye:
@jonatan-ivanov Thanks! I updated as suggested.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you for improving this! :)

PR #2977, opened 2022-01-14
Add capability to correctly export fixed buckets histogram to SignalFx

bogdandrutu:
* Fix to always include the "+Inf" bucket.
* Fix to never expire per bucket counters.
* The metrics are still reported as Gauges by default, to fix this user has to enable a property.


mateuszrzeszutek (approved):

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -170,23 +192,26 @@ protected void publish() {
...
     }
 
-    private Stream<SignalFxProtocolBuffers.DataPoint.Builder> addTimeGauge(TimeGauge timeGauge) {
+    Stream<SignalFxProtocolBuffers.DataPoint.Builder> addTimeGauge(TimeGauge timeGauge) {
         return Stream.of(addDatapoint(timeGauge, GAUGE, null, timeGauge.value(getBaseTimeUnit())));
     }
 
-    private Stream<SignalFxProtocolBuffers.DataPoint.Builder> addGauge(Gauge gauge) {
+    Stream<SignalFxProtocolBuffers.DataPoint.Builder> addGauge(Gauge gauge) {
+        if (fixHistogramBucketsType && gauge.getId().getName().endsWith(".histogram")) {
```
mateuszrzeszutek, line 200: Nit: the gauges created by `HistogramGauges` are always synthetic, adding that to the condition can make it a bit more foolproof:
```suggestion
        if (fixHistogramBucketsType
                && gauge.getId().syntheticAssociation() != null
                && gauge.getId().getName().endsWith(".histogram")) {
```

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxConfig.java
```diff
@@ -42,6 +42,15 @@ default String accessToken() {
...
     }
 
+    /**
+     * Returns {@code true} if the signalfx registry should emit histogram buckets.
+     *
+     * @return {@code true} if the signalfx registry should emit histogram buckets.
+     */
+    default Boolean fixHistogramBucketsType() {
+        return getBoolean(this, "reportHistogramBuckets").orElse(false);
+    }
```
jonatan-ivanov, line 52: Based on the implementation, it seems this does not publish "percentile histograms" but a "minimum histogram" containing the defined Service Level Objective (SLO) boundaries.

Can we match the name of the key, the name of the method, the javadoc and the variables in the registry + tests to this?

```suggestion
    /**
     * Wether to publish a minimum a histogram containing your defined service level objective (SLO) boundaries.
     *
     * @return {@code true} if the SignalFx registry should emit SLO buckets, {@code false} otherwise.
     */
    default Boolean publishSlos() {
        return getBoolean(this, "publishSlos").orElse(false);
    }
```

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -216,4 +243,30 @@ protected void publish() {
...
+            return distributionStatisticConfig;
+        }
+        double[] newSLA = sla;
+        if (!isPositiveInf(sla[sla.length - 1])) {
+            newSLA = Arrays.copyOf(sla, sla.length + 1);
+            newSLA[newSLA.length - 1] = Double.MAX_VALUE;
+        }
+        return DistributionStatisticConfig.builder()
+                // Set the expiration duration for the histogram counts to be effectively a lifetime.
+                // Without this, the counts are reset every expiry duration.
```
jonatan-ivanov, line 261: SignalFx can handle this for histogram buckets, right (including server restart and cumulative value reset)?

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -216,4 +243,30 @@ protected void publish() {
...
+                .expiry(Duration.ofNanos(Long.MAX_VALUE)) // effectively a lifetime
```
jonatan-ivanov, line 262: This has an effect on `max`:
Timers usually use `TimeWindowMax` to produce a _decaying_ max value. This change disables this decaying behavior and a `max` value that shows the highest recorder value since application has started is not very useful.

jonatan-ivanov (changes requested):

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
bogdandrutu: Yes signalfx can handle "Cumulative Counters". Ideally, would prefer the bucket counts to reset every "report" (not every x time), essentially delta counters, but that is not the behavior I can get, so this is good enough.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -216,4 +243,30 @@ protected void publish() {
     protected TimeUnit getBaseTimeUnit() {
         return TimeUnit.SECONDS;
     }
+
+    private DistributionStatisticConfig updateConfig(
+            DistributionStatisticConfig distributionStatisticConfig) {
+        // Add the +Inf bucket since the "count" resets every export.
+        double[] sla = distributionStatisticConfig.getServiceLevelObjectiveBoundaries();
```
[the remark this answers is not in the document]
bogdandrutu: Done.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -216,4 +243,30 @@ protected void publish() {
     protected TimeUnit getBaseTimeUnit() {
         return TimeUnit.SECONDS;
     }
+
+    private DistributionStatisticConfig updateConfig(
+            DistributionStatisticConfig distributionStatisticConfig) {
+        // Add the +Inf bucket since the "count" resets every export.
```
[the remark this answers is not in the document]
bogdandrutu: Done.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
bogdandrutu: I understand, but this is a problem for the histogram buckets. There is not an easy way to have both of these to reset when "report" so, I am going with the least problematic solution which indeed makes max useless.

@mateuszrzeszutek Do you think we should remove the "max" metric in this case since it is useless?

mateuszrzeszutek: `DistributionSummary` uses `TimeWindowMax` too, right?

> @mateuszrzeszutek Do you think we should remove the "max" metric in this case since it is useless?

Why not, it is useless after all. I don't really have a strong opinion about it though, if there's any reason for leaving it as it is (convention?) then I'm fine with keeping it too.

mateuszrzeszutek:
@jonatan-ivanov (and Micrometer maintainers) PTAL at this PR -- I've fixed the issues with `max()`, and made sure that all new behavior is hidden behind a new config flag.

jonatan-ivanov (approved):

jonatan-ivanov:
@bogdandrutu @mateuszrzeszutek Thank you for the change!
I rebased the branch on top of `1.9.x` so that this can go out with our next minor release (`main` is for 2.0.0 now).

PR #2994, opened 2022-02-04
Improve cacheability ignoring Java vendor and minor version in Nebula's generated manifest

jprinet:
The _Manifest_ created by [Nebula plugins](https://github.com/nebula-plugins/gradle-info-plugin) has some properties depending on _Java vendor_ and _minor version_.

An example in _micrometer-test/build/manifest/micrometer-test.properties_:
```
Created-By=11.0.12+7 (Eclipse Foundation)
Build-Java-Version=11.0.12
```

This leads to cache misses when using a different _Java vendor / minor version_ than what was used for the build which fed the cache in the first place.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -127,7 +127,9 @@ subprojects {
                         'Change',
                         'Full-Change',
                         'Branch',
-                        'Module-Origin'
+                        'Module-Origin',
+                        'Created-By',
+                        'Build-Java-Version'
                     ].each {
                         ignoreAttribute it
                         ignoreProperty it


jonatan-ivanov:
A few notes about this change:

Since we are testing multiple Java vendors and versions, this can be a potential issue, it can get us into false positive test results, e.g.: something works on one JVM and not on others but since the task is cached, it will not be re-executed so the issue won't be surfaced.

This is not an issue right now (I talked to @jprinet and also tested it) since:
1. We only have Java 8-11-17 tests right now from one vendor (Adopt), no OpenJ9 tests on CI
2. Gradle tracks the major Java version as the input of compilation and test tasks

This means that these tasks won't be cached between major versions so our Java 8-11-17 tests are re-executed and we are ok.

Once we add tests for OpenJ9 we need to check this behavior again, I _think_ it should work as of today since:
1. OpenJ9 tests are separate tasks so they are already separated from other tests (except if we want to run all tests against OpenJ9)
2. There is a better way to tell Gradle that we want to re-run test tasks for different versions/vendors: https://docs.gradle.org/current/userguide/common_caching_problems.html#java_version_tracking (thanks for @jprinet)

jonatan-ivanov:
Thanks @jprinet for fixing this and your help around the problem space! :)

PR #2997, opened 2022-02-06
Use elastic create operation instead of index 

### LINKED ISSUE #2996: Support writing to elastic data streams
Currently micrometer cannot write to elastic data streams because micrometer uses the `index` operation.

From the docs: 

> A data stream lets you store append-only time series data across multiple indices while giving you a single named resource for requests. Data streams are well-suited for logs, events, metrics, and other continuously generated data.

Data streams have been out for a while now and are the preferred mechanism for write-only, time series data because they support Index Lifecycle Management, which provides automation around index management. 

The current barrier to micrometer writing to data streams is that [requests must use an `op_type` of `create`](https://www.elastic.co/guide/en/elasticsearch/reference/current/use-a-data-stream.html#add-documents-to-a-data-stream)

Luckily this is a simple change. Currently micrometer uses the `index` operation, which is analogous to a PUT or upsert. Semantically speaking, the `create` operation is a better fit since micrometer should always be creating new documents. So, the solution is to use the `create` operation, which is a better semantic fit and works with data streams.

jeremyross:
`create` makes more sense and allows micrometer to write to elastic data streams.

Closes #2996

### DIFF
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
@@ -121,7 +121,7 @@ public class ElasticMeterRegistry extends StepMeterRegistry {
 
     private final DateTimeFormatter indexDateFormatter;
 
-    private final String createLine;
+    private final String actionLine;
 
     private volatile boolean checkedForIndexTemplate;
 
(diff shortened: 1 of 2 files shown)


shakuzen:
Is this a breaking change for users not using data streams?

shakuzen:
I don't know why the CI wasn't run for this pull request, but unit tests need to be updated for this change. They are failing now.

patpatpat123:
Hello @jeremyross ,

Hope this message finds you well.
First of all, if not anything else, big thanks for this PR!

I took the latest builds of micrometer, from 1.10 all the way to the latest milestone.
I cross-tested with Elastic cluster from 7.6, all the way to the latest 8.7.

I am observing the following: those are facts and observations:

- test case 1: with this change, if the index already exists, and is of type _plain index_, micrometer will publish the metrics inside this _plain index_.
- test case 2: with this change, if the index already exists, and is of type _data stream_, micrometer will publish the metrics inside this _data stream_.
- test case 3: with this change, if the index does **not** already exist, meaning the index does not even exist in the first place, neither plain index, neither data stream, micrometer will publish the metrics, and a _plain index_ will be created.

Above, for test cases 1, 2, and 3, by _plain index_, I mean this: (please see screenshot)



and by _data stream_ I mean this (please see screenshot)


(small nota, your change is great for test case 2. Without your/this change, test case 2 would fail, I just verified)

Just wanted to confirm with you this is your expectation for the above test cases.

I am reaching out because for test 3, an expected enhancement/feature would be:

if the index does **not** already exist, meaning the index does not even exist in the first place, neither plain index, neither data stream, micrometer will publish the metrics, and a _time series data stream_ will be created, that would be a great feature.

What do you think?

Thank you for your time reading me.

PR #3008, opened 2022-02-08
Polishing pass for 2.x

jonatan-ivanov:


micrometer-api/src/main/java/io/micrometer/api/instrument/MeterRegistry.java
```diff
@@ -92,10 +96,9 @@
 
     private final ObservationConfig observationConfig = new ObservationConfig();
 
-    @Nullable
     @Override
-    public Observation getCurrentObservation() {
-        return localObservation.get();
+    public Optional<Observation> getCurrentObservation() {
```
jonatan-ivanov, line 100: This is an important difference, I changed `MeterRegistry::getCurrentObservation` so that it returns an `Optional` instead of being `@Nullable`.

micrometer-api/src/main/java/io/micrometer/api/instrument/observation/Observation.java
```diff
@@ -227,9 +226,8 @@ default void scoped(Runnable action) {
         /**
          * Current observation available within this scope.
          *
-         * @return current observation or {@code null} if one is not present
+         * @return current observation that this scope was created by
          */
-        @Nullable
```
jonatan-ivanov, line 232: Since the `Scope` can only be created by an `Observation`, it should not be possible that this is null.

micrometer-api/src/main/java/io/micrometer/api/instrument/observation/NoopObservation.java
```diff
@@ -13,11 +13,19 @@
...
+/**
+ * No-op implementation of {@link Observation} so that we can disable the instrumentation logic.
+ *
+ * @author Jonatan Ivanov
+ * @author Tommy Ludwig
+ * @author Marcin Grzejszczak
+ *
+ * @since 2.0.0
+ */
 public class NoopObservation implements Observation {
```
shakuzen, line 29: Fine to consider this separate from this PR, but:
Is there any point in this being public API? We aren't expecting people to extend this, are we?

micrometer-api/src/main/java/io/micrometer/api/instrument/MeterRegistry.java
shakuzen: What's the motivation? Won't it make more garbage while callers still have to deal with the absent case?

micrometer-api/src/main/java/io/micrometer/api/instrument/observation/NoopObservation.java
jonatan-ivanov: I need to take a look but I think it is public because the `MeterRegistry` is using it (we can make it final `though`).

micrometer-api/src/main/java/io/micrometer/api/instrument/MeterRegistry.java
jonatan-ivanov: Simply convenience, this can be helpful dealing with the case when the current observation is not available. If you prefer `@Nullable` instead, I'm totally fine rolling this back.

shakuzen (approved):

Issue #2513, opened 2021-03-18
Apache HttpClient 5.x support


shakuzen:
A new major version (5) of the Apache HttpClient is available. It uses different package names for classes, so our support for major version 4 will not work with it. The implementation should be similar, if not the same, just with different packages.

ginkel:
@shakuzen Do you have any preferences regarding the implementation (code-copy vs. sharing code w/ the v4 implementation)? We are missing this feature and could possibly offer a contribution.

shakuzen:
Thank you for the offer to help. I have not looked into it enough to know how much could potentially be reused between the two. Generally we want to avoid a bug being fixed in one but not the other because of duplicated code, but we also don't want to overly complicate the code for the sake of reducing duplication either. So it's hard to say without taking a closer look. If you don't mind, feel free to make a judgement call on it and propose something in a pull request, and we can more easily decide with something concrete.

shakuzen:
Thank you for starting the work @tommyk-gears. We have merged a copy of the Apache HttpClient 4 support adapted for changes in Apache HC 5. I still believe having the instrumentation closer to the library would be a better experience for users (this issue would not have been needed because the instrumentation would presumably have been updated when the breaking changes were made from version 4). Unfortunately, no Apache HttpClient users have requested Micrometer support in the Apache HttpClient project that I'm aware of. The Micrometer team would be happy to advise and review any work to do such, but in the meantime, to unblock users of Apache HttpClient 5, we have made something available in Micrometer.

flozano:
any chance to have this backported to a stable release? the current version of spring stack uses apache http client 5 and it becomes impossible to have  proper metrics for an apache-client-backed RestRemplate with spring 6

jonatan-ivanov:
@flozano I don't think this should or will be backported but it will be released in 1.11.0 (planned release date: 2023-05-08).

flozano:
I see. The 1.11.0 release seems to be coming pretty soon, so it is probably OK.

Thanks for considering it.

PR #3009, opened 2022-02-09
Adding support for Apache HttpClient 5

tommyk-gears:
Adding support for Apache HttpClient 5 by simply copying and adapting the old Apache HttpClient support.

The intention of this PR is merely to have something to discuss #2513 around.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -43,6 +43,7 @@ def VERSIONS = [
         'net.sf.ehcache:ehcache:latest.release',
         'org.apache.httpcomponents:httpasyncclient:latest.release',
         'org.apache.httpcomponents:httpclient:latest.release',
+        'org.apache.httpcomponents.client5:httpclient5:latest.release',
         'org.apache.kafka:kafka-clients:2.+',
         'org.apache.kafka:kafka-streams:2.+',
         'org.apache.logging.log4j:log4j-core:2.+',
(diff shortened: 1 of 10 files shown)


shakuzen:
Thanks for the pull request. I'm wondering how the Apache HttpClient project would feel about maintaining metrics instrumentation code in their project, so that the instrumentation would be closer to the code, which would allow for easier compatibility maintenance (for example, I assume support for 5.x would already be there if the 4.x instrumentation lived with the project code). There should also be better ways to get the URI pattern if instrumentation is in the HttpClient code. In our implementation it's on the user to pass the pattern some way (e.g. in a header) to make that available.
What do you think? Would you be willing to start a discussion/enhancement request with the project maintainers there before we consider adding duplicated (from the 4.x implementation) code here?

daromnik:
Hello, will you merge this to main branch ?

nezygis:
@shakuzen Hi, any update on HttpClient 5 support?

shakuzen:
Unfortunately the implementation that this was copied from has been updated in the meantime to support the new `Observation` API. If we add support for Apache HC5, we would support the `Observation` API. This demonstrates the kind of drift and maintenance hassle of duplicated code that I mentioned in my previous comment.

It seems no one has reached out to the Apache HttpClient project requesting support for Micrometer be added there as I suggested in my previous comment here. I think it is best when such requests come from users rather than another library maintainer. We can support this in Micrometer for now.

I've opened #3598 to continue the work.

PR #3019, opened 2022-02-10
Integrate exemplars in counter and histogram. Take account of client sampling in counter

fscellos:
This PR is to add support of exemplars (with trace id) in micrometer-prometheus.

If code is instrumented with OpenTelemertey (supported by prometheus client in this implementation), exemplars with trace_id will be generated if correct header is set when requested this data (which is done by prometheus).

This PR also integrate an ExemplarSampler that is compliant with  sampling set on client side (throught Instrumentation CRD of Otel kubernetes operator for exemple). If trace is not sampled, this one won't be eligible to be choosen as examplars as it could give some unknow trace in storage backend for a navigation from a phi-percentile time series to trace system.

### DIFF
--- a/implementations/micrometer-registry-prometheus/build.gradle
+++ b/implementations/micrometer-registry-prometheus/build.gradle
@@ -4,4 +4,6 @@ dependencies {
     api 'io.prometheus:simpleclient_common'
 
     testImplementation project(':micrometer-test')
+	
+	api 'io.opentelemetry:opentelemetry-api:1.11.0'
 }
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
@@ -14,17 +14,24 @@
  * limitations under the License.
  */
 package io.micrometer.prometheus;
+import static java.util.stream.Collectors.toList;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Stream;
 
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.config.NamingConvention;
 import io.prometheus.client.Collector;
 
-import java.util.*;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Stream;
-
-import static java.util.stream.Collectors.toList;
 
 /**
  * {@link Collector} for Micrometer.
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/OtelMicroMeterSpanContextSupplier.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/OtelMicroMeterSpanContextSupplier.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2020 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.prometheus;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanId;
+import io.opentelemetry.api.trace.TraceId;
+import io.prometheus.client.exemplars.tracer.common.SpanContextSupplier;
+import io.prometheus.client.exemplars.tracer.otel.OpenTelemetrySpanContextSupplier;
+
+// A context supplier that also give information about sampling or not of current trace
+public class OtelMicroMeterSpanContextSupplier implements SpanContextSupplier {
+
+    public static boolean isAvailable() {
+        try {
+            if ("inactive".equalsIgnoreCase(System.getProperties().getProperty("io.prometheus.otelExemplars"))) {
+                return false;
+            }
+            OpenTelemetrySpanContextSupplier test = new OpenTelemetrySpanContextSupplier();
+            test.getSpanId();
+            test.getTraceId();
+            return true;
+        } catch (LinkageError ignored) {
+            // NoClassDefFoundError:
+            // Either OpenTelemetry is not present, or it is version 0.9.1 or older when
+            // io.opentelemetry.api.trace.Span did not exist.
+            // IncompatibleClassChangeError:
+            // The application uses an OpenTelemetry version between 0.10.0 and 0.15.0 when
+            // SpanContext was a class, and not an interface.
+            return false;
+        }
+    }
+
+    @Override
+    public String getTraceId() {
+        String traceId = Span.current().getSpanContext().getTraceId();
+        return TraceId.isValid(traceId) ? traceId : null;
+    }
+
+    @Override
+    public String getSpanId() {
+        String spanId = Span.current().getSpanContext().getSpanId();
+        return SpanId.isValid(spanId) ? spanId : null;
+    }
+
+    public boolean isSampled() {
+        return Span.current().getSpanContext().isSampled();
+    }
+
+}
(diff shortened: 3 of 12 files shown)


fscellos:
Reference to #2811 #2812

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusCounter.java
```diff
@@ -19,20 +19,59 @@
...
+     * sampler used is the standard one from micrometer that apply a time windows
+     * around 7 secondes for change
+     */
+    private void updateExemplar(double amount, @NonNull CounterExemplarSampler exemplarSampler) {
+        Exemplar prev;
+        Exemplar next;
+        do {
+            prev = exemplar.get();
+            next = exemplarSampler.sample(amount, prev);
+            if (next == null || next == prev) {
```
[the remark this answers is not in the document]
fscellos: @sonatype-lift ignore

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/OtelMicroMeterSpanContextSupplier.java
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.prometheus;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanId;
+import io.opentelemetry.api.trace.TraceId;
+import io.prometheus.client.exemplars.tracer.common.SpanContextSupplier;
+import io.prometheus.client.exemplars.tracer.otel.OpenTelemetrySpanContextSupplier;
```
jonatan-ivanov, line 22: Why would we need this?
We should not depend on OTel-specific things, we only want to depend on Prometheus' `SpanContextSupplier` so that users can setup the implementation of their choice (Otel, Brave, Sleuth, etc.).

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
```diff
@@ -14,17 +14,24 @@
  * limitations under the License.
  */
 package io.micrometer.prometheus;
+import static java.util.stream.Collectors.toList;
```
jonatan-ivanov, line 17: Can you revert this?

implementations/micrometer-registry-prometheus/build.gradle
```diff
@@ -4,4 +4,6 @@ dependencies {
     api 'io.prometheus:simpleclient_common'
 
     testImplementation project(':micrometer-test')
+	
+	api 'io.opentelemetry:opentelemetry-api:1.11.0'
```
jonatan-ivanov, line 8: This is not needed, see below.

jonatan-ivanov (changes requested):

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/OtelMicroMeterSpanContextSupplier.java
fscellos: Hello @jonatan-ivanov.
I introduce this because prometheus java client also use this implementation too to give use spancontextsupplier "real" implementation (here : https://github.com/prometheus/client_java/blob/master/simpleclient_tracer/simpleclient_tracer_otel/src/main/java/io/prometheus/client/exemplars/tracer/otel/OpenTelemetrySpanContextSupplier.java) 

What is not done in prometheus client is to take account of Sampling and that's what i add in this above implementation of SpanContextSupplier. The goal is not use trace id that will be not recorded to only keep useful trace on backend (isSample take account of TraceFlags to indicate above service that this trace won't be recorded).

When we use Prometheus's SpanContextSupplier, transitively we use io.opentelemetry package too in fact. So i don't know how sleuth (that use brave) work but it's mandatory to these to also have this transitive dependency to open telemetry packages.

I'm agree that is prometheus client integrate "issamppled" in their code, this dependence would be less explicit. But it exists anyway.

implementations/micrometer-registry-prometheus/build.gradle
fscellos: Hellà @jonatan-ivanov. See also my answer on your other comment.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
fscellos: Yes i will do that as soon as i come back to my office.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/OtelMicroMeterSpanContextSupplier.java
fscellos: I also take a look at current Brave (so sleuth) implementation to be sure.

Actually, Brave depends on OpenTracing API (https://github.com/openzipkin/brave/blob/master/brave/pom.xml) but OpenTracing is also deprecated  in benefice of OpenTelemetry (https://github.com/opentracing/specification/issues/163)

These APIs (OpenTracing) will be archived at the end of this year. Project Brave (and indeed Sleuth) will have to remove this dependency.

fscellos: I create an issue on prometheus client for this case : I create an issue on prometheus client for this case : https://github.com/prometheus/client_java/issues/765 (it is like russian doll to get this feature :-))

jonatan-ivanov: >I introduce this because prometheus java client also use this implementation too to give use spancontextsupplier "real" implementation
I think implementations should stay in their respective projects: OTel's `SpanContextSupplier` should be in OTel, Brave's `SpanContextSupplier` should be in Brave, Sleuth's `SpanContextSupplier` should be in Sleuth and so on. We can't have `SpanContextSupplier` implementations for every single tracing library.

>What is not done in prometheus client is to take account of Sampling
I guess if there is an issue with  Prometheus's SpanContextSupplier, that should be fixed in Prometheus not in Micrometer.

>When we use Prometheus's SpanContextSupplier, transitively we use io.opentelemetry package too in fact.
I don't think so: `io.prometheus.client.exemplars.tracer.common.SpanContextSupplier`. You can take a look at my draft for Counter exemplars, it does not contain OTel.

>So i don't know how sleuth (that use brave) work but it's mandatory to these to also have this transitive dependency to open telemetry packages.
I think we have a huge gap here. I don't see why OTel would be mandatory. What if I'm (as a user) not using it at all?

>Actually, Brave depends on OpenTracing API
The scope of that dependency is `test` so Brave uses it for testing something. :)

>These APIs (OpenTracing) will be archived at the end of this year. Project Brave (and indeed Sleuth) will have to remove this dependency.
Brave uses it for testing, Sleuth does not depend on it.

fscellos: That's ok @jonatan-ivanov .
I have a PR in progress on prometheus project to remove this dependency.
I take you inform.

fscellos: Hello @jonatan-ivanov.
Yes with this PR, nothing to do in micrometer (just some things to remove in my current PR for micrometer, like removing explicit dependencies to opentelemetry as you ask for).
But without it, we'll get "incorrect" TraceId in exemplars if sampling is applied (and i think everyone sample for traces).

fscellos: I'm not sur to clearly understand your question. I think yes, we must return null from SpanContextSupplier of prometheus client in this case. It is the goal of the PR i submit to java prometheus project.

jonatan-ivanov: I mean if the `SpanContextSupplier` does this: [see SleuthSpanContextSupplier](https://github.com/spring-cloud/spring-cloud-sleuth/blob/3.1.x/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/prometheus/prometheus/SleuthSpanContextSupplier.java), we don't need to wait till Prometheus merges your changes and releases a new version, we can go ahead with the Micrometer support as is.

fscellos: Yes it should work too. If you want i can test it next week. But we have to keep OtelMicroMeterSpanContextSupplier class (without sample method) and dependencies to otel api in this class. 
Are you agree with that ? (it seems to me that you don't want to have dependencies on otel api at all, it's why i created PR on prometheus java client side)

shakuzen:
We ended up integrating support for exemplars with a different approach, but we appreciate the pull request and work you put in on this. If you have time and interest, please do try out what we've integrated and give us any feedback on how it works (or doesn't).

fscellos:
Hello. Thank you. But i don't undertstand how you can manage sampling on trace as my pr on prometheus java client haven't be merge yet.

jonatan-ivanov:
@fscellos Please see my answer here: https://github.com/micrometer-metrics/micrometer/issues/2812#issuecomment-1079250690

Issue #3007, opened 2022-02-08
Dynatrace v2 ingestion reporting "inconsistent gauge fields"


stevegaeke-kr:
**Describe the bug**
Dynatrace is reporting an inconsistent gauge error due to the measurement count being zero, but a maximum value is still being reported.

```
ERROR i.m.dynatrace.v2.DynatraceExporterV2 - Failed metric ingestion: Error Code=400, Response Body={"linesOk":122,"linesInvalid":1,"error":{"code":400,"message":"1 invalid lines","invalidLines":[{"line":90,"error":"inconsistent gauge fields: count == 0 => min, max, sum == 0: min: 0.000000, max: 1642.000000, sum: 0.000000, count: 0"}]},...
```

It seems like the `DynatraceExporterV2` class can be enhanced to protect against this situation similar to the checks that are performed on `min` to keep it in bounds.

This is the routine I believe could be enhanced to prevent reporting max when count is zero.

```
    private Stream<String> toSummaryLine(Meter meter, HistogramSnapshot histogramSnapshot, TimeUnit timeUnit) {
        long count = histogramSnapshot.count();
        double total = timeUnit != null ? histogramSnapshot.total(timeUnit) : histogramSnapshot.total();
        double max = timeUnit != null ? histogramSnapshot.max(timeUnit) : histogramSnapshot.max();
        double min = count == 1L ? max : this.minFromHistogramSnapshot(histogramSnapshot, timeUnit);
        return this.createSummaryLine(meter, min, max, total, count);
    }
```

I'm assuming the cause is related to the note found in this section: https://micrometer.io/docs/concepts#_timers

Is the period of the Dynatrace export job synchronized with any max value clearing?

**Environment**

 - Micrometer version 1.8.1
 - Micrometer registry Dynatrace v2
 - OS: Linux, MacOS
 - Java version: 11.0.7

**To Reproduce**
How to reproduce the bug:

In SpringBoot, create a Time using:

```
    messageProcessingDuration = Timer.builder(MESSAGE_PROCESSING_DURATION)
        .register(meterRegistry);

   // Later, call messageProcessingDuration.record(duration);
   // And then wait for 2 reporting cycles. First it reports the max with count = 1.
   // In the subsequent cycle, the count becomes 0 but the max continues to be reported.
```

**Expected behavior**

If the gauge count == 0, no statistics should be reported to avoid Dynatrace reporting inconsistent gauge errors.

**Additional context**
[...]

PR #3030, opened 2022-02-15
Drop summary and timer lines with a count of 0

jonatan-ivanov:
> With this change, Dynatrace metrics lines (summaries and timers) with a count of 0 are dropped on the exporter side.
see original: https://github.com/micrometer-metrics/micrometer/pull/2970

Fixes https://github.com/micrometer-metrics/micrometer/issues/3007


jonatan-ivanov:
Thank you @pirgeo for the fix!

PR #3047, opened 2022-03-01
Add HighCardinalityTagsDetector

jonatan-ivanov:
Draft solution for detecting high cardinality tags:
- [x] The scheduler frequency and delay should be configurable
- [x] We might want to do some measurements to set a default threshold (see https://github.com/micrometer-metrics/micrometer/pull/3047#issuecomment-1133767093)
- [x] We might want this to be configurable from the MeterRegistry


wengertj:
Just a thought/question I had while looking at your draft:

There is  the `MeterRegistry.Config#onMeterAdded` callback. Could that also be an entrypoint for detecting high-cardinality tags? Some advantages might be:
* resource-friendly
* possible more user friendly too:
    * Immediate feedback when a meter is added
    * easier to configure - no polling interval
    * no need to shutdown anything

Thoughts @jonatan-ivanov ?

jonatan-ivanov:
@wengertj Yes, I think it can be used for this purposes.
I was thinking about using it but the trade-off between polling and using the listener is performance vs. convenience. I agree with you, adding this as a listener would be simpler, more convenient and it would look way nicer but on the other hand every time a new meter is created, you need to go through all of the meters that you have created so far to do the check and sometimes you can have lots of metrics. The polling solution on the other hand will do the check on an interval (e.g.: every 5m) but it will do the check even if you did not add a single meter (this can be fixed by memorizing the number of meters).

wengertj:
Fair points 👍 
I think it somewhat comes down to the question how opinionated micrometer wants to be about this - i.e will the detector be enabled by default or not. 

Related to that: I think that it would be nice if we had an easy to discover API to enable or disable the detector (in addition to documentation). Maybe something along the lines of:

```java
Metrics.globalRegistry.withDefaultHighCardinalityTagsDetector()
```

If we stick to the polling solution we might consider implementing `AutoCloseable` and shutdown the polling executor in `close`. This gives users a clear indication that there's resources that need to be closed properly.

jonatan-ivanov:
👍🏼 for `AutoCloseable`.

jonatan-ivanov:
I did some measurements with a Spring Boot based web service ([resourceater](https://github.com/jonatan-ivanov/resourceater)) that creates meters with high cardinality tags.

# Environment

Java 18, macOS (x86_64)
```
 ❯ java --version
openjdk 18 2022-03-22
OpenJDK Runtime Environment Temurin-18+36 (build 18+36)
OpenJDK 64-Bit Server VM Temurin-18+36 (build 18+36, mixed mode, sharing)
 ❯ sw_vers
ProductName:	macOS
ProductVersion:	12.4
```

# Code
You can see how meters are created in [`MeterResource`](https://github.com/jonatan-ivanov/resourceater/blob/main/src/main/java/resourceater/model/resource/micrometer/MeterResource.java) and how I start the app in the [`Makefile`](https://github.com/jonatan-ivanov/resourceater/blob/main/Makefile) (`-XX:MinHeapSize=1G -XX:MaxHeapSize=1G`).

# Measurements

The scenario was alway the same for all of the measurements:
- Start the app
- "Warm it up" by creating a single meter
- Triggering GC from VisualVM
- Creating meters
- Triggering GC from VisualVM

So in the heap graphs you see some baseline, then a drop: first GC to get the real baseline, then a slope: creating meters, then another drop (second GC to see the real utilization). The amount of heap used by the meters is basically the difference between the baseline after the drops.

I did not capture a heap dump or checked the exact numbers, I only wanted to know the approximate usage.

## 1 million Counters

As you can see creating 1 million Counters occupied a little less than 300MiB from heap.
1M-counters

## 1 million Timers

As you can see creating 1 million Timers occupied approximately 550MiB from the heap.
1M-timers

## 500k Counters plus 500k Timers

As you can see creating 500k Counters and 500k Timers occupied approximately 400MiB from heap.
500K-conters-and-500K-timers

## 2 million Timers
`java.lang.OutOfMemoryError: Java heap space`

## 1.5 million Timers

As you can see creating 1.5 million Timers occupied approximately 800MiB from the heap.

1 5M-timers

I think based on these measurements, setting the default threshold to 1M for the `HighCardinalityTagsDetector` can be an ok start since it's high enough to be an indicator that something is wrong but low enough that some apps with some breathing room have a chance to survive.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -0,0 +1,156 @@
...
+        this.delay = delay;
+        this.meterNameConsumer = meterNameConsumer;
+        this.scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+    }
+
+    /**
+     * Starts a scheduled job that checks if you have high cardinality tags.
+     */
+    public void start() {
+        this.scheduledExecutorService.scheduleWithFixedDelay(this::detectHighCardinalityTags, 0, this.delay.toMillis(),
```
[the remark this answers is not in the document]
jonatan-ivanov: @sonatype-lift ignore

jonatan-ivanov:
I'm thinking if this should be opt-in or opt-out. Right now it is opt-in.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param meterNameConsumer The action to execute if the first high cardinality tag is
+     * found
+     */
+    public HighCardinalityTagsDetector(MeterRegistry registry, int threshold, Duration delay,
+            Consumer<String> meterNameConsumer) {
+        this.registry = registry;
+        this.threshold = threshold;
+        this.delay = delay;
+        this.meterNameConsumer = meterNameConsumer;
+        this.scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
```
wengertj, line 94: Would it be worth passing in a `io.micrometer.core.instrument.util.NamedThreadFactory` and give the poller thread a descriptive name?

shakuzen:
> * We might want to do some measurements to set a default threshold

Since long ago, I've wanted to have some warning if memory used by meters exceeds a configurable % of heap memory. We could do something like that here, but maybe it doesn't make as much sense here and we should consider that separately. It just seems like we would want to warn about high cardinality before an OOM happens. The problem with an absolute threshold is that for low memory applications it will be too high. For large, monolithic applications with few instances, maybe it will be too low. It is configurable, so it probably isn't a problem. I'm just thinking how we can have a better default.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -0,0 +1,156 @@
...
+                    return Optional.of(name);
+                }
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private static void logWarning(String name) {
+        LOGGER.warn(String.format("It seems %s has high cardinality tags.", name));
```
shakuzen, line 153: It would be nice to mention how many meters there were or at least what the threshold is. Also, many users will be left wondering what to do when encountering this log message. It's hard to give generically applicable advice, but maybe something along the lines of:
> Check your configuration for the instrumentation of `{name}` to find and fix the cause of the high cardinality. If the cardinality is expected and acceptable, raise the threshold for this HighCardinalityTagsDetector.

Maybe even linking to our documentation would be good. Though right now, this is probably the only relevant section: https://micrometer.io/docs/concepts#_tag_values

shakuzen, line 153: Do we want to continually log this at warn or would the warn-then-debug logger be better? I'm leaning toward voting the latter, but I could be convinced otherwise.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.function.Consumer;
+
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
+
+/**
+ * Tries to detect high cardinality tags by checking if the amount of Meters with the same
+ * name is above a threshold. You can use this class in two ways: 1. Call findFirst and
+ * check if you get any results, if so you probably have high cardinality tags 2. Call
+ * start which will start a scheduled job that will do this check for you
```
shakuzen, line 34: I think it is worth mentioning `MeterFilter.maximumAllowableMetrics` and `MeterFilter.maximumAllowableTags` as related tools to deal with having too many meters.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
jonatan-ivanov: Definitely! Thanks!

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
jonatan-ivanov: For performance reasons at the moment we only count them till one breaches the threshold because we don't know how many of them could be.
I added the threshold and your suggestion about the actions the users can take with the link to our docs.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
jonatan-ivanov: 👍🏼

jonatan-ivanov:
>Since long ago, I've wanted to have some warning if memory used by meters exceeds a configurable % of heap memory. We could do something like that here, but maybe it doesn't make as much sense here and we should consider that separately. It just seems like we would want to warn about high cardinality before an OOM happens.

I think we can prepare this for that use-case too, e.g.: renaming it to something more general: `AnomalyDetector` and making it possible to register event handlers so that you can get a notification once something bad happened. Though we need to provide a way to inject configuration for the code that does the checks (or make the code itself injectable).

>The problem with an absolute threshold is that for low memory applications it will be too high. For large, monolithic applications with few instances, maybe it will be too low. It is configurable, so it probably isn't a problem. I'm just thinking how we can have a better default.
I'm going to check if I can figure out some usable heuristics and make the default a function of the max heap.

jonatan-ivanov:
I added a simple heuristics for the threshold based on the max heap.
I also set a maximum value for this which means no matter how much heap you have, if two million meters will share the same name, this thing will scream by default. If users decide that 2M+ is ok and they have enough heap, they can configure a value they like.

marcingrzejszczak (approved):
LGTM

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -0,0 +1,179 @@
...
+    private void logWarning(String name) {
+        WARN_THEN_DEBUG_LOGGER.log(String.format("It seems %s has high cardinality tags (threshold: %d meters).\n"
+                + "Check your configuration for the instrumentation of %s to find and fix the cause of the high cardinality (see: https://micrometer.io/docs/concepts#_tag_values).\n"
+                + "If the cardinality is expected and acceptable, raise the threshold for this %s.", name,
+                this.threshold, name, this.getClass().getSimpleName()));
+    }
+
+    private static long calculateThreshold() {
+        // half of the heap in MiB
+        long allowance = Runtime.getRuntime().maxMemory() / 1024 / 1024 / 2;
```
shakuzen, line 173: I wonder if half the max heap isn't too much for one meter name to be allowed to take before warning, considering there will be other meters too. This is a bit of a bike shed issue on what the right default is, though. I would think keeping total metrics memory overhead to 10% or less of the heap would be ideal. However, this doesn't count against all metrics registered, just metrics with the same name in one registry. If there is more than one registry, 50% * 2 is already and OOM situation.

shakuzen (commented):
Overall looks good. Just one comment about the default threshold calculation based on the heap.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
jonatan-ivanov: I think lowering the threshold is a good idea. I think it also worth calling out in the javadoc that this component will not detect if Micrometer is using a lot of memory, it only tries to detect if the app potentially has high cardinality tags. E.g.: it won't detect if one has no tags but a timestamp is concatenated to the name.

We can implement a `HighMemoryUsageDetector` or make this more generic (see: https://github.com/micrometer-metrics/micrometer/pull/3047#issuecomment-1137897921), I think the first option is more flexible.

As a reference, this is the difference in thresholds between 50% and 10%:
**50%**
```
  4MiB: 4_000
256MiB: 256_000
512MiB: 512_000
  1GiB: 1_024_000
 2+GiB: 2_000_000
```

**10%**
```
  4MiB: 0 // I will define a minimum for this, let's say 1_000
256MiB: 50_000
512MiB: 102_000
  1GiB: 204_000
  2GiB: 408_000
  4GiB: 818_000
  8GiB: 1_638_000
16+GiB: 2_000_000
```

shakuzen (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
shakuzen: Sounds good to me.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -846,6 +850,42 @@ public PauseDetector pauseDetector() {
             return pauseDetector;
         }
 
+        /**
+         * Creates and starts a new {@link HighCardinalityTagsDetector} for this registry.
+         * @return This configuration instance.
+         */
+        public Config withHighCardinalityTagsDetector() {
+            highCardinalityTagsDetector = new HighCardinalityTagsDetector(MeterRegistry.this);
```
shakuzen, line 858: Should we check if a detector is already configured and close that before replacing it? I'm not sure which would be more surprising behavior.

jonatan-ivanov: 👍🏼 fixed

Issue #2818, opened 2021-10-13
Avoid global publishing alignment


shakuzen:
> Could you confirm my understanding that all replicas will now push their points at the exact same time?

Yes, that appears to be the case now.
The distributed coordination on publication time is a side effect we probably overlooked in merging #1218. Thank you for the investigation on that. While we want consistency within a process (or registry), we don't want coordination across processes on publishing, as you suggested.
That said, the publication happening at a different millisecond across instances should not be relied upon either.

_Originally posted by @shakuzen in https://github.com/micrometer-metrics/micrometer/issues/2787#issuecomment-924054475_

shakuzen:
To be clear, the issue with global alignment on publication time is that it causes a concentration of work handling metrics from all application instances at the same time rather than naturally disbursing that work out over the step interval based on when application instances start up. This results in spikes of network traffic moving metrics to the backend, and spikes in CPU and memory usage processing all the incoming metrics on the metrics backend.

shakuzen:
I will use diagrams to illustrate the relation between step boundaries, registry start, and publications. The legend is as follows:

* `-` marks the passage of time
* `|` marks step interval boundaries
* `p` marks publications
* `s` marks when `start` is called on the `MeterRegistry`
* `S1`/`S2` indicate ordinal step intervals since the start

When a publication happens, metrics based on the values recorded during the previous step interval are published. Step boundaries are relative to epoch time.

Before the fix for #1218 the publication times would naturally be distributed throughout a step interval by application instances starting up at different times.

```
     S1         S2         S3         S4         S5
|------s---|------p---|------p---|------p---|------p---|
|---------s|---------p|---------p|---------p|---------p|
|---s------|---p------|---p------|---p------|---p------|
```

After the fix for #1218, we have the following situation.

```
     S1         S2         S3         S4         S5
|------s---|p---------|p---------|p---------|p---------|
|---------s|p---------|p---------|p---------|p---------|
|---s------|p---------|p---------|p---------|p---------|
```

Regardless of when applications startup, publication times are scheduled for the same time across all instances.

shakuzen:
One potential solution would be to record the time a `PushMeterRegistry` is constructed and pass that to StepMeters so they can align their step to the creation time of the registry, and likewise align the publish time to that. That should avoid the original issue described in #1218 and this issue. If someone would like to try implementing that, or has feedback on it, let us know.

igr:
[...]

Issue #1218, opened 2019-02-07
PushMeterRegistry doesn't align publish() with step boundaries


john-karp:
PushMeterRegistry uses `scheduledExecutorService.scheduleAtFixedRate(this::publish, config.step().toMillis(), config.step().toMillis(), TimeUnit.MILLISECONDS)` to schedule `publish()` with an initial delay of `config.step().toMillis()`. So while the period between invocations is fixed, they may run at any time with respect to the epoch.

StepDouble sets step boundaries relative to the epoch with `final long stepTime = now / stepMillis`.

As a result, PushMeterRegistry can schedule calls to publish() with arbitrary offsets relative to step boundaries. If we are unlucky, publish() will be scheduled right before a step boundary, so ordinary scheduling variations can lead to publish() sometimes being called before the boundary, and sometimes right after, leading to StepDouble sporadically duplicating or dropping data. Code example here:
https://gist.github.com/john-karp/93b86885b39f9d813345100be2a6496a

The simplest fix for PushMeterRegistry would set the initial delay of the executor service to align it with the steps. (Though I think ScheduledExecutorService cannot be trusted to keep that alignment in the long term when you have a server with ntpd active.)

izeye:
This seems to be resolved via #2393.

shakuzen:
For a visual, let me provide a diagram illustrating the problem. `-` marks the passage of time, `|` marks step interval boundaries, `p` marks publications, `s` marks when `start` is called on the `MeterRegistry`, `S1`/`S2` indicate ordinal step intervals since the start. When a publication happens, the values from the previous step interval are published. Step boundaries are relative to epoch time.

```
S1      S2         S3         S4
--s|---------p|----------|p-------p|---
```

In the above case, the `MeterRegistry` is started at the end of the step interval S1. The first publication will be scheduled one step interval later, at the end of S2. That publication will include any `StepValue` based metrics with values recorded during S1.

The problem described by this issue is illustrated by the next publish that may happen slightly later than exactly one step interval, pushing it across a step boundary. Therefore the second publication (the second `p`) happens in S4, publishing values recorded during S3. This means values recorded in S2 are never published. This is the dropping of data referred to in the original description.

Subsequently, if the next publication happened slightly sooner than scheduled, it could happen in the same step as the previous publication, resulting in publishing the same values as the previous publish. I'm not sure if this happens in practice. I can understand scheduling being late but I'm not sure how it would be earlier than expected.

[...]

PR #3054, opened 2022-03-04
Add PushMeterRegistry configuration for aligned steps between multipl…

### LINKED ISSUE #3053: Make a flag for the initial delay for PushMeterRegistry
As seen here: https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java#L76

we calculate the initial delay in order to trigger the tasks on the exact mark. For example, if `delay` is `10`s, this means that that scheduled push happens on `10s` mark, eg: 1:22:00, 1:22:10, 1:22:20...

We need this feature to be optional.

**Rationale**
If you have metrics on many places in a microservice application, imagine what happens when pushes from all services are triggered _at the same time_. It makes sense to randomize the initial delay in order to spread the load over time.

**Additional context**
N/A

igr:
…e instances of the library (closes #3053)

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -71,9 +71,12 @@ public abstract class PushMeterRegistry extends MeterRegistry {
             logger.info("publishing metrics for " + this.getClass().getSimpleName() + " every " + TimeUtils.format(config.step()));
 
             scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
-            // time publication to happen just after StepValue finishes the step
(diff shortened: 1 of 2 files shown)


igr:
Here is an example of spikes of accumulated traffic in our system, all because of wall-clock alignment:

spikes-images

This will be flatten after the patch (will update soon)

shakuzen:
Thank you for the pull request. If we were to add this configuration option, setting it to `false` would open users back up to the issue described in #1218, which I understand may be preferable to the issue you've described. We'd like to solve that issue, though, rather than introduce a configuration option. See #2818, which describes the issue. I'll add a comment there with what may be a workable solution I thought of when reviewing this pull request. If you'd be interested in implementing that, please do send us another pull request. Let us know if you have any questions.

PR #3059, opened 2022-03-09
Prevent CompositeMeter allocating on each record

BrentDouglas:
The way forEachChild was written mean each lambda has to close over the
argument which results in an allocation for each call. Instead we
slightly rewrite the forEachChild api to allow passing through the data
to each call so that the lambdas do not need to capture the argument.


BrentDouglas:
@shakuzen I updated this to avoid all the weird signatures and just return the iterable, seems a lot cleaner

shakuzen (approved):
Thanks. I've rebased onto `1.7.x` so we can include this in patch releases.

PR #2916, opened 2021-12-13
Allow file-based configuration for Dynatrace v2 exporter

pirgeo:
This PR introduces support for file-based configuration using the `DynatraceFileBasedConfigurationProvider`. It also makes sure changes in the file-based configuration are picked up, without having to restart the app to change configuration. 

This change adds another layer to the auto-configuration that is already available: Instead of falling back to the local OneAgent directly, the the file at `/var/lib/dynatrace/enrichment/endpoint/endpoint.properties` will be checked for configuration. The file will typically be provided by the [Dynatrace Operator](https://github.com/Dynatrace/dynatrace-operator) but can also just be added manually if you'd like to test it locally. I will follow up with a PR updating the documentation in `micrometer-docs` to describe the changes.

PR #3065, opened 2022-03-10
Simplify tests after Dynatrace Metric Utils 1.5.0 release

### LINKED ISSUE #3035: Dynatrace file-based configuration slow on macOS
Thanks to @jonatan-ivanov for pointing out a problem with the WatchService based approach on macOS. A native implementation seems to be missing for the macOS file systems (at least in some JDKs), leading to a fallback to a polling implementation that checks the filesystem for changes every 10s. This is consistent with what @jonatan-ivanov reported in #2916. We are currently looking into this.

Original comment for context: 

> @pirgeo I think there are some issues with the Dynatrace file poller, it seems it takes a lot of time on macOS for the file poller to detect the changes.

> The tests in this PR were failing for me on my mac (OS X 12.1, x86_64): see [this report](https://ge.micrometer.io/s/g3s2ubcarfaok).
I replaced `Thread::sleep` with awaitility it takes almost 10 seconds on my machine to see the file changes (instead of 10ms that you had). Could you please look into it if anyone else has this issue before?

> My changes: https://github.com/jonatan-ivanov/micrometer/commit/f4c1c4227bc50d52895ba57d21592c7315b2cf18 (they are unrelated since the build was failing for me at the first place).

_Originally posted by @jonatan-ivanov in https://github.com/micrometer-metrics/micrometer/issues/2916#issuecomment-1040885909_

pirgeo:
Resolves #3035

This PR simplifies the tests after we made a method for testing public in the `DynatraceFileBasedConfigurationProvider`, thanks @jonatan-ivanov for pointing it out: https://github.com/micrometer-metrics/micrometer/pull/2916#discussion_r807342246.

The newly released version of the Dynatrace utility library ([1.5.0](https://search.maven.org/artifact/com.dynatrace.metric.util/dynatrace-metric-utils-java/1.5.0/jar)) should also fix the behavior where tests take a long time on macOS, due to a FilePolling implementation that only polled for changes every 10s.


shakuzen:
Could you rebase onto the `1.9.x` branch. I'm having trouble doing so on your branch. Perhaps permission for maintainers to update the PR is not enabled? Or I'm just doing something wrong.

pirgeo:
Hey @shakuzen, thanks for taking a look. I think the problem might be that this is a fork in our organization, and I think for organizational forks `maintainers_can_modify` is false by default. I rebased onto 1.9.x, and will see if I can't allow you to modify it after all.

jonatan-ivanov:
Thanks @pirgeo!
fyi: I updated the tests and lowered the polling timeouts: https://github.com/micrometer-metrics/micrometer/commit/707508cf3a8af7565db3d57e86c631a03c1c171f

PR #3088, opened 2022-03-21
Correct tag names according to Micrometer's docs

seckin206:
See. https://micrometer.io/docs/concepts#_tag_naming

With the current implementation and using official prometheus exporter,
we end up with label names `methodType` and `statusCode`. These labels
don't follow the snake_case naming convention that prometheus follows
for labels.
Using dot for word separator makes micrometer automatically convert
label names to snake case out of the box. 
See. https://github.com/micrometer-metrics/micrometer/blob/main/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java

This PR applies the same changes raised here: https://github.com/yidongnan/grpc-spring-boot-starter/pull/645

### DIFF
--- a/micrometer-binders/src/main/java/io/micrometer/binder/grpc/AbstractMetricCollectingInterceptor.java
+++ b/micrometer-binders/src/main/java/io/micrometer/binder/grpc/AbstractMetricCollectingInterceptor.java
@@ -52,11 +52,11 @@ public abstract class AbstractMetricCollectingInterceptor {
     /**
      * The metrics tag key that belongs to the type of the called method.
      */
-    private static final String TAG_METHOD_TYPE = "methodType";
+    private static final String TAG_METHOD_TYPE = "method.type";
     /**
      * The metrics tag key that belongs to the result status code.
      */
-    private static final String TAG_STATUS_CODE = "statusCode";
+    private static final String TAG_STATUS_CODE = "status.code";
 
     /**
      * Creates a new counter builder for the given method. By default the base unit will be messages.


shakuzen:
Changing the tag name would be a breaking change for anyone relying on the original tag names. We strive to avoid such breaking changes. We are going to do a complete reassessment of meter names and tags as part of our 2.0 major version we're working on.

> These labels don't follow the snake_case naming convention that prometheus follows for labels.

Prometheus metric names certainly follow snake_case, but I'm not sure I've seen documentation about the same for labels. It's certainly not a requirement (though it isn't for metric names either). Is this more just about making the label names consistent with other label names, or is there some other issue you're facing due to the current naming?

seckin206:
> Changing the tag name would be a breaking change for anyone relying on the original tag names. 

I agree. This is a breaking change for any exporter who failed to recognize camelCase tag keys until now and failed to convert them to the desired case. (Prometheus exporter is one of such exporters where there is no functionality to convert camelCase tag keys to snake case label names.)

> Prometheus metric names certainly follow snake_case, but I'm not sure I've seen documentation about the same for labels. It's certainly not a requirement (though it isn't for metric names either). 

This is one of the recent learnings for me, too. The public documentation doesn't suggest anything regarding label names: https://prometheus.io/docs/practices/naming/
However, `promtool` have checks for labels following snake case convention. The checks for labels seem to be introduced in 2019 with this PR: https://github.com/prometheus/prometheus/pull/5515. The linting functionality currently is implemented here and used by promtool: https://github.com/prometheus/client_golang/blob/3bc8f2c651ff1abe956f970ae3d0a2d45737c501/prometheus/testutil/promlint/promlint.go#L256

> Is this more just about making the label names consistent with other label names, or is there some other issue you're facing due to the current naming?

Currently, the problem is a potential one. However, I assume it will be a real problem for us soon when we start using `promtool` more extensively for linting. I think I was naturally expecting that prometheus exporter of micrometer would export metrics such that they pass `promtool` checks.

shakuzen:
> We are going to do a complete reassessment of meter names and tags as part of our 2.0 major version we're working on.

Following up on this: we should still do this reassessment of names when we do a next major version. However, we changed plans from releasing a major version last year and currently don't know when a major version will be released.

We do now offer grpc instrumentation using the Observation API introduced in Micrometer 1.10. This uses [GrpcObservationDocumentation](https://github.com/micrometer-metrics/micrometer/blob/036b40e429e4115865f9bf1f77731f9c790c4d2a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationDocumentation.java) and `DefaultGrpcServerObservationConvention`/`DefaultGrpcClientObservationConvention`. If users switch to this instrumentation, I think the original issue will go away and they have more ways to customize now. That doesn't solve things for users who don't switch to the `Observation`-based instrumentation, but as mentioned, for backwards-compatibility reasons we don't want to change that in a breaking way. Users can still customize the behavior with a `MeterFilter` if they want a different tag naming for these. I think that might be the best we can do unless someone else has a suggestion we should consider before a major release.

Given the above, I think we can close this pull request for now (we can't merge it until working on 2.0) and suggest using the newer Observation-based instrumentation. Any objections or other ideas?

seckin206:
Thanks for the pointers! Since I opened this PR, I stopped using micrometer. Feel free to close the PR when it suits you 🙂

PR #3092, opened 2022-03-24
Moved out utils, docs and loggers to commons module

marcingrzejszczak:
we need to do this to fully abstract `micrometer-core` from `micrometer-observation`. Without this change `DocumentedMeter` depends on the `micrometer-observation` module. If you don't have `micrometer-observation` on the classpath your app will break.


micrometer-commons/build.gradle
```diff
@@ -0,0 +1,16 @@
+description 'Module containing common code'
+
+dependencies {
+    // log monitoring
+    optionalApi 'ch.qos.logback:logback-classic'
+    optionalApi 'org.apache.logging.log4j:log4j-core'
```
jonatan-ivanov, line 6: Do we really need these here?
I think we only need SLF4J (and the JDK Logger).

micrometer-commons/src/main/java/io/micrometer/common/Tag.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micrometer.observation;
+package io.micrometer.common;
```
jonatan-ivanov, line 16: `Tag` and `Tags` are tricky because those can be used by third-party instrumentations.

jonatan-ivanov (changes requested):

PR #3093, opened 2022-03-24
Add resettable DistributionSummary and Timer for Dynatrace registry

pirgeo:
related to #3007 

This PR introduces resettable `Timer` and `DistributionSummary` instruments to be used in the Dynatrace v2 exporter to ensure no max-buffering from previous export intervals ends up in the exported lines. These types also add support for tracking the minimum since the last export.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
@@ -58,8 +58,9 @@ import static io.micrometer.core.instrument.config.MeterFilterReply.NEUTRAL;
 public class DynatraceMeterRegistry extends StepMeterRegistry {
     private static final ThreadFactory DEFAULT_THREAD_FACTORY = new NamedThreadFactory("dynatrace-metrics-publisher");
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(DynatraceMeterRegistry.class);
-    private final DynatraceApiVersion apiVersion;
 
+    private final DynatraceApiVersion apiVersion;
+    private final boolean useDynatraceSummaryInstruments;
     private final AbstractDynatraceExporter exporter;
 
     @SuppressWarnings("deprecation")
@@ -71,6 +72,7 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
         super(config, clock);
 
         apiVersion = config.apiVersion();
+        useDynatraceSummaryInstruments = config.useDynatraceSummaryInstruments();
 
         if (apiVersion == DynatraceApiVersion.V2) {
             logger.info("Exporting to Dynatrace metrics API v2");
@@ -101,7 +103,7 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
 
     @Override
     protected DistributionSummary newDistributionSummary(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, double scale) {
-        if (apiVersion == DynatraceApiVersion.V2) {
+        if (apiVersion == DynatraceApiVersion.V2 && useDynatraceSummaryInstruments) {
             return new DynatraceDistributionSummary(id, clock, distributionStatisticConfig, scale);
         }
         return super.newDistributionSummary(id, distributionStatisticConfig, scale);
@@ -109,7 +111,7 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
 
     @Override
     protected Timer newTimer(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, PauseDetector pauseDetector) {
-        if (apiVersion == DynatraceApiVersion.V2) {
+        if (apiVersion == DynatraceApiVersion.V2 && useDynatraceSummaryInstruments) {
             return new DynatraceTimer(id, clock, distributionStatisticConfig, pauseDetector, exporter.getBaseTimeUnit());
         }
         return super.newTimer(id, distributionStatisticConfig, pauseDetector);
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
@@ -109,7 +109,7 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
     @Override
     protected Timer newTimer(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, PauseDetector pauseDetector) {
         if (apiVersion == DynatraceApiVersion.V2) {
-            return new DynatraceTimer(id, clock, distributionStatisticConfig, exporter.getBaseTimeUnit());
+            return new DynatraceTimer(id, clock, distributionStatisticConfig, pauseDetector, exporter.getBaseTimeUnit());
         }
         return super.newTimer(id, distributionStatisticConfig, pauseDetector);
     }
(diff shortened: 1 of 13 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
```diff
@@ -89,6 +98,22 @@ protected TimeUnit getBaseTimeUnit() {
         return this.exporter.getBaseTimeUnit();
     }
 
+    @Override
+    protected DistributionSummary newDistributionSummary(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, double scale) {
+        if (apiVersion == DynatraceApiVersion.V2) {
+            return new DynatraceDistributionSummary(id);
```
jonatan-ivanov, line 104: Should not this be aware both the distributionStatisticConfig and the scale?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
```diff
@@ -89,6 +98,22 @@ protected TimeUnit getBaseTimeUnit() {
...
+        }
+        return super.newDistributionSummary(id, distributionStatisticConfig, scale);
+    }
+
+    @Override
+    protected Timer newTimer(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig, PauseDetector pauseDetector) {
+        if (apiVersion == DynatraceApiVersion.V2) {
+            return new DynatraceTimer(id, clock, exporter.getBaseTimeUnit());
```
jonatan-ivanov, line 112: Shouldn't this be aware both the distributionStatisticConfig and the pauseDetector?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/types/DynatraceDistributionSummary.java
```diff
@@ -0,0 +1,99 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Resettable {@link DistributionSummary} implementation for Dynatrace exporters.
+ *
+ * @author Georg Pirklbauer
+ * @since 1.9.0
```
jonatan-ivanov, line 30: If you want this in 1.9, could you please rebase the PR to the `1.9.x` branch?
Otherwise could you please change the `@since` tags to `2.0.0`?

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/types/DynatraceSummary.java
```diff
@@ -0,0 +1,67 @@
...
+        if (amount < 0) {
+            return;
+        }
+        if (count == 0) {
+            min = amount;
+            max = amount;
+        } else {
+            min = Math.min(min, amount);
+            max = Math.max(max, amount);
+        }
```
jonatan-ivanov, line 40: Is this needed (if-else)?
If count is zero both min and max are zero so just calculating them without any condition should give you the right results:
```
min = Math.min(min, amount);
max = Math.max(max, amount);
```

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/types/DynatraceSummary.java
```diff
@@ -0,0 +1,67 @@
...
+package io.micrometer.dynatrace.types;
+
+/**
+ * Internal class for resettable summary statistics
+ *
+ * @author Georg Pirklbauer
+ * @since 1.9.0
+ */
+final class DynatraceSummary {
+    private long count = 0;
```
jonatan-ivanov, line 25: There are better tools than `syncronized` to make this class thread safe, could you please check implementations of `DistributionSummary`/`AbstractDistributionSummary`?
e.g.: Check `StepDistributionSummary`, I guess you need something similar but with a reseting max/min.

You can use `LongAdder` for `count`, `DoubleAdder` for `total` and `AtomicLong` for `min` and `max`.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/types/DynatraceTimer.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public DynatraceSummarySnapshot takeSummarySnapshotAndReset(TimeUnit unit) {
+        DynatraceSummarySnapshot snapshot = takeSummarySnapshot(unit);
+        summary.reset();
+        return snapshot;
+    }
+
+    // from AbstractTimer
+    @Override
+    public <T> T recordCallable(Callable<T> f) throws Exception {
```
jonatan-ivanov, line 73: This is exactly the same code as the `recordCallable` in `AbstractTimer`, you don't need to override it (I guess the same is true for the other record methods).

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -99,9 +102,9 @@ private boolean shouldIgnoreToken(DynatraceConfig config) {
 
     private DimensionList parseDefaultDimensions(Map<String, String> defaultDimensions) {
         List<Dimension> dimensions = Stream.concat(
-                defaultDimensions.entrySet().stream(),
-                staticDimensions.entrySet().stream()
-        )
+                        defaultDimensions.entrySet().stream(),
```
jonatan-ivanov, line 105: The previous indentation seemed right.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -169,7 +172,19 @@ private String createCounterLine(Meter meter, Measurement measurement) {
     }
 
     Stream<String> toTimerLine(Timer meter) {
-        return toSummaryLine(meter, meter.takeSnapshot(), getBaseTimeUnit());
+        if (!(meter instanceof DynatraceSummarySnapshotSupport)) {
```
jonatan-ivanov, line 175: How can this happen? If the registry will give you a `DynatraceTimer` in case of V2 and this class is V2 only, this can't get anything else than a `DynatraceTimer`, or can it?

jonatan-ivanov (changes requested):
I quickly went through the PR, I think introducing a new Timer and DistributionSummary is the way to solve the original issue. 👍🏼

pirgeo (commented):
Thanks for the input, @jonatan-ivanov!

I updated the PR with a bunch of changes based on your comments. Originally, I did not base the DynatraceTimer and DynatraceDistributionSummary off of AbstractTimer and AbstractDistributionSummary since those already track a Histogram, which we currently do not need. Since the underlying data structure is rather simple (only min/max/sum/count) we do not necessarily want to track the whole histogram. I must have missed the `NoopHistogram` in my first pass, thats why I implemented the interfaces directly. By basing our instruments off the Abstract instruments, the implementation actually became simpler. I set the Dynatrace instruments up in a way where the Histogram will always be a NoopHistogram. 

I also added a toggle to the DynatraceConfig now that allows for switching back to the original Micrometer-provided instruments, and rebased it on 1.9.0.

jonatan-ivanov (approved):

jonatan-ivanov:
@pirgeo Thank you very much!

PR #3096, opened 2022-03-25
Deprecated tags?

marcingrzejszczak:


micrometer-commons/src/main/java/io/micrometer/common/Tag.java
```diff
@@ -20,7 +20,7 @@
  *
  * @author Jon Schneider
  */
-public interface Tag extends Comparable<Tag> {
+public interface Tag<T extends Tag<T>> extends Comparable<T> {
```
jonatan-ivanov, line 23: Will not this cause issues/compiler warnings?

jonatan-ivanov: Also, see my other comments on https://github.com/micrometer-metrics/micrometer/pull/3097

marcingrzejszczak: Yeah it will, I think it's not needed at all

PR #3097, opened 2022-03-25
Deprecating Tag and Tags in core and moving them to commons

marcingrzejszczak:
BTW the binary compatibility plugin shows that there's a breakage between 1.9 and 1.8

```
***! MODIFIED INTERFACE: PUBLIC ABSTRACT io.micrometer.signalfx.SignalFxConfig  (not serializable)
	+++! NEW METHOD: PUBLIC(+) boolean publishCumulativeHistogram()
```


jonatan-ivanov:
The `publishCumulativeHistogram` is new in 1.9 but it is an additional default method: https://github.com/micrometer-metrics/micrometer/blob/3eb507247e1fc907ed85f0aa4daaead3e68c9375/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxConfig.java#L48

I think we are ok if this breaks serializability.

build.gradle
```diff
@@ -290,6 +293,47 @@ subprojects {
...
+                            return false
+                        } else {
+                            println "Will download and perform baseline comparison with ${compatibleVersion}"
+                            return true
+                        }
+                    }
+
+                    onlyIfNewer true
+                    compress true
+                    src "${repositories.mavenCentral().url}io/micrometer/${project.name}/${compatibleVersion}/${project.name}-${compatibleVersion}.jar"
```
jonatan-ivanov, line 318: Does `japicmp` has a `configuration`?
If so we can add a dependency like `japicmp 'io.micrometer:micrometer-core:1.8.4'` so gradle will download the jar for us and we don't need the download plugin.

micrometer-commons/build.gradle
```diff
@@ -0,0 +1,15 @@
+description 'Module containing common code'
+
+dependencies {
+    // log monitoring
+    optionalApi 'ch.qos.logback:logback-classic'
```
jonatan-ivanov, line 5: Is this needed?

micrometer-commons/src/main/java/io/micrometer/common/Tag.java
```diff
@@ -0,0 +1,37 @@
...
+    String getKey();
+
+    String getValue();
+
+    static Tag of(String key, String value) {
+        return new ImmutableTag(key, value);
+    }
+
+    @Override
+    default int compareTo(T o) {
```
jonatan-ivanov, line 33: Is this why we need the generic param?
If this is the only reason, instead of doing this, can we use a `Comparator` where we would use this?
I think the only place is
```
Arrays.sort(this.tags);
```
that can be replaced to:
```
Arrays.sort(this.tags, Comparator.comparing(Tag::getKey));
```
And later to an universal `Comparator` that can compare both tags.

build.gradle
marcingrzejszczak: I've copied it over from reactor-netty

micrometer-commons/src/main/java/io/micrometer/common/Tag.java
marcingrzejszczak: Actually, it seems that this can be at all removed - I mean the generic type.

marcingrzejszczak:
> The `publishCumulativeHistogram` is new in 1.9 but it is an additional default method:
> 
> https://github.com/micrometer-metrics/micrometer/blob/3eb507247e1fc907ed85f0aa4daaead3e68c9375/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxConfig.java#L48
> 
> I think it's ok if this breaks serializability.

Yeah, the thing is I have no idea how to set up the plugin to ignore such things. I guess we should not fail if something got modified if we're checking against `1.8` and then once `1.9` is out we should turn that flag back to `true`

PR #3098, opened 2022-03-26
Polish observation changes

izeye:
This PR polishes observation changes a bit.

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/core/tck/ObservationRegistryCompatibilityKit.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/core/tck/ObservationRegistryCompatibilityKit.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 VMware, Inc.
+ * Copyright 2022 VMware, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


micrometer-observation-test/src/main/java/io/micrometer/core/tck/ObservationRegistryCompatibilityKit.java
```diff
@@ -187,12 +191,13 @@ void checkedRunnableThrowingErrorShouldBeObserved() {
 
     @Test
     void supplierShouldBeObserved() {
+        @SuppressWarnings("unchecked")
         ObservationHandler<Observation.Context> handler = mock(ObservationHandler.class);
         when(handler.supportsContext(isA(Observation.Context.class))).thenReturn(true);
         registry.observationConfig().observationHandler(handler);
         Observation observation = Observation.createNotStarted("myObservation", registry);
 
-        String result = observation.observe((Supplier<String>) () -> {
```
jonatan-ivanov, line 195: I did these for two reasons:
1. It is explicit what is tested
2. It is safer since in some cases the cast is needed, this ensures that the right thing is tested even if we change things on the observation or on the tests side

I'm ok removing these though, what do you think?

izeye: @jonatan-ivanov Thanks for the feedback!

I see your point. I just changed it as IDE suggested. How about extracting it to a local variable to make it explicit as you mentioned while IDE won't suggest?

jonatan-ivanov: That's a great idea! :)

izeye: @jonatan-ivanov Thanks! I updated as discussed.

shakuzen (approved):

micrometer-observation-test/src/main/java/io/micrometer/core/tck/ObservationRegistryCompatibilityKit.java
jonatan-ivanov: 👍🏼 Thank you!

PR #3107, opened 2022-03-30
fixing name collision on PoolingHttpClientConnectionManagerMetricsBinder

gordon00:
Hello, two metrics have same name on PoolingHttpClientConnectionManagerMetricsBinder


shakuzen:
They have the same meter name but different tags. The ID is based on the meter name and tags, so they are still unique. If you are experiencing some issue here, please describe the problem you are facing in a new issue, ideally with a unit test or sample so that we can reproduce the issue.

PR #3111, opened 2022-04-01
Deprecating tags in core

marcingrzejszczak:
A pass on deprecating `Tag` and `Tags` from core so that the ones from `commons` should be used


micrometer-binders/src/main/java/io/micrometer/binder/jetty/JettyClientMetrics.java
```diff
@@ -67,7 +66,7 @@ public void onQueued(Request request) {
             long requestLength = Optional.ofNullable(result.getRequest().getContent())
                     .map(ContentProvider::getLength)
                     .orElse(0L);
-            Iterable<Tag> httpRequestTags = tagsProvider.httpRequestTags(result);
+            Iterable<? extends io.micrometer.common.Tag> httpRequestTags = tagsProvider.httpRequestTags(result);
```
jonatan-ivanov, line 69: I think local variables do not need to have boundaries.

shakuzen:
We're going to take a different approach to solve the problem, as this approach is too invasive, burdensome on users, and potential for subtle incompatibilities are too much.

PR #3126, opened 2022-04-13
Use more generic KeyValue terminology in Observation and Common module

marcingrzejszczak:
We want to avoid confusion with `Tag` in micrometer-core, which is specific to metrics. In commons and the observation module, it is a more generic concept call KeyValue, which could be used for a metrics tag but could also be used for arbitrarily many other things.


micrometer-observation-test/src/main/java/io/micrometer/core/tck/ObservationRegistryCompatibilityKit.java
```diff
@@ -392,41 +392,41 @@ void supplierShouldNotBeParentScopedIfParentIsNull() {
     void observationFieldsShouldBeSetOnContext() {
         AssertingHandler assertingHandler = new AssertingHandler();
         registry.observationConfig()
-                .tagsProvider(new TestTagsProvider("global"))
-                .tagsProvider(new UnsupportedTagsProvider("global"))
+                .keyValueProvider(new TestKeyValuesProvider("global"))
```
marcingrzejszczak, line 395: We missed `keyValueProvider` -> `keyValuesProvider`

Issue #1993, opened 2020-04-13
Revert Timer#max to use a windowed max rather than last publishing interval.


jkschneider:
In #1856, the behavior of `StepTimer` was changed to publish max only for the last publishing interval. For the purpose of availability monitoring, it is better to allow maximum to decay over a longer period. 

cc / @crankydillo

Further explanation as to why:

> Micrometer decays the maximum rather than aligning it to the publishing interval like it does for sum and count. If we perfectly aligned the view of maximum time to the push-interval then a dropped metrics payload means we potentially miss out on seeing a particularly high maximum value (because in the next interval we'd only consider samples that occurred in that interval). 
> 
> Practically, there are many reasons why a high maximum latency and a dropped metrics payload would be correlated. For example, if the application is under heavy resource pressure (like a saturated network interface), a response to the user for an API endpoint that is being timed (and for which a maximum value is being tracked) may be exceedingly high at the same time that a metrics post request to the monitoring system fails with a read timeout. But such conditions can be (and many times are) temporary. 
> 
> Perhaps you have a client-side load balancing strategy that recognizes that (from the client's perspective) API latency has gone up sharply for this instance that is under resource pressure, and begins preferring other instances. By relieving pressure on this instance it recovers. 
> 
> In some subsequent interval, after the instance has recovered, it's nice to be able to push a maximum latency seen during this time of trouble that would otherwise have been skipped. In fact, it's precisely these times of duress that we care about the most, not the maximum latency under fair-weather conditions!

shakuzen:
We'll also need to revert the changes made to the docs: https://github.com/micrometer-metrics/micrometer-docs/pull/120

crankydillo:
Thanks for the heads up.  I can understand how the importance of max values makes it behavior different from other 'step values'.  It appears you left the refactoring around isolating the rolling logic, which will allow us to continue align max with the last step (and live with the ramifications of missing values) without copying too much of your code.  If this comes up in the future, you might consider providing 2 versions of `StepTimer`.

Please consider making some additions to the javadoc/hosted doc/etc that makes it clear that `max` deviates from other 'step values' and why that is done.

Lastly, thanks for the project.  It's working great for us so far!

izeye:
I created https://github.com/micrometer-metrics/micrometer-docs/pull/125 to revert https://github.com/micrometer-metrics/micrometer-docs/pull/120. I also added the reason for time window max to the doc based on the comment from @jkschneider.

PR #3129, opened 2022-04-14
OTLP registry

### LINKED ISSUE #2864: OpenTelemetry Protocol (OTLP) MeterRegistry
Use Micrometer to ship metrics in [OTLP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) format for metrics.

**Rationale**
Some metrics backends and metrics intermediaries like proxies/collectors/daemons that may do some processing of metrics before ingestion will support OpenTelemetry Protocol format. If there is enough interest from users, we should support this use case.

**Additional context**
This can be separate from a MeterRegistry implementation that delegates to OpenTelemetry metrics API. By focusing on the publishing format, this makes it most similar to other `MeterRegistry` implementations.

While the rest of the OTel metrics efforts are not stable yet, the OTLP metrics format is marked stable, unblocking this work.

shakuzen:
Implementation of publishing OTLP metrics format over HTTP in Protobuf. Some features are missing (see below) and overall there is room for improvement (e.g. performance), but I wanted to get this open for feedback and get a reasonable implementation in that we can iterate on before 1.9 GA.

Some mostly arbitrary choices I've made (feedback welcome):
* Naming convention: identity
* Base time unit: milliseconds

Follow-up later:
* Configurable temporality
* Min/max - need a new release of the metrics proto and to decide how to handle it (no min current, we intentionally don't align max with the interval or the lifetime of the application)
* encoding/transport options, compression

Closes #2864


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    @Override
+    protected <T> Gauge newGauge(Meter.Id id, T obj, ToDoubleFunction<T> valueFunction) {
+        return new DefaultGauge<>(id, obj, valueFunction);
+    }
+
+    @Override
+    protected Counter newCounter(Meter.Id id) {
+        return new StartTimeAwareCumulativeCounter(id, this.clock);
```
twicksell, line 105: The folks who own my metrics backend consistently ask me if we can support delta counters as part of our migration to OTLP. Do we have any thoughts yet on an implementation for delta counters in Micrometer?

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+                .setCount(histogramSnapshot.count());
+
+        // if histogram enabled, add histogram buckets
+        if (histogramSnapshot.histogramCounts().length != 0) {
+            for (CountAtBucket countAtBucket : histogramSnapshot.histogramCounts()) {
+                histogramDataPoint.addExplicitBounds(isTimeBased ? countAtBucket.bucket(getBaseTimeUnit()) : countAtBucket.bucket());
+                histogramDataPoint.addBucketCounts((long) countAtBucket.count());
+            }
+            metricBuilder.setHistogram(Histogram.newBuilder()
+                    .setAggregationTemporality(AggregationTemporality.AGGREGATION_TEMPORALITY_DELTA)
```
shakuzen, line 203: I'm testing out sending histogram data with buckets to an Otel Collector that exports Prometheus, and I don't see a histogram exported in the Prometheus endpoint with delta temporality. Also, bucket counts should not add the counts in lower buckets like Prometheus histograms.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+    private Metric writeHistogramSupport(HistogramSupport histogramSupport) {
+        Metric.Builder metricBuilder = getMetricBuilder(histogramSupport.getId());
+        boolean isTimeBased = histogramSupport instanceof Timer;
+        HistogramSnapshot histogramSnapshot = histogramSupport.takeSnapshot();
+        long wallTimeNanos = TimeUnit.MILLISECONDS.toNanos(this.clock.wallTime());
+
+        HistogramDataPoint.Builder histogramDataPoint = HistogramDataPoint.newBuilder()
+                .addAllAttributes(getTagsForId(histogramSupport.getId()))
+                // TODO start time
+                .setTimeUnixNano(wallTimeNanos)
```
twicksell, line 192: What about the max and mins? I was curious to see if you were going to use the Measurement#getStatistic to populate this or do something different.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -0,0 +1,261 @@
...
+
+    public OtlpMeterRegistry() {
+        this(OtlpConfig.DEFAULT, Clock.SYSTEM);
+    }
+
+    public OtlpMeterRegistry(OtlpConfig config, Clock clock) {
+        this(config, clock, new HttpUrlConnectionSender());
+    }
+
+    public OtlpMeterRegistry(OtlpConfig config, Clock clock, HttpSender httpSender) {
```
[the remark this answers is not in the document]
shakuzen: It's come up before as users have wanted to write metrics to e.g. kafka before storing them.
There might not be an open issue for it. Would you mind making one so we can track it? I don't think it'll be done as part of this pull request, though.

I suppose if we plan to eventually support grpc with this registry, it may be a mistake to expose `HttpSender` in the public API now.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
shakuzen: It's configurable in `SimpleMeterRegistry`. We could make it configurable in this registry too, since OTLP the format doesn't specify a preference as far as I'm aware.
What's the motivation behind the preference for delta counts?

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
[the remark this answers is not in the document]
twicksell: That was my thought. If we don't have a good generic interface for `Sender` today, that's ok. But maybe we can encapsulate the usage of HttpSender into some protected method that is easily overridden for folks who have some other requirement.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -0,0 +1,282 @@
...
+                .map(tag -> KeyValue.newBuilder()
+                        .setKey(tag.getKey())
+                        .setValue(AnyValue.newBuilder().setStringValue(tag.getValue()).build())
+                        .build())
+                .collect(Collectors.toList());
+    }
+
+    private Iterable<KeyValue> getResourceAttributes() {
+        List<KeyValue> attributes = new ArrayList<>();
+        // TODO How to expose configuration of the service.name
```
shakuzen, line 272: In, for example, the DatadogMeterRegistry, we take configuration for the tag name to convert to the Datadog-specified `host` tag. This allows you to configure a common tag that will be published to all registries, but will be converted to what Datadog expects.

I thought of doing something similar here, but it's a bit more tricky. We're only expecting one Resource, I think. And then the question is, which meter to look at to get the tag that corresponds to the Resource `service.name`? And when?

A configuration option to set this directly instead of taking it from a configured tag is another option, which we have done in some registries. But this can result in this `service.name` info only being in your OTLP metrics, while missing from meters in other registries, which is generally not desired.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
twicksell: That's a bit beyond my expertise, but supposedly it will allow for more accurate results in our use case and improve the query performance for at least one of our metrics backends. The Wavefront folks did a short writeup on pros/cons in their docs -> https://docs.wavefront.com/delta_counters.html

Using the same config options as SimpleMeterRegistry seems reasonable, but you also need to carry the context forward and set the right `AggregationTemporality`

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
shakuzen: max/min was added to the proto but hasn't been released yet, unfortunately: https://github.com/open-telemetry/opentelemetry-proto/compare/v0.16.0...main

Though as I was implementing this, there's a bit of a mismatch here and some choices. Min/max over the lifetime of the application don't seem particularly useful, but I think that's what we would be expected to send with cumulative temporality. I tried using delta temporality but the histogram didn't seem to get exported from the OTel collector to the Prometheus scape.

We've never supported a min, though it has been asked about for a backend that requires it for its histogram type. There's also the matter that we decay the max over multiple step intervals intentionally (reasoning described [here](https://github.com/micrometer-metrics/micrometer/issues/1993)), which is a different time window than the other stats. But we can only set one start/stop time for all of them.

twicksell: Sorry, I was just reading the proto directly and didn't consider that it hadn't been released yet. 

I don't have very strong opinions here, but I'm asking because we use a Micrometer meter registry today which publishes some of these using `statistic` as a tag. So I'll need to copy it forward or come up with a migration strategy.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -0,0 +1,301 @@
...
+        attributes.add(KeyValue.newBuilder().setKey("service.name").setValue(AnyValue.newBuilder().setStringValue("unknown_service")).build());
+        attributes.add(KeyValue.newBuilder().setKey("telemetry.sdk.name").setValue(AnyValue.newBuilder().setStringValue("io.micrometer")).build());
```
shakuzen, line 293: I wasn't sure what naming convention is best here. `io.micrometer` seemed reasonable as our reverse domain. Alternatively, maybe `micrometer.core` is better as it's our automatic module name for micrometer-core.
Documentation: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#telemetry-sdk

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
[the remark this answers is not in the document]
shakuzen: For now I've made the constructor private. We can consider what to expose as the use cases for customization are brought up by users. I think this gives us the most flexibility to evolve and add features while getting something usable out sooner.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
shakuzen: I've added it as a follow-up to make the temporality configurable. We'll see about doing it before 1.9 GA if it's important to users. I've left it out of this pull request for the sake of simplicity and time to get RC1 out.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
shakuzen: I'm not quite sure what we'll do, but we have some time to think about it, so I've marked it for follow-up after this pull request.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
izeye: How about simply `micrometer`?

shakuzen: That's probably also fine. I was basing my suggestions off the wording in the docs:
> If another SDK, like a fork or a vendor-provided implementation, is used, this SDK MUST set the attribute `telemetry.sdk.name` to the fully-qualified class or module name of this SDK's main entry point or another suitable identifier depending on the language.

Also, maybe we should distinguish between micrometer-core and tracing.

Issue #3066, opened 2022-03-10
Unable to load MacOSDnsServerAddressStreamProvider on Apple Silicon with statsd meter registries


simondemartini:
**Describe the bug**
When running a statsd meter registry on an Apple Silicon based machine with native (arm64) java, Micrometer logs the following error: 

```
ERROR io.micrometer.shaded.io.netty.resolver.dns.DnsServerAddressStreamProviders - Unable to load io.micrometer.shaded.io.netty.resolver.dns.macos.MacOSDnsServerAddressStreamProvider, fallback to system defaults. This may result in incorrect DNS resolutions on MacOS.
```

**Environment**
 - Micrometer version: 1.8.3
 - Micrometer registry: statsd
 - OS: macOS
 - CPU: Apple Silicon M1 (arm64)
 - Java version: 11.0.14 (Zulu, arm64 native)

**To Reproduce**
How to reproduce the bug:
Launch a Spring Boot application with a statsd meter registry on an Apple Silicon based computer. The error and full stack trace will be logged like the following:

```
1398 [main] ERROR io.micrometer.shaded.io.netty.resolver.dns.DnsServerAddressStreamProviders - Unable to load io.micrometer.shaded.io.netty.resolver.dns.macos.MacOSDnsServerAddressStreamProvider, fallback to system defaults. This may result in incorrect DNS resolutions on MacOS.
java.lang.reflect.InvocationTargetException: null
	at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:?]
	at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[?:?]
	at jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[?:?]
	at java.lang.reflect.Constructor.newInstance(Constructor.java:490) ~[?:?]
	at io.micrometer.shaded.io.netty.resolver.dns.DnsServerAddressStreamProviders.<clinit>(DnsServerAddressStreamProviders.java:64) ~[micrometer-registry-statsd-1.8.3.jar:1.8.3]
	at io.micrometer.shaded.io.netty.resolver.dns.DnsNameResolverBuilder.<init>(DnsNameResolverBuilder.java:60) ~[micrometer-registry-statsd-1.8.3.jar:1.8.3]
[...]

PR #3139, opened 2022-04-24
Include aarch_64 netty dns resolver for Apple silicon macs

simondemartini:
Fix for #3066 

It looks like the `x86_64` classifier was already included in the shaded jar, this PR just makes sure that we also include the `aarch64` classifier in the jar too so that we don't get the errors described in #3066 

I verified the native lib is included in the shaded jar, and that it's *not* included in the generated pom as expected. 

This appears to resolve the issues I was seeing on my computer.

### DIFF
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -10,17 +10,17 @@ dependencies {
         // We do not use HTTP modules; exclude to avoid false positive CVE reports
         exclude module: 'netty-codec-http'
     }
-	// We need make sure netty dns resolution works on all macs
-	implementation 'io.netty:netty-resolver-dns-native-macos:+:osx-x86_64'
-	implementation 'io.netty:netty-resolver-dns-native-macos:+:osx-aarch_64'
+    // We need make sure netty dns resolution works on all macs
+    implementation(group: 'io.netty', name: 'netty-resolver-dns-native-macos', classifier: 'osx-x86_64')
+    implementation(group: 'io.netty', name: 'netty-resolver-dns-native-macos', classifier: 'osx-aarch_64')
 
     testImplementation project(':micrometer-test')
     testImplementation 'io.projectreactor:reactor-test'
     testImplementation 'ch.qos.logback:logback-classic'
     testImplementation 'org.awaitility:awaitility'
     // for running tests with UDS on OSX
-    testImplementation 'io.netty:netty-transport-native-kqueue:+:osx-x86_64'
-    testImplementation 'io.netty:netty-transport-native-kqueue:+:osx-aarch_64'
+    testImplementation(group: 'io.netty', name: 'netty-transport-native-kqueue', classifier: 'osx-x86_64')
+    testImplementation(group: 'io.netty', name: 'netty-transport-native-kqueue', classifier: 'osx-aarch_64')
 }
 
 nebulaPublishVerification {
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -10,6 +10,9 @@ dependencies {
         // We do not use HTTP modules; exclude to avoid false positive CVE reports
         exclude module: 'netty-codec-http'
     }
+	// We need make sure netty dns resolution works on all macs
+	implementation 'io.netty:netty-resolver-dns-native-macos:+:osx-x86_64'
+	implementation 'io.netty:netty-resolver-dns-native-macos:+:osx-aarch_64'
 
     testImplementation project(':micrometer-test')
     testImplementation 'io.projectreactor:reactor-test'
@@ -53,7 +56,7 @@ publishing {
                 asNode()
                     .dependencies
                     .dependency
-                    .findAll { ['reactor-core', 'reactor-netty-core'].contains(it.artifactId.text()) }
+                    .findAll { ['reactor-core', 'reactor-netty-core', 'netty-resolver-dns-native-macos'].contains(it.artifactId.text()) }
                     .each { it.parent().remove(it) }
             }
         }
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -96,6 +96,7 @@ subprojects {
         }
       }
       implementation platform('io.projectreactor:reactor-bom:2020.0.+')
+	  implementation platform('io.netty:netty-bom:4.1.+')
     }
   }
 }


simondemartini:
@pivotal-cla This is an Obvious Fix

implementations/micrometer-registry-statsd/build.gradle
```diff
@@ -10,6 +10,9 @@ dependencies {
         // We do not use HTTP modules; exclude to avoid false positive CVE reports
         exclude module: 'netty-codec-http'
     }
+	// We need make sure netty dns resolution works on all macs
+	implementation 'io.netty:netty-resolver-dns-native-macos:+:osx-x86_64'
+	implementation 'io.netty:netty-resolver-dns-native-macos:+:osx-aarch_64'
```
shakuzen, line 15: Since this isn't a test dependency like the `netty-transport-native-kqueue` dependency, it's probably more important that the version is aligned with other netty versions. As is, I don't think it necessarily will be aligned if the netty version of other netty dependencies is not the latest version due to the `+`. We haven't explicitly managed the netty version so far since we only transitively depend on it through reactor-netty. Maybe that should change to add these dependencies properly.

shakuzen:
It was mostly intentional that these weren't included in the shading, since we don't (currently) expect the statsd module to run on macos in production. Adding them to the shaded JAR increases the JAR size for everyone, including those not on MacOS. Due to the way shading works, there isn't a good option for users to add the dependency when they need/want it, though, so we're in a bit of a tough position on this.
That all said, these two native libraries are pretty small - less than 20 KBs. So it's probably worth it to avoid errors on developer machines.

simondemartini:
Yeah unfortunately I think any fix we add here won't be perfect, but something is needed. This PR is just one option, other options might be (in no particular order): 

1. Convincing upstream [reactor-netty](https://github.com/reactor/reactor-netty) to include macos aarch64 support by default. It looks like aarch64 was left off the default transitive dependency list when m1 support was first added [in this PR](https://github.com/reactor/reactor-netty/pull/1893#issuecomment-978113811) but I wonder if that should be revisited? Especially as Apple Silicon becomes more dominant over time. Then all we have to do is update reactor-netty 🤷 
2. Create our own additional macos shaded classifiers that include *just* the native libs in their own shaded lib, then they could be added as needed (or added by default with the ability to exclude) but this might be more work than its worth and is one more thing micrometer has to maintain.
3. Update our usage of reactor-netty to not use the native libs at all for DNS resolution. This appears to "fix" the warnings in my local testing after playing with some `resolver()` options in the client builders, but I'm not familiar enough with netty to know the performance implications of this or if it would affect non-macos systems either.

implementations/micrometer-registry-statsd/build.gradle
simondemartini: Are you thinking we should add something like `implementation platform('io.netty:netty-bom:4.1.+')` to `dependencies.gradle` like we have set up for the [reactor-bom](https://github.com/micrometer-metrics/micrometer/blob/main/dependencies.gradle#L101)?

shakuzen:
@violetagg thoughts on the previous comment?

violetagg:
@simondemartini @shakuzen 

> Convincing upstream [reactor-netty](https://github.com/reactor/reactor-netty) to include macos aarch64 support by default. It looks like aarch64 was left off the default transitive dependency list when m1 support was first added https://github.com/reactor/reactor-netty/pull/1893#issuecomment-978113811 but I wonder if that should be revisited? Especially as Apple Silicon becomes more dominant over time. Then all we have to do is update reactor-netty 🤷

For the moment we do not plan to add this dependency.

> Create our own additional macos shaded classifiers that include just the native libs in their own shaded lib, then they could be added as needed (or added by default with the ability to exclude) but this might be more work than its worth and is one more thing micrometer has to maintain.

this is an option

> Update our usage of reactor-netty to not use the native libs at all for DNS resolution. This appears to "fix" the warnings in my local testing after playing with some resolver() options in the client builders, but I'm not familiar enough with netty to know the performance implications of this or if it would affect non-macos systems either.

The native implementation is better than the default one.

simondemartini:
@shakuzen any preference on any of those options? I'm thinking either something like this PR or option 2 is probably the best. Thoughts?

implementations/micrometer-registry-statsd/build.gradle
shakuzen: The challenge as I see it is there's some level of coordination to keep the netty versions in line with what works with reactor-netty. Though in theory there shouldn't be a problem with using a newer patch version of netty than reactor-netty compiles against. So it's probably fine to just control the major and minor as you suggest.

shakuzen: I'm not sure how we let the version defined in the platform BOM be used while specifying the classifier with Gradle, though. The whole `:+:classifier` syntax was used because I didn't know how else to specify a classifier without a version.

shakuzen:
@simondemartini thank you for the reminder.
I'm fine to go with the approach in this pull request, once the version management of netty dependencies is addressed. Though I think I'd rather merge it to the `main` branch for the next feature release instead of `1.8.x` for patch releases. My thinking is since nothing is functionally broken and the error logs are only on apple silicon which I'm expecting to generally be developer machines rather than production.

implementations/micrometer-registry-statsd/build.gradle
countableSet: Looking forward to this pr being merged. 🤞🏻 Just a heads up about

> I'm not sure how we let the version defined in the platform BOM be used while specifying the classifier with Gradle, though.

You can try the more verbose form instead to exclude the version if you did want to exclude it 👍🏻 

```groovy
implementation(group: "io.netty", name: "netty-resolver-dns-native-macos", classifier: "osx-x86_64")
implementation(group: "io.netty", name: "netty-resolver-dns-native-macos", classifier: "osx-aarch_64")
```

but then the `testImplementation` below doesn't so 🤷🏻‍♀️

simondemartini: I like that syntax, and I think it helps make it more clear. If either of you don't object, I'll update the test dependencies to use this same format since that `+` is a bit confusing.

I added the `netty-bom` and used the suggested new syntax and things look good.

I took a look at gradles resolution with both `./gradlew -q :micrometer-registry-statsd:dependencies` and `./gradlew -q :micrometer-registry-statsd:dependencyInsight --dependency io.netty:netty-resolver-dns-native-macos --configuration runtimeClasspath` and it looks like the netty versions are still 4.1.77.Final which matches what it was before this change

simondemartini:
Thanks, updated the PR from your feedback and changed the target to `main`. Let me know if you have any other thoughts

shakuzen (approved):

shakuzen:
Sorry about the test failure and thank you for the investigation. I have fixed that in #3213 and rebased the pull request on those changes so it is no longer failing.

PR #3148, opened 2022-04-28
Dynatrace registry: Truncate log output

arminru:
Sometimes the response body can be quite large so we should truncate it for logging.

Also logging stack traces from errors during HTTP requests only on debug level should be sufficient.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
@@ -317,7 +317,8 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
                         response.code(),
                         StringUtils.truncate(response.body(), LOG_RESPONSE_BODY_TRUNCATION_LIMIT)));
         } catch (Throwable throwable) {
-            logger.error("Failed metric ingestion: " + throwable.getMessage(), throwable);
+            logger.error("Failed metric ingestion: " + throwable);
+            logger.debug(throwable);
         }
     }
 
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
@@ -16,6 +16,8 @@
 package io.micrometer.dynatrace.v2;
 
 import com.dynatrace.metric.util.*;
+
+import io.micrometer.common.util.StringUtils;
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.distribution.HistogramSnapshot;
 import io.micrometer.core.instrument.distribution.ValueAtPercentile;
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -310,9 +313,12 @@ private void send(List<String> metricLines) {
                     .withPlainText(body)
                     .send()
                     .onSuccess(response -> handleSuccess(metricLines.size(), response))
-                    .onError(response -> logger.error("Failed metric ingestion: Error Code={}, Response Body={}", response.code(), response.body()));
+                    .onError(response -> logger.error("Failed metric ingestion: Error Code={}, Response Body={}",
+                        response.code(),
+                        StringUtils.truncate(response.body(), LOG_RESPONSE_BODY_TRUNCATION_LIMIT)));
```
jonatan-ivanov, line 318: What do you think about signaling that the body was truncated?
E.g.: if the length of `response.body()` is more than the limit, add `"[...]"` or `(truncated)`.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -310,9 +313,12 @@ private void send(List<String> metricLines) {
...
         } catch (Throwable throwable) {
-            logger.error("Failed metric ingestion: " + throwable.getMessage(), throwable);
+            logger.error("Failed metric ingestion: " + throwable);
+            logger.debug(throwable);
```
jonatan-ivanov, line 321: Is the idea of this coming from a user?
I usually like seeing the stacktrace if an error happened, it makes it easy where the error is coming from. Logging it out on `DEBUG` level defeats that purpose, you need to know where the log event is coming from so that you can see (and read the code) that a stacktrace is available on a lower level.

jonatan-ivanov: Also, the `DEBUG` event will not contain a message so it will look weird: it will not tell you what happened and nothing guarantees that these events will be next to (or close to) each other in the logs.

If we decide that we don't want the stacktrace on `ERROR` level (I would advise against it), I would state that the user sees this stacktrace because metric ingestion failed and there is another log event about it on `ERROR` level.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
jonatan-ivanov: I think it would make sense so that it is reusable across the code base.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
jonatan-ivanov: There is [`WarnThenDebugLogger`](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-commons/src/main/java/io/micrometer/common/util/internal/logging/WarnThenDebugLogger.java) that logs at warn level at first and then logs at debug level for the rest of the events.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
arminru: Added `StringUtils.truncate` + tests for it and changed the exporter to use it.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
arminru: I changed it to always print the message as a warning and use the WarnThenDebugLogger for the stack including a reference to the other warning message. This is under the assumption that it will usually be the same error if it would occur multiple times and that the overall context is clear from the log message prefix.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you! :)

izeye:
This seems to have been lost while merging https://github.com/micrometer-metrics/micrometer/pull/3152.

shakuzen:
> This seems to have been lost while merging #3152.

Thank you for catching that. It was my fault in messing up the rebase on #3152. I believe I've restored these changes now with ccbde40f27a432bb8461c1389d8839065aedd7a1

jonatan-ivanov:
fyi: I backported these to 1.9.

PR #3155, opened 2022-05-02
Dynatrace registry: Print ignored config warning only when necessary

pirgeo:
When using the new Dynatrace instruments introduced in #3093, a warning is almost always printed that states that the DistributionStatisticConfig is ignored. This PR ensures that the warning is printed only if there is an explicit attempt to modify histogram boundaries.


jonatan-ivanov (approved):

jonatan-ivanov:
Thanks! :)

jonatan-ivanov:
fyi: I backported these to 1.9.

pirgeo:
Thank you so much!

PR #3176, opened 2022-05-13
OkHttp3 instrumentation with observation api

marcingrzejszczak:


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
```diff
@@ -102,19 +85,24 @@ private static Method getMethod(Class<?>... parameterTypes) {
 
     private final boolean includeHostTag;
 
+    private final boolean observationRegistryNoOp;
```
jonatan-ivanov, line 88: Noop?

PR #3179, opened 2022-05-14
Use consistent casing for no-op

izeye:
This PR changes to use consistent casing for "no-op".


shakuzen (approved):
Looks good to me. The change to the `isNoOp` methods will break anyone using those already.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you! :)

Issue #3190, opened 2022-05-19
NPE in TimedAspect (JdkDynamicAopProxy)


aleksey-astashenko-jemmic:
**Describe the bug**
We had some implementation classes annotated with @Timed annotation. After migrating to 1.9.0 from 1.8.5 we're having NPE in TimedAspect class. It happens because of the part of the code from `io.micrometer.core.aop.TimedAspect#timedClass` method:

```
        Method method = ((MethodSignature) pjp.getSignature()).getMethod();
        Class<?> declaringClass = method.getDeclaringClass();
        Timed timed = declaringClass.getAnnotation(Timed.class);
```

Method being called through the interface reference, and the @Timed annotation is actually on the implementation class itself. Therefore, `declaringClass` resolved into interface class, and `timed` variable is `null`. Then we get NPE down the stack. 

Am I correct in assuming that we should have here some additional logic like in the `io.micrometer.core.aop.TimedAspect#timedMethod` metod? This is how it looks like:

```
        if (timed == null) {
            method = pjp.getTarget().getClass().getMethod(method.getName(), method.getParameterTypes());
            timed = method.getAnnotation(Timed.class);
        }
```

Additionally, seems like we're missing `shouldSkip` predicate check in `io.micrometer.core.aop.TimedAspect#timedClass`, although we have it in the `io.micrometer.core.aop.TimedAspect#timedMethod`:

```
        if (shouldSkip.test(pjp)) {
            return pjp.proceed();
        }
```

**Environment**
Micrometer version [1.9.0] (works on [1.8.5])
Micrometer registry [Prometheus]
OS: [Windows/Linux]

**Additional context**
It appears that it's related to this recently merged [PR](https://github.com/micrometer-metrics/micrometer/pull/1279)
Also little bit related to this [issue](https://github.com/micrometer-metrics/micrometer/issues/1262)

shakuzen:
Thank you for the detailed report and sorry for the issues. It sounds like you've got the right idea with the fixes. Would you be interested in contributing pull requests? Ideally two separate pull requests for the two separate issues.

aleksey-astashenko-jemmic:
@shakuzen , thank you for the quick response, and sorry for the delay from my side. Created two pull requests for both issues presented: https://github.com/micrometer-metrics/micrometer/pull/3194, https://github.com/micrometer-metrics/micrometer/pull/3195

shakuzen:
[...]

PR #3195, opened 2022-05-24
Get Timed annotation from proxy target when failed

aliaksei-astashenka:
Partially resolves https://github.com/micrometer-metrics/micrometer/issues/3190, the part with having an NPE when using interfaces


jonatan-ivanov:
Shouldn't we fix this in `1.8.x`?
Right now the PR is against `main` so it will only fix `1.10.x`.

PR #3201, opened 2022-05-26
Make descriptions static where they are not

jonatan-ivanov:
Since the `description` of the `Meter`s can be different per `Meter`, we can run into weird situations when the description is dynamic/different for different `Meter.Id`s that share the same name but differ in tags for backends that does not support different description per time series.

**jvm.threads.states**
The description here (Prometheus group) is `The current number of threads having NEW state`.
This is somewhat misleading since we have time series for all of the states not just `NEW`.

```
# HELP jvm_threads_states_threads The current number of threads having NEW state
# TYPE jvm_threads_states_threads gauge
jvm_threads_states_threads{state="new",} 0.0
jvm_threads_states_threads{state="timed-waiting",} 3.0
jvm_threads_states_threads{state="blocked",} 0.0
jvm_threads_states_threads{state="waiting",} 12.0
jvm_threads_states_threads{state="terminated",} 0.0
jvm_threads_states_threads{state="runnable",} 7.0
```

**logback.events**
The description here (Prometheus group) is `Number of error level events that made it to the logs`.
This is somewhat misleading since we have time series for all of the levels not just `error`.

```
# HELP logback_events_total Number of error level events that made it to the logs
# TYPE logback_events_total counter
logback_events_total{level="info",} 6.0
logback_events_total{level="debug",} 0.0
logback_events_total{level="warn",} 0.0
logback_events_total{level="error",} 0.0
logback_events_total{level="trace",} 0.0
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
@@ -71,7 +71,7 @@ public class JvmThreadMetrics implements MeterBinder {
             for (Thread.State state : Thread.State.values()) {
                 Gauge.builder("jvm.threads.states", threadBean, (bean) -> getThreadStateCount(bean, state))
                         .tags(Tags.concat(tags, "state", getStateTagValue(state)))
-                        .description("The current number of threads having " + state + " state")
+                        .description("The current number of threads")
                         .baseUnit(BaseUnits.THREADS).register(registry);
             }
         }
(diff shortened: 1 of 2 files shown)


jack-berg (approved):

julian-zhu96:
Hi jonatan,

I found there are a log of similar cases here, such as `cache.misses`, `cache.xa.commits`, will you fix all of them?

jonatan-ivanov:
@julian-zhu96 I can go through the cache-related binders, I see the ones you mentioned, do you happen to have a list about them?

jonatan-ivanov:
@julian-zhu96 fyi: https://github.com/micrometer-metrics/micrometer/issues/3321
/cc @jack-berg

julian-zhu96:
@jonatan-ivanov No, I don't have a list. But I found this to be the case for almost half of the meters, because I lost almost half of the microservice metrics data when collected by opentelemetry collector and sent to prometheus

jonatan-ivanov:
Check with the latest snapshot, theoretically I fixed it for cache metrics.

PR #3215, opened 2022-06-10
Add a LongTaskTimer for every Observation

jonatan-ivanov:
This PR changes the behavior of `TimerObservationHandler` so that it will always create an additional `LongTaskTimer` (on top of the `Timer`) so that the active task can be observed too on-the-fly.

There is a caveat/issue with this: since the `KeyValuesProvider` is used at `stop`, none of those tags will be added to the `LongTaskTimer`. Also, the tags that are added to the `Observation` after `start` won't be added to the `LongTaskTimer`.

/cc @cppwfs


micrometer-core/src/main/java/io/micrometer/core/instrument/observation/TimerObservationHandler.java
```diff
@@ -15,10 +15,7 @@
  */
 package io.micrometer.core.instrument.observation;
 
-import io.micrometer.core.instrument.MeterRegistry;
-import io.micrometer.core.instrument.Tag;
-import io.micrometer.core.instrument.Tags;
-import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.*;
```
cppwfs, line 18: Do we allow for wildcard imports?   Annoying, aren't I? :-D

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/ObservationHandlerSample.java
```diff
@@ -42,12 +42,17 @@ public static void main(String[] args) throws InterruptedException {
...
-                .lowCardinalityKeyValue("a", "1").highCardinalityKeyValue("time", Instant.now().toString()).start();
+                .lowCardinalityKeyValue("a", "1").highCardinalityKeyValue("time", Instant.now().toString()).longTask()
+                .start();
 
         try (Observation.Scope scope = observation.openScope()) {
             Thread.sleep(1_000);
             observation.error(new IOException("simulated"));
         }
+
+        System.out.println("---");
```
cppwfs, line 53: Do we want to signal that this is before the observation stop?  i.e. `--- Before Stop ---`.   And for the System.out.println below have a similar signal `--- After Stop --- `.

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/TimerObservationHandler.java
jonatan-ivanov: Yes (I mean the wildcard imports :D), IntelliJ automatically does this, based on our [`.editorconfig`](https://github.com/micrometer-metrics/micrometer/blob/main/.editorconfig) file

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/ObservationHandlerSample.java
jonatan-ivanov: 👍🏼

shakuzen (approved):
I like this. I think it's worthwhile in most cases to have a LTT along with a Timer to see metrics on in-flight tasks. In theory, if something being timed has a timeout that's less than the step interval - meaning an active task would never be timed across two step intervals, then I guess the LTT becomes less interesting. As long as the overhead of the LTT isn't a problem, though, I think it is fine.

The difference in tags is a potential point of confusion. We should document it in the JavaDoc perhaps.

micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryCompatibilityKit.java
```diff
@@ -634,14 +635,23 @@ void recordWithSample() {
         @Test
         @DisplayName("record with stateful Observation instance")
         void recordWithObservation() {
-            Observation observation = Observation.start("myObservation", observationRegistry);
-            clock(registry).add(10, TimeUnit.NANOSECONDS);
+            Observation observation = Observation.createNotStarted("myObservation", observationRegistry)
+                    .lowCardinalityKeyValue("staticTag", "42").start();
+
+            // created after start, LongTaskTimer won't have it
+            observation.lowCardinalityKeyValue("dynamicTag", "24");
```
marcingrzejszczak, line 642: In general that shouldn't really happen. We will recommend to the users that they create a non started observation, and always use a key value provider (I think that in extremely rare cases they could set key value pairs manually)

jonatan-ivanov: The `KeyValueProvider` is the same case since it is queried at `stop` when you have all the details in the context and the `LongTaskTimer` is created at `start`.

PR #3216, opened 2022-06-11
Avoid auto-boxing in `Measurement#getValue`

marschall:
Use DoubleSupplier instead of Supplier<Double> where possible to avoid
boxing.


marschall:
I let the original constructor in for binary backwards compatibility.

Issue #1586, opened 2019-09-14
Improve Tag propagation for @Timed


jonatan-ivanov:
Currently, there isn't an easy way to add tags to a `Timer` if it was created through the `TimedAspect` using the `@Timed` annotation.

I’m thinking about implementing two things:
1.  A class ([`ExtraTagsPropagation`](https://github.com/jonatan-ivanov/micrometer-tags/blob/master/src/main/java/com/example/micrometertags/ExtraTagsPropagation.java)) with a few static methods (like SLF4J's `MDC` or Spring Cloud Sleuth's `ExtraFieldPropagation`) which add tags to a `ThreadLocal` and `TimedAspect` could get extra tags from it, e.g.:
```
@Timed
void testMethod(String param) {
    ExtraTagsPropagation.add("testTag", someValue);
    ...
}
```

2.  An annotation ([`ExtraTag`](https://github.com/jonatan-ivanov/micrometer-tags/blob/master/src/main/java/com/example/micrometertags/ExtraTag.java)) what you can put on the method params and `TimedAspect` could get the extra tags this way:

```
@Timed
void testMethod(@MetricTag("testTag") String param) {...}
```

I created a [PoC](https://github.com/jonatan-ivanov/micrometer-tags/) for this, you can see them in use in [`EchoService`](https://github.com/jonatan-ivanov/micrometer-tags/blob/master/src/main/java/com/example/micrometertags/service/EchoService.java).
The `TimedAspect` needs to be modified, for the sake of simplicity, the PoC does not do it, instead, it extends its `Function`: [`MetricsConfig`](https://github.com/jonatan-ivanov/micrometer-tags/blob/master/src/main/java/com/example/micrometertags/MetricsConfig.java).

If this is something you would accept, I would be happy to go ahead, polish this a little, modify the `TimedAspect`, add some docs and tests and open a PR.
What do you think?

shakuzen:
This is something that's been asked about before. We don't do general context propagation right now. While `ThreadLocal` should work fine in thread-based runtimes, with other runtimes (e.g. Netty, Reactor) I think there's probably more required.
I'd have to look more into this later.

jonatan-ivanov:
@shakuzen I might be missing something but since `TimedAspect` is calling the actual method synchronously (through the joinpoint) if `TimedAspect` works, `ThreadLocal` must too since you can't change threads there.
[...]

PR #3221, opened 2022-06-15
@Observed

jonatan-ivanov:


micrometer-observation/src/main/java/io/micrometer/observation/annotation/Observed.java
```diff
@@ -0,0 +1,53 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.micrometer.observation.Observation;
+
+/**
+ * @author Jonatan Ivanov
```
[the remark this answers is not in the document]
jonatan-ivanov: fixed

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -0,0 +1,175 @@
...
+ * @author Jonatan Ivanov
+ * @since 1.10.0
+ */
+@Aspect
+@NonNullApi
+public class ObservedAspect {
+
+    private static final String DEFAULT_OBSERVATION_NAME = "method.observed";
+
+    private static final Predicate<ProceedingJoinPoint> DONT_SKIP_ANYTHING = pjp -> false;
```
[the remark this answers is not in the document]
jonatan-ivanov: won't fix

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -0,0 +1,175 @@
...
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.reflect.MethodSignature;
+
+import java.lang.reflect.Method;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Predicate;
+
+/**
+ * @author Jonatan Ivanov
```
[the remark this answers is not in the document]
jonatan-ivanov: fixed

shakuzen (approved):

micrometer-observation/src/main/java/io/micrometer/observation/annotation/Observed.java
```diff
@@ -0,0 +1,55 @@
...
+     * Contextual name of the {@link Observation}.
+     * @return contextual name of the {@link Observation}
+     */
+    String contextualName() default "";
+
+    /**
+     * Low cardinality key values.
+     * @return an array of low cardinality key values.
+     */
+    String[] lowCardinalityKeyValues() default {};
```
marcingrzejszczak, line 53: Why aren't we allowing to set `highCardinalityKeyValues` ? 

Also, shouldn't we give an option to provide a `KeyValuesProvider` class ? It would have to have a default constructor (when someone doesn't want to calculate the tags from the annotated method arguments) or (maybe) a constructor that takes in `Object[]` where the array would be mapped to methods arguments? Or even maybe also a `Method` argument? Does it make any sense? :thinking:

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    private Object observe(ProceedingJoinPoint pjp, Method method, Observed observed) throws Throwable {
+        String name = observed.name().isEmpty() ? DEFAULT_OBSERVATION_NAME : observed.name();
+        Signature signature = pjp.getStaticPart().getSignature();
+        String contextualName = observed.contextualName().isEmpty() ? getContextualName(signature)
+                : observed.contextualName();
+
+        Observation observation = Observation.createNotStarted(name, new ObservedAspectContext(pjp), registry)
+                .contextualName(contextualName).lowCardinalityKeyValue("class", signature.getDeclaringTypeName())
```
marcingrzejszczak, line 141: maybe this `class` and `method` should be coming from a `DocumentedObservation`? Also maybe it should be `class.name` and `class.method` or sth like that - cause `method` can be an HTTP method? Or am I exaggerating?

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -158,6 +168,15 @@ default Observation highCardinalityKeyValue(String key, String value) {
...
     }
 
+    /**
+     * Adds multiple high cardinality key value instances. High cardinality means that the
+     * key value will have possible an unbounded number of possible values. An HTTP URL is
+     * a good example of such a key value (e.g. /foo/bar, /foo/baz etc.).
+     * @param keyValues key value instances
+     * @return this
+     */
+    Observation highCardinalityKeyValue(Iterable<KeyValue> keyValues);
```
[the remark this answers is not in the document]
jonatan-ivanov: 👍🏼

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistryAssert.java
```diff
@@ -125,6 +125,13 @@ public That hasObservationWithNameEqualToIgnoringCase(String name) {
         return new That(testObservationContext, this);
     }
 
+    public void doesNotHaveAnyObservation() {
+        List<TestObservationRegistry.TestObservationContext> contexts = this.actual.getContexts();
+        if (!contexts.isEmpty()) {
+            failWithMessage("There were %d observation(s) registered in the registry, expected 0.", contexts.size());
```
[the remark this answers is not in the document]
jonatan-ivanov: 👍🏼

micrometer-observation/src/main/java/io/micrometer/observation/annotation/Observed.java
jonatan-ivanov: >Why aren't we allowing to set highCardinalityKeyValues ?

Because attribute values of annotations must be constants, you can't have dynamic values since you can't even call a method there, only literals and constants are allowed.

>Does it make any sense?

It makes sense to me. I did write a [PoC](https://github.com/jonatan-ivanov/micrometer-tags/blob/master/src/main/java/com/example/micrometertags/service/EchoService.java) and also proposed a PR for a pretty similar feature for `@Timed` years ago 😄, see https://github.com/micrometer-metrics/micrometer/issues/1586

Check out the issue (there is also a PR), the [PoC](https://github.com/jonatan-ivanov/micrometer-tags/blob/master/src/main/java/com/example/micrometertags/service/EchoService.java) tries to do two things:
1. Add an `@ExtraTag` annotation which is like `@SpanTag` in Sleuth and somewhat can cover the `Object[]`/`Method`/`JointPont` scenario that you mentioned above.
2. An `ExtraTagsPropagation` class which is similar to Sleuth 2's `ExtraFieldPropagation`.

I think if we want to do something like this (I think it would be a nice feature) we should keep `@Timed` and `@Observed` in feature parity to a degree.
Also, I think we should discuss about two things in terms of `@Observed`:
1. Setting keyvalues based on the jointpoint/method/params (an extra annotation can cover that, see above)
2. Setting keyvalues based on the context

Can we move this to a separate issue/PR?

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
jonatan-ivanov: I'm taking a look at using the DocumentedObservation here, `class.name` maybe `method.name` makes sense we need to check the naming conventions, since right now it is `class` and `method` in `TimedAspect`.

jonatan-ivanov: fixed

shakuzen (approved):

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -0,0 +1,260 @@
...
+        public String getContextualName() {
+            return "%s";
+        }
+
+        @Override
+        public KeyName[] getLowCardinalityKeyNames() {
+            return ObservedAspectLowCardinalityKeyName.values();
+        }
+
+        public enum ObservedAspectLowCardinalityKeyName implements KeyName {
```
marcingrzejszczak, line 226: If this is not set as an enum in a separate file we will have a problem with micrometer docs generator

jonatan-ivanov: You mean `ObservedAspectObservation` should be in a separate file or both `ObservedAspectObservation` and `ObservedAspectLowCardinalityKeyName`?

I'm asking because `DocumentedObservation` says:
> If the observation contains {@link KeyName} then those need to be declared as nested enums

marcingrzejszczak: The Observation should be in a separate file and it should have a nested enum with keys

jonatan-ivanov: fixed in https://github.com/micrometer-metrics/micrometer/commit/86befc188c7da809fd5e5c5ca6c106601797f435
Thank you!

PR #3230, opened 2022-06-15
Add primitive Suppliers to Timer and LongTaskTimer

### LINKED ISSUE #1304: Add primitive Suppliers to Timer
Tools like findbugs/spotbugs will complain about needless boolean/Boolean conversion, as well as the other primitive types (int/Integer) when using ``<T> T record(Supplier<T> f)``

A simple fix would be to add these methods to the ``Timer`` interface:
``boolean record(BooleanSupplier f)``
``double record(DoubleSupplier f)``
``int record(IntSupplier f)``
``long record(LongSupplier f)``

marschall:
Fixes #1304

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -348,7 +348,16 @@ subprojects {
                     ignoreMissingClasses = true
                     includeSynthetic = true
                     // TODO remove methodExcludes when gh-3181 is resolved
-                    methodExcludes = ['io.micrometer.stackdriver.StackdriverConfig#metricTypePrefix()']
+                    methodExcludes = ['io.micrometer.stackdriver.StackdriverConfig#metricTypePrefix()',
+                            'io.micrometer.core.instrument.Timer#record(java.util.function.BooleanSupplier)',
+                            'io.micrometer.core.instrument.Timer#record(java.util.function.IntSupplier)',
+                            'io.micrometer.core.instrument.Timer#record(java.util.function.LongSupplier)',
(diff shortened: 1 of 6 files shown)


marschall:
There is a bit of duplication between `AbstractTimer` and `CompositeTimer` due to a different superclass.

I'm not sure whether this will make a noticeable performance difference but users of primitive type should no longer get `null` warnings due to `@Nullable` on `io.micrometer.core.instrument.Timer.record(Supplier<T>)`. This seems to have been the initial motivation behind #1304. Interestingly enough `io.micrometer.core.instrument.LongTaskTimer.record(Supplier<T>)` has no `@Nullable`.

shakuzen:
> Interestingly enough `io.micrometer.core.instrument.LongTaskTimer.record(Supplier<T>)` has no `@Nullable`.

That's an omission. We can fix that separately. Thank you for noticing and letting us know.

> I believe the japicmp build steps fail due these new methods being new methods on the `Timer` and `LongTaskTimer` interfaces.

Yes, though since you've provided default implementations, it should not be an issue in practice. We ran into this previously and unfortunately need to add explicit build configuration to ignore these new methods until there is a better way to deal with this behavior of japicmp and the gradle plugin. See https://github.com/shakuzen/micrometer/commit/5d3b3bd8b4119194b3bc525c4d5701363da58ee7.

shakuzen (approved):

shakuzen:
Thank you for the pull request!

PR #3240, opened 2022-06-22
StepMeterRegistry: wait for stepmeters to finalize on close

### LINKED ISSUE #1882: Counters do not increment when ElasticMeterRegistry export is enabled
_Hi! Had posted this in Slack, but thought to report it here as well pending response there_

In short: I have a Spring Boot 2.2.5 application with `spring-boot-starter-actuator` and `micrometer-registry-elastic`; when I set `management.metrics.export.elastic.enabled=true`, then my counters do not increment anymore.

Not sure if this should be a micrometer issue or related to Spring Boot; `ElasticMeterRegistry` is in micrometer whereas `ElasticMetricsExportAutoConfiguration` is in Spring Boot. But since the latter seems to mostly delegate and is rather minimal otherwise, I thought to report it here first.

**To replicate the problem:**

1. Create a new project with: https://start.spring.io/#!type=maven-project&language=java&platformVersion=2.2.5.RELEASE&packaging=jar&jvmVersion=11&groupId=com.example&artifactId=demo&name=demo&description=Demo%20project%20for%20Spring%20Boot&packageName=com.example.demo&dependencies=actuator

2. Add a dependency on micrometer-registry-elastic
```
<dependency>
	<artifactId>micrometer-registry-elastic</artifactId>
</dependency>
```

3. Create a test class containing

```java
@Test
void counterStaysAtZero() {
	DebugApplication.main("", "--management.metrics.export.elastic.enabled=true");
	// DebugApplication : Counter: 0.0 , [Measurement{statistic='COUNT', value=0.0}]
	// DebugApplication : Counter: 0.0 , [Measurement{statistic='COUNT', value=0.0}]
}
@Test
void counterAddsToOne() {
	DebugApplication.main("", "--management.metrics.export.elastic.enabled=false");
	// DebugApplication : Counter: 0.0 , [Measurement{statistic='COUNT', value=0.0}]
	// DebugApplication : Counter: 1.0 , [Measurement{statistic='COUNT', value=1.0}]
}
@SpringBootApplication
@RequiredArgsConstructor
@Slf4j
public static class DebugApplication implements ApplicationRunner {
	public static void main(String... args) {
		log.info("{}", args[1]);
		SpringApplication.run(DebugApplication.class, args).close();
	}
	private final MeterRegistry registry;
	@Override
	public void run(ApplicationArguments args) throws Exception {
		Counter counter = registry.counter("someCounter");
		log.info("Counter: {} , {}", counter.count(), counter.measure());
		counter.increment();
		log.info("Counter: {} , {}", counter.count(), counter.measure());
	}
}

If you then run both tests you will get the log output that's listed in each test method.

I would expect the Counter to increment regardless of whether metrics are exported; What can be done to make this so the counters increase?

### PR DESCRIPTION
**Pull request by monosoul**
This is a minimal change required to fix #1882 .

Introduces:

- A test case to verify the metrics published on close are complete and not published twice (with background publishing thread).
- `close()` method override in `StepMeterRegistry` to wait for stepmeters to finalize values.

As a drawback, service shutdown might take longer than usual (up to step duration).

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
@@ -102,4 +102,15 @@ public abstract class StepMeterRegistry extends PushMeterRegistry {
                 .merge(DistributionStatisticConfig.DEFAULT);
     }
 
+    @Override
+    public void close() {
+        stop();
+        try {
+            Thread.sleep(config.step().toMillis());
+        } catch (final InterruptedException e) {
+            // ignore
+        } finally {
+            super.close();
+        }
+    }
 }
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/step/StepMeterRegistryTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/step/StepMeterRegistryTest.java
@@ -22,6 +22,7 @@ import org.junit.jupiter.api.Test;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -44,6 +45,11 @@ class StepMeterRegistryTest {
     private MockClock clock = new MockClock();
 
     private StepRegistryConfig config = new StepRegistryConfig() {
+        @Override
+        public Duration step() {
+            return Duration.ofSeconds(1);
+        }
+
         @Override
         public String prefix() {
             return "test";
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java [hunk 1]
```diff
@@ -102,4 +102,15 @@ protected DistributionStatisticConfig defaultHistogramConfig() {
                 .merge(DistributionStatisticConfig.DEFAULT);
     }
 
+    @Override
+    public void close() {
```
jonatan-ivanov, line 106: The default step is `1m` (can be bigger), are you sure everyone wants to `sleep` that much and block `close`?

Also, isn't the original issues for short-lived processes? Isn't blocking their shutdown process counter-intuitive?

monosoul: > The default step is 1m (can be bigger), are you sure everyone wants to sleep that much and block close?

In our use case we don't care how long does it take to shut down the process as long as it publishes the metrics.

> Also, isn't the original issues for short-lived processes? Isn't blocking their shutdown process counter-intuitive?

The processes are probably short-lived not because people want them to be short-lived, but because the work they do is performed fast. At least that's true for our use case.

In our case we have a long running service in a container in Kubernetes, and a cronjob scheduled in Kubernetes as well that uses the same image to do some work. The time the cronjob runs depends on how much work does it have to do, but typically it takes less than a minute. We don't really care if it takes it longer to shut down as long as it is guaranteed to publish complete metrics, because we rely on those metrics for business insights and for alerts.

monosoul: Also, just for the cronjob we configured `step` duration to 5 seconds.

jonatan-ivanov: >In our use case we don't care how long does it take to shut down the process as long as it publishes the metrics. We can hide this behavior behind an option I guess?

I understand that but I think it would be a fair assumption saying that most of the users do care since it is highly likely that by default this will result in blocking the app from gracefully shutting down for a while, then a violent `SIGKILL` (~`kill -9`) of the app. By default closing the registry should be registered in a shutdown hook so once the JVM wants to terminate, it will try to close the registry and wait... and wait... until the scheduler/orchestrator gets bored that the app is not shutting down and removes the app from memory violently.

I think an opt-in config option could help but 

>The processes are probably short-lived not because people want them to be short-lived, but because the work they do is performed fast.

I'm not sure I understand this, but my point with mentioning the short-lived processes was pointing out that this change will block them being really short-lived.
Imagine an app that is deployed into a "serverless"/FaaS (Function as a Service) environment where the orchestrator can shut down the apps (JVMs) frequently (when they are idle). In this scenario, blocking the app from shutting down can be sub-optimal.

>because we rely on those metrics for business insights and for alerts.

This might be a disconnect. Micrometer is trying not to disrupt the application while it is doing its job. This means that it might not be a good choice for business metrics/accounting/financing because it can drop data points (i.e.: in case of a network issue).

shakuzen:
Thank you for the pull request. We are going to try to solve the underlying problem by sending metrics from partial step intervals on the final push. We appreciate the effort here, but ultimately we think taking the other approach will be better in general for a broad range of use cases.

PR #3255, opened 2022-06-29
New approach to naming conventions

marcingrzejszczak:
with this change we're merging semantic name provider and key values provider into a single ObservationConvention construct
users can register their conventions in the observationregistry configuration and then in the instrumentations ask for a registered convention (and provide a default if one is not registered)


micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
```diff
@@ -55,6 +55,32 @@ class SimpleObservation implements Observation {
                 .filter(handler -> handler.supportsContext(this.context))
                 .collect(Collectors.toCollection(ArrayDeque::new));
         this.filters = registry.observationConfig().getObservationFilters();
+        List<ObservationConvention<?>> observationConventions = registry.observationConfig().getObservationConventions()
+                .stream().filter(observationConvention -> observationConvention.supportsContext(this.context))
+                .collect(Collectors.toList());
+        if (!observationConventions.isEmpty()) {
+            this.keyValuesProviders.addAll(observationConventions);
+            this.context.setName(observationConventions.get(0).getName());
```
jonatan-ivanov, line 63: Can we do an `isBlank` check here? Let's say you don't want to provide a name so you return `null` or `""`.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
```diff
@@ -67,7 +67,7 @@ public class OkHttpMetricsEventListener extends EventListener {
 
     private final ObservationRegistry observationRegistry;
 
-    private final Observation.KeyValuesProvider<OkHttpContext> keyValuesProvider;
+    private final Observation.ObservationConvention<OkHttpContext> keyValuesProvider;
```
mhalbritter, line 70: I would rename this field, as it's now a `ObservationConvention`.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
```diff
@@ -91,12 +91,12 @@ public class OkHttpMetricsEventListener extends EventListener {
...
             Iterable<BiFunction<Request, Response, Tag>> contextSpecificTags) {
-        this(registry, ObservationRegistry.NOOP, new DefaultOkHttpKeyValuesProvider(), requestsMetricName, urlMapper,
-                extraTags, contextSpecificTags, emptyList(), true);
+        this(registry, ObservationRegistry.NOOP, new DefaultOkHttpObservationConvention(requestsMetricName),
+                requestsMetricName, urlMapper, extraTags, contextSpecificTags, emptyList(), true);
     }
 
     OkHttpMetricsEventListener(MeterRegistry registry, ObservationRegistry observationRegistry,
-            Observation.KeyValuesProvider<OkHttpContext> keyValuesProvider, String requestsMetricName,
+            Observation.ObservationConvention<OkHttpContext> keyValuesProvider, String requestsMetricName,
```
mhalbritter, line 99: Rename that parameter, too.

micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
```diff
@@ -55,6 +55,32 @@ class SimpleObservation implements Observation {
                 .filter(handler -> handler.supportsContext(this.context))
                 .collect(Collectors.toCollection(ArrayDeque::new));
         this.filters = registry.observationConfig().getObservationFilters();
+        List<ObservationConvention<?>> observationConventions = registry.observationConfig().getObservationConventions()
```
mhalbritter, line 58: When using the constructor with the `ObservationConvention<?>` parameter, only the given `ObservationConvention` is applied. In this case, ALL matching conventions are applied. I wonder if this is intended? Why doesn't this code use `io.micrometer.observation.ObservationRegistry.ObservationConfig#getObservationConvention`, which returns the first matching one? I must admit i'm a bit confused, so I can be totally wrong here.

micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
```diff
@@ -55,6 +55,32 @@ class SimpleObservation implements Observation {
...
+        this.handlers = registry.observationConfig().getObservationHandlers().stream()
+                .filter(handler -> handler.supportsContext(this.context))
+                .collect(Collectors.toCollection(ArrayDeque::new));
+        this.filters = registry.observationConfig().getObservationFilters();
+        this.keyValuesProviders.add(convention);
+    }
+
+    private static String name(ObservationConvention<?> convention, Context context) {
+        if (!convention.supportsContext(context)) {
+            throw new IllegalStateException("Convention [" + convention + "] can't support context [" + context + "]");
```
mhalbritter, line 81: "can't" -> "doesn't"?

wilkinsona (approved):
LGTM

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpDocumentedObservation.java
```diff
@@ -51,29 +50,26 @@ public KeyName[] getLowCardinalityKeyNames() {
...
      * @return a new {@link OkHttpDocumentedObservation}
      */
+    @SuppressWarnings("unchecked")
     static Observation of(@NonNull ObservationRegistry registry, @NonNull OkHttpContext okHttpContext,
             @NonNull String requestsMetricName,
-            @Nullable Observation.KeyValuesProvider<OkHttpContext> keyValuesProvider) {
-        Observation.KeyValuesProvider<?> provider = null;
+            @Nullable Observation.ObservationConvention<OkHttpContext> customConvention) {
+        Observation.ObservationConvention<OkHttpContext> convention = null;
         if (registry.isNoop()) {
```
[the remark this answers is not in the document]
shakuzen: But should every instrumentation check if the registry is no-op here? The call to create later will return a no-op already if the registry is a no-op registry.

PR #3258, opened 2022-07-01
WIP Http client instrumentation TCK

shakuzen:
Introduce test suite for HTTP client instrumentations to check that all implementations have the naming and tags expected. This may also help instrumentors ensure backwards compatibility of the HTTP client instrumentation across changes to it.


micrometer-test/src/main/java/io/micrometer/core/instrument/HttpClientTimingInstrumentationVerificationSuite.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * A default is provided that should be preferred by new instrumentations. Existing
+     * instrumentations that use a different value to maintain backwards compatibility may
+     * override this method to run tests with a different name used in assertions.
+     * @return name of the meter timing http client requests
+     */
+    protected String timerName() {
+        return "http.client.requests";
```
marcingrzejszczak, line 50: A future enhancement idea: We could inject the new naming convention mechanism here - that way if someone wants to see if they are compatible with a given standard they will be able to easily do it

marcingrzejszczak (approved):

micrometer-test/src/main/java/io/micrometer/core/instrument/HttpClientTimingInstrumentationVerificationSuite.java
```diff
@@ -0,0 +1,131 @@
...
+    void unmappedUrisAreCardinalityLimited(WireMockRuntimeInfo wmRuntimeInfo) {
+        stubFor(get(anyUrl()).willReturn(notFound()));
+
+        String templatedPath = "/notFound404";
+        sendHttpRequest(HttpMethod.GET, URI.create(wmRuntimeInfo.getHttpBaseUrl()), templatedPath);
+
+        Timer timer = getRegistry().get(timerName()).tags("method", "GET", "status", "404").timer();
+        // we should standardize on a value e.g. NOT_FOUND, but for backwards
+        // compatibility, assert is more lenient
+        assertThat(timer.getId().getTag("uri")).isNotEqualTo(templatedPath);
```
shakuzen, line 114: The Apache HTTP client instrumentation is failing this assertion now. We should fix that.

shakuzen: While we should definitely test and enforce this for HTTP server instrumentation, it's not the same situation for HTTP client instrumentation. Maybe relaxing this is better.

shakuzen:
Since we want to verify existing behavior is maintained, it makes more sense to me to merge this into the oldest supported branch. Rebasing to merge into 1.8.x instead of 1.9.x

PR #3261, opened 2022-07-04
Further improvement of observation convention

marcingrzejszczak:
Makes `DocumentedObservation` require either `getName()` or `getDefaultConvention()` methods to be provided in order to resolve the final name for an observation.


shakuzen (approved):
LGTM

Issue #1599, opened 2019-09-23
Accidental strict dep on LatencyUtils


codefromthecrypt:
It seems there are some guards to try to not use LatencyUtils unless pause detection is in use. However, there's still a compile dep.

```
[INFO] Caused by: java.lang.ClassNotFoundException: org.LatencyUtils.IntervalEstimator
[INFO] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:471) ~[?:?]
[INFO] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:588) ~[?:?]
[INFO] 	at org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:93) ~[zipkin-server-2.16.3-SNAPSHOT-slim.jar:?]
[INFO] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:521) ~[?:?]
[INFO] 	at io.micrometer.prometheus.PrometheusMeterRegistry.newTimer(PrometheusMeterRegistry.java:176) ~[micrometer-registry-prometheus-1.2.1.jar!/:1.2.1]
[INFO] 	at io.micrometer.core.instrument.MeterRegistry.lambda$timer$2(MeterRegistry.java:271) ~[micrometer-core-1.2.1.jar!/:1.2.1]
[INFO] 	at io.micrometer.core.instrument.MeterRegistry.getOrCreateMeter(MeterRegistry.java:576) ~[micrometer-core-1.2.1.jar!/:1.2.1]
[INFO] 	at io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:529) ~[micrometer-core-1.2.1.jar!/:1.2.1]
[INFO] 	at io.micrometer.core.instrument.MeterRegistry.timer(MeterRegistry.java:269) ~[micrometer-core-1.2.1.jar!/:1.2.1]
[INFO] 	at io.micrometer.core.instrument.Timer$Builder.register(Timer.java:477) ~[micrometer-core-1.2.1.jar!/:1.2.1]
[INFO] 	at io.micrometer.core.instrument.MeterRegistry.timer(MeterRegistry.java:396) ~[micrometer-core-1.2.1.jar!/:1.2.1]
[INFO] 	at io.micrometer.core.instrument.internal.TimedExecutorService.<init>(TimedExecutorService.java:40) ~[micrometer-core-1.2.1.jar!/:1.2.1]
[...]

PR #3263, opened 2022-07-05
Declare HdrHistogram as a runtime dependency

shakuzen:
HdrHistogram is not exposed in the API. It is an implementation detail of client-side percentiles. The scope of the dependency can be changed accordingly.

This was found while working on #1599.

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -8,8 +8,7 @@ dependencies {
     api project(":micrometer-commons")
     api project(":micrometer-observation")
 
-    // TODO(anuraaga): HdrHistogram is exposed in the micrometer API but probably shouldn't be
-    api 'org.hdrhistogram:HdrHistogram'
+    implementation 'org.hdrhistogram:HdrHistogram'
     implementation('org.latencyutils:LatencyUtils') {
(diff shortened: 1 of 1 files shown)


shakuzen:
I may have been too eager with this change. HdrHistogram's `DoubleHistogram` and `DoubleRecorder` is part of the (package-private) API in the public class TimeWindowPercentileHistogram:
https://github.com/micrometer-metrics/micrometer/blob/4f3cf5ac6ca4bb5a8672f8a0308bb1ac0868f439/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowPercentileHistogram.java#L42-L65

Though I'm not sure if that is a problem for changing the scope here or not. Outside of the local package, I don't think there is reference to any HdrHistogram classes in API that users would be able to compile against.

jonatan-ivanov (approved):

shakuzen:
With https://github.com/micrometer-metrics/micrometer/pull/3325 in now, I think we can give this a try in 1.10. User code should not compile against Micrometer API that has the HdrHistogram API in it with normal usage that I can think of. If client-side percentiles are being used, HdrHistogram is still needed at runtime. This pull request still declares HdrHistogram as needed at runtime.

PR #3285, opened 2022-07-13
Introduces GlobalObservationConvention

marcingrzejszczak:


micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -929,6 +929,16 @@ public List<KeyValuesProvider> getKeyValueProviders() {
 
     }
 
+    /**
+     * An observation convention that will be set on the {@link ObservationRegistry}.
+     *
+     * @author Marcin Grzejszczak
+     * @since 1.10.0
+     */
+    interface GlobalObservationConvention<T extends Context> extends ObservationConvention<T> {
```
jonatan-ivanov, line 938: Will this mean that users always need a subtype of `GlobalObservationConvention`?
Which would mean that they can't use an existing `ObservationConvention`.

If we want to lock this down/force the users to be explicit about this, I think it's ok, I just wanted to call out this consequence.

PR #3291, opened 2022-07-14
Added option to set a parent Observation on an existing Observation

marcingrzejszczak:
without this change in order to e.g. create spans we always need to put an observation in scope so that spans are put in thread local and a child span is being set. That doesn't make a lot of sense when we exactly know what the parent observation is.
with this change we're allowing to set a parent observation direcly on an existing observation. That way we're able to maintain a parent child relationship of observations and of e.g. spans

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
@@ -16,16 +16,12 @@
 package io.micrometer.observation;
 
 import io.micrometer.common.KeyValue;
-import io.micrometer.common.KeyValues;
-import io.micrometer.common.lang.NonNull;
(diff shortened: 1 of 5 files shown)


micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
```diff
@@ -210,6 +232,145 @@ public Observation getCurrentObservation() {
...
             this.registry.setCurrentObservationScope(previousObservationScope);
             this.currentObservation.notifyOnScopeClosed();
+            if (this.previousParentObservation != null) {
+                this.currentObservation.context.setParentObservation(this.previousParentObservation);
+            }
+        }
+
+    }
+
+    static class ImmutableContext extends Context {
```
shakuzen, line 242: This isn't used anywhere.

PR #3293, opened 2022-07-14
Receiver and SenderContext

marcingrzejszczak:
With this change we are spiking complete removal of any http / messaging / rpc abstractions and leaving a ReceiverContext and RequestReply or FireAndForget SenderContext. That way we are able to propagate any content over the wire and we are not forcing any default tagging mechanism (other than the one to be backward compatible).

It turned out that we're not properly supporting OkHttp instrumentation so I had to rewrite it to use the Interceptor approach not the EventListener one (which simplified a lot of things too).

Migration guide:
- If you were using `HttpRequest` & `HttpResponse` abstraction + the `HttpClientContext` and `HttpServerContext` you'll need to use the `RequestReplySenderContext` (for the client side) and `RequestReplyReceiverContext` (for the server side)

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpObservationInterceptor.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpObservationInterceptor.java
@@ -175,7 +175,6 @@ public class OkHttpObservationInterceptor implements Interceptor {
 
         /**
          * Add a {@link KeyValue} to any already configured tags on this Builder.
-         *
          * @param tag tag to add
          * @return this builder
          * @since 1.5.0
@@ -187,7 +186,6 @@ public class OkHttpObservationInterceptor implements Interceptor {
 
         /**
          * Add a context-specific tag.
-         *
          * @param contextSpecificTag function to create a context-specific tag
          * @return this builder
          * @since 1.5.0
@@ -208,7 +206,6 @@ public class OkHttpObservationInterceptor implements Interceptor {
          * other HTTP client metrics, this was changed to {@code target.host}, but to
          * maintain backwards compatibility the {@code host} tag can also be included. By
          * default, {@code includeHostTag} is {@literal true} so both tags are included.
-         *
          * @param includeHostTag whether to include the {@code host} tag
          * @return this builder
          * @since 1.5.0
@@ -224,7 +221,6 @@ public class OkHttpObservationInterceptor implements Interceptor {
          * These keys will be added with {@literal UNKNOWN} values when {@link Request} is
          * {@literal null}. Note that this is required only for Prometheus as it requires
          * tag match for the same metric.
-         *
          * @param requestTagKeys request tag keys
          * @return this builder
          * @since 1.3.9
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
cc @bclozel (that will be the context that you requested for - if you extend from it then propagation will happen out of the box)

micrometer-observation/src/main/java/io/micrometer/observation/transport/ReceiverContext.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public ReceiverContext(@NonNull Propagator.Getter<C> getter) {
+        this(getter, Kind.CONSUMER);
+    }
+
+    public C getCarrier() {
+        return carrier;
+    }
+
+    public void setCarrier(C carrier) {
```
jonatan-ivanov, line 60: Do we need this setter?
Also in `SenderContext`.

micrometer-observation/src/main/java/io/micrometer/observation/transport/RequestReplyReceiverContext.java
```diff
@@ -0,0 +1,59 @@
...
+    public RequestReplyReceiverContext(@NonNull Propagator.Getter<C> getter) {
+        this(getter, Kind.SERVER);
+    }
+
+    @Nullable
+    public RES getResponse() {
+        return response;
+    }
+
+    public void setResponse(RES response) {
```
jonatan-ivanov, line 55: Can we set this in the ctor and drop the setter?
Also in `RequestReplySenderContext`.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpContext.java
```diff
@@ -15,62 +15,72 @@
...
+    public OkHttpObservationInterceptor.CallState getState() {
         return state;
     }
 
     public Function<Request, String> getUrlMapper() {
         return urlMapper;
     }
 
-    public Iterable<Tag> getExtraTags() {
+    public Iterable<KeyValue> getExtraTags() {
```
jonatan-ivanov, line 75: Tag -> KeyValue (there are multiple places)

micrometer-observation/src/main/java/io/micrometer/observation/transport/ReceiverContext.java
marcingrzejszczak: Yes, cause we might first create a context and only later (if at all set the request). We already heard this feedback from @bclozel

micrometer-observation/src/main/java/io/micrometer/observation/transport/RequestReplyReceiverContext.java
marcingrzejszczak: No, cause we might first create a context and only later (if at all set the request and response). We already heard this feedback from @bclozel

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpContext.java
marcingrzejszczak: Yeah, anything wrong with that?

jonatan-ivanov: Should not it be `getExtraKeyValues` instead of `getExtraTags`?

marcingrzejszczak: I wanted to be consistent in terms of the names between metrics listener and this. If you think this should be changed please be my guest

PR #3301, opened 2022-07-19
Add getContextualName(Context c) on Observation.Convention

marcingrzejszczak:
without this change we can't provide a naming convention for objects that rely on contextual names e.g. all spans
with this change we can provide a standard naming convention for objects that rely on contextual names for e.g. spans


shakuzen (approved):
Makes sense to me.

PR #3305, opened 2022-07-20
Polishing Observation.Convention

jonatan-ivanov:


micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -1025,7 +1023,7 @@ interface KeyValuesConvention {
          * @return the new, contextual name for the observation
          */
         default String getContextualName(T context) {
-            return null;
+            return this.getName();
```
marcingrzejszczak, line 1026: Initially i thought that this should have such a value. I think i had my reasons why i set null here. Of course i don't remember those reasons anymore

PR #3312, opened 2022-07-25
Apache HTTP client instrumentation with Observation

shakuzen:
Adds Observation-based instrumentation for the Apache HTTP client while being backwards compatible. Configuring an ObservationRegistry (that is not a no-op) causes the `Observation` instrumentation to be used instead of instrumentation directly with `Timer`.


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/DefaultApacheHttpClientObservationConvention.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.httpcomponents;
+
+import io.micrometer.common.KeyValues;
+import org.apache.http.HttpRequest;
+
+import java.util.function.Function;
+
+public class DefaultApacheHttpClientObservationConvention implements ApacheHttpClientObservationConvention {
```
marcingrzejszczak, line 23: the `getContextualName()` should return sth like `HTTP Get` right? So `HTTP` word + HTTP Method of the given request?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/MicrometerHttpRequestExecutor.java
```diff
@@ -121,25 +130,96 @@ public HttpResponse execute(HttpRequest request, HttpClientConnection conn, Http
...
         }
         finally {
-            Iterable<Tag> tags = Tags.of(extraTags).and(routeTags).and(uri, method, status);
+            Tags routeTags = exportTagsForRoute ? HttpContextUtils.generateTagsForRoute(context) : Tags.empty();
+            Tags otherTags = Tags.of("method", request.getRequestLine().getMethod(), "uri", uriMapper.apply(request),
+                    "status", status.getValue());
 
-            timerSample.stop(Timer.builder(METER_NAME).description("Duration of Apache HttpClient request execution")
-                    .tags(tags).register(registry));
+            sample.stop(otherTags.and(routeTags).and(extraTags));
```
marcingrzejszczak, line 137: So if we're using the Observation API we don't want to set these tags? I'm just wondering if it should matter - shouldn't we just set all the tags upfront?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/MicrometerHttpRequestExecutor.java
```diff
@@ -195,12 +282,40 @@ public Builder exportTagsForRoute(boolean exportTagsForRoute) {
...
+        }
+
         /**
          * @return Creates an instance of {@link MicrometerHttpRequestExecutor} with all
          * the configured properties.
          */
         public MicrometerHttpRequestExecutor build() {
-            return new MicrometerHttpRequestExecutor(waitForContinue, registry, uriMapper, tags, exportTagsForRoute);
+            if (observationConvention == null) {
+                observationConvention = new DefaultApacheHttpClientObservationConvention(uriMapper, exportTagsForRoute);
```
shakuzen, line 315: I guess it will be needed to move `uriMapper` and `exportTagsForRoute` to the context. As it is, a user-provided convention would not have access to these.

marcingrzejszczak, line 315: If you initialize this at this point instead of passing this as potential `null` to the `Observation.start()` method together with a default there AFAIR you're bypassing going through registered `GlobalObservationConvention` implementations on `ObservationConfig` (check https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java#L127-L138)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/MicrometerHttpRequestExecutor.java
shakuzen: If we're using the Observation API, those tags are set by the default convention. I don't like the way this currently works because the Tag(s) objects are made even if we aren't using them.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/DefaultApacheHttpClientObservationConvention.java
shakuzen: Right. Added that now.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/ApacheHttpClientContext.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public HttpContext getApacheHttpContext() {
+        return apacheHttpContext;
+    }
+
+    public Function<HttpRequest, String> getUriMapper() {
+        return uriMapper;
+    }
+
+    public boolean exportTagsForRoute() {
```
marcingrzejszczak, line 60: `isExportTagsForRoute` ?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/DefaultApacheHttpClientObservationConvention.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return MicrometerHttpRequestExecutor.METER_NAME;
+    }
+
+    @Override
+    public String getContextualName(ApacheHttpClientContext context) {
+        return "HTTP " + getMethodString(context.getCarrier());
```
marcingrzejszczak, line 48: I've seen this hard coded quite a few times - I wonder if we can extract this somehow, somewhere? :shrug:

marcingrzejszczak (approved):

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/httpcomponents/MicrometerHttpRequestExecutorTest.java
```diff
@@ -192,12 +213,120 @@ void overrideExtraTagsDoesNotThrowAnException(@WiremockResolver.Wiremock WireMoc
...
+        }
+
+        @Override
+        public String getMethod() {
+            return method;
+        }
+
+    }
+
+    // TODO add test for status = IO_ERROR case.
```
shakuzen, line 298: I wanted to add a test for this, but I wasn't sure how. If someone knows how to test this scenario with the Apache HTTP client, let me know.  The scenarios are when the response returned from `execute` is `null` and when calling `execute` results in an `IOException`

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/DefaultApacheHttpClientObservationConvention.java
shakuzen: Since how to get the method will be different for each implementation I'm not sure we really can. This is exactly what we were trying to do with the HTTP abstractions for request/response, but we ended up going a different direction.

PR #3323, opened 2022-07-29
Scoped checked actions

jonatan-ivanov:


micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationRegistryCompatibilityKit.java
```diff
@@ -298,21 +298,47 @@ void runnableShouldBeScoped() {
...
+    }
+
+    @Test
+    void checkedRunnableShouldBeScoped() throws Throwable {
+        registry.observationConfig().observationHandler(c -> true);
+        Observation observation = Observation.start("myObservation", registry);
+        Observation.CheckedRunnable runnable = () -> assertThat(registry.getCurrentObservation()).isSameAs(observation);
+        observation.scopedChecked(runnable);
+        assertThat(registry.getCurrentObservation()).isNull();
+        assertThat(observation.getContext().getError()).isEmpty();
```
shakuzen, line 326: I think it would be good to assert that the observation is not stopped by using `scopedChecked`.

shakuzen (approved):

PR #3326, opened 2022-08-02
Generify throwable in CheckedCallable/CheckedRunnable

shakuzen:
This allows usage that catches or throws more a more specific type than `Throwable`. This is a bit more verbose when declaring a variable of type `CheckedCallable` or `CheckedRunnable`, but I expect most usage to be as a lambda or method reference, which avoids the need to do such.

@jonatan-ivanov let me know what you think. I may have overlooked some reason why not to do this. This change wasn't based on user feedback, so I don't feel strongly about doing this but I figured I would open the proposal for discussion.

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationRegistryCompatibilityKit.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationRegistryCompatibilityKit.java
@@ -149,7 +149,8 @@ public abstract class ObservationRegistryCompatibilityKit {
         registry.observationConfig().observationHandler(handler);
         Observation observation = Observation.createNotStarted("myObservation", registry);
 
-        Observation.CheckedRunnable<Throwable> checkedRunnable = () -> assertThat(registry.getCurrentObservation()).isSameAs(observation);
+        Observation.CheckedRunnable<Throwable> checkedRunnable = () -> assertThat(registry.getCurrentObservation())
+                .isSameAs(observation);
         observation.observeChecked(checkedRunnable);
         assertThat(registry.getCurrentObservation()).isNull();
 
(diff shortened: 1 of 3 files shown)


micrometer-observation/src/test/java/io/micrometer/observation/ObservationTests.java
```diff
@@ -115,4 +113,33 @@ void settingScopeMakesAReferenceOnParentContext() {
...
 
+    @Test
+    void scopedCheckedSpecificException() {
+        Service service = new Service();
+        Observation observation = Observation.start("service", registry);
+        try {
+            String s = observation.scopedChecked(service::executeCallable);
+            observation.scopedChecked(service::executeRunnable);
+        }
+        catch (IOException ignore) {
```
shakuzen, line 124: This wasn't possible before due to it being hard-coded to `Throwable`

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationRegistryCompatibilityKit.java
```diff
@@ -149,9 +149,7 @@ void checkedRunnableShouldBeObserved() throws Throwable {
         registry.observationConfig().observationHandler(handler);
         Observation observation = Observation.createNotStarted("myObservation", registry);
 
-        Observation.CheckedRunnable checkedRunnable = () -> assertThat(registry.getCurrentObservation())
```
jonatan-ivanov, line 152: These variables were there explicitly because this way we won't test a separate scenario (overloaded method) accidentally and it might help to the reader of the test too.

jonatan-ivanov (approved):

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationRegistryCompatibilityKit.java
shakuzen: Thanks. This was actually calling the wrong overload. I've reverted that part of the change now.

shakuzen:
This is a breaking change if someone is declaring a `CheckedCallable` variable somewhere like the following because `CheckedCallable` now requires two generic arguments.

```java
Observation.CheckedCallable<String> callable = () -> {
            assertThat(registry.getCurrentObservation()).isNull();
            return "test";
        };
```

CheckedRunnable will still compile because it had 0 before.
I am hoping/expecting that the impact of this breaking change will be minimal since I expect explicit declarations of `CheckedCallable` to be rare (if exist at all).

Issue #3181, opened 2022-05-16
Ignore METHOD_NEW_DEFAULT binary compatibility


shakuzen:
We check binary compatibility in the build against the previous version to catch incompatibilities introduced. This catches new default methods on interfaces, which we don't want blocking the build, since we expect this to be compatible. This came up in https://github.com/micrometer-metrics/micrometer/pull/3175#issuecomment-1125972708. It was brought up in the `japicmp` repo in https://github.com/siom79/japicmp/issues/201. However, it looks like we're missing a straightforward way to configure accepting `METHOD_NEW_DEFAULT` changes in the Gradle plugin. It has been requested in https://github.com/melix/japicmp-gradle-plugin/issues/43. We may have to do a temporary workaround until that is resolved.

shakuzen:
We can now do this more easily thanks to https://github.com/melix/japicmp-gradle-plugin/issues/43 being implemented. I've also updated it to target 1.8.x since japicmp was backported there since this issue was opened.

PR #3329, opened 2022-08-03
Datadog Dogstatsd support.

platinummonkey:
ref https://github.com/micrometer-metrics/micrometer-docs/pull/225 discussion

Make the smallest compatible change for users to migrate to dogstatsd approach. This still defaults to the prior API method currently. Many of our users already have the datadog-agent already installed collecting traces, logs and profiling information. This result in a positive performance and reliability benefit as your datadog-agent requires less network hops, and is configured to automatically handle edge related to retries and failure scenarios.  

This allows them to update their configuration as follows:

Instead of 
```
datadog.uri = https://api.datadoghq.com
datadog.apiKey = secret
```

For autodiscovery these environment variables should exist: https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/NonBlockingStatsDClient.java#L53-L58
```
# note the 3rd `/` to satisfy java.net.URI
datadog.uri = discovery:///
```

For specifying exactly (example named pipe):
```
datadog.uri = unix:///var/run/datadog.sock
```

another example for udp to the listening port
```
datadog.uri = udp://localhost:8125/
```


implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
```diff
@@ -59,9 +61,42 @@ default String hostTag() {
...
+     * Using a statsdURL will be preferred here over api endpoints and will report via the
+     * agent. Use tcp/udp/unix URI schemes to indicate this.
      */
     default String uri() {
-        return getUrlString(this, "uri").orElse("https://api.datadoghq.com");
+        return getUriString(this, "uri").orElse("https://api.datadoghq.com");
+    }
+
+    /**
+     * @return Returns true if this is a non https scheme target, to indicate to talk via
```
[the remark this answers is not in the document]
platinummonkey: @sonatype-lift ignore

implementations/micrometer-registry-datadog/build.gradle
```diff
@@ -2,6 +2,7 @@ dependencies {
     api project(':micrometer-core')
 
     implementation 'org.slf4j:slf4j-api'
+    implementation 'com.datadoghq:java-dogstatsd-client'
```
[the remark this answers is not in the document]
platinummonkey: I've raised this to our team internally to evaluate the relevancy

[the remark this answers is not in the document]
platinummonkey: response that we're tracking internally, marking this as won't fix

> java-dogstatsd-client depends on jnr-unixsocket, which depends on jnr-posix.
> 
> Because we support Java 7+, we depend on an old version of jnr-unixsocket, 0.36.
> 
> This version depends on jnr-posix which[ has a vulnerability](https://security.snyk.io/vuln/SNYK-JAVA-COMGITHUBJNR-1570422) in a function that is not used by jnr-unixsocket and java-dogstatsd-client, so does NOT directly affect the client library.
> 
> The issue will be resolved for good when we drop Java 7 support and can upgrade dependency version.

platinummonkey:
bump - any takers for review?

shakuzen (changes requested):
Thank you for the thorough pull request. I am a bit weary about adding yet another way to ship metrics to Datadog with Micrometer (with this, there are now three options), but I think it's reasonable enough to have an option to delegate to the DogStatsD client in our datadog module, given some of the benefits mentioned in https://github.com/micrometer-metrics/micrometer-docs/pull/225 such as client-side aggregation - our statsd implementation buffers lines and sends a batch but it does not aggregate them before sending.

I left a couple quick review comments, but overall this looks good. I will have to take a closer look and give it a spin, but we should probably be able to review and get something merged for our next milestone release in mid September. Thanks again for the effort here.

AhmedCh:
@shakuzen will this make it into an upcoming release? Or is there some more work left on it? 

I am looking forward to this landing as my team and I are having some issues using DataDog via Micrometer in our Spring Boot application, we don't see the same issue when using [java-dogstatsd-client](https://github.com/DataDog/java-dogstatsd-client) and I would very much like us to keep using Micrometer with Spring Boot. I am thinking might just solve our problem.

The issue is that when we use Micrometer, we observe data loss. When we compare events as recorded in Splunk (our logging tool), versus Datadog, we see 2 or 4 times more events in Splunk than Datadog. And the loss factor correlates to the size of our ECS cluster. When we had our app on 4 nodes in our ECS cluster, we had 4x data loss. After scaling down to 2 nodes, we got 2x data loss. We don't see this issue when using  java-dogstatsd-client. The issue might be something on our side, like bad config when using Micrormeter but so far we haven't found anything.

cc @platinummonkey

shakuzen:
> will this make it into an upcoming release? Or is there some more work left on it?

I'm particularly hoping to get feedback on [this review comment](https://github.com/micrometer-metrics/micrometer/pull/3329#discussion_r942267577). Once that is addressed, I will try to make time to try this out (and you or other Datadog users could also try it, which would be greatly appreciated). For inclusion in our upcoming 1.10 release, we are planning a release candidate on October 10. New features would need to be merged before then. Otherwise, it will have to wait until a future release.

> The issue is that when we use Micrometer, we observe data loss.

I'm sorry that's happening. Feel free to open a separate issue regarding it.

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -81,14 +87,75 @@ public DatadogMeterRegistry(DatadogConfig config, Clock clock, ThreadFactory thr
...
+                queryParams = queryParameters(statsdURI);
+            }
+            catch (UnsupportedEncodingException e) {
+                throw new RuntimeException(e);
+            }
+
+            if (truthy(queryParams.getOrDefault("enableAggregation", "false"))) {
+                int aggregationFlushInterval = Integer
+                        .parseInt(queryParams.getOrDefault("aggregationFlushInterval", "300"));
+                int aggregationShards = Integer.parseInt(queryParams.getOrDefault("aggregationShards", "4"));
```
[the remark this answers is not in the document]
platinummonkey: I originally did this but rebased after failing the JAPI compatibility check in CI. I can go back to that but the checks will fail - it wasn’t obvious to me if this was a hard or soft requirement check

platinummonkey:
Our original customer had moved off micrometer prior to this landing, but I can certainly come back and help finish this out as it’s a popular library.

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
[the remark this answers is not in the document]
shakuzen: Ah, sorry about that. It is indeed not obvious when getting that failure. We're hoping to improve the experience for contributors by not failing the compatibility check on new default methods with https://github.com/micrometer-metrics/micrometer/issues/3181, which was hard to implement until recently due to limitations in the plugin we're using.

AhmedCh:
> Our original customer had moved off micrometer prior to this landing, but I can certainly come back and help finish this out as it’s a popular library.

Thanks @platinummonkey, that would be much appreciated!

AhmedCh:
> I'm sorry that's happening. Feel free to open a separate issue regarding it.

Thanks, sounds good. We do have a support case open with folks at DataDog and they helping us look into this issue but since this isn't their officially supported mechanism, they might not be able to get to the bottom of this, we are also looking at things on our end. If we don't get a resolution from either approaches, I might just open an issue here.

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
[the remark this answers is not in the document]
platinummonkey: Ahhh okay I’ll go back to that then

implementations/micrometer-registry-datadog/build.gradle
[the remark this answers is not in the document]
platinummonkey: resolved previously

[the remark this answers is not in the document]
platinummonkey: @sonatype-lift ignore

platinummonkey:
the remainder of these `sonatype-lift` errors look related to what already exists and outside the scope of this PR 🙃

platinummonkey:
I merged main and now the tests fail 🥇

implementations/micrometer-registry-datadog/build.gradle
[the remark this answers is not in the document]
platinummonkey: @sonatype-lift ignoreall

shakuzen:
> I merged main and now the tests fail

Sorry about that, `main` is now fixed. Could you rebase?

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
```diff
@@ -56,12 +58,56 @@ default String hostTag() {
...
+    default boolean isStatsd() {
+        String scheme = URI.create(uri()).getScheme();
+        return !(scheme.equalsIgnoreCase("https") || scheme.equalsIgnoreCase("http"));
+    }
+
+    /**
+     * Turn this on to enable client-side (in process) aggregation for dogstatsd mode.
+     * @return Returns true if aggregation is enabled.
+     */
+    default boolean enableAggregation() {
```
shakuzen, line 75: Adding these methods to `DatadogConfig` is one approach that could work, but I'm concerned about the complexity for users of having some methods that only do anything when running in one mode but not the other. It's not unprecedented in Micrometer, but it's nice to keep things as simple as we can.

Do you have an idea how common a need it is for users to customize these aggregation configuration properties?

If the defaults are acceptable for most users, I think we could consider ways to allow customization without adding methods to `DatadogConfig`. One idea I mentioned in https://github.com/micrometer-metrics/micrometer/pull/3329#discussion_r942267577 was allowing users to provide their own `StatsDClient` to the `DatadogMeterRegistry.Builder` if they want to have the ability to fully customize. We could still allow users to get other config from the `DatadogConfig` with a method that returns a `NonBlockingStatsDClientBuilder` with e.g. the host set. Does that make sense? What do you think?

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
```diff
@@ -56,12 +58,56 @@ default String hostTag() {
         return getString(this, "hostTag").orElse("instance");
     }
 
+    /**
+     * Truthy if the configuration is defined for a dogstatsd reporting configuration.
+     * @return Returns true if this is a non https scheme target, to indicate to talk via
+     * dogstatsd.
+     */
+    default boolean isStatsd() {
```
shakuzen, line 66: I'm not sure when a user would override the value for this. Is this method just for our convenience? If so, it should probably be a private method in `DatadogMeterRegistry` instead so it is not exposed to users.

implementations/micrometer-registry-datadog/build.gradle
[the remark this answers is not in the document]
shakuzen: @sonatype-lift ignore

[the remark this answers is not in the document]
platinummonkey: @sonatype-lift ignore

platinummonkey:
currently seem blocked on 

```
./gradlew resolveAndLockAll --no-build-cache --write-locks

...
* What went wrong:
Execution failed for task ':micrometer-registry-cloudwatch2:resolveAndLockAll'.
> Could not resolve all files for configuration ':micrometer-registry-cloudwatch2:compileClasspath'.
   > Could not find software.amazon.awssdk:cloudwatch:.
     Required by:
         project :micrometer-registry-cloudwatch2
   > Could not resolve software.amazon.awssdk:cloudwatch:latest.release.
     Required by:
         project :micrometer-registry-cloudwatch2
         project :micrometer-registry-cloudwatch2 > project :micrometer-core
         project :micrometer-registry-cloudwatch2 > project :micrometer-core > project :micrometer-commons
         project :micrometer-registry-cloudwatch2 > project :micrometer-core > project :micrometer-observation
      > Could not resolve software.amazon.awssdk:cloudwatch:2.17.293.
         > Could not parse POM https://repo.maven.apache.org/maven2/software/amazon/awssdk/cloudwatch/2.17.293/cloudwatch-2.17.293.pom
            > Could not find software.amazon.awssdk:services:2.17.293.
              Searched in the following locations:
                - https://repo.maven.apache.org/maven2/software/amazon/awssdk/services/2.17.293/services-2.17.293.pom
              If the artifact you are trying to retrieve can be found in the repository but without metadata in 'Maven POM' format, you need to adjust the 'metadataSources { ... }' of the repository declaration.
      > Could not resolve software.amazon.awssdk:cloudwatch:2.17.293.
         > Could not parse POM https://repo.maven.apache.org/maven2/software/amazon/awssdk/cloudwatch/2.17.293/cloudwatch-2.17.293.pom
            > Could not find software.amazon.awssdk:services:2.17.293.
...

```

platinummonkey:
can we ignore the `sonatype-lift` here? 

1 of the failures is a nag.
2 of them actually cause compile failures and are blocking.

platinummonkey:
none of the remaining sonatype failures are related to this branch, prexisting failures on `main`

platinummonkey:
🦗 bump?

platinummonkey:
Bump

platinummonkey:
Bump

platinummonkey:
@shakuzen @AhmedCh bump?

platinummonkey:
appears another integration is failing as a result of merging `main` 

```
 What went wrong:
Execution failed for task ':micrometer-registry-wavefront:test'.
```

platinummonkey:
Bump

robert-csdisco:
@platinummonkey Hoping you get this bumped. :)   

The current statsd thing sends a line for every counter increase, which makes the metrics scale w/ the thing they're counting, which can be no bueno!

Hope this gets merged in soon!

platinummonkey:
Bump

robert-csdisco:
> Bump

Heads up @platinummonkey, the PR now looks like it acquired some merge conflicts `This branch has conflicts that must be resolved` Just in case they see your bump soon. :)

platinummonkey:
> Heads up @platinummonkey, the PR now looks like it acquired some merge conflicts `This branch has conflicts that must be resolved` Just in case they see your bump soon. :)

this is the theme yes, when a PR sits for over a year 🙃

platinummonkey:
the `.github/workflows/deploy-docs.yml` workflow also seems busted now :) irrelevant to this PR

platinummonkey:
Bump

platinummonkey:
Bump

platinummonkey:
IMG_7084

bump

platinummonkey:
IMG_7087

Bump

platinummonkey:
Bump

platinummonkey:
Bump

platinummonkey:
@shakuzen looks like you’re approving other PRs - is there anything blocking this one?

platinummonkey:
bump

platinummonkey:
bump

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -78,17 +82,62 @@ public DatadogMeterRegistry(DatadogConfig config, Clock clock) {
...
+    }
+
+    private DatadogMeterRegistry(DatadogConfig config, Clock clock, ThreadFactory threadFactory, HttpSender httpClient,
+            @Nullable StatsDClient statsdClient) {
         super(config, clock);
 
         config().namingConvention(new DatadogNamingConvention());
 
+        if (statsdClient == null && isStatsd(config.uri())) {
+            NonBlockingStatsDClientBuilder builder = new NonBlockingStatsDClientBuilder();
```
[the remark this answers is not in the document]
marcingrzejszczak: @platinummonkey what about this comment?

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -78,17 +82,63 @@ public DatadogMeterRegistry(DatadogConfig config, Clock clock) {
...
 
-    private DatadogMeterRegistry(DatadogConfig config, Clock clock, ThreadFactory threadFactory,
-            HttpSender httpClient) {
+    /**
+     * Truthy if the uri is defined for a dogstatsd reporting configuration.
+     * @return Returns true if this is a non https scheme target, to indicate to talk via
+     * dogstatsd.
+     */
+    static boolean isStatsd(String uri) {
+        String scheme = URI.create(uri).getScheme();
```
marcingrzejszczak, line 109: Are we ok if the uri creation throws an exception (when the uri is invalid) ? I would presume that that's perfectly fine

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -78,17 +82,63 @@ public DatadogMeterRegistry(DatadogConfig config, Clock clock) {
...
+            @Nullable StatsDClient statsDClient) {
         super(config, clock);
 
         config().namingConvention(new DatadogNamingConvention());
 
+        if (statsDClient == null && isStatsd(config.uri())) {
+            NonBlockingStatsDClientBuilder builder = new NonBlockingStatsDClientBuilder();
+            URI statsdURI = URI.create(config.uri());
+            if (statsdURI.getScheme().equalsIgnoreCase("tcp") || statsdURI.getScheme().equalsIgnoreCase("udp")) {
+                if (!statsdURI.getHost().equalsIgnoreCase("")) {
```
marcingrzejszczak, line 124: We could theoretically flip the equality - in case any element here is `null` (although that shouldn't really happen)

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -230,6 +344,50 @@ private void addToMetadataList(Map<String, DatadogMetricMetadata> metadata, Mete
...
 
+    // VisibleForTesting
+    void writeMetricViaStatsd(Meter.Id id, @Nullable String suffix, double value, Statistic statistic,
+            @Nullable String overrideBaseUnit) {
+        if (statsDClient == null) {
+            return;
+        }
+
+        Meter.Id fullId = id;
+        if (suffix != null)
```
marcingrzejszczak, line 355: can we do the if with braces please?

```java
if (...) {

}
```

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -230,6 +344,50 @@ private void addToMetadataList(Map<String, DatadogMetricMetadata> metadata, Mete
...
+            fullId = idWithSuffix(id, suffix);
+
+        Iterable<Tag> tags = getConventionTags(fullId);
+
+        List<String> tagsList = tags.iterator().hasNext() ? stream(tags.spliterator(), false)
+            .map(t -> "\"" + escapeJson(t.getKey()) + ":" + escapeJson(t.getValue()) + "\"")
+            .collect(toList()) : new ArrayList<String>();
+        if (config.hostTag() != null && !Objects.equals(config.hostTag(), "")) {
+            tagsList.add("host:" + config.hostTag());
```
marcingrzejszczak, line 364: Pardon my ignorance, I'm no expert as fart as this `MeterRegistry` is concerned. Why are the previous tags wrapped in json (`"escapedJsonKey:escapedJsonValue"`) and `host` is not? Shouldn't it be `"host:..."`?

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMetricMetadata.java
```diff
@@ -109,4 +109,18 @@ static String sanitizeType(Statistic statistic) {
         }
     }
 
+    static String sanitizeStatsdType(Statistic statistic) {
```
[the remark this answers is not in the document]
marcingrzejszczak: What about this comment? @platinummonkey

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -230,6 +344,50 @@ private void addToMetadataList(Map<String, DatadogMetricMetadata> metadata, Mete
...
+
+        // Create type attribute
+        switch (fullId.getType()) {
+            case COUNTER:
+                statsDClient.count(metricName, value, 1.0, tagsArray);
+                break;
+            case GAUGE:
+                statsDClient.gauge(metricName, value, 1.0, tagsArray);
+                break;
+            case LONG_TASK_TIMER:
```
marcingrzejszczak, line 379: These 3 cases are not supported in DataDog?

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/utils/SampleRegistries.java
```diff
@@ -169,6 +169,18 @@ public Duration step() {
             public String get(String k) {
                 return null;
             }
+
+            public boolean enableAggregation() {
```
marcingrzejszczak, line 173: Why is this change required?

marcingrzejszczak (commented):
Hey, I'm sorry that it's taking so long. I'm mostly focusing on the tracing part of the micrometer observability but I went through the PR so I apologize if there are any rudimentary questions.

I do understand correctly that the PR is about introducing a new way of using the `StatsDClient` in order to ship metrics to Datadog? And that way various protocols are supported `tcp`, `udp` and `unix` and default `https`?

I also have one more question, out of curiosity, why can't we work on pushing this registry out to Datadog repository (e.g. https://github.com/DataDog/java-dogstatsd-client)? That way whenever there are any changes in the Datadog API / ways of shipping metrics then nobody would have to wait for us to do any reviews. Our API is very stable as far as MeterRegistries are concerned.

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
platinummonkey: yeah, this was intentional, a bad URI could potentially drop data. The same is true of http(s) if the wrong datacenter or path is used, this would provide faster feedback if something really bad happened

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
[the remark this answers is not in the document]
platinummonkey: can definitely add

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
platinummonkey: host tag is special here as it can be set directly in the DatadogConfig. This seems to be because of the the previous direct-to-api method used prior. Normally dogstatsd agent will auto-collect the hostname(s) [cloud-environment aware] and auto-report this, but for backwards compatibility I included this specifically back in to make the migration easier if this config value is set.

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
platinummonkey: this is a fall-through clause to `distribution` type. Timer's are typically reported as distributions so you can get min/max/avg/p##..

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMetricMetadata.java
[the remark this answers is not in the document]
platinummonkey: yup not needed anymore, removing 👍

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
marcingrzejszczak: OK I see, thank you

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
marcingrzejszczak: I see, thanks

platinummonkey:
> Hey, I'm sorry that it's taking so long. I'm mostly focusing on the tracing part of the micrometer observability but I went through the PR so I apologize if there are any rudimentary questions.

Thanks for reviewing!

> I do understand correctly that the PR is about introducing a new way of using the `StatsDClient` in order to ship metrics to Datadog? And that way various protocols are supported `tcp`, `udp` and `unix` and default `https`?

A tad different - about correcting the "datadog" plugin to offer a dogstatsd method to report. In the end we should eventually remove the direct-to-api method entirely, but that would be an abrupt change without a good migration story for folks to do it in increments if they have a large number of services and a shared dependency list. This adds the ability in the existing datadog registry plugin to use dogstatsd's various methods and while still shoe-horning it into existing config. In theory registry-statsd could be updated to do non-blocking per-line publishing, but even customers that have tried this reported noticeable performance degradations and a lack of metric types (distributions for example). 

> I also have one more question, out of curiosity, why can't we work on pushing this registry out to Datadog repository (e.g. https://github.com/DataDog/java-dogstatsd-client)? That way whenever there are any changes in the Datadog API / ways of shipping metrics then nobody would have to wait for us to do any reviews. Our API is very stable as far as MeterRegistries are concerned.

Do you mean the specific `micrometer-registry-datadog` module in this repository? This might be simpler indeed.

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/utils/SampleRegistries.java
platinummonkey: I don't believe this is anymore 👍 removing

marcingrzejszczak:
> Do you mean the specific micrometer-registry-datadog module in this repository? This might be simpler indeed.

Yes. If that's possible I would love to push this code to that repo. Then you can set it up in whatever way you want and nothing will be blocking you really. Are you willing to accept such a change? What we could do then is deprecate our registry in favour of yours. We could merge this PR so that there are 2 options and then in the DataDog repository there would be the "new, proper" one. WDYT? Also I'm not saying that it would have to be that concrete repository, I'm not familiar with DataDogs repos, it was just an idea to discuss.

platinummonkey:
> Yes. If that's possible I would love to push this code to that repo. Then you can set it up in whatever way you want and nothing will be blocking you really. Are you willing to accept such a change? What we could do then is deprecate our registry in favour of yours. We could merge this PR so that there are 2 options and then in the DataDog repository there would be the "new, proper" one. WDYT? Also I'm not saying that it would have to be that concrete repository, I'm not familiar with DataDogs repos, it was just an idea to discuss.

It's very interesting, I've found a couple examples in the wild. Following up internally about this option.

marcingrzejszczak:
@platinummonkey how can I test this PR against Datadog? Could you please help me out with this?

platinummonkey:
> @platinummonkey how can I test this PR against Datadog? Could you please help me out with this?

For a full e2e test, you could sign up for a free trial account of DataDog. The setup will walk you through installing and setting up a DataDog agent. Then set the config to 

`datadog.uri = udp://localhost:8125/`

with a basic app to report some metrics. Then you could navigate to the metrics explorer view to see the metrics reported. 

If it’s easier I’m happy to take some screenshots and post a gist of a sample app. 



Regarding hosting our own registry repository - still following up but gaining popularity.

marcingrzejszczak:
Thanks for the description

> If it’s easier I’m happy to take some screenshots and post a gist of a sample app.

In the meantime please be my guest to do that :) It will definitely speed things up

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -221,21 +352,68 @@ private void addToMetadataList(Map<String, DatadogMetricMetadata> metadata, Mete
...
+            fullId = idWithSuffix(id, suffix);
+        }
+
+        Iterable<Tag> tags = getConventionTags(fullId);
+
+        List<String> tagsList = tags.iterator().hasNext() ? stream(tags.spliterator(), false)
+            .map(t -> "\"" + escapeJson(t.getKey()) + ":" + escapeJson(t.getValue()) + "\"")
+            .collect(toList()) : new ArrayList<String>();
+        if (config.hostTag() != null && !Objects.equals(config.hostTag(), "")) {
+            tagsList.add("host:" + config.hostTag());
```
shakuzen, line 383: The value of `config.hostTag()` is the name of the Micrometer tag to get the Datadog `host` tag value from. See https://github.com/micrometer-metrics/micrometer/blob/3d7a3901c68df189f185a931cee89f33c188d461/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java#L243-L247

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
```diff
@@ -72,10 +80,73 @@ default boolean descriptions() {
...
+        return getInteger(this, "maxPacketSizeBytes").orElse(-1);
+    }
+
+    /**
+     * @return {@code false} if client-side aggregation should be enabled when in
+     * Dogstatsd mode.
+     * https://docs.datadoghq.com/developers/dogstatsd/high_throughput/?code-lang=java#client-side-aggregation
+     */
+    default boolean enableAggregation() {
+        return getBoolean(this, "enableAggregation").orElse(false);
```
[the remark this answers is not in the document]
shakuzen: Actually, after looking closer, I think `false` may be the better default, since Micrometer is already aggregating and reporting once per step to DogStatsD.

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
```diff
@@ -72,10 +80,73 @@ default boolean descriptions() {
...
+    }
+
+    /**
+     * @return {@code 4} determines the number of shards in the aggregator, this feature
+     * is aimed at mitigating the effects of map locking in highly concurrent scenarios
+     * when in Dogstatsd mode.
+     * https://github.com/DataDog/java-dogstatsd-client#configuration-1
+     */
+    default Integer aggregationShards() {
+        return getInteger(this, "aggregationShards").orElse(4);
```
shakuzen, line 126: I wonder if it makes sense to expose this given the setup of Micrometer doing the aggregating and the DogStatsD client sending metrics once per interval. With that, I doubt anyone would need to configure this.

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -188,6 +290,26 @@ private Stream<String> writeTimer(Timer timer, Map<String, DatadogMetricMetadata
...
     }
 
+    private Integer writeTimerViaStatsd(Timer timer) {
+        Meter.Id id = timer.getId();
+
+        // we can't know anything about max and percentiles originating from a function
+        // timer
+        writeMetricViaStatsd(id, "count", (double) timer.count(), Statistic.COUNT, "occurrence");
+        writeMetricViaStatsd(id, "avg", timer.mean(getBaseTimeUnit()), Statistic.VALUE, null);
+        writeMetricViaStatsd(id, "sum", timer.totalTime(getBaseTimeUnit()), Statistic.TOTAL_TIME, null);
```
shakuzen, line 300: This ends up sending 3 distribution type metrics to Datadog. I'm not sure that's what we want, is it? We are aggregating Micrometer Timer recordings in Micrometer and then publishing the aggregate values once per step interval (by default, once per minute) via DogStatsD. A distribution of the 1 minute count, 1 minute sum, 1 minute average are probably not what users want. I don't think distribution type metrics will work as expected in Datadog unless all recorded values are processed by Datadog.

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -188,6 +290,26 @@ private Stream<String> writeTimer(Timer timer, Map<String, DatadogMetricMetadata
         return metrics.build();
     }
 
+    private Integer writeTimerViaStatsd(Timer timer) {
+        Meter.Id id = timer.getId();
+
+        // we can't know anything about max and percentiles originating from a function
+        // timer
```
shakuzen, line 297: Unlike `FunctionTimer`, a regular `Timer` does have a `max()` method.

platinummonkey:
> Thanks for the description
> 
> > If it’s easier I’m happy to take some screenshots and post a gist of a sample app.
> 
> In the meantime please be my guest to do that :) It will definitely speed things up

Alrighty, we've gone down the path of [creating a managed repo](https://github.com/DataDog/micrometer-registry-datadog). A sample output from a small app:

image


 We've got some things to do here in the interim, and can come back with a PR to deprecate this plugin instead.  If that sounds good, let's close this PR?

platinummonkey:
👋 alrighty, followed this through at Datadog. We've published a guide after also reviewing other plugins offered by Micrometer. 

https://docs.datadoghq.com/metrics/guide/micrometer

I'm closing this PR, and will open a new PR to mark this Datadog plugin as deprecated, then will see y'all in a few months to remove the plugin entirely. 

Thank you for the feedback and reviews! 🙇

PR #3332, opened 2022-08-05
More accurate metrics for HTTP redirect responses from Jersey

### LINKED ISSUE #3327: Uri tag replaced with REDIRECTION on all HTTP redirect responses with Jersey server
It would be nice to be able to choose if you want the URI tag to be replaced with REDIRECTION or not. In many cases it is interesting to get metrics on redirects. "304 Not modified" for example is interesting to track, same for "303 See other", how long those the server actually use to fulfil the request.

This line is the reason for the uri tag being set to REDIRECTION: https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/JerseyTags.java#L91.

**Rationale**
Uri tag replacement for redirect responses should be configureable. In many cases it is interesting to have metrics on redirections.

tinolazreg:
Fixes #3327

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/JerseyTags.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/JerseyTags.java
@@ -87,7 +87,7 @@ public final class JerseyTags {
         ContainerResponse response = event.getContainerResponse();
         if (response != null) {
             int status = response.getStatus();
-            if (isRedirection(status)) {
+            if (isRedirection(status) && event.getUriInfo().getMatchedResourceMethod() == null) {
                 return URI_REDIRECTION;
             }
             if (status == 404 && event.getUriInfo().getMatchedResourceMethod() == null) {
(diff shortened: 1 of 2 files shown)


shakuzen (approved):
Thanks! It looks good to me. Could we add a test for the case when there is no matched route?
Also, we will need the CLA signed to accept the contribution.

tinolazreg:
> Thanks! It looks good to me. Could we add a test for the case when there is no matched route? Also, we will need the CLA signed to accept the contribution.

Great! I have signed the CLA now. I tried creating a case for the test that you mentioned, but found it difficult to create a similar test with a HTTP redirection without a matching route, that ends up in the JerseyTags.

shakuzen:
> I tried creating a case for the test that you mentioned, but found it difficult to create a similar test with a HTTP redirection without a matching route, that ends up in the JerseyTags.

Thank you for trying. I tried a bit myself and also struggled to come up with something. I don't like putting something in that isn't tested, but the logic makes sense and is similar to instrumentation we have elsewhere.

I've rebased the pull request on the `1.8.x` branch so we can get this fixed in upcoming maintenance releases. The more I've thought about it, the more the previous behavior was more bug-like. Once merged, snapshots will be available if you would like to test things out with your application before the release.

Thank you for reporting and fixing this.

tinolazreg:
> > I tried creating a case for the test that you mentioned, but found it difficult to create a similar test with a HTTP redirection without a matching route, that ends up in the JerseyTags.
> 
> Thank you for trying. I tried a bit myself and also struggled to come up with something. I don't like putting something in that isn't tested, but the logic makes sense and is similar to instrumentation we have elsewhere.
> 
> I've rebased the pull request on the `1.8.x` branch so we can get this fixed in upcoming maintenance releases. The more I've thought about it, the more the previous behavior was more bug-like. Once merged, snapshots will be available if you would like to test things out with your application before the release.
> 
> Thank you for reporting and fixing this.

Thank you! Glad to be able to contribute.

PR #3343, opened 2022-08-09
Add option to disable step allignment of the report in PushRegistryConfig

bogdandrutu:
The problem with the current implementation is that if we use a 60sec (1min) step by default, all the reporting will happen at the same time across all the application on the globe.
That cause huge issues for backends or intermediary services, such as Otel Collector which has to handle a large burst, and because of that a larger footprint is needed.

Screen Shot 2022-08-09 at 4 28 35 PM

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
@@ -52,6 +52,11 @@ public interface OtlpConfig extends PushRegistryConfig {
         return getUrlString(this, "url").orElse("http://localhost:4318/v1/metrics");
     }
 
+    @Override
+    default boolean stepAlignment() {
+        return false;
+    }
+
     /**
      * Attributes to set on the Resource that will be used for all metrics published. This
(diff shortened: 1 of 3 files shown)


bogdandrutu:
Not sure what check you have, no instructions on what to do (it is expected to have a new public method):

```
***! MODIFIED INTERFACE: PUBLIC ABSTRACT io.micrometer.core.instrument.push.PushRegistryConfig  (not serializable)
	+++! NEW METHOD: PUBLIC(+) boolean stepAlignment()
```

izeye:
> Not sure what check you have, no instructions on what to do (it is expected to have a new public method):
> 
> ```
> Comparing binary compatibility of  against 
> ***! MODIFIED INTERFACE: PUBLIC ABSTRACT io.micrometer.core.instrument.push.PushRegistryConfig  (not serializable)
> 	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
> 	===  UNCHANGED SUPERCLASS: java.lang.Object (<- java.lang.Object)
> 	+++! NEW METHOD: PUBLIC(+) boolean stepAlignment()
> ```

@bogdandrutu See https://github.com/micrometer-metrics/micrometer/issues/3181

PR #3348, opened 2022-08-10
Allow Timer implementation to provide their Histogram to AbstractTimer

bogdandrutu:
This PR adds a new protected constructor to `AbstractTimer` that accepts an externally created Histogram. The usefulness of this API is proven by changing 2 implementation Prometheus and OTLP to use the new API.

For OtlpTimer to benefit more would require a change to CumulativeTimer as well (not sure if it is worth it since OtlpTimer will need to support delta, so will most likely needs it's own count/total/max), which I did not want to do in this PR, since the benefit of providing the Histogram is already visible.

This simplifies a lot the logic of overwriting what implementation of the Histogram to use in different implementations, and removes duplicate logic of replacing the fixed bucket histogram implementation.

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpTimer.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpTimer.java
@@ -15,42 +15,46 @@
  */
 package io.micrometer.registry.otlp;
 
-import io.micrometer.common.lang.Nullable;
+import io.micrometer.core.instrument.AbstractTimer;
 import io.micrometer.core.instrument.Clock;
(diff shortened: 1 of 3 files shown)


bogdandrutu:
Wrong japicmp (the super class still has it):

```
***! MODIFIED CLASS: PUBLIC io.micrometer.prometheus.PrometheusTimer  (not serializable)
	===  UNCHANGED SUPERCLASS: n.a.
	---! REMOVED METHOD: PUBLIC(-) io.micrometer.core.instrument.distribution.HistogramSnapshot takeSnapshot()
```

jonatan-ivanov:
There were a few changes around this since this PR was opened: `AbstractTimer` now has the same ctor that can accept a `Histogram` (also the OTLP registry now supports delta not only cumulative). Can this be closed?

PR #3357, opened 2022-08-18
Abstraction for instrumentating with a Timer or Observation (optionally)

shakuzen:
We have a lot of existing instrumentations with Timer that are good candidates to be rewritten with the Observation API for added extensibility. However, we cannot break compatibility for existing users by requiring an ObservationRegistry for each instrumentation. We will want to add an option to configure an ObservationRegistry on the builder for the instrumentation or as an additional constructor if there isn't a builder. Instrumentation should still work as before without providing an ObservationRegistry. This can be handled in each instrumentation, but this added class ObservationOrTimerCompatibleInstrumentation is intended to make it easier to support this case where instrumentation may be done with a Timer or Observation depending on whether an ObservationRegistry is available. By having a common abstraction for this, we can hopefully make supporting Observation in our existing instrumentations easier, and also avoid adding overhead to instrumentation by avoiding instantiating things for Observation when a Timer is used (and the opposite situation).

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2022 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.observation;
+
+import io.micrometer.common.lang.Nullable;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationRegistry;
+import io.micrometer.observation.transport.RequestReplySenderContext;
+
+import java.util.function.Supplier;
+
+/**
+ * Abstracts instrumenting code with a {@link Timer} or an {@link Observation}. This can
+ * be useful for avoiding complexity and minimizing overhead when converting
+ * instrumentation that was previously instrumented with a {@link Timer} to being
+ * optionally instrumented with an {@link Observation}. This may be necessary where
+ * backwards compatibility is a concern such that you cannot require an
+ * {@link ObservationRegistry}. If there are no backwards compatibility concerns,
+ * generally direct instrumentation with {@link Observation} should be preferred. While
+ * this was designed for use internally in micrometer-core where we have this need, it may
+ * also be useful to other libraries with pre-existing {@link Timer}-based
+ * instrumentation. If an {@link ObservationRegistry} is provided that is not the no-op
+ * registry, an {@link Observation} will be used for instrumentation. Otherwise, a
+ * {@link Timer} will be used.
+ *
+ * @param <T> context type if Observation used for instrumentation
+ */
+public class ObservationOrTimerCompatibleInstrumentation<T extends Observation.Context> {
+
+    private final MeterRegistry meterRegistry;
+
+    private final ObservationRegistry observationRegistry;
+
+    @Nullable
+    private final Observation.ObservationConvention<T> convention;
+
+    private final Observation.ObservationConvention<T> defaultConvention;
+
+    @Nullable
+    private Timer.Sample sample;
+
+    @Nullable
+    private Observation observation;
+
+    @Nullable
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
```diff
@@ -0,0 +1,141 @@
...
+ * generally direct instrumentation with {@link Observation} should be preferred. While
+ * this was designed for use internally in micrometer-core where we have this need, it may
+ * also be useful to other libraries with pre-existing {@link Timer}-based
+ * instrumentation. If an {@link ObservationRegistry} is provided that is not the no-op
+ * registry, an {@link Observation} will be used for instrumentation. Otherwise, a
+ * {@link Timer} will be used.
+ *
+ * @param <T> context type if Observation used for instrumentation
+ */
+public class ObservationOrTimerCompatibleInstrumentation<T extends Observation.Context> {
```
marcingrzejszczak, line 44: Do you add `@since` on public classes?

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
```diff
@@ -0,0 +1,141 @@
...
+            Observation.ObservationConvention<T> defaultConvention) {
+        this.meterRegistry = meterRegistry;
+        this.observationRegistry = observationRegistry == null ? ObservationRegistry.NOOP : observationRegistry;
+        this.convention = convention;
+        this.defaultConvention = defaultConvention;
+    }
+
+    private void start(Supplier<T> contextSupplier) {
+        if (observationRegistry.isNoop()) {
+            sample = Timer.start(meterRegistry);
```
marcingrzejszczak, line 99: we have  `sample = ` but then we have `this.context` but then we have `observation = `

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * If using an Observation for instrumentation and the context is a
+     * {@link RequestReplySenderContext}, set the response object on it. Otherwise, do
+     * nothing.
+     * @param response response for the RequestReplySenderContext
+     * @param <RES> type of the response
+     */
+    public <RES> void setResponse(RES response) {
+        if (observationRegistry.isNoop() || !(context instanceof RequestReplySenderContext)) {
```
marcingrzejszczak, line 115: What about `RequestReplyReceiverContext` ? We have the response there too?

shakuzen: Good catch. I had only tried this with the Apache HTTP client instrumentation, which is why I probably only thought of this. I expect we'll catch other things and add to it as we use this in more places.
Do you think it makes sense to extract the response-related methods to an interface so we could check for that here and just cast to that?

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
shakuzen: I wish the format task would normalize this for us automatically. I'll make it consistent.

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
shakuzen: In general, yes. I was still kind of on the fence when writing this whether it should be meant as an internal to Micrometer class (but needs to be public to be used from other packages) or if we should support other projects using it. I think it may make sense for some other projects that aren't having a major version when upgrading to Micrometer 1.10 but want to support Observation instrumentation in a backwards-compatible way like we do with our instrumentation.
I'll add `@since` - thanks.

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
marcingrzejszczak: We can have a checkstyle rule for this

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
```diff
@@ -0,0 +1,141 @@
...
+            return;
+        }
+        RequestReplySenderContext<?, ? super RES> requestReplySenderContext = (RequestReplySenderContext<?, ? super RES>) context;
```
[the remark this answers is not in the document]
shakuzen: @sonatype-lift ignore

jonatan-ivanov (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
shakuzen: I'll follow-up with this in a separate pull request. I've left a TODO comment for now.

Issue #2511, opened 2021-03-17
java.net.http.HttpClient metrics


Lhaaits:
I didn't find any open tickets about this, so I'm shooting in this issue.
I noticed when using the java 11 java.net.http.HttpClient, I don't see the http.client.request metrics appear in prometheus. Is this not supported yet? Or am I missing something? All other default metrics are visible, just not my calls with the HttpClient

I'm using micrometer-core 1.3.5

shakuzen:
We haven't had any requests for it so far, so we don't have instrumentation for it currently. Also, with java.net.http.HttpClient being added in Java 11, while we support down to Java 8, we would either need to do a multi-release JAR or make a separate module that is Java 11 based to include instrumentation for it, I think. I'm not aware of built-in support in Gradle or a stable plugin for multi-release JARs, which would be better than a separate module for one instrumentation. If someone has a proposal, we can consider it. I worry about the complexity that might be added to the build to support multi-release JARs right now.

Lhaaits:
That's fair. I didn't realize support was up to java 8.
I suppose as a workaround I can define my own similar timers:
`Timer.Sample sample = Timer.start();
            response = httpClient.send(request, HttpResponse.BodyHandlers.ofByteArray());
            sample.stop(Metrics.timer("http.client.request", "status", String.valueOf(response.statusCode()), "uri", uriTemplate, "method", request.method()));
`

shakuzen:
> I didn't realize support was up to java 8.

It's not up to 8 in the sense that Micrometer consumers can use later versions of Java (we build/test with the latest GA version of Java). But we support down to 8.

Maybe I was overthinking it. Since we don't need to switch classes based on Java version, perhaps we can just define a Java 11 source set and include it in the produced JAR rather than a multi-release JAR. However, I'm not sure what the correct metadata is for the JAR if part of it requires 11 while the rest is 8.

Lhaaits:
I see, so you mean it should be backwards compatible with java 8, which did not have the java.net.http.HttpClient, so you need a way to know if you should monitor this client, depending on the version, right?

shakuzen:
[...]

PR #3358, opened 2022-08-18
JDK11 HttpClient instrumentation

marcingrzejszczak:
- [x] rewrite this to use https://github.com/micrometer-metrics/micrometer/pull/3357

Related issue https://github.com/micrometer-metrics/micrometer/issues/2511


micrometer-core/src/main/java11/io/micrometer/core/instrument/binder/jdk/DefaultHttpClientObservationConvention.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public KeyValues getLowCardinalityKeyValues(HttpClientContext context) {
+        if (context.getCarrier() == null) {
+            return KeyValues.empty();
+        }
+        HttpRequest httpRequest = context.getCarrier().build();
+        KeyValues keyValues = KeyValues.of(
+                HttpClientDocumentedObservation.LowCardinalityKeys.METHOD.withValue(httpRequest.method()),
+                HttpClientDocumentedObservation.LowCardinalityKeys.URI.withValue(httpRequest.uri().toString()));
```
shakuzen, line 39: We probably want some way for users of HttpClient to specify a templated URI. Or open a follow-up issue for this, since some other http client instrumentation has potentially high cardinality uri tag by default.

shakuzen (commented):
Could we add an implementation of the http client instrumentation tck for this?

micrometer-core/src/main/java11/io/micrometer/core/instrument/binder/jdk/DefaultHttpClientObservationConvention.java
marcingrzejszczak: > We probably want some way for users of HttpClient to specify a templated URI. Or open a follow-up issue for this, since some other http client instrumentation has potentially high cardinality uri tag by default.

I couldn't find a way to pass a templated URI using `HttpClient`. Do you have an example of how to do that?

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
```diff
@@ -121,6 +124,15 @@ public <RES> void setResponse(RES response) {
         responseContext.setResponse(response);
     }
 
+    /**
+     * If using an Observation will set the error on Observation. For metrics it will set
+     * a tag {@code error}.
+     * @param throwable error that got recorded
+     */
+    public void setThrowable(Throwable throwable) {
```
marcingrzejszczak, line 132: @shakuzen I'm setting throwable here and `error` tag for metrics if there was an error (or `none` value if there was none)

shakuzen: I'm not sure if we should be doing that for Timer. That will add an `error` tag where there maybe wasn't one before. An additional tag shouldn't be a problem most of the time when querying, but it can depend how the query is written. Sometime Prometheus queries are written using a [`without` clause](https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators).

In addition to that, some instrumentation may be tagging an error on Timer in a different way - for example, existing Spring http-related instrumentation uses `exception` rather than `error`.

Another interesting side effect is verifying the instrumentation matches the DocumentedObservation. If we are strictly checking there aren't tags/keyvalues produced that aren't documented, then people need to add `error` to their DocumentedObservation low cardinality keyvalues.

shakuzen:
I'm wondering how this affects our CI builds and build requirements for contributors. We should update the contributing guide, but I'm not sure what it should say now. Does the Gradle toolchain download JDK 17 if it isn't available? If so, it feels like we'll need to reconsider the different JDK versions of our CI builds. The point was to test that the project could be built with those versions of the JDK.

micrometer-core/src/main/java11/io/micrometer/core/instrument/binder/jdk/MicrometerHttpClient.java
```diff
@@ -0,0 +1,292 @@
...
+                instrumentation.setThrowable(throwable);
+            }
+            instrumentation.setResponse(response);
+            stopObservationOrTimer(instrumentation, request, response);
+            return response;
+        });
+    }
+
+    private HttpRequest.Builder decorate(HttpRequest httpRequest) {
+        HttpRequest.Builder builder = HttpRequest.newBuilder(httpRequest.uri());
```
shakuzen, line 271: We don't need to do decoration if we're not instrumenting with Observation. It would probably be good to have a shortcut to do nothing in that case, as there will be some measurable cost to recreating the builder.

shakuzen (approved):
Looks good. We should also update the CONTRIBUTING.md.

micrometer-core/src/main/java11/io/micrometer/core/instrument/binder/jdk/MicrometerHttpClient.java
shakuzen: This is complicated to implement cleanly. I'm punting on it for now.

micrometer-core/src/main/java11/io/micrometer/core/instrument/binder/jdk/DefaultHttpClientObservationConvention.java
shakuzen: Looks like this is resolved with the `uriMapper` introduced

shakuzen:
I've updated the build configuration so if you are using JDK 11+ with Gradle, then the toolchain will use that version; otherwise JDK 17 will be used (detected locally or downloaded/installed). This better matches the behavior we had before and doesn't require a property to be passed telling the build to use the Gradle JDK for the toolchain. I've documented this in the CONTRIBUTING guide.

PR #3361, opened 2022-08-18
Basic benchmark for Observation API

jonatan-ivanov:


benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/ObservationBenchmark.java
```diff
@@ -0,0 +1,147 @@
...
+    ObservationRegistry observationRegistry;
+
+    ObservationRegistry noopRegistry;
+
+    Timer timer;
+
+    @Setup
+    public void setup() {
+        this.meterRegistry = new SimpleMeterRegistry();
+        this.timer = Timer.builder("cached.timer").tag("abc", "123").tag("index", "0").register(meterRegistry);
```
shakuzen, line 52: It looks like the cached timer has one more tag (`index`) compared to the non-cached timers. Was this intentional? Would it be better to compare with the same number of tags in all cases unless we are specifically trying to measure a difference due to more tags?

shakuzen (approved):
Looks good to me.

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/ObservationBenchmark.java
jonatan-ivanov: That is a leftover from an experiment with JIT and `NoopObservation`, my mistake, hanks for catching it, removed.

PR #3364, opened 2022-08-19
signalfx: remove unnecessary allocation if the distribution summary does not have histogram

bogdandrutu:
This optimization already in place for Timer.

### DIFF
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
@@ -54,7 +54,7 @@ final class SignalfxDistributionSummary extends AbstractDistributionSummary {
         super(id, clock, CumulativeHistogramConfigUtil.updateConfig(distributionStatisticConfig), scale, false);
         this.countTotal = new StepTuple2<>(clock, stepMillis, 0L, 0.0, count::sumThenReset, total::sumThenReset);
         max = new TimeWindowMax(clock, distributionStatisticConfig);
-        if (isDelta) {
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
```diff
@@ -54,7 +54,8 @@ final class SignalfxDistributionSummary extends AbstractDistributionSummary {
         super(id, clock, CumulativeHistogramConfigUtil.updateConfig(distributionStatisticConfig), scale, false);
         this.countTotal = new StepTuple2<>(clock, stepMillis, 0L, 0.0, count::sumThenReset, total::sumThenReset);
         max = new TimeWindowMax(clock, distributionStatisticConfig);
-        if (isDelta) {
+        if (!distributionStatisticConfig.isPublishingPercentiles()
+                && distributionStatisticConfig.isPublishingHistogram() && isDelta) {
```
jonatan-ivanov, line 58: Is it possible that someone wants to publish both percentiles and histogram (for any reason)?
In that case, I would do 
```suggestion
        if (distributionStatisticConfig.isPublishingHistogram() && isDelta) {
```

bogdandrutu: If that is a possibility, though the current AbstractTimer does not do that, I think your suggestion is good. will change the timer as well.

bogdandrutu: Done.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you! :)

PR #3367, opened 2022-08-23
Events in DocumentedObservation

marcingrzejszczak:


micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
```diff
@@ -46,7 +47,7 @@ class SimpleObservation implements Observation {
 
     private final Collection<ObservationFilter> filters;
 
-    SimpleObservation(String name, ObservationRegistry registry, Context context) {
+    SimpleObservation(@Nullable String name, ObservationRegistry registry, Context context) {
```
jonatan-ivanov, line 50: Are you sure that this can be null here?

micrometer-commons/src/main/java/io/micrometer/common/Event.java
```diff
@@ -0,0 +1,93 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.common;
+
+/**
+ * Represents an event used for documenting instrumentation.
```
jonatan-ivanov, line 19: Isn't the main goal of this class signaling arbitrary events?
See the original javadoc of `Observation.Event`.

micrometer-commons/src/main/java/io/micrometer/common/Event.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @author Marcin Grzejszczak
+ * @since 1.10.0
+ */
+public interface Event {
+
+    /**
+     * Returns event name. Use can use {@code %s} to represent dynamic entries that should
```
jonatan-ivanov, line 27: Isn't this true only for the contextualName?
Do we want to call out that the name should have low cardinality while the contextualName can have high cardinality?

micrometer-commons/src/main/java/io/micrometer/common/Event.java
```diff
@@ -0,0 +1,93 @@
...
+     * be resolved at runtime via {@link String#format(String, Object...)}.
+     * @return event name
+     */
+    String getName();
+
+    /**
+     * Returns event contextual name. Use can use {@code %s} to represent dynamic entries
```
jonatan-ivanov, line 34:
```suggestion
     * Returns event contextual name. You can use {@code %s} to represent dynamic entries
```

micrometer-commons/src/main/java/io/micrometer/common/Event.java
marcingrzejszczak: Yeah let's move it back to `Observation.Event`

micrometer-commons/src/main/java/io/micrometer/common/Event.java
marcingrzejszczak: Copy paste :see_no_evil:

micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
marcingrzejszczak: Yeah I mean look at line 60, we literally `null` the name

marcingrzejszczak:
@cppwfs this might break task cause afair you're using events already. You'll need to change `new Observation.Event(...)` to `Observation.Event.of(...)`

micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
jonatan-ivanov: I get that a `DocumentedObservation`'s `getName()` can return `null` for filtering purposes but if I take a step back, that might be valid for the `DocumentedObservation` but I think `SimpleObservation` should not accept `null` here. What should the meter handler or the tracing handler do in that case?

Also, `DocumentedObservation` has a null-check so that nullability behavior can be isolated there I guess.

PR #3372, opened 2022-08-25
Test against ObservationDocumentation in instrumentation TCK

shakuzen:
Instrumentation being tested against the TCK can optionally provide the DocumentedObservation that specifies the expected semantic naming to be used in the instrumentation. This ensures that the implementation matches the specified documentation, which may be used to generate written documentation.

This adds a new abstract class in between `InstrumentationVerificationTests` and the `HttpClientTimingInstrumentationVerificationTests`/`HttpServerTimingInstrumentationVerificationTests`. The class `InstrumentationTimingVerificationTests` contains methods and logic specifically for timing instrumentation. The goal is to leave `InstrumentationVerificationTests` generic enough that it may be used for other kinds of instrumentation verification. Arguably some of the remaining stuff in there should be moved since `ObservationRegistry` is probably only relevant when verifying some kind of timing instrumentation.

TODO:

- [x] merge and rebase on #3454 before this

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
@@ -27,6 +27,9 @@ import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import static java.util.stream.Collectors.toList;
+import static org.assertj.core.util.Streams.stream;
+
 /**
  * Assertion methods for {@code Observation.Context}s and
  * {@link Observation.ContextView}s.
@@ -194,6 +197,24 @@ public class ObservationContextAssert<SELF extends ObservationContextAssert<SELF
                 failWithMessage("Observation is missing expected keys %s.", missingKeys);
             }
         }
+        return (SELF) this;
+    }
+
+    /**
+     * Verifies that the Observation key-value keys are a subset of the given set of keys.
+     */
+    public SELF hasSubsetOfKeys(String... keys) {
+        isNotNull();
+        Set<String> actualKeys = new LinkedHashSet<>(allKeys());
+        Set<String> expectedKeys = new LinkedHashSet<>(Arrays.asList(keys));
+
+        List<String> extra = stream(actualKeys).filter(actualElement -> !expectedKeys.contains(actualElement))
+                .collect(toList());
+
+        if (extra.size() > 0) {
+            failWithMessage("Observation keys are not a subset of %s. Found extra keys: %s", keys, extra);
+        }
+
         return (SELF) this;
     }
 
--- a/micrometer-test/src/main/java/io/micrometer/core/instrument/InstrumentationTimingVerificationTests.java
+++ b/micrometer-test/src/main/java/io/micrometer/core/instrument/InstrumentationTimingVerificationTests.java
@@ -20,6 +20,7 @@ import io.micrometer.common.lang.Nullable;
 import io.micrometer.observation.docs.ObservationDocumentation;
 import io.micrometer.observation.tck.TestObservationRegistryAssert;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.extension.ExtendWith;
 
 import java.util.Arrays;
 import java.util.HashSet;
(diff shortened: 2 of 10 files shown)


jonatan-ivanov (approved):

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
```diff
@@ -197,6 +200,24 @@ else if (!extraKeys.isEmpty()) {
         return (SELF) this;
     }
 
+    /**
+     * Verifies that the Observation key-value keys are a subset of the given set of keys.
+     */
+    public SELF hasSubsetOfKeys(String... keys) {
```
shakuzen, line 206: Added this to the asserts because it may be useful to others in other places, but also I needed it for the verification with the `ObservationDocumentation` in the TCK.

micrometer-test/src/test/java11/io/micrometer/core/instrument/binder/jdk/JdkHttpClientTimingInstrumentationVerificationTests.java
```diff
@@ -13,10 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micrometer.core.instrument;
+package io.micrometer.core.instrument.binder.jdk;
```
shakuzen, line 16: The `HttpClientObservationDocumentation` is package private, which is why I moved the test here. Maybe it's better to make it public instead? Though I'm not sure why someone outside Micrometer would need to access it.

micrometer-test/src/main/java/io/micrometer/core/instrument/InstrumentationTimingVerificationTests.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        // must have all required tag keys
+        assertThat(timer.getId().getTags()).extracting(Tag::getKey).containsAll(requiredTagKeys);
+        // must not contain tag keys that aren't documented
+        assertThat(timer.getId().getTags()).extracting(Tag::getKey).isSubsetOf(allPossibleTagKeys);
+
+        if (testType == TestType.METRICS_VIA_OBSERVATIONS_WITH_METRICS_HANDLER) {
+            if (observationDocumentation.getDefaultConvention() == null) {
+                TestObservationRegistryAssert.assertThat(getObservationRegistry())
```
marcingrzejszczak, line 81: I wonder if all of that couldn't be hidden inside a new assertion in the `TestObservationRegistryAssert`

marcingrzejszczak (approved):

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
```diff
@@ -197,6 +200,24 @@ else if (!extraKeys.isEmpty()) {
...
+        isNotNull();
+        Set<String> actualKeys = new LinkedHashSet<>(allKeys());
+        Set<String> expectedKeys = new LinkedHashSet<>(Arrays.asList(keys));
+
+        List<String> extra = stream(actualKeys).filter(actualElement -> !expectedKeys.contains(actualElement))
```
jonatan-ivanov, line 211: I'm thinking if we can reuse a built-in assertion from assertJ, it has contains... methods.

jonatan-ivanov (approved):

micrometer-test/src/main/java/io/micrometer/core/instrument/InstrumentationTimingVerificationTests.java
shakuzen: Yes, probably. We can take a look after RC1 I think, as no users are asking for such an assertion now and there's other high priority stuff to work on.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
shakuzen: Probably. I think I copied this from the implementation one of assertJ's methods. Let's take a look after RC1.

PR #3379, opened 2022-08-31
HTTP server instrumentation TCK

shakuzen:
Similar to the http client instrumentation tests, these tests are meant for http server instrumentation to ensure they include the expected tags in a minimal set of scenarios that should be supported.

Pure Jetty servlet instrumentation does not even tag `uri` since getting a templated URI is not generically possible at that level. Therefore the tests are disabled for Jetty server instrumentation.

There are probably more scenarios we should test in all instrumentations. Please let me know more to add. Improvements to the usability/API are also welcome.


micrometer-test/src/test/java/io/micrometer/core/instrument/JerseyServerTimingInstrumentationVerificationTests.java
```diff
@@ -0,0 +1,88 @@
...
+    @Produces(MediaType.TEXT_PLAIN)
+    public static class TestResource {
+
+        @GET
+        public String root() {
+            return "hello";
+        }
+
+        @GET
+        @Path("hello/{name}")
```
marcingrzejszczak, line 69: I wonder if these strings can't be  static finals in the test abstract class. That way you will be sure that you're using the proper strings. WDYT?

shakuzen: I agree it would be good to have some way to avoid typos and improve discoverability beyond the current setup of it being in the JavaDoc. I'm wondering if an enum might make it easy for implementors to find all the routes that should be implemented. I'm going to draft something and see if I can get some feedback on it.

shakuzen: Ah, I tried making the enum and immediately realized that won't work for using here because the annotation needs a constant value, which is doesn't recognize enum fields/methods as constant, even though they effectively are.

shakuzen: See newly added `InstrumentedRoutes` class with string constants. WDYT?

Issue #3234, opened 2022-06-16
Support for Jetty 11


dzikoysk:
**Please describe the feature request.**
Provide implementation for Jetty 11 users. Due to several big changes, modules written for Jetty 9 are incompatible with Jetty 11 (mostly because of Jakarta EE 9) and it includes the current impl:

* https://github.com/micrometer-metrics/micrometer/tree/main/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty

**Rationale**

We're migrating to Jetty 11 from Jetty 9 that is based on top of Jakarta EE 9. 
Unfortunately micrometer currently supports only 9 and it kinda makes it impossible to operate on that module in our case.
Because we're developing an open-source library, not a private product, it will mostly affect Micrometer's users that would like to use it in further releases of [Javalin](https://github.com/javalin/javalin) framework.

**Additional context**
Should be probably provided as a new implementation as there are still Jetty 9 users that don't want to migrate.

shakuzen:
We were able to support two major versions with the same instrumentation when it came to Jersey server by removing references to the javax package classes from API signatures - https://github.com/micrometer-metrics/micrometer/pull/2766. It seems it may be tricky to do the same with Jetty. It's not a problem to have a separate class for instrumenting Jetty 11. A pull request would be welcome if someone wants to work on this.

tipsy:
Jetty 9 is has reached end of life, so this is becoming important: https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.48.v20220622

shakuzen:
I'm not sure how we can support both Jetty 9 and Jetty 11 in the same module since the jetty-server artifact has not changed coordinates. We don't want to suddenly remove our instrumentation for Jetty 9. So we're in a bit of a tough position about how to proceed. We could make a micrometer-jetty11 module, but I would like to avoid that since eventually we would want to get rid of it.

tipsy:
> I'm not sure how we can support both Jetty 9 and Jetty 11 in the same module since the jetty-server artifact has not changed coordinates. 

I'm surprised this isn't a problem that micrometer runs into all the time, most libraries break on major without changing coordinates?

> We don't want to suddenly remove our instrumentation for Jetty 9. So we're in a bit of a tough position about how to proceed. We could make a micrometer-jetty11 module, but I would like to avoid that since eventually we would want to get rid of it.

[...]

PR #3382, opened 2022-09-01
Jetty 11 module

shakuzen:
I've copied just `TimedHandler` and adapted to necessary changes for Jetty 11. I originally copied all of the classes in the same package, but I'm not sure there is any problem using the ones compiled against Jetty 9 with Jetty 11, since none of them contain `javax` package imports and the relevant Jetty API seems to be unbroken.

I have added some Jakarta-specific classes to Micrometer core for tagging, mirroring the corresponding javax servlet classes.

Jetty 11 is Java 11 baseline, so this module is also Java 11 baseline, which isn't a problem for our build anymore since we dropped the Java 8 CI build.

See #3234


micrometer-jetty11/src/test/java/io/micrometer/jetty11/TimedHandlerTest.java
```diff
@@ -0,0 +1,435 @@
+/*
+ * Copyright 2019 VMware, Inc.
```
jonatan-ivanov, line 2: Should we keep the copyright year in this case?

shakuzen:
@tipsy if you have time to review this, we'd appreciate any feedback.

shakuzen:
We can still see what the Jetty team thinks about packaging and releasing this code in a Jetty module, but I wanted to put together what it would be here first. It's mostly copy-paste from the Jetty 9 support with import changes.

tipsy:
> @tipsy if you have time to review this, we'd appreciate any feedback.

@zugazagoitia checked it out on my behalf and tested it against Javalin 5.0.0.RC1 - everything appears to be working fine :)

shakuzen:
I copied this over with the Timer-based instrumentation, but proybably it should be done with the ObservationOrTimerInstrumentation instead. We would need to add a builder to optionally pass the ObservationRegistry, and then the question is whether to leave public constructors (from the copied TimedHandler) or not.

shakuzen:
> probably it should be done with the ObservationOrTimerInstrumentation instead.

I'll have to look at that in a follow-up issue. I don't want to block getting this into the upcoming release on that.

PR #3383, opened 2022-09-01
Context allows modification of KeyValues

marcingrzejszczak:


micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -848,8 +848,9 @@ public void clear() {
          * the {@link ObservationConvention#getLowCardinalityKeyValues(Context)} method.
          * @param keyValue a key value
          */
-        void addLowCardinalityKeyValue(KeyValue keyValue) {
+        public Context addLowCardinalityKeyValue(KeyValue keyValue) {
```
jonatan-ivanov, line 851: Where would we use these?
In a handler/filter?

marcingrzejszczak: Correct - example of usage

```java
public class KubernetesObservationFilter implements ObservationFilter {
	
	@Override
	public Observation.Context map(Observation.Context context) {
		if (context instanceof AppDeployerContext) {
			RuntimeEnvironmentInfo runtimeEnvironmentInfo = ((AppDeployerContext) context).getRuntimeEnvironmentInfo();
			if (runtimeEnvironmentInfo != null) {
				return context.addLowCardinalityKeyValues(k8sTags(runtimeEnvironmentInfo.getPlatformSpecificInfo()));
			}
		}
		return context;
	}

	private KeyValues k8sTags(Map<String, String> platformSpecificInfo) {
		KeyValues keyValues = KeyValues.empty();
		if (platformSpecificInfo.containsKey("master-url")) {
			keyValues = keyValues.and(PlatformKeyName.K8S_URL.withValue(platformSpecificInfo.get("master-url")));
		}
		if (platformSpecificInfo.containsKey("namespace")) {
			keyValues = keyValues.and(PlatformKeyName.K8S_NAMESPACE.withValue(platformSpecificInfo.get("namespace")));
		}
		return keyValues;
	}
}
```

jonatan-ivanov (approved):

PR #3386, opened 2022-09-02
Remove unnecessary overrides in Timer.Builder

izeye:
This PR removes unnecessary overrides in `Timer.Builder`.


jonatan-ivanov (approved):

jonatan-ivanov:
Thank you! :)

jonatan-ivanov:
I ended-up reverting this change because it breaks binary compatibility: https://github.com/micrometer-metrics/micrometer/commit/379f326cece24422ce39f7b7ac30d74d53f42aaf

`japicmp` failed on `main` because of the removed methods, I checked it manually, it's right:
```
Exception in thread "main" java.lang.NoSuchMethodError: 'io.micrometer.core.instrument.Timer$Builder io.micrometer.core.instrument.Timer$Builder.description(java.lang.String)'
```

izeye:
@jonatan-ivanov Ah, sorry. I'm not familiar with binary compatibility, and I still don't understand why the inherited ones are not counted for it, but it'd be good to catch it earlier, so I created https://github.com/micrometer-metrics/micrometer/pull/3388.

PR #3387, opened 2022-09-02
Observation.ObservationConvention -> ObservationConvention

jonatan-ivanov:


shakuzen:
After merging this, we need to merge https://github.com/micrometer-metrics/micrometer-docs-generator/pull/24 for projects using the micrometer-docs-generator.

Issue #2394, opened 2021-01-04
Prometheus  http_server_requests_seconds_bucket{le="+Inf"} value randomly incorrect during high load


jtorkkel:
Scrape Prometheus metrics from uservice running micrometer which is configured to collect also http_server histogram and quantiles.

Query for generating Grafana heatmap, template variable filtering removed from query:
sum(rate(http_server_requests_seconds_bucket{}[$range])) by (le)

Grafana heatmap dashboard shows randomly long response time for +Inf with count -0.033 following value 0.033 in next scrape. This cause glitches to metrics,

Detailed analysis indicates that max value never exceed 4s, but there is change in values between 30.0 and +Inf.
http_server_requests_seconds_bucket{le=~".Inf|30.0"}
http_server_requests_seconds_max

Detailed dumps from Prometheus raw metrics
le="+Inf" values
99060 @1609756122.131
102076 @1609756152.134  **# too low value, 1 smaller than 30.0**
105008 @1609756182.131
107932 @1609756212.131

le="30.0" values
99060 @1609756122.131
102077 @1609756152.134  **# correct value for 30s**
105008 @1609756182.131
107932 @1609756212.131

The issue happen only during high load which hint that generating +Inf value from largest bucket value is not "atomic". 

**Environment:**
scrape interval: 30s
Micrometer 1.5.9
Java 8 192
Linux

marcingrzejszczak:
There was no interest in the issue since 2021. We’ll close it for now, and if there still will be any interest, please comment on this issue and we’ll consider re-opening it.

Issue #998, opened 2018-11-09
Timer/distribution summary snapshot values not self-consistent


john-karp:
While the name might suggest otherwise, the timer/distribution summary takeSnapshot() implementations do not atomically capture statistics, so that `count()`, `max()`, and `histogramCounts()` are not necessarily self-consistent. This can lead to situations where the total counts of the histogram buckets may not equal `snapshot.count()`. Or a downstream system reading metrics might get incorrect results computing total / count at a given time because the time periods they measured are effectively different.

jrask:
For us this is not are rare case, it happens more or less all the time so at the moment this is a blocker for us. Are there any thought about if and how this could be solved? Would be great if you could point me in the right direction.

jrask:
Thanks, so there is no plan or proposal how to solve this for "real"? My guess was that this was a concurrency issue. I could look into it, I was just hoping to perhaps get some initial intro to the problem.

What we are considering is that there are two scenarios, one which is ok and that
is when we have durations that are longer than our max SLA. In that case it would be ok with Count higher than CountAtBucket[lastValue].count() , right?

So would a workaround like this be good enough?
```
timer.count > CountAtBucket[lastValue] && Timer.max <= SLA.Max
== Rely on CountAtBucket[lastValue] as Count. Do nothing else.

timer.count > CountAtBucket[lastValue] && Timer.max > SLA.Max
== Rely on Count as Count, add an extra bucket with Timer.count -  CountAtBucket[lastValue].count
      which is from SLA_MAX TO inf.
```

jkschneider:
I think we could always rely on the count on a +Inf bucket to be the count when histograms are being recorded. They can only diverge now because count is maintained independently of histogram counts, so if a snapshot is taken between when count has been updated and the histogram is updated, they can appear inconsistent. Neither value is "more correct" in absolute terms, but consistency within a snapshot is preferable.

jrask:
Hi, I just have to leave a comment on this again since I really do not understand what happens. This is a result of a Timer being invoked with record() every 100ms and then we have steps=2s (have tried with different intervals, same result). Then everything just runs in a while true loop.

The result shows a big difference between histogram.count and histogram cumulative count. Is this really just "an atomic capture" issue? Anyway, if we are not able to get this to work properly I do not really know what to do. This is not a rare situation, it happens all the time.

It is not really useful to use count=13 men the reality is 20?

**Start**

[...]

PR #3394, opened 2022-09-06
Fix concurrency issue with cumulative bucket counts

wvanlint-square:
This fixes a concurrency issue described in #2394 where cumulative bucket counts are calculated from the beginning for each bucket, resulting in inconsistent values.

If possible, can this be backported to 1.8.x?


micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
```diff
@@ -116,6 +117,31 @@ void resetAccumulatedHistogram() {
...
+                return values.hasNext();
+            }
+
+            @Override
+            public CountAtBucket next() {
+                double value = values.next();
+                double count = currentHistogram().countAtValue(value);
+                if (cumulativeBucketCounts) {
+                    accumulation += count;
+                    return new CountAtBucket(value, accumulation);
```
shakuzen, line 136: Is overriding this method here needed? Does the `countAtValue` implementation above not take care of the logic here?

shakuzen (commented):
Thank you for the pull request. I'm having trouble understanding why this change results in different behavior. Could you help me understand that? Also, this may not change anything for the linked issue because it is about the +Inf bucket, which may get its value from the `count()` method rather than the histogram, causing the potential inconsistency.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
wvanlint-square: @shakuzen I think the problem is that the cumulative count is recomputed for each bucket.

For example, imagine if there are bucket boundaries for values 4.0 and 5.0 and the cumulative count less or equal to 4.0 has just been computed as part of `takeCountSnapshot`:
https://github.com/micrometer-metrics/micrometer/blob/994ec8bb8bea96ab072829aa50754c6cf4a3e69f/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java#L215 Then, a new value of 1.0 comes in before computing the cumulative count less or equal to 5.0. The count <= 5.0 will be greater than the count <= 4.0 even though no such value has been seen.

Does that make sense? You're probably right that a fix would need to be added for `+Inf` as well, perhaps including it in the histogram?

shakuzen: Thank you for walking me through it. That makes sense. What I missed in my initial reading of the changes is that line 133 is calling `currentHistogram().countAtValue(value)` rather than `countAtValue(value)`. This effectively makes the method `countAtValueCumulative` never called anymore, and it feels like the code should be further reworked to reflect that. The proposed approach is more efficient algorithmically, and it avoids an inconsistency. It's a step in the right direction. The snapshot still isn't atomic, though. Going off your example, if there were also a 10.0 bucket and the value recorded during taking the snapshot were 7, then the 10.0 bucket is going to have a count 1 higher than when the snapshot started being taken.

I wrote a jcstress test to capture this in https://github.com/micrometer-metrics/micrometer/pull/3396. With the changes proposed here, failure cases are reduced but not entirely eliminated.

> for `+Inf` as well, perhaps including it in the histogram?

If the buckets do not already include infinity, it will be added using the `count`. The current logic for the infinity bucket is: https://github.com/micrometer-metrics/micrometer/blob/994ec8bb8bea96ab072829aa50754c6cf4a3e69f/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L267-L279
There's some discussion on it https://github.com/micrometer-metrics/micrometer/issues/998 previously.

wvanlint-square: > This effectively makes the method countAtValueCumulative never called anymore, and it feels like the code should be further reworked to reflect that.

I removed the now unused methods.

> It's a step in the right direction. The snapshot still isn't atomic, though.

I believe our examples have slightly different properties though. This pull request only fixes the computation of cumulative counts so that they don't exhibit counts for values that were never seen. It indeed does not make the computation atomic.

Making that computation atomic would require synchronization primitives which would affect performance. The current consistency - performance trade-off could be sufficient for metrics, it is correct for some reordering of operations.

PR #3401, opened 2022-09-09
Add "createNotStarted" method that lazily creates context

ttddyy:
Add another variant of `observation` method to delay the observation context creation.  When the registry is a no-op, it fast returns the no-op observation and skips calling the supplier.  This avoids the creation of the observation context.

Ideally, the supplier should be pushed down to more method calls to further delay the context creation, but at least this gives an API that minimize the creation of context.

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/docs/DocumentedObservation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/docs/DocumentedObservation.java
@@ -19,11 +19,13 @@ import io.micrometer.common.docs.KeyName;
 import io.micrometer.common.lang.NonNull;
 import io.micrometer.common.lang.Nullable;
 import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationPredicate;
 import io.micrometer.observation.ObservationRegistry;
 import io.micrometer.observation.GlobalObservationConvention;
 import io.micrometer.observation.ObservationConvention;
 
 import java.util.Objects;
+import java.util.function.Supplier;
 
 /**
  * In order to describe your samples via e.g. enums instead of Strings you can use this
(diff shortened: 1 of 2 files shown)


micrometer-observation/src/main/java/io/micrometer/observation/docs/DocumentedObservation.java
```diff
@@ -187,6 +188,30 @@ default <T extends Observation.Context> Observation observation(@Nullable Observ
...
+     * @param customConvention convention that (if not {@code null}) will override any
+     * pre-configured conventions
+     * @param defaultConvention default convention that will be picked if there was
+     * neither custom convention nor a pre-configured one via
+     * {@link ObservationRegistry.ObservationConfig#observationConvention(GlobalObservationConvention[])}
+     * @param contextSupplier observation context supplier
+     * @param registry observation registry
+     * @return observation
+     */
+    default <T extends Observation.Context> Observation createNotStarted(
```
jonatan-ivanov, line 206: Can we also add a null check since the `registry` is `@Nullable`? See:
https://github.com/micrometer-metrics/micrometer/blob/da7c9f3fd3482d6c546aaf9d66dae30841494731/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java#L102-L105

Also, should not we do this in the `Observation` too?

jonatan-ivanov, line 206: I think we should call these methods `createNotStarted` but I'm not sure we all agreed on that.

I think either we call all of them `createNotStarted` or `observation`.

ttddyy: `registry` is non-null on this method, so don't need null check here.

I'll add documentation that this no-op registry check is a quick check and further decision(`ObservationPredicate`) would happen after context is created.

jonatan-ivanov: I'll go to an eye doctor. :D

PR #3409, opened 2022-09-13
Override default spliterator method in Tags and KeyValues

kilink:
Tags and KeyValues were inheriting their spliterator method from Iterable, which returns a Spliterator with unknown size and no characteristics set.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java
@@ -149,13 +149,18 @@ public final class Tags implements Iterable<Tag> {
 
     }
 
+    @Override
+    public Spliterator<Tag> spliterator() {
+        return Spliterators.spliterator(tags, 0, last, Spliterator.IMMUTABLE | Spliterator.ORDERED
+                | Spliterator.DISTINCT | Spliterator.NONNULL | Spliterator.SORTED);
+    }
+
     /**
      * Return a stream of the contained tags.
      * @return a tags stream
      */
     public Stream<Tag> stream() {
-        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator(),
-                Spliterator.ORDERED | Spliterator.DISTINCT | Spliterator.NONNULL | Spliterator.SORTED), false);
+        return StreamSupport.stream(spliterator(), false);
     }
 
     @Override
--- a/micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
@@ -150,13 +150,18 @@ public final class KeyValues implements Iterable<KeyValue> {
 
     }
 
+    @Override
+    public Spliterator<KeyValue> spliterator() {
+        return Spliterators.spliterator(keyValues, 0, last, Spliterator.IMMUTABLE | Spliterator.ORDERED
+                | Spliterator.DISTINCT | Spliterator.NONNULL | Spliterator.SORTED);
+    }
+
     /**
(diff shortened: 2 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java
```diff
@@ -149,13 +149,18 @@ public void remove() {
 
     }
 
+    @Override
+    public Spliterator<Tag> spliterator() {
+        return Spliterators.spliterator(tags, 0, last, Spliterator.IMMUTABLE | Spliterator.ORDERED
```
shakuzen, line 154: I doubt it makes much of a difference, but I wonder if there is a case in which passing the range `[0, last]` results in something different than passing the underlying `tags` array without bounds. Looking at the underlying implementation, it would avoid a bounds check by only passing the array.

shakuzen (approved):
Thank you for the pull request. This does seem better in principle and we should do it. I'm curious, though, if you found this due to some negative effect in your code using this.

micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java
kilink: Based on how the `dedup` method works, it looked to me like the number of elements may be less than the array length, is that not the case? If we passed the array without the start / end index, I think we'd potentially produce incorrect or duplicate results, right?

kilink:
> Thank you for the pull request. This does seem better in principle and we should do it. I'm curious, though, if you found this due to some negative effect in your code using this.

I did not observe any actual problems due to this, but it's easy to demonstrate how it can lead to inefficiencies. It is something I stumbled across and seems to be fairly common for code that extends `Iterable`. I get the sense that it's almost never a good idea to rely on the default `spliterator` method from `Iterable`!

Here are a couple of examples of why it can be good to propagate the exact size:

1. Calling `count` on a `Stream` will just return the size directly if it's [known](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/stream/ReduceOps.java#L255-L257), otherwise it's forced to do the reduction / iteration.
2. Calling `trySplit` on a `Spliterator` of unknown size will allocate an array of 1024 elements by default, which is likely much larger than a typical set of tags; granted, I think `trySplit` is more commonly used for parallel streams, but I have seen some code that [makes use of `trySplit`](https://github.com/google/guava/blob/b037a558e88c7587dd77276182a4156c490a2144/guava/src/com/google/common/collect/Streams.java#L884) even for non-parallel streams.

I didn't comb through all of the internals of `Stream`, but there are likely other optimizations that are made possible by propagating the exact size.

kilink:
I forgot to mention that an additional benefit to overriding the `spliterator` method from `Iterable`, is that even though `Tags` has its own `stream` method, users may still call `spliterator` directly, in which case they'd also get a `spliterator` with no characteristics set (SORTED, DISTINCT, etc). That means such operations would also be needlessly inefficient.

micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java
shakuzen: Yes, you are right. I had forgotten (and didn't look closely enough at) what the `dedup` method does.

jonatan-ivanov (approved):

shakuzen:
Thank you for the added details. That makes sense. I was wondering if there were examples of places in code outside of this repository that Tags spliterator is being used, so I can learn about some usage in users' code that I might not have thought about. It's a bit hard to do a GitHub search for these things given the generic method names, and of course there is a lot of code that isn't in GitHub that may be using this.

PR #3410, opened 2022-09-13
Tck with observations

marcingrzejszczak:


micrometer-test/src/main/java/io/micrometer/core/instrument/HttpClientTimingInstrumentationVerificationTests.java
```diff
@@ -49,6 +62,39 @@ public enum HttpMethod {
...
+     */
+    @Nullable
+    protected abstract CLIENT clientInstrumentedWithObservations();
+
+    /**
+     * Returns the instrumented client depending on the {@link TestType}.
+     * @return instrumented client with either {@link MeterRegistry} or
+     * {@link ObservationRegistry}
+     */
+    protected CLIENT instrumentedClient() {
```
shakuzen, line 85: This works, but I wonder if users implementing this class need the instrumented client anywhere but in the `sendHttpRequest` method implementation. Maybe we just add the instrumented http client as an argument there, and then this method can be internal (private). WDYT?

marcingrzejszczak: Great idea!

micrometer-test/src/main/java/io/micrometer/core/instrument/InstrumentationVerificationTests.java
```diff
@@ -29,4 +49,142 @@ protected MeterRegistry getRegistry() {
...
+         * Runs the tests by using the Observation API.
+         */
+        METRICS_VIA_OBSERVATIONS_WITH_METRICS_HANDLER
+
+    }
+
+    // Allows parameter resolution in BeforeEach and AfterEach
+    // Based on
+    // https://code-case.hashnode.dev/how-to-pass-parameterized-test-parameters-to-beforeeachaftereach-method-in-junit5
+    static class AfterBeforeParameterResolver implements BeforeEachMethodAdapter, ParameterResolver {
```
shakuzen, line 88: It's a shame we need to maintain this code here, but unless https://github.com/junit-team/junit5/issues/944 is implemented, I guess we don't have much of a choice.

shakuzen (approved):

Issue #3398, opened 2022-09-09
'java.lang.module.FindException: Unable to derive module descriptor' when using module path with micrometer-core 1.10.0-M5


moarychan:
**Describe the bug**
The compilation failure occurs in our CI pipeline when building using Java 11, here is a temp [URL ](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1840856&view=logs&j=c1fb1ddd-7688-52ac-4c5f-1467e51181f3&t=270374a4-721a-5366-7f60-bad5b1983827&l=17)to see the console log.

I can see the module name is correct in version 1.10.0-M3.
![image](https://user-images.githubusercontent.com/19926623/189333414-db85a325-98bf-45ef-8bf2-b6fef883be3e.png)

It works fine in our CI pipelines based on Java 8.

```console
[ERROR] Error occurred during initialization of boot layer
[ERROR] java.lang.module.FindException: Unable to derive module descriptor for D:\a\_work\1\.m2\repository\io\micrometer\micrometer-core\1.10.0-M3\micrometer-core-1.10.0-M3.jar
```

**Environment**

 - Micrometer version [1.10.0-M3]
 - OS: [windows server 2019]
 - Java version: [11.0.16.1, vendor: Eclipse Adoptium, runtime: C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\11.0.16-101\x64]

**To Reproduce**
How to reproduce the bug:
Please follow this steps:
- Checkout this branch https://github.com/moarychan/azure-sdk-for-java/tree/moary/migrate-spring-data-cosmos-to-spring-dir
- Or sparse checkout below parts:
  ```console
  /*
  !/*/
  /eng
  **/*.xml
  /sdk/cosmos
  /sdk/formrecognizer
  /sdk/parents
  /sdk/spring  
  ```

- Run the below command in the root project:
  `mvn.cmd -f pom.xml  --batch-mode --fail-at-end --settings eng/settings.xml -Dmaven.wagon.httpconnectionManager.ttlSeconds=60 -Dgpg.skip -Dmaven.javadoc.skip=true -Dcodesnippet.skip=true -Dspotbugs.skip=true -Dcheckstyle.skip=true -Drevapi.skip=true -DtrimStackTrace=false -pl com.azure:azure-cosmos, -DAZURE_TEST_DEBUG=false -Dparallel-test-playback=true -Punit -T 1C verify`

**Expected behavior**
Normal use.

[...]

PR #3411, opened 2022-09-13
Automated java module checking

marcingrzejszczak:
I've added a task called `testModules` that essentially calls

```
$ java -p module-name/build/libs/module-name-1.10.0-SNAPSHOT.jar --list-modules
```

If the module system is broken this task will fail.

The motivation to add this check is to catch issues like the one found in #3398.

TODO:

- [x] I failed to retrieve JAVA_HOME from Java Toolchain from Gradle
- [x] Maybe there's something smarter to test this?
- [x] Plug in `testModules` to the whole build

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -286,14 +286,6 @@ subprojects {
         }
     }
 
-    task testModules(type: Exec) {
-        dependsOn "jar"
-
-        commandLine "java", "-p", "${project.buildDir.absolutePath}${File.separator}libs${File.separator}/${project.name}-${project.version}.jar", "--list-modules"
-        
-        standardOutput = new ByteArrayOutputStream()
-    }
-
     if (!['samples', 'benchmarks'].find { project.name.contains(it) }) {
         apply plugin: 'nebula.maven-publish'
         apply plugin: 'nebula.maven-manifest'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -286,6 +286,14 @@ subprojects {
         }
     }
 
+    task testModules(type: Exec) {
```
jonatan-ivanov, line 289: This is failing for me because:
```
Error occurred during initialization of boot layer
java.lang.module.FindException: Unable to derive module descriptor for /.../micrometer/implementations/micrometer-registry-new-relic/build/libs/micrometer-registry-new-relic-1.10.0-SNAPSHOT.jar
Caused by: java.lang.module.FindException: Automatic-Module-Name: micrometer.registry.new.relic: Invalid module name: 'new' is not a Java identifier
```

jonatan-ivanov: Alo, this way, this is executed against the bom, the samples and the benchmarks too, I think this also fails since the bom does not have a jar.

I think it is an easy fix, we just need to put a few lines below, after this section:
```
            jar {
                manifest.attributes.put('Automatic-Module-Name', project.name.replace('-', '.'))
                metaInf {
                    from "$rootDir/LICENSE"
                    from "$rootDir/NOTICE"
                }
            }
```

build.gradle
```diff
@@ -286,6 +286,14 @@ subprojects {
...
+        dependsOn "jar"
+
+        commandLine "java", "-p", "${project.buildDir.absolutePath}${File.separator}libs${File.separator}/${project.name}-${project.version}.jar", "--list-modules"
```
jonatan-ivanov, line 292: This: `libs${File.separator}/` will cause: `libs//` but I would not bother building the path like this, since the jar task has a `archivePath` property so you can do this:

```
task testModules(type: Exec) {
    dependsOn jar
    commandLine 'java', '-p', "$jar.archivePath", '--list-modules'
    standardOutput = new ByteArrayOutputStream()
}
```

jonatan-ivanov (changes requested):

build.gradle
shakuzen: I made a separate issue for the new relic automatic module name: https://github.com/micrometer-metrics/micrometer/issues/3412

shakuzen (approved):

PR #3425, opened 2022-09-20
Polishing Observation API

jonatan-ivanov:
- `DocumentedObservation.createNotStarted` -> `observation` (to make it consistent with others)
- `ObservationRegistry.observationConvention`: removed varargs
- Make `getObservationConvention` and `isObservationEnabled` package-private in the `ObservationRegistry` (should not be used)
- Make `getError` return `Throwable`

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -742,15 +742,6 @@ public interface Observation extends ObservationView {
         }
 
         /**
-         * Optional of an error that occurred while processing the {@link Observation}.
-         * @return optional error
-         */
-        @Override
-        public Optional<Throwable> getErrorOptional() {
-            return Optional.ofNullable(this.getError());
-        }
-
-        /**
          * Error that occurred while processing the {@link Observation}.
          * @return error (null if there wasn't any)
          */
@@ -1036,12 +1027,6 @@ public interface Observation extends ObservationView {
          */
         @Nullable
         ObservationView getParentObservation();
-
-        /**
-         * Optional of an error that occurred while processing the {@link Observation}.
-         * @return optional error
-         */
-        Optional<Throwable> getErrorOptional();
 
         /**
          * Error that occurred while processing the {@link Observation}.
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -367,7 +367,7 @@ public interface Observation extends ObservationView {
      */
     @Override
     default ContextView getContextView() {
-        return this.getContext();
+        return getContext();
     }
 
     /**
@@ -863,6 +863,7 @@ public interface Observation extends ObservationView {
          * Adds a low cardinality key value - those will be appended to those fetched from
          * the {@link ObservationConvention#getLowCardinalityKeyValues(Context)} method.
          * @param keyValue a key value
+         * @return this context
          */
         public Context addLowCardinalityKeyValue(KeyValue keyValue) {
             this.lowCardinalityKeyValues.add(keyValue);
@@ -874,6 +875,7 @@ public interface Observation extends ObservationView {
          * from the {@link ObservationConvention#getHighCardinalityKeyValues(Context)}
          * method.
          * @param keyValue a key value
+         * @return this context
          */
         public Context addHighCardinalityKeyValue(KeyValue keyValue) {
             this.highCardinalityKeyValues.add(keyValue);
@@ -883,6 +885,7 @@ public interface Observation extends ObservationView {
         /**
          * Adds multiple low cardinality key values at once.
          * @param keyValues collection of key values
+         * @return this context
          */
         public Context addLowCardinalityKeyValues(KeyValues keyValues) {
             keyValues.stream().forEach(this::addLowCardinalityKeyValue);
@@ -892,6 +895,7 @@ public interface Observation extends ObservationView {
         /**
          * Adds multiple high cardinality key values at once.
          * @param keyValues collection of key values
+         * @return this context
          */
         public Context addHighCardinalityKeyValues(KeyValues keyValues) {
             keyValues.stream().forEach(this::addHighCardinalityKeyValue);
@@ -945,7 +949,7 @@ public interface Observation extends ObservationView {
     interface Event {
 
         /**
-         * Creates a {@link Event} for the given names.
+         * Creates an {@link Event} for the given names.
          * @param name The name of the event (should have low cardinality).
          * @param contextualName The contextual name of the event (can have high
          * cardinality).
@@ -965,14 +969,13 @@ public interface Observation extends ObservationView {
 
                 @Override
                 public String toString() {
-                    return "event.name='" + this.getName() + "', event.contextualName='" + this.getContextualName()
-                            + '\'';
+                    return "event.name='" + getName() + "', event.contextualName='" + getContextualName() + '\'';
                 }
             };
         }
 
         /**
-         * Creates a {@link Event} for the given name.
+         * Creates an {@link Event} for the given name.
          * @param name The name of the event (should have low cardinality).
          * @return event
          */
@@ -997,13 +1000,13 @@ public interface Observation extends ObservationView {
         }
 
         /**
-         * Creates an event for the given key name.
+         * Creates a new event with the given dynamic entries for the contextual name.
          * @param dynamicEntriesForContextualName variables to be resolved in
          * {@link Event#getContextualName()} via {@link String#format(String, Object...)}.
          * @return event
          */
         default Event format(Object... dynamicEntriesForContextualName) {
-            return Event.of(this.getName(), String.format(this.getContextualName(), dynamicEntriesForContextualName));
+            return Event.of(getName(), String.format(getContextualName(), dynamicEntriesForContextualName));
         }
 
     }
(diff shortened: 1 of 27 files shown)


micrometer-observation/src/main/java/io/micrometer/observation/docs/DocumentedObservation.java
```diff
@@ -201,14 +201,14 @@ default <T extends Observation.Context> Observation observation(@Nullable Observ
...
-     * {@link ObservationRegistry.ObservationConfig#observationConvention(GlobalObservationConvention[])}
+     * {@link ObservationRegistry.ObservationConfig#observationConventions(GlobalObservationConvention[])}
      * @param contextSupplier observation context supplier
      * @param registry observation registry
      * @return observation
      */
-    default <T extends Observation.Context> Observation createNotStarted(
-            @Nullable ObservationConvention<T> customConvention, @NonNull ObservationConvention<T> defaultConvention,
-            @NonNull Supplier<T> contextSupplier, @NonNull ObservationRegistry registry) {
+    default <T extends Observation.Context> Observation observation(@Nullable ObservationConvention<T> customConvention,
```
jonatan-ivanov, line 209: Let's discuss if we want to rename this class.
@shakuzen Did you get some feedback on this?

ttddyy: I thought we rename existing `observation` to `createNotStarted`?
Either way, it is better to be unified.

jonatan-ivanov: Yepp, that's why the let's discuss comment.
I think:
1. If we want to keep the class name as-is: it is better to rename everything to `createNotStarted`
2. If we want to rename the class to `ObservationDocumentation` (or similar): it is better to rename everything to `observation`
3. If we want to keep changes minimal: what I did

I think the best would be # 2 but it is also the most invasive.

micrometer-observation/src/main/java/io/micrometer/observation/ObservationRegistry.java
```diff
@@ -149,7 +149,7 @@ public ObservationConfig observationConvention(GlobalObservationConvention<?>...
          * @param observationConventions observation conventions
          * @return This configuration instance
          */
-        public ObservationConfig observationConvention(
+        public ObservationConfig observationConventions(
```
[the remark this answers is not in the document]
jonatan-ivanov: @sonatype-lift ignore

micrometer-observation/src/main/java/io/micrometer/observation/ObservationRegistry.java
```diff
@@ -139,7 +139,7 @@ public ObservationConfig observationFilter(ObservationFilter observationFilter)
          * @param observationConventions observation conventions
          * @return This configuration instance
          */
-        public ObservationConfig observationConvention(GlobalObservationConvention<?>... observationConventions) {
+        public ObservationConfig observationConventions(GlobalObservationConvention<?>... observationConventions) {
```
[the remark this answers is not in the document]
jonatan-ivanov: @sonatype-lift ignore

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -178,7 +178,7 @@ private void stopObservation(Observation observation, Observation.Scope scope, @
         observation.stop();
     }
 
-    public static class ObservedAspectContext extends Observation.Context {
+    static class ObservedAspectContext extends Observation.Context {
```
shakuzen, line 181: If we make this not public, then users can't make an `ObservationConvention` for `ObservedAspectContext`. Seeing as the public constructor takes an `ObservationConvention<ObservedAspectContext>`, I assume we want to allow users to do that.

jonatan-ivanov: You are right, fixed.

micrometer-observation/src/main/java/io/micrometer/observation/docs/DocumentedObservation.java
shakuzen: I think `ObservationDocumentation` better expresses what the interface is for and makes it clear that it is not an `Observation` itself. The feedback I got so far was that they agreed with that. I think it will be easier to consider that change in a separate pull request, though, so we don't block making the rest of these changes.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -742,12 +742,21 @@ public void setParentObservation(@Nullable ObservationView parentObservation) {
...
         /**
-         * Optional error that occurred while processing the {@link Observation}.
+         * Optional of an error that occurred while processing the {@link Observation}.
          * @return optional error
          */
         @Override
-        public Optional<Throwable> getError() {
-            return Optional.ofNullable(this.error);
+        public Optional<Throwable> getErrorOptional() {
+            return Optional.ofNullable(this.getError());
```
shakuzen, line 750: Maybe it is my personal preference, but if all we're doing is wrapping the other method with `ofNullable`, I feel like we may as well ask users to do that where they want to use an `Optional` rather than have both ways to maybe get the error. I would personally just have `Throwable getError()`. I don't think we have an optional variant anywhere else in our API. What is the reason we think this one method warrants an optional variant?

marcingrzejszczak:
BTW if we're chaning the `DocumetedObservation` to `ObservationDocumentation` we also need to change the `DocumentedMeter` to `MeterDocumentation` and in tracing the `DocumentedSpan` to `SpanDocumentation`. Also the code in `micrometer-docs-generator` needs to be updated

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
jonatan-ivanov: I might be biased (I like `Optional` a lot) but to me returning an optional here feels right because:
- It's a nice way to indicate the absence of a value which should be the common case here
- It tells the user very explicitly about the intention and forces them to handle both cases (presence/absence of the value)
- Reduces the chance or errors related to null
- Increases readability (I think), see:
```
context.getErrorOptional()
    .map(error -> error.getClass().getSimpleName())
    .orElse("none");
```

On the other hand, since this creates a new instance, it's less performant, that's the justification for `getError`.

shakuzen: What I meant to highlight is that none of the the other API in this repo or this module or the Context class returns Optional, so unless there is a specific reason that makes getError different from other places null may be returned (e.g. getContextualName, getParentObservation), it feels inconsistent to have an Optional variant here and not other places. For the sake of consistency, shouldn't we do one of the following: add Optional variants everywhere, not have any Optional variants, or have a reason why getError warrants an Optional variant but other places do not.

I agree it can be a nicer API, but I think the primary use of getError (and other getters on Context) will be in instrumentation on the hot path where we want the least overhead, which means the performance characteristics are especially important, and I think that's the main reason we have avoided Optional in other APIs. If a user wants to use `Optional` it's one `Optional.ofNullable` wrap away, which is all we are doing anyway. And with the nullability annotations we are using, their IDE should let them know the return value may be null.

micrometer-observation/src/main/java/io/micrometer/observation/ObservationRegistry.java
marcingrzejszczak, line 142: If we are to have `observationConventions` then we should also have `observationPredicates`, `observationHandlers` etc.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
marcingrzejszczak: I agree that maybe we should just leave `getError` return `null` and remove the `Optional` version whatsover

micrometer-observation/src/main/java/io/micrometer/observation/ObservationRegistry.java
jonatan-ivanov: Those are singular right now:
```java
public ObservationConfig observationPredicate(ObservationPredicate predicate)
public ObservationConfig observationFilter(ObservationFilter observationFilter) 
```
while `observationConventions` has varargs:
```java
public ObservationConfig observationConventions(GlobalObservationConvention<?>... observationConventions)
```

So either:
```java
public ObservationConfig observationPredicates(ObservationPredicate... predicates)
public ObservationConfig observationFilters(ObservationFilter... observationFilters) 
public ObservationConfig observationConventions(GlobalObservationConvention<?>... observationConventions)
```
or:
```java
public ObservationConfig observationPredicate(ObservationPredicate predicate)
public ObservationConfig observationFilter(ObservationFilter observationFilter) 
public ObservationConfig observationConvention(GlobalObservationConvention<?> observationConvention)
```

What do you think?

jonatan-ivanov: I checked the usages of this, even where we could use varargs we don't and call the method multiple times, I will move it to the singular version so that it will match with `observationPredicate` and `observationFilter`.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
jonatan-ivanov: 👍🏼 I removed the Optional variant.

shakuzen (approved):

PR #3427, opened 2022-09-23
Add Observation instrumentation for gRPC client and server

ttddyy:
This PR adds two gRPC interceptors - `ObservationGrpcClientInterceptor` and `ObservationGrpcServerInterceptor`.

I tried to match the data collecting from `Observation` similar to Sleuth/Brave for spans and existing instrumentation(`MetricsCollecting[Client|Server]Interceptor`).

For metrics, it is slightly different due to the underlying `Observation` infrastructure and `DefaultMeterObservationHandler`.

The existing interceptors create these metrics:
- `grpc.client.requests.sent`  (Counter)
- `grpc.client.responses.received` (Counter)
- `grpc.client.processing.duration` (Timer)
- `grpc.server.requests.sent`  (Counter)
- `grpc.server.responses.received`  (Counter)
- `grpc.server.processing.duration`  (Timer)

The new `Observation` based interceptors creates metrics:
- `grpc.client.received` (Counter)
- `grpc.client.sent` (Counter)
- `grpc.client`  (Timer)
- `grpc.client.active`  (LongTaskTimer)
- `grpc.server.received` (Counter)
- `grpc.server.sent` (Counter)
- `grpc.server` (Timer)
- `grpc.server.active` (LongTaskTimer)

The spans are nearly equivalent to what Brave provide.

image

The existing instrumentation does not have a test.
I have added one for the new interceptors that covers all GRPC use cases, 1:1, 1:stream, stream:1, stream:stream with success and failure patterns.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcClientInterceptor.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcClientInterceptor.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.micrometer.core.instrument.binder.grpc;
+
+import io.grpc.*;
+import io.grpc.Metadata.Key;
+import io.grpc.MethodDescriptor.MethodType;
+import io.micrometer.common.lang.Nullable;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationRegistry;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+/**
+ * A gRPC client interceptor that works with {@link Observation}.
+ * <p>
+ * <b>Usage:</b>
+ * </p>
+ * <pre>
+ * ManagedChannel channel = ManagedChannelBuilder.forAddress("localhost", 8080)
+ *     .intercept(new ObservationGrpcClientInterceptor(observationRegistry))
+ *     .build();
+ * channel.newCall(method, options);
+ * </pre> The instrumentation is based on the behavior of Spring Cloud Sleuth and Brave.
+ *
+ * @author Tadaya Tsuyukubo
+ * @since 1.10.0
+ */
+public class ObservationGrpcClientInterceptor implements ClientInterceptor {
+
+    private static final GrpcClientObservationConvention DEFAULT_CONVENTION = new DefaultGrpcClientObservationConvention();
+
+    private static final Map<String, Key<String>> KEY_CACHE = new ConcurrentHashMap<>();
+
+    private final ObservationRegistry registry;
+
+    @Nullable
+    private GrpcClientObservationConvention customConvention;
+
+    public ObservationGrpcClientInterceptor(ObservationRegistry registry) {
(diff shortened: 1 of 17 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcServerInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+                        (k) -> Key.of(keyName, Metadata.ASCII_STRING_MARSHALLER));
+                return carrier.get(key);
+            });
+            context.setCarrier(headers);
+
+            MethodDescriptor<ReqT, RespT> methodDescriptor = call.getMethodDescriptor();
+            String serviceName = methodDescriptor.getServiceName();
+            String methodName = methodDescriptor.getBareMethodName();
+            MethodType methodType = methodDescriptor.getType();
+            String spanName = methodDescriptor.getFullMethodName();
```
marcingrzejszczak, line 77: Can we actually try to hide the details such as metric name or span name? It's really a contextual name (I know that at the end of the day a span will be named like this but imagine if someone uses the contextual name only for logging and doesn't have tracing on the classpath).

marcingrzejszczak (approved):
Added 1 small remark but other than that this looks great!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcServerInterceptor.java
ttddyy: The contextual name is now hidden and used in convention class only.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcClientInterceptor.java
```diff
@@ -0,0 +1,105 @@
...
+            if (methodName != null) {
+                context.setMethodName(methodName);
+            }
+            context.setFullMethodName(fullMethodName);
+            context.setMethodType(methodType);
+
+            return context;
+        };
+
+        Observation observation = GrpcObservation.CLIENT.observation(this.customConvention, DEFAULT_CONVENTION,
```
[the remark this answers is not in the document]
ttddyy: @sonatype-lift ignoreall

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/GrpcObservationSample.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.io.IOException;
+
+/**
+ * Demonstrates how to use observation gRPC interceptors. To see more details, check the
+ * {@code GrpcObservationTest} in micrometer-core test.
+ *
+ * @author Tadaya Tsuyukubo
+ */
+public class GrpcObservationSample {
```
jonatan-ivanov, line 44: Should we move this to micrometer-samples or keep it here?
I'm totally fine keeping it since it is quite simple.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservation.java
```diff
@@ -0,0 +1,128 @@
...
+import io.micrometer.observation.ObservationConvention;
+import io.micrometer.observation.docs.ObservationDocumentation;
+
+/**
+ * {@link ObservationDocumentation} for gRPC.
+ *
+ * @author Tadaya Tsuyukubo
+ * @since 1.10.0
+ */
+public enum GrpcObservation implements ObservationDocumentation {
```
jonatan-ivanov, line 30: Since `DocumentedObservation` was renamed to `ObservationDocumentation`, should we rename `GrpcObservation` to to reflect that it is a documentation?

ttddyy: Renamed to `GrpcObservationDocumentation` to align with existing other documentation classes.

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/GrpcObservationSample.java
ttddyy: For now, keeping it here.
When all sample classes move to micrometer-samples, this goes as well.

PR #3431, opened 2022-09-26
Update OtlpMeterRegistryTest to run builds on Java 19

izeye:
The following tests are failing:

- `ofEmptyDoesNotAllocate()` and `andEmptyDoesNotAllocate()` in `TagsTest` are failing as their allocated bytes have been changed from 0 to 16 somehow. I'm not sure what makes the difference, so it might be better to investigate further with a dedicated issue if possible.
- `OtlpMeterRegistryTest.distributionSummaryWithHistogramBuckets()` is failing as String representation for `double` has been changed slightly. See "Double.toString(double) and Float.toString(float) may Return Slightly Different Results" item in https://jdk.java.net/19/release-notes

For the first two tests, I just split the tests into two, one for prior versions and the other for 19+.

For the second test, I changed it to compare with `double` values, not with their `String` representation.

### DIFF
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMeterRegistryTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMeterRegistryTest.java
@@ -220,10 +220,9 @@ class OtlpMeterRegistryTest {
         clock.add(OtlpConfig.DEFAULT.step());
         size.record(204);
 
-        assertThat(registry.writeHistogramSupport(size).toString()).isEqualTo("name: \"http.request.size\"\n"
-                + "unit: \"bytes\"\n" + "histogram {\n" + "  data_points {\n" + "    start_time_unix_nano: 1000000\n"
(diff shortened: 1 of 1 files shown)


shakuzen:
> For the second test, I changed it to compare with `double` values, not with their `String` representation.

That sounds reasonable. I think we should add a comment in the code why we're doing that. Otherwise surely I (and probably others) will forget and it might get buried or not easily found in git history.
We should also check if this causes any issues in metrics backends ingesting from some applications on Java 19+ and some on earlier versions. Some buckets may end up slightly different. If it's a problem, maybe we need to do something different in main code.

shakuzen:
I've opened #3436 for investigating the change in allocation.

shakuzen (approved):
I think we'll need these changes in the oldest applicable supported maintenance branch, since we will build those with JDK 19 also. That may require splitting the PR since OTLP was added in 1.9.x, but the Tags allocation tests were added to 1.8.x.

izeye:

I extracted the `TagsTest` changes into https://github.com/micrometer-metrics/micrometer/pull/3437 for the 1.8.x branch and changed the base branch for this PR to the 1.9.x branch as you suggested.

Issue #3436, opened 2022-09-27
Investigate allocation in TagsTest on Java 19+


shakuzen:
See original discovery in #3431.

> `ofEmptyDoesNotAllocate()` and `andEmptyDoesNotAllocate()` in TagsTest are failing as their allocated bytes have been changed from 0 to 16 somehow. I'm not sure what makes the difference, so it might be better to investigate further with a dedicated issue if possible.

marcingrzejszczak:
Isn't this closed via @izeye  https://github.com/micrometer-metrics/micrometer/commit/d99e5540e2f2d6e5875eadf6ae64fd44f1b5c396 ?

```java
// See https://github.com/micrometer-metrics/micrometer/issues/3436
    @Test
    @Issue("#3313")
    @DisabledIfSystemProperty(named = "java.vm.name", matches = JAVA_VM_NAME_J9_REGEX,
            disabledReason = "Sun ThreadMXBean with allocation counter not available")
    @EnabledIf("java19")
    void ofEmptyDoesNotAllocateOnJava19() {
        ofEmptyDoesNotAllocate(16);
    }
```

izeye:
@marcingrzejszczak The commit just seems to have fixed the test on Java 19+, and @shakuzen seems to have created this issue to investigate the allocation difference further.

marcingrzejszczak:
Ah, I misread it, I thought that this checks on any JDK higher or equal to 19. Can't we cahnge this to (or am I missing sth) ?

```java
@EnabledForJreRange(min = JAVA_19)
```

izeye:
@marcingrzejszczak Sure, I created https://github.com/micrometer-metrics/micrometer/pull/4512 as you suggested.

I guess the reason not using `JRE.JAVA_19` was because it wasn't available at that time or at that version on the branch.

izeye:
The tests are failing on Java 20.0.2, so I'm not sure if they have been tested on Java 20+.

See https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/5459/workflows/92926b90-6080-482b-99ff-5222df0baf84/jobs/27922

izeye:
The difference seems to happen only on JDK 19.

I checked it as follows:

- Liberica 19: 16 bytes
- Liberica 19.0.1: 16 bytes
- Liberica 20: 0 bytes
- Liberica 21.0.1: 0 bytes

[...]

PR #3437, opened 2022-09-27
Update TagsTest to run builds on Java 19

izeye:
This PR updates `TagsTest` to run builds on Java 19.

See gh-3431
See gh-3436


shakuzen (approved):
Thank you as always!

PR #3441, opened 2022-09-28
Properly pin gradle-info-plugin to 11.4.1

izeye:
`gradle-info-plugin` 11.4.1 upgrade has been done in f5fa14616fbbcd2396fc893bfe65e41caec1bcf2 but reverted in af7fa7ea58f389ccb50af0a4a22e0488d32d3d08.

This PR restores `gradle-info-plugin` 11.4.1 upgrade.

`buildscript-gradle.lockfile` used to be retained regardless of removal of dependency locks but has been removed in f2977bcf864f8c16934978be2f0c3a43b6a50687. I'm not sure if it's intentional or accidental.


jonatan-ivanov:
This happens because other Gradle plugins are bringing in the old version so when you upgrade the `buildscript-gradle.lockfile`, you get the old version. See, this: https://github.com/nebula-plugins/gradle-info-plugin/issues/99#issuecomment-1232158848.

Btw I think the situation is slightly worse than what @izeye pointed out since it seems the `buildscript-gradle.lockfile` was also removed (we have it for other branches and I think we should have it).

I asked for a fix again, hopefully we get one soon.

@izeye I have no idea how did you catch this but thank you very much!
Could you please propagate this back to `1.8.x` and also update `buildscript-gradle.lockfile` there?

izeye:
@jonatan-ivanov Sure, I changed the base branch to the 1.8.x branch.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you very much!

PR #3443, opened 2022-09-29
Take Supplier<Context> instead of Context

shakuzen:
This allows avoiding instantiation of a Context object in the case that the ObservationRegistry is a no-op. This does not overload the methods by having a variant that takes a Context and one that takes a Supplier because this becomes ambiguous for the compiler in some cases (reported to us when we added one such method to `ObservationDocumentation` previously). A lambda can be used in cases where you have a Context object to pass. Alternatively, you can implement Supplier in your custom Context class to avoid needing a lambda.

### Migration guide

Anywhere you were passing a `Context` object to an `Observation` factory method (static methods on `Observation` and instance methods on `ObservationDocumentation`), the parameter type has been changed to `Supplier<Context>` to allow for lazy instantiation of the context. There are a few options available on how to migrate your code.

#### Lambda

Using a lambda is one way to adapt code to the new API. Before where you may have had code like the following:

```java
JerseyContext jerseyContext = new JerseyContext(event);
Observation observation = JerseyObservationDocumentation.DEFAULT.start(this.jerseyObservationConvention,
                        new DefaultJerseyObservationConvention(this.metricName), jerseyContext, this.registry);
```

It can be updated to:

```java
JerseyContext jerseyContext = new JerseyContext(event);
Observation observation = JerseyObservationDocumentation.DEFAULT.start(this.jerseyObservationConvention,
                        new DefaultJerseyObservationConvention(this.metricName), () -> jerseyContext, this.registry);
```

#### Method reference

If you do not need to pass arguments to the context, you can use a method reference to the constructor.

```java
Observation.start("my.observation", Context::new, observationRegistry)
```

#### Make your context a supplier

Another alternative if you are using a custom context is to make it implement `Supplier` and then you can pass the context object as you were before. For example, an `OkHttpContext` class could be updated to the following.

```java
public class OkHttpContext extends RequestReplySenderContext<Request.Builder, Response>
        implements Supplier<OkHttpContext> {
    // ...
    @Override
    public OkHttpContext get() {
        return this;
    }
```

And then used the same as before, such as:

```java
Observation observation = OkHttpObservationDocumentation.DEFAULT.observation(this.observationConvention,
        new DefaultOkHttpObservationConvention(requestMetricName), okHttpContext, this.registry).start();
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpContext.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpContext.java
@@ -24,6 +24,7 @@ import okhttp3.Response;
 
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * A {@link SenderContext} for OkHttp3.
@@ -31,7 +32,9 @@ import java.util.function.Function;
  * @author Marcin Grzejszczak
  * @since 1.10.0
  */
-public class OkHttpContext extends RequestReplySenderContext<Request.Builder, Response> {
+@SuppressWarnings("jol")
+public class OkHttpContext extends RequestReplySenderContext<Request.Builder, Response>
+        implements Supplier<OkHttpContext> {
 
     private final Function<Request, String> urlMapper;
 
@@ -93,6 +96,11 @@ public class OkHttpContext extends RequestReplySenderContext<Request.Builder, Re
 
     public Request getOriginalRequest() {
         return originalRequest;
+    }
+
+    @Override
+    public OkHttpContext get() {
+        return this;
     }
 
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
@@ -97,13 +97,14 @@ public class ObservationOrTimerCompatibleInstrumentation<T extends Observation.C
         this.defaultConvention = defaultConvention;
     }
 
+    @SuppressWarnings("unchecked")
     private void start(Supplier<T> contextSupplier) {
         if (observationRegistry.isNoop()) {
             timerSample = Timer.start(meterRegistry);
         }
         else {
-            context = contextSupplier.get();
-            observation = Observation.start(convention, defaultConvention, context, observationRegistry);
+            observation = Observation.start(convention, defaultConvention, contextSupplier, observationRegistry);
+            context = (T) observation.getContext();
         }
     }
 
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -59,19 +59,20 @@ public interface Observation extends ObservationView {
      * @return started observation
      */
     static Observation start(String name, ObservationRegistry registry) {
-        return start(name, null, registry);
+        return start(name, () -> null, registry);
     }
 
     /**
      * Creates and starts an {@link Observation}. When no registry is passed or
      * observation is not applicable will return a no-op observation.
      * @param name name of the observation
-     * @param context mutable context
+     * @param contextSupplier mutable context
      * @param registry observation registry
      * @return started observation
      */
-    static Observation start(String name, @Nullable Context context, @Nullable ObservationRegistry registry) {
-        return createNotStarted(name, context, registry).start();
+    static <T extends Context> Observation start(String name, Supplier<T> contextSupplier,
+            @Nullable ObservationRegistry registry) {
+        return createNotStarted(name, contextSupplier, registry).start();
     }
 
     /**
@@ -84,7 +85,7 @@ public interface Observation extends ObservationView {
      * @return created but not started observation
      */
     static Observation createNotStarted(String name, @Nullable ObservationRegistry registry) {
-        return createNotStarted(name, null, registry);
+        return createNotStarted(name, () -> null, registry);
     }
 
     /**
@@ -93,14 +94,17 @@ public interface Observation extends ObservationView {
      * registry is passed or observation is not applicable will return a no-op
      * observation.
      * @param name name of the observation
-     * @param context mutable context
+     * @param contextSupplier supplier for mutable context
      * @param registry observation registry
      * @return created but not started observation
      */
-    static Observation createNotStarted(String name, @Nullable Context context,
+    static <T extends Context> Observation createNotStarted(String name, Supplier<T> contextSupplier,
             @Nullable ObservationRegistry registry) {
-        if (registry == null || registry.isNoop()
-                || !registry.observationConfig().isObservationEnabled(name, context)) {
+        if (registry == null || registry.isNoop()) {
+            return NOOP;
+        }
+        Context context = contextSupplier.get();
+        if (!registry.observationConfig().isObservationEnabled(name, context)) {
             return NOOP;
         }
         return new SimpleObservation(name, registry, context == null ? new Context() : context);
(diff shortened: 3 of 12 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
```diff
@@ -97,13 +97,14 @@ private ObservationOrTimerCompatibleInstrumentation(MeterRegistry meterRegistry,
...
+    @SuppressWarnings("unchecked")
     private void start(Supplier<T> contextSupplier) {
         if (observationRegistry.isNoop()) {
             timerSample = Timer.start(meterRegistry);
         }
         else {
-            context = contextSupplier.get();
-            observation = Observation.start(convention, defaultConvention, context, observationRegistry);
+            observation = Observation.start(convention, defaultConvention, contextSupplier, observationRegistry);
+            context = (T) observation.getContext();
```
shakuzen, line 107: Switched the order here so the supplier isn't called twice.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpContext.java
```diff
@@ -24,14 +24,17 @@
...
 /**
  * A {@link SenderContext} for OkHttp3.
  *
  * @author Marcin Grzejszczak
  * @since 1.10.0
  */
-public class OkHttpContext extends RequestReplySenderContext<Request.Builder, Response> {
+@SuppressWarnings("jol")
+public class OkHttpContext extends RequestReplySenderContext<Request.Builder, Response>
+        implements Supplier<OkHttpContext> {
```
shakuzen, line 37: Demonstrates one way to deal with this breaking change for users. Then they can pass the, in this case, OkHttpContext instead of using a lambda.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -120,21 +124,24 @@ static Observation createNotStarted(String name, @Nullable Context context,
...
+        if (registry.isNoop()) {
+            return Observation.NOOP;
+        }
         ObservationConvention<T> convention;
         if (customConvention != null) {
             convention = customConvention;
         }
         else {
-            convention = registry.observationConfig().getObservationConvention(context, defaultConvention);
+            convention = registry.observationConfig().getObservationConvention(contextSupplier.get(),
```
shakuzen, line 141: This will result in the supplier being called twice, which feels less than ideal.

micrometer-observation/src/main/java/io/micrometer/observation/docs/ObservationDocumentation.java
```diff
@@ -143,7 +141,7 @@ default Observation observation(ObservationRegistry registry) {
      * @param context observation context
      * @return observation
      */
-    default Observation observation(ObservationRegistry registry, @Nullable Observation.Context context) {
+    default Observation observation(ObservationRegistry registry, @Nullable Supplier<Observation.Context> context) {
```
shakuzen, line 144: FIXME: The Supplier shouldn't be `@Nullable` in my opinion.

micrometer-observation/src/main/java/io/micrometer/observation/docs/ObservationDocumentation.java
```diff
@@ -159,13 +157,12 @@ default Observation observation(ObservationRegistry registry, @Nullable Observat
...
      * neither custom convention nor a pre-configured one via
      * {@link ObservationRegistry.ObservationConfig#observationConvention(GlobalObservationConvention)}
-     * @param context observation context
+     * @param contextSupplier observation context
      * @param registry observation registry
      * @return observation
      */
     default <T extends Observation.Context> Observation observation(@Nullable ObservationConvention<T> customConvention,
-            @NonNull ObservationConvention<T> defaultConvention, @NonNull T context,
-            @NonNull ObservationRegistry registry) {
```
shakuzen, line 168: Unrelated to this change, but I removed the `@NonNull` annotations because they are redundant. We annotate the whole package with `@NonNullApi`.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -59,19 +59,20 @@ public interface Observation extends ObservationView {
      * @return started observation
      */
     static Observation start(String name, ObservationRegistry registry) {
-        return start(name, null, registry);
+        return start(name, () -> null, registry);
```
jonatan-ivanov, line 62: Should we provide a null context supplier so users can do something like this: `start(name, Context.NULL, registry)`?

Though if they want to pass null, they can just call the overload where this param is missing.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
jonatan-ivanov: Can we review the noop checks for the whole class, probably refactor it into  a method (or two) and do the registry based check early?
After we passed the registry based check, we can create the context once and pass it to the next method (a private overload without the supplier)?

micrometer-observation/src/main/java/io/micrometer/observation/docs/ObservationDocumentation.java
jonatan-ivanov: +1, but probably we can provide a supplier that return null, i.e.: `Context.NULL`, see my previous comment.

micrometer-observation/src/test/java/io/micrometer/observation/ObservationRegistryTest.java
```diff
@@ -73,20 +73,20 @@ void observationShouldBeNoopWhenPredicateApplicable() {
     @Test
     void observationShouldBeNoopWhenNullRegistry() {
         assertThat(Observation.start("test.timer", null)).isSameAs(NOOP);
-        assertThat(Observation.start("test.timer", new Observation.Context(), null)).isSameAs(NOOP);
+        assertThat(Observation.start("test.timer", Observation.Context::new, null)).isSameAs(NOOP);
         assertThat(Observation.createNotStarted("test.timer", null)).isSameAs(NOOP);
-        assertThat(Observation.createNotStarted("test.timer", new Observation.Context(), null)).isSameAs(NOOP);
+        assertThat(Observation.createNotStarted("test.timer", Observation.Context::new, null)).isSameAs(NOOP);
```
jonatan-ivanov, line 78: I quite like this actually. :D

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/ObservationOrTimerCompatibleInstrumentation.java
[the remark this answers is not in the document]
shakuzen: @sonatype-lift ignore

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
shakuzen: I tried adding static supplier field to `Context`, but it doesn't really save anything compared to using `Context::new` instead, so I think I'll leave it out until/unless we get feedback from users. I'll switch these `() -> null` to `Context::new` instead.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -146,21 +153,21 @@ static <T extends Context> Observation createNotStarted(@Nullable ObservationCon
      */
     static Observation start(ObservationConvention<? extends Context> observationConvention,
             @Nullable ObservationRegistry registry) {
-        return start(observationConvention, null, registry);
+        return start(observationConvention, () -> null, registry);
```
shakuzen, line 156: I couldn't get rid of this `() -> null` lambda because of generics. I wonder if we need this method. What's the use case to provide an `ObservationConvention` that doesn't target `Context` and not provide a context? If it only ever targets `Context`, then we don't need the `? extends Context` and then we can change the lambda to `Context::new`

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
shakuzen: One issue is interfaces can't have private methods. But regardless of that, it's a little hard to centralize this much more than it is, I think, due to the variety of factory methods and the different logic they have. Suggestions welcome, though.

ttddyy: What if we create a dedicated factory/builder class instead of default/static factory methods? The class would centralize noop checks for all variants of observation creations if possible.

shakuzen: Looking at this more, we have two constructors for SimpleObservation, so I think we're going to have at least two places in factory methods we need to check for a no-op registry. It turns out we end up with 3 now because there is the option to make an Observation with a name, with a convention, or to have the convention looked up. The convention look-up requires the context, so we need to get the context already at that point. I've refactored this method to not call one of the other methods and instead instantiate the SimpleObservation itself, which avoids supplying the context twice. It leads to a little code duplication, but the look-up of the convention in the registry is a notable difference that happens in the middle of the logic.

On a dedicated factory/builder class, that's an option. I worry it's a bigger change and we're basically out of time to iterate on and get feedback on such things. I don't know that it would change a lot. The no-op check that can be centralized I think is just calling `registry.isNoop()` and separately `!registry.observationConfig().isObservationEnabled(name, context)`. This is already only done in the 3 places I mentioned above because they are end up doing different things.

A tangential thing is, we're not consistent on whether we accept a `@Nullable ObservationRegistry` or not in the factory methods. We can look at that separately, I think, but we should make that consistent unless there's some reason to be different.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
shakuzen: I'll update the parameter type to `ObservationConvention<Context>` since having another Context type when a `Context` will be created doesn't make sense.

PR #3447, opened 2022-10-03
StatsdMeterRegistryPublishTest fails randomly.

andrew-krasny:
There is serverLatch variable that is initialized multiple times during the test and is used concurrently from multiple threads. This requires some form of synchronization (to guarantee 'happens-before' property between modification/observation code). At least serverLatch should be made volatile. Otherwise countdown() maight be called on the wrong instance..

### DIFF
--- a/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
+++ b/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
@@ -63,7 +63,7 @@ class StatsdMeterRegistryPublishTest {
 
     DisposableChannel server;
 
-    CountDownLatch serverLatch;
+    volatile CountDownLatch serverLatch;
 
     AtomicInteger serverMetricReadCount = new AtomicInteger();
 


jonatan-ivanov:
Thanks you for the PR!

I think I would go one step further: I would delete the entire `serverLatch` member variable and make it a local variable in every test case. In the `startServer` (that calls `countDown`), I would either add a new `CountDownLatch` parameter or a `Runnable` parameter and call `countDown` or `run` on it.

I feel this is a little more robust and passing the latch or a callback (`Runnable`) feels a better fit in functional (/reactive) style. What do you think?

andrew-krasny:
@jonatan-ivanov, thank you for taking a look at this.
Absolutely! The only reason I proposed volatile is that it allows to have a correctness fix with a minimal change and without touching the existent approach. But I totally agree that the idea of having a single serverLatch is not the best one. 
Do you want me to propose another one or do you have plans to fix it yourself?

jonatan-ivanov:
I agree with you, it is simpler and it fixes the issue, I'm fine merging the simpler fix too.
On the other hand, we should do this on `1.8.x` so that builds of older versions will enjoy this too (we are still supporting those).

You don't need to open a new PR, but could you please rebase this (or the local latch fix, whichever you want/have time for) against `1.8.x` and set the base branch of this PR to 1.8.x?

I can take care about the rest and after the fix is merged `1.8.x`, I will propagate it to `1.9.x` and `main`.

andrew-krasny:
Sounds good. Thank you!

I also tried your suggested approach ([change](https://github.com/micrometer-metrics/micrometer/compare/main...andrew-krasny:micrometer:akrasny/StatsdMeterRegistryPublishTest2)) but, surprisingly, faced another issue and didn't get to the root yet (had to postpone in favor of other things)..
Seems that there is a visibility issue with the meterRegistry.statsdConnection and the following loop doesn't always work (resumeSendingMetrics_whenServerIntermittentlyFails):
```
        if (protocol == StatsdProtocol.TCP || protocol == StatsdProtocol.UDS_DATAGRAM) {
            await().until(() -> meterRegistry.statsdConnection.get() != firstClient);
        }
```
I have a feeling that M1 is somehow more aggressive on caching than x86 ;)

jonatan-ivanov:
Tank you very much for fixing this!

PR #3450, opened 2022-10-04
Publish relative to PushMeterRegistry initialization time and align StepMeter boundaries to that

lenin-jaganathan:
This PR tries to solve this issue (https://github.com/micrometer-metrics/micrometer/issues/2818). The start time is captured when the push meter registry is initialized. For Step based meters, this is passed to the Step Value and StepTupple to align them to the registry start time and not the step start time (see - https://github.com/micrometer-metrics/micrometer/issues/1218).

TODO:

- [ ] rebase and target `1.8.x`
- [x] investigate test failures with the `alignToEpoch` config flipped

### DIFF
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
@@ -104,7 +104,7 @@ public class SignalFxMeterRegistry extends StepMeterRegistry {
 
     @Override
     protected void publish() {
-        final long timestamp = clock.wallTime();
+        final long timestamp = (clock.wallTime() / config.step().toMillis()) * config.step().toMillis();
 
         AggregateMetricSender metricSender = new AggregateMetricSender(this.config.source(),
                 this.dataPointReceiverFactory, this.eventReceiverFactory,
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
@@ -104,7 +104,7 @@ public class SignalFxMeterRegistry extends StepMeterRegistry {
 
     @Override
     protected void publish() {
-        final long timestamp = (clock.wallTime() / config.step().toMillis()) * config.step().toMillis();
+        final long timestamp = clock.wallTime();
 
         AggregateMetricSender metricSender = new AggregateMetricSender(this.config.source(),
                 this.dataPointReceiverFactory, this.eventReceiverFactory,
--- a/implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxMeterRegistryTest.java
+++ b/implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxMeterRegistryTest.java
@@ -441,11 +441,10 @@ class SignalFxMeterRegistryTest {
     void shouldNotExportCumulativeHistogramDataByDefault_Timer() {
         MockClock mockClock = new MockClock();
         SignalFxMeterRegistry registry = new SignalFxMeterRegistry(config, mockClock);
-        Timer timer = Timer.builder("my.timer")
-                .serviceLevelObjectives(Duration.ofMillis(1), Duration.ofMillis(10), Duration.ofMillis(100),
-                        Duration.ofMillis(1000))
-                .distributionStatisticExpiry(Duration.ofSeconds(10)).distributionStatisticBufferLength(1)
-                .register(registry);
+        Timer timer = Timer
+                .builder("my.timer").serviceLevelObjectives(Duration.ofMillis(1), Duration.ofMillis(10),
+                        Duration.ofMillis(100), Duration.ofMillis(1000))
+                .distributionStatisticExpiry(Duration.ofSeconds(10)).register(registry);
 
         timer.record(50, TimeUnit.MILLISECONDS);
         timer.record(5000, TimeUnit.MILLISECONDS);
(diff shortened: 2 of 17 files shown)


shakuzen:
> can you change this PR to draft mode?

Why do you think that should be done? I'm in the process of reviewing it now.

jamesmartinpp:
@shakuzen it doesn't really matter. I thought draft mode might be safer. I'm ok proceeding with the PR as is.

shakuzen (approved):
This looks good except the initial scheduling delay doesn't look right to me, and I would like to add tests to verify that logic. I will try to improve the naming of some things. And I would like to switch this new behavior to be the default, with a config option to go back to the previous behavior. I will follow-up with commits for that. And I'll rebase against and target the PR at `1.8.x`.
Thank you for the contribution.

shakuzen:
After flipping the default, there is a test failure. I've run out of time to dig into it, and I'll be on vacation next week. I can come back to this once I'm back from vacation, but feel free to look into the failure in the meantime.

lenin-jaganathan:
@shakuzen  If the default behavior is made to publish relative to the start time, a few tests will fail because of the way they were written. I will try to address those tests.

lenin-jaganathan:
@shakuzen  Addressed the test failures in SignalFx libraray.

lenin-jaganathan:
Added an additional commit to make the metrics timestamp synchronized across instances(report timestamp as the start of the current step) for SignalFx Meter Registry.

implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxMeterRegistryTest.java
```diff
@@ -457,13 +456,13 @@ void shouldNotExportCumulativeHistogramDataByDefault_Timer() {
         mockClock.add(config.step().minus(Duration.ofMillis(1)));
 
         assertThat(getDataPoints(registry, mockClock.wallTime())).hasSize(8)
-                .has(gaugePoint("my.timer.avg", 0.2525), atIndex(0)).has(counterPoint("my.timer.count", 2), atIndex(1))
+                .has(gaugePoint("my.timer.avg", 2.525), atIndex(0)).has(counterPoint("my.timer.count", 2), atIndex(1))
```
shakuzen, line 459: My review comment about investigating the cause of the failures meant understanding exactly how the behavior changed and make sure the change and test made sense. This appears to change the values so the test passes, but I don't think it's demonstrating what the test intended. We need to be clear about what changed and if that's acceptable. I'll write up a summary in another comment.

When investigating this, I found a bug in the assertion logic for `hasValue` which caused it to always return true for double values that would have an int value of 0 (like 0.2525). I'll fix that in a separate issue and rebase this pull request on it.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -104,7 +104,7 @@ else if ("https".equals(apiUri.getScheme())) {
 
     @Override
     protected void publish() {
-        final long timestamp = clock.wallTime();
+        final long timestamp = (clock.wallTime() / config.step().toMillis()) * config.step().toMillis();
```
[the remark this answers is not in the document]
shakuzen: After thinking more about it, I'm wondering why we would change this. Before this change it is sending the timestamp as the time when publishing starts. Why would we change it to the start of the step with this change? That seems like a separate decision to make not directly related to the change of this pull request. I think I'll revert this change. Please open a separate issue or pull request to consider making this change.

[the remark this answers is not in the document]
lenin-jaganathan: Earlier, the reporting was done at the top of the step (for simplicity let's assume 1min step). And if I understand right all the instances will report almost similar timestamps (i,e start of the minute). With this change, the timestamp will be de-synchronized, it will no more be the start of the minute, what I tried to do was to make it to the top of the minute.

Because when sending data from multiple instances without instanceId (or a unique identifier for an instance), the back-end of SignalFx might be able to calculate what is step a particular metric is reported at. Having said that, I will follow-up on this with a different PR after confirming with someone from SignalFx confirm this.

implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxMeterRegistryTest.java
lenin-jaganathan: As you pointed out `hasValue` had a bug all the time and the tests would have failed if it is right. And the other issue is with the TimeWindowFixedBoundaryHistogram and buffer length. That is the whole reason SignalFx Registry went for a Delta Histogram Implementation.

shakuzen: I've changed the tests back to assert what they did before as I think that's what the intention of the tests was. I've had to change the test to record later than the beginning of the step to avoid the histogram counts being rotated out when checking the published data. That whole problem is a preexisting issue that was worked around in the tests by checking the published data at a specific millisecond before the histogram would rotate but after the step-based values (count/sum/avg) would return the desired new values. We can consider if the behavior of histogram rotation is best or not separately from this pull request.

lenin-jaganathan: Thanks.

lenin-jaganathan:
@shakuzen Do you want me to address anything more on this PR?

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushRegistryConfig.java
```diff
@@ -90,6 +90,21 @@ default int batchSize() {
...
+     * this returns true, publishing is scheduled at the beginning of each step interval,
+     * relative to the Unix Epoch. This has the effect of causing all application
+     * instances publishing metrics with this configuration to publish at the same time,
+     * which may overload resources by concentrating the work to publish and ingest
+     * metrics. The more instances you have publishing metrics at the same time, the more
+     * of a problem this will be.
+     * @return false if publishing should be scheduled relative to registry instantiation
+     * time. Default is {@code false} to avoid the documented resource exhaustion issue.
+     */
+    default boolean isPushAlignedGlobally() {
```
shakuzen, line 105: We could probably spend endless time coming up with better and better names, but here's a third iteration where I've avoided using "epoch" in the name because I think it will be less clear what this does compared to focusing on the effect of pushes being aligned globally if you set this to `true`. Anyone who has feedback about the naming, feel free to share (before we release it, ideally).

I've updated the JavaDoc to explain the issue that motivated this configuration being introduced and the reason for its default.

shakuzen:
> Do you want me to address anything more on this PR?

I don't think so. I'm asking for some review internally. Once that is done, we should be good to go. If you have any thoughts on naming as I've recently updated, please share. Or if you have thoughts on the below.

On how we'll apply this where, we'll keep the default for the configuration `false` when merging this to `main` for the next feature release (1.11), which makes the publishing happen not at the same time for every instance. Out of an abundance of caution, now I would like to introduce this in patch releases (1.8.x, 1.9.x, 1.10.x) but initially with the default to `true` so there is no behavior change unless a user overrides the default. If we get some feedback on usage of this in production, we could consider flipping the default in a patch release later, as the current behavior is very problematic in large deployments. Until we get some feedback on usage, I'm not comfortable making the change in a patch release, even with an option for users to opt out of it. Another thing we're considering is making the configuration option deprecated in 1.11 because as long as there is no problem with the behavior, I'd rather not have a configuration toggle for it going forward (we didn't have one until now because we wanted consistent behavior and there wasn't a need for users to be able to choose). At some point in the future, we could consider removing the configuration altogether.

lenin-jaganathan:
Putting this in draft mode. Will close this PR in a week. For additional info see the linked issue for why this fix might not be desired.

lenin-jaganathan:
Fixed by https://github.com/micrometer-metrics/micrometer/pull/3750

PR #3453, opened 2022-10-04
Add a batching counter for grpc

taer:
Discovered that when you're using a statsd meterRegistry, every single request through the client/server interceptors will increment a counter, which then sends a message to statsd. This was causing some delays in our application, especially the grpc methods with large server streamed results

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/AbstractMetricCollectingInterceptor.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/AbstractMetricCollectingInterceptor.java
@@ -103,6 +103,8 @@ public abstract class AbstractMetricCollectingInterceptor {
 
     protected final MeterRegistry registry;
 
+    protected final boolean batchCounter;
+
     protected final UnaryOperator<Counter.Builder> counterCustomizer;
 
     protected final UnaryOperator<Timer.Builder> timerCustomizer;
@@ -114,9 +116,10 @@ public abstract class AbstractMetricCollectingInterceptor {
      * {@link MeterRegistry}. This method won't use any customizers and will only
      * initialize the {@link Code#OK OK} status.
      * @param registry The registry to use.
+     * @param batchCounter Batch request/response counters till the end
      */
-    protected AbstractMetricCollectingInterceptor(final MeterRegistry registry) {
-        this(registry, UnaryOperator.identity(), UnaryOperator.identity(), Code.OK);
+    protected AbstractMetricCollectingInterceptor(final MeterRegistry registry, boolean batchCounter) {
+        this(registry, batchCounter, UnaryOperator.identity(), UnaryOperator.identity(), Code.OK);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


jonatan-ivanov:
I think we have a PR that should superseed this one: https://github.com/micrometer-metrics/micrometer/pull/3329

Also, please read this series, and check what counters are doing in Micrometer, there are thread safety issues in this PR: https://develotters.com/posts/the-most-common-concurrency-issues-and-how-to-fix-them-part-1/

jonatan-ivanov:
Let me close this because of the Dogstatsd PR (https://github.com/micrometer-metrics/micrometer/pull/3329) that superseeds this, please let us know if you disagree.

taer:
@jonatan-ivanov So that other PR is replacing the statsdMeterRegistry and doing a "push" to the datadog statsd?

Also, I don't think there are threading issues here. The only additions are in the sendMessage and the closes. I know the onMessage calls are serial per listener. And we create a whole new listener for each server call.

I'm 99% sure the cancel and close won't be called at the same time. The docs are a bit vague as to that, only that they will only be called once.

Will that version support distributions as well?

jonatan-ivanov:
It is using the `java-dogstatsd-client` (`StatsdClient`) which is able to handle batch, retry, throttling, etc.
`StatsdCounter` is not involved so that will not send data to every increment and you should be able to configure the Datadog client what to do.

Please read the blog posts I sent you, I wrote down what's the problem there (actually there are two problems): `requestBatchCounter++` is not thread safe neither `requestBatchCounter = 0` (together with the increment). Or are you saying that these methods can't be called concurrently from separate threads (i.e.: two `sendMessage` calls)?

Check any `Counter` in Micrometer, they will not increment a primitive from a method that can be called from multiple threads).

taer:
Yeah. I've done a lot of concurrent stuff. Very familiar w/ all the atomics, sycns, mutexes, and LongAdders.  And deal w/ code that serves thousands of clients w/ 8 threads,

The 'listener' class you give the nextCall in the interceptor AFAIK will only ever be called w/ one thread at a time. The `interceptor` creates a new listener for each call, which has it's own non-shared state. It is defiantly impossible to get 2 `omMessage` calls concurrently. GRPC is all based on netty and is evented. You can only have a single event at a time per socket.

The only concern I had was the rare corner case of cancel in the server side. The docs were unclear, but the benefits of not dealing w/ locks or AtomicInteger in a critical piece of code seem to outweigh the risk of a double count.

If the new datadogMeter thing is close, then I'm ok w/ waiting. If it's far out, I'd like to have this reconsidered given the single thread access pattern of GRPC

Thanks!

taer:
BUT, the big question is..  Will this new PR be able to submit client-side distributions? I know the statsd version can. I'm less worried about the counters vs losing the client-side distros,

Thanks

taer:
If you look at how the grpc server stuff works, you get a message sent to your handler. You won't get a second until you use the call.request(int) method(mostly handled by the Stub base class). That does take an int, but the onMessage call in the interceptor is pretty much 1:1 to the invocation of the handler `onNext` call.  So if you asked for 2, you'd expect 2 `onMessage` interceptor calls as well as 2 messages in your handler.

And on the client side, it's definitely not concurrent. You're only sending one at a time.

I'd bet an ETH on that. :)

taer:
@jonatan-ivanov If the other PR is close, then I'm ok w/ closing this. But there should be no threading issues here, and all the added code is opt in as well.

Thanks

jonatan-ivanov:
My main point above was that supporting `java-dogstatsd-client` should solve this issue for everything, not only gRPC. That's the main reason I think we should not merge this. If we do this for gRPC, we should do it for everything. We need a more general solution that does not depend on special implementation and can be used with `Counter` (or other Meter's).

I think even if the counter you introduced will be called from one thread at a time, I think there still could be a thread safety issue if there are more than one threads that can call this (does not need to be concurrent): I think visibility is not guaranteed here, the second thread might not see the value that the first thread wrote even if they run serially (see the blog post I linked).

taer:
Have an ETA on the alternative then? I'm removing the interceptor and replacing it w/ my copy in the meantime. It's causing major performance issues. We can handle an accidental thread cache miss should that happen. I'm pretty sure w/ the existing version, those incredible high rates of `counter.increment()` which result in a line to statsd are also getting dropped, probably at a much higher rate than lack of memory barriers are creating.

jonatan-ivanov:
I think it should go with the 1.11.0 timeline (RC in April, GA in May) but the same timeline would apply for this PR as well.
I'm sorry it is causing performance issues but that's the burden that StatsD brings in.
Other than the Dogstatsd PR (https://github.com/micrometer-metrics/micrometer/pull/3329), you can try the following things:
1. You can use the Datadog registry (uses their HTTP API instead of StatsD, it does not have this issue)
2. [Datadog supports Prometheus](https://www.datadoghq.com/blog/monitor-prometheus-metrics/) so you might be able to use the Prometheus registry and let the datadog agent scrape your Prometheus endpints

PR #3455, opened 2022-10-05
Bump to a medium+ resource class for CI builds

shakuzen:
Our builds are running at max CPU and RAM for most of the CI build on the default `medium` resource class for docker executors. We could probably get some improvement from more resources. This bumps up one level to `medium+` and makes the resource class choice explicit in the build, instead of the implicit default.


shakuzen:
Resource usage [looks much better](https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/3461/workflows/ea4d1140-6950-4fb5-9287-cae227779a5c/jobs/18492/resources) with this bump up to the next step. I'm hoping this can avoid some of the failures we have been seeing lately of builds suddenly dying without a clear cause.

Issue #3149, opened 2022-04-28
OSGi enabled micrometer jars so that they can be deployed as OSGi bundles


jingsx:
**Please describe the feature request.**
support micrometer libraries on osgi based java runtimes.

**Rationale**
currently we are embedding the micrometer dependencies to the service bundles, which is increasing our stack size unnecessarily.
additionally we are also repackaging the micrometer dependencies to a wrapper osgi bundles, and it does not always support all the scenario

shakuzen:
Thank you for raising the enhancement request. I think this is the first time we're having OSGI support requested. Is this something you would be able to contribute? I'm not very familiar with OSGI so I do not know the kind of changes we need to make to support it, but I'm open to supporting it as long as it isn't a burden for the majority of our users who seem to not be using OSGI (I assume based on the lack of this request until now).

If you do not have time to contribute this, that is also fine. We would probably wait for votes on this issue before it were prioritized.

jingsx:
Yes, we can contribute the osgi packageing for micrometer jars. we are currently working on it, we can submit the same change for review.

jonatan-ivanov:
@jingsx Can I have a few clarifying questions? It was 10+ years ago that I built OSGi bundles so I'm not sure what has changed in the specs since then.

Is there anything else you need other than some OSGI-specific metadata in the manifest file? (`BundleActivator`?)
Can shaded dependencies (check the statsd registry) generate problems?
Is there a proper Gradle plugin to generate the OSGI-specific metadata?

driessamyn:
I raised a PR to address this: https://github.com/micrometer-metrics/micrometer/pull/3457

driessamyn:
> @jingsx Can I have a few clarifying questions? It was 10+ years ago that I built OSGi bundles so I'm not sure what has changed in the specs since then.

I'm not the original poster of this, but raised a [PR](https://github.com/micrometer-metrics/micrometer/pull/3457) with a fix, so will try to answer some of these questions.

> Is there anything else you need other than some OSGI-specific metadata in the manifest file? (`BundleActivator`?) 

I don't think so. If Micrometer used `ServiceLoader`, then there would be an impact, for example, but I have validated the core and prometheus modules to work with just the OSGi metadata.
So there isn't any overhead in maintaining the code, as it stands.

> Can shaded dependencies (check the statsd registry) generate problems? 

I don't think they do. The dependencies still get added to the JAR and are included in the `Import-Package` by default:

[...]

PR #3457, opened 2022-10-05
Add BND plugin and OSGi export core package.

driessamyn:
This change enables the generation of OSGi metadata for all JARs (so turns them into OSGi bundles).
A couple of notes:

- `exportcontents: io.micrometer.*` is a bit "brute force". Ideally, we would only export the packages that are considered "API" and people will consume in their applications. However, I'm not familiar enough with the Micrometer structure to judge whether this should be more fine-grained or if this "export all packages" approach is ok.
- OSGi will resolve all dependencies listed in the `Import-Packages` section in the manifest, unless they are explicitly marked as optional.
- Because a jar has OSGi metadata does not necessarily mean it's usable in an OSGi framework, as all the transitive dependencies need to have this metadata too.
- I have tested this change in our local OSGi application using the Prometheus registry only. I suspect not all registries will work in an OSGi environment because of the above comment. This is something that would need to be resolved in the downstream code, however.
- I added an OSGi test to validate that the core bundle, as well as the Prometheus and JMX bundles, resolve correctly in an OSGi framework. These tests can be run by executing: `./gradlew :micrometer-osgi-test:testOSGi` and have also been added to the `check` task.

I believe this PR would also resolve this issue: https://github.com/micrometer-metrics/micrometer/issues/3149


shakuzen:
@driessamyn Could you rebase on the latest `main`? I was trying to do that and see if I could get the build to pass, but I was having trouble getting the task `testOSGi` to pass after rebasing.

driessamyn:
> @driessamyn Could you rebase on the latest `main`? I was trying to do that and see if I could get the build to pass, but I was having trouble getting the task `testOSGi` to pass after rebasing.

@shakuzen I rebased with the main branch and resolve the gradle file conflicts.
I validated the testOSGi target (`./gradlew :micrometer-osgi-test:testOSGi`).

Let me know if there is anything else you need.

build.gradle
```diff
@@ -297,12 +298,23 @@ subprojects {
...
                 }
+
+                bundle {
+                    // workaround for multi-version JARs
+                    //  see https://github.com/bndtools/bnd/issues/2227
+                    bnd '''\
+-fixupmessages: '^Classes found in the wrong directory: .*'
+-exportcontents: io.micrometer.*
+'''
+                }
```
jonatan-ivanov, line 317: THe build is failing (should fail for you locally) since `micrometer-osgi-test` is a new module and gradle tries to download it for binary compatibility check.
Could you please exclude this, I can show you how:

In line 336 (I can't make a comment on that line), you can find this: https://github.com/micrometer-metrics/micrometer/blob/533bae08a95009b9dccf3afae5e139e25b0be4e4/build.gradle#L336

Could you please add the new project name to the empty list, so that it should be:
```groovy
if (!(project.name in ['micrometer-osgi-test'])) { 
```

jonatan-ivanov (changes requested):

build.gradle
driessamyn: Thanks @jonatan-ivanov .

I updated the build and fixed up some checkstyle issues as well.
The build works locally and I can see the circle-ci jdk11 also passes, but the jdk17 one is failing. I will try to have a look at that shortly.

driessamyn: @jonatan-ivanov, the Java 17 build is working as well now. The only failing one is the CLA check. I submitted the signed CLA a while ago, but not sure if I'm missing something in the process.

shakuzen: Thanks for getting the build passing!

> I submitted the signed CLA a while ago, but not sure if I'm missing something in the process.

Did you sign the individual CLA or the corporate CLA? I checked your GitHub username under the individual CLA and a signature was not found.

driessamyn: We signed the Corporate CLA (though https://cla.pivotal.io/sign/pivotal?repositoryId=micrometer-metrics/micrometer&pullRequestId=3457). I asked our legal team for confirmation, and got the following screenshot:

image

Let me know if there is anything else I can check.

shakuzen: Thanks for checking. For corporate CLAs it is either done via the company email domain or a GitHub organization. See [the FAQ](https://cla.pivotal.io/about) toward the bottom. If done via the email domain, an email address with the company email domain name needs to be associated with your GitHub account. If done via your company's GitHub organization, your GitHub account needs to be a member of that organization and that membership has to be public so the CLA tool can see you are part of that GitHub org.

build.gradle
```diff
@@ -21,6 +21,7 @@ buildscript {
         classpath 'de.undercouch:gradle-download-task:5.2.1'
         classpath 'io.spring.javaformat:spring-javaformat-gradle-plugin:0.0.38'
         classpath 'com.diffplug.spotless:spotless-plugin-gradle:6.15.0'
+        classpath 'biz.aQute.bnd:biz.aQute.bnd.gradle:6.3.1'
```
shakuzen, line 24: It looks like 6.4.0 has been released since this PR was opened. Any issue upgrading to the latest?

shakuzen (approved):

build.gradle
driessamyn: Hi @shakuzen .
I'm still struggling a bit to validate this.

The CLA was signed by GitHub user [rburnsr3](https://github.com/rburnsr3) on behalf of R3. 
We are both part of the [corda GitHub organisation](https://github.com/corda/).
Let me know if you can find a reference to that, or if there is another action on our side.

Thanks.

shakuzen: It looks like your membership to the Corda organization is not public. Check, for example, [your GitHub profile](https://github.com/driessamyn) in a private browser. If you make your membership to the Corda org public so the CLA bot can see it, I think things will work.

driessamyn: Thanks @shakuzen 
For some reason I thought that was the default setting.
I have updated my profile and the organisation should now be public.

shakuzen: I'm glad we got it sorted out. Thank you for your patience and efforts.

build.gradle
```diff
@@ -297,12 +298,23 @@ subprojects {
...
                     from "$rootDir/NOTICE"
                 }
+
+                bundle {
+                    // workaround for multi-version JARs
+                    //  see https://github.com/bndtools/bnd/issues/2227
+                    bnd '''\
+-fixupmessages: '^Classes found in the wrong directory: .*'
+-exportcontents: io.micrometer.*
+'''
```
shakuzen, line 316: @jonatan-ivanov thought it would be better if we make the indentation here match the rest of the build file. I think we can do that along with `String#stripIndent()` to have the same String as now. I can take care of this and merge.

jonatan-ivanov (commented):
👍🏼

Issue #575, opened 2018-04-24
Support for appdynamics


bjartek:
It is possible to hook micrometer into https://www.appdynamics.com/ in some way?

Relevant links from slack 
https://docs.appdynamics.com/display/PRO41/Build+a+Monitoring+Extension+Using+Java

https://docs.appdynamics.com/display/PRO42/Analytics+Events+API#AnalyticsEventsAPI-AbouttheAnalyticsEventsAPI

checketts:
I'll be working with a company that uses AppDynamics shortly and should be able to dig into this.

bjartek:
Has there been any progress on this issue?

jkschneider:
@checketts Is there anything we can do from a partner perspective to reach out and ask AppDynamics to publish a jar to maven central?

checketts:
I'm pretty disappointed with AppDynamics. It has a very hierarchical metrics concept, so I've paused on it and instead have been working toward Splunk metrics support via HEC (http event collector).

Progress has been slow on that one too, but I am making progress when I have time to work on it.

jkschneider:
I was looking at AQDL from the outside in today and it looks like you might be able to select dimensions using WHERE clauses. Combining this with [analytics functions](https://docs.appdynamics.com/display/PRO44/Analytics+Functions) seems like it _could_ work. Did you go down the same path?

checketts:
No, I didn't look into that part. I didn't know it had a query language.

I was just ramping up on AppD and trying to get the metrics in when I decided to switch focus.

bjartek:
Any progress on this?

checketts:
No I've tabled my work on this for now.

jkschneider:
@bjartek I did have a chat with an AppD engineer at Oracle CodeOne recently. Looks like we'll essentially have two implementations: a hierarchical one passing through the Java agent and a dimensional one going to the Analytics Event API. Evidently, licenses for the agent-based approach are separate from licenses for the Analytics Event API.

febinrejoe:
Any update on this issue?

shakuzen:
> Any update on this issue?

Only the information available as comments on this issue. Which license do users requesting this have so we can know what solution would work for them? Another solution would be for AppDynamics customers to request AppDynamics to support collecting Micrometer metrics from the AppDynamics Java agent.
[...]

PR #3479, opened 2022-10-08
Micrometer registry appdynamics

rjmveloso:
This PR contributes with a registry for AppDynamics #575

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -83,7 +83,8 @@ def VERSIONS = [
         libs.testcontainers.mongodb,
         libs.testcontainers,
         libs.wiremockJunit5,
-        libs.cloudwatch2
+        libs.cloudwatch2,
+        libs.appdynamics
 ]
 
 def PLATFORM_BOMS = [
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,6 @@
 [versions]
 activemq-artemis = "2.31.2"
+appdynamics = "4.5.19.29263"
 application-insights = "2.6.4"
 archunit = "1.2.1"
 asmForPlugins = "7.3.1"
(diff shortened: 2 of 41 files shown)


rjmveloso:
CircleCi is failing because there is no baseline for AppDynamics yet.

jonatan-ivanov:
@rjmveloso You can disable that check: [see here](https://github.com/micrometer-metrics/micrometer/blob/e904ffd9bca5c8079bde69eb89b27f171784ed88/build.gradle#L327)

rjmveloso:
Thank you @jonatan-ivanov, I ran it locally with `-PcompatibleVersion=SKIP`.
I believe this would be removed from the condition after a first release of the module.

rjmveloso:
> @rjmveloso You can disable that check: [see here](https://github.com/micrometer-metrics/micrometer/blob/e904ffd9bca5c8079bde69eb89b27f171784ed88/build.gradle#L327)

Think your suggestion and my previous observation relates with the issue https://github.com/micrometer-metrics/micrometer/issues/3486

rjmveloso:
It seems Gradle Plugin Portal was facing some outages leading to the following problem (version not in maven central):
```
Could not find org.ajoberstar.grgit:grgit-core:4.0.2.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/ajoberstar/grgit/grgit-core/4.0.2/grgit-core-4.0.2.pom
       - https://plugins.gradle.org/m2/org/ajoberstar/grgit/grgit-core/4.0.2/grgit-core-4.0.2.pom
 ```
 
 It is possible to force a rerun of the actions and also if possible to have a core review?

shakuzen (commented):
As far as I can tell, all tag info is dropped from publishing. Is that correct? That's a lot of valuable information lost. I'm not sure we would want to support that in the Micrometer repo. Does AppDynamics not support dimensions (tags) on metrics?

rjmveloso:
> As far as I can tell, all tag info is dropped from publishing. Is that correct? That's a lot of valuable information lost. I'm not sure we would want to support that in the Micrometer repo. Does AppDynamics not support dimensions (tags) on metrics?

You are correct. There is no direct way to publish tags with AppDynamics agent.
Maybe a solutions would be to create sub-paths for any Custom Metric for each tag, but in the end it will be just replicated metrics with the same values. I will investigate it further.

rjmveloso:
> As far as I can tell, all tag info is dropped from publishing. Is that correct? That's a lot of valuable information lost. I'm not sure we would want to support that in the Micrometer repo. Does AppDynamics not support dimensions (tags) on metrics?

@shakuzen There is no support for tags/dimensions on AppDynamics.
I cannot provide further support because I don't have access to appd anymore.
If AppDynamics have support for OpenTelemetry (as read in some thread related with the subject) I would suggest anyone interested to go for it and use otlp micrometer exporter .

Issue #3458, opened 2022-10-06
Provide a default for missing values in KeyValue


jonatan-ivanov:
Right now there is no guidance about what to do when a value is missing and users want to create a `KeyValue`.

PR #3493, opened 2022-10-18
Add KeyValue.of() variants for none values

izeye:
This PR adds `KeyValue.of()` variants for "none" values for convenience.

See gh-3458

### DIFF
--- a/micrometer-commons/src/main/java/io/micrometer/common/KeyValue.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/KeyValue.java
@@ -58,6 +58,24 @@ public interface KeyValue extends Comparable<KeyValue> {
     }
 
     /**
+     * Creates a {@link KeyValue} for the given key and {@value NONE_VALUE}.
+     * @param key key of the KeyValue
+     * @return KeyValue
+     */
+    static KeyValue of(String key) {
+        return of(key, NONE_VALUE);
+    }
+
+    /**
+     * Creates a {@link KeyValue} for the given {@link KeyName} and {@value NONE_VALUE}.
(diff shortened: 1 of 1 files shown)


shakuzen:
We discussed this internally some time ago. My feeling is that it is a bit too context-dependent for us to provide something in the API like this. For example, when it comes to an `exception` tag, the absent value makes sense as `none` if we know there was no exception, but in other cases the absent value is better represented as `unknown`. For that reason, I would rather we don't add these methods. Also we haven't seen a request from a user related to this yet. @jonatan-ivanov @marcingrzejszczak thoughts?

jonatan-ivanov:
I think if somebody does something like this, there will be a check if a `none` or non-`none` 😃 value is needed.
I think in these cases it is better to use the `NONE` constant (more explicit/readable).
e.g.:

```java
.lowCardinalityKeyvalue(KeyValue.of("something", something != null ? something : NONE));
```
or
```java
String value = something != null ? something : NONE;
.lowCardinalityKeyvalue(KeyValue.of("something", value));
```

---
**vs.**

```java
.lowCardinalityKeyvalue(something != null ? KeyValue.of("something", something) : KeyValue.of("something"));
```
or
```java
KeyValue keyValue = something != null ? KeyValue.of("something", something) : KeyValue.of("something", value);
.lowCardinalityKeyvalue(keyValue);
```

shakuzen:
I think that makes sense. Though if we wanted to take it a step further, we could accept `null` values and convert to the none constant. I think that's a bigger change and not everyone would be happy with the behavior either. I'll close this for now and we can reopen or reconsider this kind of API if we get additional feedback.

Issue #3300, opened 2022-07-18
Non-monotonic kafka metric reported as counter


jack-berg:
**Describe the bug**
[KafkaMetrics](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java#L252-L253) converts the metric `kafka.consumer.connection.count` to a counter despite the number of connections _not_ being monotonically increasing. 

This is causing issues in in the [opentelemetry micrometer shim](https://github.com/open-telemetry/opentelemetry-java/tree/main/micrometer1-shim), which relies counters being monotonically increasing for a proper translation to the opentelemetry data model. 

**Environment**
 - Micrometer version: latest
 - Micrometer registry: opentelemetry micrometer shim
 - OS: any
 - Java version: any

**To Reproduce**
Run the [KafkaClientMetricsIntegrationTest](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaClientMetricsIntegrationTest.java#L113) and observe that the printed meters should `kafka.consumer.connection.count` is a counter where it should probably be a gauge. 

**Expected behavior**
`kafka.consumer.connection.count` and other non-monotonically increasing kafka metrics should probably be gauges instead of counters. 

Originally discussed in CNCF slack [here](https://cloud-native.slack.com/archives/C024DRQ63UP/p1657828698290849).

shakuzen:
Thank you for reporting the issue.
It looks like it would be correct for us to remove the OR conditional checking for names ending in `count` when deciding whether to make a counter or gauge since all the names ending in count appear to be an active count (gauge).
https://github.com/micrometer-metrics/micrometer/blob/b617100e92f55584f3207a29c8e0338d73042b73/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java#L252

jack-berg:
> It looks like it would be correct for us to remove the OR conditional checking for names ending in count when deciding whether to make a counter or gauge

[...]

PR #3496, opened 2022-10-19
Kafka metrics ending in "-count" are gauges

taer:
I am using the kafka metrics module from mirometer. There are just 2 matrics in the set that end in count, and they're both gauges.

This is the output from the SimpleMeterRegistry
```
kafka.consumer.connection.count(COUNTER)[client.id='FOO', kafka.version='5.5.8-ccs']; count=3.0
kafka.producer.connection.count(COUNTER)[client.id='FOO', kafka.version='5.5.8-ccs']; count=2.0
```

The endsWith("total") is correct.

If you check confluent's docs here: https://docs.confluent.io/platform/current/kafka/monitoring.html#global-connection-metrics

you can see that anything ending in `-count` is a gauge style metric

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
@@ -249,7 +249,7 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
     }
 
     private Meter registerMeter(MeterRegistry registry, MetricName metricName, String meterName, Iterable<Tag> tags) {
-        if (meterName.endsWith("total") || meterName.endsWith("count")) {
+        if (meterName.endsWith("total")) {
             return registerCounter(registry, metricName, meterName, tags);
         }
         else {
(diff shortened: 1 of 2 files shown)


taer:
Although with further inspection of the metrics, it looks like another independent metric might be a dupe but with a different name.

```
kafka.consumer.connection.count(COUNTER)[client.id='FOO', kafka.version='5.5.8-ccs']; count=3.0
kafka.consumer.connections(GAUGE)[cipher='TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384', client.id='FOO', kafka.version='5.5.8-ccs', protocol='TLSv1.2']; value=3.0
```
It has some extra tags, but it's the proper type.

Still I think those 2 connection.counts should be gauges, but it's not-blocking

shakuzen (commented):
Thanks for the pull request. This was previously reported in #3300. As discussed there, this change should get us to a better state than we're currently in, but we may want to look into a more robust solution like was mentioned by Jack.
Do you think we could add some testing to make sure this stays fixed?
Also, can you rebase the changes on `1.8.x` so we can include the fix in the 1.8.x line and merge forward to 1.9.x and 1.10.0?

taer:
rebased to 1.8.   Let me look into the tests for the class

shakuzen (approved):
Thank you. I've added an assertion to the integration test as this seemed the quickest and easiest way to check the behavior change for now.

PR #3497, opened 2022-10-21
Remove sourceSets java11Test configuration in micrometer-core

izeye:
This PR removes `sourceSets` `java11Test` configuration in `micrometer-core` as it seems to be unnecessary now somehow.


shakuzen:
> it seems to be unnecessary now somehow.

I think that's due to the upgrade to the mrjar plugin 0.1.1. See https://github.com/melix/mrjar-gradle-plugin/pull/5

Issue #3502, opened 2022-10-26
Add KeyValuesTest


philwebb:
**Please describe the feature request.**
Whilst looking to contribute something to `KeyValues` I noticed there are no tests. I think `TagsTest` should be migrated.

**Rationale**
I like tests

**Additional context**
I think `TagsTest` could be used as the basis of the new class

PR #3503, opened 2022-10-26
Add support for creating KeyValues from any iterable

philwebb:
We're finding quite often that we need to adapt `Iterable<Tag>` to a `KeyValues` instance and writing code like this:

```java
Iterable<Tag> tags = this.tagsProvider.getTags(context.getUriTemplate(), context.getCarrier(),
		context.getResponse());
KeyValues keyValues = KeyValues.empty();
for (Tag tag : tags) {
	keyValues = keyValues.and(tag.getKey(), tag.getValue());
}
return keyValues;
```

It seems like this pattern might be quite common (for example you might want to adapt a `Map`). It would be nice if it could be a oneliner:

```java
return KeyValues.of(tags, Tag::getKey, Tag::getValue);
```

### DIFF
--- a/micrometer-commons/src/main/java/io/micrometer/common/KeyValue.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/KeyValue.java
@@ -15,6 +15,7 @@
  */
 package io.micrometer.common;
 
+import java.util.function.Function;
 import java.util.function.Predicate;
 
 import io.micrometer.common.docs.KeyName;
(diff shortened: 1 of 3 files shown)


philwebb:
If #3502 gets fixed before this then the test class I added will need to be rebased on top of it.

shakuzen (approved):
Sounds reasonable to me. Thank you for the pull request.

marcingrzejszczak (approved):

PR #3507, opened 2022-10-28
Verify sender propagation for HTTP server instrumentation

shakuzen:
This will catch issues in the implementation of instrumentation. For example, if the carrier is set after an Observation is started, this new test will not pass.

Server equivalent of #3504


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/JerseyContext.java
```diff
@@ -36,7 +36,7 @@ public class JerseyContext extends RequestReplyReceiverContext<ContainerRequest,
     public JerseyContext(RequestEvent requestEvent) {
         super((carrier, key) -> {
             List<String> requestHeader = carrier.getRequestHeader(key);
-            if (requestHeader.isEmpty()) {
+            if (requestHeader == null || requestHeader.isEmpty()) {
```
shakuzen, line 39: A NPE was being thrown when trying to use the Getter on a key that wasn't in the headers. The above `getRequestHeader` returns null for an absent key.

marcingrzejszczak (approved):

jonatan-ivanov (approved):

PR #3512, opened 2022-11-03
Add peer name and port to gRPC observation contexts

ttddyy:
For gRPC instrumentation:
- Add peer name and port
- Use `UNKNOWN` for unknown tag values

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/DefaultGrpcClientObservationConvention.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/DefaultGrpcClientObservationConvention.java
@@ -15,12 +15,8 @@
  */
 package io.micrometer.core.instrument.binder.grpc;
 
-import io.micrometer.common.KeyValue;
 import io.micrometer.common.KeyValues;
 import io.micrometer.core.instrument.binder.grpc.GrpcObservationDocumentation.LowCardinalityKeyNames;
-
-import java.util.ArrayList;
-import java.util.List;
 
 /**
  * Default convention for gRPC client. This class defines how to extract values from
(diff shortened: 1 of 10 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/DefaultGrpcServerObservationConvention.java
```diff
@@ -46,10 +46,17 @@ public KeyValues getLowCardinalityKeyValues(GrpcServerObservationContext context
         List<KeyValue> keyValues = new ArrayList<>();
         keyValues.add(LowCardinalityKeyNames.METHOD.withValue(context.getMethodName()));
         keyValues.add(LowCardinalityKeyNames.SERVICE.withValue(context.getServiceName()));
+        keyValues.add(LowCardinalityKeyNames.SYSTEM.withValue("grpc"));
         keyValues.add(LowCardinalityKeyNames.METHOD_TYPE.withValue(context.getMethodType().name()));
         if (context.getStatusCode() != null) {
             keyValues.add(LowCardinalityKeyNames.STATUS_CODE.withValue(context.getStatusCode().name()));
         }
+        if (context.getPeerName() != null) {
+            keyValues.add(LowCardinalityKeyNames.PEER_NAME.withValue(context.getPeerName()));
```
shakuzen, line 55: I wonder if it would be better to use a value such as `UNKNOWN` if the peer name is null. What do you think?

ttddyy: @shakuzen Yes, I agree giving `UNKNOWN` value is better rather than omitting the tag.
I have updated to specify `UNKNOWN` when value is not available(`null`).
Please check.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/DefaultGrpcClientObservationConvention.java
```diff
@@ -43,14 +39,15 @@ public String getContextualName(GrpcClientObservationContext context) {
...
-        }
-        return KeyValues.of(keyValues);
+        String statusCode = context.getStatusCode() != null ? context.getStatusCode().name() : UNKNOWN;
+        String peerPort = context.getPeerPort() != null ? context.getPeerPort().toString() : UNKNOWN;
+        return KeyValues.of(LowCardinalityKeyNames.METHOD.withValue(context.getMethodName()),
+                LowCardinalityKeyNames.SERVICE.withValue(context.getServiceName()),
+                LowCardinalityKeyNames.SYSTEM.withValue("grpc"),
+                LowCardinalityKeyNames.METHOD_TYPE.withValue(context.getMethodType().name()),
+                LowCardinalityKeyNames.STATUS_CODE.withValue(statusCode),
+                LowCardinalityKeyNames.PEER_NAME.withValue(context.getPeerName()),
```
shakuzen, line 49: I think we may have talked about this, but unfortunately I've forgotten - should peer name be low cardinality or high cardinality?

jonatan-ivanov: @ttddyy What do you think?

ttddyy: I think it is a low cardinality. The peername is essentially a hostname of the remote call target.
Different from making http calls to external network web servers, usually remote call targets should be a fixed number of servers with fixed ports within the network.
So, I don't expect the number of the target to be many.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationDocumentation.java
```diff
@@ -72,6 +72,12 @@ public String asString() {
                 return "rpc.service";
             }
         },
+        SYSTEM {
+            @Override
+            public String asString() {
+                return "rpc.system";
```
shakuzen, line 78: Could we add assertions to the tests for this if we're adding it?

shakuzen (commented):
Sorry this sat for a while. I left one comment, but I think if we get it rebased, we can get it merged.

shakuzen:
@ttddyy if you have a chance to pick this back up, let us know if it's still worth doing.

ttddyy:
@shakuzen
Originally, the change added peer name, peer port, and rpc system tags to gRPC observation. However, after reconsideration, I removed the rpc system tag. Since gRPC is the only RPC instrumentation in Micrometer, the tag always had the value `grpc`, making it redundant. I believe it's unnecessary and would only increase backend data.

I've rebased the change to `main`.

rafal-dudek:
Hello, when can we expect it to be merged? 
It should fix the issue https://github.com/micrometer-metrics/micrometer/issues/5609

shakuzen (approved):

Issue #3499, opened 2022-10-24
Improve implementation of Kafka metrics type mapping


shakuzen:
> It looks like it would be correct for us to remove the OR conditional checking for names ending in count when deciding whether to make a counter or gauge

That does appear to be the case. I implemented something similar in opentelemetry recently and inspected the type of the `KafkaMetric`'s `org.apache.kafka.common.metrics.Measurable`, and defined a [mapping](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/kafka/kafka-clients/kafka-clients-common/library/src/main/java/io/opentelemetry/instrumentation/kafka/internal/KafkaMetricRegistry.java#L36-L50) from measurable implementation to corresponding instrument type. This seemed safer to me than relying on naming naming conventions.

_Originally posted by @jack-berg in https://github.com/micrometer-metrics/micrometer/issues/3300#issuecomment-1189570193_

PR #3513, opened 2022-11-04
Improve implementation of Kafka metrics type mapping

izeye:
This PR improves implementation of Kafka metrics type mapping by applying the suggestion in the opening comment of #3499.

Closes gh-3499


shakuzen (approved):
Thank you. This looks great. I think it would be good if we test a known counter metric is correctly registered as a Counter, so that if something changes in a new version of Kafka, the test will hopefully catch it.

izeye:

I added an assertion in 30f3cd3 that hopefully covers what you suggested above. Please let me know if it doesn't.

shakuzen (approved):

PR #3529, opened 2022-11-08
Adds an option to allow removal of low & high cardinality key values

marcingrzejszczak:
without this change it's hard to mutate a context using observationfilter because you don't have access to e.g. all properties in a map or you can't remove a key value

with this change you are allowed to completely mutate the context


ttddyy (approved):

ttddyy:
The PR was merged to `main` but was backported to `1.10.x`. https://github.com/micrometer-metrics/micrometer/commit/990e8174462ddd88bbc095803b3426d4b93f0fa2

PR #3532, opened 2022-11-13
Add nullable to Observation.parentObservation

ttddyy:


shakuzen:
I think this should have been targeted at `1.10.x` so it goes in the next patch release. It seems more like a bug that we missed this than a new feature.

shakuzen:
I've cherry-picked it to `1.10.x` in 7f42bb79c49a4a0701ece8d21f957bdfcf09c45f

PR #3533, opened 2022-11-13
Publish incomplete step values on application shutdown

lenin-jaganathan:
For a Step Registry, during the graceful shutdown, the publishSafely is called and the meters are read. Since the time between the last completed step and graceful shutdown will most probably be less than the step interval, the registry reports the last completed interval data to the backend (see issue [1882](https://github.com/micrometer-metrics/micrometer/issues/1882) ). 

This PR addresses this behavior by forwarding the step registry clock by a Step on graceful shutdown and treating the last incomplete step as a new step. This reads the correct data recorded during this interval.


micrometer-core/src/main/java/io/micrometer/core/instrument/step/SkewableClock.java
```diff
@@ -0,0 +1,72 @@
...
+        return this.skewedTimeInMillis;
+    }
+
+    public Clock getOriginalClock() {
+        return this.clock;
+    }
+
+    @Override
+    public long wallTime() {
+        return clock.wallTime() + this.skewedTimeInMillis;
```
jonatan-ivanov, line 64: I'm not sure this is a good idea, many registries record the time, this is lying to them.

lenin-jaganathan: Yeah. I agree with you. I have added an additional commit that makes sure anything other than StepRegistry will see the clock as a default clock will no modifications and only StepRegistry will align to the next step on the last read.

Let me know if this is going in an wrong direction and you have a better solution that can help improve the StepRegistry implementation.

jonatan-ivanov: I still think that tampering the clock is not a good idea.
My PoC is a much bigger change (basically adding the capability to report the current data not the previous step) so when I will get back to this, I will make sure to consider this PR since it is much simpler.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -102,4 +102,16 @@ protected DistributionStatisticConfig defaultHistogramConfig() {
                 .merge(DistributionStatisticConfig.DEFAULT);
     }
 
+    @Override
+    public void close() {
+        // Move clock to start of next step and stop it.
+        if (clock instanceof StoppableClock) {
+            long now = clock.wallTime();
```
mbbeaubi, line 109: You need to publish metrics from the previous batch before incrementing the clock here, otherwise they will be lost. I would do inside the if:

if (config.enabled()) {
            publishSafely();
        }
        stop();

still call close on the super regardless after this executes.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/SkewableClock.java
mbbeaubi: > I still think that tampering the clock is not a good idea. My PoC is a much bigger change (basically adding the capability to report the current data not the previous step) so when I will get back to this, I will make sure to consider this PR since it is much simpler.

To avoid regression, could you just provide StopClock as a separate class that could be used optionally if the user wanted this behavior (by default, we just get the normal behavior).

I like this fix, but it won't solve my problem because we use APM to push metrics to elastic. We never close our meter registries, every time interval APM client just reports the values of all meters in all meter registries to the server. 

That's why I like the idea of adding the capability of a meter being able to report the current data, depending on how you implement it. If you could create a new interface like:

StepMeter {
  Iterable<Measurement> measureCurrent();  // return measurements from the current step
}

All meters that use the previous / current technique (StepCounter, StepTimer, and StepDistributionSummary) implement this.

Then inside the close method of PushMeterRegistry, look for meters of this type and push twice. The advantage of this is that APM could also look for meters of this type and publish again during its shutdown hook (of course getting APM folks to do this would be another project for me).

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
lenin-jaganathan: I get what you are saying. Basically, assuming the step as 60 seconds. If I close the registry in 1st second, there is a probability that I am still publishing values for the last step. In this scenario, just skipping publish should be sufficient. Since @jonatan-ivanov  said they are implementing a better fix for this issue, the smart thing for me is to wait until I hear back from them.

lenin-jaganathan: Another thing is, the micrometer currently doesn't have a way not to publish duplicate data for StepRegistries. E.g: StepCounter holds a value of 50 for a minute and if publish is called again within the minute same 50 will be reported which is again an issue. This scenario I mentioned above might happen when closing registries and this is exactly one of the issues we are facing. This is not just about short-lived batches, but long-running processes can over-report the value since they are reporting the value for the previous minute. This can be addressed by adding a boolean in Pushregistry and a variable to keep track of the last published in the StepMeterRegistry.

Same as above will wait for @jonatan-ivanov to comment about the fix their team is working on. I have some more things that I am happy to commit on this PR if required.

shakuzen:
Thank you for the pull request. We ended up taking a different approach to solve the problem in #3681, but we appreciate your effort and the discussion on this approach.

PR #3540, opened 2022-11-16
Add "Observation.observe(Function)" method

### LINKED ISSUE #3537: Consider to add Observation.observe(Function<C extends Observation.Context, T> function)
I have a use-case where I need to do `ResponseContext.setResponse()` after performing request-reply and withing an `Observation` scope.
So, to achieve that I have to do something like this:
```
		RequestReplyReceiverContext context = new MyCtx();

		return ObservationDoc.PROCESS.observation(this.observationConvention,
						DefaultConvention.INSTANCE,
						() -> context, this.observationRegistry)
				.observe(() -> {
					Message<?> reply = requestReply();
					context.setResponse(reply);
					return reply;
				});
```

This way I have to be sure about an `ObservationRegistry.NOOP` before calling this code.
With an `Observation.observe(Function<C extends Observation.Context, T> function)` API it would look like this:
```
		return ObservationDoc.PROCESS.observation(this.observationConvention,
						DefaultConvention.INSTANCE,
						() -> new MyCtx(), this.observationRegistry)
				.observe((context) -> {
					Message<?> reply = requestReply();
					context.setResponse(reply);
					return reply;
				});
```
and no any extra `if...else` around!

ttddyy:
Fixes #3537

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -485,19 +485,11 @@ public interface Observation extends ObservationView {
      * </ul>
      * @param runnable the {@link Runnable} to run
      */
-    @SuppressWarnings("unused")
     default void observe(Runnable runnable) {
-        start();
-        try (Scope scope = openScope()) {
+        observeWithContext((context) -> {
             runnable.run();
-        }
-        catch (Exception exception) {
-            error(exception);
-            throw exception;
-        }
-        finally {
-            stop();
-        }
+            return null; // returned value is ignored
+        });
     }
 
     default Runnable wrap(Runnable runnable) {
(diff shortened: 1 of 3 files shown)


micrometer-observation/src/test/java/io/micrometer/observation/ObservationTests.java
```diff
@@ -150,6 +153,62 @@ void scopedCheckedSpecificException() {
...
+    @Test
+    void observeWithFunction() {
+        CustomContext context = new CustomContext();
+        AtomicReference<CustomContext> contextHolder = new AtomicReference<>();
+        String result = new SimpleObservation("service", registry, context).observe((ctx) -> {
+            CustomContext customContext = (CustomContext) ctx;
+            customContext.hello();
+            contextHolder.set(customContext);
+            return "foo";
+        });
```
jonatan-ivanov, line 184: I'm not sure if it is very important but I would try to do this without initializing the `Context`. I guess that was Artem's main issue: he does not want to create a context if the observation is a noop.

```suggestion
        String result = new SimpleObservation("service", registry, CustomContext::new).observe((ctx) -> {
            CustomContext customContext = (CustomContext) ctx;
            customContext.setMessage("test");
            return "foo";
        });
```
And then verify if `customContext.getMessage()` equals to `"test"`.
For this, `CustomContext` needs a `setMessage`/`getMessage` on a `static` field (I'm not sure how ugly is that). Or you can get a reference to the observation, call `getContext` on it and do the verification against that.

shakuzen (approved):

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -587,13 +561,44 @@ default <T> Supplier<T> wrap(Supplier<T> supplier) {
...
+    @SuppressWarnings({ "unchecked" })
+    @Nullable
+    default <C extends Context, T> T observeWithContext(Function<C, T> function) {
         start();
         try (Scope scope = openScope()) {
-            return checkedCallable.call();
+            return function.apply((C) getContext());
         }
-        catch (Throwable error) {
+        catch (Exception error) {
```
jonatan-ivanov, line 601: Since its pair (`observeCheckedWithContext`) catches `Throwable`, should not we catch that here too?

jonatan-ivanov (approved):

Issue #3543, opened 2022-11-23
Fix logic to choose ObservationConvention


jonatan-ivanov:
There are two issues with the logic that chooses and applies the `ObservationConvention` on the `Observation`s.

1. The logic can choose and apply multiple conventions, also in one case it can mix them together
2. In one case the logic pics the wrong convention

Here's the old and the new decision matrix:

**OLD:**
```
1. custom     default     no pre-cofigured  -> custom
2. custom     default     pre-cofigured     -> custom on top of pre-cofigured
3. no custom  default     no pre-cofigured  -> default
4. no custom  default     pre-cofigured     -> default on top of pre-cofigured (*)
5. custom     no default  no pre-cofigured  -> custom
6. custom     no default  pre-cofigured     -> custom on top of pre-cofigured
7. no custom  no default  no pre-cofigured  -> local
8. no custom  no default  pre-cofigured     -> pre-cofigured
```
_(*): This logic is wrong: 1. they are mixed, 2. pre-cofigured should have choosen_

**NEW:**
```
1. custom     default     no pre-configured  -> custom
2. custom     default     pre-configured     -> custom (not really a valid case, use custom)
3. no custom  default     no pre-configured  -> default
4. no custom  default     pre-configured     -> pre-configured
5. custom     no default  no pre-configured  -> custom (providing default is recommended)
6. custom     no default  pre-configured     -> custom (providing default is recommended)
7. no custom  no default  no pre-configured  -> local names/tags will be used
8. no custom  no default  pre-configured     -> pre-configured
```

cc: @bclozel

PR #3544, opened 2022-11-23
Fix logic to choose ObservationConvention

jonatan-ivanov:
fixes gh-3543

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -127,6 +127,7 @@ public interface Observation extends ObservationView {
         return new SimpleObservation(name, registry, context == null ? new Context() : context);
     }
 
+    // @formatter:off
     /**
      * Creates but <b>does not start</b> an {@link Observation}. Remember to call
      * {@link Observation#start()} when you want the measurements to start. When the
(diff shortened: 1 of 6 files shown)


micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -380,8 +428,8 @@ default boolean isNoop() {
 
     /**
      * Adds an observation convention that can be used to attach key values to the
```
ttddyy, line 430: nit: change `Adds` to `Sets`.

Also, do we want to deprecate this method as we discussed offline?

The actual implementation only set the convention if it supports the context. Wondering should it also be documented?

jonatan-ivanov: Fixed in https://github.com/micrometer-metrics/micrometer/commit/3f0ca2f44b6b1fd8a343c1efd9292f63eb0d6f07, thanks for catching it.

I think we should deprecate it, but:
- I think it should happen in 1.11
- We need more discussions about the potential use-cases and consequences

I added a note about setting the convention in https://github.com/micrometer-metrics/micrometer/commit/22a81f0cf1ac3f21e3b0b1f93a2b8c5c079422e2 thank you for noticing it.

PR #3557, opened 2022-12-06
[Dynatrace] Log NaNs at DEBUG instead of WARNING

pirgeo:
When the object behind the weak reference of a Gauge meter is garbage collected, it will fall back to reporting NaN. The Dynatrace exporter would generate a warning log for each NaN generated that way. This bugfix changes the code to only log the first NaN at warning level (including some hints for the user) and all subsequent NaNs at debug level to prevent spamming the logs with lots of warnings.

Is it possible to get this into 1.9.x versions, and if so, is there anything in particular that I need to do to this PR?


arminru (approved):

pirgeo:
Rebased on https://github.com/micrometer-metrics/micrometer/tree/1.9.x and pointed the PR to that branch.

pirgeo:
@jonatan-ivanov Is there anything that needs to be done to move this PR forward?

shakuzen (approved):

PR #3559, opened 2022-12-08
Exposing extra ehCache metrics

guptasac-git:
Exposing ehCache metrics
-cacheExpiredCount
-cacheHitRatio
-writerQueueLength

Also, added below metrics for localOffHeap, localDisk, localHeap -hitCount
-missCount
-putCount
-putAddedCount
-putUpdatedCount
-removeCount


guptasac-git (commented):
Requesting for adding this

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/EhCache2Metrics.java
```diff
@@ -118,22 +118,26 @@ protected void bindImplementationSpecificMetrics(MeterRegistry registry) {
                 .tags(getTagsWithCacheName()).tags("result", "updated").description(DESCRIPTION_CACHE_PUTS_ADDED)
                 .register(registry);
 
+        FunctionCounter.builder("cache.expired.count", stats, StatisticsGateway::cacheExpiredCount)
+                .tags(getTagsWithCacheName()).description("Cache expired count").register(registry);
+
+        Gauge.builder("cache.hitRatio", stats, StatisticsGateway::cacheHitRatio).tags(getTagsWithCacheName())
```
shakuzen, line 124: We usually avoid registering metrics that you can derive from other metrics. Can't you get the hit ratio from the standard hit count and miss count metrics?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/EhCache2Metrics.java
```diff
@@ -190,6 +194,84 @@ private void recoveryTransactionMetrics(MeterRegistry registry) {
...
     }
 
+    private void localOffHeapMetrics(MeterRegistry registry) {
+        StatisticsGateway stats = getStats();
+
+        Gauge.builder("cache.local.offheap.size", stats, StatisticsGateway::getLocalOffHeapSizeInBytes)
+                .tags(getTagsWithCacheName()).description("Local off-heap size").baseUnit(BaseUnits.BYTES)
+                .register(registry);
+
+        Gauge.builder("cache.local.offheap.hitCount", stats, StatisticsGateway::localOffHeapHitCount)
```
shakuzen, line 204: Are these "counts" gauges or counters?

shakuzen (commented):
Thank you for the pull request. Sorry for the delay in reviewing. I left some review on the changes. In addition to that, this adds quite a few meters that we didn't have before. No one has requested these so far. Do you think these are generally useful for us to include by default for everyone? Or would these only be useful to some users in some situations and we should not include them by default with `EhCache2Metrics`?

PR #3562, opened 2022-12-09
Add metrics for the total number of application threads started in the JVM

lenin-jaganathan:
- generate a metric for total Threads started since the JVM start time. This is extremely useful to understand how an application is doing on thread creation and also understand the creation patterns. The existing metrics give the current status of threads but nothing about whether the threads are created newly or are existing ones. This metric will aid in such scenarios by taking deltas.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -180,17 +180,19 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
             CompositeData cd = (CompositeData) notification.getUserData();
             GarbageCollectionNotificationInfo notificationInfo = GarbageCollectionNotificationInfo.from(cd);
 
+            String gcName = notificationInfo.getGcName();
             String gcCause = notificationInfo.getGcCause();
             String gcAction = notificationInfo.getGcAction();
             GcInfo gcInfo = notificationInfo.getGcInfo();
             long duration = gcInfo.getDuration();
-            if (isConcurrentPhase(gcCause, notificationInfo.getGcName())) {
-                Timer.builder("jvm.gc.concurrent.phase.time").tags(tags).tags("action", gcAction, "cause", gcCause)
+            if (isConcurrentPhase(gcCause, gcName)) {
+                Timer.builder("jvm.gc.concurrent.phase.time").tags(tags)
+                        .tags("action", gcAction, "cause", gcCause, "gc_name", gcName)
                         .description("Time spent in concurrent phase").register(registry)
                         .record(duration, TimeUnit.MILLISECONDS);
             }
             else {
-                Timer.builder("jvm.gc.pause").tags(tags).tags("action", gcAction, "cause", gcCause)
+                Timer.builder("jvm.gc.pause").tags(tags).tags("action", gcAction, "cause", gcCause, "gc", gcName)
                         .description("Time spent in GC pause").register(registry)
                         .record(duration, TimeUnit.MILLISECONDS);
             }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -180,19 +180,17 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
             CompositeData cd = (CompositeData) notification.getUserData();
             GarbageCollectionNotificationInfo notificationInfo = GarbageCollectionNotificationInfo.from(cd);
 
-            String gcName = notificationInfo.getGcName();
             String gcCause = notificationInfo.getGcCause();
             String gcAction = notificationInfo.getGcAction();
             GcInfo gcInfo = notificationInfo.getGcInfo();
             long duration = gcInfo.getDuration();
-            if (isConcurrentPhase(gcCause, gcName)) {
-                Timer.builder("jvm.gc.concurrent.phase.time").tags(tags)
-                        .tags("action", gcAction, "cause", gcCause, "gc_name", gcName)
+            if (isConcurrentPhase(gcCause, notificationInfo.getGcName())) {
+                Timer.builder("jvm.gc.concurrent.phase.time").tags(tags).tags("action", gcAction, "cause", gcCause)
                         .description("Time spent in concurrent phase").register(registry)
                         .record(duration, TimeUnit.MILLISECONDS);
             }
             else {
-                Timer.builder("jvm.gc.pause").tags(tags).tags("action", gcAction, "cause", gcCause, "gc", gcName)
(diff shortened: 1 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -170,17 +170,19 @@ public void handleNotification(Notification notification, Object ref) {
...
+            String gcName = notificationInfo.getGcName();
             String gcCause = notificationInfo.getGcCause();
             String gcAction = notificationInfo.getGcAction();
             GcInfo gcInfo = notificationInfo.getGcInfo();
             long duration = gcInfo.getDuration();
-            if (isConcurrentPhase(gcCause, notificationInfo.getGcName())) {
-                Timer.builder("jvm.gc.concurrent.phase.time").tags(tags).tags("action", gcAction, "cause", gcCause)
+            if (isConcurrentPhase(gcCause, gcName)) {
+                Timer.builder("jvm.gc.concurrent.phase.time").tags(tags)
+                        .tags("action", gcAction, "cause", gcCause, "gc_name", gcName)
```
jonatan-ivanov, line 180: This is inconsistent to `gc` in `jvm.gc.pause` and I'm thinking if this should be a separate PR.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
```diff
@@ -66,6 +66,11 @@ public void bindTo(MeterRegistry registry) {
                 .description("The current number of live threads including both daemon and non-daemon threads")
                 .baseUnit(BaseUnits.THREADS).register(registry);
 
+        Gauge.builder("jvm.threads.started", threadBean, ThreadMXBean::getTotalStartedThreadCount).tags(tags)
```
jonatan-ivanov, line 69: Is there a daemon/live getter so that this can be consistent to `jvm.threads.daemon`/`jvm.threads.live`?

jonatan-ivanov, line 69: This should be monotonic and I guess what users are mostly interested in is not the total number of threads that has been started since the JVM started but the rate of them. So I think this must be a counter. Since there is no event you can use to increment a "normal" `Counter` but there is a (monothonic) method that gives you the current total, you can use a `FunctionCounter` instead of a `Gauge`, see the [docs](https://micrometer.io/docs/concepts#_function_tracking_counters).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
```diff
@@ -66,6 +66,11 @@ public void bindTo(MeterRegistry registry) {
...
+                .description(
+                        "The total number of threads created and also started since the Java virtual machine started")
```
jonatan-ivanov, line 71:
```suggestion
                        "The total number of threads started since the JVM started")
```

jonatan-ivanov (changes requested):
We should add tests for the changes in `JvmGcMetrics` but as I said I might move those to a separate PR.

jonatan-ivanov:
Could you please rebase this on `main`, that should solve the build failure I hope.

jonatan-ivanov:
I was thinking of Loom and the usefulness of this for virtual thread. Also if it is possible to distinguish between virtual and non-virtual threads.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
lenin-jaganathan: I see. How dumb of me to have contrasting names in consecutive conditional blocks 😟. I have no issues having this part of a separate PR.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
lenin-jaganathan: Makes sense. But still worried about Step Registry users like us. But it really makes sense to have this as a Counter.

lenin-jaganathan: Not sure what this means. Can you explain a bit more on this?

jonatan-ivanov: There are two types of threads in Java: daemon and non-daemon. See the previous few lines: `jvm.threads.daemon` and `jvm.threads.live`, see: https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html

jonatan-ivanov: Check step registries, for those this will return delta not cumulative.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
```diff
@@ -66,6 +67,11 @@ public void bindTo(MeterRegistry registry) {
                 .description("The current number of live threads including both daemon and non-daemon threads")
                 .baseUnit(BaseUnits.THREADS).register(registry);
 
+        FunctionCounter.builder("jvm.threads.started", threadBean, ThreadMXBean::getTotalStartedThreadCount).tags(tags)
+                .description(
+                        "The total number of application threads started in the JVM")
```
jonatan-ivanov, line 72:
```suggestion
                .description("The total number of application threads started in the JVM")
```

jonatan-ivanov:
The build is failing, could you please run `./gradlew format` to fix it?

lenin-jaganathan:
Yes on it @jonatan-ivanov . Having some trouble locally with switching branches. Will update this

shakuzen (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
lenin-jaganathan: Yeah. I am aware of that. But those OS implementations give us the current number of threads active. But this is more about the rate at which threads are getting created. And unfortunately, the native implementation doesn't differentiate between daemon and non-daemon for this method.

lenin-jaganathan:
@shakuzen By any chance will this metric be available in 1.10.x versions?

jonatan-ivanov:
@lenin-jaganathan Since this is a new feature, not really. 1.10.x is a maintenance  branch, we only release patch versions which can contain dependency upgrades and bugfixes but it is rare that they contain new features. This should go out early next week (hopefully on Monday) with `1.11.0-M2` and it should go GA in mid May with `1.11.0`.

lenin-jaganathan:
Thanks.

PR #3582, opened 2023-01-09
Azul Prime (Zing) C4 (GPGC) support

andrew-krasny:
Please consider the following change that will classify GPGC (C4) as generational and concurrent.
Thanks!

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -242,8 +242,18 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
     }
 
     private boolean isGenerationalGcConfigured() {
-        return ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
+        boolean result = false;
+
+        // Zing GPGC is always generational:
+        result |= ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
+                .map(MemoryPoolMXBean::getName).filter(name -> name.contains("GPGC")).findAny().isPresent();
+
+        // Keep original logic for other GCs:
+        // having more than 1 non-'tenured' pools is considered to be generational.
+        result |= ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
                 .map(MemoryPoolMXBean::getName).filter(name -> !name.contains("tenured")).count() > 1;
+
+        return result;
     }
 
     private static boolean isManagementExtensionsPresent() {
@@ -295,6 +305,13 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
                 put("partial gc", YOUNG);
                 put("global garbage collect", OLD);
                 put("Epsilon", OLD);
+                // GPGC
+                put("GPGC New", YOUNG); // old naming
+                put("GPGC Old", OLD); // old naming
+                put("GPGC New Cycles", YOUNG); // new naming
+                put("GPGC Old Cycles", OLD); // new naming
+                put("GPGC New Pauses", YOUNG); // new naming
+                put("GPGC Old Pauses", OLD); // new naming
             }
         };
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -242,18 +242,24 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
     }
 
     private boolean isGenerationalGcConfigured() {
-        boolean result = false;
-
-        // Zing GPGC is always generational:
-        result |= ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
-                .map(MemoryPoolMXBean::getName).filter(name -> name.contains("GPGC")).findAny().isPresent();
-
-        // Keep original logic for other GCs:
-        // having more than 1 non-'tenured' pools is considered to be generational.
-        result |= ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
-                .map(MemoryPoolMXBean::getName).filter(name -> !name.contains("tenured")).count() > 1;
-
-        return result;
+        // Zing GPGC is always generational and having more than one non-'tenured'
+        // pools is also considered to be generational.
+        int nonTenuredPools = 0;
+        for (MemoryPoolMXBean bean : ManagementFactory.getMemoryPoolMXBeans()) {
+            if (JvmMemory.isHeap(bean)) {
+                String name = bean.getName();
+                if (!name.contains("tenured")) {
+                    nonTenuredPools++;
+                    if (nonTenuredPools == 2) {
+                        return true;
+                    }
+                }
+                if (name.contains("GPGC")) {
+                    return true;
+                }
+            }
+        }
+        return false;
     }
 
     private static boolean isManagementExtensionsPresent() {
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
@@ -35,11 +35,13 @@ class JvmMemory {
     }
 
     static boolean isConcurrentPhase(String cause, String name) {
-        return "No GC".equals(cause) || "Shenandoah Cycles".equals(name) || "ZGC Cycles".equals(name);
+        return "No GC".equals(cause) || "Shenandoah Cycles".equals(name) || "ZGC Cycles".equals(name)
+                || (name.startsWith("GPGC") && !name.endsWith("Pauses"));
     }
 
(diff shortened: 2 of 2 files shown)


andrew-krasny:
@shakuzen,
Yes, correct! Sorry didn't know that this issue was filed ;)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -242,8 +242,18 @@ private boolean nonGenerationalGcShouldUpdateDataSize(String gcName) {
...
-        return ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
+        boolean result = false;
+
+        // Zing GPGC is always generational:
+        result |= ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
+                .map(MemoryPoolMXBean::getName).filter(name -> name.contains("GPGC")).findAny().isPresent();
+
+        // Keep original logic for other GCs:
+        // having more than 1 non-'tenured' pools is considered to be generational.
+        result |= ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
```
jonatan-ivanov, line 253: Instead of doing these tricks with the `result` and going through the stream twice, would not it be better to write the filter in a way that can handle both?
If my understanding is correct, something like this?

```java
return ManagementFactory.getMemoryPoolMXBeans().stream()
                .filter(JvmMemory::isHeap)
                .map(MemoryPoolMXBean::getName)
                .filter(name -> name.contains("GPGC") || !name.contains("tenured"))
                .findAny()
                .isPresent();
```

jonatan-ivanov (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
andrew-krasny: isPresent() is not enough, as the original code was looking for more than 1 !'tenured' beans.

So we can go with either something like this:

```
        // Zing GPGC is always generational and having more than 1 non-'tenured'
        // pools is also considered to be generational.
        int nonTenuredPools = 0;
        for (MemoryPoolMXBean bean : ManagementFactory.getMemoryPoolMXBeans()) {
            String name = bean.getName();
            if (!name.contains("tenured")) {
                nonTenuredPools++;
                if (nonTenuredPools == 2) {
                    return true;
                }
            }
            if (name.contains("GPGC")) {
                return true;
            }
        }
        return false;
```
or, slightly shorter, but little bit more cryptic and less efficient (still a single stream pass)
```
        // Zing GPGC is always generational and having more than 1 non-'tenured'
        // pools is also considered to be generational.
        return ManagementFactory.getMemoryPoolMXBeans().stream().filter(JvmMemory::isHeap)
                .map(MemoryPoolMXBean::getName)
                .mapToInt(name -> name.contains("GPGC") ? 2 : name.contains("tenured") ? 0 : 1)
                .sum() > 1;
```

jonatan-ivanov: Yepp, you are right, it needs more than 1.
I like the more readable (/less cryptic) version with the for loop. Looking at all three options, I think I would go with that. What do you think?

andrew-krasny: Agree with you, thanks! I'll update the diff

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -295,6 +311,13 @@ enum GcGenerationAge {
                 put("partial gc", YOUNG);
                 put("global garbage collect", OLD);
                 put("Epsilon", OLD);
+                // GPGC
+                put("GPGC New", YOUNG); // old naming
+                put("GPGC Old", OLD); // old naming
```
jonatan-ivanov, line 316: Just for the sake of documentation:
It seems that the new names were introduced in Azul Prime 22.12.0.0, release date: 2022-12-19 so they are still supported, see the [release notes](https://docs.azul.com/prime/release-notes#prime_stream_22_12_0_0).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -295,6 +311,13 @@ enum GcGenerationAge {
...
+                put("GPGC New Cycles", YOUNG); // new naming
```
jonatan-ivanov, line 317: Just for the sake of documentation:
This line (`GPGC New Cycles`) puzzled me a little. I built this branch locally and tested it. The meters + this map correspond what I saw in `MemoryPoolMXBean` except `GPGC New Cycles` is not listed in `getMemoryManagerNames`, see if I print out each `ManagementFactory.getMemoryPoolMXBeans()`:

```
GenPauseless New Gen (Heap memory) [GPGC New Pauses]
GenPauseless Old Gen (Heap memory) [GPGC Old Pauses]
GenPauseless Perm Gen (Heap memory) [GPGC Old Cycles]
CodeCache (Non-heap memory) [CodeCache]
Native Heap (Non-heap memory) [NativeHeap]
```

On the other hand, if I register a `NotificationListener`, I get these GC names (from GarbageCollectionNotificationInfo):
- `GPGC New Cycles`
- `GPGC New Pauses`
- `GPGC Old Cycles`
- `GPGC Old Pauses`

According to the [release notes](https://docs.azul.com/prime/release-notes#prime_stream_22_12_0_0) it is valid but I still find it weird as a GC name:
>New JMX MXBean metrics replace old metric name below `java.lang.GarbageCollector` to increase accuracy for GC monitoring added with [JDK-8265136](https://bugs.openjdk.org/browse/JDK-8265136): Previously, metric "GPGC New/Old" was providing a sum of GC pauses and concurrent GC duration. This metric is replaced by the following:
>- GPGC New/Old Cycles: duration time in ms of the concurrent GC which runs in parallel to application threads and is not stopping the application.
>- GPGC New/Old Pauses: GC pause time in ms.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
```diff
@@ -35,11 +35,13 @@ static Stream<MemoryPoolMXBean> getLongLivedHeapPools() {
     }
 
     static boolean isConcurrentPhase(String cause, String name) {
-        return "No GC".equals(cause) || "Shenandoah Cycles".equals(name) || "ZGC Cycles".equals(name);
+        return "No GC".equals(cause) || "Shenandoah Cycles".equals(name) || "ZGC Cycles".equals(name)
+                || (name.startsWith("GPGC") && !name.endsWith("Pauses"));
```
jonatan-ivanov, line 39: Ooops, shouldn't this be `"Cycles"` instead of `"Pauses"`?
This is used to record the time spent in concurrent phase and time spent in GC pause (`jvm.gc.concurrent.phase.time`, `jvm.gc.pause`). Based on the [release notes](https://docs.azul.com/prime/release-notes#prime_stream_22_12_0_0), I think this should be `"Cycles"`:
>GPGC New/Old Cycles: duration time in ms of the concurrent GC which runs in parallel to application threads and is not stopping the application.
GPGC New/Old Pauses: GC pause time in ms.

jonatan-ivanov (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
shakuzen: The condition now is starts with `GPGC` and does NOT end with `Pauses`, which would include the previous naming ("GPGC New/Old") as well as the new "GPGC New/Old Cycles". The problem with the previous naming is that it includes both concurrent and pause time. We're in a hard spot with how to count the previous naming. Either we count it as concurrent time or as pause time, but then whichever we count it as will be larger than it actually is (because it includes the other).

jonatan-ivanov: >does NOT end with

Arghhh, of course I did not see the negation. :D
One more thing about the old naming: even if the latest 11 and 17 builds are fixed by default, 8 is still using the old naming.

jonatan-ivanov (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
andrew-krasny: Indeed the motivation for these renaming was similar to what is described in JDK-8265136.
This inconsistence that you're pointing to is historical ;) But in this particular case, I believe, 'GPGC *' is what is used.

andrew-krasny:
@jonatan-ivanov, thank you for your comments and for taking time to look at this!

jonatan-ivanov:
@andrew-krasny Thank you very much for the contribution!
I added some docs and you as an `@author` to the file (I wasn't able to do it on this PR because edit for maintainers was disabled).

jeffreye:
Hi @andrew-krasny , the concurrent GC phase seems to be broken in the latest version. Meanwhile, there are new GC actions called `major GC` and `major GC pause`.

PR #3594, opened 2023-01-20
Add onMeterPublishedSuccess notification listener for Elastic Meter Registry

igdianov:
This PR adds support for elastic meter registry to notify registered listeners when a batch of meters have been successfully published to remote elastic server.

### DIFF
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
@@ -100,7 +100,7 @@ public class ElasticMeterRegistry extends StepMeterRegistry {
     @SuppressWarnings("deprecation")
     public ElasticMeterRegistry(ElasticConfig config, Clock clock) {
         this(config, clock, DEFAULT_THREAD_FACTORY,
-             new HttpUrlConnectionSender(config.connectTimeout(), config.readTimeout()), Collections.emptyList());
+                new HttpUrlConnectionSender(config.connectTimeout(), config.readTimeout()), Collections.emptyList());
     }
 
     /**
--- a/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
+++ b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java
@@ -58,6 +58,7 @@ public class ElasticMeterRegistry extends StepMeterRegistry {
 
     private static final ThreadFactory DEFAULT_THREAD_FACTORY = new NamedThreadFactory("elastic-metrics-publisher");
     static final DateTimeFormatter TIMESTAMP_FORMATTER = DateTimeFormatter.ISO_INSTANT;
+
     private final List<BiConsumer<ElasticMeterRegistry, List<Meter>>> meterPublishedSuccessListeners = new CopyOnWriteArrayList<>();
 
     private static final String ES_METRICS_TEMPLATE = "/_template/metrics_template";
@@ -432,7 +433,8 @@ public class ElasticMeterRegistry extends StepMeterRegistry {
             return this;
         }
 
-        public Builder onMeterPublishedSuccess(BiConsumer<ElasticMeterRegistry, List<Meter>> meterPublishedSuccessListener) {
+        public Builder onMeterPublishedSuccess(
+                BiConsumer<ElasticMeterRegistry, List<Meter>> meterPublishedSuccessListener) {
             meterPublishedSuccessListeners.add(meterPublishedSuccessListener);
             return this;
         }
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
What is the use-case for this feature?
Would it make sense to have a listener for failures too?
Would it make sense to have this in all push-based registries?

igdianov:
Thanks for asking. 

We want to be able to prune certian types of meters ie step timers and counters on successful push to avoid pushing them again on next push. These meters will have their values reset after push, so they are not useful to keep around.

jonatan-ivanov:
This is by design, values that are not changing are pushed because 0 has different meaning than not having the value. 0 means that the value did not change compared to the previous publish but the application is running and able to publish the metric. A missing value on the other hand means that the application is not running or does not have that metric. There is a difference and I would say they are useful at troubleshooting.

igdianov:
I understand the concept. Unfortunately, it is not possible to remove step timer and counter meters in any other way i. e. by polling meter registry, since the attempt to query the measurement will reset the meter value. We don't want to maintain the fork ourselves, so I made the contribution.

Please, also consider the testing and observability aspect of the proposed improvement since the push event notification makes it easier to see what is going on without causing the reset of measured values by queries before push happens.

jonatan-ivanov:
What I'm trying to say is that you should not remove Meters (because zero value has different meaning than not having the value). I think I still not understand why you want to remove them since they carry valuable information.

Though I think a mechanism to signal success and errors on `PushMeterRegistry` can be a good idea for other use-cases.
Let us discus this internally and get back to you, if we want to implement this, I think the scope should be every `PushMeterRegistry`.

Fyi: there is a bug that is not related to this issue but one way to fix it to introduce a way to query the current value of "step" `Meters` without resetting them.

igdianov:
@jonatan-ivanov I understand the rationale about keeping meters with zero values for tracking, but for our use case we don't want to keep them around once they are pushed to remote. 

We are using the framework to collect business metrics for business processes. Once the metric is collected and published, we want to stop tracking it. We could not use any existing api to do that, so we decided to extend the elastic registry to add the listeners. We also used the listeners api to implement tests.

I understand it may take time to implement this across all registries. In the meantime, please, consider to accept the contribution to accommodate our use case.

jonatan-ivanov:
>our use case we don't want to keep them around once they are pushed to remote.

Could you please let us know why? I would like to have a better understanding about the use-case.

>Once the metric is collected and published, we need to stop tracking it in the registry.

Could you please let us know why? I would like to have a better understanding about the use-case.

>In the meantime, please, consider to accept the contribution to accommodate our use case.

In order to accept this change we need two things:
1. A good use-case: right now what you are trying to do seems to go against the intended behavior and I would like to understand why are you trying to do this.
2. If the use-case/feature makes sense and it seems worthy to maintain it, we need to implement it for all the registries. I know it's a big change but I can help implementing it if you make your PR editable by maintainers

igdianov:
@jonatan-ivanov The main reason why we don't need to track some meters types once they are published is because all meters in our use case have high cardinality, so it does not makes any sense to leave them in the process memory after successfull push to keep pushing zero values to elastic again which consumes storage and costs.

The business processes and tasks are very long lived, i. e. 6 months, so, meter instances are created with a measurement derived from the state of the task attributes on event, i. e. duration of the completed task by user with additional tags to track business context and pushed to remote elastic. The elasticsearch is used to run aggregation queries to get analytics data. 

Since the framework allows to add and remove meters, we as users also needed have full control of the fact when meters have been published to observe and decide what to do with them. The project was missing this feature, so, I decided to make the improvements and submit our contribution  which may be helpful to community.

jonatan-ivanov:
>The main reason why we don't need to track some meters types once they are published is because all meters in our use case have high cardinality

If I'm not mistaken it's not that you don't need to but you can't really (because of high cardinality). We are not recommending the usage of high cardinality tags on metrics (see [this post](https://develotters.com/posts/high-cardinality/)), instead we are recommending the usage of a concept that was designed to handle high cardinality data (e.g.: logging, distributed tracing, etc). Btw this is pretty easy with the new Observation API.

>Since the framework allows to add and remove meters, we as users also needed have full control of the fact when meters have been published to observe and decide what to do with them

I'm not sure there is such a connection between these two, I feel this is only valid in your context (high cardinality tags) but not in general.

Also, this feature (getting notified by successful/failed push) was never requested by anyone else as fas as I know (I might be mistaken) so now that your use-case is known let me discuss this with other maintainers.

shakuzen:
@igdianov Are you only wanting to publish once before removing the meter? You could also remove these meters one step interval after recording to have basically the same effect. They are step meters, so if a publish fails, the value isn't going to be the same on the next publish anyway.

Put another way, it sounds like you want to use Micrometer to publish unaggregated (raw) recordings with high cardinality tags, and then aggregate that raw data into metrics on the backend. Micrometer's registry implementations were not designed for this use case*[^1], but you could probably make a custom registry implementation that does this.

Are you aggregating a lot of high cardinality data in the backend without any issues? Without knowing more about the context, it sounds like using distributed tracing for the high cardinality tags and limiting the metrics to low cardinality tags may make things match up better with the intended design. The new `Observation` API enables specifying high vs low cardinality tags, and you can register/create handlers to do whatever you want with the data recorded. We have default implementations of metrics and tracing handlers.

Another problem I see in the proposed implementation is that we don't track the mapping between a Meter and the time series published to a backend. The callback as is will only be called when all time series in a batch were successful. If even one time series had a problem, the success callback would not be called for any of the Meters in the same batch.

[^1]: The statsd registry does publish individual recordings but the statsd daemon does the aggregation before persisting data to a metrics backend.

jonatan-ivanov:
Based on this discussion let me close this. Please let us know if you want us to reopen.

PR #3596, opened 2023-01-20
[Dynatrace registry v1] Use header instead of query parameter

pirgeo:
In the Dynatrace v1 exporter, the API token is passed as a query parameter in the URL. On certain occasions, a misconfiguration can lead to the URL being printed in error logs. In the case where the API token is part of the URL as a query parameter, it can be leaked to the logs. We addressed that in the past (https://github.com/micrometer-metrics/micrometer/pull/3484) by redacting the token in the logs. However, we decided to update it now to use the API token in an Authorization header instead. This way, the token is no longer part of the URL, and can therefore not be leaked that way.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceExporterV1.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceExporterV1.java
@@ -67,17 +67,22 @@ public class DynatraceExporterV1 extends AbstractDynatraceExporter {
 
     private final NamingConvention namingConvention;
 
+    private static final String AUTHORIZATION_HEADER_KEY = "Authorization";
+
+    private final String authorizationHeaderValue;
+
     public DynatraceExporterV1(DynatraceConfig config, Clock clock, HttpSender httpClient) {
         super(config, clock, httpClient);
 
         this.customMetricEndpointTemplate = config.uri() + "/api/v1/timeseries/";
         this.namingConvention = new DynatraceNamingConvention(NamingConvention.dot, DynatraceApiVersion.V1);
+
+        this.authorizationHeaderValue = String.format("Api-Token %s", config.apiToken());
     }
 
     @Override
     public void export(List<Meter> meters) {
-        String customDeviceMetricEndpoint = config.uri() + "/api/v1/entity/infrastructure/custom/" + config.deviceId()
-                + "?api-token=" + config.apiToken();
+        String customDeviceMetricEndpoint = config.uri() + "/api/v1/entity/infrastructure/custom/" + config.deviceId();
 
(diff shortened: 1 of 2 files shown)


arminru (approved):

jonatan-ivanov:
@pirgeo Thanks for the contribution!
Since from the end-user perspective this "only" changes some log messages and makes things more secure, what do you think rebasing this to `1.9.x`? (As far as I remember this was fixed in `1.8.x` but that line is not supported anymore.)

pirgeo:
Great point, @jonatan-ivanov! I rebased the commits to the `1.9.x` branch and targeted it. These commits will be merged back into `main` by the maintainers, right? Thanks!

shakuzen:
> These commits will be merged back into `main` by the maintainers, right?

Yes. We always merge forward commits from oldest maintenance branch to newest.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v1/DynatraceExporterV1.java
```diff
@@ -67,17 +67,22 @@ public class DynatraceExporterV1 extends AbstractDynatraceExporter {
...
+
+    private final String authorizationHeaderValue;
+
     public DynatraceExporterV1(DynatraceConfig config, Clock clock, HttpSender httpClient) {
         super(config, clock, httpClient);
 
         this.customMetricEndpointTemplate = config.uri() + "/api/v1/timeseries/";
         this.namingConvention = new DynatraceNamingConvention(NamingConvention.dot, DynatraceApiVersion.V1);
+
+        this.authorizationHeaderValue = String.format("Api-Token %s", config.apiToken());
```
shakuzen, line 80: Perhaps it is intentional, but by reading `config.apiToken()` only in the constructor, if the config is backed by a dynamic configuration source, updates will not have any effect until a new instance of this is constructed, which I imagine only normally happens on startup of applications. Previously, we would call `config.apiToken()` on each export, allowing the possibility for dynamic updates of the configuration for API token without restarting the application.

pirgeo: Ah, you are right. I thought I was very clever by just reading it once, but it changes the behavior, and users might be relying on the original behavior (token rotation while running, etc.). I updated it, so it is read on export again. Thanks for catching this!

shakuzen:
Looks good to me. Have you had a chance to test an application using these changes yet, just to integration test that nothing surprisingly breaks with this?

pirgeo:
Yes, I did run a few manual tests internally. I ran the included sample app, once at the `1.9.x` commit and once at the commit that represents this branch. The exported data from both were accepted, and I could not tell any difference between them.

PR #3598, opened 2023-01-23
Support Apache HttpClient 5

shakuzen:
Copied and adapted from the existing instrumentation for Apache HttpClient 4. This is a follow-up to gh-3009 that started the work. This needed to be redone to incorporate the Observation API support that was added since then.

~TODO: Naming of copied classes: should we keep the same names or add `Hc5` (or similar) to the names to avoid confusion with the Apache HC 4 classes when choosing the import?~
We'll keep the names the same to make the parity more clear. IDEs and compilers will alert users if they try to use the wrong class (e.g. by importing the wrong one for the version they are using).

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/hc5/MicrometerHttpClientInterceptor.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/hc5/MicrometerHttpClientInterceptor.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2020 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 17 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/hc5/MicrometerHttpClientInterceptor.java
```diff
@@ -0,0 +1,99 @@
...
+    private final HttpResponseInterceptor responseInterceptor;
+
+    /**
+     * Create a {@code MicrometerHttpClientInterceptor} instance.
+     * @param meterRegistry meter registry to bind
+     * @param uriMapper URI mapper to create {@code uri} tag
+     * @param extraTags extra tags
+     * @param exportTagsForRoute whether to export tags for route
+     */
+    public MicrometerHttpClientInterceptor(MeterRegistry meterRegistry, Function<HttpRequest, String> uriMapper,
```
jonatan-ivanov, line 66: Do we need this and `MicrometerHttpRequestExecutor` too?
Shouldn't this use the Observation API too?

jonatan-ivanov (approved):

jonatan-ivanov:
I guess this fixes https://github.com/micrometer-metrics/micrometer/issues/2513
(which I'm not able to link to this PR).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/hc5/MicrometerHttpClientInterceptor.java
shakuzen: This is for async clients, `MicrometerHttpRequestExecutor` is for the "classic" synchronous HTTP client.
Yes, it should use the Observation API, but we didn't update it for Apache HC 4, so this is just copied what is there for now.

Issue #3546, opened 2022-11-25
OTLP registry custom header support


stse:
**Please describe the feature request.**
Provide configuration properties to provide custom headers e.g. `API-Key` which are sent whenever metrics are posted to the OTLP endpoint.

Having something like this

```yaml
management:
  otlp:
    metrics:
      export:
        step: 1s
        url: https://otlp.eu01.nr-data.net/v1/metrics
        headers:
          API-Key: otlp-endpoint-api-key
```

so that these headers are sent with each request may be this way

```java
var builder = this.httpSender.post(this.config.url());
for (final var header : this.config.headers()) {
  builder = builder.withHeader(header.key(), header.value());
}
builder.withContent("application/x-protobuf", request.toByteArray()) .send();
```

see https://github.com/micrometer-metrics/micrometer/blob/029a887eb86476036d28527ebda890289161623a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java#L108

Another solution would be to use the Open Telemetry SDK and allow to provide custom MetricExporter beans. This works very well already for tracing. Here I can provide custom SpanExporter beans.

**Rationale**
Having direct OTLP support is great but in order to connect with some OTLP endpoints, e.g. [NewRelics OTLP endpoints](https://docs.newrelic.com/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/opentelemetry-setup/) a header containing an API key has to be sent with each request.

shakuzen:
Thank you for opening the issue and describing the use case. With other MeterRegistry implementations, such a need would either be a first class concept (API key) and supported in the Config, or you would be able to customize the `HttpSender` that is used to achieve the same thing. Unfortunately neither of those are available for OtlpMeterRegistry right now. Generic customer header support sounds a bit too generic to me, but maybe it makes sense. In the initial implementation we intentionally did not expose the `HttpSender` in the API because OTLP can be sent more ways than HTTP. One option is to introduce a Builder for the registry that allows optionally customizing the `HttpSender`. I'll discuss with the team.

stse:
[...]

PR #3600, opened 2023-01-25
Configurable headers for OTLP registry

shakuzen:
OTLP collectors may require arbitrary headers, for example for authentication. This allows configuring headers to send with metrics via `OtlpConfig` or environment variables [standard to the OTLP exporter](https://opentelemetry.io/docs/reference/specification/protocol/exporter/#specifying-headers-via-environment-variables).

Resolves gh-3546

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
@@ -113,7 +113,7 @@ public interface OtlpConfig extends PushRegistryConfig {
             headersString = "".equals(headersString) ? metricsHeaders : headersString + "," + metricsHeaders;
         }
 
-        String[] keyvalues = "".equals(headersString) ? new String[] {} : headersString.split(",");
+        String[] keyvalues = Objects.equals(headersString, "") ? new String[] {} : headersString.split(",");
 
         return Arrays.stream(keyvalues).map(String::trim)
                 .filter(keyvalue -> keyvalue.length() > 2 && keyvalue.indexOf('=') > 0)
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
@@ -86,6 +86,41 @@ public interface OtlpConfig extends PushRegistryConfig {
         return resourceAttributes;
     }
 
+    /**
+     * Additional headers to send with exported metrics. This may be needed for
+     * authorization headers, for example.
+     * <p>
+     * By default, headers will be loaded from {@link #get(String)}. If that is not set,
+     * they will be taken from the environment variables
+     * {@code OTEL_EXPORTER_OTLP_HEADERS} and {@code OTEL_EXPORTER_OTLP_METRICS_HEADERS}.
+     * The header key-value pairs are expected to be in a comma-separated list in the
+     * format {@code key1=value1,key2=value2}. If a header is set in both
+     * {@code OTEL_EXPORTER_OTLP_HEADERS} and {@code OTEL_EXPORTER_OTLP_METRICS_HEADERS},
+     * the header in the latter will overwrite the former.
+     * @return a map of the headers' key-value pairs
+     * @see <a href=
+     * "https://opentelemetry.io/docs/reference/specification/protocol/exporter/#specifying-headers-via-environment-variables">OTLP
+     * Exporer headers configuration</a>
+     */
+    default Map<String, String> headers() {
+        Map<String, String> env = System.getenv();
+        String headersString = getString(this, "headers").orElse(null);
+
+        if (headersString == null) {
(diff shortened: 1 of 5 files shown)


implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
```diff
@@ -33,4 +40,40 @@ void resourceAttributesInputParsing() {
...
+    @Test
+    void headersEmptyishInputParsing() {
+        Stream<OtlpConfig> configs = Stream.of(k -> null, k -> "", k -> "  ", k -> " ,");
+        configs.forEach(config -> assertThat(config.headers()).isEmpty());
+    }
+
+    @Test
+    void headersConfigTakesPrecedenceOverEnvVars() throws Exception {
+        OtlpConfig config = k -> "header1=value1";
+        withEnvironmentVariable("OTEL_EXPORTER_OTLP_HEADERS", "header2=value")
```
shakuzen, line 52: I found the [system-stubs](https://github.com/webcompere/system-stubs) library, which allows us to test with different environment variables. We should update and add tests for the `resourceAttributes` config - it also depends on environment variables.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -86,6 +86,41 @@ default Map<String, String> resourceAttributes() {
...
+     * {@code OTEL_EXPORTER_OTLP_HEADERS} and {@code OTEL_EXPORTER_OTLP_METRICS_HEADERS},
+     * the header in the latter will overwrite the former.
+     * @return a map of the headers' key-value pairs
+     * @see <a href=
+     * "https://opentelemetry.io/docs/reference/specification/protocol/exporter/#specifying-headers-via-environment-variables">OTLP
+     * Exporer headers configuration</a>
+     */
+    default Map<String, String> headers() {
+        Map<String, String> env = System.getenv();
+        String headersString = getString(this, "headers").orElse(null);
```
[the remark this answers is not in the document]
shakuzen: This doesn't make sense unless we are using ErrorProne, which we currently are not. I would not be opposed to considering adding it. It should be a separate decision from this pull request, though. /cc @jonatan-ivanov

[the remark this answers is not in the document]
shakuzen: @sonatype-lift ignore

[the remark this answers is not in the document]
jonatan-ivanov: Yeah, let's separate them.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -86,6 +87,41 @@ default Map<String, String> resourceAttributes() {
...
+                                                                                       // headers
+            String metricsHeaders = env.getOrDefault("OTEL_EXPORTER_OTLP_METRICS_HEADERS", "").trim();
+            headersString = Objects.equals(headersString, "") ? metricsHeaders : headersString + "," + metricsHeaders;
+        }
+
+        String[] keyvalues = Objects.equals(headersString, "") ? new String[] {} : headersString.split(",");
+
+        return Arrays.stream(keyvalues).map(String::trim)
+                .filter(keyvalue -> keyvalue.length() > 2 && keyvalue.indexOf('=') > 0)
+                .collect(Collectors.toMap(keyvalue -> keyvalue.substring(0, keyvalue.indexOf('=')).trim(),
```
jonatan-ivanov, line 121: I'm not sure if it is better but one alternative of doing `indexOf` and `substring` twice would be doing a `.map(keyvalue -> .split("="))` before `collect` and use the two elements of the array in it.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
```diff
@@ -33,4 +40,40 @@ void resourceAttributesInputParsing() {
         assertThat(config.resourceAttributes()).containsEntry("k", "v").containsEntry("a", "b").hasSize(2);
     }
 
+    @Test
+    void headersEmptyishInputParsing() {
```
jonatan-ivanov, line 44: `Emptyish`? :)

jonatan-ivanov (approved):

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
shakuzen: Maybe we shouldn't support it, but a value could have `=` in it, which is why I went with the `indexOf` approach to get the first `=` only. We could use `split("=", 2)` to achieve the same effect, but it didn't feel necessarily less complicated/indirect to me. The current implementation matches what we're doing in the `resourceAttributes` config also. If we do decide to change the approach, we should probably unify it in both places.

PR #3605, opened 2023-01-27
Add ARM64 epoll for linux

taer:
I believe this is all that is needed to support UDS on ARM linux

### DIFF
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -10,6 +10,7 @@ dependencies {
         // We do not use HTTP modules; exclude to avoid false positive CVE reports
         exclude module: 'netty-codec-http'
     }
+    implementation(group: 'io.netty', name: 'netty-transport-native-epoll', classifier: 'linux-aarch_64')
 
     testImplementation project(':micrometer-test')
     testImplementation 'io.projectreactor:reactor-test'
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -11,6 +11,7 @@ dependencies {
         exclude module: 'netty-codec-http'
     }
     implementation(group: 'io.netty', name: 'netty-transport-native-epoll', classifier: 'linux-aarch_64')
+    implementation(group: 'io.netty', name: 'netty-transport-native-epoll', classifier: 'linux-x86_64')
 
     testImplementation project(':micrometer-test')
     testImplementation 'io.projectreactor:reactor-test'


taer:
The shaded jar now has this in it
```
micrometer/implementations/micrometer-registry-statsd/build/libs/META-INF/native on  addArm64Lib [$] using ☁️ dev-PowerUser 
➜ ls -l
total 336
-rw-rw-r-- 1 user user  74876 Jan 12 17:04 libio_micrometer_shaded_netty_resolver_dns_native_macos_aarch_64.jnilib
-rw-rw-r-- 1 user user  58104 Jan 12 16:55 libio_micrometer_shaded_netty_resolver_dns_native_macos_x86_64.jnilib
-rw-rw-r-- 1 user user 103728 Jan 12 14:36 libio_micrometer_shaded_netty_transport_native_epoll_aarch_64.so
-rw-rw-r-- 1 user user  94722 Jan 12 14:40 libio_micrometer_shaded_netty_transport_native_epoll_x86_64.so

```

taer:
I based on 1.9 since that was what I was using, but can rebase this on anything needed

jonatan-ivanov:
The branch is a good question (is this a feature or a bugfix? 😄).
I think if we decide to go with `1.9`, we might also want to back-port this (I can take care about it): https://github.com/micrometer-metrics/micrometer/pull/3139

@shakuzen What do you think? Should we go with `1.9` and also back-port the dns libs?

@taer Could you please double check the jar that you built and tested?
If you build it against `1.9.x`, you should have `epoll-x86` and `dns-macos-x86` (+`epoll_aarch_64` that you added) but in your list you also have `dns-macos-aarch_64` which tells me that your build was either against `1.10.x` or `main` since `1.9.x` should not have that.

implementations/micrometer-registry-statsd/build.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
         // We do not use HTTP modules; exclude to avoid false positive CVE reports
         exclude module: 'netty-codec-http'
     }
+    implementation(group: 'io.netty', name: 'netty-transport-native-epoll', classifier: 'linux-aarch_64')
```
jonatan-ivanov, line 13: In order to make this explicit, could you please also add the x86 variant?
(we did the same with dns-macos)
```suggestion
    implementation(group: 'io.netty', name: 'netty-transport-native-epoll', classifier: 'linux-aarch_64')
    implementation(group: 'io.netty', name: 'netty-transport-native-epoll', classifier: 'linux-x86_64')
```

jonatan-ivanov (changes requested):

taer:
> @taer Could you please double check the jar that you built and tested? If you build it against `1.9.x`, you should have `epoll-x86` and `dns-macos-x86` (+`epoll_aarch_64` that you added) but in your list you also have `dns-macos-aarch_64` which tells me that your build was either against `1.10.x` or `main` since `1.9.x` should not have that.

I had originally did it against master. And didn't clean.. This is the current one after a clean
```
-rw-rw-r-- 1 robert robert  58104 Dec 12 15:14 libio_micrometer_shaded_netty_resolver_dns_native_macos_x86_64.jnilib
-rw-rw-r-- 1 robert robert 103728 Dec 12 13:19 libio_micrometer_shaded_netty_transport_native_epoll_aarch_64.so
-rw-rw-r-- 1 robert robert  94722 Dec 12 13:21 libio_micrometer_shaded_netty_transport_native_epoll_x86_64.so

```

taer:
> The branch is a good question (is this a feature or a bugfix? smile).

Just something I just ran into. :)   I haven't tested it yet.. I know very little about native in java.

I'll try a quick test locally. See if it loads. The part that wasn't great was this didn't fail on ARM.. it just didn't work. Should there be some sort of error message or something in this case? Maybe there was a netty yelling on debug that I missed

jonatan-ivanov (approved):

jonatan-ivanov:
>I'll try a quick test locally. See if it loads. The part that wasn't great was this didn't fail on ARM.. it just didn't work. Should there be some sort of error message or something in this case? Maybe there was a netty yelling on debug that I missed

I'm not a big Netty wizard but as far as I know, if the native lib is not available, Netty should fallback to its Java implementation. This might not be the case of UDS (should be TCP/UDP under the hood though), it might worth creating an issue in the Netty repo if this is not an issue, a log message indeed would be nice.

taer:
Either way, I build this branch and deployed it, and it found the lib and reports to datadog! Looks happy and good to me

jonatan-ivanov:
@taer Thank you for the contribution!

PR #3615, opened 2023-02-01
[Dynatrace v2] Ensure synchronization when taking snapshots

pirgeo:
We found a timing bug that can appear when using the Dynatrace v2 exporter, usually only when recording data at high volumes. In previous versions, recording a snapshot was not a synchronized operation. That means that it was possible to write data to a DynatraceSummary, while the snapshot was being created. In very rare cases this can lead to invalid data. This PR moves the synchronization while taking a snapshot to the innermost level, and ensures that no writes can happen while a Snapshot is produced.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/types/DynatraceTimer.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/types/DynatraceTimer.java
@@ -55,9 +55,13 @@ public final class DynatraceTimer extends AbstractTimer implements DynatraceSumm
         }
     }
 
+    /**
+     * @deprecated see {@link DynatraceSummarySnapshotSupport#hasValues()}.
+     */
     @Override
+    @Deprecated
     public boolean hasValues() {
-        return count() > 0;
+        return summary.getCount() > 0;
     }
 
     @Override
@@ -67,7 +71,7 @@ public final class DynatraceTimer extends AbstractTimer implements DynatraceSumm
 
     @Override
     public DynatraceSummarySnapshot takeSummarySnapshot(TimeUnit unit) {
-        return new DynatraceSummarySnapshot(min(unit), max(unit), totalTime(unit), count());
+        return convertIfNecessary(unit, summary.takeSummarySnapshot());
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


pirgeo:
@shakuzen or @jonatan-ivanov, could you take a look? I would love to get this into the next patch release, if possible 😄

jonatan-ivanov (approved):

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/types/DynatraceTimer.java
```diff
@@ -67,7 +71,7 @@ public DynatraceSummarySnapshot takeSummarySnapshot() {
 
     @Override
     public DynatraceSummarySnapshot takeSummarySnapshot(TimeUnit unit) {
-        return new DynatraceSummarySnapshot(min(unit), max(unit), totalTime(unit), count());
+        return convertIfNecessary(unit, summary.takeSummarySnapshot());
```
jonatan-ivanov, line 74:
```suggestion
        return convertIfNecessary(summary.takeSummarySnapshot(), unit);
```

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/types/DynatraceTimer.java
```diff
@@ -77,9 +81,17 @@ public DynatraceSummarySnapshot takeSummarySnapshotAndReset() {
 
     @Override
     public DynatraceSummarySnapshot takeSummarySnapshotAndReset(TimeUnit unit) {
-        DynatraceSummarySnapshot snapshot = takeSummarySnapshot(unit);
-        summary.reset();
-        return snapshot;
+        return convertIfNecessary(unit, summary.takeSummarySnapshotAndReset());
```
jonatan-ivanov, line 84:
```suggestion
        return convertIfNecessary(summary.takeSummarySnapshotAndReset(), unit);
```

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/types/DynatraceTimerTest.java
```diff
@@ -159,6 +135,48 @@ void testDifferentTimeUnits() {
...
+    @Test
+    void testConvertIfNecessary() {
+        TimeUnit unit = TimeUnit.MILLISECONDS;
+        DynatraceTimer timer = new DynatraceTimer(ID, CLOCK, DISTRIBUTION_STATISTIC_CONFIG, PAUSE_DETECTOR, unit);
+
+        DynatraceSummarySnapshot snapshot = new DynatraceSummarySnapshot(1000., 2000., 3000., 2);
+        DynatraceSummarySnapshot unconverted = timer.convertIfNecessary(unit, snapshot);
+        assertThat(unconverted).isEqualTo(snapshot);
+
+        DynatraceSummarySnapshot converted = timer.convertIfNecessary(TimeUnit.SECONDS, snapshot);
```
jonatan-ivanov, line 147:
```suggestion
        DynatraceSummarySnapshot unconverted = timer.convertIfNecessary(snapshot, unit);
        assertThat(unconverted).isEqualTo(snapshot);

        DynatraceSummarySnapshot converted = timer.convertIfNecessary(snapshot, TimeUnit.SECONDS);
```

jonatan-ivanov (commented):
Since I applied the parameter change suggestion, the build is broken and I can't fix it on the branch but I think I can fix it with applying these suggestions. :)

jonatan-ivanov (approved):

jonatan-ivanov:
@pirgeo Thank you for fixing this!
Could you please do two more things?

1. Could you please add a note in the description that with this change recordings can be blocked during publishing for a very small amount of time while we get and reset the values for the snapshot?
2. Could you please add some example error message that Dynatrace will produce if this happens so users who get it and search on it might discover this issue?

pirgeo:
@jonatan-ivanov done! Should I send another PR for adding the missing test that you pointed out?

Issue #3560, opened 2022-12-08
Flooded with "Could not serialize meter XXX: Value was NaN" with Dynatrace plugin


szpak:
**Describe the bug**
Dynatrace exporter prints a lot of WARNs with `Could not serialize meter kafka.xxx: Value was NaN` meters.

**Environment**

I added micrometer-registry-dynatrace to my Spring Boot application to have the additional metrics exported to Dynatrace.

The pod is instrumented by the OneAgent and the exporter is able to detect that:

```
13:03:03.317  INFO 1 --- [           main] i.m.dynatrace.DynatraceMeterRegistry     : Exporting to Dynatrace metrics API v2
13:03:03.321  INFO 1 --- [           main] i.m.dynatrace.v2.DynatraceExporterV2     : Exporting to endpoint https://xxx.yyy/zzz/www/api/v2/metrics/ingest
13:03:03.385  INFO 1 --- [           main] i.m.c.instrument.push.PushMeterRegistry  : publishing metrics for DynatraceMeterRegistry every 1m
```

However, later on, after the Kafka consumer initialization, I see:

```
13:04:00.034  WARN 1 --- [trics-publisher] i.m.dynatrace.v2.DynatraceExporterV2     : Could not serialize meter kafka.consumer.coordinator.partition.lost.latency.max: Value was NaN.
13:04:00.036  WARN 1 --- [trics-publisher] i.m.dynatrace.v2.DynatraceExporterV2     : Could not serialize meter kafka.consumer.coordinator.partition.lost.latency.avg: Value was NaN.
13:04:00.049  WARN 1 --- [trics-publisher] i.m.dynatrace.v2.DynatraceExporterV2     : Could not serialize meter kafka.consumer.reauthentication.latency.avg: Value was NaN.
13:04:00.049  WARN 1 --- [trics-publisher] i.m.dynatrace.v2.DynatraceExporterV2     : Could not serialize meter kafka.consumer.reauthentication.latency.max: Value was NaN.
13:04:00.078  WARN 1 --- [trics-publisher] i.m.dynatrace.v2.DynatraceExporterV2     : Could not serialize meter kafka.consumer.coordinator.partition.lost.latency.max: Value was NaN.
13:04:00.081  WARN 1 --- [trics-publisher] i.m.dynatrace.v2.DynatraceExporterV2     : Could not serialize meter kafka.consumer.coordinator.partition.lost.latency.avg: Value was NaN.
[...]

PR #3616, opened 2023-02-02
Use Meter.Id for logging in DynatraceExporterV2

izeye:
This PR changes to use `Meter.Id` for logging in `DynatraceExporterV2` as it might help to understand #3560 or similar a bit better.


shakuzen:
Heads up to @pirgeo about this change.

shakuzen (approved):

pirgeo:
What additional info will be printed that way? I assume tags are included, right?

shakuzen:
Yes, it will include the meter name and tags:
https://github.com/micrometer-metrics/micrometer/blob/4c00160d34d490e2599002a6e8e5d1a7405160ce/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java#L346-L349

PR #3619, opened 2023-02-06
Added ability to define percentiles for OkHttpMetricsEventListener

jasonfagerberg-toast:
The default behavior seems to be to only create the 95 percentile

At Toast we mostly use 95 & 99 percentiles but I figured others may want this feature as well

Issue: https://github.com/micrometer-metrics/micrometer/issues/3620


jonatan-ivanov:
Thanks for the PR!
Since you can configure this on the `DistributionConfig`, I'm closing this PR, please let us know if I missed something and you want us to reopen. You can see the details on https://github.com/micrometer-metrics/micrometer/issues/3620

Issue #3623, opened 2023-02-08
LogbackMetrics does not protect against StackOverflowError in arbitrary Counter implementations


jonatan-ivanov:
There can be cases where incrementing a `Counter` can result in emitting log events. If that happens, the `MetricsTurboFilter` in `LogbackMetrics` can get into an infinite loop since it is triggered by emitting log events and increments a `Counter`. The scenario is the following:
1. Somebody emits a log event (e.g.: `LOGGER.info("test");`)
2. `MetricsTurboFilter` gets notified about it and increments a `Counter`
3. Incrementing the `Counter` somehow triggers emitting another log event
4. `MetricsTurboFilter` gets notified about it and increments a `Counter`
5. [...]

One real-life example to this is through using Prometheus Exemplars when incrementing a `PrometheusCounter` can invoke an `ExemplarSampler`. Since Micrometer has no control over the implementation of the `ExemplarSampler` and it can emit log events, we can get into the above scenario having an infinite loop, see: https://github.com/spring-projects/spring-boot/issues/33971

One solution to this can be ignoring the log events triggered by each meter interaction (or the effective [decision making logic](https://github.com/micrometer-metrics/micrometer/blob/827b3d9683082af4926a535ee096191939f42049/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java#L188-L208)) in `MetricsTurboFilter`

Issue #411, opened 2018-02-10
StackoverflowError while using micrometer-spring-legacy alongwith StatsD registry


aarya23:
getting into recursion as Counter.increment call results in **MetricsTurboFilter** invoking **Counter.increment** again.

Additional configuration in the springboot app:

  ```
<dependency>
            <groupId>io.micrometer</groupId>
            <artifactId>micrometer-spring-legacy</artifactId>
            <version>1.0.0-rc.7</version>
        </dependency>

 <dependency>
            <groupId>io.micrometer</groupId>
            <artifactId>micrometer-registry-statsd</artifactId>
            <version>1.0.0-rc.7</version>
        </dependency>

        <reactor.version>3.1.0.RELEASE</reactor.version>
        <spring.boot.version>1.5.4.RELEASE</spring.boot.version>
```

**StackoverlfowError:**

	at io.micrometer.core.instrument.Counter.increment(Counter.java:30)
	at io.micrometer.core.instrument.binder.logging.MetricsTurboFilter.decide(LogbackMetrics.java:102)
	at ch.qos.logback.classic.spi.TurboFilterList.getTurboFilterChainDecision(TurboFilterList.java:49)
	at ch.qos.logback.classic.LoggerContext.getTurboFilterChainDecision_0_3OrMore(LoggerContext.java:269)
	at ch.qos.logback.classic.Logger.filterAndLog_0_Or3Plus(Logger.java:373)
	at ch.qos.logback.classic.Logger.debug(Logger.java:482)
	at reactor.util.Loggers$Slf4JLogger.debug(Loggers.java:213)
	at reactor.core.publisher.Operators.onNextDropped(Operators.java:282)
	at reactor.core.publisher.UnicastProcessor.onNext(UnicastProcessor.java:317)
	**at io.micrometer.statsd.StatsdCounter.increment(StatsdCounter.java:43)**
	at io.micrometer.core.instrument.composite.CompositeCounter.lambda$increment$0(CompositeCounter.java:31)
	at io.micrometer.core.instrument.composite.AbstractCompositeMeter.lambda$forEachChild$0(AbstractCompositeMeter.java:50)
	at java.util.concurrent.CopyOnWriteArrayList.forEach(CopyOnWriteArrayList.java:890)
	at io.micrometer.core.instrument.composite.AbstractCompositeMeter.forEachChild(AbstractCompositeMeter.java:50)
[...]

PR #3627, opened 2023-02-09
Ignore metrics while incrementing counters in LogbackMetrics

jonatan-ivanov:
fixes gh-3623

### DIFF
--- a/benchmarks/benchmarks-core/build.gradle
+++ b/benchmarks/benchmarks-core/build.gradle
@@ -14,12 +14,15 @@ dependencies {
 
     jmh 'org.openjdk.jmh:jmh-core:latest.release'
 
+    jmh 'ch.qos.logback:logback-classic'
+
     // Nebula doesn't like having jmhAnnotationProcessor without jmh so we just add it twice.
     jmh 'org.openjdk.jmh:jmh-generator-annprocess:latest.release'
     jmhAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:latest.release'
 }
 
 jmh {
+    jmhVersion = '1.36'
     fork = 1
     warmupIterations = 1
     iterations = 1
--- a/benchmarks/benchmarks-core/gradle.lockfile
+++ b/benchmarks/benchmarks-core/gradle.lockfile
@@ -2,7 +2,9 @@
 # Manual edits can break the build and are not advised.
 # This file is expected to be part of source control.
 antlr:antlr:2.7.7=nohttp
+ch.qos.logback:logback-classic:1.2.11=jmh,jmhCompileClasspath,jmhRuntimeClasspath
 ch.qos.logback:logback-classic:1.2.3=checkstyle,nohttp,nohttp-cli
+ch.qos.logback:logback-core:1.2.11=jmh,jmhCompileClasspath,jmhRuntimeClasspath
 ch.qos.logback:logback-core:1.2.3=checkstyle,nohttp,nohttp-cli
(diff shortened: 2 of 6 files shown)


shakuzen:
I added a benchmark for `LogbackMetrics` that intends to measure the cost of logging with and without `LogbackMetrics`. More interesting for this change is the difference before/after this change. I did not commit the benchmark for that, but locally I made a copy of `LogbackMetrics` with the code prior to this PR and ran the benchmark against that version of LogbackMetrics, this version, and without registering any LogbackMetrics. That appears to show that this change adds significant overhead relative to the previous implementation.
The original issue was #411, and I'm not sure if the fix was intentionally limited to the Statsd implementation because of the added overhead cost, but as pointed out in https://github.com/micrometer-metrics/micrometer/issues/3623, we cannot know whether an arbitrary `Counter` implementation may log or not, so without applying the fix like this pull request, we can have the reported issue.

If we are applying this fix like this, do we still need the fix that was applied to the Statsd registry?

---
Results of a run of the benchmark in this PR on my machine:
```
Benchmark                                                  (registerLogbackMetrics)    Mode     Cnt     Score   Error  Units
LogbackMetricsBenchmark.logSomething                                          false  sample  494701     1.285 ± 0.009  us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.00                       false  sample             0.916          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.50                       false  sample             1.084          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.90                       false  sample             1.542          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.95                       false  sample             1.958          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.99                       false  sample             4.456          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.999                      false  sample             9.966          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.9999                     false  sample            38.227          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p1.00                       false  sample           640.000          us/op
LogbackMetricsBenchmark.logSomething                                           true  sample  428284     1.503 ± 0.026  us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.00                        true  sample             1.040          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.50                        true  sample             1.250          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.90                        true  sample             1.916          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.95                        true  sample             2.372          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.99                        true  sample             4.456          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.999                       true  sample            11.767          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p0.9999                      true  sample            41.536          us/op
LogbackMetricsBenchmark.logSomething:logSomething·p1.00                        true  sample          2162.688          us/op
```

Results from local benchmark comparing previous implementation:
registerLogbackMetrics
0 = don’t bind anything
1 = bind the new implementation (this pull request)
2 = bind the old implementation (before this pull request)
```
Benchmark                             (registerLogbackMetrics)  Mode  Cnt  Score   Error  Units
LogbackMetricsBenchmark.logSomething                         0  avgt    2  1.291          us/op
LogbackMetricsBenchmark.logSomething                         1  avgt    2  1.528          us/op
LogbackMetricsBenchmark.logSomething                         2  avgt    2  1.320          us/op
```

shakuzen:
> If we are applying this fix like this, do we still need the fix that was applied to the Statsd registry?

To answer my own question, I looked into it a bit, and it seems like we still need the fix in the statsd registry due to the asynchronous nature of the internal pipeline metrics recordings go through.

shakuzen (approved):

PR #3651, opened 2023-02-17
Optimize getting ThreadMetrics where getThreadInfo is repeatedly called

asasas234:
I accidentally discovered that the JvmThreadMetrics class repeatedly calls getThreadInfo and getAllThreadIds when getting the Metrics related to jvm.threads.states, resulting in a relatively wasteful CPU, so I proposed a PR to do some optimization

Here are the problems I found with the flame chart： 

image

You can see that the width of the position in my circle is not a short percentage of the entire figure.

image

The above is the problematic part of the code, it loops through all the thread states, calling getAllThreadIds and getThreadInfo once in each loop.

My optimization idea is to call getAllThreadIds and getThreadInfo only once, and group the results by State, and finally just set the Metrics based on the grouped results.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
@@ -15,6 +15,14 @@
  */
 package io.micrometer.core.instrument.binder.jvm;
 
+import java.lang.management.ManagementFactory;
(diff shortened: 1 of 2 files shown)


jonatan-ivanov (approved):

jonatan-ivanov:
Thank you for the contribution, I did some cleanup, see https://github.com/micrometer-metrics/micrometer/pull/3651/commits/4bb8b69ad32d6c68c5d8a7bfe564395a20ac6561

PR #3652, opened 2023-02-17
Update JavaDoc in RequiredSearch

CoolTomatos:
This PR includes three changes regarding the `RequiredSearch` class:
- The `findOne` method is calling `Stream::findAny` under the hood, which is contradicting to the JavaDoc, which is fixed in this PR;
- Meanwhile, renamed `findOne` to `getOne` because an exception is thrown when nothing is found, which conventionally would be named `get***`;
- While there, simplified one Optional API usage.


jonatan-ivanov:
Thank you for the contribution!
I moved your changes on top of the `1.9.x` branch and further improved the `Optional` usage.

jonatan-ivanov (approved):

PR #3657, opened 2023-02-20
Remove explicit gradle-info-plugin dependency

izeye:
This PR removes the `com.netflix.nebula:gradle-info-plugin` dependency as it doesn't seem to be used.


jonatan-ivanov:
We do use some of the info plugins (in the config file, the `contact` section). We might not need to declare the dependency manually but could you please check if the jar and the pom are the ~same with/without the plugin?

izeye:
@jonatan-ivanov Ah, you're right. It has been pinned in f5fa14616fbbcd2396fc893bfe65e41caec1bcf2 and https://github.com/micrometer-metrics/micrometer/pull/3441 temporarily, so I think we can drop the pin now.

Issue #3658, opened 2023-02-20
"jvm.threads.states" reports stale data


lenin-jaganathan:
**Describe the bug**
This [PR](https://github.com/micrometer-metrics/micrometer/pull/3651) aims at optimizing the calls to ThreadMxBean. But it ultimately makes the "jvm.threads.states" metric stale I,e this metric will always report the thread count when the class was initialized.

**Environment**
NA

**To Reproduce**
Start any application with the latest micrometer-core. Place the below piece of code which just creates threads that does nothing,
```
Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(() -> {
            new Thread(() -> {
                while (true) {
                    System.out.println("Printing");
                    try {
                        Thread.sleep(5000);
                    } catch (InterruptedException e) {
                        throw new RuntimeException(e);
                    }
                }
            }).start();
        }, 5, 5, TimeUnit.SECONDS);
new JvmThreadMetrics().bindTo(registry);
```

Export the metrics to any back-end and we will see that jvm.threads.states metric will never report any change. We can also clearly see that "jvm.threads.live" increasing which proves that there are additional threads.

**Expected behavior**
jvm.threads.states - should report the thread states at any particular point in time.

**Additional context**
This is because the logic to query the thread states was moved from Gague's Supplier to JvmThreadMetrics initialization phase. This takes the snapshot during app start time and puts it in a map that never gets updated. 

if we really see performance issues with querying the ThreadMxBean.getAllThreadIds() probably, we can query it once and use something like a countdown latch to minimize the function call and re-computation every time.
```
private volatile CountDownLatch countDownLatch = new CountDownLatch(0);
private final Map<Thread.State, Integer> threadStateCountMap = new EnumMap<>(Thread.State.class);

long getCountForThreadState(ThreadMXBean threadBean, Thread.State state) {
        this.collectThreadStateCountIfNecessary(threadBean);
        countDownLatch.countDown();
        return threadStateCountMap.get(state);
    }

private synchronized void collectThreadStateCountIfNecessary(ThreadMXBean threadMXBean) {
        if (countDownLatch.getCount() == 0) {
            Arrays.stream(Thread.State.values()).forEach(state -> threadStateCountMap.put(state, 0));

            long[] allThreadIds = threadMXBean.getAllThreadIds();
            Arrays.stream(threadMXBean.getThreadInfo(allThreadIds))
[...]

PR #3660, opened 2023-02-21
Resolve the bugs mentioned below

asasas234:
https://github.com/micrometer-metrics/micrometer/issues/3658 

To solve the bug mentioned by issuse above, my idea is to cache ThreadState and Count mapped Map by ThreadLocal, delete the state related Key by map's remove method while getting count, and delete the Map completely from ThreadLocal when the whole Map is empty.

Avoid locking, Map sharing and countDownLatch counting in this way.

I understand that the scope of the mapping cache for Thread State and Count is request level and ThreadLocal would be a bit more appropriate.

I've accessed http://localhost:8080/actuator/prometheus on my local machine and debugged the code to confirm that each time I access it, I make sure to re-fetch the number of thread states and make sure to remove the cache from ThreadLocal at the end.


asasas234:
@shakuzen @lenin-jaganathan Please help me review, thanks!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
```diff
@@ -97,4 +92,25 @@ private static String getStateTagValue(Thread.State state) {
...
+    }
+
+    private Map<Thread.State, Long> getThreadStatesGroup() {
+        ThreadMXBean threadBean = ManagementFactory.getThreadMXBean();
+        long[] allThreadIds = threadBean.getAllThreadIds();
+        Map<Thread.State, Long> stateCountGroup = Arrays.stream(threadBean.getThreadInfo(allThreadIds))
+                .collect(Collectors.groupingBy(ThreadInfo::getThreadState, () -> new EnumMap<>(Thread.State.class),
+                        Collectors.counting()));
+        for (Thread.State state : Thread.State.values()) {
+            stateCountGroup.putIfAbsent(state, null);
```
lenin-jaganathan, line 111: Putting "null" will make the Gauge reporting NaN instead of "0" which I believe conveys better. And based on exporting systems this might be breaking when Gauge returns NaN.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
```diff
@@ -15,6 +15,12 @@
  */
 package io.micrometer.core.instrument.binder.jvm;
 
+import io.micrometer.common.lang.NonNullApi;
```
lenin-jaganathan, line 18: I believe from your earlier PR `java.*` imports go first and then `io.*` imports. see [this commit](https://github.com/micrometer-metrics/micrometer/pull/3651/commits/4bb8b69ad32d6c68c5d8a7bfe564395a20ac6561) by @jonatan-ivanov

lenin-jaganathan (commented):
I will let the @jonatan-ivanov / @shakuzen  to comment on use of ThreadLocals and how much gain this will have in the context of the original problem you were trying to solve. Adding some of my concerns with this PR,

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
asasas234: This is my problem, I misremembered the default value if there is no relevant state thread, have modified the code and committed

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
asasas234: Here I compared the previous code, I did not find anything wrong, is that the order of import?

asasas234:
I can't upload attachments here, I can only take screenshots 

image

The above figure is the flame chart I got by async-profiler, you can see that this position is still relatively CPU-intensive. This figure is a screenshot of the entrance part of the flame chart, you can see that its width is not low in the whole page.

If the number of JVM threads is relatively large, then I understand that this operation is still worth optimizing, otherwise getThreadInfo and getAllThreadIds and sequential traversal of the state of each thread are to be repeated six times.

PR #3668, opened 2023-02-24
Add basic integration test for OtlpMeterRegistry and the OTel collector

### LINKED ISSUE #3534: Integration test for OTLP registry
Since there is a docker image available for the OTel Collector, we should be able to write an integration test that verifies metrics can be successfully published to the OTel Collector.

jonatan-ivanov:
Resolves #3534


implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+        //  - IDE: no telemetry_sdk_version tag
+        //  - Gradle: telemetry_sdk_version has the version number
+        whenPrometheusScraped().then().body(
+            containsString("{job=\"test\",service_name=\"test\",telemetry_sdk_language=\"java\",telemetry_sdk_name=\"io.micrometer\""),
+
+            matchesPattern("(?s)^.*test_counter\\{.+} 42\\n.*$"),
+            matchesPattern("(?s)^.*test_gauge\\{.+} 12\\n.*$"),
+
+            matchesPattern("(?s)^.*test_timer_count\\{.+} 1\\n.*$"),
+            matchesPattern("(?s)^.*test_timer_sum\\{.+} 123\\n.*$"),
```
jonatan-ivanov, line 77: Now that I'm looking at this, I don't think this is correct, this should be 123ms not 123s.

jonatan-ivanov: Seems like a collector issue, the data that we send looks like this:
```json
metrics {
  name: "test.timer"
  unit: "milliseconds"
  histogram {
    data_points {
      start_time_unix_nano: 1677210838494000000
      time_unix_nano: 1677210839021000000
      count: 1
      sum: 123.0
    }
    aggregation_temporality: AGGREGATION_TEMPORALITY_CUMULATIVE
  }
}
```

jonatan-ivanov: Created an issue to track this: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18903

shakuzen: Good catch. I guess we'll need a fix in the collector before we can change the test.

shakuzen (approved):

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
shakuzen: It seems like it might be intentional that no conversion is done currently. But then there's no way to know the unit as far as I can tell.
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/prometheus

jonatan-ivanov: I'm wondering if there are receivers that mandate a unit that is not seconds.
Or two receivers that mandate different units. Or two exporters with different units.
In all of these cases I think not doing the conversion can lead to wrong result.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+                return null;
+            }
+        };
+    }
+
+    private Response whenPrometheusScraped() {
+        // @formatter:off
+        return given()
+            .port(container.getMappedPort(9090))
+            .accept(OPENMETRICS_100)
```
shakuzen, line 115: When I tried running the collector locally, I was not able to get it to return OpenMetrics format even when sending this `Accept` header and configuring the prometheus exporter with `enable_open_metrics: true`. Debugging this test, it looks like the response is also not OpenMetrics format.

jonatan-ivanov: Great catch, I did not even notice this. This is another bug in the collector.
If I send 1.0.0 (what I should), it does not work:
```
Accept: application/openmetrics-text; version=1.0.0; charset=utf-8
```
If I send 0.0.1 (what I should not), it works:
```
Accept: application/openmetrics-text; version=0.0.1; charset=utf-8
```

See details here: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18913

jonatan-ivanov: I fixed this by requesting 0.0.1 and also included a TODO item.

Issue #2868, opened 2021-11-17
LogbackMetrics: SubstituteLoggerFactory cannot be cast to class LoggerContext


hestad:
**Describe the bug**
https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java#L56

`this(tags, (LoggerContext) LoggerFactory.getILoggerFactory());`

`ILoggerFactory` has another implementation in addition to `LoggerContext`, namely `SubstituteLoggerFactory`. We are running our application tests in parallell where our application is setup once for each test. This includes a call to LogbackMetrics() via ktor's MicrometerMetrics. In some cases when LoggerFactory.getILoggerFactory() is called in parallell you receive a SubstituteLoggerFactory instead.

**Environment**
 - Micrometer version 1.7.5
 - Micrometer registry prometheus
 - OS: macOs and Ubuntu 20
 - Java version: 16 and 17

**To Reproduce**
Possible to checkout https://github.com/navikt/su-se-bakover and run `./gradlew test`
But a smaller case would be to setup a lot of parallell tests calling LogbackMetrics() 

**Expected behavior**
>java.lang.ClassCastException: class org.slf4j.helpers.SubstituteLoggerFactory cannot be cast to class ch.qos.logback.classic.LoggerContext (org.slf4j.helpers.SubstituteLoggerFactory and ch.qos.logback.classic.LoggerContext are in unnamed module of loader 'app')
        at io.micrometer.core.instrument.binder.logging.LogbackMetrics.<init>(LogbackMetrics.java:56)
        at io.micrometer.core.instrument.binder.logging.LogbackMetrics.<init>(LogbackMetrics.java:52)
        at no.nav.su.se.bakover.web.routes.NaisRoutesKt$installMetrics$1.invoke(NaisRoutes.kt:56)
        at no.nav.su.se.bakover.web.routes.NaisRoutesKt$installMetrics$1.invoke(NaisRoutes.kt:48)
        at io.ktor.metrics.micrometer.MicrometerMetrics$Feature.install(MicrometerMetrics.kt:213)
        at io.ktor.metrics.micrometer.MicrometerMetrics$Feature.install(MicrometerMetrics.kt:171)
        at io.ktor.application.ApplicationFeatureKt.install(ApplicationFeature.kt:68)
[...]

PR #3669, opened 2023-02-27
Avoid LoggerFactory concurrent initialization

shakuzen:
It was happening before that ClassCastExceptions were being thrown from the LogbackMetrics constructor if called concurrently. When this happened, SubstituteLoggerFactory is returned and fails to cast. This attempts to avoid this situation by initializing the LoggerFactory statically in LogbackMetrics.

Resolves gh-2868

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
@@ -51,6 +51,12 @@ public class LogbackMetrics implements MeterBinder, AutoCloseable {
 
     private final Map<MeterRegistry, MetricsTurboFilter> metricsTurboFilters = new HashMap<>();
 
+    static {
+        // see gh-2868. Without this called statically, the same call in the constructor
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
```diff
@@ -51,6 +51,12 @@ public class LogbackMetrics implements MeterBinder, AutoCloseable {
 
     private final Map<MeterRegistry, MetricsTurboFilter> metricsTurboFilters = new HashMap<>();
 
+    static {
+        // see gh-2868. Without this called statically, the same call in the constructor
+        // may return SubstituteLoggerFactory and fail to cast.
+        LoggerFactory.getILoggerFactory();
```
shakuzen, line 57: This feels hacky and I'm not sure if it is a terrible idea for some reason I couldn't think of yet. The issue we're fixing largely feels like one that is probably limited to running tests. For that reason I'm not sure it is worth doing this or not, but it does seem to resolve the issue at least in our tests.

PR #3670, opened 2023-03-02
Add GC name to gc pause metric

lenin-jaganathan:
This is a follow-up PR for [3562](https://github.com/micrometer-metrics/micrometer/pull/3562). This intends to add the name of the Garbage collector to the "jvm.gc.pause" metric. Though this information might seem trivial for a single application, this really helps at the infrastructure level when 1000's of applications are running with different JDKs and different Gc configurations.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
@@ -190,22 +190,23 @@ public class JvmGcMetrics implements MeterBinder, AutoCloseable {
             CompositeData cd = (CompositeData) notification.getUserData();
             GarbageCollectionNotificationInfo notificationInfo = GarbageCollectionNotificationInfo.from(cd);
 
+            String gcName = notificationInfo.getGcName();
             String gcCause = notificationInfo.getGcCause();
             String gcAction = notificationInfo.getGcAction();
             GcInfo gcInfo = notificationInfo.getGcInfo();
             long duration = gcInfo.getDuration();
-            if (isConcurrentPhase(gcCause, notificationInfo.getGcName())) {
+
+            Tags gcTags = Tags.of("gc", gcName, "action", gcAction, "cause", gcCause).and(tags);
+            if (isConcurrentPhase(gcCause, gcName)) {
                 Timer.builder("jvm.gc.concurrent.phase.time")
-                    .tags(tags)
-                    .tags("action", gcAction, "cause", gcCause)
+                    .tags(gcTags)
                     .description("Time spent in concurrent phase")
                     .register(registry)
                     .record(duration, TimeUnit.MILLISECONDS);
             }
             else {
                 Timer.builder("jvm.gc.pause")
-                    .tags(tags)
-                    .tags("action", gcAction, "cause", gcCause)
+                    .tags(gcTags)
                     .description("Time spent in GC pause")
                     .register(registry)
                     .record(duration, TimeUnit.MILLISECONDS);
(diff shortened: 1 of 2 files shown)


shakuzen:
I like the idea and agree it would be very useful to have an identifier for the configured collector as a tag on the GC metrics. Unfortunately, I think the `gcName` on the notification is not necessarily quite what we want. From my previous investigations it has the following values. These correspond to the heap region and/or the gc phase, rather than being a single value identifying the garbage collector, which is what I think we would rather have. We could try to map from these to a collector name, but some share the same name across different collectors, and these names are not always stable across Java versions. It would be ideal if we could read the configured collector separately from these names, but I don't know of a way to do that.

G1GC
• G1 Young Generation
• G1 Old Generation

Serial
• Copy
• MarkSweepCompact

Parallel
• PS MarkSweep
• PS Scavenge

ZGC
• ZGC

Shenandoah
• Shenandoah Pauses
• Shenandoah Cycles

gencon
• scavenge
• global

balanced
• partial gc
• global garbage collect

optavgpause / optthruput
• global

metronome
• global

nogc
• Epsilon

lenin-jaganathan:
@shakuzen The list you shared is the ideal values I have intended for the tag "gc_name". The reasons for that instead of just "G1" - G1 Young Generation / G1 Old Generation makes more sense.  We had more success with this in identifying whether the Gc is for the young gen or the old gen and the name of the Gc. Today, "jvm.gc.pause" has "action" which says whether it is the end of minor or major Gc or pause, and having the Gc name will make this gcAction less useful and we can have a single metric for both the scenarios.

lenin-jaganathan:
@shakuzen  Also to add, "gc" falls under the OpenTelemetry Sematic Convention as well. Not sure how much of the OTEL convention matter in micrometer, but here is the OTEL convention on [jvm metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics) I.
>Garbage collector name and action are generally obtained via [GarbageCollectionNotificationInfo#getGcName()](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.management/com/sun/management/GarbageCollectionNotificationInfo.html#getGcName()) and [GarbageCollectionNotificationInfo#getGcAction()](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.management/com/sun/management/GarbageCollectionNotificationInfo.html#getGcAction()) respectively.

shakuzen:
> Today, "jvm.gc.pause" has "action" which says whether it is the end of minor or major Gc or pause, and having the Gc name will make this gcAction less useful and we can have a single metric for both the scenarios.

I'm not sure I understood the point about action being less useful. That may be true in the specific case of G1GC with this new tag, but we need to support many GCs and the values used for each of these things varies and is not constant across versions of Java. It may have been better if we recorded the gc time for concurrent phases and stop-the-world phases using the same metric name and a tag separating them, but that would be a breaking change at this point for users relying on the current arrangement. We can consider that separately to this pull request if you think that would be useful.

> The reasons for that instead of just "G1" - G1 Young Generation / G1 Old Generation makes more sense.

But the information about minor or major gc for g1gc is already contained in the action, so is it adding any useful information to have young/old in the gc name? Looking at other collectors, surely you can agree that `global` is not a better name than `metronome` or `gencon`. When looking at the metrics, people will not be able to distinguish between these because both will use `global`.

I would be interested to understand more about how this will be used operationally. What kinds of metric queries are you planning to use with this new tag that aren't possible currently or are difficult? Is it more about identifying the collector in use in each application or about comparing metrics values by collector?

I hope we can discuss the above more as I think this is important for making sure we're taking the right approach to provide information in the most useful way possible, and so we can understand and recommend a good approach to using this info to other users.
In the meantime, I think we can merge this feature as is for now so we can get it into a milestone and try to get wider feedback on it.

Thank you for providing the pull request and sharing your use case.

shakuzen (approved):

lenin-jaganathan:
@shakuzen  The basic idea behind having the GC information in the metric is to evaluate the performance of the GCs and also to have insight into the Gc that is actually generating this metric. I agree that for a single application "action" makes sense about the minor/major cycles. But in a hybrid environment where we have around more than 2000 applications running with different Jdk and with different Gc's for the same Jdk, information about the GC itself makes more sense. We can derive if it is a major/minor GC from the Gc name but we cannot do the same with just "action".
I do agree that with more and more GCs coming into the picture and micrometer having wider usages, we can get a wider response from users and align this.

PR #3681, opened 2023-03-10
Manual rollover of step meters for publishing partial step

### LINKED ISSUE #1882: Counters do not increment when ElasticMeterRegistry export is enabled

shakuzen:
Allows step-based meters to manually be rolled over, instead of their normal time-based rollover. This is useful in the case of publishing a partial step, such as should happen on a final publish of step meters (including when an application is shutdown before a full step interval has happened). For now, we don't see a need to expose this API publicly, so we try to keep its visibility as low as possible and mark it as internal.

This approach was chosen because it is a non-invasive option with respect to implementations of StepMeterRegistry. The changes are isolated to core code we can update. This will fix the intended problem without requiring updates to every registry implementation (possibly including custom ones we do not maintain). It avoids messing with the clock and it (hopefully) avoids making the read behavior of step meters any more complex. 

Resolves #1882
Supersedes #3533

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeter.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2023 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.step;
+
+/**
+ * Internal. Intentionally package-private.
+ */
+interface StepMeter {
+
+    /**
+     * This is an internal method not meant for general use.
+     * <p>
+     * Force a rollover of the values returned by a step meter.
+     */
+    void _manualRollover();
+
+}
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
@@ -104,4 +104,14 @@ public abstract class StepMeterRegistry extends PushMeterRegistry {
             .merge(DistributionStatisticConfig.DEFAULT);
     }
 
+    @Override
+    public void close() {
+        stop();
+        getMeters().stream()
+            .filter(meter -> meter instanceof StepMeter)
+            .map(meter -> (StepMeter) meter)
+            .forEach(StepMeter::_manualRollover);
+        super.close();
+    }
+
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepCounter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepCounter.java
@@ -26,7 +26,7 @@ import io.micrometer.core.instrument.Counter;
  *
  * @author Jon Schneider
  */
-public class StepCounter extends AbstractMeter implements Counter {
+public class StepCounter extends AbstractMeter implements Counter, StepMeter {
 
     private final StepDouble value;
 
@@ -43,6 +43,11 @@ public class StepCounter extends AbstractMeter implements Counter {
     @Override
     public double count() {
         return value.poll();
+    }
+
+    @Override
+    public void _manualRollover() {
+        value.manualRollover();
     }
 
 }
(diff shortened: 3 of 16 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeter.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.step;
+
+/**
+ * Internal. Intentionally package-private.
+ */
+interface StepMeter {
```
shakuzen, line 21: This is internal (also package-private) so it is not the most important thing to bikeshed, but I'm beginning to wonder if a more specific name would be better. `ManualRolloverCapableStepMeter` is long but descriptive.

lenin-jaganathan:
@shakuzen  Will this fix be back ported? If yes what are the versions for which we can expect this fix?

lenin-jaganathan:
@shakuzen  Will this fix be back ported? If yes what are the versions for which we can expect this fix?

shakuzen:
> Will this fix be back ported? If yes what are the versions for which we can expect this fix?

See my comment https://github.com/micrometer-metrics/micrometer/issues/1882#issuecomment-1467335481. Please do help by trying it out and letting us know if it works.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -104,4 +104,14 @@ protected DistributionStatisticConfig defaultHistogramConfig() {
             .merge(DistributionStatisticConfig.DEFAULT);
     }
 
+    @Override
+    public void close() {
+        stop();
+        getMeters().stream()
+            .filter(meter -> meter instanceof StepMeter)
+            .map(meter -> (StepMeter) meter)
+            .forEach(StepMeter::_manualRollover);
```
lenin-jaganathan, line 113: @shakuzen  I do have a concern with this particular close() method.

Scenario:
Assume that the step is 10 seconds and the application's graceful shutdown is initiated exactly or around 10th seconds. So, the scheduler would have already initiated the pushiSafely() and there are chances that close() is called around the same time. For a step meter, I believe there might be a scenario where manualRollover() will call sumThenReset() of StepValue and the pubishSafely initiated by Scheduler might call the rollCount(). This will again trigger a sumThenReset() which will make the data disappear or rest to zero.

Though this might be an edge case, registries with shorter step have more probability of running into this particular issue.

shakuzen: Thanks for taking a look and giving feedback so quickly. I think the general scenario is summarized as the registry being closed during a scheduled publish. It does sound like some data can be lost then. Would you mind opening an issue for this so we don't forget? I'm not immediately sure what we'll want to do to mitigate this. I can take a closer look tomorrow. We should be able to write a test case that simulates this scenario.

PR #3685, opened 2023-03-10
Allow custom sink for LoggingMeterRegistry

mborgraeve:
Hi,

My first PR here, of course feel free to let me know if I didn't do things properly !

Case I'm trying to solve: provide our own sink to the `LoggingMeterRegistry` so we can control how the metrics are logged. Specifically in our case we do not want them logged at `info` level, but rather at a `debug` level.

Does this make sense ?

I had a look at the test, and the use of the logger isn't tested yet. If you could provide me with some guidance about the tests you would expect that would be great !

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
@@ -48,6 +48,7 @@ import static java.util.stream.Collectors.joining;
  * Logging {@link io.micrometer.core.instrument.MeterRegistry}.
  *
  * @author Jon Schneider
+ * @author Matthieu Borgraeve
  * @since 1.1.0
  */
 @Incubating(since = "1.1.0")
@@ -75,8 +76,8 @@ public class LoggingMeterRegistry extends StepMeterRegistry {
     }
 
     /**
-     * Constructor allowing custom sink instead of a default `log::info`
-     * @param loggingSink the custom sink that will be called for each metric.
+     * Constructor allowing custom sink instead of a default {@code log::info}.
+     * @param loggingSink the custom sink that will be called for each time series.
      */
     public LoggingMeterRegistry(Consumer<String> loggingSink) {
         this(LoggingRegistryConfig.DEFAULT, Clock.SYSTEM, loggingSink);
@@ -86,7 +87,7 @@ public class LoggingMeterRegistry extends StepMeterRegistry {
      * Constructor allowing a custom sink, clock and configuration.
      * @param config the LoggingRegistryConfig
      * @param clock the Clock
-     * @param loggingSink the custom sink that will be called for each metric.
+     * @param loggingSink the custom sink that will be called for each time series.
      */
     public LoggingMeterRegistry(LoggingRegistryConfig config, Clock clock, Consumer<String> loggingSink) {
         this(config, clock, new NamedThreadFactory("logging-metrics-publisher"), loggingSink, null);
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
```diff
@@ -66,7 +66,11 @@ public LoggingMeterRegistry() {
     }
 
     public LoggingMeterRegistry(LoggingRegistryConfig config, Clock clock) {
-        this(config, clock, new NamedThreadFactory("logging-metrics-publisher"), log::info, null);
+        this(config, clock, log::info);
+    }
+
+    public LoggingMeterRegistry(LoggingRegistryConfig config, Clock clock, Consumer<String> loggingSink) {
```
jonatan-ivanov, line 72: I think the broadest ctor (the one below this) should have all the possible parameters.

jonatan-ivanov (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
jonatan-ivanov: Also, can you add some tests?

jonatan-ivanov:
I think it kind of does (to me) if you somehow want to enhance the log events.
Btw `SimpleMeterRegistry` has a `getMetersAsString` method but it is not a push meter registry so it will not emit metrics every x seconds like this registry does.

mborgraeve:
Thanks for your feedback !
I added a test with two cases for the two new constructors, made it fail locally (changing the expectedString).
I used a lambda with an assertion, that could have been a mock (the project didn't seem to import Mockito nor other mock frameworks).

Is it fine as is or does it need anything else ?

micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
```diff
@@ -66,7 +66,15 @@ public LoggingMeterRegistry() {
     }
 
     public LoggingMeterRegistry(LoggingRegistryConfig config, Clock clock) {
-        this(config, clock, new NamedThreadFactory("logging-metrics-publisher"), log::info, null);
+        this(config, clock, log::info);
+    }
+
+    public LoggingMeterRegistry(Consumer<String> loggingSink) {
```
shakuzen, line 72: A JavaDoc would be nice here.

micrometer-core/src/test/java/io/micrometer/core/instrument/logging/LoggingMeterRegistryTest.java
```diff
@@ -43,6 +46,24 @@ void defaultMeterIdPrinter() {
         assertThat(printer.id()).isEqualTo("my.gauage{tag-1=tag-2}");
     }
 
+    @Test
+    void publishUsesProvidedSinkFromConstructors() {
+        String expectedString = "my.gauage{tag-1=tag-2} value=1";
+        // simulating a mock
+        Consumer<String> sink = s -> assertThat(s).isEqualTo(expectedString);
```
shakuzen, line 53: I believe from looking at the implementation that this assertion will be run, but even if it weren't run, this test would still pass. It would be best to write the test in a way that it fails if the consumer is NOT run.

shakuzen: micrometer-core has mockito in test scope: https://github.com/micrometer-metrics/micrometer/blob/9fe11a524a7026b175093cb02227561ec575c268/micrometer-core/build.gradle#L148

mborgraeve:
I added a simple Javadoc and used Mockito, separated the two tests.

jonatan-ivanov (approved):

jonatan-ivanov:
I did some polishing (javadoc and tests), please let me know if it's not ok.
Thank you and congrats for your first PR!

mborgraeve:
That's great thanks !
I'm actually genuinely happy 🕺

PR #3702, opened 2023-03-16
Add Bnd-LastModified attribute to runtime normalization ignore list

jprinet:
This PR fixes build cache misses which started while introducing [bnd Gradle plugin](https://github.com/micrometer-metrics/micrometer/commit/7c7ed9968e7815232bdf1d8ae37e059463030bb3).

Since then, a new property `Bnd-LastModified` has appeared in the `Manifest` file. This property is changing between each build, meaning all Gradle task using a Jar containing this `Manifest` file will be impacted.

Running 2 `gradlew clean assemble` in sequence leads to build cache misses on:
- `compileKotlin`
- `javadoc`
- `testingBundle`

Screenshot 2023-03-16 at 12 23 02 PM

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -128,6 +128,7 @@ subprojects {
             runtimeClasspath {
                 metaInf {
                     [
+                            'Bnd-LastModified',
                             'Build-Date',
                             'Build-Date-UTC',
                             'Built-By',


shakuzen (approved):
Thank you for noticing this and sending a fix for it. Much appreciated.

PR #3708, opened 2023-03-22
Add supplier variant for Observation.Context.getOrDefault()

izeye:
This PR adds a supplier variant for `Observation.Context.getOrDefault()` to avoid default object creation where unnecessary.

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -1295,6 +1295,18 @@ public interface Observation extends ObservationView {
         <T> T getOrDefault(Object key, T defaultObject);
 
         /**
+         * Returns an element or default if not present.
+         * @param key key
+         * @param defaultObjectSupplier supplier for default object to return
+         * @param <T> value type
+         * @return object or default if not present
+         */
+        default <T> T getOrDefault(Object key, Supplier<T> defaultObjectSupplier) {
+            T value = get(key);
+            return value != null ? value : defaultObjectSupplier.get();
+        }
+
+        /**
          * Returns low cardinality key values.
          * @return low cardinality key values
          */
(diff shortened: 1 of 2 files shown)


micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -1294,6 +1307,15 @@ interface ContextView {
...
         <T> T getOrDefault(Object key, T defaultObject);
 
+        /**
+         * Returns an element or default if not present.
+         * @param key key
+         * @param defaultObjectSupplier supplier for default object to return
+         * @param <T> value type
+         * @return object or default if not present
+         */
+        <T> T getOrDefault(Object key, Supplier<T> defaultObjectSupplier);
```
marcingrzejszczak, line 1317: I guess that should be a default method of we don't want to break current users (assuming that anyone has ever created their own implementation of ObservationRegistry)

izeye: @marcingrzejszczak Thanks for the feedback! I updated this based on your feedback.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -1294,6 +1294,18 @@ interface ContextView {
          */
         <T> T getOrDefault(Object key, T defaultObject);
 
+        /**
+         * Returns an element or default if not present.
+         * @param key key
+         * @param defaultObjectSupplier supplier for default object to return
+         * @param <T> value type
+         * @return object or default if not present
+         */
```
jonatan-ivanov, line 1303: Could you please add `@since 1.11.0`?

izeye:
@marcingrzejszczak @jonatan-ivanov Thanks! I applied your reviews in e11b3bd.

Issue #3711, opened 2023-03-23
PushMeterRegistry scheduled publish can collide with publish on close


lenin-jaganathan:
**Describe the bug**
[gh-3681](https://github.com/micrometer-metrics/micrometer/pull/3681) address the issue for Step Registries where data for the last incomplete step is incorrect. This fix should work fine for most of the cases yet it can suffer from some corner cases where `registry.close()` can be invoked while there was already a publish operation running from the ScheduleExecutor. This **might** result in inconsistent data being reported for both steps. 

So, there might be scenarios where data for a couple of steps might be inaccurate/ data loss.

**Environment**
 - Micrometer version - 1.11.0-M2
 - Micrometer registry - StepMeterRegistry

**To Reproduce**
How to reproduce the bug:
Scenario:
Assume that the step is 10 seconds and the application's graceful shutdown is initiated exactly or around 10th seconds. So, the scheduler would have already initiated the pushiSafely() and there are chances that close() is called around the same time. For a step meter, I believe there might be a scenario where manualRollover() will call sumThenReset() of StepValue and the pubishSafely initiated by Scheduler might call the rollCount(). This will again trigger a sumThenReset() which will make the data disappear or rest to zero.

**Expected behavior**
Duplicate publish should not be triggered and data reported for any step should not miss data points.

lenin-jaganathan:
Haven't put much thought into this, but have some ideas that might prevent 2 publishes from being executed simultaneously.

See this code sample - https://github.com/micrometer-metrics/micrometer/compare/main...lenin-jaganathan:duplicate_publish?expand=1

shakuzen:
I've got a test locally I can add for this (though there's an outstanding visibility issue with calling `publishSafely` in the test outside of the same package). Feel free to open your suggested code as a pull request.

shakuzen:
The fix for this guards against publish being called on close if a scheduled publish is already happening. It does not consider if a manual call to `publish` is being called, however.

PR #3714, opened 2023-03-24
Avoid duplicate publishing of data.

lenin-jaganathan:
This PR tries to address the issue mentioned in [3711](https://github.com/micrometer-metrics/micrometer/issues/3711). This adds a boolean `publishing` and exposes a public method `isPublishing()` to PushMeterRegistry to check if the registry is publishing data or not. 
Well, this restricts the publishing of data if there is already a publish operation in place. This can be true for StepRegistries but not sure about the desired behavior for Cumulative Registries. Since, `publishSafely()` is private to PushMeterRegistry, adding this as the default behavior for all the PushRegistries.

TODO:
- [x] Add Unit tests.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -26,12 +26,15 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public abstract class PushMeterRegistry extends MeterRegistry {
 
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(PushMeterRegistry.class);
 
     private final PushRegistryConfig config;
+
+    private final AtomicBoolean publishing = new AtomicBoolean(false);
 
     @Nullable
     private ScheduledExecutorService scheduledExecutorService;
@@ -50,12 +53,28 @@ public abstract class PushMeterRegistry extends MeterRegistry {
      * Catch uncaught exceptions thrown from {@link #publish()}.
      */
     private void publishSafely() {
-        try {
-            publish();
+        if (this.publishing.compareAndSet(false, true)) {
+            try {
+                publish();
+            }
+            catch (Throwable e) {
+                logger.warn("Unexpected exception thrown while publishing metrics for " + getClass().getSimpleName(),
+                        e);
+            }
+            finally {
+                this.publishing.set(false);
+            }
         }
-        catch (Throwable e) {
-            logger.warn("Unexpected exception thrown while publishing metrics for " + getClass().getSimpleName(), e);
+        else {
+            logger.warn("Publishing is already in progress.");
         }
+    }
+
+    /**
+     * @return - true if registry is publishing metrics
+     */
+    public boolean isPublishing() {
+        return publishing.get();
     }
 
     /**
@@ -92,10 +111,10 @@ public abstract class PushMeterRegistry extends MeterRegistry {
 
     @Override
     public void close() {
+        stop();
         if (config.enabled() && !isClosed()) {
             publishSafely();
         }
-        stop();
         super.close();
     }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -71,7 +71,7 @@ public abstract class PushMeterRegistry extends MeterRegistry {
     }
 
     /**
-     * @return - true if registry is publishing metrics
+     * Returns - true if registry is publishing metrics.
      */
     public boolean isPublishing() {
         return publishing.get();
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -52,7 +52,8 @@ public abstract class PushMeterRegistry extends MeterRegistry {
     /**
      * Catch uncaught exceptions thrown from {@link #publish()}.
      */
-    private void publishSafely() {
+    // VisibleForTesting
+    void publishSafely() {
         if (this.publishing.compareAndSet(false, true)) {
             try {
                 publish();
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -67,14 +67,14 @@ public abstract class PushMeterRegistry extends MeterRegistry {
             }
         }
         else {
-            logger.warn("Publishing is already in progress.");
+            logger.warn("Publishing is already in progress. Skipping duplicate call to publishSafely().");
         }
     }
 
     /**
-     * Returns - true if registry is publishing metrics.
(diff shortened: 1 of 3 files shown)


lenin-jaganathan:
Just realized that we are using `Executors.newSingleThreadScheduledExecutor()` for scheduling publishing which will make sure, there will not be more than one publishing at a time. This makes me wonder how registries will behave when publishing takes longer than a step, I agree that this is not desired yet it makes me wonder what will happen in case of slow publishing.
Don't want to deviate from this issue but wanted to see if we can handle that in a better manner.

shakuzen:
> This makes me wonder how registries will behave when publishing takes longer than a step

From `ScheduledExecutorService#scheduleAtFixedRate` which we use in the implementation of the `start` method on `PushMeterRegistry`:

> If any execution of this task takes longer than its period, then subsequent executions may start late, but will not concurrently execute.

I agree it is separate from this issue, and I'm not sure we can necessarily do anything to detect such a situation happening, but it might be nice if we could log a warning when such a thing happens.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -50,14 +53,30 @@ protected PushMeterRegistry(PushRegistryConfig config, Clock clock) {
...
         }
-        catch (Throwable e) {
-            logger.warn("Unexpected exception thrown while publishing metrics for " + getClass().getSimpleName(), e);
+        else {
+            logger.warn("Publishing is already in progress.");
         }
     }
 
+    /**
+     * Returns - true if registry is publishing metrics.
```
shakuzen, line 74: It would be more accurate to say a scheduled publish of metrics is happening. Calls to the `publish` method will not set `isPublishing`; only calls to `publishSafely` which is private and what is scheduled.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -50,14 +53,30 @@ protected PushMeterRegistry(PushRegistryConfig config, Clock clock) {
...
+                        e);
+            }
+            finally {
+                this.publishing.set(false);
+            }
         }
-        catch (Throwable e) {
-            logger.warn("Unexpected exception thrown while publishing metrics for " + getClass().getSimpleName(), e);
+        else {
+            logger.warn("Publishing is already in progress.");
```
shakuzen, line 69: It would be good to also explicitly state that the publish will consequently be skipped.

shakuzen (commented):
Thank you for the pull request. It looks good overall. I left some comments for minor improvements. I'll add the test case after those have been addressed and merge.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -92,10 +112,10 @@ public void stop() {
 
     @Override
     public void close() {
+        stop();
```
jonatan-ivanov, line 115: I like shutting down the executor first.

micrometer-core/src/test/java/io/micrometer/core/instrument/push/PushMeterRegistryTest.java
```diff
@@ -79,17 +77,99 @@ void whenUncaughtExceptionInPublish_taskStillScheduled() throws InterruptedExcep
...
         pushMeterRegistry.close();
-        assertThat(((CountingPushMeterRegistry) pushMeterRegistry).publishCount.get()).isOne();
+        assertThat(pushMeterRegistry.publishCount.get()).isOne();
+    }
+
+    @Test
+    @Issue("#3711")
+    void scheduledPublishOverlapWithPublishOnClose() throws InterruptedException {
+        MockClock clock = new MockClock();
+        CyclicBarrier barrier = new CyclicBarrier(2);
```
jonatan-ivanov, line 98: Since the barrier is not reset, is it used because of its "all-or-nothing breakage model" and because it throws a timeout?
I guess that's also why the result of `await` can be ignored.

Since no reset is needed I believe I would used a `CountDownLatch` and checking the result of `await` to see it this was normal termination or an error, but I think using a barrier makes this simpler. 👍🏼

jonatan-ivanov (approved):
I think we should call out somewhere (issue/javadoc) that this only protects `publishSafely` but not `publish` (I think that's fine).

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -49,15 +52,32 @@ protected PushMeterRegistry(PushRegistryConfig config, Clock clock) {
...
-        try {
-            publish();
+    // VisibleForTesting
+    void publishSafely() {
+        if (this.publishing.compareAndSet(false, true)) {
+            try {
+                publish();
+            }
+            catch (Throwable e) {
+                logger.warn("Unexpected exception thrown while publishing metrics for " + getClass().getSimpleName(),
```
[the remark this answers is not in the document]
shakuzen: This is the existing log message. I'm not sure we guarantee compatibility on log message text, but we should probably consider changing existing log messages separately from this pull request.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: Yeah, I think it always should have been this way, regardless of this issue. No one ever complained, and no tests exist that would catch it being a problem, but it logically makes sense to do this first. I noticed it when working on the StepMeterRegistry's close method, but didn't end up opening an issue for it. Glad we've got it updated now. I didn't find any history on why it intentionally would have been the way it was before.

lenin-jaganathan:
Though this PR addresses https://github.com/micrometer-metrics/micrometer/issues/3711, I do have some concerns with the way `manualRollover()` will work. Especially, let's say the application shuts down closer to the end of the step which means, the next call to publish will call poll() and if the step has crossed we will roll over the data. And that also seems to be a concerning one.
Probably, we should have a signal like "stop Everything" sent to meters so that everything accumulated is not reset. Now, this makes me rethink whether `manualRollover()` will even help us in getting a solution for this without complexity. I should have raised this earlier but somehow missed it.

lenin-jaganathan:
@shakuzen / @jonatan-ivanov [ Here is the link to a gist](https://gist.github.com/lenin-jaganathan/c74bc01791c49fde6223a38155928c23) where I have a test that just exposes the scenario I said above.

As I mentioned in the earlier comment, all the meters need to know that there needs to be no more rollover of step should happen if manualRollover is called. It seems to be a bit more complicated I guess. [This is the PR](https://github.com/micrometer-metrics/micrometer/pull/3533) I was referring to earlier.

shakuzen:
> Though this PR addresses #3711, I do have some concerns with the way `manualRollover()` will work.

Thanks for thinking about this and always working to make things better. Please open a new issue describing the problem scenario and we can consider that separately. There may be some remaining specific timing cases where data will be dropped, but we're certainly in a lot better of a situation than we were before manualRollover where we would resend previous step data or send 0 if a full step had not elapsed. The data was always wrong then, and now I think we can agree it will be right almost all the time. We can still address these cases and improve the behavior going forward.

micrometer-core/src/test/java/io/micrometer/core/instrument/push/PushMeterRegistryTest.java
shakuzen: I originally tried using a CountDownLatch, but it doesn't ensure we're checking what we want to check. The same thread could count down twice, so I thought it safer to use the barrier to ensure we had two different threads waiting at the same time. I'm happy to consider ways to make the test better. Feel free to follow-up post merge. It's possible the test won't catch all kinds of possible failures, but I did check that it failed before the changes to main code at least.

PR #3727, opened 2023-03-30
MetricTag

### LINKED ISSUE #1732: @Timed annotation with dynamic tags derived from input arguments and/or output
# Use case

While `@Timed` annotation provides good starting point for annotating particular methods in a single service. In case of a platform single API can get used by completely different clients with somewhat different access patterns. Implies different notification thresholds or alert receivers. Consider a sample api 

```
    @Timed
    public Response process(BucketId bucketId, ProcessingParams params) { ... }
```

If all incoming requests are timed in a same manner a metric visualization could take the form of 

plot_all_combined

However if we were to take a look at it with a bit more precision, say plot requests to different bucket ids separately one could arrive at 

plot_all

Going further, excluding outliers paints a completely different picture 

plot_some

As it usually happens, some like 20% of requests of specific form could incur up to 80% of support burden and identifying those would make the whole monitoring exercise much more useful.

# Proposal

Introduce a new (or extend existing?) `@Timed` annotation which could take optional list of tag names as well as custom expressions similar to [SpEL](https://docs.spring.io/spring/docs/4.3.10.RELEASE/spring-framework-reference/html/expressions.html)

So that annotations could take form of 

```
    @Timed(
       inputTags = {"bucket_id", "#{[0]}"},
       outputTags = {"status": "#{[0]?.status}"}
    )
    public Response process(BucketId bucketId, ProcessingParams params) { ... }

```

# Cardinality

This, of course, is open to some abuse in cases where cardinality of input/output values becomes high. Consider what would happen in example above if `bucket_id` turns out to be a randomly generated UUIDv4. In such a case, making sure the right value/expression is provided is ultimately part of developer's responsibility. 

It is worth noting at this point the current alternative is to provide a dedicated endpoint to every specific use-case which does not seem scalable.

# References

This issue has use case similar to #1586 yet a different approach to address it. This issue is different from `extraTags` in a sense that `extraTags` are static and current issue proposes dynamically evaluated values. This issue is different from #1265 in a sense that it is protocol agnostic. It is also different from #535 in a sense that tag value is provided "declaratively"

marcingrzejszczak:
- Moves the common code from Micrometer Tracing
- Adds a setter to `TimedAspect` (too many constructors already)
- This only works for the `Timer.Builder` not for the `LongTaskTimer.Builder` (code can be modified to allow that)

fixes #1732

### DIFF
--- a/micrometer-commons/build.gradle
+++ b/micrometer-commons/build.gradle
@@ -4,6 +4,9 @@ dependencies {
     // log monitoring
     optionalApi 'ch.qos.logback:logback-classic'
 
+    // Aspects
+    optionalApi 'org.aspectj:aspectjweaver'
+
     // JUnit 5
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'com.tngtech.archunit:archunit-junit5'
--- a/micrometer-commons/build.gradle
+++ b/micrometer-commons/build.gradle
@@ -16,3 +16,14 @@ dependencies {
     testImplementation 'org.assertj:assertj-core'
     testImplementation 'org.awaitility:awaitility'
 }
+
+jar {
+    bundle {
+
+        bnd '''\
+        Import-Package: \
+            org.aspectj.*;resolution:=dynamic,\
+            *
+        '''.stripIndent()
+    }
+}
--- a/micrometer-commons/src/main/java/io/micrometer/common/annotation/TagAnnotationHandler.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/annotation/TagAnnotationHandler.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2023 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.common.annotation;
+
+import io.micrometer.common.KeyValue;
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.reflect.MethodSignature;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+/**
+ * This class is able to find all methods annotated with the Micrometer annotations. All
+ * methods mean that if you have both an interface and an implementation annotated with
+ * Micrometer annotations then this class is capable of finding both of them and merging
+ * into one set of information.
+ * <p>
+ * This information is then used to add proper tags to objects such as span or timer from
+ * the method arguments that are annotated with a tag related annotation.
+ *
+ * Code ported from Spring Cloud Sleuth.
+ *
+ * @author Christian Schwerdtfeger
+ * @since 1.11.0
+ */
+public class TagAnnotationHandler<T> {
+
+    private static final InternalLogger log = InternalLoggerFactory.getInstance(TagAnnotationHandler.class);
+
+    private final BiConsumer<KeyValue, T> tagConsumer;
+
+    private final Function<Class<? extends TagValueResolver>, ? extends TagValueResolver> resolverProvider;
+
+    private final Function<Class<? extends TagValueExpressionResolver>, ? extends TagValueExpressionResolver> expressionResolverProvider;
+
+    private final Class<? extends Annotation> tagClass;
+
+    private final BiFunction<Annotation, Object, KeyValue> toKeyValue;
+
+    public TagAnnotationHandler(BiConsumer<KeyValue, T> tagConsumer,
+            Function<Class<? extends TagValueResolver>, ? extends TagValueResolver> resolverProvider,
+            Function<Class<? extends TagValueExpressionResolver>, ? extends TagValueExpressionResolver> expressionResolverProvider,
+            Class<? extends Annotation> tagClass, BiFunction<Annotation, Object, KeyValue> toKeyValue) {
+        this.tagConsumer = tagConsumer;
+        this.resolverProvider = resolverProvider;
+        this.expressionResolverProvider = expressionResolverProvider;
+        this.tagClass = tagClass;
+        this.toKeyValue = toKeyValue;
+    }
+
+    public void addAnnotatedParameters(T objectToTag, ProceedingJoinPoint pjp) {
+        try {
+            Method method = ((MethodSignature) pjp.getSignature()).getMethod();
+            List<AnnotatedParameter> annotatedParameters = AnnotationUtils.findAnnotatedParameters(tagClass, method,
+                    pjp.getArgs());
+            getAnnotationsFromInterfaces(pjp, method, annotatedParameters);
+            addAnnotatedArguments(objectToTag, annotatedParameters);
+        }
+        catch (SecurityException ex) {
+            log.error("Exception occurred while trying to add annotated parameters", ex);
+        }
+    }
+
+    private void getAnnotationsFromInterfaces(ProceedingJoinPoint pjp, Method mostSpecificMethod,
+            List<AnnotatedParameter> annotatedParameters) {
+        Class<?>[] implementedInterfaces = pjp.getThis().getClass().getInterfaces();
+        if (implementedInterfaces.length > 0) {
+            for (Class<?> implementedInterface : implementedInterfaces) {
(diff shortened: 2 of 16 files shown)


micrometer-commons/build.gradle
```diff
@@ -4,6 +4,9 @@ dependencies {
     // log monitoring
     optionalApi 'ch.qos.logback:logback-classic'
 
+    // Aspects
+    optionalApi 'org.aspectj:aspectjweaver'
```
shakuzen, line 8: @driessamyn Could you help us with what changes need to be made for osgi related to this. The build is failing when running the osgi-test. See, for example, https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/4025/workflows/295057ca-f327-4ed1-81dd-ba46b5f0173a/jobs/21266

driessamyn: No problem, @shakuzen 

Basically, the issue can be identified here:

```
Resolution failed. Capabilities satisfying the following requirements could not be found:
    [<<INITIAL>>]
      ⇒ osgi.identity: (osgi.identity=micrometer-osgi-test-tests)
          ⇒ [micrometer-osgi-test-tests version=1.11.0.SNAPSHOT]
              ⇒ osgi.wiring.package: (&(osgi.wiring.package=io.micrometer.core.instrument.composite)(version>=1.11.0)(!(version>=2.0.0)))
                  ⇒ [micrometer-core version=1.11.0.SNAPSHOT]
                      ⇒ osgi.wiring.package: (&(osgi.wiring.package=io.micrometer.common.annotation)(version>=1.11.0)(!(version>=2.0.0)))
                          ⇒ [micrometer-commons version=1.11.0.SNAPSHOT]
                              ⇒ osgi.wiring.package: (osgi.wiring.package=org.aspectj.lang)
```

Basically, what that is saying is that `micrometer-commons` is expecting the package `org.aspectj.lang` to be "visible".
However, given this is an optional dependency, it can be marked as `dynamic` in the OSGi metadata, which means it won't try to resolve it until it's needed.

This is a patch that should resolve it: (let me know if you rather I raise a PR)

```
Index: micrometer-commons/build.gradle
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/micrometer-commons/build.gradle b/micrometer-commons/build.gradle
--- a/micrometer-commons/build.gradle	(revision dd86a65bb1f3d02da2bd61515d487d5fa91c0537)
+++ b/micrometer-commons/build.gradle	(date 1680507878716)
@@ -16,3 +16,14 @@
     testImplementation 'org.assertj:assertj-core'
     testImplementation 'org.awaitility:awaitility'
 }
+
+jar {
+    bundle {
+
+        bnd '''\
+        Import-Package: \
+            org.aspectj.*;resolution:=dynamic,\
+            *
+        '''.stripIndent()
+    }
+}
```

I did try to come up with a way top auto-generate the OSGi metadata based on the `optionalApi` configuration, but I couldn't think of a way to do this in gradle.

shakuzen: Thanks for the help. That's fixed the build now.

micrometer-commons/src/main/java/io/micrometer/common/annotation/TagAnnotationHandler.java
```diff
@@ -0,0 +1,137 @@
...
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+/**
+ * This class is able to find all methods annotated with the Micrometer annotations. All
+ * methods mean that if you have both an interface and an implementation annotated with
```
shakuzen, line 34: Could we add a test to show this works?

micrometer-core/src/main/java/io/micrometer/core/aop/MetricTag.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Execute this expression to calculate the tag value. Will be analyzed if no value of
+     * the {@link MetricTag#resolver()} was set.
+     * @return an expression
+     */
+    String expression() default "";
+
+    /**
+     * Use this bean to resolve the tag value. Has the highest precedence.
```
shakuzen, line 59: "bean" feels a little odd here. The class type given here will be used to resolve the `TagValueResolver` instance from the function given to the `MetricsTagAnnotationHandler`.

micrometer-core/src/test/java/io/micrometer/core/aop/TimedAspectTest.java
```diff
@@ -370,6 +376,58 @@ void timeClassFailure() {
...
+        TimedAspect timedAspect = new TimedAspect(registry);
+        timedAspect.setMetricsTagAnnotationHandler(
+                new MetricsTagAnnotationHandler(aClass -> tagValueResolver, aClass -> tagValueExpressionResolver));
+
+        AspectJProxyFactory pf = new AspectJProxyFactory(new MetricTagClass());
+        pf.addAspect(timedAspect);
+
+        MetricTagClass service = pf.getProxy();
+
+        service.getAnnotationForTagValueResolver("foo");
```
shakuzen, line 408: It'd be good to test that the tag value can change based on the parameter value. Even just concatenating it onto the current constant string would be good, I think.

micrometer-core/src/main/java/io/micrometer/core/aop/MetricTag.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.lang.annotation.*;
+
+/**
+ * There are 3 different ways to add tags to a span. All of them are controlled by the
+ * annotation values. Precedence is:
+ * <p>
+ * try with the {@link TagValueResolver} bean if the value of the bean wasn't set, try to
+ * evaluate a SPEL expression if there’s no SPEL expression just return a
+ * {@code toString()} value of the parameter
```
shakuzen, line 29: JavaDoc needs to be updated for metrics (it mentions span and SPEL and beans). It would also be good to make a warning that this should not be used with parameters that are going to result in high cardinality values (e.g. user-provided input) and there is no protection provided out-of-the-box for if it is used in such a way.

shakuzen (commented):
Mostly minor feedback left in comments. Looks great overall. This should be nice and offers good usability while being flexible.

The main feedback I have is a bikeshed argument about naming. In Micrometer core, I would go with `@MeterTag` instead of `@MetricTag`. We use `Meter` elsewhere there consistently, as far as I know. On the classes in commons, it feels a little odd they have "Tag" in the name even though commons has no tag references and instead has KeyValue. I wonder if we need the "Tag" part in the name in commons. It might be too generic, but what about `ValueResolver` and `ValueExpressionResolver`? They are resolving the Value part of KeyValue and we do say tag value also. Maybe it isn't a big deal either way, but it stuck out to me in reviewing.

micrometer-commons/src/main/java/io/micrometer/common/annotation/TagAnnotationHandler.java
marcingrzejszczak: We have a test for that https://github.com/micrometer-metrics/micrometer/pull/3727/files#diff-f6ba43cbb6d7919b0039dbb5cbef0b4b8b88fa8dbbb959103eb96bb10d2f9a50R46

shakuzen: Sorry, my comment wasn't clear at all. I meant to point out the part of the JavaDoc saying this works with both annotated interface and implementations, but the test only uses an annotated concrete class. So it would be nice if we also had a test to show this works with annotations on an interface, and merging the annotations from interface and concrete class implementing the interface.

micrometer-core/src/main/java/io/micrometer/core/aop/MetricTag.java
```diff
@@ -0,0 +1,64 @@
...
+     * the {@link MetricTag#resolver()} was set.
+     * @return an expression
+     */
+    String expression() default "";
+
+    /**
+     * Use this bean to resolve the tag value. Has the highest precedence.
+     * @return {@link TagValueResolver} bean
+     */
+    Class<? extends TagValueResolver> resolver() default NoOpTagValueResolver.class;
```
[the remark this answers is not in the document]
marcingrzejszczak: That's how it worked with Sleuth. The Expression resolver was e.g. the Spring Expression Language Resolver and you would have one for all expressions.

micrometer-core/src/test/java/io/micrometer/core/aop/TimedAspectTest.java
marcingrzejszczak: Most likely I didn't understand what you're saying but is there any difference between `("foo")` and `(FO + "o")`?

shakuzen: I meant that we aren't testing that the value passed to the annotated method parameter (in this case, `"foo"`) is passed to the value resolver because the value resolver used for the test returns a constant string. So if there were a bug that the value passed to the value resolver were always `null` or something, this test would still pass.

micrometer-core/src/main/java/io/micrometer/core/aop/MetricTag.java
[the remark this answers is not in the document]
shakuzen: Sounds fine to me. I just wanted to confirm my understanding. Thanks!

shakuzen (approved):

marcingrzejszczak:
I messed up the rebasing and I missed this important commit when merging this PR https://github.com/micrometer-metrics/micrometer/commit/bc59c47766ce63a0dc851cea05e5c6e45382558a :grimacing:

Issue #3729, opened 2023-03-31
Consistently add `outcome` tag to HTTP client instrumentation


kubamarchwicki:
To enhance metrics filtering, a grouping `outcome` tag would be helpful in `http_client_requests`, similar to the `outcome` tag in the `http_server_requests`.

For less matured metrics systems (like CloudWatch) it's impossible to filter by patterns, and having a similar grouping in HTTP client and server metrics would be very helpful. 

PS. I couldn't find any issue related to this one, so I created a new one. Please let me know if I've overlooked something
PS2. If such a tag makes sense, I'm happy to raise a PR and work on that.

shakuzen:
Thanks for opening the issue. Which HTTP client instrumentation are you using? We already have an outcome tag on some of them.

kubamarchwicki:
Thank you for the swift answer. I was instrumenting JDK11 `HttpClient`
I can see the `Outcome::forStatus` is used a couple of times, but primarily for server-side requests (and for JettyClient). 

I'll happily raise a couple of PRs for different clients to align it

shakuzen:
Thank you for the offer to help. We'd be happy to review pull requests for this.

PR #3732, opened 2023-03-31
Add outcome tag to HttpClient metrics

kubamarchwicki:
Fixes https://github.com/micrometer-metrics/micrometer/issues/3729

Aligning tags between `http_client_requests` metrics. A grouping outcome tag was added to all http client binders - similar to the outcome tag in `JettyClient` and the `http_server_requests`.


shakuzen (commented):
Really great work aligning all of these. Thank you so much for that. I think we can also add an `outcome` to the tags asserted in `HttpClientTimingInstrumentationVerificationTests` since it should be on all implementations now. Would you update that? Then I think it's ready to merge. Thanks again.

kubamarchwicki:
Good catch. Forgot about it. Added in acebcb039

shakuzen (approved):
Thank you a lot. Great work.

Issue #3706, opened 2023-03-22
Provide a way to give custom metric name to MicrometerHttpRequestExecutor


beatfreaker:
**Please describe the feature request.**
Currently the metric name for `MicrometerHttpRequestExecutor` is fixed to `httpcomponents.httpclient.request`, requesting to provide a way to give custom metric name like we have such option in `OkHttpMetricsEventListener`

**Rationale**
We have different libraries which use different HTTP client libraries to make API calls i.e `RestTemplate`, `OkHttp` and `HttpClient`

Now we want to make sure that all the metrics generated by all these three clients are reported with same metric name. I know we can customize spring-boot's `RestTemplate` and `OkHttp` client metric name to `httpcomponents.httpclient.request` so that all metrics are reported under `httpcomponents.httpclient.request` but we like to name the metric as `http.client.requests` so need an option to provide custom metric name to `MicrometerHttpRequestExecutor`

**Additional context**

jonatan-ivanov:
Since okhttp has it I think it would make sense to add it to this one too. Are you up to filing a PR for this?

In the meantime: you can customize any `Meter` using a `MeterFilter`, in your case:
```java
new MeterFilter() {
    @Override
    public Meter.Id map(Meter.Id id) {
       if(id.getName().equals("httpcomponents.httpclient.request")) {
          return id.withName("http.client.requests");
       }
       return id;
    }
}
```
See the docs: https://micrometer.io/docs/concepts#_transforming_metrics

beatfreaker:
@jonatan-ivanov Sure, you can assign this issue to me. I will work on a PR to resolve it.

beatfreaker:
Hi @jonatan-ivanov ,

Should I create PR against `main` branch?

Also is it fine to implement a similar capability in `MicrometerHttpClientInterceptor`?

jonatan-ivanov:
>Should I create PR against main branch?

Yes, since this is a new functionality (it's not really a bug: a `MeterFilter` can be used to change the name).

>Also is it fine to implement a similar capability in MicrometerHttpClientInterceptor?

[...]

PR #3734, opened 2023-04-01
Provide an option to customize meter name for HttpClient metrics

beatfreaker:
Provides an option to customize meter name for metrics generated for requests that goes through `HttpClient` and `HttpAsyncClient`

Resolved gh-3706

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/MicrometerHttpClientInterceptor.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/httpcomponents/MicrometerHttpClientInterceptor.java
@@ -68,11 +68,12 @@ public class MicrometerHttpClientInterceptor {
      * @param uriMapper URI mapper to create {@code uri} tag
      * @param extraTags extra tags
      * @param exportTagsForRoute whether to export tags for route
+     * @param meterName meter name
      */
     public MicrometerHttpClientInterceptor(MeterRegistry meterRegistry, Function<HttpRequest, String> uriMapper,
-            Iterable<Tag> extraTags, boolean exportTagsForRoute) {
+            Iterable<Tag> extraTags, boolean exportTagsForRoute, String meterName) {
         this.requestInterceptor = (request, context) -> timerByHttpContext.put(context,
-                Timer.resource(meterRegistry, METER_NAME)
+                Timer.resource(meterRegistry, meterName)
                     .tags("method", request.getRequestLine().getMethod(), "uri", uriMapper.apply(request)));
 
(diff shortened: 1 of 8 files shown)


shakuzen:
Thank you for the pull request. It looks like CircleCI is having trouble loading the CI config. Could you try logging into CircleCI once? I believe that should fix the issue. You may need to update the branch to trigger another build if you can't do it from CircleCI's UI.

shakuzen (commented):
Thanks for the pull request. This mostly looks good, but I guess we haven't really figured out what to do about [this comment](https://github.com/micrometer-metrics/micrometer/issues/3706#issuecomment-1481548955). If you make a test where `ObservationRegistry` is configured on the instrumentation, I think you will see that your custom name does not get applied to the created Timer.

beatfreaker:
@shakuzen added tests that results in showing that the custom metric name is not supported when `ObservationRegistry` is configured.
Also, the Javadoc added on `Builder metricName(String name)` stating that overridden metric name works only if `ObservationRegistry` is not configured should be fine right?

shakuzen:
It's fine in so much as the behavior is consistent with the JavaDoc, but I don't think it's ideal behavior to be adding at this point where we would expect people in general to be migrating to using the ObservationRegistry in instrumentation and yet we're adding a new configuration method that does not do anything in that case. We could make it so DefaultApacheHttpClientObservationConvention took the configured name so that it would be used for the observation too unless a custom convention is configured. The unfortunate thing is that currently `DefaultApacheHttpClientObservationConvention` is designed as a singleton, so that's going to be broken if we make a constructor that takes a name. The no arg constructor can continue to use the existing default name, though. I think that might be a better path forward. @jonatan-ivanov what do you think?

jonatan-ivanov:
I haven’t checked/tried this but if we want to make this configurable for that instrumentation, we might be able to provide a convention that extends the default and overrides the name to the user provided one but if we do that why isn’t the user the one then who provides it in the first place?

shakuzen:
> why isn’t the user the one then who provides it in the first place?

They certainly could provide a custom convention, but the main issue is that we will be adding a configuration for name that does nothing in case observation-based instrumentation is used. Users could simply pass a name for Timer-based instrumentation but have to provide a custom convention to do the same with Observation-based instrumentation.

I made [this commit](https://github.com/micrometer-metrics/micrometer/commit/72aa1a67b52cae6df1aea70952a85fb0fb2df7b5) demonstrating what I was suggesting. It's certainly not without flaws. We didn't initially intend for default conventions to be used that way. I'm not sure which way is better. (I didn't make the equivalent changes to the hc5 package since it was to demonstrate my suggestion for review first)

Issue #522, opened 2018-03-27
Metrics support for Netty allocators and event executors


zeagord:
It will be nice to see metrics added around Netty. For example,

1. Number of threads in the eventloop
2. Processing time in the eventloop
3. Throughput

tkp1n:
I recently added the following netty-micrometer glue code to a spring boot project to monitor memory usage by the default `PooledByteBufAllocator`. 
Although not in your list, this may be useful to some...

<details>
<summary>lots of code hidden...</summary>

```java
package x.y.z;

import io.micrometer.core.instrument.Metrics;
import io.micrometer.core.instrument.Tags;
import io.netty.buffer.*;
import org.springframework.context.annotation.Configuration;

import javax.annotation.PostConstruct;

@Configuration
public class NettyMetricsConfig {
    private static final String NETTY = "netty";
    private static final String ALLOC = "alloc";
    private static final String POOLED = "pooled";
    private static final String UNPOOLED = "unpooled";

    private static final String MEMORY = "memory";
    private static final String DIRECT = "direct";
    private static final String HEAP = "heap";

    private static final String ARENA = "arena";

    private static final String CHUNK = "chunk";

    private static final String THREAD = "thread";
    private static final String LOCAL = "local";

    private static final String CACHE = "cache";
    private static final String SIZE = "size";
    private static final String TINY = "tiny";
    private static final String SMALL = "small";
    private static final String NORMAL = "normal";
    private static final String HUGE = "huge";

    private static final String SUBPAGE = "subpage";
    private static final String CHUNKLIST = "chunklist";
    private static final String NUMBER = "number";

    private static final String USED = "used";
    private static final String COUNT = "count";

    private static final String ALLOCATION = "allocation";
    private static final String DEALLOCATION = "deallocation";
    private static final String ACTIVE = "active";
    private static final String BYTE = "byte";

    private static final String ELEMENT = "element";
    private static final String MAX = "max";
    private static final String MIN = "min";
    private static final String AVAILABLE = "available";
    private static final String PAGE = "page";
    private static final String USAGE = "usage";

    @PostConstruct
    public void configureNettyMetrics() {
        PooledByteBufAllocatorMetric pooledMetric = PooledByteBufAllocator.DEFAULT.metric();

        Tags pooled = Tags.of(ALLOC, POOLED);

[...]

PR #3742, opened 2023-04-04
Add metrics support for Netty 4.x

bclozel:
This commit adds two new `MeterBinder` implementations for instrumenting
Netty 4.x: `NettyAllocatorMetrics` and `NettyEventExecutorMetrics`.

`NettyAllocatorMetrics` will instrument any `ByteBufAllocatorMetricProvider`
and gather information about heap/direct memory allocated; additional
metrics are provided for pooled allocators.

`NettyEventExecutorMetrics` will instrument `EventExecutor` (typically,
`EventLoop` instances) and count the number of pending tasks for each.

Metrics and tags are described in the `NettyMeters` class.

Closes gh-522

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyAllocatorMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyAllocatorMetrics.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2023 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.netty4;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.binder.MeterBinder;
+import io.netty.buffer.ByteBufAllocatorMetric;
+import io.netty.buffer.ByteBufAllocatorMetricProvider;
+import io.netty.buffer.PooledByteBufAllocator;
+import io.netty.buffer.PooledByteBufAllocatorMetric;
+
+/**
+ * {@link MeterBinder} for Netty memory allocators.
+ *
+ * @author Brian Clozel
+ * @since 1.11.0
+ * @see NettyMeters
+ */
+public class NettyAllocatorMetrics implements MeterBinder {
+
+    private final ByteBufAllocatorMetricProvider allocator;
+
+    /**
+     * Create a binder instance for the given allocator.
+     * @param allocator the {@code ByteBuf} allocator to instrument
+     */
+    public NettyAllocatorMetrics(ByteBufAllocatorMetricProvider allocator) {
+        this.allocator = allocator;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        int allocatorId = this.allocator.hashCode();
+
+        ByteBufAllocatorMetric allocatorMetric = this.allocator.metric();
+        Tags tags = Tags.of(NettyMeters.AllocatorMeterTags.ID.asString(), String.valueOf(allocatorId),
+                NettyMeters.AllocatorMeterTags.ALLOCATOR_TYPE.asString(), this.allocator.getClass().getSimpleName());
+
+        Gauge
+            .builder(NettyMeters.ALLOCATOR_MEMORY_USED.getName(), allocatorMetric,
+                    ByteBufAllocatorMetric::usedHeapMemory)
+            .tags(tags.and(NettyMeters.AllocatorMemoryMeterTags.MEMORY_TYPE.asString(), "heap"))
+            .register(registry);
+
+        Gauge
+            .builder(NettyMeters.ALLOCATOR_MEMORY_USED.getName(), allocatorMetric,
+                    ByteBufAllocatorMetric::usedDirectMemory)
+            .tags(tags.and(NettyMeters.AllocatorMemoryMeterTags.MEMORY_TYPE.asString(), "direct"))
+            .register(registry);
+
+        if (this.allocator instanceof PooledByteBufAllocator) {
+            PooledByteBufAllocator pooledByteBufAllocator = (PooledByteBufAllocator) this.allocator;
+            PooledByteBufAllocatorMetric pooledAllocatorMetric = pooledByteBufAllocator.metric();
+
+            Gauge
+                .builder(NettyMeters.ALLOCATOR_MEMORY_PINNED.getName(), pooledByteBufAllocator,
+                        PooledByteBufAllocator::pinnedHeapMemory)
+                .tags(tags.and(NettyMeters.AllocatorMemoryMeterTags.MEMORY_TYPE.asString(), "heap"))
+                .register(registry);
+
+            Gauge
+                .builder(NettyMeters.ALLOCATOR_MEMORY_PINNED.getName(), pooledByteBufAllocator,
+                        PooledByteBufAllocator::pinnedDirectMemory)
+                .tags(tags.and(NettyMeters.AllocatorMemoryMeterTags.MEMORY_TYPE.asString(), "direct"))
+                .register(registry);
+
+            Gauge
+                .builder(NettyMeters.ALLOCATOR_POOLED_ARENAS.getName(), pooledAllocatorMetric,
+                        PooledByteBufAllocatorMetric::numHeapArenas)
+                .tags(tags.and(NettyMeters.AllocatorMemoryMeterTags.MEMORY_TYPE.asString(), "heap"))
+                .register(registry);
+            Gauge
+                .builder(NettyMeters.ALLOCATOR_POOLED_ARENAS.getName(), pooledAllocatorMetric,
+                        PooledByteBufAllocatorMetric::numDirectArenas)
+                .tags(tags.and(NettyMeters.AllocatorMemoryMeterTags.MEMORY_TYPE.asString(), "direct"))
+                .register(registry);
+
+            Gauge
+                .builder(NettyMeters.ALLOCATOR_POOLED_CACHE_SIZE.getName(), pooledAllocatorMetric,
+                        PooledByteBufAllocatorMetric::normalCacheSize)
+                .tags(tags.and(NettyMeters.AllocatorPooledCacheMeterTags.CACHE_TYPE.asString(), "normal"))
+                .register(registry);
+            Gauge
+                .builder(NettyMeters.ALLOCATOR_POOLED_CACHE_SIZE.getName(), pooledAllocatorMetric,
+                        PooledByteBufAllocatorMetric::smallCacheSize)
+                .tags(tags.and(NettyMeters.AllocatorPooledCacheMeterTags.CACHE_TYPE.asString(), "small"))
+                .register(registry);
+
+            Gauge
+                .builder(NettyMeters.ALLOCATOR_POOLED_THREADLOCAL_CACHES.getName(), pooledAllocatorMetric,
+                        PooledByteBufAllocatorMetric::numThreadLocalCaches)
+                .tags(tags)
+                .register(registry);
+
+            Gauge
+                .builder(NettyMeters.ALLOCATOR_POOLED_CHUNK_SIZE.getName(), pooledAllocatorMetric,
+                        PooledByteBufAllocatorMetric::chunkSize)
+                .tags(tags)
+                .register(registry);
+        }
+    }
+
+}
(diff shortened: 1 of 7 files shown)


franz1981:
In case you are interested, it is possible to add these additional metrics out of Netty event loops : https://github.com/netty/netty/pull/9080

Sadly I have never completed it, but it should e possible

And the same, here: https://github.com/netty/netty/pull/11293#issuecomment-855341497
To use a sentinel periodic task, per event loop, to measure busyness of it

bclozel:
This PR is focusing on Netty 4.x - Netty 5.x is still in alpha version, we can add that support anytime.
I'll add a few notes here to highlight important points regarding the binder setup and the metrics themselves.

## MeterBinder setup

[The Reactor team is using a `ConcurrentMap` cache](https://github.com/reactor/reactor-netty/blob/10c127a1bd09c9f22550e3423cd3ffbafa797d8e/reactor-netty-core/src/main/java/reactor/netty/transport/ByteBufAllocatorMetrics.java#L57) to avoid binding metrics to the same allocator/event loop multiple times. Since allocator and event loop resources can be configured in multiple ways, the project chose to instrument those lazily at runtime as they are encountered during channel initialization.

I initially adopted that approach but rolled it back for two reasons:

* this is quite unusual in `MeterBinder` implementations
* it works for Reactor Netty as it relies on a single registry, but supporting multiple registries would need a more complicated setup

Maybe this cache could be handled still as a Reactor Netty opinion and still leverage the binders provided here?

## Metrics names configuration

In my initial proposal I said that I would try to provide a way to customize metric names. Because of the number and structure of metrics, this PR doesn't allow that. Instead, I think that libraries and apps could use a `MeterFilter` to rewrite metric names on the fly with a custom prefix. Is that acceptable? See the next section for the actual metrics.

## Metrics

* `netty.allocator.memory.used` - Size of memory used by the allocator, in bytes
  Tags: `"id"` (unique id for the allocator), `"allocator.type"` (pooled, unpooled), `"memory.type"` (heap, direct)
* `"netty.allocator.memory.pinned"` - Size of memory used by allocated buffers, in bytes.
  Tags: `"id"` (unique id for the allocator), `"allocator.type"` (pooled, unpooled), `"memory.type"` (heap, direct)
* `"netty.allocator.pooled.arenas"` - Number of Arenas for a pooled allocator.
  Tags: `"id"` (unique id for the allocator), `"allocator.type"` (pooled, unpooled), `"memory.type"` (heap, direct)
* `"netty.allocator.pooled.cache.size"` - Size of the cache for a pooled allocator, in bytes.
  Tags: `"id"` (unique id for the allocator), `"allocator.type"` (pooled, unpooled), `"cache.type"` (normal, small)
* `"netty.allocator.pooled.threadlocal.caches"` - Number of ThreadLocal caches for a pooled allocator.
  Tags: `"id"` (unique id for the allocator), `"allocator.type"` (pooled, unpooled)
* `"netty.allocator.pooled.chunk.size"` - Size of memory chunks for a pooled allocator, in bytes.
  Tags: `"id"` (unique id for the allocator), `"allocator.type"` (pooled, unpooled)
* `"netty.eventexecutor.tasks.pending"` - Number of pending tasks in the event executor.
  Tags: `"name"` (unique name for the event executor)

This PR does not instrument the DNS infrastructure and I didn't dig much in that area. I'm not sure we can use a `DnsQueryLifecycleObserver` to record Timers, maybe only Counters are possible?

## Prometheus format sample

Here is a sample of Prometheus format I captured while testing the instrumentation on a running Netty server.

```
# HELP netty_eventexecutor_tasks_pending  
# TYPE netty_eventexecutor_tasks_pending gauge
netty_eventexecutor_tasks_pending{name="nioEventLoopGroup-3-2",} 0.0
netty_eventexecutor_tasks_pending{name="nioEventLoopGroup-3-1",} 0.0
netty_eventexecutor_tasks_pending{name="nioEventLoopGroup-3-5",} 0.0
netty_eventexecutor_tasks_pending{name="nioEventLoopGroup-3-4",} 0.0
netty_eventexecutor_tasks_pending{name="nioEventLoopGroup-3-3",} 0.0
# HELP netty_allocator_memory_used  
# TYPE netty_allocator_memory_used gauge
netty_allocator_memory_used{allocator_type="pooled",id="814169746",memory_type="heap",} 2.097152E7
netty_allocator_memory_used{allocator_type="pooled",id="814169746",memory_type="direct",} 2.097152E7
# HELP netty_allocator_pooled_cache_size  
# TYPE netty_allocator_pooled_cache_size gauge
netty_allocator_pooled_cache_size{allocator_type="pooled",cache_type="small",id="814169746",} 256.0
netty_allocator_pooled_cache_size{allocator_type="pooled",cache_type="normal",id="814169746",} 64.0
# HELP netty_allocator_memory_pinned  
# TYPE netty_allocator_memory_pinned gauge
netty_allocator_memory_pinned{allocator_type="pooled",id="814169746",memory_type="heap",} 5734400.0
netty_allocator_memory_pinned{allocator_type="pooled",id="814169746",memory_type="direct",} 0.0
# HELP netty_allocator_pooled_arenas  
# TYPE netty_allocator_pooled_arenas gauge
netty_allocator_pooled_arenas{allocator_type="pooled",id="814169746",memory_type="heap",} 16.0
netty_allocator_pooled_arenas{allocator_type="pooled",id="814169746",memory_type="direct",} 16.0
# HELP netty_allocator_pooled_threadlocal_caches  
# TYPE netty_allocator_pooled_threadlocal_caches gauge
netty_allocator_pooled_threadlocal_caches{allocator_type="pooled",id="814169746",} 5.0
# HELP netty_allocator_pooled_chunk_size  
# TYPE netty_allocator_pooled_chunk_size gauge
netty_allocator_pooled_chunk_size{allocator_type="pooled",id="814169746",} 4194304.0

```

shakuzen:
> Instead, I think that libraries and apps could use a `MeterFilter` to rewrite metric names on the fly with a custom prefix. Is that acceptable?

As someone who isn't a Netty expert, I think so. I'd love to hear from others who know more about Netty usage than me, though. In the past, the strong need to customize the metric name at the binder level came from there being multiple instances of the instrumented thing with potentially different tags on it, like ExecutorServiceMetrics. Will there be multiple instances of the binder in an app with a need to distinguish between the metrics from each? Netty being shaded was something I thought about, but if the package is different, these binders won't be usable anyway.

> This PR does not instrument the DNS infrastructure and I didn't dig much in that area.

I think it's fine to leave that as out-of-scope for this PR. If any users would like us to add this, please open an issue requesting it (or a pull request).

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetricsTests.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    void shouldHaveTaskPendingMetric() throws Exception {
+        Set<String> names = new LinkedHashSet<>();
+        DefaultEventLoopGroup eventExecutors = new DefaultEventLoopGroup();
+        eventExecutors.spliterator().forEachRemaining(eventExecutor -> {
+            if (eventExecutor instanceof SingleThreadEventExecutor) {
+                SingleThreadEventExecutor singleThreadEventExecutor = (SingleThreadEventExecutor) eventExecutor;
+                names.add(singleThreadEventExecutor.threadProperties().name());
+                new NettyEventExecutorMetrics(eventExecutor).bindTo(this.registry);
```
shakuzen, line 49: Without being a Netty expert, having users loop like this to bind each feels a bit weird to me. Would there be a reason a user would want metrics for some event executors in an EventExecutorGroup but not others? I wonder if we should use a higher level abstraction in NettyEventExecutorMetrics and add metrics for each executor for users rather than make them bind each one individually.

violetagg:
> In case you are interested, it is possible to add these additional metrics out of Netty event loops : [netty/netty#9080](https://github.com/netty/netty/pull/9080)
> 
> Sadly I have never completed it, but it should be simple
> 
> And the same, here: [netty/netty#11293 (comment)](https://github.com/netty/netty/pull/11293#issuecomment-855341497) To use a sentinel periodic task, per event loop, to measure busyness of it

Yep that's something that Reactor Netty has also in its backlog https://github.com/reactor/reactor-netty/issues/1433

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetricsTests.java
violetagg: I don't get this question.
Basically you have `EventLoopGroup` with `EventLoop`s. Every `EventLoop` has a name and a queue with pending tasks. What you are proposing is to have metrics on the `EventLoopGroup` is that correct?

shakuzen: I'm saying we should have metrics on all of the EventLoops like this test does, but without making users do `new NettyEventExecutorMetrics(eventExecutor).bindTo(this.registry)` for each individual `EventExecutor`. Instead we could take the `EventLoopGroup` as a parameter and register metrics for each `EventExecutor` so users only need to call, e.g. `new NettyEventExecutorMetrics(eventLoopGroup).bindTo(this.registry)` once rather than iterating over each element like now. Basically the question is does it make sense to make things as granular as they are now? It only makes sense to me if there is a case you would only want metrics for some EventLoops in a group but not all of them. If you always want all of them, we should just take the group as a parameter and iterate internally so users don't have to. Does that make more sense?

violetagg: I agree with you, it is just easier in Reactor Netty at this point to do it per `EventLoop`. Definitely you want metrics for all `EventLoop`s in the `EventLoopGroup`.

bclozel: I guess I could add a constructor variant that takes the entire `EventLoopGroup`?
This implementation was targeting the "lazy" case where the current `EventLoop` is found in the given channel during its initialization. Something like:

```
@Override
public void initChannel(SocketChannel channel) throws Exception {
  ByteBufAllocator alloc = channel.alloc();
  if (alloc instanceof ByteBufAllocatorMetricProvider) {
    // this concurrent check must be implemented by micrometer users
    if (isAllocatorInstrumented(alloc)) {
      new AllocatorMetrics(((ByteBufAllocatorMetricProvider)alloc)).bindTo(prometheusRegistry);
    }
  }
  // this concurrent check must be implemented by micrometer users
  if (isEventLoopInstrumented(channel.eventLoop())) {
    new EventExecutorMetrics(channel.eventLoop()).bindTo(prometheusRegistry);
  }
  channel.pipeline().addLast(new HttpRequestDecoder());
  channel.pipeline().addLast(new HttpResponseEncoder());
  channel.pipeline().addLast(new CustomHttpServerHandler());
}
```

violetagg:
>  I think that libraries and apps could use a MeterFilter to rewrite metric names on the fly with a custom prefix. Is that acceptable?

We will need to change the name if we want to keep backwards compatibility ...

bclozel:
> In case you are interested, it is possible to add these additional metrics out of Netty event loops : [netty/netty#9080](https://github.com/netty/netty/pull/9080)
> 
> Sadly I have never completed it, but it should be simple
> 
> And the same, here: [netty/netty#11293 (comment)](https://github.com/netty/netty/pull/11293#issuecomment-855341497) To use a sentinel periodic task, per event loop, to measure busyness of it

@franz1981 @violetagg I think this type of instrumentation really belongs in Netty directly. I'd be happy to expand metrics here once this API is available in Netty. We do maintain more involved instrumentations, but they usually rely on official extension points that are not likely to change.

bclozel:
I've just pushed additional changes in a separate commit that:

* change the "allocator.type" Tag for allocator metrics now hold the actual Java simple class name, e.g. "UnpooledByteBufAllocator"
* the binder for executor metrics now accept `Iterable<EventExecutor>`, which means both `EventLoopGroup` and `EventLoop` types are compatible.

I will squash this commit before merging this PR, once we're done with the review cycle.

franz1981:
@bclozel 

> I think this type of instrumentation really belongs in Netty directly. I'd be happy to expand metrics here once this API is available in Netty


For this one https://github.com/netty/netty/pull/9080 I think it's fine; but please consider that in Netty we don't modify public APIs unless marking them as `Unstable` (at least in Netty 4.1, for Netty 5, no idea)


> https://github.com/netty/netty/pull/11293#issuecomment-855341497

For this one, is different, because is something that Netty cannot provide nor decide by it's own, if the dynamic that allow it work is clear: if not, I can better explain it here instead.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyAllocatorMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+import io.netty.buffer.PooledByteBufAllocatorMetric;
+
+/**
+ * {@link MeterBinder} for Netty memory allocators.
+ *
+ * @author Brian Clozel
+ * @since 1.11.0
+ * @see NettyMeters
+ */
+public class NettyAllocatorMetrics implements MeterBinder {
```
shakuzen, line 34: Could we add a typical usage example to a JavaDoc? I don't know if it will be common knowledge for Netty users from the API defined here how/where to get the type to pass to the constructor.

shakuzen (approved):
Looks good to me. I would like to get rid of `id` on the allocator metrics if possible because it doesn't seem particularly meaningful to a user looking at the metrics. Due to my lack of experience with Netty, I don't know if there would ever be multiple instances of the same type of allocator in the same app to instrument. If not, it seems like we could get rid of `id`. However, it does seem theoretically possible for their to be multiple allocator instances of the same type.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyAllocatorMetrics.java
bclozel: @alesj we've added [code snippets in the reference documentation](https://micrometer.io/docs/ref/netty). Does this work for you?

Issue #3745, opened 2023-04-06
SignalFx Naming Convention creates too many regex objects


lenin-jaganathan:
**Describe the bug**
While exporting metrics (publish()), the SignalFx registry converts the tags into the SignalFx Dimension object. To stick to the SignalFx [tag naming convention](https://docs.splunk.com/Observability/metrics-and-metadata/metric-names.html#dimension-name-requirements), it does a lot of the regex (5 to be precise) matching which created too much garbage objects.

Scenario: 10,000 metrics (Note: this is 10,000 metrics and not 10,000 meters. For, one timer we will have 4 metrics generated in case of SignalFx.) exported by application, with an average of 8 tags per metric.

For the above scenario, SignalFx creates the following

- Create one Datapoint object for each metric (Since we have histograms as separate data points we end up creating 10,000 of these objects per minute)

- Creates one Dimension Object per dimension per metric. (Total = 80,000 per minute)

- 5 java.util.regex.Matcher objects per dimension. (Total = 400,000)

All the above are garbage collected.

Ref code: https://github.com/micrometer-metrics/micrometer/blob/b8ad913e53067908118fc97133ed6452b4796c42/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java#L81-L99

**To Reproduce**
Create a SignalFx Registry and register 10,000 meters with x tags each. Observe the memory increase right at the Step start time (when publishing happens). We will easily see (50,000 * x) number of regex objects created

**Expected behavior**
Garbage creation during publishing should be optimized to create as little as possible.

lenin-jaganathan:
Another way of doing it is in the new meter () construct (paying the one-time cost), where we can validate the tags and add them with the right value to the meter itself. This way we can completely eliminate the re-construction of Objects during publish time (which happens very very frequently)

lenin-jaganathan:
Well, using newMeter() construct will not be possible because of the implementation of registerMeterIfNecessary which checks if the meter exists. Will re-think about any better options.

PR #3747, opened 2023-04-07
Reduce Regex object creation in SignalFx Naming Convention

lenin-jaganathan:
This reduces the allocation of regex objects while publishing to SignalFx. This to an extent alters the existing behavior in the below cases,

Cases with the previous implementation:

- If a tag starts with "sf_", it strips it off. But it doesn't guard against something like "sf_sf_"
- If a tag starts with "\_" , it strips it off. If it starts with "__" it strips first "_" and adds an "a" to the beginning
- If a tag starts does not start with an (a-z) or (A-Z), it adds an "a" at the beginning.

This PR makes the behavior consistent to prefix with "a" for all the above cases, thus making the tags stick to naming standards.

- If a tag starts with "aws_|gcp_|azure_", earlier implementation, logged a message but this PR ignores it.

Reference Issue: https://github.com/micrometer-metrics/micrometer/issues/3745

### DIFF
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
@@ -75,7 +75,7 @@ public class SignalFxNamingConvention implements NamingConvention {
 
         if (conventionKey.startsWith("sf_") || !isAlphabet(conventionKey.charAt(0))) {
             logger.log(conventionKey
-                    + " doesn't adhere to SignalFx naming standards. Appending 'a' to the tag/dimension key.");
+                    + " doesn't adhere to SignalFx naming standards. Prefixing the tag/dimension key with 'a'.");
             conventionKey = "a" + conventionKey;
         }
 
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
@@ -106,7 +106,8 @@ public class SignalFxNamingConvention implements NamingConvention {
 
         if (conventionKey.startsWith("aws_") || conventionKey.startsWith("gcp_")
                 || conventionKey.startsWith("azure_")) {
-            logger.log("'" + conventionKey + "' (original name: '" + key + "') is not a valid tag key. "
+            String finalConventionKey = conventionKey;
+            logger.log(() -> "'" + finalConventionKey + "' (original name: '" + key + "') is not a valid tag key. "
                     + "Must not start with any of these prefixes: aws_, gcp_, or azure_. "
                     + "Please rename it to conform to the constraints. "
                     + "If it comes from a third party, please use MeterFilter to rename it.");
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
@@ -27,8 +27,8 @@ import java.util.regex.Pattern;
 /**
  * {@link NamingConvention} for SignalFx.
  *
- * See
- * https://developers.signalfx.com/metrics/data_ingest_overview.html#_criteria_for_metric_and_dimension_names_and_values
+ * @see <a href="https://docs.splunk.com/Observability/metrics-and-metadata/metric-names.html">
+ *     Naming conventions for metrics and dimensions</a>
  *
  * @author Jon Schneider
  * @author Johnny Lim
@@ -72,14 +72,14 @@ public class SignalFxNamingConvention implements NamingConvention {
     @Override
     public String tagKey(String key) {
         String conventionKey = delegate.tagKey(key);
+        conventionKey = PATTERN_TAG_KEY_DENYLISTED_CHARS.matcher(conventionKey).replaceAll("_");
 
-        if (conventionKey.startsWith("sf_") || !isAlphabet(conventionKey.charAt(0))) {
+        if (conventionKey.startsWith("sf_") || !Character.isLetter(conventionKey.charAt(0))) {
             logger.log(conventionKey
                     + " doesn't adhere to SignalFx naming standards. Prefixing the tag/dimension key with 'a'.");
             conventionKey = "a" + conventionKey;
         }
 
-        conventionKey = PATTERN_TAG_KEY_DENYLISTED_CHARS.matcher(conventionKey).replaceAll("_");
         return StringUtils.truncate(conventionKey, KEY_MAX_LENGTH);
     }
 
@@ -89,10 +89,6 @@ public class SignalFxNamingConvention implements NamingConvention {
     public String tagValue(String value) {
         String formattedValue = StringEscapeUtils.escapeJson(delegate.tagValue(value));
         return StringUtils.truncate(formattedValue, TAG_VALUE_MAX_LENGTH);
-    }
-
-    private static boolean isAlphabet(char c) {
-        return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
     }
 
 }
--- a/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
+++ b/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
@@ -27,9 +27,9 @@ import java.util.regex.Pattern;
 /**
  * {@link NamingConvention} for SignalFx.
  *
- * @see <a href="https://docs.splunk.com/Observability/metrics-and-metadata/metric-names.html">
- *     Naming conventions for metrics and dimensions</a>
- *
+ * @see <a href=
+ * "https://docs.splunk.com/Observability/metrics-and-metadata/metric-names.html"> Naming
+ * conventions for metrics and dimensions</a>
  * @author Jon Schneider
  * @author Johnny Lim
  */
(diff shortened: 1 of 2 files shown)


shakuzen:
> * If a tag starts with "aws_|gcp_|azure_", earlier implementation, logged a message but this PR ignores it.

Is that no longer a limitation in SignalFx?

lenin-jaganathan:
It still is! But the earlier implementation doesn't do anything except logging. The problem with this is this is done for every metric publishing interval. I would need help from someone from SignalFx to weigh in on these changes (maybe @bogdandrutu ?).

But If the logging behavior is something that is expected, then I would do that while creating a new meter which is done only once. (in the newMeter() constructs and would not do that in every publishing cycle.)

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
```diff
@@ -106,4 +91,8 @@ public String tagValue(String value) {
         return StringUtils.truncate(formattedValue, TAG_VALUE_MAX_LENGTH);
     }
 
+    private static boolean isAlphabet(char c) {
```
atoulme, line 94: Can we use `Character.isLetter` instead?

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
```diff
@@ -81,21 +73,14 @@ public String name(String name, Meter.Type type, @Nullable String baseUnit) {
...
+                    + " doesn't adhere to SignalFx naming standards. Appending 'a' to the tag/dimension key.");
             conventionKey = "a" + conventionKey;
         }
-        if (PATTERN_TAG_KEY_DENYLISTED_PREFIX.matcher(conventionKey).matches()) {
-            String finalConventionKey = conventionKey;
-            logger.log(() -> "'" + finalConventionKey + "' (original name: '" + key + "') is not a valid tag key. "
-                    + "Must not start with any of these prefixes: aws_, gcp_, or azure_. "
-                    + "Please rename it to conform to the constraints. "
-                    + "If it comes from a third party, please use MeterFilter to rename it.");
-        }
```
atoulme, line 97: Rather than deleting this, maybe we can apply a heuristic to rename those? Could we add them to line 76 with the following condition:
`conventionKey.startsWith("aws_") || conventionKey.startsWith("gcp_") || conventionKey.startsWith("azure_")`

implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxNamingConventionTest.java
```diff
@@ -31,8 +31,11 @@ class SignalFxNamingConventionTest {
 
     @Test
     void tagKey() {
-        assertThat(convention.tagKey("_boo")).isEqualTo("boo");
-        assertThat(convention.tagKey("sf_boo")).isEqualTo("boo");
+        assertThat(convention.tagKey("_boo")).isEqualTo("a_boo");
+        assertThat(convention.tagKey("__boo")).isEqualTo("a__boo");
+        assertThat(convention.tagKey("sf_boo")).isEqualTo("asf_boo");
+        assertThat(convention.tagKey("sf__boo")).isEqualTo("asf__boo");
+        assertThat(convention.tagKey("sf_sf_boo")).isEqualTo("asf_sf_boo");
```
atoulme, line 38: I added this test:
`assertThat(convention.tagKey("sf.boo")).isEqualTo("sf_boo");`
So it seems possible to inadvertently create `sf_` prefixed dimensions.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
```diff
@@ -81,21 +73,14 @@ public String name(String name, Meter.Type type, @Nullable String baseUnit) {
...
-        return StringUtils.truncate(conventionKey, KEY_MAX_LENGTH); // 1
+
+        conventionKey = PATTERN_TAG_KEY_DENYLISTED_CHARS.matcher(conventionKey).replaceAll("_");
```
atoulme, line 82: I would move this up - see below for the test case I added.

atoulme (commented):
Thank you for your contribution! I work for Splunk and have reviewed this PR. Please see my feedback in comments.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
```diff
@@ -81,21 +73,14 @@ public String name(String name, Meter.Type type, @Nullable String baseUnit) {
     public String tagKey(String key) {
         String conventionKey = delegate.tagKey(key);
 
-        conventionKey = START_UNDERSCORE_PATTERN.matcher(conventionKey).replaceAll(""); // 2
-        conventionKey = SF_PATTERN.matcher(conventionKey).replaceAll(""); // 2
-
-        conventionKey = PATTERN_TAG_KEY_DENYLISTED_CHARS.matcher(conventionKey).replaceAll("_");
-        if (!START_LETTERS_PATTERN.matcher(conventionKey).matches()) { // 3
+        if (conventionKey.startsWith("sf_") || !isAlphabet(conventionKey.charAt(0))) {
```
bogdandrutu, line 76: This does not generate the same output as before. For example in case of `START_UNDERSCORE_PATTERN` that was just removed not adding an "a", same for `SF_PATTERN`. Do I miss anything?

lenin-jaganathan: Yes, that is right. And I tried to lay that down explicitly in the PR description that this PR alters some of the existing behavior for better memory optimization. If you guys have a better approach to this, I would be more happy to wait for that implementation.

More about this is described in this issue - https://github.com/micrometer-metrics/micrometer/issues/3745

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
lenin-jaganathan: I explicitly decided against that for the reason that it will return true for Unicode letters.

E.g: Character.isLetter('à') will also return true.


Having said that, if we are moving 
> conventionKey = PATTERN_TAG_KEY_DENYLISTED_CHARS.matcher(conventionKey).replaceAll("_");

as for the first condition, I guess this will replace all those with "_" and then I can rely on Character.isLetter().

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
lenin-jaganathan: I would rather rely on you guys on this. Just a note, earlier implementations never did anything for these prefixes other than a logging message. 

Do you want to add 'a' for these prefixes also?

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
```diff
@@ -77,25 +69,23 @@ public String name(String name, Meter.Type type, @Nullable String baseUnit) {
     // letters, numbers, underscores _ and hyphens - . This requirement is expressed in
     // the following regular expression:
     // ^[a-zA-Z][a-zA-Z0-9_-]*$
+    /**
+     * @see <a href=
+     * "https://docs.splunk.com/Observability/metrics-and-metadata/metric-names.html#dimension-names-and-value-standards">Dimension
```
shakuzen, line 74: I think it'd be better to move this to the class-level JavaDoc and get rid of the old link that's currently there.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
bogdandrutu: I don't have a "better" approach, but I do have a recommendation to try to keep the current behavior if possible. For example by replacing the use of `SF_PATTERN.matcher(conventionKey).replaceAll("")` with something equivalent with less memory usage like:

```java
private static final SF_PREFIX = "sf_";
....

if (conventionKey.startsWith(SF_PREFIX)) {
  conventionKey = conventionKey.substring(SF_PREFIX.length())
}
```

lenin-jaganathan: @bogdandrutu I fixed the code to retain existing behavior. I used recursive matching to eliminate any "sf_" and "_" as prefixes.
The basic idea behind this change is that the tag key convention should be as simple as possible when the key adheres to the SignalFx convention. Please review the changes and feel free to comment on this.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
atoulme: I like what @bogdandrutu said in a different comment - since you're intent on performance gains, let's just focus on those and keep the behavior the same for now.

atoulme (approved):

shakuzen (approved):
If everyone's happy with how things are now, we're ready to merge. Since we ended up preserving the prior behavior (or at least the behavior users should have relied upon) and have more tests now, I would be open to merging this into 1.9.x so we can release it in patch versions.

lenin-jaganathan:
That would be really great if we can get this on 1.9.x

shakuzen:
> That would be really great if we can get this on 1.9.x

Sure. Please rebase onto and target the PR at `1.9.x`

lenin-jaganathan:
@shakuzen I have re-based the change onto 1.9.x and updated the target branch as well.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
```diff
@@ -80,22 +72,47 @@ public String name(String name, Meter.Type type, @Nullable String baseUnit) {
...
+                    + " doesn't adhere to SignalFx naming standards. Prefixing the tag/dimension key with 'a'.");
             conventionKey = "a" + conventionKey;
         }
-        if (PATTERN_TAG_KEY_DENYLISTED_PREFIX.matcher(conventionKey).matches()) {
-            String finalConventionKey = conventionKey;
-            logger.log(() -> "'" + finalConventionKey + "' (original name: '" + key + "') is not a valid tag key. "
+
+        if (conventionKey.startsWith("aws_") || conventionKey.startsWith("gcp_")
+                || conventionKey.startsWith("azure_")) {
+            logger.log("'" + conventionKey + "' (original name: '" + key + "') is not a valid tag key. "
```
bogdandrutu, line 109: This may be slower and less performant than the previous implementation when logging is not available.

bogdandrutu (approved):

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxNamingConvention.java
lenin-jaganathan: Thanks for catching the change, I had changed this to the existing behaviour.

shakuzen:
Thank you all for the reviews.

Issue #3717, opened 2023-03-25
Async support for HttpSender


lenin-jaganathan:
**Please describe the feature request.**
HttpUrlConnectionSender currently sends data in a synchronous fashion I,e the metrics are sent and we wait for the response blocking the thread. Since the publishing happens on a dedicated thread this might seem problematic when combined with timeouts for long pending requests. But, when applications have thousands of metrics that need to be batched and sent, which might result in metrics being measured for different intervals rather than for the same step.

Scenario:
Assume an application has 15,000 meters in the registry and the registry has a default batch size of 10,000. The scheduler calls the publishing method at 00:00:00. Reading meters is a faster task and we would complete reading 10,000 meters in x milliseconds but most of the registries call HttpUrlConnectionSender.send() for this first batch. Since this is a blocking I/O this depends on network latency and other external factors. And the reading of the next 5000 meters is delayed by the response time of the first HTTP request. This ensures that the first and second batch of meters doesn't measure the measurement for the same period of time.

This is mostly true for most of the registries that export each Histogram bucket as a Gauge. 1000 timers with 25 buckets can easily take up 25k meters and reading this in batches and sending them synchronously will take away the Step concept since the data sent is no more normalized to a fixed window.

**Rationale**
To make sure that we are reporting the measurements for the same step period. (more so that we minimize the gap that is there in the reading of data).

Probably, something like an AsyncHttpSender which will return a CompleteableFuture instead of blocking the thread. And all these futures can be awaited once all of them are submitted.

shakuzen:
I'm not sure we need an async abstraction to deal with the problem you mentioned. We just need to separate the steps of preparing the data to send and sending it, no?

lenin-jaganathan:
https://github.com/micrometer-metrics/micrometer/issues/2818 is not a completely related issue. But, [this comment](https://github.com/micrometer-metrics/micrometer/issues/2818#issuecomment-1334845320) of mine again points to registry implementations taking longer time to complete the publishing.

shakuzen:
The batching feature was implemented because some backends have a limit on the number of metrics that can be in a single request.

> But if you are referring to just constructing the payload at once but just splitting them into multiple requests, I guess that will lead to duplicate implementations in most of the Registries whereas a client that can return a CompletableFuture immediately can help in such scenarios.

[...]

PR #3750, opened 2023-04-10
Start publish at randomized time in next step

shakuzen:
This avoids different instances calling publish at the same time by scheduling them at a time randomly chosen within the next step upon start being called. It will avoid scheduling publishing at the last 20% of a step interval to avoid the issue described in gh-1218 where publishing could take long enough to finish in a subsequent step after it started. An additional background thread is added for StepMeterRegistry that will poll all meters at the start of the step, separate from when publishing happens. This is so the values published are the ones recorded in the previous step. We need rollCount to be called for step meters as early as possible in the step to minimize the number of recordings reported for the wrong step. This is not perfect, but it is no worse than the status quo.

Resolves gh-2818

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -22,6 +22,7 @@ import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.util.TimeUtils;
 
+import java.util.Random;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
@@ -31,6 +32,11 @@ import java.util.concurrent.atomic.AtomicBoolean;
 public abstract class PushMeterRegistry extends MeterRegistry {
 
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(PushMeterRegistry.class);
+
+    // Make sure we do not schedule publishing at the end of a step interval by only
+    // scheduling in the beginning X percent of the interval.
+    // VisibleForTesting
+    static final double PERCENT_RANGE_OF_RANDOM_PUBLISHING_OFFSET = 0.8;
 
     private final PushRegistryConfig config;
 
@@ -97,9 +103,8 @@ public abstract class PushMeterRegistry extends MeterRegistry {
                     + TimeUtils.format(config.step()));
 
             scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
-            // time publication to happen just after StepValue finishes the step
             long stepMillis = config.step().toMillis();
-            long initialDelayMillis = stepMillis - (clock.wallTime() % stepMillis) + 1;
+            long initialDelayMillis = calculateInitialDelay();
             scheduledExecutorService.scheduleAtFixedRate(this::publishSafely, initialDelayMillis, stepMillis,
                     TimeUnit.MILLISECONDS);
         }
@@ -119,6 +124,16 @@ public abstract class PushMeterRegistry extends MeterRegistry {
             publishSafely();
         }
         super.close();
+    }
+
+    // VisibleForTesting
+    long calculateInitialDelay() {
+        long stepMillis = config.step().toMillis();
+        Random random = new Random();
+        long randomOffsetWithinStep = (long) (stepMillis * random.nextDouble()
+                * PERCENT_RANGE_OF_RANDOM_PUBLISHING_OFFSET);
+        long offsetToStartOfNextStep = stepMillis - clock.wallTime() % stepMillis;
+        return offsetToStartOfNextStep + 2 + randomOffsetWithinStep;
     }
 
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -132,7 +132,7 @@ public abstract class PushMeterRegistry extends MeterRegistry {
         Random random = new Random();
         long randomOffsetWithinStep = (long) (stepMillis * random.nextDouble()
                 * PERCENT_RANGE_OF_RANDOM_PUBLISHING_OFFSET);
-        long offsetToStartOfNextStep = stepMillis - clock.wallTime() % stepMillis;
+        long offsetToStartOfNextStep = stepMillis - (clock.wallTime() % stepMillis);
         return offsetToStartOfNextStep + 2 + randomOffsetWithinStep;
     }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -35,7 +35,6 @@ public abstract class PushMeterRegistry extends MeterRegistry {
 
     // Schedule publishing in the beginning X percent of the step to avoid spill-over into
     // the next step.
-    // VisibleForTesting
     private static final double PERCENT_RANGE_OF_RANDOM_PUBLISHING_OFFSET = 0.8;
 
     private final PushRegistryConfig config;
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -33,10 +33,10 @@ public abstract class PushMeterRegistry extends MeterRegistry {
 
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(PushMeterRegistry.class);
 
-    // Make sure we do not schedule publishing at the end of a step interval by only
-    // scheduling in the beginning X percent of the interval.
+    // Schedule publishing in the beginning X percent of the step to avoid spill-over into
+    // the next step.
     // VisibleForTesting
-    static final double PERCENT_RANGE_OF_RANDOM_PUBLISHING_OFFSET = 0.8;
+    private static final double PERCENT_RANGE_OF_RANDOM_PUBLISHING_OFFSET = 0.8;
 
     private final PushRegistryConfig config;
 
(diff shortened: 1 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -121,4 +126,14 @@ public void close() {
         super.close();
     }
 
+    // VisibleForTesting
+    long calculateInitialDelay() {
+        long stepMillis = config.step().toMillis();
+        Random random = new Random();
```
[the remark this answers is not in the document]
shakuzen: @sonatype-lift ignore

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -104,6 +110,26 @@ protected DistributionStatisticConfig defaultHistogramConfig() {
...
     }
 
+    @Override
+    public void start(ThreadFactory threadFactory) {
+        super.start(threadFactory);
+
+        if (config.enabled()) {
+            this.meterPollingService = Executors.newSingleThreadScheduledExecutor(threadFactory);
+
+            this.meterPollingService.scheduleAtFixedRate(this::pollMetersToRollover, getInitialDelay(),
```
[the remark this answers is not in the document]
shakuzen: @sonatype-lift ignore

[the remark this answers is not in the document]
shakuzen: @sonatype-lift ignore

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -97,9 +103,8 @@ public void start(ThreadFactory threadFactory) {
                     + TimeUtils.format(config.step()));
 
             scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
-            // time publication to happen just after StepValue finishes the step
             long stepMillis = config.step().toMillis();
-            long initialDelayMillis = stepMillis - (clock.wallTime() % stepMillis) + 1;
+            long initialDelayMillis = calculateInitialDelay();
```
lenin-jaganathan, line 107: This in a way enforces registries to complete the publish() in 20% of the step time. And with the issues referenced in https://github.com/micrometer-metrics/micrometer/issues/3717, this could be a serious issue for users, unless there is an option to preserve the existing behavior.

If the step registries can publish within 20% of the step time, then I believe that alone would ensure that we are reading the data for right step.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -116,4 +142,22 @@ public void close() {
...
 
+    /**
+     * This will poll the values from meters, which will cause a roll over for Step-meters
+     * if past the step boundary. This gives some control over when roll over happens
+     * separate from when publishing happens.
+     */
+    // VisibleForTesting
+    void pollMetersToRollover() {
+        this.getMeters()
+            .forEach(m -> m.match(gauge -> null, Counter::count, Timer::count, DistributionSummary::count,
```
lenin-jaganathan, line 153: I see the point in just calling count on timers will be faster, but I also could see this can get trickier where the timer snapshots(max/histograms especially if things are step based) will not be in sync with the counts and totals.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: If you aren't publishing within 20% of the step time, that's probably a separate issue to look into. I would think there must be something wrong with the registry implementation, high cardinality, not enough CPU/memory, or not enough resources for your metrics backend. Prior to Micrometer 1.7, we published at any arbitrary time in the step, and I don't think anyone reported publishing taking so long. If publishing takes a long time, I don't think the answer is to publish at the beginning of the step to give the most time possible. That just exacerbates the issue because it will slow down shared resources (network, metrics backend) when all of your instances try to use them at the same time.

> If the step registries can publish within 20% of the step time, then I believe that alone would ensure that we are reading the data for right step.

Not if we're scheduling at the end of the step. That's the whole point. Registries still need to publish in a reasonable amount of time, and we need to look into cases where that isn't happening.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
shakuzen: I think I might be missing the point you're making. Max and histograms rotate on writing and reading, so they don't have the same need for this as count/total do. The max and histograms rotation are not aligned to the epoch step anyway.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
lenin-jaganathan: > Not if we're scheduling at the end of the step.

I agree with this. But, what I meant is if 20% is reasonable enough time to publish we can do the publishing anytime. But, yeah that will bring the issue of Step's not reporting data for the particular step. Fair enough.

But, I would probably have some config to have the old behaviour for at least next 1 or 2 versions. I will leave you guys to decide on that.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
lenin-jaganathan: Yeah, they are. Oh, wait this is on StepRegistry which should be fine. Thanks for the clarification. I might need to do some testing with the case of SignalFxTimer which is an abstract one in the Step registry. I will get back to that.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -121,4 +126,14 @@ public void close() {
         super.close();
     }
 
+    // VisibleForTesting
+    long calculateInitialDelay() {
+        long stepMillis = config.step().toMillis();
+        Random random = new Random();
+        long randomOffsetWithinStep = (long) (stepMillis * random.nextDouble()
+                * PERCENT_RANGE_OF_RANDOM_PUBLISHING_OFFSET);
+        long offsetToStartOfNextStep = stepMillis - clock.wallTime() % stepMillis;
```
jonatan-ivanov, line 135: Can we make it more obvious?
```suggestion
        long offsetToStartOfNextStep = stepMillis - (clock.wallTime() % stepMillis);
```

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -121,4 +126,14 @@ public void close() {
...
+        return offsetToStartOfNextStep + 2 + randomOffsetWithinStep;
```
jonatan-ivanov, line 136: Can we add a comment that says something like:
Wait till the next step + a little more randomly somewhere between 0 and 80% of the step interval + 2ms?

Btw why 2ms?

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -116,4 +142,22 @@ public void close() {
...
+                    meter -> null, meter -> null, FunctionCounter::count, FunctionTimer::count, meter -> null));
+    }
+
+    private long getInitialDelay() {
+        long stepMillis = config.step().toMillis();
+        // schedule one millisecond into the next step
+        return stepMillis - clock.wallTime() % stepMillis + 1;
```
jonatan-ivanov, line 160: Can we make it more obvious?
```suggestion
        return stepMillis - (clock.wallTime() % stepMillis) + 1;
```

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -116,4 +142,22 @@ public void close() {
         super.close();
     }
 
+    /**
+     * This will poll the values from meters, which will cause a roll over for Step-meters
+     * if past the step boundary. This gives some control over when roll over happens
+     * separate from when publishing happens.
```
jonatan-ivanov, line 148: I'm thinking if we can get into trouble when the rollover and the publisher schedulers are "too close" to each other and since there is some indeterminism in them (they are not scheduling at the exact time), at one schedule the order of them are the opposite so we lose one step worth of data: publish, rollover, rollover, publish.

jonatan-ivanov (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
lenin-jaganathan: I agree that this can be reordered. But is that really a worry? Maybe or May not be. 

The calls to poll() of a StepMeter is meant to be idempotent for a give step. We take a snapshot once and use that for calls until next step approach which doesn't care if it is publisher or pollService. The advantage with pollService is that it will make these measurements align more to step i,e since the pollService only rollsOver the step, it is meant to be faster and finish rolling over all the meters within some ms so the data polled during publishing is exactly for 00:00:00.xxx to 00:00:01.yyy. Here xxx the completion of last poll and yyy is the completion of current poll.

Please ignore, if your question is completely different 🙂. 

@shakuzen  Do you think it makes sense to have StepMeter interface have something like rollOver() method which will specifically do that to make things more explicit?  or may be that is too much for what we are doing ?

shakuzen: I talked to @jonatan-ivanov about this a little bit ago. This is fine since polling things multiple times in the same step interval has no effect beyond the first time. It doesn't matter which of the scheduled tasks happens first. We just want polling to happen at the beginning of the step.

> Do you think it makes sense to have StepMeter interface have something like rollOver() method which will specifically do that to make things more explicit? or may be that is too much for what we are doing ?

I considered it, but I think it's too much for now. If we see a need for it later we can introduce it, but hopefully it is clear enough now what we are doing and why.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: > why 2ms? Because the rollover is +1?

Yes.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: We try very hard to not break compatibility, which means it is very hard for us to remove something once exposed as public API. This is one of the reasons I try very hard to not make things configurable that in theory we believe users should not need to adjust. I would like to get more feedback on this with RC1 and we can consider making it configurable in GA or the next release if there's a need. Let's look into it more.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: We can, but I think Lift might complain about unnecessary parentheses 😄

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: I added a polish commit with a comment explaining. I also changed the implementation slightly so that the 2ms isn't increasing the latest time a publish can be scheduled. It shouldn't matter for anything but very small steps (like the test case is using).

ttddyy: nit: jitter might be a good variable name for the random value.

PR #3764, opened 2023-04-12
Adding observeChecked alias methods to avoid ambiguous method

quaff:
`observation.observeChecked(proceedingJoinPoint::proceed)` cannot be compiled due to "The method observeChecked(Observation.CheckedRunnable<Throwable>) is ambiguous for the type Observation"

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -520,6 +520,10 @@ public interface Observation extends ObservationView {
         });
     }
 
+    default <E extends Throwable> void observeCheckedRunnable(CheckedRunnable<E> checkedRunnable) throws E {
+        observeChecked(checkedRunnable);
+    }
+
     default <E extends Throwable> CheckedRunnable<E> wrapChecked(CheckedRunnable<E> checkedRunnable) throws E {
         return () -> observeChecked(checkedRunnable);
     }
@@ -567,6 +571,11 @@ public interface Observation extends ObservationView {
     @Nullable
     default <T, E extends Throwable> T observeChecked(CheckedCallable<T, E> checkedCallable) throws E {
         return observeCheckedWithContext((context) -> checkedCallable.call());
+    }
+
+    @Nullable
+    default <T, E extends Throwable> T observeCheckedCallable(CheckedCallable<T, E> checkedCallable) throws E {
+        return observeChecked(checkedCallable);
     }
 
     default <T, E extends Throwable> CheckedCallable<T, E> wrapChecked(CheckedCallable<T, E> checkedCallable) throws E {


jonatan-ivanov:
Hi,

Thank you for the PR!
Can I have a few questions around it?

To me it seems what you are doing is the opposite of overloading. `observeChecked` has overloads it seems you don't want that but have a separate method.

Have you checked `ObservedAspect`? Could it work for you?

I'f I'm trying to do what you described (using `org.aspectj.lang.ProceedingJoinPoint`), I'm getting a different error:
```
Reference to 'proceed' is ambiguous, both 'proceed()' and 'proceed(Object[])' match
```

Both of them should be resolved by `.observeChecked(() -> pjp.proceed());` as you can see in https://github.com/micrometer-metrics/micrometer/blob/ac917e2cd3d9b0da37a5eceb3a2f403a6391e1db/micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java#L150

Another thing I'm not getting is how `proceed` can be a (Checked)`Runnable`? It should return some result so it should be a (Checked)`Callable`.

Also, I don't think we should have an `observeChecked(CheckedRunnable)` and `observeCheckedRunnable(CheckedRunnable)` without the former being deprecated.

quaff:
I need to explicit cast to `(Observation.CheckedCallable<Object, Throwable>)`
```java
Object result = Observation.createNotStarted("test", observationRegistry).observeChecked((Observation.CheckedCallable<Object, Throwable>) pjp::proceed);
```

eclipse jdt:

javac:

jonatan-ivanov:
I get that from the description but your PR is only about `CheckedRunnable`. Also, this does not answer any of my other questions above, could you please help me with those too?
Also, I don't think you need to cast, you can do this:
```java
.observeChecked(() -> pjp.proceed()); 
```

quaff:
> I get that from the description but your PR is only about `CheckedRunnable`. Also, this does not answer any of my other questions above, could you please help me with those too? Also, I don't think you need to cast, you can do this:
> 

`.observeChecked(() -> pjp.proceed())` works but method reference is better choice, that's the purpose of this PR.

jonatan-ivanov:
Could you please answer my other questions above too?

quaff:
> Could you please answer my other questions above too?

1. Yes I'm doing opposite of overloading, I've amended the description.
2. `ObservedAspect` is using `observeChecked(() -> pjp.proceed())` not wanted `observeChecked(pjp::proceed)`
3. I'm fine with dropping `observeCheckedRunnable(CheckedRunnable)` from this PR, `observeChecked(CheckedRunnable)` can be used for method reference.
```java
observeChecked(action::execute); // works fine

interface Action {
  void execute() throws Exception;
}
```

marcingrzejszczak:
Hey @quaff, thank you for the PR. We've discussed this internally and decided that we don't want to add more methods to the current API since you already have 2 options, one involves using a lambda and no casting, the other method reference and casting.

PR #3766, opened 2023-04-12
Reduces overall memory allocation in DynatraceExporterV2 while exporting

belks:
In applications with a lot of metrics the Dynatrace exporter collects all lines to be exported in one list before sending them in batches. This allocates quite a lot of unnecessary memory.
This PR changes this behavior so that batches are sent immediately and already sent lines are not kept in memory.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
@@ -17,9 +17,9 @@ package io.micrometer.dynatrace.v2;
 
 import com.dynatrace.metric.util.*;
 import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.distribution.HistogramSnapshot;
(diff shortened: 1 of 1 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -139,9 +136,22 @@ public void export(List<Meter> meters) {
         // Lines that are too long to be ingested into Dynatrace, as well as lines that
         // contain NaN or Inf values are not returned from "toMetricLines", and are
         // therefore dropped.
```
shakuzen, line 138: This comment should probably also be moved to where `toMetricLines` is being called.

shakuzen (approved):
Thanks for the pull request. @pirgeo let us know if you have any concerns with this change.

pirgeo (approved):
No concerns from my side!

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
belks: Done. Moved the comment closer to ` toMetricLines`

PR #3768, opened 2023-04-13
Polish "Add metrics support for Netty 4.x"

izeye:
This PR polishes the "Add metrics support for Netty 4.x" changes a bit.

See gh-3742

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -99,7 +99,7 @@ dependencies {
     optionalApi 'org.apache.tomcat.embed:tomcat-embed-core'
     optionalApi 'org.glassfish.jersey.core:jersey-server'
     optionalApi 'io.grpc:grpc-api'
-	optionalApi 'io.netty:netty-transport'
+    optionalApi 'io.netty:netty-transport'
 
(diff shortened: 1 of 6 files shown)


jonatan-ivanov (approved):

izeye:
@jonatan-ivanov Oop, sorry. I fixed them now.

franz1981:
@jonatan-ivanov what about the `allocatorId` ? I didn't yet opened an issue but would be better to:
- use a `CopyOnWriteArraySet` to verify a certain allocator isn't observed yet
- and a static atomic long to create runtime fake identity for the allocator just registered

bclozel:
@franz1981 the current design assumes that there are various ways to instrument Netty (and the allocator) here:

* manually during startup, when resources like allocators are manually set and known. In this case, there would be no need to maintain the list of already registered allocators
* at runtime, lazily, when an allocator is seen during channel init. In this case, it's a good idea to maintain such a list and prevent duplicate instrumentations

We can't choose this for libraries, this is why we chose to remain neutral in Micrometer's instrumentation. Does this make sense?

PR #3776, opened 2023-04-20
Compile sources with '-parameters' to retain method parameter names

quaff:
Fix warning raised by spring
```
o.s.c.LocalVariableTableParameterNameDiscoverer WARN Using deprecated '-debug' fallback for parameter name resolution. Compile the affected code with '-parameters' instead or avoid its introspection: io.micrometer.core.aop.CountedAspect
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,7 +90,7 @@ subprojects {
         tasks {
             compileJava {
                 options.encoding = 'UTF-8'
-                options.compilerArgs << '-Xlint:unchecked' << '-Xlint:deprecation'
+                options.compilerArgs << '-Xlint:unchecked' << '-Xlint:deprecation' << '-parameters'
 
                 sourceCompatibility = javaTargetVersion
                 targetCompatibility = javaTargetVersion


shakuzen:
It'd be good to understand when/why this is happening. It may be that Spring should not even be doing introspection in this case. Compiling with `-parameters` masks the issue of unnecessary introspection. I see https://github.com/spring-projects/spring-framework/issues/29612 and https://github.com/spring-projects/spring-framework/issues/30103 were fixed as recently as 6.0.7. If you still get the issue on the latest framework version, can you get a stacktrace or share a minimal sample that reproduces the issue?

quaff:
```java
@SpringBootTest
@ExtendWith(OutputCaptureExtension.class)
public class ApplicationTests {
    
    @Test
    void shouldNotOutputDeprecatedLogging(CapturedOutput output){
        assertThat(output).doesNotContain("Compile the affected code with '-parameters' instead or avoid its introspection: io.micrometer.core.aop.CountedAspect"); // failed
    }
}

```
```groovy
plugins {
	id 'java'
	id 'org.springframework.boot' version '3.0.5'
	id 'io.spring.dependency-management' version '1.1.0'
}

group = 'com.example'
version = '0.0.1-SNAPSHOT'
sourceCompatibility = '17'

repositories {
	mavenCentral()
}

dependencies {
	implementation 'org.springframework.boot:spring-boot-starter-aop'
	implementation 'org.springframework.boot:spring-boot-starter-web'
	implementation 'org.springframework.boot:spring-boot-starter-actuator'
	testImplementation 'org.springframework.boot:spring-boot-starter-test'
}

tasks.named('test') {
	useJUnitPlatform()
}
```
[countedaspect.zip](https://github.com/micrometer-metrics/micrometer/files/11283490/countedaspect.zip)

quaff:
It's weird that `TimedAspect` is not affected, it's very similar with `CountedAspect`.

shakuzen:
Thank you for the quick response and the reproducer. I'm asking someone from the Spring Framework team to take a look to determine if there is something to fix on the framework side.

> It's weird that `TimedAspect` is not affected, it's very similar with `CountedAspect`.

I agree, and I'm not sure why it's different.

quaff:
It triggered by `@Around("@annotation(counted)")`, the name `counted` is binding to method parameter `Counted counted`, method parameter inspection is required,`-debug` is deprecated, we should switch to `-parameter`.

https://github.com/micrometer-metrics/micrometer/blob/9b61c23a925130d039171f43f69ec14299507bec/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java#L185-L186

shakuzen:
Thank you for the analysis. I think we can change the aspect code so that referencing a named parameter is not necessary. I don't see any reason why we needed to do it that way.

shakuzen:
@quaff with https://github.com/micrometer-metrics/micrometer/pull/3780 merged, could you try with the latest Micrometer snapshot `1.10.7-SNAPSHOT` to see if that resolves the issue for you?

jonatan-ivanov:
Superseded by #3780

Issue #3781, opened 2023-04-20
Support primitive types for Observation tags


nikoncode:
Modern tracing backends support boolean, double and long values in attribute, it might be used for future filtering. For example: http.retry_count > 3.

I added support of these types in tracing library https://github.com/micrometer-metrics/tracing/issues/234, but I want to make this work starting from Observation (Convention) layer. It requires modification of KeyValue class.

maciejwalkowiak:
Did you mean @marcingrzejszczak?

nikoncode:
Yes, I am sorry.

maciejwalkowiak:
No worries it happens systematically 😊

marcingrzejszczak:
@nikoncode we are looking into this. Please be patient

marcingrzejszczak:
We will postpone the decision  to merge this until the next minor release. We need to think this through

nikoncode:
If you want my opinion:
* we probably should even start exposing types from KeyName level, because it's used to declare attribute.
* since types "kind of" exporter specific, we can try to make all places which operates with types abstract (for example setTags), to be able to support various types for different backends or use conversion to string as fallback.
* we can get inspiration from opentelemetry https://github.com/open-telemetry/opentelemetry-java/blob/9f600741edbc0094ca9828b8873cf854a02f1917/api/all/src/main/java/io/opentelemetry/api/common/AttributeKey.java#L22, they Key is parameterized and use types as enum for future processing.
* I think it's not an option to prevent creation of non-string attributes, because:
  * tracing specifications and modern backends support DSL which might be based on attribute type (Jaeger, Tempo, etc.). <> for numbers, and things like IN for array.
  * I like micrometer-tracing concept and want to stick to the spring boot 3 defaults to instrument in-house transport clients with variant of opentelemetry spec.
* I can try to assist with implementation, because I feel direct pain with it. But it's not possible without strategic decision from your side. And thanks for letting me implement this at least on span level, I will use it to avoid call of specific opentelemetry methods.

marcingrzejszczak:
> I can try to assist with implementation, because I feel direct pain with it. But it's not possible without strategic decision from your side. And thanks for letting me implement this at least on span level, I will use it to avoid call of specific opentelemetry methods.

No problem. Can you elaborate more about the direct pain of not having this? Backends should allow you to do comparison operations like `>` or `<` even on string values :thinking:

marcingrzejszczak:
[...]

PR #3782, opened 2023-04-20
support long, double and boolean for KeyValue #3781

nikoncode:


nikoncode:
It maybe better to have type starting from KeyName level, because it's used to declare some attributes. I like how it's done in otel, but might be hard to do the same because of compatibility reasons.

https://github.com/open-telemetry/opentelemetry-java/blob/9f600741edbc0094ca9828b8873cf854a02f1917/api/all/src/main/java/io/opentelemetry/api/common/AttributeKey.java#L22

shakuzen:
Thanks for the pull request. I'm closing until we better understand the negative impact of not providing this. See comments in #3781

viacheslav-dobrynin:
@shakuzen, hello!

We are faced with the problem of the absence of types in micrometer.
More specifically, langfuse expects an integer type for the tag, and when it receives a string, it does not work as expected.

See https://github.com/orgs/langfuse/discussions/10097#discussioncomment-15059893 for details.

Issue #3774, opened 2023-04-17
DeltaHistogram in SignalFx registry doesn't align with count and total


lenin-jaganathan:
**Describe the bug**
https://github.com/micrometer-metrics/micrometer/pull/3750 fixed the issue for distributed publishing of metrics. For StepRegistries, this implementation took an approach of scheduling a thread to poll the meters at the start of the minute and exporting at a different point in time. An unexpected outcome of this fix is that [DeltaHistogram](https://github.com/micrometer-metrics/micrometer/blob/main/implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/DeltaHistogramCounts.java) implementation of SignalFx registry is broken. Since the polling thread will call the count() on Timer at the start of the minute, the count and total are rolled over for the StepWindow. But the actual reading (`takeSnapShot()`) of the timer can happen at any point in time between the start of the step and 80% of the step (maybe even later if batching is involved). This breaks the Delta Histogram Implementation of the SignalFx registry.

 
**Environment**
 All

 - Micrometer version 1.11.0
 - Micrometer registry - SignalFx

**To Reproduce**
How to reproduce the bug:

Set-Up:

* Create a SIgnalFx Meter Registry with a Step of 60 seconds (any step would do for that matter. The 60s is chosen for explaining the behavior).
* Create a Timer and configure SLAs / Histograms.
* Increment the timer once per second.

Scenario:

Assume that the https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java#L129-L138 returns 20th second. So, publishing will happen on the 20th second.
* The poll service rollover count at the 0th second. 
* The scheduled publish happens on the 20th second when the metric is published the count will be a snapshot at the 0th second whereas the Histogram count will be a snapshot at the 20th second. So, for the first publication, the histogram count will be (count+20).

In real-case scenarios, the data recorded will not be uniform, so this Histogram will report a different snapshot that is misaligned with the count and total

**Expected behavior**
Irrespective of the publishing time, the count metric and the count of histogram buckets should match.

lenin-jaganathan:
[...]

PR #3793, opened 2023-04-29
Improve StepBucketHistogram

lenin-jaganathan:
This polishes/fixes [gh-3749](https://github.com/micrometer-metrics/micrometer/pull/3749)
* eliminate null check on noValue
* to support exporting cumulative buckets, more so in a view that this can be leveraged in other libraries as well. (see - https://github.com/micrometer-metrics/micrometer/issues/3774)
* fix explicit histogram buckets (SLO), was zero previously

### DIFF
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpDeltaMeterRegistryTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpDeltaMeterRegistryTest.java
@@ -151,7 +151,7 @@ class OtlpDeltaMeterRegistryTest extends OtlpMeterRegistryTest {
         timer.record(111, TimeUnit.MILLISECONDS);
 
         HistogramDataPoint histogramDataPoint = writeToMetric(timer).getHistogram().getDataPoints(0);
-        assertThat(histogramDataPoint.getExplicitBoundsCount()).isZero();
+        assertThat(histogramDataPoint.getExplicitBoundsCount()).isEqualTo(4);
         this.stepOverNStep(1);
         assertHistogram(writeToMetric(timer), TimeUnit.MINUTES.toNanos(1), TimeUnit.MINUTES.toNanos(2), "milliseconds",
                 3, 198, 111);
@@ -243,7 +243,7 @@ class OtlpDeltaMeterRegistryTest extends OtlpMeterRegistryTest {
         assertHistogram(writeToMetric(ds), 0, TimeUnit.MINUTES.toNanos(1), "bytes", 0, 0, 0);
 
         HistogramDataPoint histogramDataPoint = writeToMetric(ds).getHistogram().getDataPoints(0);
-        assertThat(histogramDataPoint.getExplicitBoundsCount()).isZero();
+        assertThat(histogramDataPoint.getExplicitBoundsCount()).isEqualTo(4);
         this.stepOverNStep(1);
         assertHistogram(writeToMetric(ds), TimeUnit.MINUTES.toNanos(1), TimeUnit.MINUTES.toNanos(2), "bytes", 3, 198,
                 111);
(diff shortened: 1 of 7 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepValue.java
```diff
@@ -34,13 +35,18 @@
...
+    private final AtomicLong lastInitPos;
 
-    private volatile V previous = noValue();
+    private volatile V previous;
 
     public StepValue(final Clock clock, final long stepMillis) {
+        this(clock, stepMillis, null);
+    }
+
+    protected StepValue(final Clock clock, final long stepMillis, @Nullable final V initValue) {
```
shakuzen, line 46: Does this need to be protected? It looks like it could be package-private instead.

shakuzen:
It looks like OTelCollectorIntegrationTest is failing. Perhaps something has changed in a recently published version of the OTel Collector docker image.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepValue.java
lenin-jaganathan: StepBucketHistogram is part of the distribution package so this needs to be protected to use this constructor.

lenin-jaganathan:
@shakuzen The failures in the OTEL collector test are mostly related to your comment [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18903#issuecomment-1446216103) (at least I think so). 

Basically, the test expects the timer to have a name like "test_timer_count", but it seems the name is changed to test_timer_<timeunit>_count. The same goes for the counter as well, where test_counter is now test_counter_total. Haven't digged in deep but I don't see any related PR's there.

lenin-jaganathan:
@shakuzen / @jonatan-ivanov  Does it makes sense for the OtlpStepTimer to be re-named as OtlpDeltaTimer? This is something that I am not able to choose either but I have a slight preference for renaming this as OtlpStepMeters as OtlpDeltaMeter

jonatan-ivanov (approved):

jonatan-ivanov:
@lenin-jaganathan Thank you very much for the PR!
I fixed the build on all of our branches, also rebased this PR to `main`, see details here: https://github.com/micrometer-metrics/micrometer/issues/3796

The summary is: our tests depend on the (wrong) behavior of the OTel collector, some of them were fixed which was a breaking change, so our tests were also broken.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpDeltaMeterRegistryTest.java
```diff
@@ -151,7 +151,7 @@ void timerWithHistogram() {
         timer.record(111, TimeUnit.MILLISECONDS);
 
         HistogramDataPoint histogramDataPoint = writeToMetric(timer).getHistogram().getDataPoints(0);
-        assertThat(histogramDataPoint.getExplicitBoundsCount()).isZero();
+        assertThat(histogramDataPoint.getExplicitBoundsCount()).isEqualTo(4);
```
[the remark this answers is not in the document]
lenin-jaganathan: yes that's right. It make sures as soon as timer object is created, we'll have buckets on takeSnapshot() call. Though this doesn't solve anything for OTLP, it address issues when StepBucketHistogram is used in registries where buckets are registered as Gauges.

jonatan-ivanov:
>Does it makes sense for the OtlpStepTimer to be re-named as OtlpDeltaTimer? This is something that I am not able to choose either but I have a slight preference for renaming this as OtlpStepMeters as OtlpDeltaMeter

Sorry, I missed this comment while I was reviewing/played with this.
My two cents:
- Since it is package-private, it does not have a big significance, we can change anytime.
- To me the `Step*` makes a little bit more sense from the perspective of consistency. I'm thinking that instead of renaming, maybe it would make sense calling out the reason why it does not extend `StepTimer` in the Javadoc .

Issue #3773, opened 2023-04-17
OTLP registry publishes incorrect delta values on close


shakuzen:
See https://github.com/micrometer-metrics/micrometer/pull/3625#pullrequestreview-1376259552. The fix made in https://github.com/micrometer-metrics/micrometer/issues/1882 does not currently help `OtlpMeterRegistry` because it does not extend `StepMeterRegistry`.

PR #3794, opened 2023-04-30
Avoid global publishing alignment in OTLP registry

lenin-jaganathan:
This applies [gh-3750](https://github.com/micrometer-metrics/micrometer/pull/3750) for the OtlpDelta flavor. This is more of copying some of the code from StepMeterRegistry for now as this will be critical for the right functionality.

References:
- https://github.com/micrometer-metrics/micrometer/issues/2818
- https://github.com/micrometer-metrics/micrometer/pull/3750

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
@@ -30,6 +30,7 @@ import io.micrometer.core.instrument.push.PushMeterRegistry;
 import io.micrometer.core.instrument.step.StepCounter;
 import io.micrometer.core.instrument.step.StepFunctionCounter;
 import io.micrometer.core.instrument.step.StepFunctionTimer;
+import io.micrometer.core.instrument.step.StepMeterRegistry;
 import io.micrometer.core.instrument.util.MeterPartition;
 import io.micrometer.core.instrument.util.NamedThreadFactory;
 import io.micrometer.core.instrument.util.TimeUtils;
@@ -46,6 +47,8 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.function.DoubleSupplier;
@@ -76,9 +79,12 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
 
     private final Resource resource;
 
+    private final io.opentelemetry.proto.metrics.v1.AggregationTemporality otlpAggregationTemporality;
+
     private long deltaAggregationTimeUnixNano = 0L;
 
-    private final io.opentelemetry.proto.metrics.v1.AggregationTemporality otlpAggregationTemporality;
+    @Nullable
+    private ScheduledExecutorService meterPollingService;
 
     public OtlpMeterRegistry() {
         this(OtlpConfig.DEFAULT, Clock.SYSTEM);
@@ -100,6 +106,31 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
         this.setDeltaAggregationTimeUnixNano();
         config().namingConvention(NamingConvention.dot);
         start(DEFAULT_THREAD_FACTORY);
+    }
+
+    @Override
+    public void start(ThreadFactory threadFactory) {
+        super.start(threadFactory);
+
+        if (config.enabled() && isDelta()) {
+            this.meterPollingService = Executors.newSingleThreadScheduledExecutor(threadFactory);
+            this.meterPollingService.scheduleAtFixedRate(this::pollMetersToRollover, getInitialDelay(),
+                    config.step().toMillis(), TimeUnit.MILLISECONDS);
+        }
+    }
+
+    @Override
+    public void stop() {
+        super.stop();
+        if (this.meterPollingService != null) {
+            this.meterPollingService.shutdown();
+        }
+    }
+
+    @Override
+    public void close() {
+        stop();
+        super.close();
     }
 
     @Override
@@ -249,6 +280,27 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
                 .setAggregationTemporality(otlpAggregationTemporality)
                 .build())
             .build();
+    }
+
+    /**
+     * This will poll the values from meters, which will cause a roll over for Step-meters
+     * if past the step boundary. This gives some control over when roll over happens
+     * separate from when publishing happens. This method is almost the same as the one in
+     * {@link StepMeterRegistry} it is subtly different from it in that this uses
+     * {@code takeSnapshot()} to roll over the timers/summaries as OtlpDeltaTimer is using
+     * a {@code StepValue} for maintaining distributions.
+     */
+    // VisibleForTesting
+    void pollMetersToRollover() {
+        this.getMeters()
+            .forEach(m -> m.match(gauge -> null, Counter::count, Timer::takeSnapshot, DistributionSummary::takeSnapshot,
+                    meter -> null, meter -> null, FunctionCounter::count, FunctionTimer::count, meter -> null));
+    }
+
+    private long getInitialDelay() {
+        long stepMillis = config.step().toMillis();
+        // schedule one millisecond into the next step
+        return stepMillis - (clock.wallTime() % stepMillis) + 1;
     }
 
     // VisibleForTesting
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpDeltaMeterRegistryTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpDeltaMeterRegistryTest.java
@@ -29,6 +29,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 
 import static io.micrometer.registry.otlp.AggregationTemporality.DELTA;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
 
 class OtlpDeltaMeterRegistryTest extends OtlpMeterRegistryTest {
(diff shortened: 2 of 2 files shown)


implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpDeltaMeterRegistryTest.java
```diff
@@ -311,4 +312,69 @@ void testMetricsStartAndEndTime() {
...
+            .builder("timer.function", functionCount, AtomicLong::get, AtomicLong::get, MILLISECONDS)
+            .register(registry);
+
+        // before rollover
+        assertSum(writeToMetric(counter), 0, TimeUnit.MINUTES.toNanos(1), 0);
+        assertThat(functionCounter.count()).isZero();
+        assertThat(functionTimer.count()).isZero();
+        assertThat(functionTimer.totalTime(MILLISECONDS)).isZero();
+
+        this.stepOverNStep(1);
```
jonatan-ivanov, line 335: The implementation looks ok but I don't think we are testing here what we should.
Since the test is initialized with adding a step to the clock, this is already the second step. This is even weirder if you add `counter.increment();` right after the `Counter` creation (which I think we should), in that case `counter.count()` should be 0 but will be 1 since we are already passed the step boundary.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpDeltaMeterRegistryTest.java
```diff
@@ -311,4 +312,69 @@ void testMetricsStartAndEndTime() {
...
+        counter.increment();
+        functionCount.incrementAndGet();
+        assertSum(writeToMetric(counter), TimeUnit.MINUTES.toNanos(1), TimeUnit.MINUTES.toNanos(2), 0);
+        assertThat(writeToMetric(functionCounter).getSum().getDataPoints(0).getAsDouble()).isEqualTo(15);
+        assertThat(writeToMetric(functionTimer).getHistogram().getDataPoints(0).getSum()).isEqualTo(15);
+        assertThat(writeToMetric(functionTimer).getHistogram().getDataPoints(0).getCount()).isEqualTo(15);
+    }
+
+    @Test
+    void scheduledRolloverTimer() {
```
jonatan-ivanov, line 349: Could you please add a test for `DistributionSummary` too?

jonatan-ivanov (changes requested):

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -251,6 +282,24 @@ private Metric writeSum(Meter meter, DoubleSupplier count) {
             .build();
     }
 
+    /**
+     * This subtly different from {@link StepMeterRegistry#pollMetersToRollover()} in that
+     * this uses takeSnapshot() to rollover the timers/summaries as OtlpDeltaTimer is
+     * using a StepValue for maintaining distributions.
+     */
```
jonatan-ivanov, line 289:
```suggestion
    /**
     * This will poll the values from meters, which will cause a roll over for Step-meters
     * if past the step boundary. This gives some control over when roll over happens
     * separate from when publishing happens. This method is almost the same as the one in
     * {@link StepMeterRegistry} it is subtly different from it in that this uses
     * {@code takeSnapshot()} to roll over the timers/summaries as OtlpDeltaTimer is using
     * a {@code StepValue} for maintaining distributions.
     */
```

jonatan-ivanov (changes requested):

jonatan-ivanov:
@lenin-jaganathan I rebased this to `main` to fixed the build.

shakuzen:
In general, I still think it'd be best to extend StepMeterRegistry instead of copying all of this over to get the behavior that is already provided by StepMeterRegistry.

lenin-jaganathan:
> In general, I still think it'd be best to extend StepMeterRegistry instead of copying all of this over to get the behavior that is already provided by StepMeterRegistry.

@shakuzen I am really not comfortable with doing this change by extending StepMeterRegistry. This will result in an incorrect relationship, which will say OTLP is a StepRegistry which it is not. Below is the Javadoc for StepMeterRegistry,
> Registry that step-normalizes counts and sums to a rate/second over the publishing interval.

However, code duplication also smells equally bad. And we also have https://github.com/micrometer-metrics/micrometer/issues/3773 which will most probably have the same code duplication issues. However, I am up for making the PollService and RollOverService as has-a relationship to avoid this duplication.

cc: @jonatan-ivanov

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -102,6 +108,31 @@ private OtlpMeterRegistry(OtlpConfig config, Clock clock, HttpSender httpSender)
         start(DEFAULT_THREAD_FACTORY);
     }
 
+    @Override
+    public void start(ThreadFactory threadFactory) {
+        super.start(threadFactory);
+
+        if (config.enabled() && isDelta()) {
+            this.meterPollingService = Executors.newSingleThreadScheduledExecutor(threadFactory);
+            this.meterPollingService.scheduleAtFixedRate(this::pollMetersToRollover, getInitialDelay(),
```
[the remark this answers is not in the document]
lenin-jaganathan: @sonatype-lift ignore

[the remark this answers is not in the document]
lenin-jaganathan: @sonatype-lift ignore

jonatan-ivanov:
I'm not really keen on extending `StepMeterRegistry` either and in some cases it does not help a lot I think, in some cases it does but there are trade-offs.
I created a few versions (there are more since they can be combined):

1. I merged together the two PRs (global alignment + final close): https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:otlp-push-registry-all-in-one
- This keeps the registry on being a `PushMeterRegistry` (moving to `StepMeterRegistry` and then away from it is breaking)
- But it has duplicated code with `StepMeterRegistry`
- And it makes the `_closingRollover` method of `OtlpStepTuple2` and `StepValue` `protected` (I don't think this is a big issue)

2. I also created a version so that `OtlpMeterRegistry extends StepMeterRegistry` and keeps most of the things similar to the first solution: https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:otlp-step-registry-all-in-one

- `OtlpMeterRegistry` will be a `StepMeterRegistry` and `OtlpConfig` will be a `StepRegistryConfig`, changing this will be breaking
- We should override `start` since it is different: we should start `meterPollingService` in case of delta but not in case of cumulative; we can also do this (`meterPollingService` needs to be changed to `protected` in `StepMeterRegistry` and this does not feel good):
```java
if (meterPollingService != null && isCumulative()) {
	this.meterPollingService.shutdownNow();
	this.meterPollingService = null;
}
```
- `pollMetersToRollover` is different: `takeSnapshot` (otlp) is called instead of `count` (step), also we should have it in the otlp registry at least as a `package-private` method to make it visible for tests; in order to do this, it needs to be changed to `protected` in `StepMeterRegistry`
- `getInitialDelay` is the same, can be removed
- `stop` can be removed
- `newMeter` and `newGauge` can be removed
- `close` is slightly different: we should only call `closingRollover` in case of delta (should not cause issues if called in case of cumulative); also `OtlpStepTimer` and `OtlpStepDistributionSummary` need to be handled differently, since they are not `StepMeter`s.
- `OtlpStepTuple2` cannot be removed, neither `_closingRollover` methods (`OtlpStepTimer` and `OtlpStepDistributionSummary`)
- `histogram` in `AbstractTimer` should stay `protected` (I needed to change it from `private`, fyi: it is `protected` in `AbstractDistributionSummary` too)
- `closingRollover` in `StepTuple2` and `StepValue` should stay `protected` (I needed to change it from `package-private`)

3. Make `StepMeter` public and let `OtlpStepTimer` and `OtlpStepDistributionSummary` implement it (`OtlpMeterRegistry extends StepMeterRegistry`): https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:otlp-step-registry-and-step-meter
- `close` can be removed from the registry
- `StepMeter` is `public` as well as some `_closingRollover` methods

4. Utilize `StepDistributionSummary` and `StepTimer`: https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:otlp-step-registry-and-steptimer-stepsummary
- `close` can be removed from the registry
- `_closingRollover` still need to be protected in `StepValue`
- It has unnecessary fields (max on Otlp and Step Timer/Distribution)

I think another option is to make `OtlpMeterRegistry` a router and have a `OtlpCumulativeMeterRegistry` and `OtlpStepMeterRegistry` but it would not solve most of the problems above, though it would make future changes non-breaking. I also think that redesigning a few things (inject an implementation of Max, better support for injecting Histogram with support of `_closingRollover`) would make things better and simpler.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpDeltaMeterRegistryTest.java
lenin-jaganathan: Thanks for catching that. I missed updating the meter in the same step. Fixed this.

lenin-jaganathan:
@jonatan-ivanov did try something like,

PollService{
    pollsMetersOnStepBoundary();
    closingRollOver();
}

And then, StepMeterRegistry will have a has-a relationship with PollService and Otlp can also have that in the case of DeltaFlavour.

I can try this, but I will be out on vacation soon(by the end of this week maybe), so not sure if I can try this before the next release. I also want to fix some of the broken behaviors in SignalFxRegistry too so I bet I will not have enough time to try these things. But just adding them as an idea.

jonatan-ivanov:
@lenin-jaganathan I haven't tried it. I think the `PollService` approach is a good idea to make injecting that logic in a more flexible way, though part of the logic is still somewhat duplicated and it also makes `closingRollOver` public.

Not in this PR but I think we have a bigger issue: we should be able to inject Max and Histogram. Once we can do that `OtlpStepTimer` and `OtlpStepDistributionSummary` can extend `StepTimer` and `StepDistributionSummary` so that there will be no need for the custom logic (and we can move that out to a separate component that you are suggesting).

My plan is to make changes in this release that have minimal user impact (basically how these PRs are right now) so that we can continue working on this and fix it properly without breaking changes.

Have fun on your vacation! :)

jonatan-ivanov (approved):

PR #3798, opened 2023-05-02
Publish values on close in OTLP registry

jonatan-ivanov:
The fix for partial publish on close was not applied to OTLP registry.

Closes gh-3773
See gh-1882

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpStepTuple2.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpStepTuple2.java
@@ -13,40 +13,24 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package io.micrometer.registry.otlp;
 
 import io.micrometer.core.instrument.Clock;
-import io.micrometer.core.instrument.step.StepValue;
+import io.micrometer.core.instrument.step.StepTuple2;
 
-import java.util.concurrent.atomic.DoubleAccumulator;
 import java.util.function.Supplier;
 
 /**
- * A {@link StepValue} implementation that tracks max over a Step Interval.
- *
- * @author Lenin Jaganathan
- * @since 1.11.0
+ * This is an internal class not meant for general use. The only reason to have this class
+ * so that {@code OtlpMeterRegistry} can call {@code _closingRollover} on
+ * {@code StepTuple2} and the method does not need to be public.
  */
-class StepMax extends StepValue<Double> {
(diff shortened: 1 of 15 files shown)


jonatan-ivanov:
@lenin-jaganathan If you are up to and have some time, would you mind taking a look at this?

lenin-jaganathan:
@jonatan-ivanov  Linking to my comment on the other PR which tries to fix rollover at the top of the minute - https://github.com/micrometer-metrics/micrometer/pull/3794#issuecomment-1530909924

It seems this also falls in the same bag of somehow duplicating things from StepRegistry but we can't inherit OTLPRegistry from step registry as it is not just a StepRegistry. Just thinking if we could combine these two PRs and come-up with something better with lesser duplications. cc: @shakuzen

jonatan-ivanov:
@lenin-jaganathan This PR is a little more problematic than your but since we started the discussion there, see my answer there: https://github.com/micrometer-metrics/micrometer/pull/3794#issuecomment-1532346339

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpStepTuple2.java
```diff
@@ -0,0 +1,41 @@
...
+import io.micrometer.core.instrument.step.StepTuple2;
+
+import java.util.function.Supplier;
+
+/**
+ * This is an internal method not meant for general use. The only reason to have this
+ * class so that {@code OtlpMeterRegistry} can call {@code _closingRollover} on
+ * {@code StepTuple2} and the method does not need to be public.
+ */
+class OtlpStepTuple2<T1, T2> extends StepTuple2<T1, T2> {
```
lenin-jaganathan, line 29: Now that StepTuple2 has been copied over to OTLP. I guess the better this would be for this to support StepN, which can hold count, total, max, (maybe step-histogram too) so that it makes sure that complete rotation of StepMeter happens with minimal overhead. 
Do you think will it be hard to have StepHolder which can hold 'N' items aligned to same step and should be rotated together?

jonatan-ivanov: The concern I have is instead of `poll1` and `poll2` there should be a `poll(index)` as well as you should provide a collection of "noValue" and a collection of `Supplier`.
I think this can be chaotic rather quickly.

I would like to look into improving step-histogram and max (we don't have an interface for that), so that we can make things simpler and more robust. Once we are there, this idea can be handy but right now I think I would prefer separate types.

PR #3799, opened 2023-05-02
Align DeltaHistogram in SignalFx registry with count and total

lenin-jaganathan:
fixes https://github.com/micrometer-metrics/micrometer/issues/3774, and partially [this](https://github.com/micrometer-metrics/micrometer/issues/3845) one too 

Known Issues,

- [x] ~~Need help from someone from SignalFx to confirm the expected behaviors when both percentiles and sla are enabled on a Timer. This will help in writing failing tests based on the expected behavior and then fixing the.~~ - Kept the old behavior as is and fixed the issue by using StepBucketHistogram for Delta Usecases.
- [x]  pollMetersToRollover() in StepMeterregistry should be overridden so that when the SignalFx registry runs in deltaHistogram mode, it rolls over the step and reports the right data.


lenin-jaganathan:
@atoulme / @bogdandrutu Need your input on the first item in the description. I believe this is critical to be part of 1.11.0 to align with the latest changes made to the micrometer-core.

breedx-splk:
Hey @lenin-jaganathan. We set up a test to examine the current behavior of the Timer with both percentiles and slo enabled. You can see the code and the results here: https://github.com/breedx-splk/micrometer_perc_sla_timer

The current behavior results in 4 gauges and 2 sums.

Most interestingly are the `my.timer.histogram` and `my.timer.percentile` gauges. The `my.timer.histogram` has a timeseries that corresponds with each of the SLO values configured into the timer, and the `my.timer.percentile` has a timeseries per configured percentile.

Given that some users might rely on this existing behavior, it would be ideal to maintain the shape of those metrics. 👍🏻

lenin-jaganathan:
@breedx-splk Can you have a look at this PR? Kept the existing behavior in SignalFxMeterRegistry for histograms and only the DeltaHistogram is altered so that it address the issue (i,e align with Sum and Count).

jonatan-ivanov:
I need to take a deeper look but since it seems a bug it makes sense to fix it in a patch release. Also, if we decide to backport it to earlier releases (1.9.x), maybe this can be fixed earlier.

breedx-splk:
> @breedx-splk Can you have a look at this PR? Kept the existing behavior in SignalFxMeterRegistry for histograms and only the DeltaHistogram is altered so that it address the issue (i,e align with Sum and Count).

Hey @lenin-jaganathan . Respectfully, without knowing much about micrometer implementation internals and not knowing much about the `SignalFxMeterRegistry` implementation, I'm having a hard time following what you're trying to fix here. I've read the description in #3774 maybe 10 times and have tried to recreate it [here](https://github.com/breedx-splk/micrometer_perc_sla_timer/blob/main/src/main/java/com/splunk/example/BasicMicrometerTimerApp.java#L42).

> Expected behavior: Irrespective of the publishing time, the count metric and the count of histogram buckets should match.

I don't think I follow this. With delta histograms turned on in the registry, I see a `Sum` _named_ `my.timer.histogram` for each of the configure SLO values. In my sample code, there are two -- one for the 1.75 slo and one for the 0.999 slo, and a third one with the same name for +Inf. So, on the first report I see something like this:

* Sum `my.timer.histogram` (le=0.999) value = 82.000
* Sum `my.timer.histogram` (le=1.75) value = 82.000
* Sum `my.timer.histogram` (le=+Inf) value = 82.000
* Sum `my.timer.count` (le=0.999) value = 51

I take this to mean that the "histogram" for the slo has 3 buckets -- one for each value and one for +inf. Am I understanding that incorrectly? Is the problem then that the 51 and the 82 don't match? 

Regardless, the count does seem to be notably lower than the actual number of ticks issued by the app.

lenin-jaganathan:
@jonatan-ivanov That would be good. Just a note,
* The broken feature is not part of 1.9.x so this fix is not needed there.
* https://github.com/micrometer-metrics/micrometer/pull/3750 was introduced in 1.11.x, so the feature is working in 1.10.x just fine.

So, the only major version where this is broken is 1.11.0, which is why I based this on 1.11.x. It would be nice to have some feedback on this from you guys and get this in time for the next patch.


> Respectfully, without knowing much about micrometer implementation internals and not knowing much about the SignalFxMeterRegistry implementation, I'm having a hard time following what you're trying to fix here

@breedx-splk I get you and I don't want to be hard on you here. But yeah, the exact thing you mentioned in your comment where the histogram reports 82 values and the count reports 51 is the issue this PR is trying to fix. (Do note that this is the behavior when `publishCumulativeHistogram` and `publishDeltaHistogram` are enabled).
Thanks for confirming this from the SignalFx side. As suggested in this PR (and with the test cases added in [this commit](https://github.com/micrometer-metrics/micrometer/pull/3799/commits/b01178904ecfafa874f05d2e9d8d9fa2387f5d8c) ), this PR keeps all the existing behavior and addresses only what is described in the bug. 

@jonatan-ivanov / @shakuzen Need your help in getting this merged.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxTimer.java
```diff
@@ -37,70 +35,46 @@
...
-    @Override
-    public double totalTime(TimeUnit unit) {
-        return TimeUtils.nanosToUnit(countTotal.poll2(), unit);
-    }
-
-    @Override
-    public double max(TimeUnit unit) {
-        return max.poll(unit);
+        if (stepBucketHistogram != null) {
+            final long nanoAmount = (long) TimeUtils.convert(amount, unit, TimeUnit.NANOSECONDS);
```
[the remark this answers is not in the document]
lenin-jaganathan: @sonatype-lift ignore

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
```diff
@@ -16,88 +16,61 @@
...
-final class SignalfxDistributionSummary extends AbstractDistributionSummary {
-
-    private final LongAdder count = new LongAdder();
-
-    private final DoubleAdder total = new DoubleAdder();
-
-    private final StepTuple2<Long, Double> countTotal;
-
-    private final TimeWindowMax max;
+final class SignalfxDistributionSummary extends StepDistributionSummary {
```
lenin-jaganathan, line 34: This makes SignaFxMeters simple and makes them closely aligns with Step contracts. The earlier attempt to use AbstractDistributionSummary was due to the lack of flexibility on Histogram which was fixed in 1.11.0 by adding a new constructor which takes Histogram rather than DistributionStatistic Config.

lenin-jaganathan:
Got it. Thanks @jonatan-ivanov for the explanation 👍🏽.

I guess 1.11.1 is scheduled for first week June right? So ajy help with the review will be appreciated. Thanks a lot.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
lenin-jaganathan: A positive side effect of using the StepMeters in StepRegistry might fix https://github.com/micrometer-metrics/micrometer/issues/3845.

I guess this fixes https://github.com/micrometer-metrics/micrometer/issues/3845.

lenin-jaganathan:
@jonatan-ivanov Any help with the review is appreciated.
Thank you!!

lenin-jaganathan:
@jonatan-ivanov/ @shakuzen  Any help with the review is appreciated. This is kind of a blocker for us to move to 1.11.x which has some fixes and improvements. Basically, if this makes it in time for 1.11.1, it would be really helpful.

Thank you!!

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -260,4 +260,20 @@ protected TimeUnit getBaseTimeUnit() {
...
 
+    @Override
+    protected void pollMetersToRollover() {
+        if (config.publishDeltaHistogram()) {
+            // When delta histogram is enabled, force a takeSnapShot on the timer to roll
+            // Over the count, total and
+            // histograms altogether.
+            this.getMeters()
+                .forEach(m -> m.match(gauge -> null, Counter::count, Timer::takeSnapshot,
+                        DistributionSummary::takeSnapshot, meter -> null, meter -> null, FunctionCounter::count,
```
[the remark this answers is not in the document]
shakuzen: I have a POC locally for what I'm tentatively calling StepMeterRegistryVerifier, where you can register consumers to be run in subsequent steps, and it takes care of calling `pollMetersToRollover` for you at the beginning of the step. For example, I converted the test added in this pull request to use it as follows:

```java
@Test
@Issue("3774")
void deltaHistogramCountsShouldMatchWithTimerCount() {
    MockClock mockClock = new MockClock();
    SignalFxMeterRegistry registry = new SignalFxMeterRegistry(cumulativeDeltaConfig, mockClock);
    Duration[] buckets = new Duration[] { Duration.ofMillis(10), Duration.ofMillis(50), Duration.ofMillis(100) };
    Timer timer = Timer.builder("my.timer").serviceLevelObjectives(buckets).register(registry);

    StepMeterRegistryVerifier.using(registry).addStepConsumer(reg -> {
        timer.record(5, TimeUnit.MILLISECONDS);
        timer.record(20, TimeUnit.MILLISECONDS);
        timer.record(175, TimeUnit.MILLISECONDS);
        timer.record(2, TimeUnit.MILLISECONDS);
    }).addStepConsumer(reg -> {
        // Recording for next step has started. But these values should be reported
        // only in then next step.
        timer.record(5, TimeUnit.MILLISECONDS);
        timer.record(25, TimeUnit.MILLISECONDS);
        timer.record(75, TimeUnit.MILLISECONDS);
        timer.record(125, TimeUnit.MILLISECONDS);
        timer.record(500, TimeUnit.MILLISECONDS);

        // Current is not elapsed, so only the values recorded before start of this
        // step should be reflected yet.
        // Assert that data recorded for previous step is available.
        List<SignalFxProtocolBuffers.DataPoint> dataPoints = getDataPoints(registry, mockClock.wallTime());
        assertThat(dataPoints).hasSize(8).has(gaugePoint("my.timer.avg", 0.0505), atIndex(0))
                .has(counterPoint("my.timer.count", 4), atIndex(1))
                .has(allOf(counterPoint("my.timer.histogram", 4), bucket("+Inf")), atIndex(2))
                .has(allOf(counterPoint("my.timer.histogram", 2), bucket(buckets[0])), atIndex(3))
                .has(allOf(counterPoint("my.timer.histogram", 3), bucket(buckets[1])), atIndex(4))
                .has(allOf(counterPoint("my.timer.histogram", 3), bucket(buckets[2])), atIndex(5))
                .has(gaugePoint("my.timer.max", 0.5), atIndex(6))
                .has(counterPoint("my.timer.totalTime", 0.202), atIndex(7));
    }).addStepConsumer(reg -> {
        List<SignalFxProtocolBuffers.DataPoint> dataPoints = getDataPoints(registry, mockClock.wallTime());
        // Assert that data recorded for previous step is available.
        assertThat(dataPoints).hasSize(8).has(gaugePoint("my.timer.avg", 0.146), atIndex(0))
                .has(counterPoint("my.timer.count", 5), atIndex(1))
                .has(allOf(counterPoint("my.timer.histogram", 5), bucket("+Inf")), atIndex(2))
                .has(allOf(counterPoint("my.timer.histogram", 1), bucket(buckets[0])), atIndex(3))
                .has(allOf(counterPoint("my.timer.histogram", 2), bucket(buckets[1])), atIndex(4))
                .has(allOf(counterPoint("my.timer.histogram", 3), bucket(buckets[2])), atIndex(5))
                .has(gaugePoint("my.timer.max", 0.5), atIndex(6))
                .has(counterPoint("my.timer.totalTime", 0.73), atIndex(7));
    }).verify();

    registry.close();
}
```

shakuzen (commented):
Thanks for the pull request, and for updating it with a separate test. I think overall it's good. My biggest concern is opening up API we'll need to maintain in a patch release. I'd like to explore more if we can reasonably avoid that.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
```diff
@@ -16,88 +16,61 @@
...
-        max = new TimeWindowMax(clock, distributionStatisticConfig);
+        super(id, clock, distributionStatisticConfig, scale, stepMillis, defaultHistogram(clock,
+                CumulativeHistogramConfigUtil.updateConfig(distributionStatisticConfig, isDelta), false));
+
         if (distributionStatisticConfig.isPublishingHistogram() && isDelta) {
-            deltaHistogramCounts = new DeltaHistogramCounts();
+            stepBucketHistogram = new StepBucketHistogram(clock, stepMillis,
+                    DistributionStatisticConfig.builder()
+                        .serviceLevelObjectives(CumulativeHistogramConfigUtil
+                            .addPositiveInfBucket(distributionStatisticConfig.getServiceLevelObjectiveBoundaries()))
```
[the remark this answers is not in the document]
lenin-jaganathan: Ha! Thanks for pointing this out. Fixed it. I was hoping that `publishingHistogram()` will guard and then realized that it is not the same.

implementations/micrometer-registry-signalfx/src/test/java/io/micrometer/signalfx/SignalFxMeterRegistryTest.java
```diff
@@ -255,7 +256,16 @@ void shouldExportDeltaHistogramData_Timer(Duration[] buckets, double[] percentil
         // Advance time, so we are in the "next" step where currently recorded values will
         // be reported.
         mockClock.add(config.step());
+        registry.pollMetersToRollover(); // Poll service rolls the step meters.
 
+        // Recording for next step has started. But these values should be reported only
+        // in step.
+        timer.record(5, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
lenin-jaganathan: Addressed.

implementations/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
```diff
@@ -16,88 +16,69 @@
...
-    @Override
-    public double totalAmount() {
-        return countTotal.poll2();
-    }
-
-    @Override
-    public double max() {
-        return max.poll();
+        if (stepBucketHistogram != null) {
+            stepBucketHistogram.poll();
```
shakuzen, line 69: A comment about why this is here would be good for more easily understanding it in the future while looking at the code. Same in the Timer implementation.

micrometer-test/src/main/java/io/micrometer/core/instrument/step/PollingAwareMockStepClock.java
```diff
@@ -0,0 +1,80 @@
...
+
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.util.TimeUtils;
+
+/**
+ * A clock meant to be used for testing {@link StepMeterRegistry}. This clock does the
+ * {@link StepMeterRegistry#pollMetersToRollover()} whenever the step is crossed thus
+ * simulating the expected behaviour of step meters.
+ */
+public class PollingAwareMockStepClock implements Clock {
```
shakuzen, line 31: I'm not sure how meaningful or noticed it has been in the lifetime of Micrometer, but we do have an `@Incubating` annotation that I think we should put on this class in case it makes sense to make some breaking changes (or even remove) it at some point.

shakuzen (approved):
Looks good to me. I left a couple comments on small things to fix. Thanks for the work on this.

micrometer-test/src/main/java/io/micrometer/core/instrument/step/PollingAwareMockStepClock.java
lenin-jaganathan: Yeah. It completely make sense to add something like that and introduce future changes on this and mature this. I will work on this today evening my time and make sure that I add a commit by EOD today.

On the same note, I think it is now essential that all the StepRegistry tests uses this clock or the StepVerifier you proposed, so that this poll behaviour is not excluded in the tests. Since, the behaviour is a scheduled thread it is easy to jot account for this tests and miss out on something like what this issue tries to solve. I will probably open an separate issue to make a change like that.

lenin-jaganathan:
Addressed the open queries and comments.

Issue #3797, opened 2023-05-01
Apache HTTP async client metrics use HttpContext in weird way, NPE is thrown


flozano:
**Describe the bug**
I am getting a NPE when using properly (I think, as I follow the javadoc in that class) MicrometerHttpClientInterceptor.

See https://github.com/micrometer-metrics/micrometer/issues/1886#issuecomment-1529851272 for all the details; I was initially thinking it was just a weird design decision and therefore just commented there, but now I am convinced it's a bug (after reviewing the different implementations of HttpContext interface in apache httpcomponents client 4.x, I am not sure it can be safely used as a map key).

**Environment**
JDK 17
MacOS
Apache httpcomponents async client 4.x

**To Reproduce**
use the interceptor exactly as described.

**Additional context**
all the details are in https://github.com/micrometer-metrics/micrometer/issues/1886#issuecomment-1529851272

shakuzen:
I don't know a particular reason that implementation was chosen. I also don't know why you would be getting a NPE because of it, though. Is this something you can reproduce in a test case? I think the context would need to be shared by multiple requests for that to happen. And if that is the case, I wonder why we wouldn't have the same issue if we try storing the Micrometer sample in the HttpContext. Though I agree it does seem like a reasonable approach, beside this issue.

flozano:
I was not successful reproducing it in a test, but it shows up consistently in a deployed server. 

First thing I need to look at is that this client sends a BasicHttpContext containing some authentication configuration. Maybe the HttpContext seen by the initial interceptor is the one I send, and the response interceptor for some reason sends an IOSession-backed one.

Another wild speculation is that for some reason (number of I/O threads, time of server to respond), the server takes a while to respond and the HttpContext is not exactly the same instance (it's backed by same IOSession and therefore contains the same attributes, but it's not the same HttpContext instance). This is just guess, as it's the only explanation I can think of.

Or maybe it's a combination of both. The NPE however definitely happens, and the interceptors are properly set to the client, so something is going on.

If any of those (or both) was the cause, by using the interface of the HttpContext it should be fixed, as it is a stronger expectation on the API contract that the HttpContext interface provides the same bag of attributes to both the request and the response interceptor. Current code has an expectation about the HttpContext not only containing the same info, but also about being exactly the same instance when received by both interceptors. I am not sure it can be assumed this is the case.

I will keep trying to reproduce in isolated way it by tweaking httpcomponents async client I/O reactor and server response time.

[...]

PR #3801, opened 2023-05-02
Use (Async) ExecChainHandler to measure IOExceptions (#3800)

### LINKED ISSUE #3800: Instrumentations for Apache HttpComponents do not meter errors and leak memory
**Describe the bug**
An HttpAsyncClient instrumented with MicrometerHttpClientInterceptor does not meter IO errors.
For example, requests resulting in a `ConnectTimeoutException`, `SocketTimeoutException` are not metered.

`MicrometerHttpRequestExecutor` however does meter IO errors by setting the `status` key to IO_ERROR.

**Environment**
 - Micrometer version [**1.11.0-RC1**]
 - Micrometer registry [**prometheus**]
 - Java version: [**17**]

**To Reproduce**
How to reproduce the bug:
Make an instrumented client to run into some sort of IO Exception, ConnectTimeoutException, SocketTimeoutException, etc.

**Expected behavior**
- The request is counted, and elapsed time is measured.
- The Tag **status** has the value IO_ERROR
- The Tag **outcome** is UNKNOWN.

**Additional context**
The Apache Http Components instrumentation does not support the tag **exception**, which is quite commonly used by other http client instrumentations.

cachescrubber:
Use `HttpRequestRetryStrategy` to measure IOExceptions.
Fixes #3800, #3797, #3706


cachescrubber:
The only solution I could find to measure errors in `HttpAsyncClient` is to use a delegating `HttpRequestRetryStrategy` which measures after the very last retry. WDYT? Are there other possibilities? The code would need some polishing, I just want verify the suggested solution with you.

cachescrubber:
FTR: https://issues.apache.org/jira/browse/HTTPCLIENT-2274

cachescrubber:
This is becoming something bigger then expected.

The current implementation for both async (MicrometerHttpClientInterceptor) and classic (MicrometerHttpRequestExecutor) is implemented in a way that

* metrics do not include the time needed to establish a connection.
* metrics do not count failed connection attempts.
* async metrics even do not include IO errors for established connection (original topic here)

Some experimentation using a `HttpRequestRetryStrategy` to meter IO errors felt awkward and I was not able to capture errors during connection establishment.

So I checked with the [HttpClient Team on their advise](https://issues.apache.org/jira/browse/HTTPCLIENT-2274) how to instrument HttpClient (async and classic). They recommend that we should use (Async) ExecChainHandler to capture execution time metrics.

I update this PR for (currently hc5 only) and implemented the suggested instrumentation.

cachescrubber:
HttpClient has a Feature to execute automatic retries in case of special return codes (503) or retrieable IO exceptions. Default setup (`DefaultHttpRequestRetryStrategy`) is to retry 1 time after a 1 sec delay.

It is possible to meter the http transaction combined or individually.

**combined** means

- measure the total time over all attempts incl. the inferred delays.
- record the outcome of the last attempt.

**individual** means

- measure time for each individual request - **excluding** the inferred delays.
- record the outcome for each attempt.

bclozel:
Superseded by #3800 and merged with [cb86bc2](https://github.com/micrometer-metrics/micrometer/commit/cb86bc2e3b727df5c323adf2f5e4bd4b8bf693d6). Thanks so much for your contribution @cachescrubber !

cachescrubber:
Hi @bclozel I just took the time to review your changes. Nice to see the `ObservationRegistry` in action without the intermediate `ObservationOrTimerCompatibleInstrumentation`. Looks much cleaner!

I think you missed on important aspect of my original PR. I introduced support to `meterRetries` (see https://github.com/micrometer-metrics/micrometer/pull/3801#issuecomment-1547252927). Basically, the (Async)HttpClient has a build in retry capability which needs special treatment in the instrumentation, otherwise the observations are not correct.

bclozel:
Hey @cachescrubber - I left this out on purpose, because the behavior is inconsistent between the "classic" and "async" flavours and I'm afraid there's nothing we can do about it.

The `AsyncExecChain.Scope#execCount` (the number of executions, > 1 meaning it's a retry) is only available to the async variant. There is no such thing in `ExecChain.Scope` and the `execCount` information in `HttpRequestRetryExec` is not made available to the rest of the chain. I initially thought about adding a `"retry.count"="1"` as a key value but this won't work for the "classic" version.

Also, setting up the micrometer `ChainHandler` before the retry one does not yield the same behavior in "classic" and "async":

* in "classic", we only see a single execution from the `ObservationExecChainHandler` and record only one observation
* in "async", the `ObservationExecChainHandler` is called for each retry, creating a new observation each time

Because of this, I don't think we can consistently produce information as `Keyvalue` or directly in `Observation.Context`. If you have suggestions let me know.

cachescrubber:
Hi @bclozel - you are right  - instrumentation is a bit different between classic and async. This is why I introduced 

* ApacheHttpClientMetricsBinder#instrument(org.apache.hc.client5.http.impl.async.HttpAsyncClientBuilder)
* ApacheHttpClientMetricsBinder#instrument(org.apache.hc.client5.http.impl.classic.HttpClientBuilder)

this way, the actual placement of the handler inside the handler-chain becomes an implementation detail, an no longer the responsibility of the user. Also, the `(Async)ExecChainHandler` implementations were package-private, not extensible.

I'm pretty sure I covered the retry scenarios with integration tests showing consistent behavior for both async and classic. Please see (in my last version)

* hc5.ApacheHttpClientMetricsBinderTest#retriesAreMetered_overall
* hc5.ApacheHttpClientMetricsBinderTest#retriesAreMetered_individually
* hc5.ApacheHttpClientMetricsBinderTest#testPositiveOutcomeAfterRetry_overall
* hc5.ApacheHttpClientMetricsBinderTest#testPositiveOutcomeAfterRetry_individual

I actually think the tests would show that specially the async meters not correct like they are right now.

Next to the placement within the handler-chain,  `(scope.execCount.get() == 1)` is the crucial part in the `MeteringAsyncExecChainHandler` in order to get correct results.

I think I could submit a new PR starting with a new Integration Test showing/demonstrating the issue. Did you removed the Wiremock based tests on purpose? Because I'm not sure how to mock the required behaviour.

bclozel:
> instrumentation is a bit different between classic and async. This is why I introduced
>
>   ApacheHttpClientMetricsBinder#instrument(org.apache.hc.client5.http.impl.async.HttpAsyncClientBuilder)
>   ApacheHttpClientMetricsBinder#instrument(org.apache.hc.client5.http.impl.classic.HttpClientBuilder)

I'd rather not tie our instrumentation to implementations details of another ExecChain implementation. Given the inconsistencies and the lack of support for cancellation overall, I don't think this would be wise.

> this way, the actual placement of the handler inside the handler-chain becomes an implementation detail, an no longer the responsibility of the user. Also, the (Async)ExecChainHandler implementations were package-private, not extensible.

I don't think your initial PR was supporting cancellations and the implementation was still leaking observations. With this in place, I think that the micrometer ExecChain must be configured before the retry one because of internal lifecycle issues with the connection resources. Maybe that's something we should document on the `ObservationExecChainHandler` (although we're already documenting that it should be configured first).

> I actually think the tests would show that specially the async meters are not correct like they are right now.

I think they would be consistent with the behavior of the library: in the async case retries trigger the entire chain, whereas in the classic case this happens internally.

> Next to the placement within the handler-chain, (scope.execCount.get() == 1) is the crucial part in the MeteringAsyncExecChainHandler in order to get correct results.

I see where you're coming from, but this merely hides the fact that several separate requests were made. If I understand correctly, if retries are not metered, there is a single observation that does not cover all retries in the async mode. Depending on how the setup is made, this could be the opposite for the classic case.

With all that in mind, I don't think we should try to compensate for the inconsistencies in the library itself.

cachescrubber:
> I don't think your initial PR was supporting cancellations and the implementation was still leaking observations. With this in place, I think that the micrometer ExecChain must be configured before the retry one because of internal lifecycle issues with the connection resources. 

Yes, my initial PR did not support cancellations. I created another PR reinstating my meterRetries support over your much improved HandlerChain implementation. It would be great if you could have a look at #3941. Is it still leaking observations or does your async cancellation support also kicks in here?

PR #3810, opened 2023-05-05
Add DeferredScope

ttddyy:
`DeferredScope` allows creating a scope instance without opening it.

For example, when creating, opening, and closing a scope need to happen independently, this `DeferredScope` can be used as:

```java
// create a scope object without starting it
DeferredScope scope = observation.deferredScope();

// two different callbacks for start and stop
StartCallback startCallback = new StartCallback(scope);
StopCallback stopCallback = new StopCallback(scope);

// callback will be invoked later by underlying framework
return new Listener(startCallback, stopCallback);
```


ttddyy:
The original usecase no longer needs this change because of the different approach in https://github.com/micrometer-metrics/micrometer/pull/3836.
closing.

Issue #3678, opened 2023-03-08
Provide a simple way to make decisions based on the parent in ObservationPredicate


jonatan-ivanov:
This issue was originally discussed in https://github.com/spring-projects/spring-boot/issues/34400
The scenario is the following: if Spring Security is used and users want to ignore `/actuator` endpoints, there is no easy way to do it today because Spring Security does not add details about the request. One potential solution would be checking the parent (root) Observation but it is not available through the context when the predicate is tested.

There is a [reproducer](https://github.com/spring-projects/spring-boot/issues/34400#issuecomment-1459658488) in the issue mentioned above.

Also see: https://github.com/spring-projects/spring-security/issues/12854

braunsonm:
Thanks for opening this @jonatan-ivanov 
Can you clarify if the intention of this issue is to also have the parent set if the parent was ignored in a previous predicate?

ie:

```java
@Configuration
public class TestConfiguration {
    @Bean
    ObservationPredicate ignoreActuator() {
        return (s, context) -> ignoreActuatorRecursive(context);
    };

    public boolean ignoreActuatorRecursive(Observation.ContextView context) {
        if (context instanceof ServerRequestObservationContext serverRequestObservationContext) {
            HttpServletRequest carrier = serverRequestObservationContext.getCarrier();
            return !carrier.getServletPath().startsWith("/actuator"); // <----- Line (*)
        }

        if (context.getParentObservation() != null) {  // <-- Will the parent still be set if it was ignored in line (*)?
            return ignoreActuatorRecursive(context.getParentObservation().getContextView());
        }

        return true;
    }
}
```

It would be great for the parent to remain set even if the parent observation was ignored in a previous predicate. The reason being so that you can properly ignore any subsequent observations deriving from a root span that was ignored.

jonatan-ivanov:
No, and there the parent can't be really set i these cases. If an Observation is ignored we will create a noop one that is hopefully eliminated by JIT.

quaff:
@braunsonm `context.getParentObservation()` will always be null in `ObservationPredicate`, I'm trying to fix it by  https://github.com/micrometer-metrics/micrometer/pull/3867.

PR #3843, opened 2023-05-17
Introduce withParentOnly for @Observed

quaff:
Take tracing for example, we may no want to create root span by @Observed, they should be created if a root http span present.


marcingrzejszczak:
Can you please elaborate on this? What problem are you trying to solve? What's currently not working?

quaff:
> Can you please elaborate on this? What problem are you trying to solve? What's currently not working?

I don't want span created by @Observed as root of trace, It must be child of other span. 
For example, it shouldn't be started if called by `@Scheduled` trigger, but I want it be included in a complete http trace.

jonatan-ivanov:
I'm not sure I clearly get what you are trying to do so I'm going to make assumptions. Based on this:
>For example, it shouldn't be started if called by @Scheduled trigger, but I want it be included in a complete http trace.

I think you have a method that you annotate with `@Observed` and that method is called from different places, e.g.: from `@Scheduled` and from a `@Controller`. So what you need is basically a conditional `@Observed`.

I think there are two options to do this:

1. Create an ObservationPredicate that checks if there is in a context of http and ignores it if it isn't. Here's an [example](https://github.com/jonatan-ivanov/teahouse/blob/a4e27451998d05480f236d93d49842cac750ecc1/core/src/main/java/org/example/teahouse/core/actuator/config/CommonActuatorConfig.java#L59), it's a little bit different but it can detect if your observation is in the context of http.
2. Do not put `@Observed` on your `@Scheduled` method and use `Observation.createNotStarted(...).observe(methodReference);` in your `@Controller` instead.

I think this issue is related: https://github.com/micrometer-metrics/micrometer/issues/3678 and I would rather solve this for any `Observation` than only for `@Observed`.

quaff:
> I think you have a method that you annotate with `@Observed` and that method is called from different places, e.g.: from `@Scheduled` and from a `@Controller`. So what you need is basically a conditional `@Observed`.

You are right.
> 
> I think there are two options to do this:
> 
> 1. Create an ObservationPredicate that checks if there is in a context of http and ignores it if it isn't. Here's an [example](https://github.com/jonatan-ivanov/teahouse/blob/a4e27451998d05480f236d93d49842cac750ecc1/core/src/main/java/org/example/teahouse/core/actuator/config/CommonActuatorConfig.java#L59), it's a little bit different but it can detect if your observation is in the context of http.
> 2. Do not put `@Observed` on your `@Scheduled` method and use `Observation.createNotStarted(...).observe(methodReference);` in your `@Controller` instead.

I vote for option 1, option 2 is viable but invasive.

I think compare to this PR there is a little drawback of `ObservationPredicate`, objects are created and then dropped, it will increase overhead, hope it's insignificant.

jonatan-ivanov:
>I think compare to this PR there is a little drawback of ObservationPredicate, objects are created and then dropped, it will increase overhead, hope it's insignificant.

You should have only one `ObservationPredicate` instance that does this so this means one extra object in your whole application lifecycle. Also, using an `ObservationPredicate` should be more lightweight than this PR since in this PR a new `Observation.Context` and a new `Observation` will be created for every call (and dropped). With a predicate only the `Observation.Context` will be created (and dropped) there will not be a new `Observation` instance (we cannot spare creating the context since that's what your decision is based of).

quaff:
Please close this if https://github.com/micrometer-metrics/micrometer/pull/3867 get merged, then we could use `ObservationPredicate` like this:
```java
	@Bean
	ObservationPredicate noParentlessDatabaseObservations() {
		return (name, context) -> {
			if (context instanceof LettuceObservationContext || context instanceof DataSourceBaseContext) {
				return context.getParentObservation() != null;
			}
			return true;
		};
	}
```

marcingrzejszczak:
Closing since we did this https://github.com/micrometer-metrics/micrometer/pull/3867

PR #3854, opened 2023-05-22
Make DistributionStatisticConfig extensible

mauhiz:


jonatan-ivanov:
Let's discuss this on the issue first before we would talk about what is needed to implement it.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
```diff
@@ -71,6 +72,25 @@ public class DistributionStatisticConfig implements Mergeable<DistributionStatis
     @Nullable
     private Integer bufferLength;
 
+    public DistributionStatisticConfig() {
+    }
+
+    public DistributionStatisticConfig(DistributionStatisticConfig original) {
```
jonatan-ivanov, line 78: I'm not sure I getting why this is needed, should not this be the concern of `merge` instead?

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
```diff
@@ -271,7 +291,15 @@ public double[] getServiceLevelObjectiveBoundaries() {
...
     public static class Builder {
 
-        private final DistributionStatisticConfig config = new DistributionStatisticConfig();
+        protected final DistributionStatisticConfig config;
+
+        public Builder() {
+            config = new DistributionStatisticConfig();
+        }
+
+        protected Builder(DistributionStatisticConfig config) {
```
jonatan-ivanov, line 300: I'm not sure I get this either a `Builder` should create an instance of `DistributionStatisticConfig`, not receive one.

micrometer-core/src/test/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfigExtensibilityTest.java
```diff
@@ -0,0 +1,155 @@
...
+                sub.extraParam = extraParam;
+            }
+            return sub;
+        }
+
+        public static SubBuilder subBuilder() {
+            return new SubBuilder(new SubDistributionStatisticConfig());
+        }
+
+        public static class SubBuilder extends Builder {
```
jonatan-ivanov, line 65: Is this why you need to receive a `DistributionStatisticConfig` in `DistributionStatisticConfig.Builder`?

I don't think `SubDistributionStatisticConfig.Builder` should extend `DistributionStatisticConfig.Builder`.

mauhiz: Yes. Which way would you recommend?

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
mauhiz: I thought it's the only way that that the subclass's constructor can copy all fields to a subclass instance, when merging

Issue #3863, opened 2023-05-30
StepRegistry publish on close before scheduled publish loses previous step data


lenin-jaganathan:
**Describe the bug**
https://github.com/micrometer-metrics/micrometer/issues/1882 was fixed by introducing the StepMeter Contract and doing a closing rollover. This fixed the original issue but https://github.com/micrometer-metrics/micrometer/pull/3750 introduced a polling thread that rolls the step meters at the start of the step. 

These 2 fixes combined produce a grey area where publishing is happening at the middle of the step and the registry closes. In this scenario, the data accumulated for the previous step will be lost and will be never reported.

**Environment**
 - Micrometer version - 1.11.0
 - Micrometer registry - StepMeterRegistry

** To Reproduce **
How to reproduce the bug:
Use any Step registry for testing,
* start the registry and create an increment counter every second.
* close the registry exactly during publishing.

If the step is 60 seconds and the counter is incremented every second, it will report a value of 60 for every step. At the end of 1st minute, rollOver will happen and the data to be reported is `60` (but this is not published yet). Let's assume that publishing is scheduled on the 30th second of every minute. If the app initiates a shutdown on the 20th second, StepRegistry will do a closingRollover on StepMeters causing the accumulated values (`20` is accumulated for the current step) will be published. But, the data that is accumulated for the last completed step(`60`) was never published and gets lost.

```
S1      S2         S3         S4
---|r----p-----|r----p------|r----p------|r--c

r - rollover meters at the end of the step
p - actual publish is started
c - closing Rollover is performed.
```

As you can see from the above the data for the last "r"(last completed step) was never published. Since the app shutdown is initiated closingRollover will update the value to be published with the current value, thus data for the previous minute is dropped.

**Expected behavior**
Publish on close should not cause data loss for the previous step.

lenin-jaganathan:
@shakuzen Can this be updated to point to 1.11.x milestone?

PR #3864, opened 2023-05-30
Publish completed step values on close (StepRegistry)

lenin-jaganathan:
fixes https://github.com/micrometer-metrics/micrometer/issues/3863

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
@@ -41,6 +41,8 @@ public abstract class PushMeterRegistry extends MeterRegistry {
 
     private final AtomicBoolean publishing = new AtomicBoolean(false);
 
+    private long lastScheduledPublishStartTime = 0L;
+
     @Nullable
     private ScheduledExecutorService scheduledExecutorService;
 
@@ -60,6 +62,7 @@ public abstract class PushMeterRegistry extends MeterRegistry {
     // VisibleForTesting
     void publishSafely() {
         if (this.publishing.compareAndSet(false, true)) {
+            this.lastScheduledPublishStartTime = clock.wallTime();
             try {
                 publish();
             }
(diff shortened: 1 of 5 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/step/StepMeterRegistryTest.java
```diff
@@ -342,6 +321,57 @@ void publishOnCloseCrossesStepBoundary() {
         assertThat(registry.publishedFunctionTimerTotals.pop()).isEqualTo(53);
     }
 
+    @Test
+    @Issue("#3863")
+    void shouldPublishLastCompletedStepOnClosingRollover() {
```
shakuzen, line 326: Nit pick on the test name: how about `shouldPublishLastCompletedStepWhenClosingBeforeScheduledPublish`?

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -60,6 +62,7 @@ protected PushMeterRegistry(PushRegistryConfig config, Clock clock) {
     // VisibleForTesting
     void publishSafely() {
         if (this.publishing.compareAndSet(false, true)) {
+            this.lastScheduledPublishStartTime = clock.wallTime();
```
shakuzen, line 65: I don't think it's a problem with the current implementation, but it feels like it would be more correct to set this directly from the scheduled task rather than in this method. After all, this method *could* be called from outside scheduling. I'm thinking updating the scheduling to something like:

```java
scheduledExecutorService.scheduleAtFixedRate(() -> {
    this.lastScheduledPublishStartTime = clock.wallTime();
    publishSafely();
}, initialDelayMillis, stepMillis, TimeUnit.MILLISECONDS);
```

What do you think?

shakuzen (approved):
Looks good. I left a nit and a question. This should fix things for the case of closing before scheduled publishing in a step. I think there will still be issues if close happens while scheduled publishing is happening, but we can worry about that separately from this change. That should be a much rarer case.

lenin-jaganathan:
the build failure seems unrelated to this PR and maybe a flaky one or something. I don't have permission to restart the job.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: Yeah, considering everything identified so far it seems like it will be better to leave it as is for now.

PR #3865, opened 2023-05-30
Proper scope handling in gRPC client instrumentation

ttddyy:
Prior to this change, scope in gRPC client instrumentation was not properly handled when client methods are called on different threads.
This change properly populates scope on callback methods.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/grpc/GrpcAsyncTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/grpc/GrpcAsyncTest.java
@@ -17,9 +17,8 @@
 package io.micrometer.core.instrument.binder.grpc;
 
 import com.google.common.util.concurrent.ListenableFuture;
-import io.grpc.ManagedChannel;
-import io.grpc.Metadata;
-import io.grpc.Server;
+import io.grpc.*;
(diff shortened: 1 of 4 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/grpc/GrpcAsyncTest.java
```diff
@@ -119,6 +124,76 @@ void simulate_trace_in_async_requests() {
...
+        ob.observe(() -> {
+            StreamObserver<SimpleRequest> requestObserver = asyncStub.clientStreamingRpc(responseObserver);
+
+            // perform onNext on a separate thread
+            AtomicBoolean onNextDone = new AtomicBoolean();
+            new Thread(() -> {
+                requestObserver.onNext(request);
+                onNextDone.set(true);
+            }).start();
+            await().untilTrue(onNextDone);
```
jonatan-ivanov, line 179: I'm thinking if it is better to use a `CountDownLatch` instead of Awaitility for this.
```java
CountDownLatch onNextDone = new CountDownLatch();
new Thread(() -> {
    requestObserver.onNext(request);
    onNextDone.countDown();
    }).start();
onNextDone.await();
```

Also, maybe using `CompletableFuture` (uses the common pool) instead of creating a new `Thred` might be better.

jonatan-ivanov (approved):

PR #3875, opened 2023-06-05
use seconds for OTLP export

zeitlinger:
use seconds for otlp export, as specified in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/8355


jonatan-ivanov:
Thanks for the PR!
Your link points to an issue in the Micrometer Bridge. I guess you meant this as "specified": https://github.com/open-telemetry/opentelemetry-specification/pull/3312

Based on the PR, the semantic conventions spec says that:
>When instruments are measuring durations, seconds (i.e. `s`) SHOULD be used.

I found this somewhat confusing. Why does the semantic conventions spec says anything about how to measure time, shouldn't it be in the SDK specs? Can one convention define one unit and another one define a different one? Why is it specified at all since OTel does not have a Timer concept.

Based on this, I don't think we should merge this PR since:
- Micrometer always measures duration in nanos and then converts it to the unit of the registry. No OTel specification should have any affect on "measuring durations" in Micrometer.
- Users of Micrometer may or may not use the OTel semantic conventions (please note that semantic conventions spec should not have any effect on OTLP) and OTLP does not define a time unit
- This is a breaking change

What do you think?

lenin-jaganathan:
@jonatan-ivanov Do you think it still makes sense to address the referenced issue? I mean providing configurable "baseUnit"?

zeitlinger:
> @lenin-jaganathan Definitely, especially because the OTel Collector is not converting time units to the backend that is forwarding to so the component that emits OTLP have to. This is very unfortunate but it's not a huge change on Micrometer's side.

I've given it a try: https://github.com/micrometer-metrics/micrometer/pull/3883

zeitlinger:
> * Users of Micrometer may or may not use the OTel semantic conventions

This is a key question - how does micrometer fit into OpenTelemetry ecosystem?

OTLP is a building block, but the otel semantic conventions are an important building block as well - becoming more important as they are stabilizing. 

One use case is an otel JVM dashboard. Would micrometer aim to support this once the semantic conventions are stable?

jonatan-ivanov:
Yes, we are planning to look into providing support for OTel Semconv (we also contributed to it).

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/3883

lenin-jaganathan:
@jonatan-ivanov Just wanted to get more clarity on your comment
> we are planning to look into providing support for OTel Semconv.

Does this mean JVM Metrics generated by micrometer will adhere to [OTEL Sem conv](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/runtime-environment-metrics/#jvm-metrics) for JVM Metrics?
I mean some MeterFilter similar to something [here](https://github.com/micrometer-metrics/micrometer/blob/main/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusRenameFilter.java) to convert all the JVM metrics to OTEL semconv equivalent?

And are there plans on adding filters for other binders as well. E.g: OTEL has sem conv for HTTP metrics and how is micrometer planning to adapt it?

jonatan-ivanov:
>Does this mean JVM Metrics generated by micrometer will adhere to [OTEL Sem conv](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/runtime-environment-metrics/#jvm-metrics) for JVM Metrics?
I mean some MeterFilter similar to something [here](https://github.com/micrometer-metrics/micrometer/blob/main/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusRenameFilter.java) to convert all the JVM metrics to OTEL semconv equivalent?

We can't promise it **will** since right now semconv is not stable but once it becomes stable and gains popularity we are planning to look into providing that feature to the users. This can be implemented via Micrometer providing `MeterFilter`s or adding variants/flags to the current instrumentation to produce data that is compliant to semconv. We must keep current behavior too (to not introduce breaking chnges).

>And are there plans on adding filters for other binders as well. E.g: OTEL has semconv for HTTP metrics and how is micrometer planning to adapt it?

I think the above should be applicable everywhere it makes sense (http, mongo, etc). I think the "how" will be a question for every single instrumentation.

zeitlinger:
> Yes, we are planning to look into providing support for OTel Semconv (we also contributed to it).

that's great to hear.

I think it would make sense to come back to this topic (making seconds the default base unit) together with the otel semconv then - because both topics will help users to re-use dashboards for all of their java applications.

PR #3581, opened 2023-01-08
Make cumulative count computations in `takeCountSnapshot` more efficient

ThePumpingLemma:
Instead of recomputing the cumulative sum again for each `CountAtBucket`, keep a running tally during iteration

PR #3922, opened 2023-06-20
Don't add `PercentileHistogramBuckets` when `supportsAggregablePercentiles` is false

### LINKED ISSUE #3904: Using publishPercentileHistogram breaks histogram values
**Describe the bug**

After upgrading from micrometer `1.10.6` to `1.11.0` we noticed test failures in our application regarding histograms.
Our application is the elastic APM Java Agent. While we do not use micrometer ourselves, we do provide instrumentations for capturing user defined metrics via micrometer. For these instrumentations we have unit tests, which are the ones which started to fail after upgrading to `1.11.0`. 

The failure we are seeing is that `DistributionSummaries` report invalid `CountAtBuckets` when also using `publishPercentileHistogram()`. See the test case and the comments for details.

**Environment**

 - Micrometer version `1.11.0`, same issue occurs with `1.11.1`
 - Micrometer registry `SimpleMeterRegistry`

**To Reproduce**

```java
    @Test
    void verifyHistogram() {
        SimpleMeterRegistry registry = new SimpleMeterRegistry();
        DistributionSummary meter = DistributionSummary
                .builder("my_histogram")
                .serviceLevelObjectives(5, 50, 95)
                .publishPercentileHistogram() //when commenting out this line, everything works fine on 1.11.0+
                .register(registry);

        for(int val : new int[]{22, 55, 66, 98}) {
            meter.record(val);
        }

        HistogramSnapshot snapshot = meter.takeSnapshot();
        CountAtBucket[] countAtBuckets = snapshot.histogramCounts();

        assertThat(countAtBuckets)
                .extracting(CountAtBucket::bucket)
                .containsExactly(5.0, 50.0, 95.0);

        //this assertion fails on 1.11.0+: The bucket values are [0.0, 0.0, 0.0]
        assertThat(countAtBuckets)
                .extracting(CountAtBucket::count)
                .containsExactly(0.0, 1.0, 3.0);
    }
```

**Expected behavior**

I would expect the bucket values to be as shown in the test and not depend on the presence/absence of `publishPercentileHistogram()`. If this is not a bug and instead caused by wrong usage from our side, please let me know.

shakuzen:
There was a mismatch between the underlying buckets used in a histogram and the buckets that would be returned in a HistogramSnapshot. This resulted in the wrong counts due to a change in the way we calculate the bucket count (see gh-3581), which assumes underlying buckets = buckets to return. This change makes the underlying buckets used the same as the buckets that will be returned, since there is no benefit of using more buckets than we will return.

A test has been added to the TCK to cover this case. It needs to be overly flexible since we don't know at the TCK level whether a MeterRegistry uses cumulative bucket counts or whether it supports aggregable percentile histogram buckets.

Resolves #3904

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
@@ -60,11 +60,6 @@ public class TimeWindowFixedBoundaryHistogram extends AbstractTimeWindowHistogra
         NavigableSet<Double> histogramBuckets = distributionStatisticConfig
             .getHistogramBuckets(supportsAggregablePercentiles);
 
-        Boolean percentileHistogram = distributionStatisticConfig.isPercentileHistogram();
-        if (percentileHistogram != null && percentileHistogram && supportsAggregablePercentiles) {
-            histogramBuckets.addAll(PercentileHistogramBuckets.buckets(distributionStatisticConfig));
-        }
-
         this.buckets = histogramBuckets.stream().filter(Objects::nonNull).mapToDouble(Double::doubleValue).toArray();
         initRingBuffer();
     }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
@@ -61,7 +61,7 @@ public class TimeWindowFixedBoundaryHistogram extends AbstractTimeWindowHistogra
             .getHistogramBuckets(supportsAggregablePercentiles);
 
         Boolean percentileHistogram = distributionStatisticConfig.isPercentileHistogram();
-        if (percentileHistogram != null && percentileHistogram) {
+        if (percentileHistogram != null && percentileHistogram && supportsAggregablePercentiles) {
             histogramBuckets.addAll(PercentileHistogramBuckets.buckets(distributionStatisticConfig));
         }
 
@@ -110,7 +110,7 @@ public class TimeWindowFixedBoundaryHistogram extends AbstractTimeWindowHistogra
 
     /**
      * For recording efficiency, we turn normal histogram into cumulative count histogram
-     * only on calls to {@link #countsAtValues(Iterator<Double>)}.
+     * only on calls to {@link FixedBoundaryHistogram#countsAtValues(Iterator)}.
      */
     @Override
     Iterator<CountAtBucket> countsAtValues(Iterator<Double> values) {
--- a/micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryCompatibilityKit.java
+++ b/micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryCompatibilityKit.java
@@ -322,22 +322,13 @@ public abstract class MeterRegistryCompatibilityKit {
             HistogramSnapshot snapshot = summary.takeSnapshot();
             CountAtBucket[] countAtBuckets = snapshot.histogramCounts();
 
-            assertThat(countAtBuckets).satisfiesAnyOf(
-                    MeterRegistryCompatibilityKit.this::assertHistogramNotSupportingPercentileHistogramBuckets,
-                    MeterRegistryCompatibilityKit.this::assertHistogramSupportingPercentileHistogramBuckets);
+            assertHistogramBuckets(countAtBuckets);
         }
 
     }
 
-    private void assertHistogramNotSupportingPercentileHistogramBuckets(CountAtBucket[] countAtBuckets) {
-        // only SLO buckets added because percentile histogram buckets are not supported
-        assertThat(countAtBuckets).extracting(CountAtBucket::bucket).containsExactly(5.0, 50.0, 95.0);
-
-        assertThat(countAtBuckets).extracting(CountAtBucket::count).containsExactly(0.0, 1.0, 3.0);
-    }
-
-    private void assertHistogramSupportingPercentileHistogramBuckets(CountAtBucket[] countAtBuckets) {
-        // percentile histogram buckets will be there, but assert SLO buckets are present
+    private void assertHistogramBuckets(CountAtBucket[] countAtBuckets) {
+        // percentile histogram buckets may be there, assert SLO buckets are present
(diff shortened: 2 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
```diff
@@ -61,7 +61,7 @@ public TimeWindowFixedBoundaryHistogram(Clock clock, DistributionStatisticConfig
             .getHistogramBuckets(supportsAggregablePercentiles);
 
         Boolean percentileHistogram = distributionStatisticConfig.isPercentileHistogram();
-        if (percentileHistogram != null && percentileHistogram) {
+        if (percentileHistogram != null && percentileHistogram && supportsAggregablePercentiles) {
```
shakuzen, line 64: This is the only change to main code. This makes sure we don't add buckets unless they will be used. If a registry does not `supportsAggregablePercentiles` then it should not do anything when `percentileHistogram` is enabled.

micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryCompatibilityKit.java
```diff
@@ -300,6 +301,68 @@ void histogramCounts() {
...
+                    MeterRegistryCompatibilityKit.this::assertHistogramSupportingPercentileHistogramBuckets);
+        }
+
+    }
+
+    private void assertHistogramNotSupportingPercentileHistogramBuckets(CountAtBucket[] countAtBuckets) {
+        // only SLO buckets added because percentile histogram buckets are not supported
+        assertThat(countAtBuckets).extracting(CountAtBucket::bucket).containsExactly(5.0, 50.0, 95.0);
+
+        assertThat(countAtBuckets).extracting(CountAtBucket::count).containsExactly(0.0, 1.0, 3.0);
```
shakuzen, line 336: There's unfortunately no point in the assertions in this method since the other method will always be true when these assertions are true, and we need the lax conditions there for the reasons in the comments.
I'll simplify by getting rid of this method.

shakuzen:
An alternative to this change would have been reverting https://github.com/micrometer-metrics/micrometer/pull/3581. If there's some reason this change causes other issues (which aren't being caught by existing tests), we can consider that option.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
```diff
@@ -60,11 +60,6 @@ public TimeWindowFixedBoundaryHistogram(Clock clock, DistributionStatisticConfig
         NavigableSet<Double> histogramBuckets = distributionStatisticConfig
             .getHistogramBuckets(supportsAggregablePercentiles);
 
-        Boolean percentileHistogram = distributionStatisticConfig.isPercentileHistogram();
-        if (percentileHistogram != null && percentileHistogram) {
-            histogramBuckets.addAll(PercentileHistogramBuckets.buckets(distributionStatisticConfig));
```
shakuzen, line 65: An earlier version added `&& supportsAggregablePercentiles` to the if condition here, which makes this correct, but upon a closer look, this was unnecessary in the first place. `PercentileHistogramBuckets` are already conditionally added in the call immediately before this removed code: `DistributionStatisticConfig#getHistogramBuckets`.

ThePumpingLemma:
@shakuzen I took a look at this and came up with a different approach that fixes the provided test case:

```
diff --git a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryHistogram.java b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryHistogram.java
index 3f6e0a5d2..042a10814 100644
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryHistogram.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryHistogram.java
@@ -33,13 +33,6 @@ class FixedBoundaryHistogram {
         this.isCumulativeBucketCounts = isCumulativeBucketCounts;
     }

-    long countAtValue(double value) {
-        int index = Arrays.binarySearch(buckets, value);
-        if (index < 0)
-            return 0;
-        return values.get(index);
-    }
-
     void reset() {
         for (int i = 0; i < values.length(); i++) {
             values.set(i, 0);
@@ -73,8 +66,13 @@ class FixedBoundaryHistogram {
         return low < buckets.length ? low : -1;
     }

+    int getCountIndex(double value) {
+        return Math.max(Arrays.binarySearch(buckets, value), 0);
+    }
+
     Iterator<CountAtBucket> countsAtValues(Iterator<Double> values) {
         return new Iterator<CountAtBucket>() {
+            private int previousCountIndex = 0;
             private double cumulativeCount = 0.0;

             @Override
@@ -85,13 +83,17 @@ class FixedBoundaryHistogram {
             @Override
             public CountAtBucket next() {
                 double value = values.next();
-                double count = countAtValue(value);
+
                 if (isCumulativeBucketCounts) {
-                    cumulativeCount += count;
+                    int currentCountIndex = getCountIndex(value);
+                    for (int i = previousCountIndex; i <= currentCountIndex; i++) {
+                        cumulativeCount += FixedBoundaryHistogram.this.values.get(i);
+                    }
+                    previousCountIndex = currentCountIndex + 1;
                     return new CountAtBucket(value, cumulativeCount);
                 }
                 else {
-                    return new CountAtBucket(value, count);
+                    return new CountAtBucket(value, FixedBoundaryHistogram.this.values.get(getCountIndex(value)));
                 }
             }
         };
```

I noticed that the cumulative sum in https://github.com/micrometer-metrics/micrometer/pull/3581 didn't sum all of the values between successive indexes returned by the binary search. This should hopefully revert to the previous behavior while retaining the performance improvement.

shakuzen:
@ThePumpingLemma thank you for taking a look. It's much appreciated. I think we could consider making the algorithm more robust, but I don't know that it's needed. The issue arose because there were cases where the histogram used more buckets internally than it would return when taking a snapshot. If the buckets used internally match the buckets returned in a snapshot, I don't think we run into the case your proposed change solves for, but I may be missing something. If I'm not missing something, I think we can consider the change separately from the change here. What do you think?

ThePumpingLemma:
@shakuzen The original code had a loop which ensured that all counts were summed, which did handle the bucket count mistmatches: https://github.com/micrometer-metrics/micrometer/pull/3581/files#diff-0cdfd1340b54d9c8884a8993dd43dd9bff1749050bfa6144233eabff4981b23eL159-L160

Given it's probably just robustness change at this point, agreed that it can be be considered separately. Thanks for fixing this!

Issue #3914, opened 2023-06-19
StepMeterRegistry publishes on close even if config.enabled is false


felixkrull-neuland:
**Describe the bug**
When a `StepMeterRegistry` is closed, it calls `publish` without first checking if `config.enabled()` returns true [^1]. This is in contrast to `PushMeterRegistry` which explicitly checks for enabled before calling `publish` in `close` [^2]

[^1]: https://github.com/micrometer-metrics/micrometer/blob/413f2b79edebcd1c2c85bd4394a87ec121612444/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java#L146
[^2]: https://github.com/micrometer-metrics/micrometer/blob/413f2b79edebcd1c2c85bd4394a87ec121612444/micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java#L134

The consequence is that in our integration tests, DatadogMeterRegistry tries to publish to Datadog when the registry is closed even though publishing is disabled for that environment. It's not a huge deal in general because setups that shouldn't publish metrics also don't have API keys configured, but the big error message in the log is annoying.

**Environment**

 - Micrometer 1.11.1
 - Datadog registry

**Expected behavior**
`StepMeterRegistry.close` should also check for enabled before calling `publish`.

PR #3924, opened 2023-06-20
StepMeterRegistry publishes on close even if config.enabled is false 

lenin-jaganathan:
Fixes https://github.com/micrometer-metrics/micrometer/issues/3914

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -125,9 +125,6 @@ dependencies {
     optionalApi project(':micrometer-observation')
     optionalApi 'io.micrometer:context-propagation'
 
-    // Common Tests
-    testImplementation project(':micrometer-test')
-
     // kotlin
     optionalApi 'org.jetbrains.kotlin:kotlin-reflect'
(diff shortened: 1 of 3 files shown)


micrometer-core/build.gradle
```diff
@@ -125,6 +125,9 @@ dependencies {
     optionalApi project(':micrometer-observation')
     optionalApi 'io.micrometer:context-propagation'
 
+    // Common Tests
+    testImplementation project(':micrometer-test')
```
shakuzen, line 129: micrometer-test depends on micrometer-core. This is a circular dependency.

lenin-jaganathan: Yeah, I just realised. I can remove this as this is not essential for this change.

shakuzen: Since it is used in test scope only, I think we can use it, on further thought. I didn't look into the details of its use yet to give any thought to whether it makes sense or not, though.

shakuzen (approved):
Thank you.

PR #3926, opened 2023-06-21
Constant version of otel collector for integration test

### LINKED ISSUE #3925: OTelCollectorIntegrationTest failing due to breaking change in OTel Collector
In https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.80.0, a previous breaking change that broke our integration test has been reverted, breaking our integration tests again.

> prometheusreceiver, prometheusexporter, prometheusremotewrite: Disable pkg.translator.prometheus.NormalizeName feature gate by default (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/23208)
The feature gate pkg.translator.prometheus.NormalizeName was enabled prematurely while translation
on the prometheus receiver was incomplete. To address this, the feature gate has been reverted back to alpha status.
This will remain the case until the translation on the receiver side aligns with the translation on the exporter side,
or until it is replaced with a configuration option or completely removed. To maintain the current behavior, you can
enable the feature gate using the --feature-gates=pkg.translator.prometheus.NormalizeName command argument.
However, please note that the translation in the prometheus receiver is a subject to future changes.

shakuzen:
We should have reproducible builds and maintenance branches frequently breaking on new versions of the otel collector blocks other work like unrelated bug fixes. This change fixes the version and manages it in the build file. It's not the same as how other dependencies are managed (no version range, no lock file), but it's maybe an improvement and keeps versions in build files rather than in a Java test class. Once merged to `main` it can be set to `latest` until that branches off as a maintenance branch.

This does add a little work to run the integration test from an IDE because the system property configuration needs to be added to your IDE run config.

I'm open to feedback on taking a different approach. This is just what I thought of on first try.

Closes #3925

### DIFF
--- a/implementations/micrometer-registry-otlp/build.gradle
+++ b/implementations/micrometer-registry-otlp/build.gradle
@@ -13,6 +13,10 @@ dependencies {
     testImplementation 'org.awaitility:awaitility'
 }
 
+dockerTest {
+    systemProperty 'otel-collector-image.version', '0.79.0'
+}
+
 // new module in 1.9. This should be removed in later branches.
 japicmp.enabled = false
 downloadBaseline.enabled = false
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
@@ -20,8 +20,10 @@ import io.restassured.response.Response;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.utility.DockerImageName;
 
 import java.time.Duration;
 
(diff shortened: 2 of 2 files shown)


implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
```diff
@@ -46,12 +48,25 @@ class OTelCollectorIntegrationTest {
...
         .withCommand("--config=/etc/" + CONFIG_FILE_NAME)
         .withClasspathResourceMapping(CONFIG_FILE_NAME, "/etc/" + CONFIG_FILE_NAME, READ_ONLY)
+        .waitingFor(Wait.forLogMessage(".*Everything is ready.*", 1))
         .withExposedPorts(4318, 9090); // HTTP receiver, Prometheus exporter
 
+    private static String getCollectorImageVersion() {
+        String version = System.getProperty("otel-collector-image.version");
+        if (version == null) {
+            throw new IllegalStateException(
+                    "System property 'otel-collector-image.version' is not set. This should be set in the build configuration for running from the command line. If you are running OTelCollectorIntegrationTest from an IDE, set the system property to the desired collector image version.");
```
shakuzen, line 65: This error message is intended to make it obvious what to do for anyone who tries to run the test in the IDE for the first time, which will fail (unlike before this change).

implementations/micrometer-registry-otlp/build.gradle
```diff
@@ -13,6 +13,10 @@ dependencies {
     testImplementation 'org.awaitility:awaitility'
 }
 
+dockerTest {
+    systemProperty 'otel-collector-image.version', '0.79.0'
```
shakuzen, line 17: Rather than use the latest version (`0.80.0` now) which requires changes to the test assertions, I thought it better to keep the maintenance branches unchanged, and we will update only the `main` branch to use the latest version and adapt to the changes needed. Let me know if anyone disagrees with that direction.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
```diff
@@ -46,12 +48,25 @@ class OTelCollectorIntegrationTest {
...
 
+    private static final DockerImageName COLLECTOR_IMAGE = DockerImageName
+        .parse("otel/opentelemetry-collector-contrib:" + getCollectorImageVersion());
+
     @Container
-    private final GenericContainer<?> container = new GenericContainer("otel/opentelemetry-collector-contrib")
+    private final GenericContainer<?> container = new GenericContainer(COLLECTOR_IMAGE)
         .withCommand("--config=/etc/" + CONFIG_FILE_NAME)
         .withClasspathResourceMapping(CONFIG_FILE_NAME, "/etc/" + CONFIG_FILE_NAME, READ_ONLY)
+        .waitingFor(Wait.forLogMessage(".*Everything is ready.*", 1))
```
shakuzen, line 58: Maybe not needed, but it seemed like this was a better indication of being ready to receive traffic than the port being open (which I believe was the default wait strategy used before).

shakuzen:
I'll merge for now so our builds can stop being broken. Feel free to leave feedback post-merge.

Issue #3934, opened 2023-06-23
OTLP missing bucket count for positive infinity


lenin-jaganathan:
**Describe the bug**
https://github.com/open-telemetry/opentelemetry-proto/blob/4967b51b5cb29f725978362b9d413bae1dbb641c/opentelemetry/proto/metrics/v1/metrics.proto#L426-L427

The above spec from OpenTelemetry specifies that the count of bucket_counts should be one greater than the count of explicit_bounds with the last bucket_count representing the count of values between last bucket and Positive Infinity. The current implementation in Micormeter doesn't track the values between the last bucket and POSITIVE_INFINITY by default. Even if PSOITIVE_INFINITY is added as an SLO value, the micrometer will add an explicit_bound with the value  POSITIVE_INFINITY since the micrometer does a strict 1-1 mapping of values-count.

**Environment**
 - Micrometer version - 1.11.1 (but it should also exist with 1.10.x)
 - Micrometer registry - OTLP

**To Reproduce**
How to reproduce the bug:
Configure an OTLPMeterRegistry and add a Timer with histograms enabled.

**Expected behavior**
bucket_count should be one greater than explicit_bounds and it should represent values between last bucket and positive infinity.

shakuzen:
Thanks for bringing this to our attention. Maybe it's just me, but comments in a proto definition don't seem like the best place to specify these things. There's no mention of this in the [published documentation for histograms](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram). Could you open an issue with OpenTelemetry to get the relevant info reflected in the documentation? I'm also wondering how OTLP producers are supposed to catch these kinds of issues. It would have been nice if we got some validation failure or at least warning somewhere. Neither the proto nor the OTel collector complain about bucket_count having the wrong number of buckets.

lenin-jaganathan:
Maybe the word "must" in the proto is the cue for producers. I will open up an issue on the open telemetry docs to address this. 

What I see is some of the exporters we are using currently, just ignore all the buckets because the count is not one greater than buckets. I did spend a lot of time debugging but using a LoggingExporter which just prints things as is and I don't see any logs for that. I have to read the proto and try other SDKs (I used NodeJs) to see the difference is the number of counts.

lenin-jaganathan:
https://github.com/micrometer-metrics/micrometer/pull/3935 - PR for fixing this. Somehow default linking of issues to PR didn't work.

[...]

PR #3935, opened 2023-06-24
Add +Infinity bucket count to OTLP Histogram

lenin-jaganathan:
Fixes https://github.com/micrometer-metrics/micrometer/issues/3934

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
@@ -45,6 +45,7 @@ import io.opentelemetry.proto.resource.v1.Resource;
 
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -516,4 +532,25 @@ static Histogram getHistogram(Clock clock, DistributionStatisticConfig distribut
...
+        if (sloBoundaries == null || sloBoundaries.length == 0) {
+            // When there are no SLO's associated with DistributionStatisticConfig we will
+            // add one with Positive
+            // Infinity. This will make sure we always have POSITIVE_INFINITY, and the
+            // NavigableSet will make sure
+            // duplicates if any will be ignored.
+            return EMPTY_SLO_WITH_POSITIVE_INF;
+        }
+
+        boolean containsPositiveInf = Arrays.stream(sloBoundaries).anyMatch(value -> value == Double.POSITIVE_INFINITY);
```
lenin-jaganathan, line 547: Technically we can also use Double.MAX_VALUE for representing this. But, Double.POSITIVE_INFINITY is already part of the default buckets and hence I chose to prefer POSITIVE_INFINITY over MAX_VALUE.

shakuzen (approved):

shakuzen:
It seems we don't have any test in OtlpDeltaMeterRegistryTest that verifies this change. It would be good to have one, but I'll merge anyway and we can follow-up with that.

lenin-jaganathan:
@shakuzen [These assertions](https://github.com/lenin-jaganathan/micrometer/blob/6187fb3fbdeb357f71e2dbd71fc31e1317ffefc5/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMeterRegistryTest.java#L203-L229) in the common tests is used by DeltaRegistry which covers this use-case. It would be better for the tests in Cumulative also to use assertion on the Proto objects rather than the String representation since it is very hard to read and make any changes to the String representations used in OTLPCumulativeRegistry. Maybe that is what we might need to address.

PR #3937, opened 2023-06-26
Fix Javadoc for ObservationThreadLocalAccessor(ObservationRegistry)

izeye:
This PR sets the `instance` static field in the `ObservationThreadLocalAccessor(ObservationRegistry)` constructor as it seems to have been missed based on its Javadoc.


marcingrzejszczak:
I don't think we should be setting this because for the `ContextRegistry` default instantiation, the default constructor will be called. If someone is manually updating a `ContextRegistry` then they should manually update the `instance`. They might also be using an instance for tests and don't want the test instance to be used in other parts of the system.

izeye:
@marcingrzejszczak Thanks for the feedback! I fixed its Javadoc instead.

PR #3938, opened 2023-06-26
Use different name for StepMeterRegistry thread that polls to roll over

### LINKED ISSUE #3837: Thread that rolls over meters in StepMeterRegistry shares same name with publishing thread
Originally reported by @pirgeo in https://github.com/micrometer-metrics/micrometer/issues/3832. I can see how it is confusing to have multiple threads using the naming of the thread actually used for doing the publishing, despite only one of them doing publishing. It may be a good idea to name the thread that polls meters differently to avoid this confusion. Though if we do not use the `ThreadFactory` passed to the `start` method, we cannot (easily) name the thread specific to the underlying registry. Initially I thought this would not be a problem and it made more sense to name the thread something like `step-meter-registry-poller`, but in the case where a user has multiple step meter registries in use, this will also be hard to understand.

> However, when I switch over to Micrometer 1.11.0-RC1, I suddenly have two publishing threads ( `dynatrace-metrics-publisher` and `dynatrace-metrics-publisher-2`).

izeye:
This is an attempt to resolve #3837.

Closes gh-3837

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
@@ -26,6 +26,7 @@ import io.micrometer.core.instrument.internal.DefaultGauge;
 import io.micrometer.core.instrument.internal.DefaultLongTaskTimer;
 import io.micrometer.core.instrument.internal.DefaultMeter;
 import io.micrometer.core.instrument.push.PushMeterRegistry;
+import io.micrometer.core.instrument.util.NamedThreadFactory;
 
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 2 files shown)


shakuzen (commented):
Thanks for the pull request. I'm leaning toward it being better to unconditionally use a different name rather than try to guess if a user wants to use the same `ThreadFactory` with the poller thread based on the type of the `ThreadFactory`. Until we understand a good use case for that, I think it's better we leave it as an implementation detail. What do you think?

izeye:
@shakuzen Thanks for the feedback! I thought that it's safer to respect the provided `ThreadFactory` if it's unknown, but if you're comfortable with doing it unconditionally, I'm okay with it. Just let me know your final preference on it.

shakuzen:
Yes, let's go with the passed `ThreadFactory` being for the publishing thread only. We can update the JavaDoc to make this more explicit. Let's keep the poller thread an internal detail until we understand the user need to customize the thread via a `ThreadFactory`.

izeye:
@shakuzen Thanks! I updated this accordingly.

shakuzen (approved):
Thanks

PR #3939, opened 2023-06-27
Polish "SimpleObservation#makeCurrent()"

ttddyy:
- Remove outdated javadoc
- Reduce loop and duplicated logic


micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java
```diff
@@ -366,29 +356,22 @@ public void reset() {
          */
         @Override
         public void makeCurrent() {
+            Deque<SimpleObservation> observations = new ArrayDeque<>();
```
marcingrzejszczak, line 359: After our fixes in `ObservationThreadLocalAccessor` calling this method shouldn't be needed anymore. Do you have any use case for it?

ttddyy: No, after the gRPC fix, I don't have a particular usage for `makeCurrent` method "anymore".

jonatan-ivanov:
So as I understand, this is not needed anymore right?
Please let us know if this is not correct so that we can reopen.

Issue #3943, opened 2023-06-28
ObservationThreadLocalAccessor.restore() generating errors in tests


nvervelle:
**Describe the bug**
When upgrading from spring-boot 3.0.4 (micrometer 1.10.4) to spring-boot 3.0.8 or 3.1.1 (micrometer 1.11.1), some of my tests are now failing with a message comparing a `NoopObservation` with null :

> Observation `<io.micrometer.observation.NoopObservation@390cd91e>` to which we're restoring is not the same as the one set as this scope's parent observation `<null>` . Most likely a manually created Observation has a scope opened that was never closed. This may lead to thread polluting and memory leaks

It seems related to #3831 which introduced `NullObservation` and the verification in `ObservationThreadLocalAccessor.restore()`.

**Environment**

 - Micrometer version 1.11.1
 - OS: Ubuntu
 - Java version: openjdk version "17.0.1" 2021-10-19

**To Reproduce**

I'm trying to extract a test case from my own code to show how to reproduce it. I will post it here once I'm able to reproduce it on a sample application.

For the moment, it seems related to observations created by `Mono.tap(Micrometer.observation(registry))`.

Exemple of test that fails with the new version :
```java
  @Test
  @DisplayName("Save when the bucket does not exist. Must terminate with errors.")
  void saveWhenBucketDoesNotExist() throws IOException {
    // GIVEN
    final Resource fileResource = ClasspathResourceLoader.fromFile("input/ID-card.png");

    // WHEN
    final Mono<FileReference> responseFileReferenceMono =
        s3Repository.save(fileResource.getFile().toPath());

    // THEN
    StepVerifier.create(responseFileReferenceMono)
        .expectError(NoSuchBucketException.class)
        .verify();
  }
```

Code under test :
```java
  public Mono<FileReference> save(final Path path) {
    return save(AsyncRequestBody.fromFile(path), path.toFile().length());
  }

  private Mono<FileReference> save(
      final AsyncRequestBody asyncRequestBody,
      final long length) {
    return Mono.fromFuture(() -> putObject(asyncRequestBody, fileReference, length))
        .name("s3 put-object")
        .doOnError(NoSuchBucketException.class, e -> LOGGER.error("Unknown bucket"))
        .tap(Micrometer.observation(observationService.registry()))
        .thenReturn(fileReference);
  }

  private CompletableFuture<PutObjectResponse> putObject(
      final AsyncRequestBody asyncRequestBody,
      final long length) {
[...]

PR #3945, opened 2023-06-29
:bug: Avoid potential race condition with Observation.NOOP

nvervelle:
It seems that in some situations `Observation.NOOP` can be null even if `NoopObservation.INSTANCE` is properly set (see [comment in issue](https://github.com/micrometer-metrics/micrometer/issues/3943#issuecomment-1612566866)).

This PR removes `NoopObservation.INSTANCE` so that there's no duplication and no risk of race condition on `Observation.NOOP`. This change resolves the NPE reported in the [comment](https://github.com/micrometer-metrics/micrometer/issues/3943#issuecomment-1612566866), but is not sufficient to resolve the problem reported by #3943

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/NoopObservation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/NoopObservation.java
@@ -29,14 +29,9 @@ import io.micrometer.common.lang.Nullable;
  */
 final class NoopObservation implements Observation {
 
-    /**
-     * Instance of {@link NoopObservation}.
-     */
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
Can you create this PR against 1.10.x branch?

nvervelle:
> Can you create this PR against 1.10.x branch?

Sure @marcingrzejszczak , I've created PR #3946

marcingrzejszczak:
thanks! closing in favour of https://github.com/micrometer-metrics/micrometer/pull/3946

PR #3946, opened 2023-06-29
3943: :bug: Avoid potential race condition with Observation.NOOP

nvervelle:
Same as #3945 but for 1.10.x branch


marcingrzejszczak:
thank you!

PR #3948, opened 2023-06-29
fix support for "seconds" base time unit

zeitlinger:
histograms have a lower bound of 1s if you use seconds as the base time unit

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
@@ -22,6 +22,7 @@ import io.micrometer.core.instrument.internal.Mergeable;
 import java.time.Duration;
 import java.util.NavigableSet;
 import java.util.TreeSet;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.LongStream;
 
 /**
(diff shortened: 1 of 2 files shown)


zeitlinger:
@jonatan-ivanov does this make sense in this way?

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
```diff
@@ -110,7 +117,11 @@ public NavigableSet<Double> getHistogramBuckets(boolean supportsAggregablePercen
         NavigableSet<Double> buckets = new TreeSet<>();
 
         if (percentileHistogram != null && percentileHistogram && supportsAggregablePercentiles) {
-            buckets.addAll(PercentileHistogramBuckets.buckets(this));
+            long scale = baseTimeUnit.toMillis(1);
```
lenin-jaganathan, line 120: Trying to understand why this is a problem? I believe this is already taken care of since the buckets are always converted to Nanos and recording happens in Nanos. When exporting things are again converted to the baseTimeunit(). Am I missing something?

zeitlinger: I just double checked and saw that I'm using the a different histogram implementation that I thought I had (the one from the OpenTelemetry SDK).

PR #3949, opened 2023-06-30
add DistributionStatisticConfig.toBuilder()

zeitlinger:
Add method to turn DistributionStatisticConfig into a builder - if you just want to change a single property from the default


shakuzen:
Thanks for the pull request and sorry this took so long to get reviewed. Could you provide more details on the situation in which this would be useful? I want to make sure we understand the use case and that it's still needed with the latest API.

shakuzen:
@zeitlinger if you have a chance, could you take a look at my previous comment?

zeitlinger:
can't remember - let's close

shakuzen:
Thanks for the response. If you do remember or other users have a use case for this, let's discuss and we can always reopen in the future. My first comment here was a little cryptic with reference to "the latest API" and I don't exactly remember but perhaps I was referring to https://github.com/micrometer-metrics/micrometer/pull/4097, in case anyone in the future wonders.

PR #3950, opened 2023-06-30
Ensure OTLP histograms always contain bucket counts

jack-berg:
The `micrometer-registry-otlp` artifact emits histograms without the [bucket_counts](https://github.com/open-telemetry/opentelemetry-proto/blob/c4dfbc51f3cd4089778555a2ac5d9bc093ed2956/opentelemetry/proto/metrics/v1/metrics.proto#L428) field.

Although this field is optional and the value is redundant in cases where `explicit_bounds` are not set (if `explicit_bounds` are unset its a single bucket histogram with range [-infinity, infinity] and therefore the single entry in `bucket_counts` is redundant with the total `count` field). 

Its in these types of cases that `micrometer-registry-otlp` omits `bucket_counts`. This PR adjusts the behavior to _always_ include `bucket_counts`, which makes the messages easier to consume.

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
@@ -45,6 +45,7 @@ import io.opentelemetry.proto.resource.v1.Resource;
 
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executors;
@@ -393,10 +394,11 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
                     .addExplicitBounds(isTimeBased ? countAtBucket.bucket(getBaseTimeUnit()) : countAtBucket.bucket());
                 histogramDataPoint.addBucketCounts((long) countAtBucket.count());
             }
-            metricBuilder.setHistogram(io.opentelemetry.proto.metrics.v1.Histogram.newBuilder()
-                .setAggregationTemporality(otlpAggregationTemporality)
-                .addDataPoints(histogramDataPoint));
-            return metricBuilder.build();
+        }
+        // else treat as single bucket histogram
+        else {
+            histogramDataPoint.addAllExplicitBounds(Collections.emptyList());
+            histogramDataPoint.addBucketCounts(count);
         }
 
         return metricBuilder
@@ -415,7 +417,9 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
                     .setStartTimeUnixNano(getStartTimeNanos((functionTimer)))
                     .setTimeUnixNano(getTimeUnixNano())
                     .setSum(functionTimer.totalTime(getBaseTimeUnit()))
-                    .setCount((long) functionTimer.count()))
+                    .setCount((long) functionTimer.count())
+                    .addAllExplicitBounds(Collections.emptyList())
+                    .addBucketCounts((long) functionTimer.count()))
                 .setAggregationTemporality(otlpAggregationTemporality))
             .build();
     }
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpCumulativeMeterRegistryTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpCumulativeMeterRegistryTest.java
@@ -91,11 +91,11 @@ class OtlpCumulativeMeterRegistryTest extends OtlpMeterRegistryTest {
         timer.record(111, TimeUnit.MILLISECONDS);
         clock.add(otlpConfig().step());
         timer.record(4, TimeUnit.MILLISECONDS);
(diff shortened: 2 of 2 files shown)


implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpCumulativeMeterRegistryTest.java
```diff
@@ -91,11 +91,11 @@ void timer() {
         timer.record(111, TimeUnit.MILLISECONDS);
         clock.add(otlpConfig().step());
         timer.record(4, TimeUnit.MILLISECONDS);
-        assertThat(writeToMetric(timer).toString()).isEqualTo(
-                "name: \"web.requests\"\n" + "description: \"timing web requests\"\n" + "unit: \"milliseconds\"\n"
-                        + "histogram {\n" + "  data_points {\n" + "    start_time_unix_nano: 1000000\n"
-                        + "    time_unix_nano: 60001000000\n" + "    count: 4\n" + "    sum: 202.0\n" + "  }\n"
-                        + "  aggregation_temporality: AGGREGATION_TEMPORALITY_CUMULATIVE\n" + "}\n");
+        assertThat(writeToMetric(timer).toString()).isEqualTo("name: \"web.requests\"\n"
```
lenin-jaganathan, line 94: I guess if we are making a change at the lower layer(in this case OTEL protocol). It would be great to add tests that cover both Cumulative and Delta cases.
@jack-berg Do you mind if I add additional commits with tests?

lenin-jaganathan:
I am confused by the use of "must" in [this metrics proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L381-L383) where it says,

> If the histogram does not contain the distribution of values, then both "explicit_bounds" and "bucket_counts" must be omitted and only "count" and "sum" are known.

May be this should be fixed to reflect what is expected of producers. Though it is not a specification, it still seems to be misleading.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpCumulativeMeterRegistryTest.java
jack-berg: Please do!

lenin-jaganathan: I just realized that I wouldn't have permission. But that's not a big deal, adding things to OtlpMeterRegistryTest#assertHistogram to assert the bucket list is always non-zero.

The ideal way to do it should be to assert that the sum of values in the buckets_count list is equal to the count. But due to [this open issue](https://github.com/micrometer-metrics/micrometer/issues/3934), the numbers won't match up when values fall into Positive Infinity. So for now asserting that bucketCounts is always Positive would be a good start.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -393,10 +394,11 @@ Metric writeHistogramSupport(HistogramSupport histogramSupport) {
...
                 histogramDataPoint.addBucketCounts((long) countAtBucket.count());
             }
-            metricBuilder.setHistogram(io.opentelemetry.proto.metrics.v1.Histogram.newBuilder()
-                .setAggregationTemporality(otlpAggregationTemporality)
-                .addDataPoints(histogramDataPoint));
-            return metricBuilder.build();
+        }
+        // else treat as single bucket histogram
+        else {
+            histogramDataPoint.addAllExplicitBounds(Collections.emptyList());
```
shakuzen, line 400: Unless I'm missing something, this line doesn't serve any purpose. Adding an empty list won't effectively change anything, right?

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpCumulativeMeterRegistryTest.java
shakuzen: The linked issue is now fixed, which means this PR will need to be rebased to resolve conflicts.

shakuzen (commented):
In addition to the point by @lenin-jaganathan about the comment in the proto which says we "must" not do what is proposed here, I don't understand why the changes here "makes the messages easier to consume". Could you expand on that, please? An OTLP consumer should accept OTLP that meets the specification, and the currently produced OTLP meets specification as far as I understand. I'm confused about the notion of easier. This will add more data to send over the wire and parse for every histogram while providing no information that wasn't already available.

jack-berg:
> I am confused by the use of "must" in [this metrics proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L381-L383) where it says,
>
>> If the histogram does not contain the distribution of values, then both "explicit_bounds" and "bucket_counts" must be omitted and only "count" and "sum" are known.

I missed that comment in the proto. That makes it pretty clear that the current behavior is in fact aligned with the spec. Closing this PR. Sorry for the confusion.

Issue #3891, opened 2023-06-09
Unnecessary ThreadLocal overhead in LogbackMetrics for disabled log levels


onukristo:
**Describe the bug**

An application is using Logback and `io.micrometer.core.instrument.binder.logging.LogbackMetrics` is enabled, via Logback turbo filters system. Notice that this is the case of almost every Spring Boot service out there.

Now, let's say an application is using `log.debug("Foo")` somewhere, and logging level for that logger is set to INFO.
The expectation of developers here is that this call will be super cheap.

However the `log.debug` call ends up in `io.micrometer.core.instrument.binder.logging.MetricsTurboFilter#decide`, which now does one ThreadLocals get, one put and one removal.

And, accessing thread local is actually relatively heavy operation.

If the application has written that logging statement as following:

```java
if (log.isDebugEnabled()){
  log.debug("Foo");
}
```

Then the issue does not happen, because following block in LogbackMetrics eliminates the thread local accesses.

```java
        // When filter is asked for decision for an isDebugEnabled call or similar test,
        // there is no message (ie format)
        // and no intention to log anything with this call. We will not increment counters
        // and can return immediately and
        // avoid the relatively expensive ThreadLocal access below. See also logbacks
        // Logger.callTurboFilters().
        if (format == null) {
            return FilterReply.NEUTRAL;
        }
```

But `if (log.isDebugEnabled())` clause is rarely used in java world.

Profiler example as well to illustrate the issue: 
![image](https://github.com/micrometer-metrics/micrometer/assets/5344757/41e3d0cb-fa4d-46f7-be93-6756b151a85c)

The profiling was the thing, that made me wonder - how come we have so many log.debug stacks visible in our flame graphs?

**Environment**
- Micrometer version 1.10.7
- Spring Boot 3.0.7
- logback-core, logback-classic: 1.4.7

**To Reproduce**
1. Create a default Spring Boot Application or enable LogbackMetrics in a simple java app with Logback as logging system.
2. Set logging level to INFO.
3. Add log.debug("Foo") statement.
4. Step into it.

**Expected behavior**
`log.debug("Foo")` is expected to be lightning fast in java ecosystem.

I think the fix is just about replacing 
```java
if (format == null) {
  return FilterReply.NEUTRAL;
}
```

with

```java
if (!level.isGreaterOrEqual(logger.getEffectiveLevel())){
   return FilterReply.NEUTRAL;
}
```

**Additional context**
[...]

PR #3952, opened 2023-07-03
Avoid potential overhead of ThreadLocal.remove in LogbackMetrics

spmason:
In one of our apps we're seeing that over time the size of the ThreadLocalMap underpinning the ThreadLocal class gets large (due to other libraries overusing/abusing ThreadLocals). When this happens `ThreadLocal.remove` can start to become a performance bottleneck as it does some `expungeStaleEntry` calls that iterate over the whole map.

Because logger calls can be so frequent (even if they're suppressed this code is still called) the `TurboFilter` gets called _a lot_, and thus so does `ThreadLocal.remove`. Therefore Micrometer ends up slowing down all our code.

I agree that technically this isn't Micrometer's fault, but the potential for it to happen in any application is still there and MM should really do what it can to avoid becoming a bottleneck.

So this PR changes `LogbackMetrics` to remove the need for it to call `ThreadLocal.remove` and thus the potential for this bug. It does this by _not clearing the ThreadLocal after use_ and instead setting it to a 'holder' class that has its own boolean flag that gets flipped on/off inside `ignoreMetrics`.

Running the LogbackBenchmarks this seems to be possibly faster than the regular code even when ThreadLocals aren't overloaded, so it seems like a decent improvement overall

Of course the tradeoff is that now micrometer 'pollutes' all threads with an extraneous ThreadLocal value that it never cleans up. However:

1. If a thread doesn't log(?) then it won't get a ThreadLocal. If it does then the chances are it will log again (soon?) so the ThreadLocal will be needed again then anyway.
2. It's not a big object, one boolean
3. It should be cleaned up when the threads die

In developing this fix I did try a couple of other solutions:
1. Just call `ThreadLocal.set(false)` instead of `remove`. Unfortunately `set` does some similar cleanup to `remove` so the issue would still be there
2. Use an `AtomicBoolean` instead of a custom class to hold the flag. This wasn't performant enough as `AtomicBoolean` does a load of locking etc as it's meant to be used across threads - as these flags are explicitly only used on a single thread that locking is unnecessary overhead
3. Use a boolean[] as the flag holder. This seems a bit too ugly to me

Note that there is a similar issue to this one that might have helped this situation (https://github.com/micrometer-metrics/micrometer/issues/3891) as most of the time this is happening for me on some suppressed `logger.debug` calls. However even without that fix this would still be an issue for every non-suppressed logger call that my app made.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
@@ -47,7 +47,7 @@ import static java.util.Collections.emptyList;
 @NonNullFields
 public class LogbackMetrics implements MeterBinder, AutoCloseable {
 
-    static ThreadLocal<Boolean> ignoreMetrics = new ThreadLocal<>();
+    static ThreadLocal<BooleanHolder> ignoreMetrics = new ThreadLocal<>();
 
     private final Iterable<Tag> tags;
 
@@ -123,12 +123,17 @@ public class LogbackMetrics implements MeterBinder, AutoCloseable {
      * runnable.
      */
     public static void ignoreMetrics(Runnable r) {
-        ignoreMetrics.set(true);
+        BooleanHolder flag = ignoreMetrics.get();
+        if (flag == null) {
+            flag = new BooleanHolder();
+            ignoreMetrics.set(flag);
+        }
+        flag.set(true);
         try {
             r.run();
         }
         finally {
-            ignoreMetrics.remove();
+            flag.set(false);
         }
     }
 
@@ -141,6 +146,17 @@ public class LogbackMetrics implements MeterBinder, AutoCloseable {
         }
     }
 
+    static class BooleanHolder {
+        private boolean flag;
+
+        public boolean get() {
+            return flag;
+        }
+
+        public void set(boolean flag) {
+            this.flag = flag;
+        }
+    }
 }
 
 @NonNullApi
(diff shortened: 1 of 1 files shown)


jonatan-ivanov:
Thank you for the PR and sorry that we did not get back to you sooner.

The leak that this change is causing might not be a big issue for apps that are (re)using platform threads but with [virtual threads](https://openjdk.org/jeps/425) this can be an issue: virtual threads can be created and dropped at a high rate so you can end-up with a lot of `BooleanHolder` instances. I tested this with JDK 22 and I was able to create a significant amount of leak (well over 1GiB) just by doing this (a lot) even without introducing a `BooleanHolder`:
```java
Thread.ofVirtual().start(() -> threadLocal.set(Boolean.TRUE));
```

I can see the following solutions:
- Logback implementing this natively (through JMX?) so we don't need the filter neither the thread-local (please create an issue for Logback if you think this would work for you)
- [Scoped Values](https://openjdk.org/jeps/464) can solve this issue but the feature is still in preview
- Making this somehow overridable, I'm not sure how though since `ignoreMetrics` is `static`

shakuzen:
> Because logger calls can be so frequent (even if they're suppressed this code is still called) the `TurboFilter` gets called _a lot_, and thus so does `ThreadLocal.remove`. Therefore Micrometer ends up slowing down all our code.

Since this pull request was originally opened we've addressed the part about the ThreadLocal being called when the log level is not enabled in https://github.com/micrometer-metrics/micrometer/issues/3891, so this issue should be at least somewhat mitigated.

I'm concerned about the following mentioned tradeoff this solution introduces:

> Of course the tradeoff is that now micrometer 'pollutes' all threads with an extraneous ThreadLocal value that it never cleans up.

So I'm a bit weary to adopt this solution as we have to think about how it may affect users with different run-time characteristics and usage in their applications.

The whole use of ThreadLocal to avoid potentially infinitely recursive calls in the case a Counter implementation's increment method logs is unfortunate because I think in practice it only came about because of the Statsd implementation. Though the general idea stands that at the level of logging instrumentation we can't know that an Counter implementation doesn't itself log.

jonatan-ivanov:
Duplicate of https://github.com/micrometer-metrics/micrometer/pull/6882

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/6882

jonatan-ivanov:
@spmason Even if we went to another direction, thank you very much for the PR and your efforts.

PR #3971, opened 2023-07-10
Deprecates observeWithContext and observeWithContextChecked

marcingrzejszczak:
these methods are will not work when an observation is filtered out and becomes noop. That's because the context type will no longer be C it will be Context

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -19,6 +19,8 @@ import io.micrometer.common.KeyValue;
 import io.micrometer.common.KeyValues;
 import io.micrometer.common.lang.NonNull;
 import io.micrometer.common.lang.Nullable;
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
Related issue https://github.com/spring-projects/spring-integration/issues/8664

ttddyy (approved):

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -48,6 +50,8 @@
  */
 public interface Observation extends ObservationView {
 
+    InternalLogger logger = InternalLoggerFactory.getInstance(Observation.class);
```
izeye, line 53: The `logger` is exposed as `public`. It seems to be an accidental leak.

marcingrzejszczak: Darn, it is, let me fix it

PR #3986, opened 2023-07-19
Removes the need to set OR on OTLA

marcingrzejszczak:
without this change we are required to set OR on OTLA so that when NullObservation gets created we will call the proper handlers to e.g. clear tracing values.

with this change we no longer require to set OR on OTLA because we will reuse the OR that was set on the Observation before clearing happens in OTLA via `setValue()`. When there was no Observation before calling `setValue()` we just skip creating any scopes. That way we are always reusing the ObservationRegistry that is used between Observations.

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/ObservationView.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/ObservationView.java
@@ -26,6 +26,15 @@ import io.micrometer.observation.Observation.ContextView;
 public interface ObservationView {
 
     /**
+     * Returns the {@link ObservationRegistry} attached to this observation.
+     * @return corresponding observation registry
+     * @since 1.10.10
+     */
+    default ObservationRegistry getObservationRegistry() {
+        return ObservationRegistry.NOOP;
+    }
+
+    /**
      * Returns the {@link ContextView} attached to this observation.
      * @return corresponding context
      */
--- a/micrometer-observation/src/test/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessorTests.java
+++ b/micrometer-observation/src/test/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessorTests.java
@@ -15,10 +15,14 @@
  */
 package io.micrometer.observation.contextpropagation;
 
+import io.micrometer.context.ContextAccessor;
 import io.micrometer.context.ContextExecutorService;
 import io.micrometer.context.ContextRegistry;
 import io.micrometer.context.ContextSnapshot;
+import io.micrometer.context.ContextSnapshotFactory;
+import io.micrometer.observation.NullObservation;
 import io.micrometer.observation.Observation;
+import io.micrometer.observation.Observation.Context;
 import io.micrometer.observation.ObservationHandler;
 import io.micrometer.observation.ObservationRegistry;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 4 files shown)


micrometer-observation/src/main/java/io/micrometer/observation/ObservationView.java
```diff
@@ -25,6 +25,15 @@
  */
 public interface ObservationView {
 
+    /**
+     * Returns the {@link ObservationRegistry} attached to this observation.
+     * @return corresponding observation registry
+     * @since 1.10.10
+     */
+    default ObservationRegistry getObservationRegistry() {
```
chemicL, line 33: Are we able to operate on the `ObservationRegistry.ObservationConfig` level instead? All the calls to the registry to obtain the current `Observation`/`Scope` can go to the static, global instance.

micrometer-observation/src/test/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessorTests.java
```diff
@@ -122,16 +150,69 @@ void capturedThreadLocalValuesShouldBeCapturedRestoredAndCleared()
...
         child.stop();
         parent.stop();
 
         then(child.getEnclosingScope()).isNull();
         then(parent.getEnclosingScope()).isNull();
     }
 
+    @Test
+    void uses2DifferentObservationRegistries() {
+        AtomicReference<String> reference = new AtomicReference<>("");
```
chemicL, line 172: rename to `currentHandler` perhaps?

micrometer-observation/src/test/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessorTests.java
```diff
@@ -15,34 +15,50 @@
...
     ContextRegistry registry = new ContextRegistry();
 
     @BeforeEach
     void setup() {
-        observationRegistry.observationConfig().observationHandler(new TracingHandler());
-        registry.registerThreadLocalAccessor(new ObservationThreadLocalAccessor(observationRegistry));
+        registry.registerThreadLocalAccessor(new ObservationThreadLocalAccessor());
+
+        ContextRegistry.getInstance().removeContextAccessor(mapContextAccessor);
+        ContextRegistry.getInstance().registerContextAccessor(mapContextAccessor);
```
chemicL, line 61: what's happening here? :)

chemicL (changes requested):
Please consider not coupling the `ObservationView` with the registry even further in the public API - perhaps we can evolve the usage in the future to have `ObservationRegistry` be simply global and static for `ThreadLocal` access, while the `ObservationConfig` is something that can vary from `Observation` to `Observation`.

micrometer-observation/src/main/java/io/micrometer/observation/ObservationView.java
marcingrzejszczak: They cannot because we need an `ObservationRegistry` to create an `Observation`, not an `ObservationConfig`

marcingrzejszczak:
> Please consider not coupling the ObservationView with the registry even further in the public API - perhaps we can evolve the usage in the future to have ObservationRegistry be simply global and static for ThreadLocal access, while the ObservationConfig is something that can vary from Observation to Observation.

We can create a separate issue for that. I don't see real value of doing such a change at this point

chemicL (approved):

PR #3989, opened 2023-07-20
Added ObservedValve for Tomcat

marcingrzejszczak:
to have that working we've added additional convention and contexts around javax and jakarta classes


bclozel:
Thanks Marcin for working on this!

I'm wondering about the instrumentation for the JAX RS client bits. This is using the `ClientRequestFilter` and `ClientResponseFilter` contracts for this. It seems that if the connection somehow fails, or if a filter fails, a `javax.ws.rs.client.ResponseProcessingException` is thrown directly and this exception does not go through the response filter. This means that observations would not be stopped in those cases and that the instrumentation would miss a bunch of cases: filter errors and I/O errors in general. This reminds me of the work we recently did on the HttpComponents client where we switched to a different approach. Is there a better way to instrument clients here. Maybe through the `*Invoker` contracts by wrapping? We would need to see some concrete usage with a client to see if that's feasible.

I'm also wondering about the same problems on the server side with the `ContainerRequestFilter` and `ContainerResponseFilter` contracts. There, it seems that any exception (thrown by a JAX-RS endpoint or a filter) will be handled by an `ExceptionMapper`. In this case we might not need a different instrumentation approach but we might need a consistent ordering of that filter in the chain. Should it be ordered so that all other filters had a chance to process the response so that we don't record the wrong information? How should that be enforced?

Now about the Tomcat Valve. I'm not super familiar with this contract, but I remember that the `asyncSupported` flag is off by default and that this can play a important part since requests might not be instrumented for the async I/O case. I think that for regular async support, you might not need to do something special for async dispatches (like in the Servlet filter case), but I'm not entirely sure and I think this should be checked.

In general, in the Spring Framework instrumentation I tried to provide a lot of flexibility with custom names for observations and custom conventions (by providing constructor variants), but I'm wondering if this doesn't clutter the API in the end. Maybe this would be better handled with other contracts like the observation filter? I think this really depends on the use case. Maybe here JAX RS implementations would like to have a choice to come up with their own custom name?

marcingrzejszczak:
> I'm wondering about the instrumentation for the JAX RS client bits. This is using the ClientRequestFilter and ClientResponseFilter contracts for this. It seems that if the connection somehow fails, or if a filter fails, a javax.ws.rs.client.ResponseProcessingException is thrown directly and this exception does not go through the response filter. This means that observations would not be stopped in those cases and that the instrumentation would miss a bunch of cases: filter errors and I/O errors in general. This reminds me of the work we recently did on the HttpComponents client where we switched to a different approach. Is there a better way to instrument clients here. Maybe through the *Invoker contracts by wrapping? We would need to see some concrete usage with a client to see if that's feasible.
>I'm also wondering about the same problems on the server side with the ContainerRequestFilter and ContainerResponseFilter contracts. There, it seems that any exception (thrown by a JAX-RS endpoint or a filter) will be handled by an ExceptionMapper. In this case we might not need a different instrumentation approach but we might need a consistent ordering of that filter in the chain. Should it be ordered so that all other filters had a chance to process the response so that we don't record the wrong information? How should that be enforced?

Yup I fully agree. I don't know if these are the right contracts to be used for proper wrapping. I think I'll need some help from people working with Jakarta API directly. I asked about it in the RestEASY tracker https://issues.redhat.com/browse/RESTEASY-3356?focusedId=22781362&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22781362 and still am waiting for some feedback.

> Now about the Tomcat Valve. I'm not super familiar with this contract, but I remember that the asyncSupported flag is off by default and that this can play a important part since requests might not be instrumented for the async I/O case. I think that for regular async support, you might not need to do something special for async dispatches (like in the Servlet filter case), but I'm not entirely sure and I think this should be checked.

We had it turned on for Sleuth for ages and it worked well :shrug: 

> In general, in the Spring Framework instrumentation I tried to provide a lot of flexibility with custom names for observations and custom conventions (by providing constructor variants), but I'm wondering if this doesn't clutter the API in the end. Maybe this would be better handled with other contracts like the observation filter? I think this really depends on the use case. Maybe here JAX RS implementations would like to have a choice to come up with their own custom name?

I see. I guess it would be good for someone using JAX RS to chime in and give us feedback. Do you know anyone we can mention here? :)

bclozel:
> > Now about the Tomcat Valve. I'm not super familiar with this contract, but I remember that the asyncSupported flag is off by default and that this can play a important part since requests might not be instrumented for the async I/O case. I think that for regular async support, you might not need to do something special for async dispatches (like in the Servlet filter case), but I'm not entirely sure and I think this should be checked.
> 
> We had it turned on for Sleuth for ages and it worked well 🤷

My bad, I missed that it was calling `setAsyncSupported(true)` already in the constructor.

> > In general, in the Spring Framework instrumentation I tried to provide a lot of flexibility with custom names for observations and custom conventions (by providing constructor variants), but I'm wondering if this doesn't clutter the API in the end. Maybe this would be better handled with other contracts like the observation filter? I think this really depends on the use case. Maybe here JAX RS implementations would like to have a choice to come up with their own custom name?
> 
> I see. I guess it would be good for someone using JAX RS to chime in and give us feedback. Do you know anyone we can mention here? :)

Not really unfortunately. I think [Helidon is integrating with tracing libraries](https://github.com/helidon-io/helidon/tree/helidon-3.x/tracing) directly. Maybe something there that we can take a look at?

marcingrzejszczak:
So I think we're using the proper APIs here. We just need to add support for error cases https://github.com/helidon-io/helidon/blob/helidon-3.x/tracing/jersey/src/main/java/io/helidon/tracing/jersey/AbstractTracingFilter.java#L125C9-L139

marcingrzejszczak:
However for the client side there's also the interceptor https://github.com/helidon-io/helidon/blob/helidon-3.x/tracing/jersey-client/src/main/java/io/helidon/tracing/jersey/client/ClientTracingInterceptor.java . I'll check it out

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/AbstractDefaultHttpClientRequestObservationConvention.java
```diff
@@ -0,0 +1,140 @@
...
+        else {
+            return METHOD_NONE;
+        }
+    }
+
+    private KeyValue status(@Nullable Integer statusCode) {
+        if (statusCode == null) {
+            return STATUS_CLIENT_ERROR;
+        }
+        return KeyValue.of(CommonLowCardinalityKeys.STATUS, String.valueOf(statusCode));
```
jonatan-ivanov, line 94: To make this consistent to the other methods:

```suggestion
        if (statusCode != null) {
            return KeyValue.of(CommonLowCardinalityKeys.STATUS, String.valueOf(statusCode));
        }
        return STATUS_CLIENT_ERROR;
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/AbstractDefaultHttpClientRequestObservationConvention.java
```diff
@@ -0,0 +1,140 @@
...
+        if (statusCode != null) {
+            return HttpOutcome.forStatus(false, statusCode);
+        }
+        return HttpOutcome.forStatus(false, statusCode);
+    }
+
+    protected KeyValues getHighCardinalityKeyValues(@Nullable URI uri, @Nullable String userAgent) {
+        // Make sure that KeyValues entries are already sorted by name for better
+        // performance
+        return KeyValues.of(requestUri(uri), userAgent(userAgent));
```
jonatan-ivanov, line 123: I think user-agent is a great idea and we should do it in Spring Framework too.
/cc @bclozel

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/AbstractDefaultHttpClientRequestObservationConvention.java
```diff
@@ -0,0 +1,140 @@
...
+    private static final KeyValue EXCEPTION_NONE = KeyValue.of(CommonLowCardinalityKeys.EXCEPTION, KeyValue.NONE_VALUE);
+
+    private static final KeyValue HTTP_URL_NONE = KeyValue.of(CommonHighCardinalityKeys.HTTP_URL, KeyValue.NONE_VALUE);
+
+    private static final KeyValue USER_AGENT_NONE = KeyValue.of(CommonHighCardinalityKeys.USER_AGENT_ORIGINAL,
+            KeyValue.NONE_VALUE);
+
+    protected String getContextualName(String lowercaseMethod) {
+        return "http " + lowercaseMethod;
+    }
```
jonatan-ivanov, line 61: Shouldn't we add the `url`/`uri` too?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/AbstractDefaultHttpServerRequestObservationConvention.java
```diff
@@ -0,0 +1,154 @@
...
+            }
+            return keyValues;
+        }
+        catch (Exception ex) {
+            return KeyValues.empty();
+        }
+    }
+
+    protected KeyValues getHighCardinalityKeyValues(String requestUri) {
+        return KeyValues.of(httpUrl(requestUri));
```
jonatan-ivanov, line 101: Shouldn't we add the user-agent too just like we did with the client?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/HttpObservationDocumentation.java
```diff
@@ -0,0 +1,297 @@
...
+            public String asString() {
+                return "url.scheme";
+            }
+        },
+
+        /**
+         * Name of HTTP request method or {@value KeyValue#NONE_VALUE} if the request was
+         * not received properly.
+         */
+        METHOD {
```
jonatan-ivanov, line 128: There is also a `HTTP_REQUEST_METHOD`. Do we need both?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/HttpObservationDocumentation.java
```diff
@@ -0,0 +1,297 @@
...
+            public String asString() {
+                return "method";
+            }
+
+        },
+
+        /**
+         * HTTP response raw status code, or {@code "UNKNOWN"} if no response was created.
+         */
+        STATUS {
```
jonatan-ivanov, line 139: There is also a `HTTP_RESPONSE_STATUS_CODE`. Do we need both?

micrometer-jakarta/build.gradle
```diff
@@ -0,0 +1,116 @@
...
+
+    // Aspects
+    optionalApi 'org.aspectj:aspectjweaver'
+
+    optionalApi 'org.eclipse.jetty:jetty-server'
+    // jakarta servlet
+    optionalApi 'jakarta.servlet:jakarta.servlet-api:5.+'
+    optionalApi 'org.eclipse.jetty:jetty-client:12.0.+'
+    optionalApi 'org.apache.tomcat.embed:tomcat-embed-core:10.1.+'
+    optionalApi 'org.glassfish.jersey.core:jersey-server:3.1.+'
```
jonatan-ivanov, line 39: Shouldn't we add everything that has a version here to `dependencies.gradle`?

dependencies.gradle
```diff
@@ -75,6 +75,9 @@ def VERSIONS = [
         'org.hsqldb:hsqldb:2.5.+', // 2.6.0 requires JDK 11.
         'org.jooq:jooq:3.14.+',
         'org.jsr107.ri:cache-ri-impl:1.0.0',
+        'org.jboss.resteasy:resteasy-client:6.2.4.Final',
+        'org.jboss.resteasy:resteasy-netty4:6.2.4.Final',
+        'javax.validation:validation-api:1.0.0.GA',
```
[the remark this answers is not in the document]
marcingrzejszczak: I prefer to manually set the version so that it will be easier to migrate to TOML.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/AbstractDefaultHttpClientRequestObservationConvention.java
marcingrzejszczak: not on the client side https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#name

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/HttpObservationDocumentation.java
marcingrzejszczak: Yes, this one is for backward compatibility (our existing instrumentations tag like that). The other one is following semantic conventions.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/http/HttpObservationDocumentation.java
marcingrzejszczak: Yes, this one is for backward compatibility (our existing instrumentations tag like that). The other one is following semantic conventions.

micrometer-jakarta/build.gradle
marcingrzejszczak: No because we have tomcat in version 8 for micrometer-core and 10 for jakarta. I will double check if some of those can't be moved

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/ObservedValve.java
```diff
@@ -0,0 +1,97 @@
...
+import io.micrometer.observation.ObservationRegistry;
+
+/**
+ * A {@link Valve} that creates {@link Observation}.
+ *
+ * @author Marcin Grzejszczak
+ * @since 1.12.0
+ * @see HttpObservationDocumentation
+ */
+public class ObservedValve extends ValveBase {
```
shakuzen, line 40: Some guidance in the JavaDoc on when a user would configure this instrumentation would be good. Users should avoid double instrumentation that can happen when configuring instrumentation at multiple layers in the same stack for HTTP request processing. For example, I would expect users should not configure this and servlet-level or jax-rs or Spring WebMvc instrumentation.

micrometer-jakarta/build.gradle
```diff
@@ -0,0 +1,116 @@
...
+}
+
+dependencies {
+    api project(":micrometer-core")
+    api project(":micrometer-commons")
+    api project(":micrometer-observation")
+    optionalApi 'jakarta.ws.rs:jakarta.ws.rs-api' // jakarta
+
+    // Aspects
+    optionalApi 'org.aspectj:aspectjweaver'
```
shakuzen, line 32: I didn't notice any AOP in this module. Is this used?

micrometer-jakarta/build.gradle
```diff
@@ -0,0 +1,116 @@
...
+
+    optionalApi 'org.eclipse.jetty:jetty-server'
```
shakuzen, line 34: I'm also not seeing where jetty server or client classes are used.

micrometer-jakarta/src/main/java/io/micrometer/jakarta/instrument/binder/http/ObservationAutoDiscoverable.java
```diff
@@ -0,0 +1,39 @@
...
+import jakarta.ws.rs.ConstrainedTo;
+import jakarta.ws.rs.RuntimeType;
+import jakarta.ws.rs.core.FeatureContext;
+import org.glassfish.jersey.internal.spi.AutoDiscoverable;
+
+/**
+ * Binds observation into Jersey.
+ */
+@ConstrainedTo(RuntimeType.CLIENT)
+public class ObservationAutoDiscoverable implements AutoDiscoverable {
```
shakuzen, line 27: I'm not familiar with how libraries provide these `AutoDiscoverable` implementations, but the JavaDoc mentions declaring the implementation in a file in `META-INF/services` named `org.glassfish.jersey.internal.spi.AutoDiscoverable`. Otherwise I guess this won't be auto-discovered, but as I said, I'm not familiar with this.

Looking at [the documentation](https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/deployment.html#deployment.autodiscoverable), it says this is internal, so I wonder if we should be providing this implementation at all.

> Auto discovery functionality is in Jersey supported by implementing an internal `AutoDiscoverable` Jersey SPI. This interface is not public at the moment, and is subject to change in the future, so be careful when trying to use it.

micrometer-jakarta/src/main/java/io/micrometer/jakarta/instrument/binder/http/ObservationHttpJakartaInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+import static io.micrometer.jakarta.instrument.binder.http.ObservationHttpJakartaClientFilter.OBSERVATION_SCOPE_PROPERTY;
+
+/**
+ * A post-invocation client interceptor. Will stop an observation started in
+ * {@link ObservationHttpJakartaClientFilter} in case of exceptions.
+ *
+ * @author Marcin Grzejszczak
+ * @since 1.12.0
+ */
+public class ObservationHttpJakartaInterceptor implements PostInvocationInterceptor {
```
shakuzen, line 34: This should be named more specifically to Jersey, since it implements a Jersey SPI.

micrometer-jakarta/build.gradle
```diff
@@ -0,0 +1,116 @@
...
+    optionalApi 'jakarta.servlet:jakarta.servlet-api'
+    optionalApi 'org.eclipse.jetty:jetty-client:12.0.+' // can't move to dependencies.gradle, we have 2 versions of this lib (check micrometer-core)
+    optionalApi 'org.apache.tomcat.embed:tomcat-embed-core:10.1.+' // can't move to dependencies.gradle, we have 2 versions of this lib (check micrometer-core)
+    optionalApi 'org.glassfish.jersey.core:jersey-server:3.1.+' // can't move to dependencies.gradle, we have 2 versions of this lib (check micrometer-core)
+    // jakarta JMS
+    optionalApi 'jakarta.jms:jakarta.jms-api'
+
+    // log monitoring
+    optionalApi 'ch.qos.logback:logback-classic'
+    optionalApi 'org.apache.logging.log4j:log4j-core'
```
shakuzen, line 45: There's no logging instrumentation in this module; these should be removed.

micrometer-jakarta/build.gradle
```diff
@@ -0,0 +1,116 @@
...
+
+    optionalApi project(':micrometer-observation')
+    optionalApi 'io.micrometer:context-propagation'
```
shakuzen, line 48: I don't see any context-propagation specific instrumentation. This should be removed. `micrometer-observation` is already declared as `api` so it can be removed from `optionalApi`.

micrometer-core/build.gradle
```diff
@@ -67,6 +67,8 @@ jar {
 dependencies {
     api project(":micrometer-commons")
     api project(":micrometer-observation")
+    optionalApi 'javax.ws.rs:javax.ws.rs-api' // javax
```
shakuzen, line 70: I'm confused why this is needed now but it wasn't prior to this pull request. I get a compilation error in `JerseyTags` when I remove this, but that class in micrometer-core is unchanged.

micrometer-jakarta/src/main/java/io/micrometer/jakarta/instrument/binder/http/ObservationHttpJakartaInterceptor.java
shakuzen: I suppose this would be a problem in an environment that is using a JAX-RS implementation not including Jersey. I don't know how much of a problem that is, but as far as I understand, there will be issues with the filter linked in the JavaDoc if there are exceptions and an interceptor like this isn't configured.

micrometer-jakarta/src/main/java/io/micrometer/jakarta/instrument/binder/http/HttpJakartaServerRequestObservationContext.java
```diff
@@ -0,0 +1,54 @@
...
+ * observations.
+ * <p>
+ * This context also extends {@link RequestReplyReceiverContext} for propagating tracing
+ * information during HTTP request processing.
+ *
+ * @author Brian Clozel
+ * @author Marcin Grzejszczak
+ * @since 1.12.0
+ */
+public class HttpJakartaServerRequestObservationContext
```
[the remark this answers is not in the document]
shakuzen: I've updated the naming of all related classes. The JavaDocs still need an update.

micrometer-jakarta/src/main/java/io/micrometer/jakarta/instrument/binder/http/ObservationHttpJakartaInterceptor.java
shakuzen: Ultimately, this is a blocker for providing JAX-RS client instrumentation. This was ran into by the OTel Java instrumentation as well in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5430. There's more information about the limitation in JAX-RS itself in https://github.com/jakartaee/rest/issues/684. So I don't know that we can provide JAX-RS client instrumentation without it being buggy. For now, we should not merge this code until/unless we can find an alternative solution.

shakuzen: @marcingrzejszczak The existing tests do not demonstrate the issue mentioned in the linked issues. Running the tests against a WireMock endpoint stubbed as follows demonstrates the issue:

```java
wmRuntimeInfo.getWireMock().register(WireMock.get("/connectionReset").willReturn(WireMock.aResponse().withFault(Fault.CONNECTION_RESET_BY_PEER)));
```

The observation will be started but never stopped in this case. The Jersey specific `ObservationJerseyClientInterceptor` doesn't have any effect in the `ObservationJaxRsHttpClientFilterTests` because RESTEasy is used. This demonstrates why I don't think we can offer JAX-RS client instrumentation: it leaks observations when there is no response from the server. Rather we can add Jersey client instrumentation to jersey-micrometer, where with the Jersey interceptor we can ensure no leaks. It's unfortunate we can't instrument at the JAX-RS level for the client side, but we don't want to put out instrumentation with a known leak.

<details>
<summary>Full test code and failure</summary>

Test method added to `ObservationJaxRsHttpClientFilterTests`
```java
@Test
void clientFilterShouldWorkWithJakartaHttpClientForExceptionsWithoutResponse(WireMockRuntimeInfo wmRuntimeInfo) {
    wmRuntimeInfo.getWireMock().register(WireMock.get("/connectionReset").willReturn(WireMock.aResponse().withFault(Fault.CONNECTION_RESET_BY_PEER)));

    try (Client client = ClientBuilder.newClient()) {
        client.register(new ObservationJaxRsHttpClientFilter(observationRegistry, null));
        final WebTarget target = client
            .target("http://localhost:" + wmRuntimeInfo.getHttpPort() + "/connectionReset");
        Exception exception = null;
        try (Response response = target.request().get()) {
            BDDAssertions.fail("Response should not be returned");
        } catch (Exception e) {
            exception = e;
        }
        BDDAssertions.then(exception).isNotNull().isInstanceOf(ProcessingException.class);
    }

    wmRuntimeInfo.getWireMock()
        .verifyThat(WireMock.getRequestedFor(WireMock.urlEqualTo("/connectionReset"))
            .withHeader("foo", WireMock.equalTo("bar")));

    TestObservationRegistryAssert.then(observationRegistry)
        .hasSingleObservationThat()
        .hasBeenStarted()
        .hasBeenStopped()
        .hasError();
}
```

Test outcome
```
java.lang.AssertionError: Observation is not stopped

	at io.micrometer.jakarta9.instrument.binder.http.jaxrs.client.ObservationJaxRsHttpClientFilterTests.clientFilterShouldWorkWithJakartaHttpClientForExceptionsWithoutResponses(ObservationJaxRsHttpClientFilterTests.java:112)
	at java.base/java.lang.reflect.Method.invoke(Method.java:578)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

</details>

mkjensen:
Hi,

What is the status of this?

It looks pretty interesting to us. We have a product using Spring Boot 3.2 consisting of both legacy code using RESTEasy (via [RESTEasy Spring Boot Starter](https://github.com/resteasy/resteasy-spring-boot)) and newer code using Spring MVC.

It would be nice if we could easily get the same metrics for both, specifically the `http.server.requests` metric.

Thanks.

thomas-fra:
> Hi,
> 
> What is the status of this?
> 
> It looks pretty interesting to us. We have a product using Spring Boot 3.2 consisting of both legacy code using RESTEasy (via [RESTEasy Spring Boot Starter](https://github.com/resteasy/resteasy-spring-boot)) and newer code using Spring Web MVC.
> 
> It would be nice if we could easily get the same metrics for both, specifically the `http.server.requests` metric.
> 
> Thanks.

+1 
We are migrating from TomEE to Wildfly and so from cxf that come with metrics providers to RestEasy that have nothing :(
@marcingrzejszczak ?

PR #3996, opened 2023-07-25
Add exemplar support to _count

jonatan-ivanov:
Prometheus did not support exemplars for Summaries and it did not support exemplars for anything else than buckets on Histograms. Since Prometheus now supports exemplars on every time series, we can add exemplars to Summary/Histogram _count, see: https://github.com/prometheus/prometheus/issues/11982

Also related: https://github.com/micrometer-metrics/micrometer/pull/3460

Since the support for this [was added](https://github.com/prometheus/prometheus/pull/11984) in Prometheus sever in [`2.43.0`](https://github.com/prometheus/prometheus/releases/tag/v2.43.0) and the previous behavior was dropping the whole scrape response, this change (as-is in this draft) will break users who are using older Prometheus servers ([`2.43.0` was released on 2023-03-21](https://github.com/prometheus/prometheus/releases/tag/v2.43.0)).
We should figure out how to handle this based on the [Prometheus support terms](https://prometheus.io/docs/introduction/release-cycle/): it seems the oldest version that is still supported by Prometheus maintainers but does not have this functionality will reach EOL by 2023-07-31 (in a ~week).

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusDistributionSummary.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusDistributionSummary.java
@@ -15,14 +15,17 @@
  */
 package io.micrometer.prometheus;
 
+import io.micrometer.common.lang.NonNull;
 import io.micrometer.common.lang.Nullable;
 import io.micrometer.core.instrument.AbstractDistributionSummary;
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.distribution.*;
+import io.prometheus.client.exemplars.CounterExemplarSampler;
 import io.prometheus.client.exemplars.Exemplar;
-import io.prometheus.client.exemplars.HistogramExemplarSampler;
+import io.prometheus.client.exemplars.ExemplarSampler;
 
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.atomic.DoubleAdder;
 import java.util.concurrent.atomic.LongAdder;
 
@@ -47,10 +50,16 @@ public class PrometheusDistributionSummary extends AbstractDistributionSummary {
     @Nullable
     private final Histogram histogram;
 
-    private boolean exemplarsEnabled = false;
+    @Nullable
+    private final ExemplarSampler exemplarSampler;
+
+    @Nullable
+    private final AtomicReference<Exemplar> lastExemplar;
+
+    private boolean histogramExemplarsEnabled = false;
 
     PrometheusDistributionSummary(Id id, Clock clock, DistributionStatisticConfig distributionStatisticConfig,
-            double scale, HistogramFlavor histogramFlavor, @Nullable HistogramExemplarSampler exemplarSampler) {
+            double scale, HistogramFlavor histogramFlavor, @Nullable ExemplarSampler exemplarSampler) {
         super(id, clock,
                 DistributionStatisticConfig.builder()
                     .percentilesHistogram(false)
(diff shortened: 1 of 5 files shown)


jonatan-ivanov:
@HaloFour If you are interested and willing to spend the time on this, would you mind reviewing and/or testing it?

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusDistributionSummary.java
jonatan-ivanov, line 1: Both here and in `PrometheusTimer`, the last sampled exemplars can be recorded in two ways:
1. `PrometheusHistogram` records it so that the exemplar will not be sampled twice (one `sample` call in `PrometheusHistogram`): `Histogram` use-case (there are buckets + `_count`)
2. If that is not available, `PrometheusDistributionSummary`/`PrometheusTimer` record it directly: `Summary` use-case (no buckets, just `_count`)

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusHistogram.java
```diff
@@ -107,16 +113,20 @@ private void updateExemplar(double value, @Nullable TimeUnit sourceUnit, @Nullab
...
+        Exemplar nextExemplar;
 
         double exemplarValue = (sourceUnit != null && destinationUnit != null)
                 ? TimeUtils.convert(value, sourceUnit, destinationUnit) : value;
         do {
-            prev = exemplars.get(index);
-            next = exemplarSampler.sample(exemplarValue, bucketFrom, bucketTo, prev);
+            previusBucketExemplar = exemplars.get(index);
+            previousLastExemplar = lastExemplar.get();
+            nextExemplar = exemplarSampler.sample(exemplarValue, bucketFrom, bucketTo, previusBucketExemplar);
```
jonatan-ivanov, line 125: Please notice that here and both in `PrometheusDistributionSummary`/`PrometheusTimer` the value of the last exemplar is the recorded value. I think this is expected, but the value for `_count` should not be the value that was recorded but (in these cases) `1.0`.
Because of this the value is set in the registry (see my comment on it).

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java
```diff
@@ -295,8 +296,14 @@ public DistributionSummary newDistributionSummary(Meter.Id id,
...
 
-                samples.add(
-                        new Collector.MetricFamilySamples.Sample(conventionName + "_count", tagKeys, tagValues, count));
+                if (lastExemplar == null) {
+                    samples.add(new Collector.MetricFamilySamples.Sample(conventionName + "_count", tagKeys, tagValues,
+                            count));
+                }
+                else {
+                    samples.add(new Collector.MetricFamilySamples.Sample(conventionName + "_count", tagKeys, tagValues,
+                            count, copyExemplarWithNewValue(1.0, lastExemplar)));
```
jonatan-ivanov, line 305: Here I'm setting a new value for the `Exemplar`. This is needed because the value of the last recorded Exemplar is the last recorded (and sampled) value by `PrometheusDistributionSummary`/`PrometheusTimer`. That conforms the "contract" for being the value of the `.lastExemplar()` (which can be used elsewhere, i.e. in `_sum`) but for `_count` the value should be `1.0`.

HaloFour:
Looks simple enough, and the scraped metrics look good.  Are you thinking of adding the exemplar to `_sum` also or is that not really necessary?

I'm trying to figure out which version of Prometheus we're using and whether I can test the updated format with it.

jonatan-ivanov:
>Are you thinking of adding the exemplar to _sum also or is that not really necessary?

I'm not sure how would that be used. Do you do a `rate` on the `_sum` or aggregate it somehow where you don't use `_count` too? It's a very simple change but I would only add it if there is a user need for it.

shakuzen:
> it seems the oldest version that is still supported by Prometheus maintainers but does not have this functionality will reach EOL by 2023-07-31 (in a ~week).

Since it is now EOL, I think we can proceed with shipping this feature in our upcoming minor release and see what feedback we get from users. At least for now, I would say it seems reasonable that we don't target versions of Prometheus that are not supported by the project itself.

shakuzen (approved):

PR #4003, opened 2023-07-27
remove metrics from MeterRegistry when shutting down InstrumentedQueuedThreadPool

### LINKED ISSUE #4000: Remove metrics when shutting down InstrumentedQueuedThreadPool
remove metrics from MeterRegistry when shutting down InstrumentedQueuedThreadPool
Something like:

```java
    @Override
    protected void doStop() throws Exception {
       Set<Tag> tagSet = new HashSet<>();
        tags.forEach(tagSet::add);
        Stream.of("jetty.threads.config.min",
                        "jetty.threads.config.max",
                        "jetty.threads.busy",
                        "jetty.threads.jobs",
                        "jetty.threads.current",
                        "jetty.threads.idle")
                .flatMap(name -> meterRegistry.find(name).meters().stream())
                .map(Meter::getId)
                .filter(id -> new HashSet<>(id.getTags()).containsAll(tagSet))
                .forEach(meterRegistry::remove);
        super.doStop();
    }
```

**Rationale**
Clean up resources when not needed anymore

**Additional context**
The dropwizard metrics  jetty InstrumentedQueuedThreadPool does something [similar](https://github.com/dropwizard/metrics/blob/release/4.2.x/metrics-jetty11/src/main/java/io/dropwizard/metrics/jetty11/InstrumentedQueuedThreadPool.java#L103), and it seems like a good idea to do here as well.

I am happy to open a PR for this with tests if it is agreed that this should exist.

Thanks

wakingrufus:
remove metrics from MeterRegistry when shutting down  JettyServerThreadPoolMetrics/InstrumentedQueuedThreadPool

fixes #4000

### Rationale
Clean up resources when not needed anymore

### Additional context
The dropwizard metrics jetty InstrumentedQueuedThreadPool does something [similar](https://github.com/dropwizard/metrics/blob/release/4.2.x/metrics-jetty11/src/main/java/io/dropwizard/metrics/jetty11/InstrumentedQueuedThreadPool.java#L103), and it seems like a good idea to do here as well.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPool.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 VMware, Inc.
+ * Copyright 2024 VMware, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
  */
 package io.micrometer.core.instrument.binder.jetty;
 
+import io.micrometer.common.lang.Nullable;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
@@ -38,6 +39,9 @@ public class InstrumentedQueuedThreadPool extends QueuedThreadPool {
     private final MeterRegistry registry;
 
     private final Iterable<Tag> tags;
+
+    @Nullable
+    private JettyServerThreadPoolMetrics threadPoolMetrics;
 
     /**
      * Default values for the instrumented thread pool.
@@ -112,8 +116,16 @@ public class InstrumentedQueuedThreadPool extends QueuedThreadPool {
     @Override
     protected void doStart() throws Exception {
         super.doStart();
-        JettyServerThreadPoolMetrics threadPoolMetrics = new JettyServerThreadPoolMetrics(this, tags);
+        threadPoolMetrics = new JettyServerThreadPoolMetrics(this, tags);
         threadPoolMetrics.bindTo(registry);
+    }
+
+    @Override
+    protected void doStop() throws Exception {
+        if (threadPoolMetrics != null) {
+            threadPoolMetrics.close();
+        }
+        super.doStop();
     }
 
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 VMware, Inc.
+ * Copyright 2024 VMware, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,13 +15,18 @@
  */
 package io.micrometer.core.instrument.binder.jetty;
 
+import io.micrometer.common.lang.Nullable;
 import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.MeterBinder;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.eclipse.jetty.util.thread.ThreadPool;
 import org.eclipse.jetty.util.thread.ThreadPool.SizedThreadPool;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * {@link MeterBinder} for Jetty {@link ThreadPool}.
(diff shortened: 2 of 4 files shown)


wakingrufus:
Hi @shakuzen would you mind taking a look at this PR? Thanks

wakingrufus:
@jonatan-ivanov I would really appreciate if someone could take a look at this PR. Thanks!
Sorry for the direct ping, but if I am going to ping someone random, I figured it might as well be someone who I have played duck bowling with ;)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPool.java
```diff
@@ -116,4 +120,18 @@ protected void doStart() throws Exception {
         threadPoolMetrics.bindTo(registry);
     }
 
+    @Override
+    protected void doStop() throws Exception {
+        Set<Tag> tagSet = new HashSet<>();
+        tags.forEach(tagSet::add);
+        Stream
+            .of("jetty.threads.config.min", "jetty.threads.config.max", "jetty.threads.busy", "jetty.threads.jobs",
+                    "jetty.threads.current", "jetty.threads.idle")
```
shakuzen, line 129: I worry about keeping this list in sync with changes in `JettyServerThreadPoolMetrics`. It would be better to keep track of the metrics in `JettyServerThreadPoolMetrics` and make it `AutoCloseable`. Then `doStop` here just needs to close its instance of `JettyServerThreadPoolMetrics`. You can see an example of where we have done this in `KafkaMetrics`.

shakuzen (changes requested):
Sorry for taking so long to review, and thanks for opening the pull request.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
```diff
@@ -38,52 +44,81 @@
...
             .register(registry);
     }
 
+    @Override
+    public void close() throws Exception {
+        if (registry != null) {
+            Set<Tag> tagSet = new HashSet<>();
+            tags.forEach(tagSet::add);
+            Stream.of(MIN, MAX, BUSY, JOBS, CURRENT, IDLE)
+                .flatMap(name -> registry.find(name).meters().stream())
```
shakuzen, line 117: This can fail to find the meters by name depending on the meter filters configured. That's why keeping a reference to the meters registered (or their `Id`) for directly removing them is the safe way to do this. See the implementation in `KafkaMetrics` for an example.

shakuzen (changes requested):
Thank you for quickly updating the pull request. I've left another review.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
wakingrufus: Ah I see what you mean now. Thanks again, I will update this

wakingrufus: Ok, it is updated. Please take another look. Thanks again!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
```diff
@@ -41,49 +46,80 @@
...
             .description("The number of idle threads in the pool")
             .tags(tags)
             .register(registry);
+        registeredMeterIds.add(idleGauge.getId());
+    }
+
+    @Override
+    public void close() throws Exception {
+        if (registry != null) {
+            registeredMeterIds.forEach(registry::remove);
```
jonatan-ivanov, line 121: If we think that things will be reused and these meters need to be removed from the registry (reading the issue I'm still not sure or I missed something) we should definitely empty `registeredMeterIds` too.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPoolTest.java
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright 2019 VMware, Inc.
```
jonatan-ivanov, line 2: Should be 2024.
Could you please also update the year in other files too?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
shakuzen: I don't think it would be expected for an instance of `JettyServerThreadPoolMetrics` to be reused, though I suppose it is theoretically possible. In `InstrumentedQueuedThreadPool`, it will create a new instance in `doStart`.

wakingrufus: I agree that an instance of JettyServerThreadPoolMetrics should not be reused, but emptying `registeredMeterIds` seems like a good idea just to avoid potential footguns. I'll make that change

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPoolTest.java
wakingrufus: Yes. is `VMware, Inc.` still the correct copyright holder?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPoolTest.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Collections;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class InstrumentedQueuedThreadPoolTest {
+
+    @Test
+    void test_metrics() throws Exception {
```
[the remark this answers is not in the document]
wakingrufus: Done. Thanks for the feedback!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
wakingrufus: This is done

wakingrufus:
I have made all requested changes. Please take another look.
Thanks again!

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/InstrumentedQueuedThreadPoolTest.java
shakuzen: Yes, thanks.

shakuzen (approved):

shakuzen:
Thank you for the pull request and patience and all the quick turnarounds on feedback.

PR #4006, opened 2023-08-01
[Dynatrace v2] Export unit/description to Dynatrace

### LINKED ISSUE #3979: Dynatrace Registry: Export units and descriptions
**Please describe the feature request**
The Micrometer API allows for defining a `description` and `baseUnit` upon Meter creation.
The Dynatrace metrics ingestion API used by the Dynatrace Micrometer registry allows for specifying units and descriptions for ingested metrics: https://www.dynatrace.com/support/help/shortlink/metric-ingestion-protocol#metadata

It would be great if the Dynatrace registry would make full use of both APIs and also export the unit and description for each metric if they are specified by the user.

**Rationale**
The Meter description can describe a metric in more detail than just the metric name itself.
The Meter unit can help when interpreting metrics and is useful to tell whether different metrics are comparable with one another.

**Additional context**
https://www.dynatrace.com/support/help/shortlink/metric-ingestion-protocol#metadata
https://www.dynatrace.com/support/help/shortlink/custom-metric-metadata

cc @pirgeo

pirgeo:
resolves #3979

Metadata (unit and description) can be sent to the Dynatrace metrics API in a special format. This addition will automatically create these lines if unit and/or description are set on the respective Micrometer meter. The feature is turned on by default and can be turned off by toggling the `exportMeterMetadata` switch.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -158,7 +158,7 @@ public interface DynatraceConfig extends StepRegistryConfig {
         if (apiVersion() == V1) {
             return false;
         }
-        return getBoolean(this, "exportDynatraceMetadata").orElse(true);
+        return getBoolean(this, "exportMeterMetadata").orElse(true);
     }
 
     @Override
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -125,6 +125,9 @@ public interface DynatraceConfig extends StepRegistryConfig {
      * @since 1.8.0
      */
     default boolean enrichWithDynatraceMetadata() {
+        if (apiVersion() == V1) {
+            return false;
+        }
         return getBoolean(this, "enrichWithDynatraceMetadata").orElse(true);
     }
 
@@ -141,6 +144,21 @@ public interface DynatraceConfig extends StepRegistryConfig {
             return false;
         }
         return getBoolean(this, "useDynatraceSummaryInstruments").orElse(true);
+    }
+
+    /**
+     * Toggle whether to export metadata (unit and description) to the Dynatrace backend
+     * for the V2 version of this exporter. Metadata will be exported by default from
+     * Micrometer version 1.12.0. This setting has no effect for the (legacy) Dynatrace
+     * Exporter v1
+     * @return true if metadata should be exported, false otherwise.
+     * @since 1.12.0
+     */
+    default boolean exportMeterMetadata() {
+        if (apiVersion() == V1) {
+            return false;
+        }
+        return getBoolean(this, "exportDynatraceMetadata").orElse(true);
     }
 
     @Override
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
@@ -120,7 +120,9 @@ public interface DynatraceConfig extends StepRegistryConfig {
     }
 
     /**
-     * Return whether to enrich with Dynatrace metadata.
+     * Return whether to enrich with Dynatrace metadata. Dynatrace metadata is provided by
+     * the Dynatrace OneAgent or Dynatrace Kubernetes Operator and helps put metrics
+     * emitted via the Micrometer exporter in context more easily.
      * @return whether to enrich with Dynatrace metadata
      * @since 1.8.0
      */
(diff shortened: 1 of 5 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -133,40 +133,63 @@ private DimensionList parseDefaultDimensions(Map<String, String> defaultDimensio
...
         if (!batch.isEmpty()) {
             send(batch);
         }
     }
 
-    private Stream<String> toMetricLines(Meter meter) {
-        return meter.match(this::toGaugeLine, this::toCounterLine, this::toTimerLine, this::toDistributionSummaryLine,
-                this::toLongTaskTimerLine, this::toTimeGaugeLine, this::toFunctionCounterLine,
-                this::toFunctionTimerLine, this::toMeterLine);
+    private void trySendBatch(List<String> batch, int partitionSize) {
```
arminru, line 173: The `try` wording had me think of an operation that's likely to fail or put in a try/catch block, which is not the case here.
Maybe `sendBatchIfFull` or `sendBatchIfPartitionFull` might be a clearer name but either should work.

arminru (approved):

pirgeo:
I rebased onto main, but it seems I am now conflicting with the dependency lock state. Should I rebase onto an older version or can I fix this some other way?

shakuzen:
> I rebased onto main, but it seems I am now conflicting with the dependency lock state. Should I rebase onto an older version or can I fix this some other way?

Did you rebase onto the upstream `main` or on a stale version of `main` in your fork? I can't reproduce the issue seen with the latest `main` in this repo (upstream). I couldn't re-run the job either since it is running on CircleCI for your fork.

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
```diff
@@ -143,6 +146,21 @@ default boolean useDynatraceSummaryInstruments() {
...
 
+    /**
+     * Toggle whether to export metadata (unit and description) to the Dynatrace backend
+     * for the V2 version of this exporter. Metadata will be exported by default from
+     * Micrometer version 1.12.0. This setting has no effect for the (legacy) Dynatrace
+     * Exporter v1
+     * @return true if metadata should be exported, false otherwise.
+     * @since 1.12.0
+     */
+    default boolean exportMeterMetadata() {
```
shakuzen, line 157: I don't know exactly what the `enrichWithDynatraceMetadata()` controls, but I wonder if users will be confused by this `exportMeterMetadata()` co-existing with it. What's the motivation to have a configuration method at all for this? Is there a concern of performance issues for users with many meters?

pirgeo: `enrichWithDynatraceMetadata()` retrieves Dynatrace-specific metadata from the OneAgent, if it's available. This helps Dynatrace put the Micrometer metrics in the right context in our UI.
`exportMeterMetadata()` controls whether the metadata set on the meters in Micrometer should be exported at all. There is a performance overhead, and since metadata does not change often, users might want to set it once and then turn off this feature. Dynatrace will remember the metadata you set before. We have seen that turning on this feature can increase the volume of exported data significantly, and want to give users (especially ones that export huge volumes of data) an out. Egress can get quite expensive in certain situations, and exporting redundant data does not help that.

I have thought about other ways to tackle this, but those approaches usually have the downside of having to keep state between exports and would be a much bigger refactoring. I might tackle that at a later point in time. We have users asking for the possibility of exporting metadata, and we want to make sure that users who do not want to have the overhead can opt out.

Do you think adding more documentation on these methods would make it more clear? I will also update the Micrometer docs when this feature is added.

shakuzen (approved):

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceConfig.java
pirgeo: I added some more context to the Javadocs. I also pointed out that setting that toggle to false is equivalent to configuring a MeterFilter. I think there is value in having it be easy to toggle with a Spring Boot configuration, especially since it might be a somewhat resource-heavy feature.

pirgeo (commented):
I went on an (extended) wild goose chase to find out why the CI was failing on the Dynatrace fork. The gist of it is: The CI builds (and even local builds for that matter) rely on the git tags published on the Micrometer repo. In our CI (and when I just checked out the forked repo locally), the last tag was 1.7.0-M1. For that reason, all our builds tried to build against 1.7.0-SNAPSHOT:
```
Inferred project: micrometer, version: 1.7.0-SNAPSHOT
```
Some time in the last few weeks, a lockfile upgrade led to an incompatibility with 1.7.0:
```
> Could not resolve all files for configuration ':micrometer-test:java11TestCompileClasspath'.
   > Resolved 'org.hdrhistogram:HdrHistogram:2.1.12' which is not part of the dependency lock state
   > Resolved 'io.micrometer:micrometer-core:1.9.5' which is not part of the dependency lock state
```
 Copying over the tag for 1.12.0-M2 into our fork fixed all the problems. Now we build against 1.12.0-SNAPSHOT again, and all is right with the world. Don't ask me how much time I spent looking at the two git repos and not understaning why one main branch builds and the other one doesn't when they are at the exact same commit 😅

shakuzen:
> Don't ask me how much time I spent looking at the two git repos and not understaning why one main branch builds and the other one doesn't when they are at the exact same commit 😅

Sorry you had to go through that, but thank you for getting to the bottom of it, resolving it, and sharing the cause with us. I'll try to keep that in mind should anyone run into a similar issue in the future.

PR #4028, opened 2023-08-11
Allow specifying the meterNameConsumer for HighCardinalityTagsDetector

sean-heller:
Currently, it's not possible to specify the `meterNameConsumer`, so folks are just left with the default functionality of logging a warning then logging debug messages. 

This change would allow for users to specify their own consumer - one use case for this would be to add a new metric that is marked every time there is a metric with high cardinality detected, which could be used for alerting (vs. just looking through logs).

Let me know what you think! Thanks!

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/HighCardinalityTagsDetectorTests.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/HighCardinalityTagsDetectorTests.java
@@ -90,6 +90,21 @@ class HighCardinalityTagsDetectorTests {
         assertThat(highCardinalityTagsDetector.findFirst()).isEmpty();
     }
 
+    @Test
+    void addsCustomMeterNameConsumer() {
+        TestCustomMeterNameConsumer customMeterNameConsumer = new TestCustomMeterNameConsumer();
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/HighCardinalityTagsDetectorTests.java
```diff
@@ -107,4 +122,21 @@ public String getName() {
 
     }
 
+    private static class TestCustomMeterNameConsumer implements Consumer<String> {
+
+        @Nullable
```
alexey-vasin, line 127: Why do you need such annotations in test code? Seems like an overkill to me.

alexey-vasin (approved):

micrometer-core/src/test/java/io/micrometer/core/instrument/HighCardinalityTagsDetectorTests.java
jonatan-ivanov: It's a good practice even in tests, also, it will eliminate IDE complaints/warnings.

shakuzen (approved):

Issue #4032, opened 2023-08-16
Add a timestamp to Event in Observation


marcingrzejszczak:
sometimes we want to pass a timestamp when an event happened. Currently we don't allow that. Example - https://github.com/line/armeria/pull/4980#discussion_r1252703954

jonatan-ivanov:
Is the use-case getting to know about an event that happened in the past and you know when did the event happen?
Right now we only know the timestamp of current events in the present (when `onEvent` is called; similarly to `onStart`/`onStop`).

I think getting to know about past event can be a fair use-case, my two cents:
- Back in the day, the first  PoC of the ~~Recording~~ Observation API had non-instantaneous (`Observation` as of today) and instantaneous (`Event` as of today) events, both of them hold the timestamp to tell when they happened (was "more stateful" compared to what we have today), I think this is the same what this issue is for
- We removed this capability and made it "less stateful" so that it can be more lightweight but one of the the trade-offs was this
- I think it can be ok to introduce a timestamp on `Event` if there's need for it but if we do it, it will be a little different than `Observation` since that one does not have a timestamp and both suppose to record that something happened
- Also the timestamp will be there even if it is not needed
- As an alternative, I'm also thinking about introducing an `Event` implementation that can have a timestamp (`PastEvent`, `EventWithTimeStamp` ? :|) but adding a timestamp to the event can be also ok

marcingrzejszczak:
The example is available here https://github.com/line/armeria/pull/4980#discussion_r1252703954 . So in Armeria they are gathering timing information about certain events happening in the system. Then they can pass it to annotate e.g. the spans. In case of an observation api we can't really use that information because the event ignores it (and spans ignore it too)

PR #4034, opened 2023-08-17
Add timestamp (wall time) information on the Event 

marcingrzejszczak:
fixes gh-4032

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -1271,22 +1271,19 @@ public interface Observation extends ObservationView {
          * @return event
          */
         static Event of(String name, String contextualName) {
-            return new Event() {
-                @Override
-                public String getName() {
-                    return name;
-                }
+            return new SimpleEvent(name, contextualName);
+        }
 
-                @Override
-                public String getContextualName() {
-                    return contextualName;
-                }
-
-                @Override
-                public String toString() {
-                    return "event.name='" + getName() + "', event.contextualName='" + getContextualName() + '\'';
-                }
-            };
+        /**
+         * Creates an {@link Event} for the given names and timestamp (wall time).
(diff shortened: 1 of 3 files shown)


micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -1282,9 +1285,62 @@ public String getContextualName() {
...
                 }
 
+                @Override
+                public long getTimestamp() {
+                    return this.timestamp;
+                }
+
+                @Override
+                public TimeUnit getTimestampTimeUnit() {
+                    return TimeUnit.MILLISECONDS;
```
jonatan-ivanov, line 1295: Is this needed? Can we just say in the javadocs that it is the current wall time in ms?

jonatan-ivanov (commented):
I'm not sure I understand the use-case (I made a [comment on the issue](https://github.com/micrometer-metrics/micrometer/issues/4032#issuecomment-1682748138)).

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
marcingrzejszczak: That's what I did originally. I even named the method `getTimestampInMs` or sth like that. With 2 methods there is no doubt about what the units are and you are able to modify it as you please. Without that you can make easy mistakes (like I did when I've forgotten that timestamp in Brave is micro not milliseconds :shrug: ). WDYT @shakuzen ?

jonatan-ivanov:
The things I changed in the PR:
- Rebased it on `main`.
- Moved the anonymous `Event` classes out to a `SimpleEvent` class to avoid code duplication (and anonymous classes).
- I consolidated the interface changes into a single `getWallTime` method. Please let me know what you think about this, we can always add a `getWallTimeUnit` as an extra if it will be confusing (tried to document it to make it more clear).

PR #4038, opened 2023-08-19
Add xxxKeyValue() variants with KeyValue to ObservationContextAssert

izeye:
While working on #4037, `xxxKeyValue()` variants with `KeyValue` in `ObservationContextAssert` seem to be useful, so this PR adds them.


marcingrzejszczak (approved):

marcingrzejszczak:
Great idea, I'm really surprised that we've forgotten to add it earlier. Thanks!

PR #4044, opened 2023-08-24
remove total suffix for counter in prometheus

### LINKED ISSUE #2881: Remove suffix _total for counters in PrometheusNamingConvention
There is no need of adding _total suffix in counter metric when using prometheus registry in PrometheusNamingConvention class.

Since version 0.10 of prometheus-simpleclient they are adding the suffix _total in counter metrics.
This is done in [Collectors](https://github.com/micrometer-metrics/micrometer/blob/8e7f66670cbf33189f9aa3af99943444c5f2643a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java#L65).

cuichenli:
fix #2881


jonatan-ivanov:
Hey @cuichenli,
Could you please help us a little to understand this better and answer the questions here: https://github.com/micrometer-metrics/micrometer/issues/2881#issuecomment-1754394345

I think it's also worth noticing that this is a breaking change.

Issue #4047, opened 2023-08-26
Add Prefix Support for Metric Names in CacheMeterBinder


fartzy:
**Please describe the feature request.**
I would like to modify the [CacheMeterBinder](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java) abstract class to support prefixing metric names. As of now, the metric names are [hard coded ](https://github.com/micrometer-metrics/micrometer/blob/a56b968ba5b3db9b5e4a4feac813080783a16f5f/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java#L78) when the metrics are created in the `public final void bindTo(MeterRegistry registry)` method and then registered. Specifically, I propose adding a method `getMetricNamePrefix()` which returns a prefix string that can be added to the front of all metric names in - which will be in modifying the `bindTo()` method. This allows for better customization and adaptability in various system settings. 

**Rationale**
Without the proposed feature, it is impossible to rename metrics without deleting the existing cache metrics in the `bindImplementationSpecificMetrics()` method and re-registering them with new, prefixed names. This capability is particularly useful for specialized systems like [ActiveMQ-Artemis](https://issues.apache.org/jira/projects/ARTEMIS/issues/ARTEMIS-4306), which adhere to unique metric naming conventions. For instance, these cache metrics could be applied to the [authorization and authentication caches](https://issues.apache.org/jira/projects/ARTEMIS/issues/ARTEMIS-4306) utilized for access control within the message broker. Deleting, recreating, and re-registering metrics is an inefficient way to align metric names. As such, the ability to add prefixes to metrics is essential for the seamless integration of Micrometer's cache metrics with ActiveMQ-Artemis or any other systems that prefer strict naming conventions for metrics. This feature would eliminate the need to delete and re-register cache metrics, thereby simplifying cache metric management.

**Additional context**
**Backwards Compatibility:** This change is backward compatible because the default prefix is an empty string. This ensures that, by default, the existing metric names will not change, effectively appending an empty string to the current metric name.
**Test Coverage:** I have also added a `CacheMetricsBehaviorTest` class  to verify this new functionality.
**Customization:** This feature enhances Micrometer's flexibility by allowing users to define custom metric name prefixes for cache metrics which can be particularly useful in enterprise or specialized settings.

jonatan-ivanov:
[...]

PR #4048, opened 2023-08-26
Add Metric Name Prefixing in CacheMeterBinder

fartzy:
**Summary**
This PR aims to fix issue #4047.  This allows the ability to add prefixes to metric names in the `CacheMeterBinder` class, making it more customizable and adaptable to systems with specific metric naming conventions (e.g., ActiveMQ-Artemis). This change is backward-compatible.

**Changes**
Introduced `getMetricNamePrefix()` method in `CacheMeterBinder` to fetch a prefix for metric names.
Introduced a `DEFAULT_EMPTY_METRIC_NAME_PREFIX` constant with an empty string as the default metric name prefix.
Prefixed metric names in five places in the `public final void bindTo(MeterRegistry registry)` method in `CacheMeterBinder`.

**Why is this important?**
Currently, Micrometer's `CacheMeterBinder` doesn't provide a straightforward way to customize metric names, which is useful for integrating seamlessly with systems like [ActiveMQ-Artemis](https://activemq.apache.org/components/artemis/documentation/latest/metrics.html) that follow unique metric naming conventions. This PR eliminates the need to unregister and re-register metrics, thereby simplifying metric management.

**Backwards Compatibility**
This feature is backward-compatible because the default prefix is an empty string. By default, existing metric names will not change, effectively appending an empty string to the current metric name.

**Test Coverage**
`AbstractCacheMetricsTest` class has been created to include tests that verify this new functionality, ensuring that all the common cache metrics have the correct prefix.

**Usage Example**
After this change, you can override the getMetricNamePrefix() method in a subclass to set a custom prefix for metric names.

```java
@Override
protected String getMetricNamePrefix() {
    return "custom_prefix_";
}
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
@@ -40,6 +40,10 @@ public abstract class CacheMeterBinder<C> implements MeterBinder {
 
     private static final String DESCRIPTION_CACHE_GETS = "The number of times cache lookup methods have returned a cached (hit) or uncached (newly loaded or null) value (miss).";
 
+    // Default metric name prefix is set to an empty string, indicating no prefix by
+    // default. This constant allows subclasses to use or override it as needed.
+    private static final String DEFAULT_EMPTY_METRIC_NAME_PREFIX = "";
+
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
@fartzy Please see this comment: https://github.com/micrometer-metrics/micrometer/issues/4047#issuecomment-1867583957 on the issue.

Issue #3872, opened 2023-05-31
Metrics lost on close if scheduled publish is in progress


pirgeo:
As pointed out in https://github.com/micrometer-metrics/micrometer/issues/3832#issuecomment-1561822331: 
> the JVM simply does not wait for Micrometer to finish publishing. This is intentional (Micrometer should not disturb your app)

Therefore, in some cases, the last export is not finished because the JVM shuts down before the export is completed. This can happen for exports triggered by a timer (periodic exports), as well as for exports triggered by a shutdown signal (e.g. hitting the actuator shutdown endpoint in a Spring Boot app).

I think adding a configuration option to wait for the exporter thread to finish its export would be helpful. This could be combined with a timeout to define the maximum time to wait until the app shuts down even if the data is not exported, in order to not block the shutdown indefinitely in case of network issues. By setting this timeout to 0, Micrometer would behave in the same way as it does today.

**Rationale**
From my research on #3832, I realized that it is possible that apps shut down without exporting data collected between the previous export and the shutdown signal. For certain metrics, this might be relevant information. I believe it should be the choice of the person running the application whether to have the application run for a few extra cycles in order to ensure data completeness. Additionally, for apps running only for a short period of time, there is a chance that _no data_ is exported if the first export would only happen after the app is shut down. By ensuring the export is complete before shutting down, even short-lived apps will always be able to export their data.

shakuzen:
Can you confirm if my assumption in my previous comment is true for what you're seeing?

pirgeo:
This is in line with what you wrote above. I originally saw the "Duplicate call to publish" message, meaning the scheduled publish was in progress while I tried to shut the app down. I have not had problems with unfinished exports when the shutdown-publish is responsible.
It seems to be that very rare case where the publishing is already in progress (because it's scheduled) and then the app gets a shutdown signal and exits before the publishing is finished.

pirgeo:
Any news on this?

pirgeo:
@shakuzen I am thinking about how this could best be tackled since it potentially can have quite the impact (e.g. losing data about how many background queues were shut down before the app shut down because the publishing does not complete). I was able to come up with two ways of dealing with this:

1. Make it possible to turn off the randomly delayed publishing. However, with the changes as they are now (with the atomic variable checking if publishing is already in progress), the lost data might not actually have anything to do with the random delay. If the publishing is in progress at the same time as the app shuts down, this problem can still appear.
[...]

Issue #3832, opened 2023-05-11
"Duplicate call to publish" error since 1.11.0-RC1


pirgeo:
**Describe the bug**
As part of some internal testing with our Dynatrace meter registry, I recently saw the following error message for the first time: 
```
Publishing is already in progress. Skipping duplicate call to publish().
```
This log message was introduced in https://github.com/micrometer-metrics/micrometer/pull/3714 and first released with Micrometer 1.11.0-RC1. 

I was trying to debug this a little bit, but it's really hard to reproduce (which is a good thing, I guess 😄). One thing that I did find out was that some change in 1.11.0-RC1 seems to lead to multiple publishing threads being created. E.g., when running the registry with micrometer core version 1.9.10 (the default for Spring Boot 2.7.11), I get 1 thread named `dynatrace-metrics-publisher`, which is responsible for pushing metrics every X seconds. The thread is created using the [default thread factory](https://github.com/micrometer-metrics/micrometer/blob/main/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java#LL59C75-L59C75), which is an instance of [NamedThreadFactory](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/util/NamedThreadFactory.java). Everything is fine and seems to work in this version.

However, when I switch over to Micrometer 1.11.0-RC1, I suddenly have two publishing threads ( `dynatrace-metrics-publisher` and  `dynatrace-metrics-publisher-2`). It, therefore, seems as if the NewThread method on the ThreadFactory is called twice. I am not quite sure how this happens, since I assume the threads are created [here](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java#L118) and the SingleThreadScheduledExecutor should not create two threads (maybe?). 

I have also tried another registry (I randomly picked the elastic registry), and it seems to also be creating two export threads instead of one. Hence, I believe this might have something to do with the base class (maybe StepMeterRegistry?). Have you seen this before? Any ideas where it could come from? As I said, I am having a hard time getting a reproducer ready (getting two threads is easy, but they seem to not interfere most of the time). I assume the error comes from a race condition when both threads try to export at the same time. This would also explain why it only shows up once in a blue moon. 

**Environment**

 - Micrometer version: 1.11.0-RC1
[...]

PR #4062, opened 2023-09-07
Allow waiting instead of dropping data in `duplicate call to publish()`.

pirgeo:
Fixes https://github.com/micrometer-metrics/micrometer/issues/3872

When the scheduled publish is already in progress when `close()` is called on the registry, the `duplicate call to publish()` error will show up and the data from the last export will be dropped. This is the intended behavior in Micrometer (influence the application lifecycle as little as possible). See #3872 and #3832 for more details. 

To honor the commitment to not influencing shutdown, I implemented this in a way where the current behavior does not change unless the user explicitly opts in. For the Dynatrace exporter, I added a higher timeout by default, as multiple of our customers asked us to ensure they can rely on the final export. It is still possible to opt-out for users of the dynatrace-registry by setting the timeout to zero.


shakuzen:
Thank you for the pull request and the tests and investigation around this. I'll close this since we went the route without the new configuration for timeout.
Superseded by #4287

Issue #4052, opened 2023-08-28
Builder pattern using Tags leads to high CPU overhead


DanielThomas:
We discovered some significant CPU hotspots in a service and found several places in our frameworks that were doing a significant amount of unnecessary work due to the  sort/dedup on every `Tags` class operation. In our case we were either using `Tags` as a builder with up to a dozen calls to `Tags::and` or has several calls to `tags(String, String)` on the builders.

Appears when this commit landed it switched all builders to use`Tags` internally, so incur the sort/dedup penalty for every call, making building tags quite expensive:

https://github.com/micrometer-metrics/micrometer/commit/b7bbea3d6935aaf794d538cf5f6da5155c9bb93f#diff-a64eec4db3aaf5a50e01e46efed55b3eecb772bcbe0ad929ef59b455c7a5ec2c

This makes `Tags` a poor choice for intermediate `Tags` state and if sort/dedupe can't be avoided, there should probably be a mutable builder class backing these usages instead. If that overhead can't be avoided, I'd argue that the builder-style methods on `Tags` should be deprecated, because it's easy to get into trouble and not realise it.

The starting state for the builders is also `Tags.EMPTY` so `Tags::and` could probably use an optimization tha returns the incoming `Tags` instance if you're calling `and` against `Tags.EMPTY`.

**Environment**

 - Micrometer version 1.11.3

**To Reproduce**

Use multiple calls to `tags` 

**Expected behavior**

Tag construction incurs as little overhead as possible.

**Additional context**

https://github.com/Netflix/dgs-framework/pull/1612

DanielThomas:
I should add, if there's the possibility of avoiding this completely, that'd be great - before these changes, these accounted for 50%(!) of the CPU overhead for the application. We've gotten that down a bunch, but it's unavoidable when doing the final measurement:

<img width="1145" alt="Screenshot 2023-09-08 at 6 10 51 pm" src="https://github.com/micrometer-metrics/micrometer/assets/1479220/13e10ce6-5af3-490b-9a74-fd090217fc8c">

shakuzen:
Thank you for the report and details. It's much appreciated. We would need to look into this carefully. Unfortunately, I think the complete motivation/context in which the original changes were made (almost 5 years ago, for version 1.1.0) is lost at this point. So we'd really need to dig into all things that could be affected by making changes around this.

> I should add, if there's the possibility of avoiding this completely, that'd be great

[...]

Issue #2275, opened 2020-09-28
Possible performance issues with usage of ....builder


Hronom:
We observing that solution like this:
```
            return Timer
                    .builder("execution.time")
                    .tag("executionName", name)
                    .tag("executionFullName", fullName)
                    .publishPercentiles(MetricsConstraints.defaultTimerPercentiles)
                    .register(Metrics.globalRegistry)
```
is slow compare to:
```
        return timersByFullName.computeIfAbsent(fullName, s -> Timer
                .builder("execution.time")
                .tag("executionName", name)
                .tag("executionFullName", fullName)
                .publishPercentiles(MetricsConstraints.defaultTimerPercentiles)
                .register(Metrics.globalRegistry));
```
In second solution we use `private static final ConcurrentHashMap<String, Timer> timersByFullName = new ConcurrentHashMap<>();`.

Why such performance degradation happens in micrometer?
Is it because we have a lot of metrics in `Metrics.globalRegistry` so it's becomes slow compare to local map `timersByFullName`.

What is the advice you can provide here?
Will first solution win, if we will use instead of `Metrics.globalRegistry`  - local registry for this metrics and add it in `Metrics.globalRegistry.add(localMeterRegistry);`

Thanks!

marcingrzejszczak:
Can you provide some proof to your findings? Do you have any benchmarks?

shakuzen:
> Why such performance degradation happens in micrometer?

There is a cost to the builder but it provides the most flexibility and usability. In your alternative, you need to compute a correct key that uniquely identifies each Timer. I don't know whether you are doing that given the code snippet, but note that a meter is uniquely identified by its name and full set of tags, after any registered meter filters are applied.
We of course want to keep the overhead of using the builder as low as possible, but we need to keep usability. And for most users, the cost should be negligible, but of course it depends on the use case. We're open to suggestions and are always striving to improve the performance of Micrometer. Let us know if you've identified something we can improve.

jonatan-ivanov:
Could you please check if this feature could be useful for you: https://github.com/micrometer-metrics/micrometer/pull/4097?

lenin-jaganathan:
I understand that there is a performance overhead associated with creating meters. See more info in [this repository ](https://github.com/lenin-jaganathan/micrometer-benchmark) where I have tried to benchmark the results.

[...]

Issue #535, opened 2018-04-03
Add shortcut to assign dynamic tags to Meters


gavlyukovskiy:
From slack thread it would be nice to have ability to define only dynamic part in metric while metric with static tags can be defined only once:
```
registry.timer("my.timer", "static", "tag", "response", response.status()).record(1);
```
->
```
Timer myTimer = registry.timer("my.timer", "static", "tag")
...
myTimer.withTags("response", response.status()).record(1);
```
where `withTags` returns new instance of timer with additional tags attached.

jkschneider:
Ingenious usability hack.

zeratul021:
@jkschneider is there a recommended way of solving this issue?
E.g. I want to have counter for registered users and the dynamic tag would be their country of origin.
TIA

jkschneider:
Sorry this won't get done for 1.1.0. It is actually much harder than it would initially seem. We don't habitually maintain a reference to the registry on each meter, so spawning a completely new metric is difficult. Also, for timers we don't tend to preserve all the options (like `DistributionSummaryConfig`) that we used to create a timer after it exists. Technically there is no way to go back and introduce this without a breaking API change.

jkschneider:
@zeratul021 It is possible to use dynamic tags now though, simply by registering it as you use it. If the meter already exists, Micrometer will return the existing one.

zeratul021:
@jkschneider thanks for the input. Yeah, I'm using it exactly like that. Seems there is nothing wrong with it for now.

u6f6o:
@jkschneider if I am not mistaken, this pattern (register again) cannot be applied to gauges without building a new gauge every time? In opposite to the other meters, the registry returns the gauge value instead of a wrapper representing the gauge (which makes perfect sense imo).

I currently have to migrate a lot of existing, dynamic gauges and came up with this workaround to register dynamic gauges: 

```java 
private final ConcurrentHashMap<Meter.Id, AtomicLong> dynamicGauges = new ConcurrentHashMap<>();
```

```java
private void handleDynamicGauge(String meterName, String labelKey, String labelValue, Long snapshot) {
    Meter.Id id = new Meter.Id(meterName, Tags.of(labelKey, labelValue), null, null, GAUGE);

    dynamicGauges.compute(id, (key, current) -> {
        if (current == null) {
            AtomicLong initialValue = new AtomicLong(snapshot);
            registry.gauge(key.getName(), key.getTags(), initialValue);
            return initialValue;
[...]

PR #4092, opened 2023-09-21
Add Meter.Cache for dynamic meter registration

qweek:
**Summary**
This PR aims to add Meter.Cache for dynamic meter registration. This allows to create meters with dynamic tags, making it more customizable and adaptable to systems with tags depending on external sources. This change is backward-compatible.

**Changes**
Introduce additional register() method in specific Meter to create Meter.Provider for ad-hoc Meter creation.

**Why is this important?**
Currently, Micrometer's Meter doesn't provide a straightforward way to customize tags without additional allocations in runtime. This PR eliminates the need to build, deduplicate and sort Tags on each call and allow to re-use metrics builder, thereby simplifying metric management.

**Backwards Compatibility**
This feature is backward-compatible because it doesn't affect existing register() method.

**Test Coverage**
MeterCacheTest class has been created to include tests that verify this new functionality, ensuring that all basic metrics supports dynamic registration.

**Usage Example**
After this change, you can use new register() method to create metric with dynamic tags.

```
Meter.Provider<String, Counter> provider = Counter.builder("counter")
            .register(registry, value -> Tags.of("source", value));

void request(String source) {
    provider.get(source).increment();
}
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Counter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Counter.java
@@ -18,6 +18,7 @@ package io.micrometer.core.instrument;
 import io.micrometer.common.lang.Nullable;
 
 import java.util.Collections;
+import java.util.function.Function;
 
 /**
  * Counters monitor monotonically increasing values. Counters may never be reset to a
@@ -128,6 +129,14 @@ public interface Counter extends Meter {
          * @return A new or existing counter.
          */
         public Counter register(MeterRegistry registry) {
+            return register(registry, tags);
+        }
+
+        public <K> Meter.Provider<K, Counter> register(MeterRegistry registry, Function<K, Tags> provider) {
+            return new Meter.Cache<>(key -> register(registry, tags.and(provider.apply(key))));
+        }
+
+        private Counter register(MeterRegistry registry, Tags tags) {
             return registry.counter(new Meter.Id(name, tags, baseUnit, description, Type.COUNTER));
         }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/DistributionSummary.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/DistributionSummary.java
@@ -24,6 +24,7 @@ import io.micrometer.core.instrument.distribution.ValueAtPercentile;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 /**
  * Track the sample distribution of events. An example would be the response sizes for
@@ -395,6 +396,14 @@ public interface DistributionSummary extends Meter, HistogramSupport {
          * @return A new or existing distribution summary.
          */
         public DistributionSummary register(MeterRegistry registry) {
+            return register(registry, tags);
+        }
+
+        public <K> Meter.Provider<K, DistributionSummary> register(MeterRegistry registry, Function<K, Tags> provider) {
+            return new Meter.Cache<>(key -> register(registry, tags.and(provider.apply(key))));
+        }
+
+        private DistributionSummary register(MeterRegistry registry, Tags tags) {
             return registry.summary(new Meter.Id(name, tags, baseUnit, description, Type.DISTRIBUTION_SUMMARY),
                     distributionConfigBuilder.build(), scale);
         }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Gauge.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Gauge.java
@@ -20,6 +20,7 @@ import io.micrometer.core.annotation.Incubating;
 import io.micrometer.core.instrument.distribution.HistogramGauges;
 
 import java.util.Collections;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.function.ToDoubleFunction;
 
@@ -192,6 +193,19 @@ public interface Gauge extends Meter {
          * @return A new or existing gauge.
          */
         public Gauge register(MeterRegistry registry) {
+            return register(registry, obj, tags);
+        }
+
+        public <K> Meter.Provider<K, T> register(MeterRegistry registry, Supplier<T> generator,
+                Function<K, Tags> provider) {
+            return new Meter.Cache<>(key -> {
+                final T obj = generator.get();
+                register(registry, obj, tags.and(provider.apply(key)));
+                return obj;
+            });
+        }
+
+        private Gauge register(MeterRegistry registry, T obj, Tags tags) {
             return registry.gauge(new Meter.Id(name, tags, baseUnit, description, Type.GAUGE, syntheticAssociation),
                     obj, strongReference ? new StrongReferenceGaugeFunction<>(obj, f) : f);
         }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
@@ -23,7 +23,9 @@ import io.micrometer.core.instrument.distribution.HistogramGauges;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -481,6 +483,29 @@ public interface Meter {
     }
 
     default void close() {
+    }
+
+    interface Provider<K, V> {
+
+        V get(@Nullable K key);
+
+    }
+
+    class Cache<K, V> implements Provider<K, V> {
+
+        private final Map<K, V> cache = new ConcurrentHashMap<>();
+
+        private final Function<K, V> mappingFunction;
+
+        public Cache(Function<K, V> mappingFunction) {
+            this.mappingFunction = mappingFunction;
+        }
+
+        @Override
+        public V get(@Nullable K key) {
+            return cache.computeIfAbsent(key, mappingFunction);
+        }
+
     }
 
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
@@ -29,6 +29,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
 import java.util.function.DoubleSupplier;
+import java.util.function.Function;
 import java.util.function.IntSupplier;
 import java.util.function.LongSupplier;
 import java.util.function.Supplier;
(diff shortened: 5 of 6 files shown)


qweek:
Related issues that may have benefit from this change:
https://github.com/micrometer-metrics/micrometer/issues/4052
https://github.com/micrometer-metrics/micrometer/pull/2285

jonatan-ivanov:
Thank you for the PR! We are planning to do something similar (please see below), it seems this PR shows a different way to implement that.
My two cents:

>This allows to create meters with dynamic tags

Just to clarify things: Micrometer is capable of doing this, the `register` method is basically a `createOrGet` backed by a `Map` so if you call register with different (dynamic) tags, it will create new meters, if you call it with the same tags, it will return the meter that was registered previously. (It seems you are aware of this, I just wanted to call this out since other users might not be.)

>This PR eliminates the need to build, deduplicate and sort Tags on each call and allow to re-use metrics builder, thereby simplifying metric management.

I'm not sure the PR does this or it works as intended. Builders are not meant to be shared (they are mutable) and as far as I can tell (maybe I misunderstood something), every time you ask for a new (tag) value, the builder will be modified (`tags.and` modifies the builder) so I guess we have a shared mutability (concurrency) issue through it. Or am I misunderstanding something?

I think this is somewhat similar to this approach that we are planning to implement in 1.12 (in the next 2 weeks 🤞🏼): https://github.com/micrometer-metrics/micrometer/issues/535#issuecomment-1485992683
The approach in the comment is not caching anything (it's basically a syntax-sugar/shortcut) but it feels a bit more explicit/straightforward with tags.
```
timerProvider.get("first.value").record(latency); // in this PR
```
vs.
```
timerProvider.apply(Tags.of("name", name)).record(latency); // in the linked comment
```

I'm not sure the cache is needed or it's a real benefit in general. Also, items are never removed from the cache, e.g.: meters can be removed from the registry so this is a memory leak by definition. Sometimes meters have to be removed and re-registered, see the Kafka instrumentation. 🙁
In the long run, the cache might have some memory impact. Not just because of the leak I mentioned above but because it basically doubles the references to every `Meter` that is created through it (the registry also stores them). Also, since the key can be anything, it might not be apparent to the users that it must have a fair `haschCode` and `equals` implementation.  I'm also thinking if having a lot of short-lived builder instances is better or not from the memory perspective since they will be eliminated by the GC while the cache won't be.

qweek:
> If you call it with the same tags, it will return the meter that was registered previously.
Yep, as you mentioned, I'm aware of this and `registerMeterIfNecessary()` method still do unnecessary work before call to `getOrCreateMeter()`, because it applies all filters in `getMappedId()`.
> builder will be modified (tags.and modifies the builder)
No, I intentionally use immutable operations.
Compare existing code
```register(registry, tags.and(provider.apply(key)))```
and mutable "classic" version
```
tags(provider.apply(key));
register(registry, tags);
```
In second case, as side effect you will always override internal builder tags.
Expected behaviour (added similar test case)
```
var provider = Counter.builder("my.counter")
            .tag("default.tag", "default.value")
            .register(registry, value -> Tags.of("key", value));
provider.get("first.value"); // Tags: ("default.tag", "default.value") and ("key", "first.value")
provider.get("second.value"); // Tags: ("default.tag", "default.value") and ("key", "second.value")
```
> https://github.com/micrometer-metrics/micrometer/issues/535#issuecomment-1485992683
Yes, it similar approach, but it will have same problems as described before: unnecessary work for process Tags and build Meter.
Anyway, it will help to solve our problem (our existing solution is much worse, because we need to use reflection due to discussed immutability problems with tags).

According to possible memory leaks, I agree that it's possible, but can be described in documentation. It doesn't affect existing code and we still have possibility to create meters with usual approach. In some cases we already have methods which create meters, but didn't have way to close it without specific logic.
```
// internal gauge is not returned to user
T MeterRegistry.gauge(String name, Iterable<Tag> tags, @Nullable T stateObject, ToDoubleFunction<T> valueFunction)
```
> having a lot of short-lived builder instances is better
Unfortunately no, it was one of the main reasons why we implement this change, because existing approaches highly affect Young GC, lead to minor pauses and also use a lot of CPU.

**Comment by jonatan-ivanov** [comment]
>registerMeterIfNecessary() method still do unnecessary work before call to getOrCreateMeter(), because it applies all filters in getMappedId().

I think that is necessary in some use cases. You might not use `MeterFilter` but other users do.

>No, I intentionally use immutable operations. Compare existing code `register(registry,tags.and(provider.apply(key)))` and mutable "classic" version

Indeed; sorry, somehow I misread/interpret it. That extra private `register` method that basically reuses all the fields of the `Builder` except the `tags` so that tags are injectable seems like a great idea. Btw this can also help to greatly simplify the approach in https://github.com/micrometer-metrics/micrometer/issues/535#issuecomment-1485992683 since that always makes a new copy of the builder but with the extra `register` I think copying the `Builder` is not needed anymore since `register` takes care of "copying" (reusing the builder fields).

>Yes, it similar approach, but it will have same problems as described before: unnecessary work for process Tags and build Meter.

With your idea using the extra `register` method, I think it is simple to convert the original approach (https://github.com/jonatan-ivanov/micrometer/commit/f47c83b63672547aba5ede8db5982a1763e80299) into something that does not need to copy the builder (https://github.com/jonatan-ivanov/micrometer/commit/228b2b019cbd170aef6130fd4528dd3f589fd163). Please notice that you might have a slightly different goal than #535. That one is about API usability while I think you also want to spare some work.

>According to memory leaks, I agree that it's possible, but can be described in documentation.

I don't think this would be fortunate, users are using Micrometer in various ways and I don't think providing them a seemingly harmless method that just creates meters but can leave garbage behind is a good idea.

>In some cases we already have methods which create meters, but didn't have way to close it without specific logic.
```
// internal gauge is not returned to user
T MeterRegistry.gauge(String name, Iterable<Tag> tags, @Nullable T stateObject, ToDoubleFunction<T> valueFunction)
```

This is by design but you can still get a reference to this `Gauge` if you want to remove it: `registry.find(...).gauge()`.

>Unfortunately no, it was one of the main reasons why we implement this change, because existing approaches highly affect Young GC, lead to minor pauses and also use a lot of CPU.

From the young-gen perspective, I think you are right, but I was talking about the overall impact: "if having a lot of short-lived builder instances is better or not from the memory perspective". The trade-off is basically more garbage in young-gen (your pain point) vs. increased heap usage on the long run.

Could you please check the changes I made on the proposed solution for #535 that now does not need a builder thanks to your idea https://github.com/jonatan-ivanov/micrometer/commit/228b2b019cbd170aef6130fd4528dd3f589fd163 and tell me how much would it help in your use-case?

I think I would still prefer being able to pass all the extra tags instead of just one value:
```java
timerProvider.get("first.value").record(latency);
//vs.
timerProvider.apply(Tags.of("name", name)).record(latency);
```
I think that approach is more explicit (readable), looking at only `timerProvider.get("first.value")`, it's not apparent what should I expect.

qweek:
> You might not use MeterFilter but other users do.

I didn't mean that filters itself are unnecessary (we actually use them), but when you convert MeterId -> MappedMeterId it is usually expected that after applying all the filters you will get the same Id.
> I think copying the Builder is not needed anymore since register takes care of "copying" (reusing the builder fields).

I happy that you like idea with `register(registry, tags)` method. The only addition to (https://github.com/jonatan-ivanov/micrometer/commit/228b2b019cbd170aef6130fd4528dd3f589fd163), is that I would still prefer an interface like `Meter.Provider { M withTags(Tags extraTags); }` instead of a generic function. 

On the contrary, it seems to me that the idea of copying a builder can be very useful when implementing a factory :)
I would rather want to have something like this:
1. Add private copy constructor and public `Builder copy()` method
2. Add method that convert builder to factory, like
`<F extends Factory> toFactory(MeterRegistry registry, BiFunction<Builder, MaterRegistry, F> constructor)`
4. Factory interface should have only `close()` method, then users will be able to add any methods and implementations, like `Counter withTags(Tags extraTags)`

As you said, Gauge is different case, to make it work same way as other meters we need:
1. Additional methods on Builder to override `stateObject(T stateObject)` and `valueFunction(...)`
2. Have access to `stateObject` from Gauge class
3. Then we can implement Factory methods to add dynamic tags for cases like
 `withObject("objectName", object)` for multiple objects as tags <- 1 property
 `withProperty("propertyName", State::getPropertyX)`  for 1 object -> multiple properties as tags

Tell me what you think about this, if you are interested in such an implementation, I can make a proof of concept.

jonatan-ivanov:
>when you convert MeterId -> MappedMeterId it is usually expected that after applying all the filters you will get the same Id

I'm not sure I follow/understand, you need to look up the Meters anyway, right? Should we create a new issue for this?

>I would still prefer an interface like Meter.Provider { M withTags(Tags extraTags); } instead of a generic function.

We usually don't do that but I can see some advantages of it, I added a comment to the PR: https://github.com/micrometer-metrics/micrometer/pull/4097#pullrequestreview-1656119221

>On the contrary, it seems to me that the idea of copying a builder can be very useful when implementing a factory :)

Yeah, that's what I did originally (copy ctor) but with your idea, the extra Builder object can be eliminated. Also, I'm not sure it is needed to create meters with different builder settings (other than tags), what would be the use case for that? I think the other settings of the builder do not create a new meter when you change them so do not seem that useful to me.

>As you said, Gauge is different case, to make it work same way as other meters we need:

I feel doing something like this to `Gauge` or other "async" Meters `FunctionXyz` is a slippery slope and can be misused easily. For dynamic tagging gauges, you might want to look at `MultiGauge` where you can register gauges dynamically by providing the dynamic tags: https://micrometer.io/docs/concepts#_multi_gauge

lenin-jaganathan:
@jonatan-ivanov We have also faced the issue of using the default builder pattern to record values in a Meter.

There are 2 big concerns we had,

- Recording using the builder is a slower process and since this is in the hot path, it becomes super critical to have the recording as light as possible
- High memory allocation during the recording of data and hence increased Young Gen activity. 

See [this repo](https://github.com/lenin-jaganathan/micrometer-benchmark) where I have a few benchmarks run where there was a clear issue with the current approach to record values.

In the above example, I have commented out the meter filter to just show that there is still a clear difference if there are no meter filters involved. But in real uses, we will have multiple meter filters and that will only be slower (also increase allocation). And when these meter filters involve adding tags/manipulating tags/changing the name of the meter it just gets worse. 

We were forced to use a Cache for directly interacting with the meters to have better performance. This is also not straightforward, as we need to handle meter removal and invalidating the cache when meters are removed and other cases. Yet we encourage our internal users to use the cache to have an improved performance number.

jonatan-ivanov:
@lenin-jaganathan Could you please check if the new `withRegistry` feature (since 1.12.0) can help mitigating/eliminating those issues? It should eliminate the need of a new builder when you would like to attach tags dynamically, see: https://github.com/micrometer-metrics/micrometer/pull/4097


@qweek What do you think, can https://github.com/micrometer-metrics/micrometer/pull/4097 supersede this PR?

lenin-jaganathan:
I explored the option provided in the new version to attach tags dynamically. But, my ask still differs a bit from what is being discussed here. My concern is calling the register on the registry seems to be a costly operation (especially with meter filters this creates a lot of Garbage and exec time is also elevated highly). 

What I was looking for is a cache-like solution that will return the existing meters for a given set of meter names, tags, etc.

lenin-jaganathan:
I think https://github.com/micrometer-metrics/micrometer/issues/2275 relates more to the issue I discussed and I will add it here.

PR #4093, opened 2023-09-22
Ignore noop children when polling composite meters

jonatan-ivanov:
When a CompositeMeterRegistry is used and one of its CompositeMeters is polled (its value is fetched), the returned value can depend on the order of the registries inside of the composite if the composite contains a registry that has any NoopMeters.

Example: a CompositeMeterRegistry contains two registries, A and B. We create a counter in the composite and increment it once. After this both A and B contain one counter but lets say that in A the counter is ignored so it will be noop.
When the count called on CompositeCounter, it can return either 0 (if NoopCounter was used)
or 1 (if the non-noop counter was used).

In order to fix this, we can ignore the NoopMeters when Meters are polled in a composite registry.

Closes gh-1441

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/composite/AbstractCompositeMeter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/composite/AbstractCompositeMeter.java
@@ -18,11 +18,11 @@ package io.micrometer.core.instrument.composite;
 import io.micrometer.core.instrument.AbstractMeter;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.noop.NoopMeter;
 import io.micrometer.core.lang.Nullable;
 
 import java.util.Collections;
 import java.util.IdentityHashMap;
-import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -49,9 +49,11 @@ abstract class AbstractCompositeMeter<T extends Meter> extends AbstractMeter imp
     }
 
     T firstChild() {
-        final Iterator<T> i = children.values().iterator();
-        if (i.hasNext())
-            return i.next();
+        for (T next : children.values()) {
+            if (!(next instanceof NoopMeter)) {
+                return next;
+            }
+        }
 
         // There are no child meters. Return a lazily instantiated no-op meter.
         final T noopMeter = this.noopMeter;
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/composite/CompositeMeterRegistryTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/composite/CompositeMeterRegistryTest.java
@@ -28,18 +28,23 @@ import io.micrometer.core.instrument.step.StepRegistryConfig;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 import reactor.core.publisher.Flux;
 import reactor.core.scheduler.Schedulers;
 
 import java.time.Duration;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
 
+import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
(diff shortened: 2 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/composite/AbstractCompositeMeter.java
```diff
@@ -49,9 +49,11 @@ final Iterable<T> getChildren() {
     }
 
     T firstChild() {
-        final Iterator<T> i = children.values().iterator();
-        if (i.hasNext())
-            return i.next();
+        for (T next : children.values()) {
```
jonatan-ivanov, line 52: This _should not_ be on the "hot path" (i.e.: recording a measurement or retrieving meters) but it will be called every time a value is fetched for publishing.

I'm thinking if there might be another way to implement this: modifying the `children` collection so that it has some priority ordering so the non-noop meters in it take precedence and will be "in front" of the collection (if any). That solution though might be more problematic since it seems a bit more complicated and it can be on a "hotter" path (meter creation).

micrometer-core/src/main/java/io/micrometer/core/instrument/composite/AbstractCompositeMeter.java
```diff
@@ -49,9 +49,11 @@ final Iterable<T> getChildren() {
...
+            if (!(next instanceof NoopMeter)) {
+                return next;
+            }
+        }
 
         // There are no child meters. Return a lazily instantiated no-op meter.
         final T noopMeter = this.noopMeter;
```
jonatan-ivanov, line 59: If we are at this point, the `children` collection was either empty or contains only noop meters. In the second case we could return it the first value (like we did before this change) but I'm not sure it worth the hassle.

stefano-salmaso:
You are merging this into `micrometer-metrics:1.9.x`, why not in the latest?

micrometer-core/src/main/java/io/micrometer/core/instrument/composite/AbstractCompositeMeter.java
stefano-salmaso: Have you seen [this PR](https://github.com/micrometer-metrics/micrometer/pull/1443/commits/fa1c14e5b3cdf1139df7cfa39a38ee8e025dce28)? The approach is completely different but certainly more efficient. We should understand if there is any hidden side effect if the meter is not added..

jonatan-ivanov: I guess https://github.com/micrometer-metrics/micrometer/pull/1442 is also connected.
I would not do what is in https://github.com/micrometer-metrics/micrometer/pull/1443, it is somewhat similar to the priority ordering solution I explained before so its path is "hotter" and it omits meters from the registry.

jonatan-ivanov:
>You are merging this into micrometer-metrics:1.9.x, why not in main?

Because I think we should consider this as a bug not a new feature.
If we would merge this in `main` only `1.12.x` and newer versions will have this fix. If we merge this into `1.9.x`, then that and newer versions will have the fix (1.9, 1.10, 1.11, 1.12, ...)

klau-nagy-cs:
Hi,
Are there any news on this PR? Do you know when this fix will be merged and released?
We have the same problem using version 1.11 and this fix would solve our issue.
Thank you!

jkemming:
Hi @jonatan-ivanov,

I've [implemented a solution](https://github.com/micrometer-metrics/micrometer/compare/main...jkemming:micrometer:gh-1441) on top of your PR that addresses the "hot path" issue, please have a look. Feel free to incorporate the changes here, or if you want I can open a separate PR.

Also, there are two alternative solutions I have considered. If any of those sound better to you, please let me know:

* If atomic updates to `children` and `firstNonNoopChild` aren't required, the inner class could essentially be inlined.
* Alternatively, we could keep the calculation on the "hot path" and cache the result, clearing the cache when `children` is updated.

jonatan-ivanov:
@klau-nagy-cs 
>Are there any news on this PR? Do you know when this fix will be merged and released?
We have the same problem using version 1.11 and this fix would solve our issue.

Szia, No news, we need to review and discuss it. Adding 👍🏼 to the issue could help with prioritizing.


@jkemming
>I've [implemented a solution](https://github.com/micrometer-metrics/micrometer/compare/main...jkemming:micrometer:gh-1441) on top of your PR that addresses the "hot path" issue, please have a look. Feel free to incorporate the changes here, or if you want I can open a separate PR.

Hi, I'm not sure I understand, My comment about the "hot path" is saying that what I'm doing in the PR should not be on the hot path so it should not have a hot path issue. That part of the code should only called once per every publishing but depending on the setup add/remove registries could be even less frequent, we should definitely consider this option too.

jkemming:
I'm afraid I misunderstood your comment then, sorry for that... Should there be anything I can do to help, please let me know.

benkeil:
Can we merge this?

slovdahl (approved):

hoa2clj:
Hello!
Do you have any updates on this PR? Or do you know when it might be merged and released?
Thank you in advance!

PR #4097, opened 2023-09-23
Add shortcuts to assign dynamic tags to Meters

jonatan-ivanov:
This feature is a shortcut to create/get certain meters in a simpler way using the builder of that meter but without the need of recreating the builder every time.

This is what dynamic tagging looks like before this change (a bit longer and multiple builder instances will be created):
```java
private final MeterRegistry registry = ...;

void favoriteColor(Color color) {
    Counter.builder("colors")
        .tags(Tags.of("color", color.name()))
        .register(registry)
        .increment();
}

```
And this is what you can do after (a bit shorter and one builder instance will be created):
```java
private final MeterRegistry registry = ...;
private final MeterProvider<Counter> meterProvider = Counter.builder("colors").withRegistry(registry);

void favoriteColor(Color color) {
    meterProvider.withTags(Tags.of("color", color.name())).increment();
}
```

Closes gh-535
See gh-4092

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
@@ -35,6 +35,7 @@ import static java.util.Collections.singletonList;
  * A named and dimensioned producer of one or more measurements.
  *
  * @author Jon Schneider
+ * @author Jonatan Ivanov
  */
 public interface Meter {
 
@@ -477,6 +478,26 @@ public interface Meter {
         public Meter register(MeterRegistry registry) {
             return registry.register(new Meter.Id(name, tags, baseUnit, description, type), type, measurements);
         }
+
+    }
+
+    /**
+     * Convenience interface to create new meters from tags based on a common
+     * "template"/builder. See usage in Meter implementations, e.g.: {@code Timer},
+     * {@code Counter}
+     *
+     * @param <T> Meter type
+     * @since 1.12.0
+     */
+    interface MeterProvider<T extends Meter> {
+
+        /**
+         * Registers (creates a new or gets an existing one if already exists) Meters
+         * using the provided tags.
+         * @param tags Tags to attach to the Meter about to be registered
+         * @return A new or existing Meter
+         */
+        T withTags(Iterable<? extends Tag> tags);
 
     }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
@@ -499,6 +499,27 @@ public interface Meter {
          */
         T withTags(Iterable<? extends Tag> tags);
 
+        /**
+         * Registers (creates a new or gets an existing one if already exists) Meters
+         * using the provided tags.
+         * @param tags Tags to attach to the Meter about to be registered
+         * @return A new or existing Meter
+         */
+        default T withTags(String... tags) {
+            return withTags(Tags.of(tags));
+        }
+
+        /**
+         * Registers (creates a new or gets an existing one if already exists) Meters
+         * using the provided tags.
+         * @param key the tag key to add
+         * @param value the tag value to add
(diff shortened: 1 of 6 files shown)


gavlyukovskiy:
This looks absolutely awesome!

micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
```diff
@@ -430,6 +427,18 @@ public Builder description(String description) {
...
 
+        /**
+         * Convenience method to create new meters from the builder that only differ in
+         * tags. This method can be used for dynamic tagging by creating the builder once
+         * and applying the dynamically changing tags using the returned {@link Function}.
+         * @param registry A registry to add the meter to, if it doesn't already exist.
+         * @return A {@link Function} that returns a meter based on the provided tags.
+         * @since 1.12.0
+         */
+        public Function<Tags, Timer> with(MeterRegistry registry) {
```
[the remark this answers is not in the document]
shakuzen: I think a dedicated type would be useful, since users are going to keep a reference to this to use it. That does increase the importance somewhat on getting the appropriate naming and API before release.

jonatan-ivanov:
@gavlyukovskiy @qweek I modified this a little bit, what do you think?

jonatan-ivanov:
This feature is a shortcut to create certain meters in a simpler way using the builder of that meter but without the need of recreating the builder every time.

This is what dynamic tagging looks like before this change (a bit shorter and multiple builder instances will be created):
```java
private final MeterRegistry registry = ...;

void favoriteColor(Color color) {
    Counter.builder("colors")
        .tags(Tags.of("color", color.name()))
        .register(registry)
        .increment();
}

```
And this is what you can do after (a bit shorter and one builder instance will be created):
```java
private final MeterRegistry registry = ...;
private final MeterProvider<Counter> meterProvider = Counter.builder("colors").withRegistry(registry);

void favoriteColor(Color color) {
    meterProvider.withTags(Tags.of("color", color.name())).increment();
}
```

micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
```diff
@@ -480,6 +481,26 @@ public Meter register(MeterRegistry registry) {
...
+     */
+    interface MeterProvider<T extends Meter> {
+
+        /**
+         * Registers (creates a new or gets an existing one if already exists) Meters
+         * using the provided tags.
+         * @param tags Tags to attach to the Meter about to be registered
+         * @return A new or existing Meter
+         */
+        T withTags(Iterable<? extends Tag> tags);
```
steven-sheehy, line 500: Can this also support overloaded methods for `String... tags` and `String key, String value` like most metrics have?

steven-sheehy, line 500: Why `withTags()` versus the more commonly used elsewhere `tags()`?

shakuzen: Yes, I think we can add default implementations for those that delegate to this method for convenience.

shakuzen: I don't want to speak for @jonatan-ivanov, but I think the idea was to differentiate from the `tags()` method on the builders because those return a builder and can chain, whereas this method returns a registered `Meter` and cannot be chained to add more tags. I spent some time trying to think of better naming, but I was not able to come up with something. This is definitely something we're hoping to get feedback on and improve. Ideally, we would not change the naming after RC1, but it may be worth making an exception in this case if we cannot come up with something better before merging this.

jonatan-ivanov: 👍🏼 I added them.

jonatan-ivanov: Yeah, `tags` might be confusing since that is on the `Builder` as well as `MeterProvider`.

PR #4102, opened 2023-09-28
Support API Token Type from the Wavefront SDK

jonatan-ivanov:
API Token type was added to the Wavefront SDK so that Wavefront can support multiple types of tokens/auth mechanisms: https://github.com/wavefrontHQ/wavefront-sdk-java/releases/tag/v3.4.1

### DIFF
--- a/implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
+++ b/implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
@@ -15,6 +15,8 @@
  */
 package io.micrometer.wavefront;
 
+import com.wavefront.sdk.common.clients.service.token.TokenService;
+import com.wavefront.sdk.common.clients.service.token.TokenService.Type;
 import io.micrometer.common.lang.Nullable;
 import io.micrometer.core.instrument.config.validate.InvalidReason;
 import io.micrometer.core.instrument.config.validate.Validated;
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
```diff
@@ -111,6 +113,14 @@ default String source() {
         });
     }
 
+    default TokenService.Type apiTokenType() {
```
jonatan-ivanov, line 116: Should we surface this from the Wavefront SDK vs. create a wrapper of it in Micrometer?
We do surface SDK types in case of [stackdriver](https://github.com/micrometer-metrics/micrometer/blob/973581c0839435ad0dcb5a8731d645f21c121f23/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverConfig.java#L104) but not in case of [otlp](https://github.com/micrometer-metrics/micrometer/blob/973581c0839435ad0dcb5a8731d645f21c121f23/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java#L102).

oppegard (approved):

implementations/micrometer-registry-wavefront/src/main/java/io/micrometer/wavefront/WavefrontConfig.java
shakuzen: I think the difference between OTLP and this is that we are only passing this to the Wavefront SDK. We are not doing anything with it ourselves. Whereas with OTLP, we need to do things differently depending on the value of the aggregation temporality enum. If we use a type we don't control and a new value is added, we can't handle that properly; so it feels better to limit the values to those we knowingly can support. In this case, it's up to the Wavefront SDK to support the values in `TokenService.Type`, so I think this is fine.

shakuzen (approved):

shakuzen:
We can handle it as a follow-up, but is there some documentation we can point to so users know the format and parts expected for the token?

jonatan-ivanov:
@oppegard Is there such docs? (see Tommy's question above)

Issue #3489, opened 2022-10-16
DistributionStatisticConfig object is mutable 


gopircr:
**Describe the bug**

Builder to create DistributionStatisticConfig object doesnt follow the standard [Builder design pattern ](https://blogs.oracle.com/javamagazine/post/exploring-joshua-blochs-builder-design-pattern-in-java) and hence the state of the config object can potentially  be modified after it is created. 

**Environment**

 - Micrometer version 1.9.5
 

**To Reproduce**

 ```
       DistributionStatisticConfig.Builder  b = DistributionStatisticConfig.builder()
                .bufferLength(1000)
                .maximumExpectedValue(300.0);
        
        DistributionStatisticConfig config = b.build();
        
        // prints 1000
        System.out.println(config.getBufferLength()); 

        //builder is reused to create more config objects
        b.bufferLength(400);

        //side effect of updating the builder. It prints 400
        System.out.println("again-"+config.getBufferLength()); 

```

**Expected behavior**

The standard builder pattern is used to create immutable objects and if the desired behavior is to allow mutability then builder name can be a misnomer and can cause confusion.   Its perhaps a common practice that builder itself can be reused to create more objects of the same type ( in this case its config object ).

marcingrzejszczak:
What problem would this change actually fix?

PR #4104, opened 2023-10-03
add copy() method to Builders

qweek:
**Summary**
This PR allows to reuse builders as templates for dynamic tags, factories, etc.

**Changes**
Introduce additional copy() method in all specific Meter's Builders, Tags and DistributionStatisticConfig.

**Why is this important?**
Currently, Micrometer doesn't provide a straightforward way to reuse builders. This PR eliminates the need to fully describe all fields for each similar builder, thereby simplifying metric management.

**Backwards Compatibility**
This feature is partially backward-compatible:
- it doesn't affect existing methods and constructors for Meter's Builders and Tags
- but it changes DistributionStatisticConfig behavior and make it consistent with common Builder pattern
    (fix bug https://github.com/micrometer-metrics/micrometer/issues/3489)

**Test Coverage**
CopyableTest class has simple test that verify new functionality, will extend it to cover all cases if you agree with design.

**Usage Example**
After this change, you can use new copy() method to create metric with dynamic tags
```
final Counter.Builder template = Counter.builder("counter");
template.copy().tags("tagA", "valueA").register(registry);
template.copy().tags("tagB", "valueB").register(registry);
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/DistributionStatisticConfig.java
@@ -49,34 +49,37 @@ public class DistributionStatisticConfig implements Mergeable<DistributionStatis
     public static final DistributionStatisticConfig NONE = builder().build();
 
     @Nullable
-    private final Boolean percentileHistogram;
+    private Boolean percentileHistogram;
 
     @Nullable
-    private final double[] percentiles;
+    private double[] percentiles;
 
     @Nullable
-    private final Integer percentilePrecision;
+    private Integer percentilePrecision;
 
     @Nullable
-    private final double[] serviceLevelObjectives;
+    private double[] serviceLevelObjectives;
 
     @Nullable
-    private final Double minimumExpectedValue;
+    private Double minimumExpectedValue;
 
     @Nullable
-    private final Double maximumExpectedValue;
+    private Double maximumExpectedValue;
 
     @Nullable
-    private final Duration expiry;
+    private Duration expiry;
 
     @Nullable
-    private final Integer bufferLength;
+    private Integer bufferLength;
 
     public static Builder builder() {
         return new Builder();
     }
 
-    public DistributionStatisticConfig(@Nullable Boolean percentileHistogram, @Nullable double[] percentiles,
+    public DistributionStatisticConfig() {
+    }
+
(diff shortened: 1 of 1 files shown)


qweek:
Also suggest to make breaking change and convert public `DistributionStatisticConfig` and `DistributionStatisticConfig.Builder` constructors to protected.
Now you can create `new DistributionStatisticConfig()`, but you can't change private fields.
You can create `new DistributionStatisticConfig.Builder()`, but also have `builder()` method with same functionality.

jonatan-ivanov:
Let me collect similar/involved issues/PRs here:
- https://github.com/micrometer-metrics/micrometer/issues/3489
- https://github.com/micrometer-metrics/micrometer/pull/3949
- https://github.com/micrometer-metrics/micrometer/pull/3854
- https://github.com/micrometer-metrics/micrometer/pull/4092
- https://github.com/micrometer-metrics/micrometer/pull/4097

I think we should gather the use cases for this and what we can provide for this today, what changes might needed.

qweek:
Some thoughts:

- https://github.com/micrometer-metrics/micrometer/issues/3489
Worth to fix. DistributionStatisticConfig will be immutable if we convert public constructor to protected.

- https://github.com/micrometer-metrics/micrometer/pull/3949
I'm not sure that I understand use case. Usually DistributionStatisticConfig is hidden in Builders, so maybe it this case be covered with reusable Builders.

- https://github.com/micrometer-metrics/micrometer/pull/3854
I think it shouldn't be implemented this way. As I understand you had to create your own Builder to work with extendable DistributionStatisticConfig, so instead you can add additional fields in custom Builder. It's one of reasons why I made all Builder's constructors protected.

- https://github.com/micrometer-metrics/micrometer/pull/4092
Cache is only 1 use case for factories, so it's better to make it more flexible.

- https://github.com/micrometer-metrics/micrometer/pull/4097
I like this approach. It didn't create additional objects, but also cover only 1 use case and we need to add some custom logic above to make it work in other cases.

I prefer to split this problem into 2 parts:
- make more flexible builders (make it reusable, add methods to gauge, maybe add name() method to builders)
- collect use cases and find better solution for factories (delayed construction, caches, etc)
- 
Right now, I know 2 use cases related to tags (dynamic tags and gauge with different objects/functions per dynamic tag)

jonatan-ivanov:
@qweek Would https://github.com/micrometer-metrics/micrometer/pull/4097 solve your first use-case with dynamic tags and would using `MultiGauge` help in the second?

PR #4106, opened 2023-10-05
Expand dependabot config to update everything

shakuzen:
Now that we have converted to using a version catalog TOML file (see gh-4095), we can configure dependabot to update all of our dependencies. This will reduce the workload on maintainers and keep things up-to-date better, avoiding human error.

The dependabot configuration separates out "build dependencies" which are things like build plugins, which should not directly affect users of Micrometer. We want to keep these versions and build configuration in general consistent between all supported branches. Hence, we will accept minor version upgrades, since these should not affect users of our library. A pull request will only be opened to the oldest supported branch, and the change should be able to be cleanly merged forward, due to keeping build configuration and versions consistent between supported branches. Such upgrades do not need to be included in release notes.

The dependabot configuration for "non-build dependencies" is configured for each supported branch. We typically only upgrade patch versions of these dependencies in maintenance branches. On the development branch `main` we will accept minor version upgrades. These upgrades will be included in the release notes, and having a pull request to each branch, even for the same upgrade, makes generating release notes automatically easier.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,19 +5,96 @@ registries:
     url: https://plugins.gradle.org/m2
     username: dummy # Required by dependabot
     password: dummy # Required by dependabot
+  maven-central:
+    type: maven-repository
+    url: https://repo1.maven.org/maven2/
+    username: dummy # Required by dependabot
+    password: dummy # Required by dependabot
 updates:
-  - package-ecosystem: "gradle"
-    directory: "/"
-    target-branch: "1.9.x" # oldest OSS supported branch
-    allow:
-      - dependency-name: "com.gradle*"
-    registries:
-      - gradle-plugin-portal
-    schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 10
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,19 +5,96 @@ registries:
...
+      # only upgrade minor and patch versions for build dependencies
+      - dependency-name: "*"
+        update-types:
+          - version-update:semver-major
     registries:
       - gradle-plugin-portal
+      - maven-central
+# Non-build dependencies; target every supported branch
+  - package-ecosystem: gradle
+    directory: /gradle
```
shakuzen, line 55: Dependabot config does not let you define multiple configurations for the same combination of package ecosystem, directory, and target branch. Hence we hack around this by specifying a different directory for "non-build dependencies", which we happen to know will all be defined in the TOML file in the `/gradle` folder. This differentiates it from the configuration for the "build dependencies" which uses `/` as the directory because they will be defined in both `settings.gradle` in the root as well as the TOML file in `/gradle` and setting `/` as the directory covers both.

PR #4201, opened 2023-10-10
Move Jakarta JMS instrumentation to micrometer-jakarta10

shakuzen:
Introduces a new module micrometer-jakarta10 and moves the previously merged JMS instrumentation there. The reason for this is that we tend to try to support a range of versions when possible for our users, and if Jakarta makes breaking changes in the future, we would not be able to support that in a single module (such as micrometer-core, or micrometer-jakarta) unless the new Jakarta artifacts used different coordinates.

The Jakarta JMS instrumentation was introduced in #4007. This moves it, unchanged, to the new module for the above mentioned reason.
This is a portion of the changes that were proposed in #3989.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -327,7 +327,7 @@ subprojects {
 
             check.dependsOn("testModules")
 
-            if (!(project.name in ['micrometer-osgi-test'])) {
+            if (!(project.name in ['micrometer-jakarta10'])) {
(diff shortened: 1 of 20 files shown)


shakuzen:
In 9ba22645b56881e123d51b55979a5c0d6ae25d0f I lowered the baseline to Jakarta EE 9. This will allow more people to use the instrumentation and there's no reason I know of to baseline on 10 now.

Issue #4202, opened 2023-10-10
JMS Context properties    'X-B3-TraceId' is not a valid java identifier.


tw-atroehrsm:
**Describe the bug**

Exception when publising a JMS message. Not all characters are allowed according to JMS naming standard, i.e hyphen.

```
	
jakarta.jms.JMSRuntimeException: AMQ139012: The property name 'X-B3-TraceId' is not a valid java identifier.

	at org.apache.activemq.artemis.jms.client.ActiveMQMessage.checkProperty(ActiveMQMessage.java:911)
	at org.apache.activemq.artemis.jms.client.ActiveMQMessage.setStringProperty(ActiveMQMessage.java:669)
	at io.micrometer.core.instrument.binder.jms.JmsPublishObservationContext.lambda$new$0(JmsPublishObservationContext.java:40)
	at io.micrometer.tracing.handler.PropagatingSenderTracingObservationHandler.lambda$onStart$0(PropagatingSenderTracingObservationHandler.java:60)
	at brave.propagation.B3Propagation$Format$1.inject(B3Propagation.java:46)
```

**Environment**

 - Micrometer version [e.g. 1.7.1]  1.12.0-M3
 
  - Micrometer registry [e.g. prometheus] prometheus
 
 - OS: [e.g. macOS] macOS
 
 - Java version: [e.g. output of `java -version`]
 
 openjdk 17.0.6 2023-01-17
OpenJDK Runtime Environment Temurin-17.0.6+10 (build 17.0.6+10)
OpenJDK 64-Bit Server VM Temurin-17.0.6+10 (build 17.0.6+10, mixed mode, sharing)

**To Reproduce**
How to reproduce the bug:

Use Spring Boot 3.2.0-M3 with artemis-starter and tracing enabled. Publish a message.

**Expected behavior**
Escape / Unescape Property Names to comply with JMS Java Identifier Rules
https://camel.apache.org/components/4.0.x/jms-component.html#_message_header_mapping

**Additional context**

tw-atroehrsm:
https://issues.apache.org/jira/browse/ARTEMIS-2700 some issue from 2020 where someone stumbled upon this.

tw-atroehrsm:
Otherwise I would be happy if the escaping would just work within the SB3 world.. as thats the only place we use JMS

shakuzen:
> Im not sure what happend if we switch from B3-* to b3 single header. Any experiences with migration?

[...]

PR #4203, opened 2023-10-10
Escape/Unescape JMS String properties in contexts

bclozel:
Prior to this commit, both `JmsPublishObservationContext` and `JmsProcessObservationContext` would set propagation headers as String properties on the JMS `Message`.
This would cause issues with JMS implementations, as the JMS specification only allows valid Java identifiers as keys. Typically, a `X-B3-TraceId` header fails this requirement as it contains "-" characters.

This commit ensures that the following characters are escaped when publishing messages, and unescaped when receiving messages:

* "-" becomes "_HYPHEN_"
* "." becomes "_DOT_"

See gh-4202

### DIFF
--- a/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsProcessObservationContext.java
+++ b/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsProcessObservationContext.java
@@ -26,7 +26,10 @@ import jakarta.jms.Message;
  * messages}.
  * <p>
  * The inbound tracing information is propagated by looking it up in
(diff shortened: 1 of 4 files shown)


bclozel:
Some of the logic here is backwards, I need to update the PR before it gets reviewed or merged.

bclozel:
It's all good now. Note that this change is only required when tracing headers contain "-" characters, like `X-B3-TraceId`. W3C headers are not impacted.

jonatan-ivanov (approved):

bclozel:
Closing because of https://github.com/micrometer-metrics/micrometer/issues/4202#issuecomment-1757183048

PR #4247, opened 2023-10-16
[Dynatrace v2] Only serialize metadata when set on the meter

pirgeo:
(Bugfix)

In #4006, we added support for metrics metadata (unit/description). While working on something unrelated, we realized that the exporter will try to serialize metadata to the Dynatrace transport format. When no metadata is set on the meter (e.g. custom metrics introduced by users) the serialization library will log a warning stating that metadata serialization had been called without setting data. This can lead to many log lines.

Since it is perfectly fine to create meters without metadata, the exporter should not attempt to create metadata lines if no metadata was set on the metric.


jonatan-ivanov:
Thank you for the PR!
I simplified it a little and added a test: https://github.com/micrometer-metrics/micrometer/commit/9df4469588570fb8c9c89cf57520eefea7da8440

PR #4262, opened 2023-10-23
Remove obsolete dependency locking configuration

izeye:
This PR removes obsolete dependency locking configuration.


jonatan-ivanov (approved):

jonatan-ivanov:
Thank you!

Issue #4056, opened 2023-09-02
DataDog plugin triggers blocking due to infinite metric metadata update request failures


platinummonkey:
fixed in https://github.com/micrometer-metrics/micrometer/pull/3329 but it seems the original responders have gone completely AWOL. Starting this issue purposefully to raise attention.

shakuzen:
Sorry for the lack of response on the pull request. It's on our radar but we haven't had the bandwidth to review it lately.

As for the title of this issue, could you provide more details? Metadata is sent at the end of each publish for metrics that haven't had metadata successfully sent yet. Where is blocking happening?

platinummonkey:
We block because micrometer logic infinitely spams in a DoS fashion :)

shakuzen:
Thanks for the clarification on the blocking. I don't understand the infinitely part still. Once an application has successfully sent metadata for a metric, it does not send it again for the life of that application. Or at least that's the intended behavior. An attempt to send the same metadata should only happen if it did not get a successful response or if the application is restarted or it's a different instance.

platinummonkey:
Right this can happen from a 429, 404, 403, 401, or 400 it appears to not take into account the upstream response. Regardless back to the other reason for the PR is the current method is inefficient and lossy (in the event of errors) push format compared to dogstatsd backed by the agents retry behavior and optimized intake paths.

shakuzen:
If there's additional error handling that should be done, we'd appreciate details of that. While I understand your preferred solution is the PR, the fact is all users are not going to instantly switch over to that in a new minor version. So if there is a bug we should fix, we'd be interested in helping users (and it sounds like it could alleviate some unnecessary overhead on the Datadog side). The existing direct-to-datadog approach would not entirely go away any time soon even if the PR is merged, after all.

platinummonkey:
I think this can suffice as a bug report for that. Feel free to change the issue description, but not currently interested in solving that issue. It’s non optimized for us, using a very very old intake path, lossy and we actually prefer people switch default to the statsd approach. It also comes with additional distribution support.

Is there a specific reason to hold onto the direct to api approach indefinitely that can be clearly articulated? (Aside from the need to migrate)

For the time remaining I am shifting my attention back to #3329 - I’ll remerge upstream and fix the new conflicts.

platinummonkey:
This is still a problem and affecting customers

platinummonkey:
Still the same bump

PR #4283, opened 2023-10-26
Deprecate DatadogMeterRegistry

platinummonkey:
The replacement option is detailed in https://docs.datadoghq.com/metrics/guide/micrometer/ by other otel/prometheus plugins

Related to https://github.com/micrometer-metrics/micrometer/pull/3329 follow up

### DIFF
--- a/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
+++ b/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
@@ -62,7 +62,8 @@ public class DatadogMeterRegistry extends StepMeterRegistry {
      * @param config Configuration options for the registry that are describable as
      * properties.
      * @param clock The clock to use for timings.
-     * @deprecated Datadog plugin is deprecated, see the Datadog guide https://docs.datadoghq.com/metrics/guide/micrometer/
+     * @deprecated Datadog plugin is deprecated, see the Datadog guide
+     * https://docs.datadoghq.com/metrics/guide/micrometer/
      */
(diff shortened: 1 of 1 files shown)


jonatan-ivanov:
Could you please help me to understand the intent of this? Is the HTTP API of Datadog deprecated? If not, why should the registry be?
I understand that Datadog supports multiple formats/protocols but HTTP is still available for the users, right?

platinummonkey:
Can certainly add some references here!

This registry has some bugs and documented a couple of those issues here: 
- https://github.com/micrometer-metrics/micrometer-docs/pull/225#issuecomment-1647962627
- https://github.com/micrometer-metrics/micrometer/issues/4056

We will be recommending against using the DatadogRegistry here as a result of the above issues. In addition given there are other compatible plugins that support open-source initiatives we contribute to, this makes sense for us to continue to invest in those (like otel). 

The otel and agent prometheus/openmetrics integrations are very well supported integrations and the agent is optimized to handle the failure modes and aggregations appropriate for Datadog customers.


So yes, for now it is still possible to use those HTTP endpoints directly, but we won't be continuing to support this integration directly as noted above. If you wish to keep it and close this, I'd probably recommend of the top of my mind: 
- fixing the issue of infinite loops PUTs to the metric metadata endpoints
- gracefully buffering and handling any failures at the API level to not drop data
- developing a way to exclude certain tags and exclude certain metrics for customers that are cost-conscious without requiring them make a code change.

jonatan-ivanov:
Let me close this since we are still planning to look into using Dogstatsd in the `DatadogMeterRegistry` sometimes later (as far as I remember there were some issues in that implementation that should be fixed).

Issue #4218, opened 2023-10-11
ObservationGrpcServerInterceptor does not store observation in a context


AleksanderBrzozowski:
**Please describe the feature request.**
Let's say that we have a gRPC server written in kotlin with coroutines support enabled:

```kotlin
class ExampleServer(private val registry: ObservationRegistry) :ServiceGrpcKt.CoroutineImplBase() {
    override suspend fun exampleMethod(request: Request): Response {
        // logic
    }
```

And let's assume that `ObservationGrpcServerInterceptor` was added as an interceptor.

When we try to get the current observation from the registry, it always return null, which is expected as suspended function are ran in a different thread:

```kotlin
override suspend fun exampleMethod(request: Request): Response {
    this.registry.currentObservation // returns null
}
```

I wonder how we can fix this, so that we can get the current observation the same as we do for blocking API.

It would be nice if the existing function could be used to retrieve the current observation created by the interceptor:
```kotlin
class ExampleServer(private val registry: ObservationRegistry) :ServiceGrpcKt.CoroutineImplBase() {
    override suspend fun exampleMethod(request: Request): Response {
        coroutineContext.currentObservation()
    }
```

**Rationale**
It is currently not possible to create a child observation based on the parent one that was created in the gRPC Server Interceptor. 

**Additional context**
https://github.com/micrometer-metrics/micrometer/pull/3256

jonatan-ivanov:
So it seems there is a missing instrumentation and not the existing one having some bugs, right?

AleksanderBrzozowski:
@jonatan-ivanov
Yeah, exactly, you can see my PR to find how this might be added 🙂 

I don't think that the discussion you have mentioned is related, though. It seems to be about context propagation to the REST calls, and in my case it is gRPC call 🙂 

I will be happy if you can take a look at the PR that I provided 🙂

marcingrzejszczak:
Thank you @AleksanderBrzozowski for your PR. We will look into that ASAP (we haven't forgotten about your PR :) )

AleksanderBrzozowski:
@marcingrzejszczak Thanks! 🙂 I haven't forgotten either! 😄

AleksanderBrzozowski:
@marcingrzejszczak Hello! :) How is it going? Were you able to take a look at the PR? :)

marcingrzejszczak:
Not yet, the PR however needs some conflict resolution.

AleksanderBrzozowski:
[...]

PR #4284, opened 2023-10-27
add support for Observation propagation in kotlin gRPC coroutine server

AleksanderBrzozowski:
This commit contains changes that allow to capture current observation when using kotlin gRPC coroutine server.

In order to properly propagate current Observation to a coroutine server method, we need to propagate Observation as a context element by extending `CoroutineContextServerInterceptor`.

Moreover, current Observation needs to be somehow allowed to capture by `CoroutineContextServerInterceptor`. To do this, we need to open current Observation scope inside `ObservationGrpcServerInterceptor`. It is important to keep these two interceptors in a proper order - first, we need to open current Observation scope, and later create a context element based on it.

Fixes: #4218

### DIFF
--- a/micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObservationCoroutineContextServerInterceptor.kt
+++ b/micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObservationCoroutineContextServerInterceptor.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2013-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.micrometer.core.instrument.kotlin
+
+import io.grpc.Metadata
+import io.grpc.ServerCall
+import io.grpc.kotlin.CoroutineContextServerInterceptor
+import io.micrometer.observation.ObservationRegistry
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * This interceptor is meant to propagate observation context to a kotlin coroutine gRPC server method.
+ *
+ * Usage:
+ *
+ * ```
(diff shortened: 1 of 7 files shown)


micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObservationCoroutineContextServerInterceptor.kt
```diff
@@ -0,0 +1,48 @@
...
+ *         .intercept(ObservationGrpcServerInterceptor(observationRegistry))
+ *         .build();
+ * server.start()
+ * ```
+ *
+ * Please remember that order of interceptors matters, and it has to be the same as it is in the example above.
+ *
+ * @since ??? TODO should we set this property?
+ */
+class ObservationCoroutineContextServerInterceptor(
```
AleksanderBrzozowski, line 42: I wonder how to document this change, and where to put example code? WDYT?

marcingrzejszczak (approved):
I cannot comment on the Koltin part cause I don't know Kotlin well enough but other than that LGTM

marcingrzejszczak:
cc @ttddyy

micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObservationCoroutineContextServerInterceptor.kt
AleksanderBrzozowski: What about docs? Shouldn't we mention how to set it up for kotlin gRPC services that use coroutines? 🙂

AleksanderBrzozowski: I've set `since` property and made a new commit - I guess that we would squash them before merge? 🙂

micrometer-core/src/test/kotlin/io/micrometer/core/instrument/kotlin/binder/grpc/GrpcCoroutinesTest.kt
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright 2013-2023 the original author or authors.
```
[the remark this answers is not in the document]
AleksanderBrzozowski: Yeah, good point, I've set it to 2024 for all new files 🙂

ttddyy (approved):

AleksanderBrzozowski:
@marcingrzejszczak Could you re-review? 🙏

marcingrzejszczak (approved):

marcingrzejszczak:
Great work @AleksanderBrzozowski , thank you for your contribution!

AleksanderBrzozowski:
Yeah, no worries!

Issue #1433, opened 2019-05-23
Allow Tomcat's metrics to be bound when its MBean registry has been disabled


wilkinsona:
Tomcat 9.0.20 has [introduced](https://bz.apache.org/bugzilla/show_bug.cgi?id=63361) an option to disable its MBean registry. This improves startup time and reduces Tomcat's heap footprint by ~2MB. Spring Boot will [disable the registry by default in 2.2](https://github.com/spring-projects/spring-boot/issues/16498).

A downside of disabling the registry is that a number of the MBeans that are used by `TomcatMetrics` are no longer available. We'll work around this in Boot by documenting that users should enable the registry if they want certain Tomcat metrics to be available. Assuming it's possible, I think it would be better if Micrometer was less reliant on Tomcat's MBeans for its Tomcat metrics. The HTTP session metrics do not rely on an MBean, perhaps some or even all of the others could take a similar approach?

shakuzen:
I would love to not rely on MBeans at all if possible. I think we couldn't find alternatives for the rest. @markt-asf is there a way to retrieve the same information we're getting from the MBeans now? See [TomcatMetrics](https://github.com/micrometer-metrics/micrometer/blob/c80ab3da6b6b4ba386f05c1cc12ee60581f8d010/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java). Specifically, we're using the Tomcat MBeans for ThreadPool, StringCache, Servlet, and GlobalRequestProcessor.

shakuzen:
I would also note, though, that we've had requests from users to go the other direction and get session metrics from an MBean because when deploying as a WAR they don't have access to the `Manager` in application code (see #1620). The maintenance and testing of the code to retrieve from MBeans has been painful, though, and I'd prefer a strongly typed API.

markt-asf:
There are various mechanisms. All involve some form of hack because they require jumping from application code to Tomcat's internals and Tomcat is designed to explicitly block that when running under a `SecurityManager`.

The simplest is probably implementing `ContainerServlet` which will give you access to the `Wrapper` (Tomcat's internal container for a Servlet) and from there you can navigate to wherever you wish. `getParent()` will get you the `Context`, another `getParent()` the `Host` and then the `Engine`. From there you can get to the `Service` and the `Server` which is the root of the object tree.

[...]

PR #4285, opened 2023-10-27
Allow using TomcatMetrics without `Manager` on the compilation classpath

jjank:
See relevant Slack discussion: https://micrometer-metrics.slack.com/archives/C662HUJC9/p1698128026070889

EDIT: I've added Javadoc to the new constructors 👍

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
@@ -68,6 +68,35 @@ public class TomcatMetrics implements MeterBinder, AutoCloseable {
 
     private volatile String jmxDomain;
 
+    /**
+     * Similar to {@link TomcatMetrics#forStandalone(Iterable, MBeanServer)} but more
+     * convenient as it makes use of {@link TomcatMetrics#getMBeanServer()} to get the
+     * default {@link MBeanServer}
+     */
+    public static TomcatMetrics forStandalone(Iterable<Tag> tags) {
+        return new TomcatMetrics(null, tags, getMBeanServer());
+    }
+
+    /**
+     * This factory method is intended for users of a standalone Tomcat instance - i.e.
+     * build a {@code .war} artifact and deploy it into a Tomcat instance by copying it
+     * into {@code $CATALINA_HOME\webapps}
+     * <p>
+     * In this scenario applications usually do not have
+     * {@code org.apache.catalina.Manager} on the (compilation)-classpath
+     * <p>
+     * This particular constructor avoids the need of having
+     * {@code org.apache.catalina.Manager} available at compilation time e.g. by means of
+     * the {@code provided} scope in Maven
+     */
+    public static TomcatMetrics forStandalone(Iterable<Tag> tags, MBeanServer mBeanServer) {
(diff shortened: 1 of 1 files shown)


jonatan-ivanov:
I tested this the following way:
1. Published it to my maven local: `./gradlew pTML`
2. Created an empty Gradle project using `gradle init`
3. Added `mavenLocal()` and `micrometer-core`
4. Wrote a psvm app

build.gradle
```gradle
apply plugin: 'java'

repositories {
    mavenCentral()
    mavenLocal()
}

dependencies {
    implementation 'io.micrometer:micrometer-core:1.12.0-SNAPSHOT'
}
```

Demo.java
```java
MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
MeterRegistry registry = new SimpleMeterRegistry();

try (TomcatMetrics tomcatMetrics = new TomcatMetrics(Tags.empty())) {
	tomcatMetrics.bindTo(registry);
}
try (TomcatMetrics tomcatMetrics = new TomcatMetrics(Tags.empty(), mBeanServer)) {
	tomcatMetrics.bindTo(registry);
}
```

The first ctor (`TomcatMetrics(Iterable<Tag>)`) goes through the compiler without any problems. The second one (`TomcatMetrics(Iterable<Tag>, MBeanServer)`) does not:
```
java: cannot access org.apache.catalina.Manager
  class file for org.apache.catalina.Manager not found
```

I guess because in the first case the compiler can decide which ctor to  call without looking into the others (one argument) while in the second case, it needs to look into the types. I think even if we add the first ctor, it can make this unsafe since future ctors we add can cause compile errors. For example if I add a one-arg ctor with the `Manager` (`TomcatMetrics(Manager)`), the first case above fails too with the same compiler error.

jjank:
Hi @jonatan-ivanov,

my bad I guess I jumped to a conclusion to quickly 🙈 

I agree that it's not a safe option to add any of the two ctors. Additionally it's confusing because the class offers so many constructors.

The only alternative I could think of is static factory methods. Something along the lines of:
```java
public static TomcatMetrics forStandalone(Iterable<Tag> tags, MBeanServer mBeanServer) {
        return new TomcatMetrics(null, tags, mBeanServer);
    }

// It's debatable whether this method is _really_ necessary. Requring users to call `TomcatMetrics::getMBeanServer` seems reasonable
public static TomcatMetrics forStandalone(Iterable<Tag> tags) {
        return new TomcatMetrics(null, tags, getMBeanServer());
    }
```

This would solve some issues:
* Although `forStandalone` specifically is up for discussion but obviously there at least is a chance of naming
* Safer evolvment of the API

I've updated my PR with the static factory methods. But feel free to not accept the PR at all if you think it is not worth the extra API-surface

jjank:
@jonatan-ivanov I've rebased my branch onto main, now the build passes 😅

jonatan-ivanov:
As far as I remember we further discussed this on Slack and the conclusion was that if you add `org.apache.tomcat.embed:tomcat-embed-core` as a `provided` dependency (compile-only), you can reference the `Manager` class which will be provided by Tomcat runtime.

With that, let me close this PR. Also, we are planning to look into providing a JMX-fee/Manager-free implementation of `TomcatMetrics`, see:
- https://github.com/micrometer-metrics/micrometer/pull/1620
- https://github.com/micrometer-metrics/micrometer/issues/1433

Please let s know if my I didn't remember correctly and you want us to reopen this PR.

PR #4286, opened 2023-10-29
Add Git commit and build info metrics

### LINKED ISSUE #3154: Provide binders for Git and Build information
**Description**
There has been some discussion at the Spring Boot project about exposing information like the Git commit ID and Maven coordinates as metrics, similar in style to the existing JvmInfoMetrics binder. For consistency, we would provide the same information that Spring Boot's Actuator exposes on its /actuator/info endpoint.

The binders would receive the data via constructor arguments and add a constant gauge with tags for all attributes we want to expose. If some values are null, we would add a marker string ("unknown", "null", "N/A") to make sure all tags are always present.

Example:

```
git_info{app="some-service", branch="main", id="cafecafe", time="2022-04-26T18:39:18Z"} 1.0
build_info{app="some-service", name="Some Service", artifact="some-service", group="org.example", version="1.7", time="2022-04-28T12:01:07Z"} 1.0
```

This issue is intended to discuss whether this is a worthwhile addition and to discuss the design if the team decides to proceed.

**Rationale**
Adding metrics binders to Micrometer would provide standardization of these metrics across all users of Micrometer, not just Spring Boot. The new metrics would be useful in Grafana dashboards, for example, to display which exact commit/version is currently deployed.

**Additional context**

- [PR 30628](https://github.com/spring-projects/spring-boot/pull/30628) at Spring Boot
- [GitProperties](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/info/GitProperties.html)
- [BuildProperties](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/info/BuildProperties.html)

ardetrick:
Closes https://github.com/micrometer-metrics/micrometer/issues/3154.

I was a bit unsure how closely I should be following the [Spring Boot PR](https://github.com/spring-projects/spring-boot/pull/30628) referenced in the issue. This implementation is more explicit in which attributes are allowed and how they are supplied to the binder. But there could be an alternative approach that builds the info objects using an instance of `Properties`. 

Other naming specific questions I have are:
- Is `unknown` an acceptable default value for the tag? Would you rather it go in a specific class rather than in `MeterBinder`?
- Is `GitInfo` preferable to `GitCommitInfo`?
- Are the tag names inline with expectations?

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/build/BuildInfoMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/build/BuildInfoMetrics.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2026 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.micrometer.core.instrument.binder.build;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+import java.time.Instant;
+
+import static java.util.Collections.emptyList;
+import static java.util.Objects.requireNonNull;
+
+public class BuildInfoMetrics implements MeterBinder {
+
+    private final BuildInfo buildInfo;
+
+    private final Iterable<Tag> tags;
+
+    public BuildInfoMetrics(BuildInfo buildInfo) {
+        this(buildInfo, emptyList());
+    }
+
+    public BuildInfoMetrics(BuildInfo buildInfo, Iterable<Tag> tags) {
+        requireNonNull(buildInfo, "buildInfo");
+        requireNonNull(tags, "tags");
+        this.buildInfo = buildInfo;
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        requireNonNull(registry, "registry");
+        Gauge.builder("build.info", () -> 1L)
+            .description("Build information")
+            .strongReference(true)
+            .tag("group", buildInfo.getGroup().orElse(DEFAULT_TAG_VALUE))
+            .tag("artifact", buildInfo.getArtifact().orElse(DEFAULT_TAG_VALUE))
+            .tag("name", buildInfo.getName().orElse(DEFAULT_TAG_VALUE))
+            .tag("version", buildInfo.getVersion().orElse(DEFAULT_TAG_VALUE))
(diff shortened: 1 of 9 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/build/BuildInfoMetrics.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.time.Instant;
+
+import static java.util.Objects.requireNonNull;
+
+public class BuildInfoMetrics implements MeterBinder {
+
+    private final BuildInfo buildInfo;
+
+    public BuildInfoMetrics(BuildInfo buildInfo) {
```
lenin-jaganathan, line 31: May be one more constructor to add user provided tags to metrics?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/build/BuildInfoMetrics.java
```diff
@@ -0,0 +1,50 @@
...
+    public void bindTo(MeterRegistry registry) {
+        requireNonNull(registry, "registry");
+        Gauge.builder("build.info", () -> 1L)
+            .description("Build information")
+            .strongReference(true)
+            .tag("group", buildInfo.getGroup().orElse(DEFAULT_TAG_VALUE))
+            .tag("artifact", buildInfo.getArtifact().orElse(DEFAULT_TAG_VALUE))
+            .tag("name", buildInfo.getName().orElse(DEFAULT_TAG_VALUE))
+            .tag("version", buildInfo.getVersion().orElse(DEFAULT_TAG_VALUE))
+            .tag("timestamp", buildInfo.getTimestamp().map(Instant::toString).orElse(DEFAULT_TAG_VALUE))
```
lenin-jaganathan, line 46: My initial thoughts on the `timestamp` was whether it is a good candidate for metrics. But, I guess it should be fine since it will most probably be tied to versions and timestamp is more of an informative tag than anything.

lenin-jaganathan: But is this really for tracking build info of dependencies? or the service itself?

ardetrick: I think it could be used for either the service and/or one or many dependencies. This implementation does not impose an opinion. Ultimately it was added to try and meet parity with the [SpringBoot](https://github.com/spring-projects/spring-boot/blob/dbe882aced8656b41d673b2542256c227bc528cd/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/buildinfo/BuildInfoProperties.java#L100) equivalent class.

One use case for a dependency would be to use this metric to assert that "all production services are using SpringBoot version 3.1.x or a version with a `timestamp` value from the last 3 months."

One could case for the service might be to use this metric in combination with something like "deployment.info" (which does not exist in Micrometer right now, but for the purposes of this discussion, lets say it captures at least a `timestamp` value of when the runtime was deployed). You could then use these two metrics together to understand how long it takes to get an application into production starting from the time the deployable artifact was created. This example is a bit more complicated and has it's flaws (such as rollbacks, rolling deployments, auto-scaling) but shared it as a potential idea.

Ultimately I agree with you, the timestamp value may end up being more informational than anything but I could imagine use cases for it to be useful and it meets SpringBoot feature parity..

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/build/BuildInfoMetrics.java
ardetrick: This has been [added](https://github.com/micrometer-metrics/micrometer/pull/4286/commits/c467a76710dc0cba22c59596b244737264ab9142).

PR #4315, opened 2023-11-04
Add support for `AutoShutdownDelegatedExecutorService`

nscuro:
This executor is returned by `Executors#newSingleThreadExecutor` under Java 21.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
@@ -296,7 +296,8 @@ public class ExecutorServiceMetrics implements MeterBinder {
             if (className.equals("java.util.concurrent.Executors$DelegatedScheduledExecutorService")) {
                 monitor(registry, unwrapThreadPoolExecutor(executorService, executorService.getClass()));
             }
-            else if (className.equals("java.util.concurrent.Executors$FinalizableDelegatedExecutorService")) {
+            else if (className.equals("java.util.concurrent.Executors$FinalizableDelegatedExecutorService")
+                    || className.equals("java.util.concurrent.Executors$AutoShutdownDelegatedExecutorService")) {
                 monitor(registry,
                         unwrapThreadPoolExecutor(executorService, executorService.getClass().getSuperclass()));
             }


shakuzen (changes requested):
Thank you for the pull request. Would you please add tests that verify the metrics work with Executors of the new type added?

jonatan-ivanov:
@nscuro Could you please add some tests?

shakuzen:
I went ahead and added a test for it (to our micrometer-java21 module's tests since it requires JDK 21+):

```java
// See gh-4315
@Test
void threadPoolMetricsWith_AutoShutdownDelegatedExecutorService() throws InterruptedException {
    CountDownLatch latch = new CountDownLatch(1);
    ExecutorService unmonitored = Executors.newSingleThreadExecutor();
    assertThat(unmonitored.getClass().getName()).isEqualTo("java.util.concurrent.Executors$AutoShutdownDelegatedExecutorService");
    ExecutorService monitored = ExecutorServiceMetrics.monitor(registry, unmonitored, "test");
    monitored.execute(latch::countDown);
    assertThat(latch.await(100, TimeUnit.MILLISECONDS)).isTrue();
    assertThat(registry.get("executor.completed")
        .tag("name", "test")
        .functionCounter()
        .count()).isEqualTo(1L);
}
```

The test still fails; just the warning logged is different. So I don't think this PR effectively changes anything, unless I'm missing something.

```
Jan 08, 2025 3:40:38 PM io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics unwrapThreadPoolExecutor
INFO: Cannot unwrap ThreadPoolExecutor for monitoring from java.util.concurrent.Executors$DelegatedExecutorService due to java.lang.reflect.InaccessibleObjectException: Unable to make field private final java.util.concurrent.ExecutorService java.util.concurrent.Executors$DelegatedExecutorService.e accessible: module java.base does not "opens java.util.concurrent" to unnamed module @60985c7c

io.micrometer.core.instrument.search.MeterNotFoundException: Unable to find a meter that matches all the requirements at once. Here's what was found:
   FAIL: No meter with name 'executor.completed' was found.
   FAIL: No meters with type function counter were found.
   FAIL: No meters have the required tag 'name'.
```

Given the above, I'll close this PR. If we have another way of enabling the additional metrics, let's explore that, but I'm not sure it's worth changing the logged warning from one thing to another. Let me know if I've missed anything.

izeye:
`--add-opens java.base/java.util.concurrent=ALL-UNNAMED` seems to be required for this change to work.

I created https://github.com/micrometer-metrics/micrometer/pull/5811 on top of this PR and the test that @shakuzen wrote above.

Issue #3846, opened 2023-05-18
Current step data is not published on shutdown when publishing is in progress


lenin-jaganathan:
**Describe the bug**
https://github.com/micrometer-metrics/micrometer/issues/1882 was fixed by introducing the StepMeter Contract and doing a closing rollover. This fixed the original issue but https://github.com/micrometer-metrics/micrometer/pull/3750 introduced a polling thread that rolls the step meters at the start of the step. These 2 fixes combined produce a grey area where publishing is happening at the middle of the step and the registry closes. In this scenario, the data accumulated for the current step will be lost and will be never reported.

**Environment**
 - Micrometer version - 1.11.0
 - Micrometer registry - StepMeterRegistry

**To Reproduce**
How to reproduce the bug:
Use any Step registry for testing,
- start the registry and create an increment counter every second.
- close the registry exactly during publishing.

If the step is 60 seconds and the counter is incremented every second, it will report a value of 60 for every step. If the registry publishes at the 30th second and closes also happens at the 30th second, then publishing will be skipped as there is already publishing that is in progress. Also, the closingRollOver will be skipped as the data is already getting published. The publish will report a value of 60 (the value for the last step) and the data for the current step will never be published.

**Expected behavior**
Data for the current step should be published on close and publish happening simultaneously.

jonatan-ivanov:
Is this the same thing we are discussing here: https://github.com/micrometer-metrics/micrometer/issues/3832?

lenin-jaganathan:
I believe they are related.

PR #4318, opened 2023-11-06
Wait for in-progress publish before closing step rollover

shakuzen:
If a scheduled publish is in progress when `close` is called, `StepMeterRegistry` was not performing the closing rollover and was not doing a subsequent publish of the final partial step.

Wait for the in-progress publish to finish before the closing rollover is called, which is followed by a final publish of the partial step (in `PushMeterRegistry#close`).

Fixes gh-3846

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
@@ -243,8 +243,8 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
     @Override
     public void close() {
         stop();
-        if (!isPublishing() && isDelta()) {
-            if (!isDataPublishedForCurrentStep()) {
+        if (config.enabled() && isDelta() && !isClosed()) {
+            if (!isDataPublishedForCurrentStep() && !isPublishing()) {
                 // Data was not published for the current step. So, we should flush that
                 // first.
                 try {
@@ -255,6 +255,9 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
                             "Unexpected exception thrown while publishing metrics for " + getClass().getSimpleName(),
                             e);
                 }
+            }
+            else if (isPublishing()) {
+                waitForInProgressScheduledPublish();
             }
             getMeters().forEach(this::closingRollover);
         }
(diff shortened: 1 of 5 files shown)


lenin-jaganathan:
I am afraid OTLPMeterRegistry will also need this fix copied over.

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
```diff
@@ -136,7 +136,12 @@ public void close() {
         super.close();
     }
 
-    private void waitForInProgressScheduledPublish() {
+    /**
+     * Wait until scheduled publishing by {@link PushMeterRegistry} completes, if in
```
jonatan-ivanov, line 140: Should we add a note that this should not be called in most of the cases by the users?

jonatan-ivanov (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/push/PushMeterRegistry.java
shakuzen: Only implementations of `PushMeterRegistry` should have access to call it, so I didn't think about mentioning it. Implementations may need to call it like StepMeterRegistry (or OtlpMeterRegistry). What do you think?

jonatan-ivanov: Sounds reasonable. 👍🏼

shakuzen:
> I am afraid OTLPMeterRegistry will also need this fix copied over.

I was planning to do it in a follow-up pull request, but I've gone ahead and included the changes in this pull request. Please take a look. Thanks.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -243,8 +243,8 @@ protected DistributionStatisticConfig defaultHistogramConfig() {
     @Override
     public void close() {
         stop();
-        if (!isPublishing() && isDelta()) {
-            if (!isDataPublishedForCurrentStep()) {
+        if (config.enabled() && isDelta() && !isClosed()) {
```
shakuzen, line 246: These aren't directly related to this change but were out-of-sync with logic in StepMeterRegistry which also applies to OTLP.

lenin-jaganathan: This is going to be problem given this is a duplicate code being copied around.

shakuzen: I agree, which is why I wanted to avoid it in the first place, but unfortunately things aren't so easy. We'll have to look again at what we can do to remove the need for the duplication without opening up too much, but that's definitely out of scope for this PR.

Issue #4324, opened 2023-11-09
Support Jetty 12 in JettyConnectionMetrics 


shakuzen:
See #4261. Currently, if JettyConnectionMetrics is used with Jetty 12, a `NoClassDefFoundError` will happen whenever the `onClosed` callback of the Connection Listener is called.

```
INFO  o.e.j.i.AbstractConnection Failure while notifying listener imcibj.JettyConnectionMetrics@7786a3c6{STARTED} java.lang.NoClassDefFoundError: org/eclipse/jetty/server/HttpConnection
```

https://github.com/micrometer-metrics/micrometer/blob/214f038e45645b69bf1609cb5a486716c3238863/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java#L165-L170

We are trying to determine whether the connection instrumented is a server or client connection, and we are doing that with an instanceof check against `HttpConnection`, which is in the `jetty-server` artifact. This is not very robust as there are other server implementations of `Connection`. It also causes the above mentioned error in conjunction with the relocation of the `HttpConnection` class to a different package.

PR #4325, opened 2023-11-09
Improve `JettyConnectionMetrics` connection type detection

shakuzen:
The `HttpConnection` class was relocated to a different package in Jetty 12 which was causing a `NoClassDefFoundError` when doing an instanceof check on it. The goal of it is to determine whether the connection instrumented is a server connection or a client connection. However, it specifically checks against `HttpConnection` and assumes that all server connections will be an instanceof `HttpConnection` and any not an instanceof it are client connections. This is brittle because Jetty supports more implementations of `Connection` on the server side than `HttpConnection` and there could in theory be an arbitrary implementation where we do not know whether it is a server or client connection (it could also be neither).

This instead checks whether the package name contains `server` or `client` or neither. This is admittedly also brittle, but given the known implementations of `Connection` provided by the Jetty project, this pattern generally seems to hold, and it is at least more honest using `UNKNOWN` when our heuristic fails. It also avoids the `NoClassDefFoundError` when using `JettyConnectionMetrics` with Jetty 12.

Resolves gh-4324


micrometer-jetty11/build.gradle
```diff
@@ -2,11 +2,7 @@ description 'Micrometer instrumentation for Jetty 11'
 
 dependencies {
     api project(":micrometer-core")
-    api(libs.jetty11Server) {
-        version {
-            strictly libs.jetty11Server.get().version
```
shakuzen, line 7: This was no longer needed since I removed the jetty dependencies from `dependencies.gradle` in lieu of referring to the type-safe version catalog everywhere for jetty.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -162,10 +161,16 @@ public void onClosed(Connection connection) {
         }
 
         if (sample != null) {
-            String serverOrClient = connection instanceof HttpConnection ? "server" : "client";
+            String type = "UNKNOWN";
+            if (connection.getClass().getName().contains("server")) {
```
shakuzen, line 165: We may want to consider switching to make the qualified type name the value of the `type` tag, but that would be breaking for anyone that relies on the current values. Hence I thought it better to consider that in a future version. This change should keep the prior values in most common cases, I hope.

shakuzen:
The jdk11 build is failing because Jetty 12 requires Java 17. We could skip the jetty12 sample on jdk <17 or we could remove the jdk11 build.

jonatan-ivanov (approved):

Issue #4100, opened 2023-09-26
Deprecate Jersey server instrumentation in favour of the jersey-micrometer module in Jersey


marcingrzejszczak:
https://github.com/eclipse-ee4j/jersey/pull/5391

senivam:
question - why Micrometer does not support ``@Counted`` annotation for Jersey? As far as I see, there is only ``@Timed`` annotation processing for ``MetricsRequestEventListener``.

shakuzen:
@senivam I think it was just missed to add that support to the Jersey instrumentation. Since we are deprecating the code here in favor of the code in the Jersey repo, please open an issue there and we can see about adding support for that.

shakuzen:
Note that only Jersey server instrumentation was added to the jersey-micrometer module. We need to add the Jersey client instrumentation there before deprecating that.

PR #4326, opened 2023-11-09
Add missing metrics for Jersey in cases of client errors

adpaste:
We are using Jersey integration and we have noticed that some 40x errors (such as `406 Not Acceptable` or `415 Unsupported Media Type`) are not reported in the metrics.

After some debugging, we found out that the root cause of this issue is that only one kind of client exception (`NotFoundException`) is taken into account during metric collection.

This pull request should fix the problem, by taking into account all types of `ClientErrorExceptions`.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListenerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListenerTest.java
@@ -27,6 +27,7 @@ import org.junit.jupiter.api.Test;
 
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.core.Application;
+import javax.ws.rs.core.MediaType;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -148,6 +149,11 @@ class MetricsRequestEventListenerTest extends JerseyTest {
         }
         catch (Exception ignored) {
         }
+        try {
+            target("produces-text-plain").request(MediaType.APPLICATION_JSON).get();
+        }
+        catch (Exception ignored) {
+        }
 
         assertThat(registry.get(METRIC_NAME)
             .tags(tagsFrom("/throws-exception", "500", "SERVER_ERROR", "IllegalArgumentException"))
@@ -161,6 +167,11 @@ class MetricsRequestEventListenerTest extends JerseyTest {
 
         assertThat(registry.get(METRIC_NAME)
             .tags(tagsFrom("/throws-mappable-exception", "410", "CLIENT_ERROR", "ResourceGoneException"))
+            .timer()
+            .count()).isEqualTo(1);
+
+        assertThat(registry.get(METRIC_NAME)
+            .tags(tagsFrom("root", "406", "CLIENT_ERROR", "NotAcceptableException"))
             .timer()
             .count()).isEqualTo(1);
     }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListener.java
@@ -73,7 +73,7 @@ public class MetricsRequestEventListener implements RequestEventListener {
 
         switch (event.getType()) {
             case ON_EXCEPTION:
-                if (!isNotFoundException(event)) {
+                if (!isClientError(event)) {
                     break;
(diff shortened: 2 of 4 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListenerTest.java
```diff
@@ -163,6 +169,11 @@ void exceptionsAreMappedCorrectly() {
             .tags(tagsFrom("/throws-mappable-exception", "410", "CLIENT_ERROR", "ResourceGoneException"))
             .timer()
             .count()).isEqualTo(1);
+
+        assertThat(registry.get(METRIC_NAME)
+            .tags(tagsFrom("root", "406", "CLIENT_ERROR", "NotAcceptableException"))
```
shakuzen, line 174: The URI being `root` seems incorrect. Shouldn't we be able to capture `produces-text-plain` here?

shakuzen (changes requested):
Thanks for reporting the issue and providing a fix along with a test. Could you take a look at the one comment I had and rebase the pull request on `1.9.x` so we can accept it as a bug fix? Also, note that the Jersey instrumentation was added to a jersey-micrometer module in the Jersey repo (see https://github.com/eclipse-ee4j/jersey/pull/5391). We plan to deprecate the code in the Micrometer repo in favor of that (see #4100). So please send a pull request over there too if you have the time and consider migrating to using that over what will be deprecated here.

Thanks again!

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListenerTest.java
adpaste: It seems that when the requested accepted media type is different than the media type supported by the endpoint, the `uri` is not matched by Jersey and by default is resolved to `root`.

We could probably change this behavior [here](https://github.com/micrometer-metrics/micrometer/blob/f5f3e013516172948e497769daa48496fc78224f/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jersey/server/JerseyTags.java#L97), but I'm not sure what kind of `uri` would be more desired in such case.

adpaste:
Thanks for the feedback. I replied to the comment, and rebased the PR to `1.9.x`.
I'll look into migrating to `jersey-micrometer`.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListenerTest.java
shakuzen: I'm not seeing a good way we can get the URI template we want since the request was not matched to a resource. In that case, I think it would be best to use the value `UNKNOWN` for the `uri` tag rather than `root`. Using `root` makes it seem like the request was for a resource mapped to the root path, but it isn't necessarily and in the test case it definitely isn't.

shakuzen: @adpaste would you be willing to update the PR to make sure `UNKNOWN` is used for the `uri` tag instead of `root` in this case?

adpaste: @shakuzen Sure, I've updated the PR with the proposed change. Since I'm not fully familiar with the codebase, could you please confirm if this approach looks alright to you?

shakuzen (approved):

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jersey/server/MetricsRequestEventListenerTest.java
shakuzen: Thanks. Looks good to me.

shakuzen:
It looks like this broke tests in the `1.11.x` branch. I'm looking into it now.
https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/6107/workflows/6c026dde-bd49-4bb1-a8d7-bdb0f07168b3/jobs/30900

shakuzen:
I was able to fix it by applying similar changes to JerseyKeyValues (which is a copy of JerseyTags for Observation-based instrumentation). See https://github.com/micrometer-metrics/micrometer/commit/66c1d4772019146b47ac27e4bcdd324000cedc7f

PR #4409, opened 2023-11-29
Update samples to Spring Boot 2.7.18

izeye:
This PR updates samples to Spring Boot 2.7.18.

### DIFF
--- a/samples/micrometer-samples-boot2-reactive/build.gradle
+++ b/samples/micrometer-samples-boot2-reactive/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'java'
-    id 'org.springframework.boot' version '2.7.16'
+    id 'org.springframework.boot' version '2.7.18'
 }
 
 apply plugin: 'io.spring.dependency-management'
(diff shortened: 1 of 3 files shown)


izeye:
Dependabot [handles Gradle Enterprise Gradle plugin](https://github.com/micrometer-metrics/micrometer/pull/4109) but doesn't seem to handle other Gradle plugins like this and https://github.com/micrometer-metrics/micrometer/pull/4410.

shakuzen:
Unfortunately dependabot's support of the Java ecosystem and Gradle is limited in multiple ways. This might be another such limitation. It may work if we migrate the plugin declaration to reference something in the toml version catalog.

shakuzen:
That said, in regards to the spring samples, I think we should rather work to remove them from the micrometer repo and make sure anything we want covered is included in the micrometer-samples or other repo. Having them in the micrometer repo makes a circular dependency, which isn't the end of the world for a sample, but it's also unnecessary if we can do the same thing in a samples repo.

jonatan-ivanov:
Let me merge this so that we are using the latest version and then we can move them to the samples repo.

jonatan-ivanov:
Thank you!

PR #4483, opened 2023-12-08
Avoid IllegalArgumentException on DefaultLongTaskTimer when percentiles is empty

### LINKED ISSUE #4482: DefaultLongTaskTimer class throws an IllegalArgumentException when percentiles is empty
**Describe the bug**
The `DefaultLongTaskTimer` class throws an `IllegalArgumentException` when `double[] percentiles` is not `null`, but it is empty.

**Environment**
 - Micrometer version [1.11.6, 1.12.0]
 - Micrometer registry [prometheus]
 - OS: [macOS]
 - Java version: [openjdk 17.0.5 2022-10-18]

**To Reproduce**
How to reproduce the bug:

```java
@Test
public void testWhenPercentilesEmptyThrowsIllegalArgumentException() {
    final var id = new Meter.Id("my.metric", Tags.empty(), "seconds", null, Type.LONG_TASK_TIMER);
    final var config = DistributionStatisticConfig.builder()
            .percentiles(new double[] {})
            .build();

    final var timer = new DefaultLongTaskTimer(id, Clock.SYSTEM, TimeUnit.SECONDS, config, true);
    Assertions.assertThrowsExactly(IllegalArgumentException.class, () -> {
        timer.takeSnapshot();
    });
}
```

Also, in a Spring Boot 3.1.6 environment, if you set the metrics distribution to an empty configuration, the error throws when accessing `/actuator/prometheus`. (Empty configuration is useful when you have a default configuration and, for a specific environment you want it to be `null`).

```
management.metrics.distribution.percentiles.http.server.requests=
```

**Expected behavior**

Considering the actual implementation of `DefaultLongTaskTimer`, I believe that it was intended to handle empty arrays nicely, the same way when it is `null`.

```java
Queue<Double> percentilesRequested = new ArrayBlockingQueue<>(
        distributionStatisticConfig.getPercentiles() == null ? 1
                : distributionStatisticConfig.getPercentiles().length);
double[] percentilesRequestedArr = distributionStatisticConfig.getPercentiles();
if (percentilesRequestedArr != null && percentilesRequestedArr.length > 0) {
    Arrays.stream(percentilesRequestedArr).sorted().boxed().forEach(percentilesRequested::add);
}
```

marcelopbarros:
closes #4482

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
@@ -128,10 +128,10 @@ public class DefaultLongTaskTimer extends AbstractMeter implements LongTaskTimer
 
     @Override
     public HistogramSnapshot takeSnapshot() {
-        Queue<Double> percentilesRequested = new ArrayBlockingQueue<>(
-                distributionStatisticConfig.getPercentiles() == null ? 1
-                        : distributionStatisticConfig.getPercentiles().length);
         double[] percentilesRequestedArr = distributionStatisticConfig.getPercentiles();
+        Queue<Double> percentilesRequested = new ArrayBlockingQueue<>(
+                percentilesRequestedArr == null || percentilesRequestedArr.length == 0 ? 1
+                        : distributionStatisticConfig.getPercentiles().length);
         if (percentilesRequestedArr != null && percentilesRequestedArr.length > 0) {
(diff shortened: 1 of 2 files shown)


shakuzen (commented):
Thank you for the pull request. The change looks good to me. Would you also add a test for this case? To fix the build failing due to format, you can run `./gradlew format` and it should fix the formatting for you.

shakuzen:
Once those small things are addressed, I think this is ready to merge. We'll want to rebase and target the `1.10.x` branch to fix this in the oldest OSS supported maintenance branch and we can merge forward from there.

marcelopbarros:
> Would you also add a test for this case?

Sure, I've just added.

Issue #4492, opened 2023-12-14
`micrometer-registry-statsd` POM file is invalid


nathankooij:
**Describe the bug**
The newly published POM files for the `io.micrometer:micrometer-registry-statsd` are invalid, as they are missing `version` declarations for the `io.netty:netty-transport-native-epoll` artifacts (for both classifiers). This means in build errors, like:

```
[WARNING] Unable to obtain POM for artifact: io.micrometer:micrometer-registry-statsd:jar:1.12.1:compile
org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for io.netty:netty-transport-native-epoll:jar:linux-aarch_64 is missing. @ line 34, column 17
[ERROR] 'dependencies.dependency.version' for io.netty:netty-transport-native-epoll:jar:linux-x86_64 is missing. @ line 40, column 17
```

The diff between [1.12.0](https://search.maven.org/artifact/io.micrometer/micrometer-registry-statsd/1.12.0/jar) and [1.12.1](https://search.maven.org/artifact/io.micrometer/micrometer-registry-statsd/1.12.1/jar):

```diff
--- 1.12.0	2023-12-14 10:09:05
+++ 1.12.1	2023-12-14 10:09:18
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.micrometer</groupId>
   <artifactId>micrometer-registry-statsd</artifactId>
-  <version>1.12.0</version>
+  <version>1.12.1</version>
   <name>micrometer-registry-statsd</name>
   <description>Application monitoring instrumentation facade</description>
   <url>https://github.com/micrometer-metrics/micrometer</url>
@@ -28,20 +28,18 @@
     <dependency>
       <groupId>io.micrometer</groupId>
       <artifactId>micrometer-core</artifactId>
-      <version>1.12.0</version>
+      <version>1.12.1</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-transport-native-epoll</artifactId>
-      <version>4.1.101.Final</version>
       <classifier>linux-aarch_64</classifier>
       <scope>runtime</scope>
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-transport-native-epoll</artifactId>
-      <version>4.1.101.Final</version>
[...]

PR #4493, opened 2023-12-14
Remove shaded netty dependency from POM

shakuzen:
The dependency is shaded and so it should not be in the POM. We missed excluding this one when it was added in gh-3605.

Fixes gh-4492

Before:
```xml
  <dependencies>
    <dependency>
      <artifactId>micrometer-core</artifactId>
      <version>1.9.18-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-transport-native-epoll</artifactId>
      <classifier>linux-aarch_64</classifier>
      <scope>runtime</scope>
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-transport-native-epoll</artifactId>
       <classifier>linux-x86_64</classifier>
       <scope>runtime</scope>
     </dependency>
    <dependency>
      <groupId>com.google.code.findbugs</groupId>
      <artifactId>jsr305</artifactId>
      <version>3.0.2</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
  </dependencies>
```

After:
```xml
  <dependencies>
    <dependency>
      <artifactId>micrometer-core</artifactId>
      <version>1.9.18-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>com.google.code.findbugs</groupId>
      <artifactId>jsr305</artifactId>
      <version>3.0.2</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
  </dependencies>
```

### DIFF
--- a/implementations/micrometer-registry-statsd/build.gradle
+++ b/implementations/micrometer-registry-statsd/build.gradle
@@ -58,7 +58,7 @@ publishing {
                 asNode()
                     .dependencies
                     .dependency
-                    .findAll { ['reactor-core', 'reactor-netty-core'].contains(it.artifactId.text()) }
+                    .findAll { ['reactor-core', 'reactor-netty-core', 'netty-transport-native-epoll'].contains(it.artifactId.text()) }
                     .each { it.parent().remove(it) }
             }
         }


shakuzen:
I'd like to verify that UDS protocol continues to work on Linux arm64 with this change. I expect it should, but we don't run CI on Linux arm64.

shakuzen:
I tested using a sample app running in a container for linux aarch64 with the patched version of the statsd jar. I used netcat to check UDS datagrams received from the application.

PR #4496, opened 2023-12-18
Use LongAdder in CumulativeTimer and CumulativeDistributionSummary

izeye:
This PR changes to use `LongAdder` in `CumulativeTimer` and `CumulativeDistributionSummary` like other implementations (ex. `PrometheusTimer`).


shakuzen (approved):
Thanks. I noticed this as well when looking at your other pull request. I'm not sure why these were different but it does seem better to use the Adder classes.

PR #4499, opened 2023-12-19
Add Meter for process CPU time

### LINKED ISSUE #4498: Add Meter for process CPU time
I would like to have a meter for the accumulated CPU time used by a JVM process.

**Rationale**
We use the accumulated CPU time in derived metrics; e.g., we derive the percentage of time spent with garbage collection from the accumulated CPU time and the GC times.

**Additional context**
We currently add the meter in our applications, but we think it would be generally useful.

simonzn:
Closes #4498

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
@@ -18,6 +18,7 @@ package io.micrometer.core.instrument.binder.system;
 import io.micrometer.common.lang.NonNullApi;
 import io.micrometer.common.lang.NonNullFields;
 import io.micrometer.common.lang.Nullable;
+import io.micrometer.core.instrument.FunctionCounter;
 import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
@@ -72,6 +73,9 @@ public class ProcessorMetrics implements MeterBinder {
     @Nullable
     private final Method processCpuUsage;
 
+    @Nullable
+    private final Method processCpuTime;
+
     public ProcessorMetrics() {
         this(emptyList());
     }
@@ -83,6 +87,7 @@ public class ProcessorMetrics implements MeterBinder {
         Method getCpuLoad = detectMethod("getCpuLoad");
         this.systemCpuUsage = getCpuLoad != null ? getCpuLoad : detectMethod("getSystemCpuLoad");
         this.processCpuUsage = detectMethod("getProcessCpuLoad");
+        this.processCpuTime = detectMethod("getProcessCpuTime");
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
```diff
@@ -83,6 +91,14 @@ public ProcessorMetrics(Iterable<Tag> tags) {
         Method getCpuLoad = detectMethod("getCpuLoad");
         this.systemCpuUsage = getCpuLoad != null ? getCpuLoad : detectMethod("getSystemCpuLoad");
         this.processCpuUsage = detectMethod("getProcessCpuLoad");
+        this.processCpuTime = detectMethod("getProcessCpuTime");
+        if (Objects.equals(System.getProperty("com.ibm.lang.management.OperatingSystemMXBean.isCpuTime100ns"),
```
marcingrzejszczak, line 95: I'm not an expert in this, but why are we assuming that this is `true` by default?

simonzn: No, the default unit is 1 ns, but may be changed to 100 ns via this system property. See https://eclipse.dev/openj9/docs/api/jdk11/jdk.management/com/ibm/lang/management/OperatingSystemMXBean.html#getProcessCpuTime()

marcingrzejszczak:
Can you please fix the build? I think there are issues with checkstyle

simonzn:
It seems the checkstyle error was fixed on main; rebased. Afterwards tests failed with a `ClassCastException` - fixed as well.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
```diff
@@ -83,6 +89,13 @@ public ProcessorMetrics(Iterable<Tag> tags) {
         Method getCpuLoad = detectMethod("getCpuLoad");
         this.systemCpuUsage = getCpuLoad != null ? getCpuLoad : detectMethod("getSystemCpuLoad");
         this.processCpuUsage = detectMethod("getProcessCpuLoad");
+        this.processCpuTime = detectMethod("getProcessCpuTime");
+        if (Boolean.getBoolean("com.ibm.lang.management.OperatingSystemMXBean.isCpuTime100ns")) {
```
marcingrzejszczak, line 93: Are any other properties we should check here?

simonzn: According to the documentation of [com.sun.management.OperatingSystemMXBean](https://docs.oracle.com/javase/8/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getProcessCpuTime--) and [com.ibm.lang.management.OperatingSystemMXBean](https://www.ibm.com/docs/api/v1/content/SSYKE2_8.0.0/openj9/api/jdk8/jre/management/extension/com/ibm/lang/management/OperatingSystemMXBean.html#getProcessCpuTime--) this is the only property that relates to `getCpuTime()`. I checked the documentation from Java 8 to 21.

marcingrzejszczak: That's for the IBM's SDK, I'm wondering if there aren't any other that provide something similar

simonzn: Not sure how to provide a conclusive answer here.
* I suppose all OpenJDK builds provide com.sun.management.OperatingSystemMXBean, which always returns ns.
* We already covered OpenJ9 above.
* GraalVM implemented substitutions which returned ns on POSIX-OSes and threw an `UnsupportedFeatureError`, which cannot be caught if I understand correctly; no idea why they didn't return `-1` as per the specification. These substitutions where removed with 22.2.0 (https://github.com/oracle/graal/pull/4383/).

I didn't find documentation for Azul Platform Prime. If we want to be on the safe side we could simply provide the value as is and remove the base unit, but that could be confusing as well.

Any suggestions how to proceed with this?

marcingrzejszczak: I think that what we currently have is a good start. We can always add new entries when necessary. WDYT @shakuzen @jonatan-ivanov ?

jonatan-ivanov: I'm thinking if Micrometer should be trying to be "smart" here or not.
I think if the JVM returns something gibberish/unexpected, it might not be the job of Micrometer to try to fix that. This might get even worse not only because there are many Java distributions but this behavior can change throughout versions.

I'm thinking if Micrometer should return the value of `getProcessCpuTime` as is instead of trying to cover all the cases that every versions of every supported JVMs could come up with.

It seems that newer versions of OpenJ9 return nanos, see the [code](https://github.com/eclipse-openj9/openj9/blob/c1df83d33e08556bc8f2eb91652ae608b342a3c4/jcl/src/jdk.management/share/classes/com/ibm/lang/management/OperatingSystemMXBean.java#L115-L125):
```java
	/**
	 * Returns total amount of time the process has been scheduled or
	 * executed so far in both kernel and user modes. Returns -1 if the
	 * value is unavailable on this platform or in the case of an error.
	 *
	 * Note that as of Java 8 SR5 the returned value is in 1 ns units, unless the system property
	 * com.ibm.lang.management.OperatingSystemMXBean.isCpuTime100ns is set to "true".
	 *
	 * @return process CPU time.
	 */
	public long getProcessCpuTime();
```

I was not able to find what is really "Java 8 SR5" and if releases before that are are still supported or not (SR4?) but I will try to ask around.

jonatan-ivanov: I got a bit more info on this: Java 8 SR5 was the Java fix pack version from September 2017. So this was already fixed in a Java version that is now over 6 years old (the fix is actually older than Micrometer 1.0.0 itself). If anyone is still running on it, they might have bigger issues, I don't think Micrometer should be smart here, we should just use the value as-is.

simonzn: Thanks for investigating, I will remove the check & multiplication. Should we then assume every implementation is compliant with the specification and leave the base unit as is, for documentation purposes?

jonatan-ivanov: Yes, I think so.

simonzn: Ok, done

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you for the contribution!
I rebased the PR on top of `main`, polished a bit, added a few more asserts and merged in.
It should be available as a snapshot in a few minutes and go out with our next milestone release next month.

PR #4502, opened 2023-12-20
Minor change: Align the types of elements on which @Timed and @Counted can be applied

baeyenwo:
Add `ElementType.ANNOTATION_TYPE`, `ElementType.TYPE` as targetTypes to `@Counted`

`@Timed` and `@Counted` are used in very similar ways. 
However the type of elements where the annotation is allowed is not aligned.

This pr proposes to fix the inconsistency.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java
@@ -35,7 +35,7 @@ import java.lang.annotation.*;
  * @see io.micrometer.core.aop.CountedAspect
  */
 @Inherited
-@Target(ElementType.METHOD)
+@Target({ ElementType.ANNOTATION_TYPE, ElementType.TYPE, ElementType.METHOD })
 @Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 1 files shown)


marcingrzejszczak (approved):

shakuzen (changes requested):
We should add tests that prove this works on the new targets.

shakuzen:
I tried adding a test to `CountedAspectTest` for this and it fails. We'll need to update the `CountedAspect` implementation to handle the new targets before we would want to merge this.

```java
@Test
void countClass() {
    CountedClassService service = getAdvisedService(new CountedClassService());

    service.hello();

    assertThat(meterRegistry.get("class.counted")
        .tag("class", "io.micrometer.core.aop.CountedAspectTest$CountedClassService")
        .tag("method", "hello")
        .counter()
        .count()).isEqualTo(1);
}

@Counted("class.counted")
static class CountedClassService {
    String hello() {
        return "hello";
    }

    void fail() {
        throw new RuntimeException("Oops");
    }
}
```

shakuzen:
I've opened #4746 for tracking this work.

PR #4510, opened 2023-12-24
Use spin-wait hints

marschall:
Use spin-wait hints in the spin locks on Java 11 and later.

See: [JEP 285 Spin-Wait Hints](https://openjdk.org/jeps/285)

A couple of notes:
- I had to copy and paste the `package-info.java` otherwise the `JavadocPackage` rule would fail.
- Since I only found spin locks in the `io.micrometer.core.instrument.composite` I created a package private class there. An other option would be a public class in `io.micrometer.core.instrument.internal`.
- I used a multi-release class with version Java 11 since that is already in place.
- We could reduce memory consumption further by replacing the `AtomicBoolean childrenGuard` with a `volatile int childrenGuard` and a `static final AtomicIntegerFieldUpdater`. As long as we need to be backwards compatible we can unfortunately not use a `VarHandle`.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/composite/AbstractCompositeMeter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/composite/AbstractCompositeMeter.java
@@ -82,6 +82,7 @@ abstract class AbstractCompositeMeter<T extends Meter> extends AbstractMeter imp
                     childrenGuard.set(false);
                 }
             }
+            SpinLockSupport.onSpinWait();
         }
     }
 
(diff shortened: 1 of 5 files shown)


marcingrzejszczak:
Thank you for the PR. Is there any issue that this PR addresses? What problems are we trying to solve here? Is there a performance problem that currently you can measure that we should be fixing?

marschall:
See the linked JEP for a discussion on the benefits of hints for spin loops.

No, I do not currently have a productive or synthetic example where the absence of the hint causes a measurable issue. I was not looking for this, I noticed it when browsing the code for a different purpose.

jonatan-ivanov:
@marschall Could you please look into our benchmarks and check if this change has any effect on performance?

I think it should not really since where this is happening is not on the hot path. In fact this should only happen once in the lifecycle of an application (creating the registry).
Please let us know if you think differently and we can reopen this PR.

Issue #3231, opened 2022-06-16
Enhance code logic in PrometheusNamingConvention


EricGao888:
**Please describe the feature request.**
* There are two `switch` statements in `PrometheusNamingConvention`, which might be combined as one.

https://github.com/micrometer-metrics/micrometer/blob/393de2ffd8140d4e665bc0851d22b8772ee4c4a0/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java#L56-L79

**Rationale**
* Current logic makes developers a little bit confused.

**Additional context**
* I'd like to submit a PR to fix this.

marcingrzejszczak:
Thank you for filing the issue and sorry for the delay, please file a PR :)

EricGao888:
> Thank you for filing the issue and sorry for the delay, please file a PR :)

done

marcingrzejszczak:
Currently this introduces more problems than fixes things. Let's close it for now.

PR #4513, opened 2023-12-28
[Improvement] Merge switch block in PrometheusNamingConvention

EricGao888:
* This PR closes: #3231

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java
@@ -55,18 +55,10 @@ public class PrometheusNamingConvention implements NamingConvention {
 
         switch (type) {
             case COUNTER:
-            case DISTRIBUTION_SUMMARY:
-            case GAUGE:
-                if (baseUnit != null && !conventionName.endsWith("_" + baseUnit))
-                    conventionName += "_" + baseUnit;
-                break;
-        }
-
-        switch (type) {
-            case COUNTER:
                 if (!conventionName.endsWith("_total"))
                     conventionName += "_total";
                 break;
+            case DISTRIBUTION_SUMMARY:
             case TIMER:
             case LONG_TASK_TIMER:
                 if (!timerSuffix.isEmpty() && conventionName.endsWith(timerSuffix)) {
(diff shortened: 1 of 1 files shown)


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java
```diff
@@ -53,20 +53,12 @@ public PrometheusNamingConvention(String timerSuffix) {
...
-            case COUNTER:
-            case DISTRIBUTION_SUMMARY:
-            case GAUGE:
-                if (baseUnit != null && !conventionName.endsWith("_" + baseUnit))
-                    conventionName += "_" + baseUnit;
-                break;
-        }
-
         switch (type) {
             case COUNTER:
```
marcingrzejszczak, line 57: That's not the same for counter. For counter it would do 

```java
                    conventionName += "_" + baseUnit;
```

and then

```
if (!conventionName.endsWith("_total"))
                    conventionName += "_total";
```

whereas in your case it does only the conventionName part.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java
```diff
@@ -53,20 +53,12 @@ public PrometheusNamingConvention(String timerSuffix) {
...
                 if (!conventionName.endsWith("_total"))
                     conventionName += "_total";
                 break;
+            case DISTRIBUTION_SUMMARY:
```
marcingrzejszczak, line 61: For `DISTRIBUTION_SUMMARY` it would do

```java
if (baseUnit != null && !conventionName.endsWith("_" + baseUnit))
                    conventionName += "_" + baseUnit;
```

and now it does 

```java
 if (!timerSuffix.isEmpty() && conventionName.endsWith(timerSuffix)) {
                    conventionName += timerSuffix + "_seconds";
                }
```

marcingrzejszczak (changes requested):
As far as I see it this PR at completely changes the logic. I don't think this is correct

EricGao888:
> As far as I see it this PR at completely changes the logic. I don't think this is correct

Thanks for the education. My completely misunderstanding with `switch` statement. Will fix it.

Issue #3873, opened 2023-06-01
Jetty Connector Metrics are only reported onClosed


kelunik:
**Describe the bug**

`io.micrometer.core.instrument.binder.jetty.JettyConnectionMetrics` reports metrics only in `onClosed`. Thus, connections receiving lots of traffic / messages but that are kept open, will never be visible in the metrics, then create a high peak at closure.

**Environment**

 - Micrometer version: 1.10.6
 - Micrometer registry: Simple + Elastic APM Agent
 - OS: Linux
 - Java version: 17.0.7

**To Reproduce**

```java
Server server = new Server(createThreadPool());
ServerConnector connector = new ServerConnector(server, 2, -1, tls, alpn, h2);
connector.addBean(new JettyConnectionMetrics(MetricsRegistry.register(), connector, List.of()));
connector.setPort(1234);
server.addConnector(connector);
```

Keep the connection in the client open (e.g. OkHttp and send a request every minute - default idle time is 5 minutes).

**Expected behavior**

Traffic being visible in the reported metrics.

marcingrzejszczak:
So what would be the expected outcome? Would you want the metrics to be available as gauges?

kelunik:
No, counters are fine, but metrics should be recorded when bytes are sent / received instead of on connection closure.

marcingrzejszczak:
I see, so instead of doing a listener on the connection, most likely we should be using the `NetworkTrafficListener` and use the `incoming` and `outgoing` methods? 

```java
public interface NetworkTrafficListener
{
    /**
     * <p>Callback method invoked when a connection from a remote client has been accepted.</p>
     * <p>The {@code socket} parameter can be used to extract socket address information of
     * the remote client.</p>
     *
     * @param socket the socket associated with the remote client
     */
    default void opened(Socket socket)
    {
    }

    /**
     * <p>Callback method invoked when bytes sent by a remote client arrived on the server.</p>
     *
     * @param socket the socket associated with the remote client
     * @param bytes the read-only buffer containing the incoming bytes
     */
    default void incoming(Socket socket, ByteBuffer bytes)
    {
    }

    /**
     * <p>Callback method invoked when bytes are sent to a remote client from the server.</p>
     * <p>This method is invoked after the bytes have been actually written to the remote client.</p>
     *
     * @param socket the socket associated with the remote client
     * @param bytes the read-only buffer containing the outgoing bytes
     */
[...]

PR #4514, opened 2023-12-29
Record Jetty bytes in/out metrics through network listeners

marcingrzejszczak:
before this change we were dumping in / out bytes from the connection upon the connection close which could result in big traffic whenever the connection got closed after this change we're measuring the bytes through `NetworkTrafficListener` that hooks in whenever actual in / out bytes are being processed

fixes gh-3873

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetricsTest.java
@@ -27,8 +27,8 @@ import org.eclipse.jetty.client.HttpClient;
 import org.eclipse.jetty.client.api.Request;
 import org.eclipse.jetty.client.util.StringContentProvider;
 import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.NetworkTrafficServerConnector;
 import org.eclipse.jetty.server.Server;
-import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.util.component.LifeCycle;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
@@ -44,12 +44,14 @@ class JettyConnectionMetricsTest {
 
     private Server server = new Server(0);
 
-    private ServerConnector connector = new ServerConnector(server);
+    private NetworkTrafficServerConnector connector = new NetworkTrafficServerConnector(server);
 
     private CloseableHttpClient client = HttpClients.createDefault();
 
     void setup() throws Exception {
-        connector.addBean(new JettyConnectionMetrics(registry));
+        JettyConnectionMetrics metrics = new JettyConnectionMetrics(registry);
+        connector.addBean(metrics);
+        connector.addNetworkTrafficListener(metrics);
         server.setConnectors(new Connector[] { connector });
         server.start();
     }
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetricsTest.java
```diff
@@ -44,12 +44,14 @@ class JettyConnectionMetricsTest {
...
-    private ServerConnector connector = new ServerConnector(server);
+    private NetworkTrafficServerConnector connector = new NetworkTrafficServerConnector(server);
 
     private CloseableHttpClient client = HttpClients.createDefault();
 
     void setup() throws Exception {
-        connector.addBean(new JettyConnectionMetrics(registry));
+        JettyConnectionMetrics metrics = new JettyConnectionMetrics(registry);
+        connector.addBean(metrics);
+        connector.addNetworkTrafficListener(metrics);
```
shakuzen, line 54: This looks like new setup that isn't documented in the class JavaDoc or done in the static `addToAllConnectors` method. It will be a behavior change that the bytes in/out will only be available if this is set, right?

shakuzen (commented):
I'm not sure about the difference in the number of bytes counted. It would be good to understand that. The description of the metrics should be updated as well.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetricsTest.java
marcingrzejszczak: Yes, that's why I'm showing this in the tests that bytes are not equal to what they were used to.

shakuzen:
@joakime do you have some time to help us understand this? When changing from counting the bytes in/out on the connection to a network listener and we're getting different numbers.

joakime:
Depending on where the integration point is in the handler tree you will either see the bytes as initiated by the webapp, or the bytes as sent on the network.

The most obvious and largest difference would be if a GzipHandler is in place.
You could be getting the bytes as seen before the compression or after the compression.

The techniques that micrometer were using in Jetty 9/10/11 were only seeing the bytes passing through the HttpOutput / HttpInput classes.
Which is not network bytes.

Using the `org.eclipse.jetty.io.NetworkTrafficListener` you will get everything on the network.

* The Headers
* The Body
* The Transfer-Encoding differences (like chunking overhead)
* The HTTP Trailers
* The TLS negotiation
* The TLS overhead
* The HTTP protocol overhead (like in HTTP/2)

While in the techniques you used in Jetty 9/10/11 on micrometer you saw only the Body before Transfer-Encoding.

shakuzen:
Thanks for the explanation. One effect of changing from counting the bytes via `Connection.Listener` to `NetworkTrafficListener` seems to be the loss of the ability to count the bytes in/out for `HttpClient`, because there's no way to register a `NetworkTrafficListener` on an `HttpClient` as far as I can tell; it can only be registered on the server-side. Is that right?

joakime:
I agree, it would be nice to have HttpClient also participate in NetworkTrafficListener ...

So I opened https://github.com/jetty/jetty.project/issues/11491 to do just that!

shakuzen:
Thank you for that. I've subscribed to the issue. We can update our tests/documentation when that is available.

PR #4517, opened 2023-12-31
Fix concurrency in grpc test

ttddyy:
one-liner fix for the variable which could be accessed in parallel.


izeye:
This is not a user-facing bug, so it would be better to be labeled as "type: task". It also appears in the "Bug Fixes" section of the release notes that you might want to change.

See https://github.com/micrometer-metrics/micrometer/releases/tag/v1.11.8

jonatan-ivanov:
Thank you!
I fixed the label, the release notes for `1.11.8` and also for [`1.12.2`](https://github.com/micrometer-metrics/micrometer/releases/tag/v1.12.2).

PR #4561, opened 2024-01-04
Make route tags optional in OkHttpMetricsEventListener

kelunik:
Connecting to different hosts / ports leads to metric explosion without disabling these tags.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
@@ -76,9 +76,6 @@ public class OkHttpMetricsEventListener extends EventListener {
 
     private static final String TAG_VALUE_UNKNOWN = "UNKNOWN";
 
-    private static final Tags TAGS_TARGET_UNKNOWN = Tags.of(TAG_TARGET_SCHEME, TAG_VALUE_UNKNOWN, TAG_TARGET_HOST,
-            TAG_VALUE_UNKNOWN, TAG_TARGET_PORT, TAG_VALUE_UNKNOWN);
-
     @Nullable
     private static Method getMethod(Class<?>... parameterTypes) {
         try {
@@ -101,6 +98,12 @@ public class OkHttpMetricsEventListener extends EventListener {
 
     private final Iterable<Tag> unknownRequestTags;
 
+    private final boolean includeTargetSchemeTag;
+
+    private final boolean includeTargetHostTag;
+
+    private final boolean includeTargetPortTag;
+
     private final boolean includeHostTag;
 
     // VisibleForTesting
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
Hey, thank you for the PR @kelunik . Is this fixing an issue? Why can't you use a `MeterFilter` to remove these?

kelunik:
Hi @marcingrzejszczak, we indeed experienced too high cardinality with these tags to a point where metrics started getting dropped, because they couldn't be delivered in time. However, this became only an issue in our case combined with the fact that Elastic seems to still report metrics with `0` if there are no further events but the JVM keeps running, so the number of metrics to report kept growing, eventually leading to the drop.

I wasn't aware of `MeterFilter`, ignoring the tags there will likely work as well. The documentation says:

> `ignoreTags(String…​)`: Drops matching tag keys from every meter. This is particularly useful when a tag provably comes to have too high cardinality and starts stressing your monitoring system or costing too much but you cannot change all the instrumentation points quickly.

_"cannot change all the instrumentation points quickly"_ sounds like this is more like an escape hatch and directly changing the instrumentation points is the preferred solution? I'd still prefer to directly disable these tags, as it keeps the metrics configuration in one place instead of one part in the registry config and another in the event listener config.

marcingrzejszczak:
I'm just wondering because nobody else was complaining about the cardinality issue of these tags. We prefer not to add new features if there are existing ways to mitigate issues.

kelunik:
@marcingrzejszczak Usually, the request patterns are rather similar, I guess, so the issue doesn't appear. We had used the listener in other places as well before without any issues, but the one use case we have uses dynamic ephemeral ports, so runs into this issue.

marcingrzejszczak:
I would advocate for closing this PR until we get more interest from the community. WDYT @jonatan-ivanov @shakuzen ?

shakuzen:
I don't especially like adding flags for each tag. However, cardinality explosion with the default instrumentation is also undesirable. HTTP client instrumentation can be difficult to do in such a way it necessarily avoids cardinality issues due to difficulty in tagging `uri` in a low cardinality way out-of-the-box and the cardinality of host/port depends on usage/environment. For that reason, in other instrumentations we have offered a flag for route tags (see MicrometerHttpRequestExecutor for Apache HttpClient 5). Alternatively we've had the concept of a tags provider (see JettyClientTagsProvider) where we can have a default that users can override. Basically, we haven't been consistent and I'm not sure what the best path forward to unify on is, but maybe there is another approach we could take to allow disabling this for some clients where it is known ahead of time route tags are not wanted.

At least it is a good idea to cross reference the MeterFilter documentation from the documentation specifically for OkHttp (and other HTTP clients) and mention this potential cardinality problem. To that end, @kelunik would you be willing to open a GitHub issue to capture your concerns while we figure out what approach makes sense?

The MeterFilter approach should work for users affected in the meantime, and I would also note that we have the following convenience MeterFilter which I believe Spring Boot uses for the HTTP client instrumentations they configure to avoid unbounded cardinality.

> `maximumAllowableTags(String meterNamePrefix, String tagKey, int maximumTagValues, MeterFilter onMaxReached)`: Places an upper bound on the number of tags produced by the matching series.

shakuzen:
Thank you for opening that. We'll take a holistic look at what we can do to improve this across implementations.
We'll close this for now, then.

Issue #4053, opened 2023-08-29
OTLPMeterRegistry should adhere to OpenTelemetry data model for producing metrics


lenin-jaganathan:
The current implementation of OTLPMeterRegistry seems to violate some of the principles of [Metric Datamodel specified by OpenTelemetry.
](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md)

see below for the data model copied over from OTEL proto,

```
  Metric
------------+
name        |
description |
unit        |     +------------------------------------+
data        |---> |Gauge, Sum, Histogram, Summary, ... |
------------+     +------------------------------------+

 Data [One of Gauge, Sum, Histogram, Summary, ...]
-----------+
...        |  // Metadata about the Data.
points     |--+
-----------+  |
              |      +---------------------------+
              |      |DataPoint 1                |
              v      |+------+------+   +------+ |
           +-----+   ||label |label |...|label | |
           |  1  |-->||value1|value2|...|valueN| |
           +-----+   |+------+------+   +------+ |
           |  .  |   |+-----+                    |
           |  .  |   ||value|                    |
           |  .  |   |+-----+                    |
           |  .  |   +---------------------------+
           |  .  |                   .
           |  .  |                   .
           |  .  |                   .
           |  .  |   +---------------------------+
           |  .  |   |DataPoint M                |
           +-----+   |+------+------+   +------+ |
           |  M  |-->||label |label |...|label | |
           +-----+   ||value1|value2|...|valueN| |
                     |+------+------+   +------+ |
                     |+-----+                    |
                     ||value|                    |
                     |+-----+                    |
                     +---------------------------+
```

But the micrometer's producer creates N number of metrics with the same. We basically create one metric object per meter and add exactly one data point to it. This violates the specification mentioned [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#opentelemetry-protocol-data-model).
The specification goes on to mention how conflicts can be resolved in cases where description/unit mismatches. Micrometer OTLPMeterRegistry should honor these conventions for the following reasons,

* The OpenTelemetry consumer recommendation says that "Consumers MAY reject OpenTelemetry Metrics data containing semantic errors (i.e., more than one Metric identity for a given name, Resource, and Scope).". This can be problematic for micrometer users if they run into consumers who reject these data points.
[...]

PR #4564, opened 2024-01-12
Group OTLP data points under same metric

lenin-jaganathan:
This PR improves the OTLP export by effectively leveraging OpenTelemetry Line Protocol to encode metrics. This will both reduce the network bytes as well as the memory allocation we do in the JVM. I have attached a sample of what is changing. This will provide huge gains since in a real-world environment where there will only be a limited set of metrics that differ only in the tags/attributes (usually 100's - 1000s of time series for the same metric name and we will avoid duplicate publishing of the metadata). This will partially resolve https://github.com/micrometer-metrics/micrometer/issues/4053 by optimizing the publishing and eliminating most of the semantic errors.

### What is not addressed?
1. When a metric name has multiple descriptions (this shouldn't be an ideal usage but the micrometer allows this), we consider this a separate metric and add it to a new metric. (OTLP suggestion is to prefer the longest string but with micrometer design, this will need a lot of changes at the core and how a metric is treated in micrometer).
2. When a metric name has multiple base units, we will consider that a separate metric. (OTLP suggestion is to convert the data if possible.)

However, the above 2 will be extreme cases where the same metric name will have these properties with differing descriptions or baseUnit. Addressing this will need to have a different approach to how we identify metrics in micrometers and this corner case can be continued to be resolved as part of the mentioned issue.

### Comparison before and after:
Let's assume 2 counters - counter{tag=key1}, counter{tag=key2}. The description of the counters is "Test counter" and the base unit is "byte".

<details>
<summary>DataExported before this change </summary>

```
Metric #0
Descriptor:
     -> Name: counter
     -> Description: Test counter
     -> Unit: bytes
     -> DataType: Sum
     -> IsMonotonic: true
     -> AggregationTemporality: Delta
NumberDataPoints #0
Data point attributes:
     -> key: Str(value2)
StartTimestamp: 2024-01-12 08:50:25 +0000 UTC
Timestamp: 2024-01-12 08:50:30 +0000 UTC
Value: 1.000000
Metric #1
Descriptor:
     -> Name: counter
     -> Description: Test counter
     -> Unit: bytes
     -> DataType: Sum
     -> IsMonotonic: true
     -> AggregationTemporality: Delta
NumberDataPoints #0
Data point attributes:
     -> key: Str(value1)
StartTimestamp: 2024-01-12 08:50:25 +0000 UTC
Timestamp: 2024-01-12 08:50:30 +0000 UTC
Value: 1.000000
```
</details>

<details>
<summary>DataExported after this change </summary>

```
Metric #0
Descriptor:
     -> Name: counter
     -> Description: Test counter
     -> Unit: bytes
     -> DataType: Sum
     -> IsMonotonic: true
     -> AggregationTemporality: Delta
NumberDataPoints #0
Data point attributes:
     -> key: Str(value2)
StartTimestamp: 2024-01-12 08:50:25 +0000 UTC
Timestamp: 2024-01-12 08:50:30 +0000 UTC
Value: 1.000000
NumberDataPoints #1
Data point attributes:
     -> key: Str(value1)
StartTimestamp: 2024-01-12 08:50:25 +0000 UTC
Timestamp: 2024-01-12 08:50:30 +0000 UTC
Value: 1.000000
```
</details>

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
@@ -20,6 +20,7 @@ import io.micrometer.common.util.internal.logging.InternalLogger;
 import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
 import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.distribution.*;
 import io.micrometer.core.instrument.distribution.Histogram;
@@ -34,7 +35,6 @@ import io.micrometer.core.instrument.step.StepFunctionTimer;
 import io.micrometer.core.instrument.step.StepMeterRegistry;
 import io.micrometer.core.instrument.util.MeterPartition;
 import io.micrometer.core.instrument.util.NamedThreadFactory;
-import io.micrometer.core.instrument.util.TimeUtils;
 import io.micrometer.core.ipc.http.HttpSender;
 import io.micrometer.core.ipc.http.HttpUrlConnectionSender;
 import io.opentelemetry.proto.collector.metrics.v1.ExportMetricsServiceRequest;
@@ -52,13 +52,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-import java.util.function.DoubleSupplier;
 import java.util.function.ToDoubleFunction;
 import java.util.function.ToLongFunction;
-import java.util.stream.Collectors;
-
-import static io.opentelemetry.proto.metrics.v1.AggregationTemporality.AGGREGATION_TEMPORALITY_CUMULATIVE;
-import static io.opentelemetry.proto.metrics.v1.AggregationTemporality.AGGREGATION_TEMPORALITY_DELTA;
 
 /**
  * Publishes meters in OTLP (OpenTelemetry Protocol) format. HTTP with Protobuf encoding
@@ -83,11 +78,9 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
 
     private final Resource resource;
 
-    private final io.opentelemetry.proto.metrics.v1.AggregationTemporality otlpAggregationTemporality;
+    private final AggregationTemporality aggregationTemporality;
 
     private final TimeUnit baseTimeUnit;
-
-    private long deltaAggregationTimeUnixNano = 0L;
 
     // Time when the last scheduled rollOver has started. Applicable only for delta
     // flavour.
@@ -112,9 +105,7 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
         this.baseTimeUnit = config.baseTimeUnit();
         this.httpSender = httpSender;
         this.resource = Resource.newBuilder().addAllAttributes(getResourceAttributes()).build();
-        this.otlpAggregationTemporality = AggregationTemporality
-            .toOtlpAggregationTemporality(config.aggregationTemporality());
-        setDeltaAggregationTimeUnixNano();
+        this.aggregationTemporality = config.aggregationTemporality();
         config().namingConvention(NamingConvention.dot);
         start(DEFAULT_THREAD_FACTORY);
     }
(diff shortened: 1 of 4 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -140,15 +131,10 @@ public void stop() {
 
     @Override
     protected void publish() {
-        if (isDelta()) {
```
marcingrzejszczak, line 143: Isn't this a breaking change?

marcingrzejszczak (approved):
WDYT @shakuzen @jonatan-ivanov ?

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
lenin-jaganathan: Nope. This is being moved to the OTLPMetricConverter([see here](https://github.com/micrometer-metrics/micrometer/blob/56b3607a68840fad0f6f3d5287f3f34cca8521a4/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java#L67)).

lenin-jaganathan:
@marcingrzejszczak Did you get a chance to check back on this?

shakuzen:
> 1. When a metric name has multiple descriptions (this shouldn't be an ideal usage but the micrometer allows this), we consider this a separate metric and add it to a new metric.

This has come up in the past. I think we should try to update all of our instrumentation to avoid having different descriptions. If we don't already have an issue to track this, feel free to open one, and we can try to tackle that in this upcoming minor.

lenin-jaganathan:
I will write an follow-up to have more finer control of what should be defined as a meter.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMeterRegistryTest.java
```diff
@@ -52,6 +52,11 @@ abstract class OtlpMeterRegistryTest {
 
     abstract OtlpConfig otlpConfig();
 
+    @AfterEach
+    void clearRegistry() {
+        registry.clear();
```
shakuzen, line 57: This shouldn't be necessary. Each test will run with a new instance of `OtlpMeterRegistry`

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
```diff
@@ -0,0 +1,346 @@
...
+
+    OtlpMetricConverter(Clock clock, Duration step, TimeUnit baseTimeUnit,
+            AggregationTemporality aggregationTemporality, NamingConvention namingConvention) {
+        this.clock = clock;
+        this.step = step;
+        this.aggregationTemporality = aggregationTemporality;
+        this.otlpAggregationTemporality = AggregationTemporality.toOtlpAggregationTemporality(aggregationTemporality);
+        this.baseTimeUnit = baseTimeUnit;
+        this.namingConvention = namingConvention;
+        this.deltaTimeUnixNano = (clock.wallTime() / step.toMillis()) * step.toNanos();
```
shakuzen, line 67: Isn't this (in combination with where the constructor is used in `publish`) changing from setting the delta time once per publish to setting it once per batch? I guess we don't have a test that would catch such a change.
Was the change intentional? It seems unrelated to the main purpose of the pull request.

shakuzen (approved):
Looks good overall. I left one question.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
lenin-jaganathan: Yeah, this was intentional to this per batch reason being, in normal scenarios where the meters are all read and marshaled into OTLP protocol completed within the step this would mean no change except that computation of delatTimeUnixNano happens multiple times (once per batch). NOTE: Irrespective of how many computations happen the value is the same.

In case, there are too many and all of them are not read within the same step interval (like half are read and some spill over), the value would have rolled over to the next step, and re-computing per batch would at least set the right timestamp for that step. But maybe the right thing would be to re-compute for every meter. That looks sub-optimal to me given that if this scenario happens already the metrics are flawed and users have to re-configure their steps to accommodate publishing all the meters. What do you think?

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMeterRegistryTest.java
lenin-jaganathan: removed it.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
shakuzen: I think it's fine to make the change, but it should be called out as a change in behavior that was made. This is both for review purposes and tracking/announcing changes to users. It's not clear from the title or description that this would be changed as part of this.

lenin-jaganathan: I have also updated the description of PR to reflect the same.

PR #4582, opened 2024-01-17
Downgraded OkHttp

marcingrzejszczak:
related to https://github.com/micrometer-metrics/micrometer/pull/4523


gradle/libs.versions.toml
```diff
@@ -58,7 +58,7 @@ mockito = "5.5.0"
 mongo = "4.8.2"
 netty = "4.1.104.Final"
 newrelic-api = "5.14.0"
-okhttp = "5.0.0-alpha.11" # TODO is compiling against an alpha version intentional?
+okhttp = "4.11.0"
```
shakuzen, line 61: Downgraded from 4.12.0 to 4.11.0. This better matches the 5.0 alpha we were compiling against in terms of Kotlin dependencies.

micrometer-core/build.gradle
```diff
@@ -1,5 +1,5 @@
 plugins {
-    id "org.jetbrains.kotlin.jvm" version "1.7.22"
+    alias(libs.plugins.kotlin17)
```
shakuzen, line 2: We will continue using the same Kotlin plugin version. We don't want to upgrade the minor version in a maintenance release.

micrometer-core/build.gradle
```diff
@@ -82,6 +82,7 @@ dependencies {
     optionalApi 'io.micrometer:context-propagation'
 
     // kotlin
+    optionalApi enforcedPlatform('org.jetbrains.kotlin:kotlin-bom')
```
shakuzen, line 85: Consistently enforces the kotlin library versions match the plugin version. Without `enforcedPlatform`, Gradle was resolving higher versions than we wanted.

shakuzen (commented):
I copied some tests over to a sample module that builds with Kotlin 1.9 so we are checking our Kotlin-related code with the latest version supported while we compile against the oldest version we support (Kotlin 1.7)

PR #4583, opened 2024-01-17
[Dynatrace] Logging verbosity: Move metadata discrepancy logging to WarnThenDebug

pirgeo:
Many metrics have different descriptions based on their attributes (e.g., `log4j2.events` with tag `loglevel=info` has description `Number of info level log events`, while the same metric key with tag `loglevel=trace` has the description `Number of trace level log events`). Every time a discrepancy like that is detected, the Dynatrace exporter will log a warning. Consequently, many warnings will be logged, and continue to be logged during the runtime of the process. In order to reduce the number of warning logs produced, this PR introduces a WarnThenDebugLogger for this particular log. 

While working on this, I realized that the other WarnThenDebugLoggers in the Dynatrace exporter are actually using the same instance (since they are initialized with the same class type). I fixed that by introducing thin wrappers for different WarnThenDebugLoggers. There is no change in application behaviour, just in the logging output.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/WarnThenDebugLoggers.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/WarnThenDebugLoggers.java
@@ -21,10 +21,9 @@ import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
  * This internal class holds loggers that are used in {@link DynatraceExporterV2}. They
  * are all just extending the default {@link WarnThenDebugLogger}. It is necessary to
  * extend them, because the {@link WarnThenDebugLogger} does not allow creating a new
- * logger with just a name (a class object has to be passed). If the WarnThenDebugLogger
- * is created with the same class type multiple times, the same instance will be returned.
- * This would mean that in a scenario with multiple loggers, only the first call would log
- * the warning, and all subsequent calls will log at debug.
+ * logger with just a name (a class object has to be passed). Creating the
+ * WarnThenDebugLogger with the same class multiple times makes it impossible to test, as
+ * the MockLoggerFactory will ignore multiple loggers with the same name.
  */
 class WarnThenDebugLoggers {
 
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/WarnThenDebugLoggers.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/WarnThenDebugLoggers.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2024 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.dynatrace.v2;
+
+import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
+
+/**
+ * This internal class holds loggers that are used in {@link DynatraceExporterV2}. They
+ * are all just extending the default {@link WarnThenDebugLogger}. It is necessary to
+ * extend them, because the {@link WarnThenDebugLogger} does not allow creating a new
+ * logger with just a name (a class object has to be passed). If the WarnThenDebugLogger
+ * is created with the same class type multiple times, the same instance will be returned.
+ * This would mean that in a scenario with multiple loggers, only the first call would log
+ * the warning, and all subsequent calls will log at debug.
+ */
+class WarnThenDebugLoggers {
+
+    static class StackTraceLogger extends WarnThenDebugLogger {
+
+        public StackTraceLogger() {
+            super(StackTraceLogger.class);
+        }
+
+    }
+
+    static class NanGaugeLogger extends WarnThenDebugLogger {
+
+        public NanGaugeLogger() {
+            super(NanGaugeLogger.class);
+        }
+
+    }
+
+    static class MetadataDiscrepancyLogger extends WarnThenDebugLogger {
+
+        public MetadataDiscrepancyLogger() {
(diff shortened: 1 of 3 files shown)


marcingrzejszczak:
Thank you for your PR. Can you please elaborate more on what is currently not working and how this fixes things? I don't fully understand things.

pirgeo:
Absolutely! Micrometer allows multiple descriptions for one metric name. In Dynatrace, currently we only accept one description for all metrics of a metric key. Let me elaborate on my example above. Let's assume they were created like this (this is probably not true, but for the sake of the example it doesn't matter, I think):

```java
Counter.Builder infoCounter = Counter.builder("log4j2.events")
    .tag("loglevel", "info")
    .description("Number of info level log event");
Counter.Builder traceCounter = Counter.builder("log4j2.events")
    .tag("loglevel", "trace")
    .description("Number of trace level log event");
```

They both have the same metric name (`log4j2.events`). When serializing, I would have to choose which of the descriptions I send to Dynatrace (since I can only set one of the two). Which meter is serialized first decides which description should be set - and it might be very different between exports and runs.

To circumvent this, we check for inconsistencies between descriptions before we export, and log a warning if they don't match. The problem I am trying to address with this PR is that we log a warning _every time this happens_, and _on each export_ - that can be a lot of warning logs. Additionally, this is not a problem that will just right itself - it requires intervention by the application author, so continuously logging the warning does not help the person looking at the logs at all. 

Hope that makes it a bit clearer - thanks for taking a look!

marcingrzejszczak (approved):

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/WarnThenDebugLoggers.java
```diff
@@ -0,0 +1,55 @@
...
+import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
+
+/**
+ * This internal class holds loggers that are used in {@link DynatraceExporterV2}. They
+ * are all just extending the default {@link WarnThenDebugLogger}. It is necessary to
+ * extend them, because the {@link WarnThenDebugLogger} does not allow creating a new
+ * logger with just a name (a class object has to be passed). If the WarnThenDebugLogger
+ * is created with the same class type multiple times, the same instance will be returned.
+ * This would mean that in a scenario with multiple loggers, only the first call would log
+ * the warning, and all subsequent calls will log at debug.
```
shakuzen, line 27: I don't think that's accurate. The two instances will use the same logger but the level is still controlled by the instance flag in each.

I wrote the following test:

```java
@Test
void twoInstancesSameLoggerClass() {
    WarnThenDebugLogger first = new WarnThenDebugLogger(WarnThenDebugLoggerTest.class);
    WarnThenDebugLogger second = new WarnThenDebugLogger(WarnThenDebugLoggerTest.class);

    first.log("hi");
    second.log("hello");
}
```

And the log output is:
```
16:19:03.684 [main] WARN io.micrometer.common.util.internal.logging.WarnThenDebugLoggerTest - hi Note that subsequent logs will be logged at debug level.
16:19:03.684 [main] WARN io.micrometer.common.util.internal.logging.WarnThenDebugLoggerTest - hello Note that subsequent logs will be logged at debug level.
```

pirgeo: I think you are right, the description is off. I wanted to assert on the different `WarnThenDebugLogger`s, to ensure that all warning messages are triggered. Initially, I was using three separate `WarnThenDebugLogger`s to do that (without the subclasses). There is only one constructor for WarnThenDebugLogger and it takes a `Class` parameter: https://github.com/micrometer-metrics/micrometer/blob/16dadc0b8b98a11a82c5e0eda0a723db392a0854/micrometer-core/src/main/java/io/micrometer/core/util/internal/logging/WarnThenDebugLogger.java#L37-L39
Since there is no obvious choice of which Class type to put there, I at first put `WarnThenDebugLogger.class`. 
In the tests, we are using the `MockLoggerFactory` to `get` these Loggers. There, I have two choices of how to retrieve the Loggers: 
https://github.com/micrometer-metrics/micrometer/blob/16dadc0b8b98a11a82c5e0eda0a723db392a0854/micrometer-test/src/main/java/io/micrometer/core/util/internal/logging/MockLoggerFactory.java#L46-L48 and
https://github.com/micrometer-metrics/micrometer/blob/16dadc0b8b98a11a82c5e0eda0a723db392a0854/micrometer-test/src/main/java/io/micrometer/core/util/internal/logging/MockLoggerFactory.java#L57-L59
Since there is no way of creating the `WarnThenDebugLogger` with a name instead of the `Class` object, all three used the `WarnThenDebugLogger.class` to create, and then the same `WarnThenDebugLogger.class` to retrieve them from the `MockLoggerFactory`. But: the `MockLoggerFactory` cannot distinguish between these instances, since they all have the same name. Consequently, the tests were very flaky, because it is not clear which `WarnThenDebugLogger` was being returned by `mockLoggerFactory.getLogger(WarnThenDebugLogger.class)`. 
In order to work around this, I created the new classes that basically set a specific "name" (by using the `Class` type of the subclass) to the separate `WarnThenDebugLogger`s. This way, there will be three separate `WarnThenDebugLogger`s in the `MockLoggerFactory` (each with a different name), and I can pull the right one to assert on. 

Do you think this is the wrong approach? Alternatively, I could add API surface to the WarnThenDebugLogger to allow creating one with a specific name. Happy to discuss other alternatives as well!

pirgeo: Updated the description in f05701bef79bbad872674002b29e8bdfb16c5b55

shakuzen (approved):

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/WarnThenDebugLoggers.java
pirgeo: Yep, sounds good! Thank you!

izeye:
> Many metrics have different descriptions based on their attributes (e.g., `log4j2.events` with tag `loglevel=info` has description `Number of info level log events`, while the same metric key with tag `loglevel=trace` has the description `Number of trace level log events`).

This seems to be similar to https://github.com/micrometer-metrics/micrometer/pull/3201. I created https://github.com/micrometer-metrics/micrometer/pull/4864 to try to resolve this.

izeye:
> Since there is no way of creating the `WarnThenDebugLogger` with a name instead of the `Class` object,

I also created https://github.com/micrometer-metrics/micrometer/pull/4865 to try to resolve this.

pirgeo:
Great news, thank you!

PR #4591, opened 2024-01-18
Update osgi import for okhttp after downgrade

shakuzen:
I don't understand why this change is needed upon downgrading, but without it the `testOSGI` task prints an exception and runs indefinitely.

@driessamyn could you shed any light on this? We downgraded okhttp from a 5.0 alpha version to 4.11.0 and the osgi test started exhibiting the behavior above. This seems to fix it, but I don't understand why the change was needed. Separately, it would be good if the `testOSGI` task would fail instead of hang when this happens.


driessamyn:
> Merging to fix the broken build, but I'd still like to understand this and know if we can improve things.

Let me try to replicate this and investigate.

PR #4595, opened 2024-01-18
URL decode headers from environment variables in OtlpConfig

zeitlinger:
Follow up for https://github.com/open-telemetry/opentelemetry-java/pull/6164

Blocked by https://github.com/open-telemetry/opentelemetry-specification/issues/3832

### DIFF
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
@@ -62,8 +62,8 @@ class OtlpConfigTest {
     @Test
     void headersUseEnvVarWhenConfigNotSet() throws Exception {
         OtlpConfig config = k -> null;
-        withEnvironmentVariable("OTEL_EXPORTER_OTLP_HEADERS", "header2=value")
-            .execute(() -> assertThat(config.headers()).containsEntry("header2", "value").hasSize(1));
+        withEnvironmentVariable("OTEL_EXPORTER_OTLP_HEADERS", "header2=va%20lue")
+            .execute(() -> assertThat(config.headers()).containsEntry("header2", "va lue").hasSize(1));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


marcingrzejszczak (approved):

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
```diff
@@ -100,8 +101,17 @@ void headersConfigTakesPrecedenceOverEnvVars() throws Exception {
     @Test
     void headersUseEnvVarWhenConfigNotSet() throws Exception {
         OtlpConfig config = k -> null;
-        withEnvironmentVariable("OTEL_EXPORTER_OTLP_HEADERS", "header2=value")
-            .execute(() -> assertThat(config.headers()).containsEntry("header2", "value").hasSize(1));
+        withEnvironmentVariable("OTEL_EXPORTER_OTLP_HEADERS", "header2=va%20lue")
+            .execute(() -> assertThat(config.headers()).containsEntry("header2", "va lue").hasSize(1));
```
shakuzen, line 105: Let's also add a test to verify that unencoded values continue to work since they did before.
`header2=va lue` should continue to work.

shakuzen (approved):
It sounds like it's a bug we didn't do this before. If so, we should include it in maintenance releases. What do you think?

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
zeitlinger: @shakuzen done

zeitlinger:
> It sounds like it's a bug we didn't do this before. If so, we should include it in maintenance releases. What do you think?

I agree - see https://github.com/open-telemetry/opentelemetry-specification/issues/3832#issuecomment-1906234313

shakuzen:
Thanks. I've rebased the pull request on the `1.11.x` branch. We'll merge forward from there to include it in our upcoming round of maintenance releases.

PR #4597, opened 2024-01-19
Downgrade okhttp to 4.11.0 upgrade Kotlin to 1.9

shakuzen:
OkHttp 4.12.0 depends on a version of okio that ends up requiring Kotlin 1.9. We only have an optional dependency on OkHttp so we could upgrade to 4.12.0, but I would rather not compile against a newer version of OkHttp than our users can use if they are on our minimum supported version of Kotlin (1.7). For that reason, compiles against the latest version of OkHttp that is compatible with Kotlin 1.7 - OkHttp 4.11.0.
Uses Kotlin 1.9 to compile but remains compatible with Kotlin 1.7.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -88,6 +88,8 @@ subprojects {
 
             toolchain {
                 languageVersion = javaLanguageVersion
+                sourceCompatibility = javaTargetVersion
+                targetCompatibility = javaTargetVersion
             }
         }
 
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -117,6 +119,9 @@ subprojects {
             compileTestJava {
                 options.encoding = 'UTF-8'
                 options.compilerArgs << '-Xlint:unchecked' << '-Xlint:deprecation'
+
+                sourceCompatibility = 11
```
jonatan-ivanov, line 123: Would it make sense to do something like this?
```
ext.javaTestTargetVersion = JavaVersion.VERSION_11
...
sourceCompatibility = javaTestTargetVersion
sourceCompatibility = javaTestTargetVersion
```
or
```
sourceCompatibility = JavaVersion.VERSION_11
sourceCompatibility = JavaVersion.VERSION_11
```

shakuzen: I'm not sure one is better than the other, but consistency would be good. I've updated it to align to your suggestion since that was already in use in some places.

Issue #4607, opened 2024-01-22
Remove unnecessary call to getConventionName in PrometheusMeterRegistry


davidmestr:
We need some customization in the collector name, but in `PrometheusMeterRegistry` and `MicrometerCollector`, the collector name is calculated two times (and the second is difficult to customize):

- In `PrometheusMeterRegistry` (`applyToCollector` method) the collector name is retrieved from `getConventionName(Meter.Id id)` method, in order to be used as a key in the internal `collectionMap`.
- But, when the `MicrometerCollector` is created, the collector name is calculated another time in the `MicrometerCollector` constructor.

We propose to unify the collector name passing the first calculated name (in `PrometheusMeterRegistry.getConventionName(id)`) to `MicrometerCollector` constructor.

With this easy change, it is posible to uncouple the metric name to the collector name, and resolve some existent issues related to some Prometheus metric name restrictions that doesn't exist in Micrometer. 

We also provide a [PR](https://github.com/micrometer-metrics/micrometer/pull/4608) with the small changes in the `PrometheusMeterRegsitry.applyToCollector` and `MicrometerCollector` classes.

PR #4608, opened 2024-01-22
feat: Unify micrometer collector name

davidmestr:
We propose to unify the collector name passing the first calculated name (in `PrometheusMeterRegistry.getConventionName(id)`) to `MicrometerCollector` constructor.

With this easy change, it is posible to uncouple the metric name to the collector name, and resolve some existent issues related to some Prometheus metric name restrictions that doesn't exist in Micrometer.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/MicrometerCollector.java
@@ -51,6 +51,13 @@ class MicrometerCollector extends Collector implements Collector.Describable {
         this.help = config.descriptions() ? Optional.ofNullable(id.getDescription()).orElse(" ") : " ";
     }
 
+    public MicrometerCollector(String name, Meter.Id id, NamingConvention convention, PrometheusConfig config) {
+        this.id = id;
+        this.conventionName = name;
+        this.tagKeys = id.getConventionTags(convention).stream().map(Tag::getKey).collect(toList());
+        this.help = config.descriptions() ? Optional.ofNullable(id.getDescription()).orElse(" ") : " ";
+    }
+
     public void add(List<String> tagValues, Child child) {
         children.put(tagValues, child);
     }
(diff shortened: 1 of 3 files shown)


marcingrzejszczak:
Hey @davidmestr thank you for the PR. Can you tell us if there's an issue that this PR fixes?

davidmestr:
You can find more details in https://github.com/micrometer-metrics/micrometer/issues/4607.

The issue is that if you overwrites the method `PrometheusMeterRegistry.getConventionName(id)`, the name (the key) in  internal `collectionMap` is different that the `MicrometerCollector` name.

marcingrzejszczak:
I understand. Is there a chance to have a test added to this change?

davidmestr:
The main test is that existent test are still working, but with the change, the `PrometheusNamingConvention.name()` is only call one time (performance improvement). More over, we add a test with `PrometheusMeterRegistry.getConventionName(id)` overwrited.

shakuzen:
> resolve some existent issues related to some Prometheus metric name restrictions that doesn't exist in Micrometer.

Could you elaborate on this? I can see the slight performance improvement, but I'm not sure if this achieves anything else. If you wanted a prefix as demonstrated in the added test, you could configure a custom `NamingConvention` without the changes in this PR. https://docs.micrometer.io/micrometer/reference/concepts/naming.html#page-title

So is this change needed functionally for anything or is it just a performance improvement?

shakuzen:
> the internal key collectionMap and the name of the collector can be different

Only if the `NamingConvention` can return a different value for the same input, right? That's what I'm curious about seeing in a test. What's the use case in which they end up being different? Both places are calling `getConventionName` on the same `Id` with the same `NamingConvention`.

davidmestr:
> Only if the `NamingConvention` can return a different value for the same input, right? 
Yes, our name convention calls an external system, and it can happen.

shakuzen:
> Yes, our name convention calls an external system, and it can happen.

I'd say that's unexpected and there's a high likelihood that such behavior could cause problems even after this change. We don't generally expect that a different convention name could be returned for the same input. Anyway, we can merge this change for the performance aspect of it.

shakuzen:
Thank you for the pull request. I've polished in https://github.com/micrometer-metrics/micrometer/commit/f2fa7a3f5422c40fc02d4bfa4decfc9e1b77fdfb

PR #4724, opened 2024-02-05
Use Dynatrace snapshot for LongTaskTimer (1.12.x)

pirgeo:
Similar to https://github.com/micrometer-metrics/micrometer/pull/3093, this PR introduces a Dynatrace-specific instrument for the LongTaskTimer. This is necessary because we still see instances of LongTaskTimer data rejected by the Dynatrace API. We tried to fix this in https://github.com/micrometer-metrics/micrometer/pull/3985, which reduced the number of errors, but did not completely solve the problem.

--- 
**The long version (an example):**

I have (albeit unsuccessfully) tried to reproduce this issue, and am now pretty sure that it has to do with a race condition. For example, this is one of the inconsistent datapoints we saw: 
```
min <= avg <= max doesn't hold: min: 44.809818, max: 44.812933, sum: 89.628272, count: 3, avg: 29.876091
```
These values are calculated in the `DefaultLongTaskTimer.takeSnapshot` method: https://github.com/micrometer-metrics/micrometer/blob/9a19c9c1dd345cdd97bc67089c6a1c367762d61c/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L130-L216 We are estimating the min as the 0% quantile of the data, since the min is not tracked by the LongTaskTimer instrument. The 0% quantile is calculated in the first section of the `takeSnapshot` method ([L131-L201](https://github.com/micrometer-metrics/micrometer/blob/9a19c9c1dd345cdd97bc67089c6a1c367762d61c/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L131-L201)). Duration and max are calculated down below: 
https://github.com/micrometer-metrics/micrometer/blob/9a19c9c1dd345cdd97bc67089c6a1c367762d61c/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L203-L204 Finally, in line 214, the count is retrieved (by calling `activeTasks.size()`): 
https://github.com/micrometer-metrics/micrometer/blob/9a19c9c1dd345cdd97bc67089c6a1c367762d61c/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L214 

These numbers are calculated in consecutive steps, which are _not synchronized_. If we look at the numbers for the error data again, this seems to be reflected:
1. The min is calculated as the 0% quantile, thus it is the smallest number: `44.809818`. 
2. Just after that, the duration (total) is calculated and it turns out to be `89.628272`. Note that this is _very close_ to min+max (which comes out to `89.622751`), so there must be two active tasks that have been running for approximately the same time at this point. 
3. The max is calculated: `44.812933`. It is very close to the min, so there were either two tasks running for almost the exact same length of time, or there was only one task, and the calculation of the total is off (which seems unlikely).
Note that we know that there must have been at most one or two tasks running at this point in time.
4. The count is retrieved by getting  `activeTasks.size()`. It is `3`, so there must be three items in the collection. Note however the notice on the [`ConcurrentLinkedDeque#size()`](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/ConcurrentLinkedDeque.html#size()) method: 

>Beware that, unlike in most collections, this method is NOT a constant-time operation. Because of the asynchronous nature of these deques, determining the current number of elements requires traversing them all to count them. **Additionally, it is possible for the size to change during execution of this method, in which case the returned result will be inaccurate. Thus, this method is typically not very useful in concurrent applications.**

(emphasis mine)

The only way I can see this happening, is if a new task was added to be recorded between the time the min/max/sum values were calculated, and the point in time where the count is retrieved. In a case like `http.server.requests.active`, it is actually quite likely that a request starts/finishes _while_ the export is still running.

We have also seen other examples that point to the size of the collection changing while the snapshot is being prepared. One example I saw had failed to create a 0% quantile because there were no activeTasks to estimate the quantile from, but had a `count > 0`. This means, that when the method started there were 0 activeTasks, but when it reached the point where the count was retrieved, suddenly there was at least one activeTask. 

**TL;DR:** min, max, sum, and count retrieval are not synchronized and are therefore potentially inconsistent if tasks finish/start during the calculation of the values.

--- 
**Proposed fix:** As we did for multiple instruments in https://github.com/micrometer-metrics/micrometer/pull/3093, we suggest to replace the LongTaskTimer with the DynatraceLongTaskTimer for the DynatraceMeterRegistry, which records data in a synchronized way and ensures that the data is consistent (as per the Dynatrace definition) at the time of export. As an added benefit, the min is automatically tracked in the DynatraceSnapshot, so it does not have to be estimated at export time.

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/types/DynatraceLongTaskTimerTest.java
+++ b/implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/types/DynatraceLongTaskTimerTest.java
@@ -46,7 +46,7 @@ class DynatraceLongTaskTimerTest {
     private static final Offset<Double> TOLERANCE = Offset.offset(0.000001);
 
     @Test
-    void testValuesAreRecorded() throws InterruptedException {
+    void singleTaskValuesAreRecorded() throws InterruptedException {
         DynatraceLongTaskTimer ltt = new DynatraceLongTaskTimer(ID, CLOCK, TimeUnit.MILLISECONDS,
                 DISTRIBUTION_STATISTIC_CONFIG, false);
         ExecutorService executorService = Executors.newSingleThreadExecutor();
@@ -60,8 +60,8 @@ class DynatraceLongTaskTimerTest {
                 taskHasBeenRunningLatch.countDown();
 
                 try {
-                    // wait until the snapshot has been taken.
-                    stopLatch.await(300, TimeUnit.MILLISECONDS);
+                    // wait until the snapshot has been taken
+                    assertThat(stopLatch.await(300, TimeUnit.MILLISECONDS)).isTrue();
                 }
                 catch (InterruptedException e) {
                     throw new RuntimeException(e);
@@ -69,7 +69,7 @@ class DynatraceLongTaskTimerTest {
             });
         });
 
-        taskHasBeenRunningLatch.await(300, TimeUnit.MILLISECONDS);
+        assertThat(taskHasBeenRunningLatch.await(300, TimeUnit.MILLISECONDS)).isTrue();
 
         DynatraceSummarySnapshot snapshot = ltt.takeSummarySnapshot();
         // can release the background task
@@ -90,7 +90,7 @@ class DynatraceLongTaskTimerTest {
      * other thread as well and time is basically "double-counted".
      */
     @Test
-    void testTwoValuesAreRecorded() throws InterruptedException {
+    void parallelTasksValuesAreRecorded() throws InterruptedException {
         // use the system clock, as it is much easier to understand than when using the
         // MockClock (When using MockClock, adding to the clock in two separate threads
         // basically means that two things happen at the same time).
@@ -108,8 +108,8 @@ class DynatraceLongTaskTimerTest {
                     Thread.sleep(70);
                     taskHasBeenRunningLatch.countDown();
 
-                    // wait until the snapshot has been taken.
-                    stopLatch.await(300, TimeUnit.MILLISECONDS);
+                    // wait until the snapshot has been taken
+                    assertThat(stopLatch.await(300, TimeUnit.MILLISECONDS)).isTrue();
                 }
                 catch (InterruptedException e) {
                     throw new RuntimeException(e);
@@ -123,8 +123,8 @@ class DynatraceLongTaskTimerTest {
                     Thread.sleep(30);
                     taskHasBeenRunningLatch.countDown();
 
-                    // wait until the snapshot has been taken.
-                    stopLatch.await(300, TimeUnit.MILLISECONDS);
+                    // wait until the snapshot has been taken
+                    assertThat(stopLatch.await(300, TimeUnit.MILLISECONDS)).isTrue();
                 }
                 catch (InterruptedException e) {
                     throw new RuntimeException(e);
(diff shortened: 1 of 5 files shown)


implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/types/DynatraceLongTaskTimerTest.java
```diff
@@ -0,0 +1,153 @@
...
+        // the first Thread has been "running" for ~30ms at the time of recording and will
+        // supply the min
+        assertThat(snapshot.getMin()).isGreaterThanOrEqualTo(30);
+        // the second Thread has been "running" for ~70ms at the time of recording and
+        // will supply the max
+        assertThat(snapshot.getMax()).isGreaterThanOrEqualTo(70).isLessThan(100);
+        // the min is greater than 30, and the max is greater than 70, so together they
+        // have to be greater than 100.
+        assertThat(snapshot.getTotal()).isGreaterThan(100);
+        assertThat(snapshot.getCount()).isEqualTo(2);
```
jonatan-ivanov, line 150: This might not be very close to the reality and that could make these tests quite brittle. Also, the this is failing for me, on my machine, running the tests from my IDE, min and max are both over 100ms.

I can see the test being more intuitive using the system clock but for robustness, I might use `MockClock`.

pirgeo:
Thanks for taking a look @jonatan-ivanov. I am pretty sure I set maintainers can modify to true, but something is going on with these organizational forks. I merged your branch and used the MockClock in the test. Thank you for the suggestions!

I'd love to backport it to older versions - I wasn't aware that they were still being released though, the latest 1.9.x version I could find was https://github.com/micrometer-metrics/micrometer/releases/tag/v1.9.17. Should I still rebase it on 1.9.x, or is it enough to rebase on 1.11.x?

pirgeo:
Here's a diff for 1.9.x actually: https://github.com/micrometer-metrics/micrometer/compare/1.9.x...dynatrace-oss-contrib:micrometer:dynatrace-longtasktimer-1.9.x?expand=1. There were a few conflicts mainly around the metadata things. When merging forward, we need to ensure that doesn't get lost.

jonatan-ivanov:
>I am pretty sure I set maintainers can modify to true, but something is going on with these organizational forks.

Yeah, I don't think you can do anything on your side, the maintainers edit flag should not give me edit access to the Dynatrace fork (I guess maybe except if you would be an org admin but I'm not sure). It should work from a fork under your own account but I guess Dynatrace does not want that so we need to handle it this way. :)

>I merged your branch and used the MockClock in the test.

Thank you!

>I'd love to backport it to older versions - I wasn't aware that they were still being released though, the latest 1.9.x version I could find was https://github.com/micrometer-metrics/micrometer/releases/tag/v1.9.17. Should I still rebase it on 1.9.x, or is it enough to rebase on 1.11.x?

1.9.x is commercial support only: https://micrometer.io/support/ for that branch we only fix things if there is a customer request. But since this change belong to your other fixes made on 1.9.x quite strongly, I was thinking that it would be nice to put this next to it (maybe somebody will need it in 1.9/1.10). I'm debating if we should do this there or not since it is more work, let me take a look at your other PR and see. Thank you for opening it!

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/4780

Issue #3828, opened 2023-05-09
Support jOOQ 3.15 and later


kousalik:
**Describe the bug**
Class CreateTableColumnStep has been removed from the jOOQ implementation in this [commit](https://github.com/jOOQ/jOOQ/commit/2393afc7f2409f2f4a51500b24dd2af643dd3f4c) in February '22. 
Using the MetricsDSLContext with a version without that class causes a runtime exception.

**Environment**
 - Micrometer version: 1.10.6
 - jOOQ version: 3.18.3
 - OS: any
 - Java version: any

**To Reproduce**
How to reproduce the bug:
Attempt to use MetricsDSLContext implementation with one of the latest jOOQ releases. Fails at runtime due to `java.lang.ClassNotFoundException: org.jooq.CreateTableColumnStep`

**Expected behavior**
Normal function.

jonatan-ivanov:
Right now we compile against JOOQ 3.14.x, we should look into upgrading it, thank you for reporting it.

shakuzen:
@diastremskii thank you for the offer and starting work on it. I can understand how that sounds like an appealing solution and it may work in practice, but since the API of the class will be different between the multi-releases, it sounds like it goes against the design of the multi-release jar feature. [JEP 238](https://openjdk.org/jeps/238) has the following to say about it:

> Every version of the library should offer the same API; investigation is required to determine whether this should be strict backwards compatibility where the API is exactly the same (byte code signature equality), or whether this can be relaxed to some degree without necessarily enabling the introduction of new enhancements that would blur the notion of one unit of release. This may imply, at a minimum, that a public class present in a release-specific directory should also be present in the root, though it need not be present in an earlier release directory. The run-time system will not verify this property, but tooling can and should detect such API compatibility issues, and a library method may also be provided to perform such varification (for example on java.util.jar.JarFile).

Given that, I'm not sure we should try to go that route. It would be more ideal if we can offer instrumentation that isn't affected by so many unrelated changes in jOOQ and therefore is compatible across a much wider range of versions.

shakuzen:
[...]

Issue #4508, opened 2023-12-22
Deprecate jooQ support


marcingrzejszczak:
more context here https://github.com/micrometer-metrics/micrometer/issues/2686

marcingrzejszczak:
We could consider leveraging the multirelease jar to support this? (check this [issue](https://github.com/micrometer-metrics/micrometer/issues/3828))

shakuzen:
I commented on that proposal over in https://github.com/micrometer-metrics/micrometer/issues/2686 but I've proposed another way we could perhaps salvage jOOQ instrumentation there. Let's put deprecation on hold to see if we can get some feedback from our jOOQ users on that.

shakuzen:
#4727 is a path to supporting the latest versions of jOOQ so deprecation should no longer be needed.

PR #4727, opened 2024-02-07
Compatibility with jOOQ 3.15 and later

shakuzen:
Implementing jOOQ's `DSLContext` interface was causing revlock and subsequently frequent breaking. Now that jOOQ OSS supported versions no longer work with Java 11, we cannot compile against them to update our implementation `MetricsDSLContext`. This switches to extending `DefaultDSLContext` instead which eliminates a lot of boilerplate implementation and should be more compatible. A sample has been added that runs the tests against the latest version of jOOQ to ensure MetricsDSLContext works with it while compiling against the oldest OSS version of jOOQ that supports Java 8.

Resolves gh-3828
Obviates gh-4508 for now

This does break binary compatibility, as reported by japi. I think that's going to be unavoidable given the circumstances and it is limited to the jOOQ integration. Users will need to recompile their code that uses `MetricsDSLContext`. I am not aware of any libraries that configure the jOOQ integration, so it should mostly be used in application code, making the recompile requirement less of an issue. Typical practice is to recompile application code when changing dependency versions.


shakuzen:
It looks like since jOOQ 3.17, open source versions only support Java 17+. I've updated the sample to only run with 17+.

samples/micrometer-samples-jooq/src/test/java/io/micrometer/samples/jooq/MetricsDSLContextTest.java
```diff
@@ -0,0 +1,209 @@
...
+import static org.jooq.impl.DSL.*;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Tests for {@link MetricsDSLContext} run on the latest version of jOOQ.
+ *
+ * @author Jon Schneider
+ * @author Johnny Lim
+ */
+class MetricsDSLContextTest {
```
shakuzen, line 46: This class is copied from micrometer-core with only import/JavaDoc changes.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/MetricsDSLContext.java
```diff
@@ -136,4972 +116,678 @@ public DSLContext tags(Iterable<Tag> tags) {
     }
 
     @Override
-    public Schema map(Schema schema) {
-        return context.map(schema);
+    public <R extends Record> SelectWhereStep<R> selectFrom(Table<R> table) {
```
ferdinand-swoboda, line 119: Any specific reason this is not 
```suggestion
public <R extends Record> SelectWhereStep<R> selectFrom(TableLike<R> table) {
```
?

shakuzen: Because we compile against jooq 3.14 and `TableLike` does not exist in that version.

PR #4748, opened 2024-02-13
fix(GRPC interceptor): customizable tag names

monosoul:
This PR changes GRPC's `MetricCollectingClientInterceptor` and `MetricCollectingServerInterceptor` to allow tag names customization.

### The problem:
Both interceptors use `service` tag name for **_GRPC_** service names (think endpoint names), while in most cases this tag name is used for the actual service name (e.g. deployment name in k8s). This causes inconsistency in metrics and there is no way to customize the tag name at the moment.

### The solution:
Technically, both interceptors allow providing `counterCustomizer` and `timerCustomizer`, but both are just functions on `Counter`/`Timer` Builder, and builders don't have any methods to **_remove_** the tags added before.

So I had two options here:

1. Add a `remove` method to the builder interface (which would be kinda weird I think).
2. Allow passing custom tag names via the constructor.

I went for the second option here. While technically it wasn't necessary to customize other tag names, I did that too, so that I can have consistent tag names for those metrics.

What I eventually went for in our app is: `grpc.service`, `grpc.method`, `grpc.methodType` and `grpc.statusCode`.

I guess, I could've as well just changed the tag names here to those values, but I imagine that could break dashboards/alerts for some folks :smile:

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/AbstractMetricCollectingInterceptor.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/AbstractMetricCollectingInterceptor.java
@@ -41,25 +41,33 @@ import java.util.function.UnaryOperator;
  */
 public abstract class AbstractMetricCollectingInterceptor {
 
+    protected static final String DEFAULT_TAG_SERVICE_NAME = "service";
+
     /**
      * The metrics tag key that belongs to the called service name.
      */
-    private static final String TAG_SERVICE_NAME = "service";
+    private final String tagServiceName;
+
+    protected static final String DEFAULT_TAG_METHOD_NAME = "method";
 
     /**
      * The metrics tag key that belongs to the called method name.
      */
-    private static final String TAG_METHOD_NAME = "method";
+    private final String tagMethodName;
+
+    protected static final String DEFAULT_TAG_METHOD_TYPE = "methodType";
 
     /**
      * The metrics tag key that belongs to the type of the called method.
      */
-    private static final String TAG_METHOD_TYPE = "methodType";
+    private final String tagMethodType;
+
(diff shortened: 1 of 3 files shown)


shakuzen:
Thank you for the pull request. I think this is similar to https://github.com/micrometer-metrics/micrometer/pull/4561#issuecomment-1909539595. Rather than add constructor arguments for this, I think we need to take a more holistic approach to this for maintainability and being able to evolve things. That likely means something like a tags provider for these, but we need to talk about a consistent approach going forward. I'll discuss it with @jonatan-ivanov.

monosoul:
I guess another way thing here works be to make the methods to create builders non-static in `AbstractMetricCollectingInterceptor`, then I'll be able to extend `MetricCollectingClientInterceptor` and the other one, and override those.

jonatan-ivanov:
I know this is about the `MetricCollecting*Interceptor` classes but this should be already possible if you use the other flavor of instrumentation that uses the Observation API since there you have the chance to provide your own `Observation.Convention`. I think in practice this would mean something like this:
```java
public class CustomGrpcClientObservationConvention extends DefaultGrpcClientObservationConvention {
    @Override
    public KeyValues getLowCardinalityKeyValues(GrpcClientObservationContext context) {
        KeyValues keyValues = super.getLowCardinalityKeyValues(context);
        return keyValues.and(KeyValues.of("your.own", context.getCarrier().getSomething()));
    }
}
```

We can either come up with a similar pattern for Meters or we can encourage users to use the `Observation`-based instrumentation instead.

shakuzen:
That's a good point @jonatan-ivanov. @monosoul are you using (or could you use) the Observation-based instrumentation for gRPC or do you need this in the Timer-based instrumentation specifically? See https://docs.micrometer.io/micrometer/reference/reference/grpc.html

Issue #4749, opened 2024-02-14
customObservationConvention is not respected for MicrometerHttpClient


provataros:
**Describe the bug**
Even though you can provide a customObservationConvention when building the MicrometerHttpClient, on observationStop, it still uses the DefaultHttpClientObservationConvention.getName() to get the convention name, completely disregarding possible overrides on the customObservationConvention ( MicrometerHttpClient#236)

**Environment**
All environments, master branch

**Expected behavior**
Use customObservationConvention .getName() instead

**Additional context**
Add any other context about the problem here, e.g. related issues.

jonatan-ivanov:
I think you are using the instrumentation with its `Timer` flavor and it only creates a `Timer` not an `Observation`. If you want to utilize the `Observation.Convention` you should use the instrumentation in a way that it creates an instrumentation otherwise you can override the name of the `Timer` using a `MeterFilter`.

/cc @shakuzen

provataros:
I eventually patched this, and added the additional tags on the observationConvention, on my 
company's branch. Once I make sure this works as expected, I will open a proper PR

jonatan-ivanov:
Didn't you already open #4751?

provataros:
I mean, i will include additional improvements that are related to this one. Or should i open a different issue for those?

shakuzen:
As Jonatan pointed out, the part you are changing in the pull request is for the `Timer`-based instrumentation (when an `ObservationRegistry` isn't configured). Please write a unit test that demonstrates what is not working as expected. I think it will be more clear then and we'll be able to make sure it stays fixed in the future. It may not have been clear before how to setup this instrumentation. Just yesterday I added documentation about it; see [the documentation](https://docs.micrometer.io/micrometer/reference/1.13-SNAPSHOT/reference/java-httpclient.html). If you are not passing an `ObservationRegistry`, then the Observation-related things like a custom `ObservationConvention` will not be used.

PR #4751, opened 2024-02-14
respect customObservationConvention on MicrometerHttpClient

provataros:
The PR is just a proposal for the changes


shakuzen:
Thanks for the pull request. Is there an underlying issue that you've run into? Could you describe that so we are aware?

jonatan-ivanov:
@shakuzen See: https://github.com/micrometer-metrics/micrometer/issues/4749

PR #4755, opened 2024-02-17
Make LongTaskTimer creation conditional in DefaultMeterObservationHandler

jonatan-ivanov:
This change makes it possible to enable/disable `LongTaskTimer` creation in `DefaultMeterObservationHandler`. If users do not need the `LongTaskTimer`, they can turn it off and spare CPU cycles and the extra time series that will not be created this way.

Usage example:
```java
ObservationRegistry observationRegistry = ObservationRegistry.create();
observationRegistry.observationConfig()
    .observationHandler(new DefaultMeterObservationHandler(meterRegistry, LONG_TASK_TIMER));
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/observation/DefaultMeterObservationHandler.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/observation/DefaultMeterObservationHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 VMware, Inc.
+ * Copyright 2024 VMware, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import io.micrometer.core.instrument.*;
 import io.micrometer.observation.Observation;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/observation/DefaultMeterObservationHandler.java
```diff
@@ -42,17 +43,38 @@ public class DefaultMeterObservationHandler implements MeterObservationHandler<O
...
     public DefaultMeterObservationHandler(MeterRegistry meterRegistry) {
         this.meterRegistry = meterRegistry;
+        this.shouldCreateLongTaskTimer = true;
+    }
+
+    /**
+     * Creates the handler with the defined Meters to use when the handler records the
+     * received events.
+     * @param meterRegistry the MeterRegistry to use
+     * @param metersToIgnore the Meters that should not be created when events are handled
```
shakuzen, line 61: I would change "events" to "Observations" so people aren't confused with `Event`s that are part of Observations.

shakuzen (approved):
Overall looks good. I left some naming polish suggestions. I worry a bit it will be surprising for users they can only pass LTTs to be ignored given the API, but at least not being able to pass other things makes it clear what is supported. I suppose we can tackle expanding the support given user requests.

PR #4756, opened 2024-02-19
JettyClientMetrics compatibility with Jetty 12

### LINKED ISSUE #4609: Make JettyClientMetrics compatible with Jetty12
Micrometer JettyClientMetrics not working with Jetty 12, getting NoClassDefFoundError  exception, since Jetty 12 renamed packages, 
https://github.com/micrometer-metrics/micrometer/blob/1.12.x/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyClientMetrics.java

See imports below, which makes reference to old packages found in Jetty11
import org.eclipse.jetty.client.**api**.ContentProvider;
import org.eclipse.jetty.client.**api**.Request;
import org.eclipse.jetty.client.**api**.Result;

**Rationale**
Springboot [3.2.x](https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html#appendix.dependency-versions) was released adding Jetty12 support, but latest micrometer-core version 1.12.2 is not supporting them.

**Additional context**
This behavior is similar to tickets:
https://github.com/micrometer-metrics/micrometer/issues/4261
https://github.com/qos-ch/logback/issues/719
https://github.com/akkinoc/logback-access-spring-boot-starter/issues/374

shakuzen:
Copy JettyClientMetrics and related classes to a micrometer-jetty12 module from micrometer-core (based on Jetty 9) and migrate the Jetty API accordingly.

https://eclipse.dev/jetty/documentation/jetty-12/programming-guide/index.html#pg-migration-11-to-12

Resolves #4609

### DIFF
--- a/micrometer-jetty12/src/main/java/io/micrometer/jetty12/client/JettyClientKeyValues.java
+++ b/micrometer-jetty12/src/main/java/io/micrometer/jetty12/client/JettyClientKeyValues.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2024 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.jetty12.client;
+
+import io.micrometer.common.KeyValue;
+import io.micrometer.common.lang.Nullable;
+import io.micrometer.common.util.StringUtils;
+import io.micrometer.core.instrument.binder.http.Outcome;
+import org.eclipse.jetty.client.Request;
+import org.eclipse.jetty.client.Result;
+import org.eclipse.jetty.http.HttpStatus;
+
+import java.util.function.BiFunction;
+import java.util.regex.Pattern;
+
+/**
+ * Factory methods for {@link KeyValue} associated with a request-response exchange that
+ * is handled by Jetty {@link org.eclipse.jetty.client.HttpClient}.
+ *
+ * @author Jon Schneider
+ * @since 1.13.0
+ */
+public final class JettyClientKeyValues {
+
+    private static final KeyValue URI_NOT_FOUND = KeyValue.of("uri", "NOT_FOUND");
+
+    private static final KeyValue URI_REDIRECTION = KeyValue.of("uri", "REDIRECTION");
+
+    private static final KeyValue URI_ROOT = KeyValue.of("uri", "root");
+
+    private static final KeyValue EXCEPTION_NONE = KeyValue.of("exception", "None");
+
+    private static final KeyValue EXCEPTION_UNKNOWN = KeyValue.of("exception", "UNKNOWN");
+
+    private static final KeyValue METHOD_UNKNOWN = KeyValue.of("method", "UNKNOWN");
+
+    private static final KeyValue HOST_UNKNOWN = KeyValue.of("host", "UNKNOWN");
+
+    private static final KeyValue STATUS_UNKNOWN = KeyValue.of("status", "UNKNOWN");
+
+    private static final Pattern TRAILING_SLASH_PATTERN = Pattern.compile("/$");
+
+    private static final Pattern MULTIPLE_SLASH_PATTERN = Pattern.compile("//+");
+
+    private static final KeyValue OUTCOME_UNKNOWN = KeyValue.of("outcome", "UNKNOWN");
+
+    private JettyClientKeyValues() {
+    }
+
+    /**
+     * Creates a {@code method} KeyValue based on the {@link Request#getMethod() method}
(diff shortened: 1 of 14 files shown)


shakuzen:
The micrometer-samples-jetty12 module can be removed after this and #4753 are merged and the test there is added to the micrometer-jetty12 module. A question that's tangential to this pull request is whether we copy `JettyConnectionMetrics` from micrometer-core to micrometer-jetty12 or keep it there undeprecated since it works with Jetty 9 through 12.

micrometer-jetty12/src/test/java/io/micrometer/jetty12/client/JettyClientMetricsTest.java
```diff
@@ -0,0 +1,161 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.CountDownLatch;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@WireMockTest
```
shakuzen, line 38: I reworked the tests to use WireMock since that seems easier and more similar to other tests we have.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/client/JettyClientTagsProvider.java
```diff
@@ -0,0 +1,60 @@
...
+     * For client metric to be usefully aggregable, we must be able to time everything
+     * that goes to a certain endpoint, regardless of the parameters to that endpoint.
+     * @param result The result which also contains the original request.
+     * @return A URI pattern with path variables and query parameter unsubstituted.
+     * @deprecated since 1.11.0 in favor of
+     * {@link JettyClientMetrics#builder(MeterRegistry, BiFunction)} to configure the uri
+     * pattern function
+     */
+    @Deprecated
+    String uriPattern(Result result);
```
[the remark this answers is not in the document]
shakuzen: I'm not sure there is anything to get rid of after the removal of the deprecated constructor and static builder method. In fact, I un-deprecated this in the micrometer-jetty12 module. Ultimately, if we support instrumentation with just a MeterRegistry, we need a way to customize tags and `JettyClientTagsProvider` is that way. The Observation-based instrumentation doesn't use it but at least they can share the BiFunction by default that is passed to the static builder method.

shakuzen:
@joakime if you have some time to take a look at this, it's the client-side instrumentation updated for Jetty 12. It's mostly a copy of the existing Jetty client instrumentation for Jetty 9, but if you spot anything that looks unusual or could be improved, please let us know.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/client/JettyClientKeyValues.java
```diff
@@ -0,0 +1,163 @@
...
+     * {@code result}.
+     * @param result the request result
+     * @return the outcome KeyValue derived from the status of the response
+     */
+    public static KeyValue outcome(@Nullable Result result) {
+        if (result == null) {
+            return OUTCOME_UNKNOWN;
+        }
+        return Outcome.forStatus(result.getResponse().getStatus()).asKeyValue();
+    }
```
joakime, line 161: With HTTP/1 the existence of the status code is just what the response commit buffer had, not that the request/response is complete or successful (depending on the mode used in HTTP/1 there could be a socket exception, or even just a normal socket closure with incomplete data see "close-delimited message" on the HTTP/1 specs). 

With HTTP/2 and HTTP/3 the status code is just what was seen on the first buffer.
Those responses can also be canceled via the HTTP/2 and HTTP/3 protocols and be incomplete, these would be seen as protocol specific exceptions.

joakime:
I don't see protocol/version specific details in these metrics.

I'm especially concerned about that kind of metric in light of PR https://github.com/jetty/jetty.project/pull/11368

shakuzen:
> I don't see protocol/version specific details in these metrics.

Thanks for pointing that out and sharing the Jetty PR. Right now it may be more straightforward to tag the protocol, but it sounds like after that PR, there may be layering of protocols and that makes it complicated to map that to tags. It's certainly something to explore, though. Maybe it'd be best to open an issue for it and track it separately so we don't block merging this on that. What do you think?

micrometer-jetty12/src/main/java/io/micrometer/jetty12/client/JettyClientKeyValues.java
shakuzen: Those are interesting scenarios and I don't think we're properly handling them right now. It sounds like something we could write a test for and then figure out how to (or if we can) improve all the different HTTP instrumentations (/cc @bclozel). For HTTP client instrumentation we have a small suite of tests that we run all of them against; see [HttpClientTimingInstrumentationVerificationTests](https://github.com/micrometer-metrics/micrometer/blob/63cfcbccd57dabb9cfffe77b613626ded47803d2/micrometer-test/src/main/java/io/micrometer/core/instrument/HttpClientTimingInstrumentationVerificationTests.java) which is extended in this pull request to run the instrumentation against it in Jetty12ClientTimingInstrumentationVerificationTests.

Are there existing tests in Jetty for this scenario that we could get inspiration from to add to our test suite?

We haven't done a good job of testing with HTTP/2 or HTTP/3 so far, so that's another point of improvement.

jonatan-ivanov (approved):

Issue #3236, opened 2022-06-17
JvmHeapPressureMetrics have incorrect base unit


fkrull:
**Describe the bug**
The unit for the metrics "jvm.memory.usage.after.gc" and "jvm.gc.overhead" in `JvmHeapPressureMetrics.java` is set as percent, but the values are in range [0...1]. This confuses Datadog (and possibly other registries?) which shows these metrics as e.g. "0.1%" instead of 10%.

**Environment**

 - Micrometer 1.9.0
 - Datadog

**Expected behavior**
The unit should match the value range so that Datadog can display it properly. Datadog seems to support a "fraction" unit that works for this range.

marcingrzejszczak:
We can't change the values in a way that would be backward compatible. Are you saying that what we send to Datadog is in a format that is not correct?

fkrull:
No, I'm saying that the value range does not match the unit. The values for that metric are 0-1, but the unit is "percent" so Datadog expects 0-100. So then when you graph these metrics, Datadog will automatically show you readings like "1% GC overhead" because of the unit it receives, and then you're like, "oh, 1%, that's not very much at all" when in fact it is 100% because the value range doesn't match the unit. And then you have to know that the value range and the unit don't match and multiply with 100 or explicitly override the unit to make it show correct things.

(Unless any of this changed in the past year and a half, I can't really check right now.)

izeye:
It seems to be technically wrong and misleading as reported by @fkrull. I created https://github.com/micrometer-metrics/micrometer/pull/4765 to try to resolve it.

shakuzen:
Thanks for the clarification @fkrull and sorry the issue was closed. Thanks to @izeye's pull request, going forward the unit will not be specified. If you would like to test it out, the changes are available in 1.3.0-SNAPSHOT in the snapshot repo now before the 1.3.0-M2 pre-release.

izeye:
@shakuzen Typos in the versions for 1.13.0-SNAPSHOT and 1.13.0-M2.

fkrull:
@shakuzen @izeye I can't really test this out any more, but thanks for getting this fixed!

PR #4765, opened 2024-02-22
Drop wrong base units in JvmHeapPressureMetrics

izeye:
This PR drops wrong base units in `JvmHeapPressureMetrics`.

See gh-3236


shakuzen (approved):
It would be nice to replace it with a correct unit that works for all backends. However, I'm not sure all backends will agree on what that will be. I see it was suggested in the linked issue using `fraction` but Prometheus for instance seems [to suggest](https://prometheus.io/docs/practices/naming/#base-units) `ratio` instead. Given that, maybe no unit is best.

Issue #4754, opened 2024-02-15
Better support working with Observation from Kotlin


ilya40umov:
Right now (as far as I can tell), there is no extra support for using `Observation` API from Kotlin code.

For example, to create a new `Observation` in normal, blocking code, one first needs to invent a convenience function like this:
```
private fun <T> ObservationRegistry.observeNotNull(name: String, block: () -> T): T =
    Observation.start(name, this).observe(Supplier { block() })!!
```

And when it comes to wrapping a suspending block of code within `Observation` it gets even more complicated:
```
private suspend fun <T> ObservationRegistry.observe(name: String, block: suspend () -> T): T {
    val observation = Observation.createNotStarted(name, this)
    observation.start()
    return try {
        mono(observationRegistry.asContextElement() + Dispatchers.Unconfined) {
            block()
        }.contextWrite { context ->
            // XXX this seems to be the safest option of making the new observation current
            context.put(ObservationThreadLocalAccessor.KEY, observation)
        }.awaitSingle()
    } catch (error: Throwable) {
        throw error
    } finally {
        // XXX or maybe it's safer to stop this in `doOnTerminate` of the created mono?
        observation.stop()
    }
}
``` 

At the same time, when working with Otel APIs directly, there is an option to construct a span/context and put it into scope via Otel's [ContextExtensions.kt](https://github.com/open-telemetry/opentelemetry-java/blob/main/extensions/kotlin/src/main/kotlin/io/opentelemetry/extension/kotlin/ContextExtensions.kt)

shakuzen:
Thanks for opening the issue. Is this something you would be willing to make a pull request for, @ilya40umov? I don't regularly use Kotlin so it will probably be easier for folks who do to suggest the changes.

ilya40umov:
@shakuzen I'll definitely try looking into this (not sure sure if I'll have time for this next week though).

ilya40umov:
So far I was able to improve a bit upon the second extension function I posted originally, so now it looks as follows:

```
private suspend fun <T> ObservationRegistry.observe(name: String, block: suspend () -> T): T {
    val observation = Observation.start(name, this)
    val coroutineContext = currentCoroutineContext().minusKey(Job.Key)
    return mono(coroutineContext + observationRegistry.asContextElement()) {
        block()
    }.contextWrite { context ->
        context.put(ObservationThreadLocalAccessor.KEY, observation)
    }.doOnTerminate {
        observation.stop()
[...]

PR #4772, opened 2024-02-26
Kotlin-specific ObservationRegistry API

ilya40umov:
This PR introduces a couple of extension functions on `ObservationRegistry` to help using `Observation` API from Kotlin code.

- `ObservationRegistry.observeAndGet` allows to observe a non-suspending block of code:
  - this method is analogous to `Observation.observe`, but accepts a block of code instead of expecting a Java Supplier
  - this method will work with both nullable and non-nullable generic type (while usage of Supplier makes Kotlin compiler think the result of the nested block is always nullable)
- `ObservationRegistry.observeAndAwait` allows to observe a suspending block of code:
  - this method is again similar to `Observation.observe`, but will accept a suspending block of code
  - by using `openScope().use { observationRegistry.asContextElement() }` it manages to create an instance of `KotlinObservationContextElement` with the new observation captured as "current"

Initial idea: Why add those to `ObservationRegistry` and not to `Observation`? Mostly for consistency. I.e. `observeAndAwait` needs  access to `ObservationRegistry` anyway, so it's easy to add it to the registry as an extension. While `observeAndGet` could indeed be made into an extension function on `Observation`, however it still needs to be called something other than `observe`, as otherwise Intellij can't resolve which `observe` version the code is referring to.

**Update:** I realized that Observation also contains registry it was created for, so I created an alternative PR which adds both of the extension functions to `Observation` class: https://github.com/micrometer-metrics/micrometer/pull/4823 

gh-4754

### DIFF
--- a/micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObserveAndAwait.kt
+++ b/micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObserveAndAwait.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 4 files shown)


ilya40umov:
@shakuzen would you be the one to ask for review on this PR? 🙏

shakuzen:
> would you be the one to ask for review on this PR?

I will try to find someone to take a look. Some folks are off this week, though.

ilya40umov:
I realized that `Observation` also contains the registry it was created for, so I created an alternative PR which adds both of the extension functions to Observation class as a possible alternative: https://github.com/micrometer-metrics/micrometer/pull/4823

Feel free to choose which one would make more sense for Micrometer project to include.

shakuzen:
> I realized that `Observation` also contains the registry it was created for, so I created an alternative PR which adds both of the extension functions to Observation class as a possible alternative: #4823

Thanks. I think `Oberservation` would be the better place for something like this. Closing this in favor of #4823

Issue #3877, opened 2023-06-06
ArrayIndexOutOfBoundsException when using LongTaskTimer


rdehuyss:
**Describe the bug / it may be a question**
I'm seeing `ArrayIndexOutOfBoundsException` in my logs when using `LongTaskTimer`.

**Environment**
 - Micrometer version 1.11.0
 - Micrometer registry: SimpleMeterRegistry
 - OS: macOS
 - Java version: `openjdk version "17.0.1" 2021-10-19`

**To Reproduce**
```java
    @Test
    void reproduceIssue() throws InterruptedException {
        SimpleMeterRegistry simpleMeterRegistry = new SimpleMeterRegistry();

        LongTaskTimer timer = LongTaskTimer.builder("jobrunr.jobs")
                .publishPercentiles(0.25, 0.5, 0.75, 0.8, 0.9, 0.95)
                .publishPercentileHistogram()
                .register(simpleMeterRegistry);
        LongTaskTimer.Sample start = timer.start();
        Thread.sleep(3640);
        System.out.println(simpleMeterRegistry.getMetersAsString());
        Thread.sleep(2640);
        System.out.println(start.stop());

        System.out.println(simpleMeterRegistry.getMetersAsString());
    }
```

This results in the following output:
```
08:46:23.327 [Test worker] WARN  i.m.c.i.internal.DefaultGauge - Failed to apply the value function for the gauge 'jobrunr.jobs.percentile'. Note that subsequent logs will be logged at debug level.
java.lang.ArrayIndexOutOfBoundsException: Index 5 out of bounds for length 5
	at io.micrometer.core.instrument.distribution.HistogramGauges.lambda$new$9(HistogramGauges.java:119)
	at io.micrometer.core.instrument.internal.DefaultGauge.value(DefaultGauge.java:53)
	at io.micrometer.core.instrument.Measurement.getValue(Measurement.java:54)
	at io.micrometer.core.instrument.simple.SimpleMeterRegistry.toString(SimpleMeterRegistry.java:206)
	at io.micrometer.core.instrument.simple.SimpleMeterRegistry.lambda$toString$1(SimpleMeterRegistry.java:195)
jobrunr.jobs(LONG_TASK_TIMER)[]; active_tasks=1.0, duration=3.651603291 seconds
jobrunr.jobs.percentile(GAUGE)[phi='0.25']; value=3.653785708 seconds
[...]

PR #4782, opened 2024-03-01
[Dynatrace] Add artificial zero percentile in Meter creation rather than a MeterFilter

### LINKED ISSUE #4750: Zero percentile can be missing from Dynatrace meters where expected
**Describe the bug**
When using the Spring Boot configuration in application.yaml to specify percentiles for Meters, that configuration will overwrite all other configurations.

I am filing this as a bug, but am not sure if this is expected behavior. It is not the behavior I expected, so I know of at least one example where this behavior breaks the DynatraceExporter. 

**Environment**
The bug has been present in Micrometer versions used in Spring Boot for a long time. Due to it only showing up in a situation where percentiles are configured in multiple places, this is a rare one to show up. I know that it was present in 1.9.x, and it is possible to reproduce with versions up to the one on main right now. 
 - Micrometer version `all released versions` 
 - Micrometer registry `SimpleMeterRegistry, DynatraceMeterRegistry`, probably all other registries as well, but I just tested with these two. 
 - OS: OS independent
 - Java version: java version independent

**To Reproduce**
See the reproducer at https://github.com/pirgeo/springboot-percentileconfig. There is a step-by-step example in the readme.

**Expected behavior**
When using multiple ways of configuring percentiles, I would expect percentile values to be merged, not overwritten.

pirgeo:
resolves #4750

See the issue for more context. This is mostly a refactoring of where the percentiles are registered (in the MeterFilter vs. in the creation of Meters)

CC @shakuzen

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
@@ -145,7 +145,7 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
     protected LongTaskTimer newLongTaskTimer(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig) {
         if (useDynatraceSummaryInstruments) {
             return new DynatraceLongTaskTimer(id, clock, exporter.getBaseTimeUnit(), distributionStatisticConfig,
-                false);
+                    false);
         }
 
         DistributionStatisticConfig config = distributionStatisticConfig;
@@ -179,7 +179,7 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
     }
 
     private DistributionStatisticConfig addZeroPercentileIfMissing(Meter.Id id,
-                                                                   DistributionStatisticConfig distributionStatisticConfig) {
+            DistributionStatisticConfig distributionStatisticConfig) {
         double[] percentiles;
 
         double[] configPercentiles = distributionStatisticConfig.getPercentiles();
@@ -195,7 +195,8 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
             zeroPercentileMeterFilter.addMeterId(id);
         }
         else {
-            // Zero percentile is explicitly added to the config, no need to add it to drop list.
+            // Zero percentile is explicitly added to the config, no need to add it to
+            // drop list.
             return distributionStatisticConfig;
         }
 
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
@@ -60,6 +60,10 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
 
     private final boolean useDynatraceSummaryInstruments;
 
+    private final boolean shouldAddZeroPercentile;
+
+    private final DenyZeroPercentileMeterFilter zeroPercentileMeterFilter = new DenyZeroPercentileMeterFilter();
+
     private final AbstractDynatraceExporter exporter;
 
     @SuppressWarnings("deprecation")
@@ -73,17 +77,22 @@ public class DynatraceMeterRegistry extends StepMeterRegistry {
         super(config, clock);
 
         useDynatraceSummaryInstruments = config.useDynatraceSummaryInstruments();
+        // zero percentile is needed only when using the V2 exporter and not using
+        // Dynatrace summary instruments.
+        shouldAddZeroPercentile = config.apiVersion() == DynatraceApiVersion.V2 && !useDynatraceSummaryInstruments;
 
         if (config.apiVersion() == DynatraceApiVersion.V2) {
             logger.info("Exporting to Dynatrace metrics API v2");
             this.exporter = new DynatraceExporterV2(config, clock, httpClient);
-            // Not used for Timer and DistributionSummary in V2 anymore, but still used
-            // for the other timer types.
-            registerMinPercentile();
         }
         else {
             logger.info("Exporting to Dynatrace metrics API v1");
             this.exporter = new DynatraceExporterV1(config, clock, httpClient);
+        }
+
+        if (shouldAddZeroPercentile) {
+            // zero percentiles automatically added should not be exported.
+            this.config().meterFilter(zeroPercentileMeterFilter);
         }
 
         start(threadFactory);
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/DynatraceMeterRegistry.java
```diff
@@ -121,77 +132,70 @@ protected Timer newTimer(Meter.Id id, DistributionStatisticConfig distributionSt
...
+            return metersWithArtificialZeroPercentile.contains(id.getName()) && "0".equals(id.getTag("phi"));
+        }
+
+        @Override
+        public MeterFilterReply accept(Meter.Id id) {
+            return hasArtificialZeroPercentile(id) ? DENY : NEUTRAL;
+        }
+
+        public void addMeterId(Meter.Id id) {
+            metersWithArtificialZeroPercentile.add(id.getName() + ".percentile");
```
shakuzen, line 166: Why not use a Set of `Id` instead of this?

shakuzen (commented):
Overall looks good, I think. You can rebase now with the changes to LTT. Are you planning to add some tests?

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/DynatraceMeterRegistryTest.java
```diff
@@ -151,6 +153,142 @@ void shouldNotTrackPercentilesWithDynatraceSummary() throws Throwable {
...
+
+        // create instruments with an explicit 0 percentile. This should be exported.
+        Timer timer = Timer.builder("my.timer").publishPercentiles(0, 0.5, 0.99).register(registry);
+        DistributionSummary distributionSummary = DistributionSummary.builder("my.ds")
+            .publishPercentiles(0, 0.5, 0.99)
+            .register(registry);
+        // For LongTaskTimer, the 0 percentile is not tracked as it doesn't clear the
+        // "interpolatable line" threshold defined in DefaultLongTaskTimer#takeSnapshot().
+        // see shouldTrackPercentilesWhenDynatraceSummaryInstrumentsNotUsed for a test
+        // that exports LongTaskTimer percentiles
```
pirgeo, line 259: This comment is referring to these three lines: https://github.com/micrometer-metrics/micrometer/blob/e3329b8f49c4e0eeec6fc24b10edd9f3c16af040/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L143-L145
They drop the 0 percentile (`percentilesAboveInterpolatableLine` won't contain the `0` percentile anymore)

I'm not sure if this is a bug or working as intended.

pirgeo: I dug a little deeper, and I think the problem lies in the fact that the 0% HistogramGauge is never registered.
1. The LongTaskTimer is registered, with the explicit zero percentile.
2. The histogram gauges get registered. To do so, an initial snapshot is created: 
https://github.com/micrometer-metrics/micrometer/blob/e3329b8f49c4e0eeec6fc24b10edd9f3c16af040/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/HistogramGauges.java#L100
3. At this point, the number of active tasks is 0. The LongTaskTimer is currently in the process of being registered. Thus, the calculation for `percentilesAboveInterpolatableLine` looks as follows:
https://github.com/micrometer-metrics/micrometer/blob/e3329b8f49c4e0eeec6fc24b10edd9f3c16af040/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L143-L145
Since the number of `activeTasks` is 0, the calculation in the `filter` effectively evaluates to `p * (0 + 1) > 0`, which is equivalent to `p > 0`. This, of course, is a condition that the 0 percentile (where `p==0`) explicitly does not fulfil. The 0 percentile is dropped here.
4. Because the 0 percentile was dropped in the initial `takeSnapshot`, it will not be registered as a HistogramGauge because the `percentileValues` call has only percentiles > 0: https://github.com/micrometer-metrics/micrometer/blob/e3329b8f49c4e0eeec6fc24b10edd9f3c16af040/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/HistogramGauges.java#L103
5. Since the 0th percentile gauge is never registered, it is also never exported. 

Later, when the app is running and there are active tasks, this is caught in the `if` here: https://github.com/micrometer-metrics/micrometer/blob/e3329b8f49c4e0eeec6fc24b10edd9f3c16af040/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L151
However, on the initial `takeSnapshot`, https://github.com/micrometer-metrics/micrometer/blob/e3329b8f49c4e0eeec6fc24b10edd9f3c16af040/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L161-L165 evaluates to an empty list, as there are no active tasks during registration, and the 0th percentile is never registered.

shakuzen: We do have at least one bug in this area of the code. See https://github.com/micrometer-metrics/micrometer/issues/3877. I'll take a closer look tomorrow.

shakuzen: Just wanted to update here that I didn't get a chance to look into it more today due to some other work that needs to get done first. I'll come back to this once that's done.

pirgeo:
Hi @shakuzen, 
I just realized that we never merged this PR, so I'd like to revive it again. I guess I should retarget it against 1.14 now?

pirgeo:
@shakuzen @jonatan-ivanov Do you think we could revive this one and get it into 1.14? Anything to be done from my side to do so?

shakuzen (approved):

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/DynatraceMeterRegistryTest.java
shakuzen: It'll take some more time to look into the LTT issue, I think. But the changes in this pull request don't change the status quo, I think. So nothing is worse related to the bug. I think we're fine to merge these changes and follow-up on #3877 separately. Let me know if I'm missing anything.

pirgeo: That's right, there is no change in the status quo, it only fixes a bug where the 0 percent quantile wasn't added automatically when explicitly specifying percentiles.

pirgeo:
Thank you for taking a look!

PR #4823, opened 2024-03-02
Kotlin-specific Observation API

ilya40umov:
This PR introduces a couple of extension functions on `Observation` to help using it from Kotlin code.
It's an alternative version of the following PR: https://github.com/micrometer-metrics/micrometer/pull/4772, which would introduce the same API, but on `ObservationRegistry`-level.

- `Observation.observeAndGet` allows to observe a non-suspending block of code:
  - this method is analogous to `Observation.observe`, but accepts a block of code instead of expecting a Java Supplier
  - this method will work with both nullable and non-nullable generic type (while usage of Supplier makes Kotlin compiler think the result of the nested block is always nullable)
- `Observation.observeAndAwait` allows to observe a suspending block of code:
  - this method is again similar to `Observation.observe`, but will accept a suspending block of code
  - by using `openScope().use { observationRegistry.asContextElement() }` it manages to create an instance of `KotlinObservationContextElement` with the new observation captured as "current"

Fixes gh-4754


ilya40umov:
Hm. Looks like some unrelated tests are failing `MongoMetricsCommandListenerTest > shouldCreateFailedCommandMetricWithCustomSettings() `.  🤔

micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObserveAndAwait.kt
```diff
@@ -0,0 +1,47 @@
...
+ * - Stops the [Observation]
+ *
+ * @param block the block of code to be observed
+ * @return the result of executing the provided block of code
+ */
+suspend fun <T> Observation.observeAndAwait(block: suspend () -> T): T {
+    start()
+    return try {
+        withContext(
+            openScope().use { observationRegistry.asContextElement() },
```
shakuzen, line 37: A consideration here is that this implementation depends on the context-propagation library, while Observation and ObservationRegistry don't directly - they optionally depend on it. Since we added the extension function `asContextElement()` to `ObservationRegistry`, that gives precedent to this kind of thing, but I'm not sure we should do this. That said, I don't know what a good alternative is. Basically, the issue is that this would fail at runtime if the user didn't have context-propagation on their classpath, which could happen if say they only care about producing metrics from Observation and not tracing. It feels bad to have a class that otherwise can be used without a library except an API (extension function) that will fail without a specific dependency.

shakuzen (commented):
Thank you for the pull request, and sorry it took so long to get reviewed. The PR looks reasonable but there's one issue I think we need to figure out what, if anything, we can/should do. I'm raising it with the team.

micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObserveAndAwait.kt
jensbaitingerbosch: Is there any update on this issue? I stumbled accross that PR as I wanted to provied a solution for the same problem. On one hand, a solution that is not perfect is better then no solution at all, on the other hand, @shakuzen, what 'context-propagation' api are you talking about? There is the coroutines-core (org.jetbrains.kotlinx:kotlinx-coroutines-core), which I would have anyway in the project if I used coroutines and it could be documented that this function needs that library.

micrometer-core/src/main/kotlin/io/micrometer/core/instrument/kotlin/ObserveAndAwait.kt
```diff
@@ -0,0 +1,47 @@
...
+ * - Starts the [Observation]
+ * - Puts the [Observation] into [CoroutineContext]
+ * - Calls the provided [block] withing the augmented [CoroutineContext]
+ * - Signals the error to the [Observation] if any
+ * - Stops the [Observation]
+ *
+ * @param block the block of code to be observed
+ * @return the result of executing the provided block of code
+ */
+suspend fun <T> Observation.observeAndAwait(block: suspend () -> T): T {
```
jensbaitingerbosch, line 33: I suggest to make the receivers nullable and internally handle that there is no observation. 

Motivation is that the client can simply call ```
observationRegistry.currentObservation.observeAndAwait{
   // some business code
} 
```
and can be sure that the business code is executed regardless if there is a current observation or not
```suggestion
suspend fun <T> Observation?.observeAndAwait(block: suspend () -> T): T {
    if (this == null) return block()
```

Issue #4824, opened 2024-03-04
OSGi import of javax.annotation and javax.annotation.meta is incorrect


ponziani:
**Describe the bug**
The manifest file of micrometer-core and micrometer-commons contains the Import-Package "javax.annotation;version=
 "[3.0,4)",javax.annotation.meta;version="[3.0,4)". However, these are not the correct version ranges, as explained in https://github.com/jhy/jsoup/issues/1616#issuecomment-958684482

This results in our Apache Karaf application no longer starting after upgrading to Micrometer 1.11

`Caused by: org.apache.felix.resolver.reason.ReasonException: Unable to resolve micrometer-commons/1.11.9: missing requirement [micrometer-commons/1.11.9] osgi.wiring.package; filter:="(&(osgi.wiring.package=javax.annotation)(version>=3.0.0)(!(version>=4.0.0)))"
        at org.apache.felix.resolver.Candidates$MissingRequirementError.toException(Candidates.java:1341)
        ... 17 more
`

**Environment**
 - Micrometer version 1.11.9
 - Java version: 17

**Additional context**
As suggested in the jsoup-link, the import could be made optional.

shakuzen:
Resolved by #4825

PR #4825, opened 2024-03-04
OSGi import of javax.annotation is incorrect #4824

ponziani:
This PR fixes the problem described in https://github.com/micrometer-metrics/micrometer/issues/4824 : `The manifest file of micrometer-core and micrometer-commons contains the Import-Package "javax.annotation;version=
"[3.0,4)",javax.annotation.meta;version="[3.0,4)".`


shakuzen:
@driessamyn do you have time to review this?

driessamyn (approved):
This looks correct. 👍

ponziani:
Thanks. I wasn't sure if we could even exclude the javax.annotation package. Do you know, or are you able to run a decent test?

Also, I had some troubles because the com.sun.management is in the Import-Packages as well. I managed to work around it in my Karaf application, but if that import is also not really necessary, it would be better to leave it out as well.

shakuzen:
I'll merge for now, but if @driessamyn has a chance to address @ponziani's question above, we can address it post merge if there is anything to address. Thank you both.

Issue #4693, opened 2024-02-02
"Failed to publish metrics to OTLP receiver" error message contains no actionable context


climategadgets:
**Describe the bug**
Under some circumstances, a message `Failed to publish metrics to OTLP receiver` with an exception will be logged, usually observed with `java.net.ConnectException: Connection refused` as a root cause.

This message does not contain any information that allows to determine where exactly the meter registry is trying to connect, or any other relevant context, for that matter. Since it is possible that the registry is being automatically instantiated by frameworks (Quarkus, Spring, etc.) and that process is usually not logged, the only way to fix the problem is to guess what configuration item needs to be fixed.

**Environment**

Environment agnostic, here: https://github.com/micrometer-metrics/micrometer/blob/061dbf69e908ce530e11b6e34082b01197263df6/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java#L182

**To Reproduce**
How to reproduce the bug:

Have the framework instantiate the registry with no configuration present, or wrong configuration

**Expected behavior**
Error message must contain enough information to determine what needs to be changed to fix the problem.
* Minimum: `Connection to $host:$port refused`
* Preferred: `Connection to $host:$port refused, check the value of x.y.z configuration keyword`

It is understood that "connection refused" may not be the only root cause here. This doesn't change the problem definition - not enough context information to fix the problem.

shakuzen:
Thank you for the detailed issue. It sounds like a good enhancement to me. Would you be up for making a pull request for it?

climategadgets:
> Thank you for the detailed issue. It sounds like a good enhancement to me. Would you be up for making a pull request for it?

Let me see what I can do.

climategadgets:
Below is a quick fix; sure thing it needs test cases but `OtlpMeterRegistry` is pretty closed up (`httpSender`, the main source of problems, can't be easily injected) so I wanted to make sure that this goes in the direction you guys approve before investing time in tests.

Please advise.

```
diff --git a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
index ef3eb90d1..0131c2565 100644
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
[...]

PR #4830, opened 2024-03-05
Announce OtlpMeterRegistry configuration found at startup

climategadgets:
**Existing Behavior**
`OtlpMeterRegistry` says nothing at startup.

**Updated Behavior**
`OtlpMeterRegistry` emits a startup message allowing to identify and troubleshoot the OTLP receiver endpoint.

**Additional Context**
#4693 

@pivotal-cla This is an Obvious Fix

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
@@ -125,6 +125,15 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
             this.meterPollingService.scheduleAtFixedRate(this::pollMetersToRollover, getInitialDelay(),
                     config.step().toMillis(), TimeUnit.MILLISECONDS);
         }
+
+        if (config.enabled()) {
+            sayHello();
+        }
+    }
+
+    private void sayHello() {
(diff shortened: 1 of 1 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -135,6 +135,15 @@ public void start(ThreadFactory threadFactory) {
...
                     config.step().toMillis(), TimeUnit.MILLISECONDS);
         }
+
+        if (config.enabled()) {
+            sayHello();
+        }
+    }
+
+    private void sayHello() {
+        logger.info("Publishing metrics for {} to {} with resource attributes {}", getClass().getSimpleName(),
```
shakuzen, line 145: Right now this will result in two log messages on start. One from `PushMeterRegistry` like:
```
publishing metrics for OtlpMeterRegistry every 1m
```

And now this also.

```
Publishing metrics for OtlpMeterRegistry to http://localhost:50901/v1/metrics with resource attributes {}
```

I think it may be confusing to users to see both messages. They may think two registries are publishing. I'll open a proposal to make the log message on start from PushMeterRegistry customizable and then this can be rebased on top of that if we merge it.

shakuzen: #4848 is now merged and these changes can be adapted to use that.

Issue #4856, opened 2024-03-15
Do not re-apply MeterFilters to IDs when registering if unnecessary


lenin-jaganathan:
**Please describe the feature request.**
To interact with a meter (recording values), the preferred pattern documented is to use the Builder to create a meter and interact with it. For example, the following code would create a counter,

``` java
Counter counter = Counter
    .builder("counter")
    .baseUnit("beans") // optional
    .description("a description of what this counter does") // optional
    .tags("region", "test") // optional
    .register(registry);
```

Once the counter is created we would call the `counter.increment()` to increment the counter. For a simple counter like this, we can save it in a variable and call the increment method whenever we want to increment. Now, if we want to add a tag whose value will be available only during runtime, we cannot store it in a simple variable. There are two ways to do it,
1. Directly call the increment method on the object returned by the builder. Since, the registry will make sure that only one meter will be registered for a `name` and `tag` set we would get the same object all the time and can call increment on it. Again, so simple it seems but in a real-world scenario, applications would be having meterfilters registered which will have to be run every single time we invoke the builder. Since, `Meter.Id` is immutable we keep on creating new meters if we want to add tags/modify tags/modify the name, etc. But all the code runs and would eventually return an already created meter in the registry. This becomes a huge overhead in the real-world use case where we interact with meters thousands of times per second and generate a very high amount of garbage data.
2. To work around the problem in `1`, we tried [a few different things](https://github.com/lenin-jaganathan/micrometer-benchmark/tree/main) and eventually used a `HashMap` to store the different tags as the key and the returned meter as value. So, the next time you need the meter for a tag and name combination we would avoid all the intermediate steps and directly return the actual meter. This surely had a huge performance gain for us which is available in the above-mentioned repo. It opened up a few issues like,
i. What if a meter is removed from the registry? Then our cache will be holding reference to a meter that is no longer reported and would be incrementing that. We then listened to the onMeterRemoved and removed the meters from the cache.
ii. What if someone added a meterfilter? That would basically invalidate all the meters. Micrometer doesn't enforce strict rules about when a filter can be added. But we sort approached meter registry state as a immutable thing and didn't want to invalidate the entire cache because that micrometer doesn't invalidate meters when meter filters are registered.

To address the above mentioned problem and to improve the overall performance of micrometer meters, a built-in way to map the preMapped Id's to mappedId's with O(1) retrieval time with minimal garbage generation could be introduced.

Reference:
[...]

PR #4857, opened 2024-03-15
Improve performance of meter interaction

lenin-jaganathan:
Fixes https://github.com/micrometer-metrics/micrometer/issues/4856,

This PR introduces the following changes,
* Keeps a map to do the look-up between 'Id` without meter filters to the actual meter. When create is called on a meter, we first check if a mapping exists and immediately return the mapped meter from the actual map that holds the meters.
* When meter filters are added to the registry, the meters registered until this point are maintained in the set and when the create operation happens on these meters, we re-compute the mapping similarly to keep the existing behavior.

TODO:

- [x] Add additional tests to MeterRegistryTest

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -103,6 +103,8 @@ public abstract class MeterRegistry {
 
     private final Map<Id, Meter> preFilterIdToMeterMap = new HashMap<>();
 
+    private final Set<Id> stalePreFilterIds = ConcurrentHashMap.newKeySet();
+
     /**
      * Map of meter id whose associated meter contains synthetic counterparts to those
      * synthetic ids. We maintain these associations so that when we remove a meter with
@@ -617,7 +619,7 @@ public abstract class MeterRegistry {
             Function<Meter.Id, ? extends Meter> noopBuilder) {
 
         Meter m = preFilterIdToMeterMap.get(originalId);
-        if (m != null) {
+        if (m != null && !unmarkStaleId(originalId)) {
             return m;
         }
 
@@ -664,6 +666,15 @@ public abstract class MeterRegistry {
         }
 
         return m;
+    }
+
+    /**
+     * Marks the ID as no longer stale if it is stale. Otherwise, does nothing.
+     * @param originalId id before any filter mapping has been applied
+     * @return {@code true} if the id is stale
+     */
+    private boolean unmarkStaleId(Id originalId) {
+        return !stalePreFilterIds.isEmpty() && stalePreFilterIds.remove(originalId);
     }
 
     private boolean accept(Meter.Id id) {
@@ -724,7 +735,14 @@ public abstract class MeterRegistry {
         if (meterMap.containsKey(mappedId)) {
             synchronized (meterMapLock) {
                 final Meter removedMeter = meterMap.remove(mappedId);
-                preFilterIdToMeterMap.values().removeIf(meter -> meter.equals(removedMeter));
+                Iterator<Meter> iterator = preFilterIdToMeterMap.values().iterator();
+                while (iterator.hasNext()) {
+                    Meter next = iterator.next();
+                    if (next.equals(removedMeter)) {
+                        stalePreFilterIds.remove(next.getId());
+                        iterator.remove();
+                    }
+                }
                 if (removedMeter != null) {
                     Set<Id> synthetics = syntheticAssociations.remove(mappedId);
                     if (synthetics != null) {
@@ -762,10 +780,6 @@ public abstract class MeterRegistry {
         /**
          * Append a list of common tags to apply to all metrics reported to the monitoring
          * system.
-         * <p>
-         * </p>
-         * <strong>NOTE: A no-op operation if meters are already registered to the
-         * registry.</strong>
          * @param tags Tags to add to every metric.
          * @return This configuration instance.
          */
@@ -777,10 +791,6 @@ public abstract class MeterRegistry {
          * Append a list of common tags to apply to all metrics reported to the monitoring
          * system. Must be an even number of arguments representing key/value pairs of
          * tags.
-         * <p>
-         * </p>
-         * <strong>NOTE: A no-op operation if meters are already registered to the
-         * registry.</strong>
          * @param tags MUST be an even number of arguments representing key/value pairs of
          * tags.
          * @return This configuration instance.
@@ -792,16 +802,13 @@ public abstract class MeterRegistry {
         /**
          * Add a meter filter to the registry. Filters are applied in the order in which
          * they are added.
-         * <p>
-         * </p>
-         * <strong>NOTE: A no-op operation if meters are already registered to the
-         * registry.</strong>
          * @param filter The filter to add to the registry.
          * @return This configuration instance.
          */
         public synchronized Config meterFilter(MeterFilter filter) {
             if (!meterMap.isEmpty()) {
                 logWarningAboutLateFilter();
+                stalePreFilterIds.addAll(preFilterIdToMeterMap.keySet());
             }
             MeterFilter[] newFilters = new MeterFilter[filters.length + 1];
             System.arraycopy(filters, 0, newFilters, 0, filters.length);
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -102,10 +102,12 @@ public abstract class MeterRegistry {
     private final Map<Id, Meter> meterMap = new ConcurrentHashMap<>();
 
     // writes guarded by meterMapLock, reads can be stale
-    private final Map<Id, Meter> preFilterIdToMeterMap = new HashMap<>();
+    // VisibleForTesting
+    final Map<Id, Meter> preFilterIdToMeterMap = new HashMap<>();
 
     // not thread safe; only needed when MeterFilter configured with Meters registered
-    private final Set<Id> stalePreFilterIds = new HashSet<>();
+    // VisibleForTesting
+    final Set<Id> stalePreFilterIds = new HashSet<>();
 
     /**
      * Map of meter id whose associated meter contains synthetic counterparts to those
@@ -742,11 +744,11 @@ public abstract class MeterRegistry {
         if (meterMap.containsKey(mappedId)) {
             synchronized (meterMapLock) {
                 final Meter removedMeter = meterMap.remove(mappedId);
-                Iterator<Meter> iterator = preFilterIdToMeterMap.values().iterator();
+                Iterator<Map.Entry<Id, Meter>> iterator = preFilterIdToMeterMap.entrySet().iterator();
                 while (iterator.hasNext()) {
-                    Meter next = iterator.next();
-                    if (next.equals(removedMeter)) {
-                        stalePreFilterIds.remove(next.getId());
+                    Map.Entry<Id, Meter> nextEntry = iterator.next();
+                    if (nextEntry.getValue().equals(removedMeter)) {
+                        stalePreFilterIds.remove(nextEntry.getKey());
                         iterator.remove();
                     }
                 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -101,9 +101,11 @@ public abstract class MeterRegistry {
      */
     private final Map<Id, Meter> meterMap = new ConcurrentHashMap<>();
 
+    // writes guarded by meterMapLock, reads can be stale
     private final Map<Id, Meter> preFilterIdToMeterMap = new HashMap<>();
 
-    private final Set<Id> stalePreFilterIds = ConcurrentHashMap.newKeySet();
+    // not thread safe; only needed when MeterFilter configured with Meters registered
+    private final Set<Id> stalePreFilterIds = new HashSet<>();
 
     /**
      * Map of meter id whose associated meter contains synthetic counterparts to those
@@ -619,7 +621,7 @@ public abstract class MeterRegistry {
             Function<Meter.Id, ? extends Meter> noopBuilder) {
 
         Meter m = preFilterIdToMeterMap.get(originalId);
-        if (m != null && !unmarkStaleId(originalId)) {
+        if (m != null && !isStaleId(originalId)) {
             return m;
         }
 
@@ -661,11 +663,16 @@ public abstract class MeterRegistry {
                     }
                     meterMap.put(mappedId, m);
                     preFilterIdToMeterMap.put(originalId, m);
+                    unmarkStaleId(originalId);
                 }
             }
         }
 
         return m;
+    }
+
+    private boolean isStaleId(Id originalId) {
+        return !stalePreFilterIds.isEmpty() && stalePreFilterIds.contains(originalId);
     }
 
     /**
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -94,11 +94,14 @@ public abstract class MeterRegistry {
 
     private final More more = new More();
 
-    // Even though writes are guarded by meterMapLock, iterators across value space are
-    // supported
-    // Hence, we use CHM to support that iteration without ConcurrentModificationException
-    // risk
+    /**
+     * Even though writes are guarded by meterMapLock, iterators across value space are
+     * supported. Hence, we use CHM to support that iteration without
+     * ConcurrentModificationException risk.
+     */
     private final Map<Id, Meter> meterMap = new ConcurrentHashMap<>();
+
+    private final Map<Id, Meter> preFilterIdToMeterMap = new HashMap<>();
 
     /**
      * Map of meter id whose associated meter contains synthetic counterparts to those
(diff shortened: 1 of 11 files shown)


shakuzen (commented):
Thank you very much for this pull request. It's a great idea for a performance gain for everyone with common usage scenarios and hopefully it will remove the need for bespoke caching when using Micrometer for particularly performance sensitive users.

That said, I'm a bit worried about the side effects of this as it currently is. The biggest thing we want to avoid is causing metrics to break in an unexpected way that is not caught until deployed to production.

Before this pull request, the issue with meters registered before a MeterFilter is configured is that they will not have that filter applied to them. If getOrCreate is called again, then the filter will be applied and a new meter will be created (assuming the filter alters the ID), and if a reference is not maintained to the first meter created (before the MeterFilter was configured), then the first meter will be essentially orphaned. Calls to `removeByPreFilterId` will not remove the orphaned meters.  Calls to `remove` would still work if a reference to the meter or id of the "orphaned" meter was used.

I worry some users may inadvertently be relying on this behavior and it may not be an issue for them if they're not maintaining a reference to a meter registered before a MeterFilter is configured.

Giving a concrete, albeit contrived, example below:

```java
@Test
void meterRegistrationBeforeAndAfterMeterFilterRegistration() {
    Counter c1 = registry.counter("counter");
    registry.config().meterFilter(MeterFilter.commonTags(Tags.of("common", "tag")));
    Counter c2 = registry.counter("counter");
    
    Timer t1 = registry.timer("timer");
    
    registry.remove(c1);
    registry.remove(c2);
}
```

Before this pull request, `c1` and `c2` will be different meters. `c2` and `t1` will have the common tags; `c1` will not. Each call to `remove` will remove the different meters.

After this pull request (as it currently is), `c1` and `c2` will be the same meter, without the common tags. The first remove call will remove the only counter registered. This is what I worry will be surprising and breaking behavior that won't necessarily get caught until it is too late.

I have some ideas I'd like to try to see if we can get to some compromise where the performance gains are there most of the time but we still allow for late MeterFilter configuration while warning about it, which I'll do in a separate pull request. Timeline wise, we should definitely get the warning log in the RC to start getting user feedback on scenarios that would be affected by this as soon as possible so we can hopefully get more confident about restricting when a MeterFilter can be configured. I would also like get this performance enhancement in as well, but we'll see where we end up on trying to retain the MeterFilter behavior more. I suspect MeterFilters shouldn't be configured so late or often even if they're after some meters have been registered.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -772,14 +790,20 @@ public Config commonTags(String... tags) {
...
+         * registry.</strong>
          * @param filter The filter to add to the registry.
          * @return This configuration instance.
          */
         public synchronized Config meterFilter(MeterFilter filter) {
-            MeterFilter[] newFilters = new MeterFilter[filters.length + 1];
-            System.arraycopy(filters, 0, newFilters, 0, filters.length);
-            newFilters[filters.length] = filter;
-            filters = newFilters;
+            if (meterMap.isEmpty()) {
```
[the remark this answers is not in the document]
shakuzen: Here's the promised PR: #4917

shakuzen:
[Here's](https://github.com/micrometer-metrics/micrometer/compare/main...shakuzen:micrometer:track-stale-meters) an implementation of an idea to maintain existing behavior while allowing this optimization in cases when a MeterFilter hasn't been configured since the meter was last registered. See the two tests added checking the existing behavior and the new behavior. I'll look more into concurrency and benchmarking tomorrow.

lenin-jaganathan:
I was away for a while and hence didn't get back on this one quickly. I will have a look at the implementations today. Meanwhile, one additional thing we might want to consider might be How solving this would affect a composite meter registry and its initial design choices. Left a comment [here](https://github.com/micrometer-metrics/micrometer/pull/4917#issuecomment-2036252946).

shakuzen:
I think we're ready to add the commits over [here](https://github.com/micrometer-metrics/micrometer/compare/main...shakuzen:micrometer:track-stale-meters) to this pull request. I've added concurrency and benchmarks. I'll share more info when I'm back at the computer later, but feel free to leave any feedback in the meantime. I'm hopeful we can get this into the 1.13.0 RC scheduled for Monday.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -780,6 +815,7 @@ public Config commonTags(String... tags) {
         public synchronized Config meterFilter(MeterFilter filter) {
             if (!meterMap.isEmpty()) {
                 logWarningAboutLateFilter();
+                stalePreFilterIds.addAll(preFilterIdToMeterMap.keySet());
```
lenin-jaganathan, line 818: Should we synchronize this on `meterMapLock`? Given that meterfilter updation is very very rare, this could potentially avoid any meter creation-related activities when filters are getting updated WDYT.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -710,12 +739,18 @@ public Meter removeByPreFilterId(Meter.Id preFilterId) {
...
+        if (meterMap.containsKey(mappedId)) {
             synchronized (meterMapLock) {
-                m = meterMap.remove(mappedId);
-                if (m != null) {
+                final Meter removedMeter = meterMap.remove(mappedId);
+                Iterator<Meter> iterator = preFilterIdToMeterMap.values().iterator();
+                while (iterator.hasNext()) {
+                    Meter next = iterator.next();
+                    if (next.equals(removedMeter)) {
+                        stalePreFilterIds.remove(next.getId());
```
lenin-jaganathan, line 749: Shouldn't we remove this on the key of the value being removed from preFilterId?

The below test would fail,

``` java
@Test
void meterRegistryShouldBeWipedCompletely() {
    registry.config().commonTags("application", "abcservice");
    Counter c1 = registry.counter("counter");
    registry.config().commonTags("common", "tag");
    
    registry.remove(c1.getId());
    assertThat(registry.getMeters()).isEmpty();
    assertThat(registry.preFilterIdToMeterMap).isEmpty();
    assertThat(registry.stalePreFilterIds).isEmpty();
}
```

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -695,7 +723,8 @@ public Meter remove(Meter meter) {
     @Incubating(since = "1.3.16")
     @Nullable
     public Meter removeByPreFilterId(Meter.Id preFilterId) {
-        return remove(getMappedId(preFilterId));
+        final Meter meterToRemove = preFilterIdToMeterMap.get(preFilterId);
```
lenin-jaganathan, line 726: In case of a meterFilter being added to the registry, before this change this would apply all the latest meter filters and remove only the mapped Id.

Now, we are removing this based on preFilterId, shouldn't this remove both instances, since both would have the same preFilterId? The current implementation would remove only the mappedId and leave the initially mapped Id orphan without a preFilterId reference. This might get tricky as we would have removed the old mapping  when we applied the new mappings.

lenin-jaganathan (commented):
Looking for some more potential edge cases, but looks good except for one potential failing case.

At some point would also be good to understand what this would mean in terms of memory usage. Should be fairly low since most of things seem to be references, but you never know!!

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
lenin-jaganathan: You could also argue that since the mapping changed the old mapping is no longer valid and the old mapping being orphan is expected.

But in this case, I would rather see the orphan meter getting removed rather than sitting idle. But, then there is the problem you shared with references being held and interactions on the orphan meter.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: Good catch. Thanks.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -612,8 +620,15 @@ private Id getMappedId(Id id) {
...
-        Meter m = meterMap.get(mappedId);
+            Function<Meter.Id, ? extends Meter> noopBuilder) {
+
+        Meter m = preFilterIdToMeterMap.get(originalId);
+        if (m != null && !isStaleId(originalId)) {
+            return m;
+        }
+
+        Id mappedId = getMappedId(originalId);
+        m = meterMap.get(mappedId);
```
lenin-jaganathan, line 631:
```suggestion
        m = meterMap.get(mappedId);
        
        if(m != null) {
            // If the mapping exists and the meter is marked stale, then this meter is no longer stale.
            synchronized (meterMapLock) {
                if(isStaleId(originalId)) {
                    unmarkStaleId(originalId);
                }
            }
            return m;
        }
```

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -94,12 +94,21 @@ public abstract class MeterRegistry {
...
+     */
     private final Map<Id, Meter> meterMap = new ConcurrentHashMap<>();
 
+    // writes guarded by meterMapLock, reads can be stale
+    // VisibleForTesting
+    final Map<Id, Meter> preFilterIdToMeterMap = new HashMap<>();
+
+    // not thread safe; only needed when MeterFilter configured with Meters registered
+    // VisibleForTesting
+    final Set<Id> stalePreFilterIds = new HashSet<>();
```
lenin-jaganathan, line 110: Also, should we make setters that return unmodifiable collections and make these collections private?

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: I believe the behavior is the same before and after unless I'm missing some case. I wrote this test, for example, and it passes before and after this PR.

```java
@Test
void removeByPreFilterIdAfterAddingFilter() {
    Counter c1 = registry.counter("counter");
    registry.config().commonTags("common", "tag");
    Counter c2 = registry.counter("counter");

    assertThat(registry.removeByPreFilterId(c1.getId())).isSameAs(c2).isNotSameAs(c1);
    assertThat(registry.getMeters()).containsExactly(c1);
}
```

> But, then there is the problem you shared with references being held and interactions on the orphan meter.

Because of this, I don't think we should be changing this behavior yet. I can understand how the behavior can be surprising from a perspective, but changing it has more of a chance of being surprising to someone who (perhaps unknowingly) depends on the current behavior.

shakuzen:
Thank you for the additional feedback. I think we will have to consider remaining feedback post-RC1. It's very late for me and we need to get this release out. Feel free to share any other feedback you have. I just wanted to let you know why we're going forward before handling it.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -94,12 +94,21 @@ public abstract class MeterRegistry {
...
+    /**
+     * Even though writes are guarded by meterMapLock, iterators across value space are
+     * supported. Hence, we use CHM to support that iteration without
+     * ConcurrentModificationException risk.
+     */
     private final Map<Id, Meter> meterMap = new ConcurrentHashMap<>();
 
+    // writes guarded by meterMapLock, reads can be stale
+    // VisibleForTesting
+    final Map<Id, Meter> preFilterIdToMeterMap = new HashMap<>();
```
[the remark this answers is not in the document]
lenin-jaganathan: I am not sure why that would happen. Do you know of a scenario where this could happen?

shakuzen:
Sharing benchmark results from my M1 Macbook Pro:

This PR:
```
Benchmark                                                                   Mode  Cnt      Score    Error   Units
MeterRegistrationBenchmark.registerExisting                                 avgt    5      0.021 ±  0.001   us/op
MeterRegistrationBenchmark.registerExisting:gc.alloc.rate                   avgt    5  12807.403 ± 53.471  MB/sec
MeterRegistrationBenchmark.registerExisting:gc.alloc.rate.norm              avgt    5    144.000 ±  0.001    B/op
MeterRegistrationBenchmark.registerExisting:gc.count                        avgt    5   1206.000           counts
MeterRegistrationBenchmark.registerExisting:gc.time                         avgt    5    625.000               ms
MeterRegistrationBenchmark.registerExistingWithProvider                     avgt    5      0.024 ±  0.001   us/op
MeterRegistrationBenchmark.registerExistingWithProvider:gc.alloc.rate       avgt    5  11486.191 ± 48.498  MB/sec
MeterRegistrationBenchmark.registerExistingWithProvider:gc.alloc.rate.norm  avgt    5    144.000 ±  0.001    B/op
MeterRegistrationBenchmark.registerExistingWithProvider:gc.count            avgt    5   1081.000           counts
MeterRegistrationBenchmark.registerExistingWithProvider:gc.time             avgt    5    555.000               ms
MeterRegistrationBenchmark.registerNew                                        ss  400      7.839 ±  0.688   us/op
MeterRegistrationBenchmark.registerNew:gc.alloc.rate                          ss  400     90.960 ±  5.493  MB/sec
MeterRegistrationBenchmark.registerNew:gc.alloc.rate.norm                     ss  400   4059.800 ± 61.780    B/op
MeterRegistrationBenchmark.registerStale                                      ss  400      6.026 ±  0.908   us/op
MeterRegistrationBenchmark.registerStale:gc.alloc.rate                        ss  400     67.791 ±  4.128  MB/sec
MeterRegistrationBenchmark.registerStale:gc.alloc.rate.norm                   ss  400   4023.130 ± 61.860    B/op
```

With Micrometer 1.13.0-M2:

```
Benchmark                                                                   Mode  Cnt     Score    Error   Units
MeterRegistrationBenchmark.registerExisting                                 avgt    5     0.123 ±  0.001   us/op
MeterRegistrationBenchmark.registerExisting:gc.alloc.rate                   avgt    5  4211.074 ± 47.241  MB/sec
MeterRegistrationBenchmark.registerExisting:gc.alloc.rate.norm              avgt    5   272.000 ±  0.001    B/op
MeterRegistrationBenchmark.registerExisting:gc.count                        avgt    5   704.000           counts
MeterRegistrationBenchmark.registerExisting:gc.time                         avgt    5   338.000               ms
MeterRegistrationBenchmark.registerExistingWithProvider                     avgt    5     0.101 ±  0.001   us/op
MeterRegistrationBenchmark.registerExistingWithProvider:gc.alloc.rate       avgt    5  4083.215 ± 36.611  MB/sec
MeterRegistrationBenchmark.registerExistingWithProvider:gc.alloc.rate.norm  avgt    5   216.000 ±  0.001    B/op
MeterRegistrationBenchmark.registerExistingWithProvider:gc.count            avgt    5   682.000           counts
MeterRegistrationBenchmark.registerExistingWithProvider:gc.time             avgt    5   322.000               ms
MeterRegistrationBenchmark.registerNew                                        ss  400    10.195 ±  1.950   us/op
MeterRegistrationBenchmark.registerNew:gc.alloc.rate                          ss  400    92.470 ±  6.339  MB/sec
MeterRegistrationBenchmark.registerNew:gc.alloc.rate.norm                     ss  400  4256.750 ± 61.338    B/op
MeterRegistrationBenchmark.registerStale                                      ss  400     6.338 ±  0.849   us/op
MeterRegistrationBenchmark.registerStale:gc.alloc.rate                        ss  400    90.097 ±  5.519  MB/sec
MeterRegistrationBenchmark.registerStale:gc.alloc.rate.norm                   ss  400  4017.450 ± 61.371    B/op

```

Of note, registering existing meters appears about 5x faster and generates about half as much garbage.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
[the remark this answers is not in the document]
shakuzen: Thanks for the review @tsegismont. I added jcstress tests with the intent of catching such kinds of issues. The current tests are probably lacking scenarios that may run into problems like you mentioned. Could you share something we could add to there to verify the issue and make sure it stays fixed in the future?

lenin-jaganathan:
Retrospectively thinking about what sort of memory implications this will have when we have the below scenarios,
1. A meter with dynamicTagValues like uri with userId and meter-filter imposes some limit?
E.g:
```java
new MeterFilter() {
    private final Set<String> observedTagValues = ConcurrentHashMap.newKeySet();

    @Override
        String value = id.getTag("uri");
        if (value != null) {
            if (!observedTagValues.contains(value)) {
                if (observedTagValues.size() >= 100) {
                    maskedTag = Tag.of("uri, "MASKED");
                    return id.withTag(maskedTag);
                }
                observedTagValues.add(value);
            }
        }
        return id;
    }
}
```
The above meterfilter adds a dummy URI if the number of URIs exceeds a limit (in this case 100). Once 100 unique URI's, are added, this will map all the new ID to a single meter. But, after this fix the mapping will be the same but we will keep on adding the prefilter ID's to the map which will grow indefinitely.


If filters return a NOOP meter we are fine since we don't store them anyway. But, in the above scenario we are altering the URI tag alone to make sure we don't lose the count of requests but still managing tags to a reasonable count.

shakuzen:
@lenin-jaganathan that does sound like a problem for unbounded cardinality tags if users are doing something like you showed. I'm not really sure what we can do about it. Would you open a separate issue to track it? Ideas welcome on solutions, too.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: I assume you meant getters. That's a good idea. I've done it in https://github.com/micrometer-metrics/micrometer/commit/f1471c573d2067ca508387f67815b6b5858d53fc

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: You're right. This issue would be unfortunate and somewhat common to run into, I suspect. I've committed a fix in c63b93a5ab9678e7ffe7053c08a6b9f843364542

izeye:
> Of note, registering existing meters appears almost 6x faster and generates about half as much garbage.

@shakuzen It seems to be more than double, not half. Am I missing something?

izeye:
> > Of note, registering existing meters appears almost 6x faster and generates about half as much garbage.
> 
> @shakuzen It seems to be more than double, not half. Am I missing something?

@shakuzen Oops, sorry. I missed the increased throughput from its calculation and should have seen the "norm" value instead. Please ignore this comment.

Issue #4089, opened 2023-09-18
Upgrade to Apache HttpComponents Client 5.3.0


cachescrubber:
**Apache HttpComponents follow-up.**

This is a follow-up to the changes made to the HttpComponents instrumentation in #3800 and #3941, specially to the latter. To solve the issues left I filed [HTTPCLIENT-2291](https://issues.apache.org/jira/browse/HTTPCLIENT-2291) and asked for clarification of the documentation in regard to the exec chain behavior in case of retries. The javadoc for `(Async)ExecChainHandler` has been [updated in this pr](https://github.com/apache/httpcomponents-client/pull/480). Both changes are scheduled with the release **5.3.0** of `httpclient5`.

With the changes in **5.3.0** the `ObservationExecChainHandler` could be safely positioned **before** or **after** the retry exec chain element. Async or classic exec chain behavior would be identical. In micrometer, only the javadoc of `ObservationExecChainHandler` had to be adjusted accordingly, the current implementation would be just fine.

As an opinionated version of the instrumentation I would recommend to document the following:

***meter request as a whole, without individual retries***

`addExecInterceptorFirst("micrometer", new ObservationExecChainHandler(observationRegistry));`

***meter retries individually***

`addExecInterceptorAfter(ChainElement.RETRY.name(), "micrometer", new ObservationExecChainHandler(observationRegistry));`

I don't know exactly when httpclient5  5.3.0 will be released, but IMO it will be rather short term. So with a bit of luck it might be picked up by micrometer 1.12.0.

@bclozel FYI

marcingrzejszczak:
Hey @cachescrubber, are you willing to file a PR?

cachescrubber:
Hi @marcingrzejszczak I can file a PR.

Ideally I could create a Section on Apache HttpComponents instrumentation in the [Reference Instrumentations](https://docs.micrometer.io/micrometer/reference/reference.html) section of the documentation. Would that be feasible?

Target version would be 1.13, right?

marcingrzejszczak:
Correct on both questions. Target 1.13, and reference instrumentations section :)

cachescrubber:
Ok, fine. One more question - I there a timeline for removal of the deprecated http-client 4 supports and the initial Interceptor / requestExecutor? The deprecations are not marked forRemoval and no further information is given in the javadoc.

marcingrzejszczak:
[...]

PR #4860, opened 2024-03-18
Document Retry Strategy Considerations for Apache HttpComponents

cachescrubber:
# Document Retry Strategy Considerations for Apache HttpComponents

Document how to configure the micrometer exec handler for specific retry strategy behaviour.

See #4089


shakuzen (approved):
Excellent addition to the documentation and tests. Thank you for this. I'll rebase on the `1.12.x` branch so we can get it in those versions of the documentation going forward since this doesn't rely on any new feature introduced after 1.12.x.

cachescrubber:
Hi @shakuzen, we should not merge it to **1.12.x**, since the free placement of the exec chain handler over both classic and async requires httpcomponents **5.3.x.**, hence it is related #4089 .

shakuzen:
@cachescrubber Thanks. I missed that. The need for 5.3 or later should be called out in the documentation, because regardless of what version we compile against, it does not mean users are necessarily using that version.

docs/modules/ROOT/pages/reference/httpcomponents.adoc
```diff
@@ -1,12 +1,14 @@
 [[overview]]
 = Apache HttpComponents Client Instrumentation
 
-IMPORTANT: This section requires usage of Apache HttpComponents Client at least in version 5.
+IMPORTANT: This section requires usage of Apache HttpComponents Client at least in version *5.3*.
```
shakuzen, line 4: It's just the retry handling that requires 5.3, right? This makes it sound like 5.3 is required for the instrumentation to work at all.

PR #4867, opened 2024-03-21
Add Exemplars support to Prometheus 1.x

jonatan-ivanov:
Issues:

1. prometheus-metrics-tracer-initializer is needed:
We would only need SpanContext from prometheus-metrics-tracer-common
but ExemplarSampler imports SpanContextSupplier so that is also needed
from prometheus-metrics-tracer-initializer.

2. Testing is not easy because of the rate limiter since the Prometheus
Java Client does not have a clock abstraction that we could mock/fake.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/DefaultExemplarSamplerFactory.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/DefaultExemplarSamplerFactory.java
@@ -16,6 +16,7 @@
 package io.micrometer.prometheusmetrics;
 
 import io.prometheus.metrics.config.ExemplarsProperties;
+import io.prometheus.metrics.config.PrometheusProperties;
 import io.prometheus.metrics.core.exemplars.ExemplarSampler;
 import io.prometheus.metrics.core.exemplars.ExemplarSamplerConfig;
 import io.prometheus.metrics.tracer.common.SpanContext;
@@ -29,9 +30,9 @@ import java.util.concurrent.ConcurrentMap;
  * @author Jonatan Ivanov
  * @since 1.13.0
  */
-public class DefaultExemplarSamplerFactory implements ExemplarSamplerFactory {
+class DefaultExemplarSamplerFactory implements ExemplarSamplerFactory {
 
-    private final ExemplarsProperties properties = ExemplarsProperties.builder().build();
+    private final ExemplarsProperties exemplarProperties = PrometheusProperties.get().getExemplarProperties();
 
     private final ConcurrentMap<Integer, ExemplarSamplerConfig> exemplarSamplerConfigsByNumberOfExemplars = new ConcurrentHashMap<>();
 
@@ -46,14 +47,15 @@ public class DefaultExemplarSamplerFactory implements ExemplarSamplerFactory {
     @Override
     public ExemplarSampler createExemplarSampler(int numberOfExemplars) {
         ExemplarSamplerConfig config = exemplarSamplerConfigsByNumberOfExemplars.computeIfAbsent(numberOfExemplars,
-                key -> new ExemplarSamplerConfig(properties, numberOfExemplars));
+                key -> new ExemplarSamplerConfig(exemplarProperties, numberOfExemplars));
         return new ExemplarSampler(config, spanContext);
     }
 
     @Override
     public ExemplarSampler createExemplarSampler(double[] histogramClassicUpperBounds) {
         ExemplarSamplerConfig config = exemplarSamplerConfigsByHistogramUpperBounds.computeIfAbsent(
-                histogramClassicUpperBounds, key -> new ExemplarSamplerConfig(properties, histogramClassicUpperBounds));
+                histogramClassicUpperBounds,
+                key -> new ExemplarSamplerConfig(exemplarProperties, histogramClassicUpperBounds));
         return new ExemplarSampler(config, spanContext);
     }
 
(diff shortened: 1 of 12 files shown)


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/DefaultExemplarSamplerFactory.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Default implementation of {@link ExemplarSamplerFactory}.
+ *
+ * @author Jonatan Ivanov
+ * @since 1.13.0
+ */
+public class DefaultExemplarSamplerFactory implements ExemplarSamplerFactory {
+
+    // TODO: Should we expose ExemplarsProperties and let users inject it through a ctor?
+    private final ExemplarsProperties properties = ExemplarsProperties.builder().build();
```
shakuzen, line 35: Since we aren't implementing the behavior controlled by this, I think it's okay for users to rely on the configuration methods documented by the Prometheus client: https://prometheus.github.io/client_java/config/config/
As long as our usage allows these methods to work, then I don't think we need to expose something specifically.

jonatan-ivanov: We don't allow changing these defaults and no one asked for this for the old client either.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -85,17 +93,17 @@ public PrometheusMeterRegistry(PrometheusConfig config, PrometheusRegistry regis
      * @param config configuration
      * @param registry prometheus registry
      * @param clock clock
-     * @param exemplarSampler exemplar sampler
+     * @param exemplarSamplerFactory exemplar sampler factory
      */
     public PrometheusMeterRegistry(PrometheusConfig config, PrometheusRegistry registry, Clock clock,
-            @Nullable ExemplarSampler exemplarSampler) {
+            @Nullable ExemplarSamplerFactory exemplarSamplerFactory) {
```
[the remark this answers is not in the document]
jonatan-ivanov: The advantage of providing an `ExemplarSamplerFactory` is that users can control sampling behavior, i.e.: tweak sampling config or completely redefine its logic.
We provided this possibility with the 0.x client (since we let them inject `ExemplarSampler`) and I remember users overriding it to remove exemplars on `_count`. 

Maybe it can also help testing since right now you need to wait 90ms between recordings, otherwise no Exemplar will be sampler, with a custom sampler or tweaking with sampling config, this can be avoided.

I'm ok replacing it to `SpanContext` since it is simpler and if users will ask for it we can add an extra ctor.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/DefaultExemplarSamplerFactory.java
shakuzen: I looked into the implementation a bit. I assumed that using the builder would load the values from the locations specified in the documentation and use those unless programmatically overwritten. That's not the case. To get that sort of behavior, we should be using `PrometheusPropertiesLoader`, I think.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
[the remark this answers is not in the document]
shakuzen: I think if we make sure the Prometheus client can be configured via its standard way (see other comment thread), users will be able to control exemplar sampling interval and exemplars on count etc without us exposing a separate interface or accepting a properties object, or am I still missing something?

If I'm not missing anything, I think we can make these new classes not public to limit new API we need to maintain. It'd be great to get feedback from users on if the API we're exposing makes sense for their needs or not. Unfortunately we won't have much time since this won't make it in until the RC.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/DefaultExemplarSamplerFactory.java
jonatan-ivanov: `PrometheusProperties` keeps a singleton instance created by `PrometheusPropertiesLoader` and it's `get` method returns this instance so using that should be as simple as:
```java
private static final ExemplarsProperties EXEMPLARS_PROPERTIES = PrometheusProperties.get().getExemplarProperties();
```
and
```java
private static final ExporterProperties EXPORTER_PROPERTIES = PrometheusProperties.get().getExporterProperties();
```

But to me the question is more like: should we really do this and depend on PrometheusProperties? Or instead we should expose these properties (if we want to) in `PrometheusConfig`? In the case of `ExemplarsProperties` this might make some sense since we are using the defaults and this let the users override them (though it would be a valid question why they cannot override it through `PrometheusConfig`).
But in the case of `ExporterProperties` we are not using the default (`exemplarsOnAllMetricTypes` is false by default) so either we use the defaults and drop the extra exemplars by default or figure out how to let the users override this by using `ExporterProperties`.

jonatan-ivanov: I did this here: https://github.com/micrometer-metrics/micrometer/pull/4867/commits/228d8598a47caa19bbc11a150aa8253424f284fb
With this, you can override `ExemplarsProperties` but not `ExporterProperties`.

PR #4917, opened 2024-04-02
Warn when a MeterFilter is configured after a meter is registered

shakuzen:
In normal usage we would expect this should not happen, but there is nothing enforcing or encouraging it to not happen. This warns users via logging that it is happening and advises to correct it, if possible. Otherwise, we want to learn about why it is not fixable. The problem with configuring MeterFilters after a Meter has been registered is that such filters will not be applied to previously registered meters. This can result in a mix of metrics with all filters applied and only some filters applied, which is potentially hard to notice.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -16,6 +16,8 @@
 package io.micrometer.core.instrument;
 
 import io.micrometer.common.lang.Nullable;
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
 import io.micrometer.core.annotation.Incubating;
 import io.micrometer.core.instrument.Meter.Id;
 import io.micrometer.core.instrument.config.MeterFilter;
@@ -776,11 +778,29 @@ public abstract class MeterRegistry {
          * @return This configuration instance.
          */
         public synchronized Config meterFilter(MeterFilter filter) {
+            if (!meterMap.isEmpty()) {
+                logWarningAboutLateFilter();
+            }
             MeterFilter[] newFilters = new MeterFilter[filters.length + 1];
             System.arraycopy(filters, 0, newFilters, 0, filters.length);
             newFilters[filters.length] = filter;
             filters = newFilters;
             return this;
+        }
+
+        private void logWarningAboutLateFilter() {
+            InternalLogger logger = InternalLoggerFactory.getInstance(MeterRegistry.this.getClass());
+            String baseMessage = "A MeterFilter is being configured after a Meter has been registered to this registry. All MeterFilters should be configured before any Meters are registered. If that is not possible or you have a use case where it should be allowed, let the Micrometer maintainers know at https://github.com/micrometer-metrics/micrometer/issues/4920.";
+            if (logger.isDebugEnabled()) {
+                String stackTrace = Arrays.stream(Thread.currentThread().getStackTrace())
+                    .map(StackTraceElement::toString)
+                    .collect(Collectors.joining("\n\tat "));
+                logger.debug(baseMessage + "\n" + stackTrace);
+            }
+            else {
+                logger.warn(baseMessage
+                        + " Enable DEBUG level logging on this logger to see a stack trace of the call configuring this MeterFilter.");
+            }
         }
 
         /**
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/MeterRegistryLoggingTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/MeterRegistryLoggingTest.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2024 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import com.innoq.junit.jupiter.logging.Logging;
+import com.innoq.junit.jupiter.logging.LoggingEvents;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.assertj.core.api.InstanceOfAssertFactories;
+import org.junit.jupiter.api.Test;
+import org.slf4j.LoggerFactory;
+
+import static org.assertj.core.api.Assertions.as;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for expected logging from {@link MeterRegistry}.
+ */
+@Logging
+class MeterRegistryLoggingTest {
+
+    MeterRegistry registry = new SimpleMeterRegistry();
+
+    @Test
+    void meterRegistrationBeforeMeterFilterConfig(LoggingEvents logEvents) {
+        registerMetricsAndConfigure();
+
+        assertThat(logEvents.withLevel(Level.DEBUG)).isEmpty();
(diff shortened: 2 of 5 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -776,6 +778,19 @@ public Config commonTags(String... tags) {
          * @return This configuration instance.
          */
         public synchronized Config meterFilter(MeterFilter filter) {
+            if (!meterMap.isEmpty()) {
+                InternalLogger logger = InternalLoggerFactory.getInstance(MeterRegistry.this.getClass());
+                String baseMessage = "A MeterFilter is being configured after a Meter has been registered to this registry. This is unexpected and should be corrected so all MeterFilters are configured before any Meters are registered. If that is not possible or you have a use case where it should be allowed, contact the Micrometer maintainers.";
```
shakuzen, line 783: "contact the Micrometer maintainers" is a bit ambiguous what users should do. This was somewhat intentional because whatever logging message we go with will live forever in the versions of Micrometer we release with this, while the way to contact us may change over time. Maybe opening a GitHub issue and providing a link to that will be future-proof enough.

shakuzen, line 783: "This is unexpected" begs the question why, which the log message doesn't explain now, but it is already a long log message, so I was avoiding expanding more. If we do end up including a link to a GitHub issue, maybe that will alleviate this problem as we can expand more easily (and update the explanation) in a GitHub issue than we can in a log message.

jonatan-ivanov: Should we say
```suggestion
                String baseMessage = "A MeterFilter is being configured after a Meter has been registered to this registry. This is unexpected and should be corrected so all MeterFilters are configured before any Meters are registered. If that is not possible or you have a use case where it should be allowed, contact the Micrometer maintainers (see: https://micrometer.io).";
```

On the Micrometer landing page, maybe we could have a contact section that groups the slack, github and twitter links together in a drop-down menu item.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -776,6 +778,22 @@ public Config commonTags(String... tags) {
...
+                if (logger.isDebugEnabled()) {
+                    String stackTrace = Arrays.stream(Thread.currentThread().getStackTrace())
+                        // skip getStackTrace frame and the call within this method
+                        .skip(2)
```
jonatan-ivanov, line 787: How brittle is this? Can this bite us later?

micrometer-core/src/test/java/io/micrometer/core/instrument/MeterRegistryLoggingTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.slf4j.LoggerFactory;
+
+import static org.assertj.core.api.Assertions.as;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for expected logging from {@link MeterRegistry}.
+ */
+@Logging
+class MeterRegistryLoggingTest {
```
jonatan-ivanov, line 36: Should we/can we add an extra test that checks the whole stacktrace (with multiple elements)?

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: The JavaDoc has the following warning:

> Some virtual machines may, under some circumstances, omit one or more stack frames from the stack trace. In the extreme case, a virtual machine that has no stack trace information concerning this thread is permitted to return a zero-length array from this method.

For what it's worth, I did try with a few different JVM versions/vendors and skip(2) worked fine, but that's not a great guarantee. It does feel like the first two frames should be relatively stable since it is the getStackTrace call and the call to the containing method. If there were user code in there, it would be more risky.

That all said, it's safest to just not skip anything at the expense of the stack trace being more verbose with a couple frames we don't need. I'll update to remove the skipping.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: Linking to the website is less ambiguous but I think it still leaves people wondering which method is best. Maybe any is fine, but that can split the conversation across different places and lose context for everyone. I will open a GitHub issue and include a link to it in the log message. That way we can update the latest info and everyone can read it before reporting to us.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: Without skipping (and I changed the test code a bit), we end up with a stack trace like this (with IntelliJ collapsing a bunch of reflection/junit/intellij frames):

```
16:01:44.236 [main] DEBUG i.m.c.i.simple.SimpleMeterRegistry - A MeterFilter is being configured after a Meter has been registered to this registry. This is unexpected and should be corrected so all MeterFilters are configured before any Meters are registered. If that is not possible or you have a use case where it should be allowed, contact the Micrometer maintainers.
java.base/java.lang.Thread.getStackTrace(Thread.java:1602)
	at io.micrometer.core.instrument.MeterRegistry$Config.meterFilter(MeterRegistry.java:785)
	at io.micrometer.core.instrument.MeterRegistry$Config.commonTags(MeterRegistry.java:759)
	at io.micrometer.core.instrument.MeterRegistry$Config.commonTags(MeterRegistry.java:771)
	at io.micrometer.core.instrument.MeterRegistryLoggingTest.configureCommonTags(MeterRegistryLoggingTest.java:79)
	at io.micrometer.core.instrument.MeterRegistryLoggingTest.registerMetricsAndConfigure(MeterRegistryLoggingTest.java:75)
	at io.micrometer.core.instrument.MeterRegistryLoggingTest.meterRegistrationBeforeMeterFilterConfigWithDebugLogging(MeterRegistryLoggingTest.java:57) <31 internal lines>
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541) <9 internal lines>
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541) <28 internal lines>
```

The first two frames would have been skipped before.

micrometer-core/src/test/java/io/micrometer/core/instrument/MeterRegistryLoggingTest.java
shakuzen: I updated the test to check for the 3 lines of the stack trace within the test code (what would be equivalent to user code in a real scenario). Do you think that's sufficient?

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: I've opened https://github.com/micrometer-metrics/micrometer/issues/4920 for tracking this.

lenin-jaganathan:
FFT: 
1. Should we deal with the composite meter registries differently?
2. How would this likely affect the parent and child registries and the addition of meter filters in composite registries.

shakuzen:
Thanks for the review. I didn't particularly think about composite registries. What kind of scenarios are you thinking might require special attention?

micrometer-core/src/test/java/io/micrometer/core/instrument/MeterRegistryLoggingTest.java
jonatan-ivanov: Yes, looks great!

jonatan-ivanov (approved):

lenin-jaganathan:
> Thanks for the review. I didn't particularly think about composite registries. What kind of scenarios are you thinking might require special attention?

One particular case I am aware of is with `Metrics.globalRegistry`. We use spring-boot actuator for wiring up the registries and then add them to the global one. There are third-party libraries that add metrics directly to the global registries and sometimes this happens before we add any meter filters.

Why we add filters to global registry?
This is to avoid overusing memory when third party libraries / users add meters directly to the global registry.

shakuzen:
Thanks for the feedback. That's good to be aware of.

> There are third-party libraries that add metrics directly to the global registries and sometimes this happens before we add any meter filters.

If whatever is configuring everything (e.g. Spring Boot auto-config) can be made to not do what will result in registering meters until after filters have been configured, this should be solvable. It's just made harder by there being a "hidden" dependency on the global MeterRegistry that e.g. Spring Boot doesn't and can't automatically know about.

It's not clear what we can/should change about this pull request for this yet, so let's get this into RC1 and collect feedback. Let's try to figure out as many of these scenarios as we can in the upcoming month before GA based on testing with the RC1 version. Hopefully we can find a good solution for these.

PR #4921, opened 2024-04-04
Allow customizing PrometheusProperties via PrometheusConfig

jonatan-ivanov:


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusConfig.java
```diff
@@ -58,6 +60,19 @@ default Duration step() {
...
+    /**
+     * @return an instance of {@link Properties} that contains Prometheus Java Client
+     * config entries, for example
+     * {@code io.prometheus.exporter.exemplarsOnAllMetricTypes=true}.
+     * @see https://prometheus.github.io/client_java/config/config/
+     */
+    @Nullable
+    default Properties prometheusProperties() {
+        Properties properties = new Properties();
+        properties.setProperty("io.prometheus.exporter.exemplarsOnAllMetricTypes", "true");
```
jonatan-ivanov, line 72: We enable exemplars by default.
If somebody overrides this method, they can do this to reuse the Micrometer defaults but also define their own (see the tests):
```java
@Override
public Properties prometheusProperties() {
    Properties properties = new Properties();
    properties.putAll(PrometheusConfig.super.prometheusProperties());
    properties.setProperty("io.prometheus.exemplars.sampleIntervalMilliseconds", "42");
    return properties;
}
```

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusConfig.java
```diff
@@ -58,6 +60,19 @@ default Duration step() {
...
 
+    /**
+     * @return an instance of {@link Properties} that contains Prometheus Java Client
+     * config entries, for example
+     * {@code io.prometheus.exporter.exemplarsOnAllMetricTypes=true}.
+     * @see https://prometheus.github.io/client_java/config/config/
+     */
+    @Nullable
+    default Properties prometheusProperties() {
+        Properties properties = new Properties();
```
shakuzen, line 71: I wonder if we should try parsing from `get` to populate the default Properties. See, for example, what we do in `OtlpConfig#resourceAttributes`. This would allow Micrometer users to use this with e.g. a `.properties` file and implementing only `get`.

shakuzen (approved):
Looks good. I left some feedback we can consider in a follow-up if we want to pursue it.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusConfig.java
jonatan-ivanov: I think that's a good idea.

shakuzen: I opened #4922 for it

PR #4951, opened 2024-04-08
[Dynatrace] debug log in exporter when no meters are registered

pichlermarc:
Adds a debug log for when no meters are passed to `DynatraceExporterV2#export()` and returns early as any code after this is a no-op.

Currently, when no meters are passed to export, we don't send any requests nor do we do any other operations that would indicate that the exporter being called on any log-level, thus making it harder for users and/or support engineers to diagnose this trivial case.

cc @pirgeo


pichlermarc:
(keeping as draft until I've sorted out the CLA situation :slightly_smiling_face:)

shakuzen (approved):
Looks good to me. Thanks.

Issue #4954, opened 2024-04-09
Improve memory usage of StepBucketHistogram


lenin-jaganathan:
While configuring Histograms the expectation is that 16 bytes would be used per bucket (8 bytes - for bucket bound, 8 bytes - for a count of values at a bound). But it seems using StepBucketHistogram (also TimeWindowFixedBoundaryHistogram) uses 8 additional bytes in addition to the required bytes. This is because these implementations seem to hold a local copy of buckets which costs for additional memory footprint.

Expected Behaviour:
Histogram only uses (2 * 8) bytes per bucket.

shakuzen:
Following up on https://github.com/micrometer-metrics/micrometer/pull/4955#issuecomment-2395707968, @lenin-jaganathan is there anything left to do for this issue, or is it taken care of by #4955?

lenin-jaganathan:
I missed the comment. Will have a look and update.

lenin-jaganathan:
We are good. Resolved by https://github.com/micrometer-metrics/micrometer/pull/4955

PR #4955, opened 2024-04-09
Reduce memory footprint of StepBucketHistogram

lenin-jaganathan:
Partially resolves https://github.com/micrometer-metrics/micrometer/issues/4954

I think we can also apply this safely to TimeWindowFixedBoundaryHistogram.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryHistogram.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryHistogram.java
@@ -27,14 +27,32 @@ class FixedBoundaryHistogram {
 
     private final boolean isCumulativeBucketCounts;
 
+    /**
+     * Creates a FixedBoundaryHistogram which tracks the count of values for each bucket
+     * bound).
+     * @param buckets - sorted bucket boundaries
+     * @param isCumulativeBucketCounts - whether the count values should be cumulative
+     * count of lower buckets and current bucket.
+     */
     FixedBoundaryHistogram(double[] buckets, boolean isCumulativeBucketCounts) {
         this.buckets = buckets;
         this.values = new AtomicLongArray(buckets.length);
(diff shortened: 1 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryHistogram.java
```diff
@@ -73,28 +93,49 @@ else if (value > key)
         return low < buckets.length ? low : -1;
     }
 
-    Iterator<CountAtBucket> countsAtValues(Iterator<Double> values) {
+    Iterator<CountAtBucket> countsAtValues(Iterator<Double> buckets) {
```
lenin-jaganathan, line 96: Can we just use `getCountsAtBucket()` instead of this? This makes a call to `countAtBucket()` for every bucket which in turn does a binary search to find the bucket.  It makes sense if we need to get the counts for only a subset of buckets, but we don't do that anywhere (why would we even do that?). But at the end, all we need is a continuous list of buckets and their count which we can get by iterating on the values directly.

There were a few issues solved around this area hence not changing anything at this point. But, 
my claim would be the `buckets` would not change after the histogram is created so we should always rely on the buckets from the underlying implementation (in this case FixedBucketHistogram).

shakuzen: > It makes sense if we need to get the counts for only a subset of buckets, but we don't do that anywhere (why would we even do that?).

I'm not sure why we would but we did at some point (inadvertently) do that, which caused a regression when code in this area was changed - see https://github.com/micrometer-metrics/micrometer/commit/bbb5585e60cc8c7aee481b3e90a1429196e58427. Unless we find some reason to do that, I agree we don't have a use case for it now. I'd be willing to explore a refactor that gets rid of this, but not in maintenance branches, I think.

shakuzen:
@lenin-jaganathan what's remaining after this to resolve https://github.com/micrometer-metrics/micrometer/issues/4954?

Issue #4957, opened 2024-04-10
Document support for Prometheus Client 1.x


jonatan-ivanov:
Prometheus Client 1.x support was added in https://github.com/micrometer-metrics/micrometer/pull/4846, we also need to add docs.

PR #4967, opened 2024-04-17
Document support for Prometheus Client 1.x

jonatan-ivanov:
Closes gh-4957


docs/modules/ROOT/pages/implementations/prometheus.adoc
```diff
@@ -175,3 +207,52 @@ Actually, the `PrometheusMeterRegistry` has a shortcut for this, so you can do t
 ----
 registry.throwExceptionOnRegistrationFailure();
 ----
+
+== Exemplars
+
+Exemplars are metadata that you can attach to the value of your time series. They can reference reference data outside of your metrics. A common use case is storing tracing information (`traceId`, `spanId`). Exemplars are not tags/dimensions (labels in Prometheus terminology), they will not increase cardinality since they belong to the values of the time series.
```
shakuzen, line 213:
```suggestion
Exemplars are metadata that you can attach to the value of your time series. They can reference data outside of your metrics. A common use case is storing tracing information (`traceId`, `spanId`). Exemplars are not tags/dimensions (labels in Prometheus terminology), they will not increase cardinality since they belong to the values of the time series.
```

shakuzen (approved):
Looks good.
For additional (future?) enhancements I was hoping to see a section on configuration. It would also be nice to make the docs cleaner if we had a way to toggle between the new/old Prometheus client so only the relevant information is shown at a time. I don't know how difficult something like that would be to do, but I know other documentation has such things for e.g. code snippets in different languages.

jonatan-ivanov:
>I was hoping to see a section on configuration.

You mean listing the Prometheus configs that you can use?
Or just point to the Prometheus docs site?

>It would also be nice to make the docs cleaner if we had a way to toggle between the new/old Prometheus client so only the relevant information is shown at a time. I don't know how difficult something like that would be to do, but I know other documentation has such things for e.g. code snippets in different languages.

I really like that idea, there is something similar in the Boot docs in the [Structuring Your Code](https://docs.spring.io/spring-boot/3.3/reference/using/structuring-your-code.html) section ([source](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/using/structuring-your-code.adoc)) but the [`include-code` extension](https://github.com/spring-io/asciidoctor-extensions?tab=readme-ov-file#include-code) does not seem to be able to do this.

shakuzen:
> You mean listing the Prometheus configs that you can use?
> Or just point to the Prometheus docs site?

Mostly just linking to the Prometheus Java client's documentation on the configuration it supports.

Issue #4968, opened 2024-04-17
Use Mockito 4 in observation-test


violetagg:
**Describe the bug**
```
org/mockito/Mockito has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

**Environment**
 - Micrometer version: 1.13.0-SNAPSHOT
 - OS: macOS
 - Java version: jdk 8

**To Reproduce**
Here is a reproducible example
https://github.com/violetagg/GH-4968

When you run `org.example.MicrometerMockitoTest#test` you should be able to see
```
org/mockito/Mockito has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
java.lang.UnsupportedClassVersionError: org/mockito/Mockito has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:473)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at org.example.MicrometerMockitoTest.test(MicrometerMockitoTest.java:10)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #4969, opened 2024-04-17
Use Mockito 4 in observation-test

jonatan-ivanov:
Mockito 5 is not compatible with Java 8.

Closes gh-4968

### DIFF
--- a/micrometer-test/build.gradle
+++ b/micrometer-test/build.gradle
@@ -32,7 +32,7 @@ dependencies {
     api 'com.github.tomakehurst:wiremock-jre8-standalone'
     java11TestImplementation 'com.github.tomakehurst:wiremock-jre8-standalone'
 
-    api 'org.mockito:mockito-core'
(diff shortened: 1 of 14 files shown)


micrometer-test/build.gradle
```diff
@@ -32,7 +32,7 @@ dependencies {
     api 'com.github.tomakehurst:wiremock-jre8-standalone'
     java11TestImplementation 'com.github.tomakehurst:wiremock-jre8-standalone'
 
-    api 'org.mockito:mockito-core'
+    api libs.mockitoCore4
```
shakuzen, line 35: We should make sure to call out this change from Mockito 5 to 4 in the release notes.

shakuzen (approved):

micrometer-test/build.gradle
jonatan-ivanov: I added the `release-notes` label to the issue.

PR #4979, opened 2024-04-21
Validate cloudwatch2 registry using localstack

sullis:
this PR uses the testcontainers-localstack module

https://java.testcontainers.org/modules/localstack/

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -9,7 +9,7 @@ awaitility = "4.2.1"
 # legacy SDK
 aws-cloudwatch = "1.12.696"
 caffeine = "2.9.3"
-cloudwatch2 = "2.25.26"
+awsSdk2 = "2.25.26"
 colt = "1.2.0"
 dagger = "2.51.1"
 dropwizard-metrics = "4.2.25"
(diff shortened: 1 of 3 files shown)


sullis:
Ready for review

shakuzen:
Thank you for the pull request. This is something we've been wanting to do for a long time, and this looks like it should be a good approach.

gradle/libs.versions.toml
```diff
@@ -9,7 +9,7 @@ awaitility = "4.2.1"
 # legacy SDK
 aws-cloudwatch = "1.12.696"
 caffeine = "2.9.3"
-cloudwatch2 = "2.25.26"
+awsSdk2 = "2.25.26"
```
jonatan-ivanov, line 12: `cloudwatch2` might be a better name since this is not the whole AWSD SDK.

gradle/libs.versions.toml
```diff
@@ -95,7 +95,8 @@ assertj = { module = "org.assertj:assertj-core", version.ref = "assertj" }
 awaitility = { module = "org.awaitility:awaitility", version.ref = "awaitility" }
 aws-javaSdkCloudwatch = { module = "com.amazonaws:aws-java-sdk-cloudwatch", version.ref = "aws-cloudwatch" }
 caffeine = { module = "com.github.ben-manes.caffeine:caffeine", version.ref = "caffeine" }
-cloudwatch2 = { module = "software.amazon.awssdk:cloudwatch", version.ref = "cloudwatch2" }
+cloudwatch2 = { module = "software.amazon.awssdk:cloudwatch", version.ref = "awsSdk2" }
+awsSdk2NettyClient = { module = "software.amazon.awssdk:netty-nio-client", version.ref = "awsSdk2" }
```
jonatan-ivanov, line 99: Can't we use the default client shipped with the cloudwatch dependency?

implementations/micrometer-registry-cloudwatch2/build.gradle
```diff
@@ -8,4 +8,7 @@ dependencies {
 
     testImplementation project(':micrometer-test')
     testImplementation(libs.mockitoCore)
+    testImplementation(libs.testcontainers.localstack)
+    testImplementation(libs.awsSdk2NettyClient)
+    testImplementation(libs.logbackLatest)
```
jonatan-ivanov, line 13: Why do we need this?

jonatan-ivanov (changes requested):
Thank you for the PR, we definitely would like to have more integration tests!

Though I think in this form this might not be a real integration test yet or I might misunderstand what is happening here.

As the PR looks right now, I don’t get why localstack is needed. It seems verification happens against the mock (spy) so wether there was an http API call to CloudWatch (what would make this an integration test) is irrelevant from the test's point of view.

I think in order to make this an integration test and justify the usage of localstack, the test should call the AWS SDK to check if the published data is in localstack (CloudWatch) or not otherwise this could be a unit test with a different mocked/fake client without localstack (we might already have that kind of tests).

Could you please check if you can call the same cloudwatch client to get the data that was published by the registry and run verification against that data?

Issue #4983, opened 2024-04-22
AnnotationHandler can't see method from upper class


bpernetupgrade:
No, I don't think that was intended. Could you open an issue for it? I guess we're missing a test case to cover that scenario since we didn't notice it.

_Originally posted by @shakuzen in https://github.com/micrometer-metrics/micrometer/pull/4506#discussion_r1574050235_

shakuzen:
This will be available shortly in published snapshot versions (`1.11.12-SNAPSHOT`, `1.12.6-SNAPSHOT`, `1.13.0-SNAPSHOT`) from the snapshot repo. @bpernetupgrade would you mind testing out snapshots to make sure there are no remaining issues related to this for you?
Thanks @MartinUhlen for so quickly providing a fix and a test for this.

PR #4985, opened 2024-04-24
fix: AnnotationHandler should read annotations from parent class

MartinUhlen:
Fixes issue #4983 which is a regression in pull request #4506

`AnnotationHandler` should be able to read annotations from parent class.

The fix is really simple (too simple? why was it there in the first place?), just removing a redundant line in `AnnotationHandler`. 

The new test case and all existing test cases passes.


shakuzen:
> The fix is really simple (too simple? why was it there in the first place?), just removing a redundant line in `AnnotationHandler`.

I'm not sure, but it looks like it may have been a mistake. I don't see the line in the original pull request https://github.com/micrometer-metrics/micrometer/pull/3727, but I see it in the follow-up commit https://github.com/micrometer-metrics/micrometer/commit/bc59c47766ce63a0dc851cea05e5c6e45382558a. I'm not sure where it came from.

shakuzen (approved):
Thank you for the quick turnaround on fixing this. I've rebased on the `1.11.x` branch so we can merge forward and include it in all affected versions.

shakuzen:
@bpernetupgrade would you mind taking a look? Does the test capture the scenario you had in mind? Any other tests you think we should add?

bpernetupgrade:
> @bpernetupgrade would you mind taking a look? Does the test capture the scenario you had in mind? Any other tests you think we should add?

@shakuzen I am not super familiar with AspectJ, but yeah the test scenario is correct and matched the situation I encountered. Thanks.

Issue #2410, opened 2021-01-15
OkHttpMetricsEventListener always returns NOT_FOUND as uri for 404


johntowell:
This bit of code is breaking our metrics, why would we not pass 404 to the uriMapper?

```    private String getUriTag(CallState state, @Nullable Request request) {
        if (request == null) {
            return TAG_VALUE_UNKNOWN;
        }
        return state.response != null && (state.response.code() == 404 || state.response.code() == 301)
                    ? "NOT_FOUND" : urlMapper.apply(request);
    }```

checketts:
In other code 404s are grouped together to avoid a tag explosion (and corresponding memory implications).

When you say 'breaking' your metrics, is it causing some sort of crash? Or merely that your metrics are incomplete?

johntowell:
> In other code 404s are grouped together to avoid a tag explosion (and corresponding memory implications).
maybe make it a configuration? with default enabled?

> When you say 'breaking' your metrics, is it causing some sort of crash? Or merely that your metrics are incomplete?
yes incomplete, we track all 400s by endpoint in grafana to see which ones are failing. And all the 404 are grouped together and look like `GET - NOT_FOUND (404)` so we can't tell which endpoint is failing.

checketts:
> maybe make it a configuration? with default enabled?

Sounds good to me. Would you be interested in contributing that?

johntowell:
That's fair, but I don't have the time right now

shakuzen:
> maybe make it a configuration? with default enabled?

While I'm not necessarily opposed to this being possible to configure (though honestly even that gives me pause for concern), I don't think a simple on-off switch is the best way for that. It adds a configuration option for all users to consider and is too easy to flip, forget, and then have an outage in production due to it.

After looking into this a bit, I think it is already possible to configure, since we added the `contextSpecificTags` in 1.5. See [this `tag` method on the Builder](https://www.javadoc.io/static/io.micrometer/micrometer-core/1.6.3/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.Builder.html#tag(java.util.function.BiFunction)) where you can configure a function to overwrite the `uri` tag. This feels a bit hackish but that's probably how it should feel given the potential for unbounded cardinality tag explosion.

Let us know if that workaround is sufficient for your use case.

johntowell:
[...]

PR #4987, opened 2024-04-26
Add parameter to customize status codes mapped as 'NOT_FOUND' in OkHttpMetricsEventListener

tangcent:
### Problem Description
In my project, I frequently request resources via URLs formatted as `/resource/{id}`. Some resources may not exist or may have expired, resulting in a `404` response. It's crucial for me to monitor the proportion of such invalid (possibly expired) resources. Using the `urlMapper` functionality of `OkHttpMetricsEventListener`, I convert these URLs from `/resource/xxxxx` to a templated form `/resource/{id}` for consistent metric collection. But the current implementation aggregates all `404` responses under a generic `NOT_FOUND` tag, which obfuscates the visibility of specific URLs or patterns contributing to these errors.

As discussed in #2410, the `OkHttpMetricsEventListener` in Micrometer currently handles specific HTTP status codes, such as `404`, by mapping them to a `NOT_FOUND` tag to mitigate tag cardinality issues. However, this behavior is fixed and does not allow for customization based on user-specific needs or different application contexts.

### Proposed Solution
Introduce flexibility in the `OkHttpMetricsEventListener` that enables users to define custom sets of status codes that should be treated as `NOT_FOUND`. Additionally, provide the option to configure whether to create URI tags for requests resulting in these status codes, which could include or exclude `404` among others.
### Benefits

- Customization: Users gain control over how status codes are handled, allowing them to adapt metrics collection to better suit their operational and monitoring needs.

- Reduced Complexity: Provides a way to handle exceptions to general tagging rules without adding significant complexity to the monitoring setup.

- Enhanced Usability: Makes the `OkHttpMetricsEventListener` more versatile and suitable for a wider range of use cases by accommodating different monitoring strategies.

This change would make the `OkHttpMetricsEventListener` more adaptable to various monitoring needs, enhancing its overall utility and effectiveness in real-world scenarios.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
@@ -95,6 +95,8 @@ public class OkHttpMetricsEventListener extends EventListener {
 
     private final Function<Request, String> urlMapper;
 
+    private final Set<Integer> statusCodesMappedAsNotFound;
+
     private final Iterable<Tag> extraTags;
 
     private final Iterable<BiFunction<Request, Response, Tag>> contextSpecificTags;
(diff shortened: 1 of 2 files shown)


shakuzen:
Did you see my comment on the linked issue?

> After looking into this a bit, I think it is already possible to configure, since we added the contextSpecificTags in 1.5. See [this tag method on the Builder](https://www.javadoc.io/static/io.micrometer/micrometer-core/1.6.3/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.Builder.html#tag(java.util.function.BiFunction)) where you can configure a function to overwrite the uri tag.

Would that not be sufficient for your use case and customizing this behavior?

shakuzen:
@tangcent I'm still curious about my previous question, when you have a chance to respond.

tangcent:
@shakuzen yes, that makes sense! thanks for the clarification. I’ll go ahead and close this PR. lmk if there's anything else I can do. :)

shakuzen:
I've opened #5812 to track the larger change.

PR #4989, opened 2024-04-30
Remove space at beginning of line in doc

tdinev:
Remove a leading space at the beginning of a line in the `Concepts` documentation page.


jonatan-ivanov:
Thank you for the PR! Could you please rebase it against `1.12.x` so that this fix will be available in 1.12 and in 1.13 too (I will take care about forward merging it)?

PR #5055, opened 2024-05-07
MeterTags wrapper to support multiple MeterTag-s on a single method parameter

### LINKED ISSUE #4081: Improve MeterTag to allow extracting multiple keys/value tags
The '@MeterTag' feature lets you add dynamic tags, taking a value from a parameter. Currently, you can only extract one key/value pair because annotation can't be repeated and I did not see a way to define a list of key/value pairs. 
It would be helpful to allow for extracting multiple key-value pairs.

**Rationale**
For APIs with numerous fields, it would be beneficial to create multiple tags for improved filtering.

**Additional context**

smaxx:
…ethod parameter

Resolves #4081

### DIFF
--- a/micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationUtils.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationUtils.java
@@ -19,10 +19,8 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.util.Arrays;
+import java.util.LinkedList;
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
-import java.util.stream.Stream;
 
 /**
  * Utility class that can verify whether the method is annotated with the Micrometer
@@ -41,15 +39,16 @@ final class AnnotationUtils {
     static List<AnnotatedParameter> findAnnotatedParameters(Class<? extends Annotation> annotationClazz, Method method,
             Object[] args) {
         Parameter[] parameters = method.getParameters();
-        return IntStream.range(0, parameters.length).boxed().flatMap(parameterIndex -> {
-            Parameter parameter = parameters[parameterIndex];
+        List<AnnotatedParameter> result = new LinkedList<>();
+        for (int i = 0; i < parameters.length; i++) {
+            Parameter parameter = parameters[i];
             Annotation[] annotations = parameter.getAnnotationsByType(annotationClazz);
-            if (annotations.length == 0) {
-                return Stream.empty();
-            }
-            return Arrays.stream(annotations)
-                .map(annotation -> new AnnotatedParameter(parameterIndex, annotation, args[parameterIndex]));
-        }).collect(Collectors.toList());
+            final int parameterIndex = i;
+            Arrays.stream(annotations)
+                .map(annotation -> new AnnotatedParameter(parameterIndex, annotation, args[parameterIndex]))
+                .forEach(result::add);
+        }
+        return result;
     }
 
 }
--- a/micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationUtils.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationUtils.java
@@ -17,8 +17,12 @@ package io.micrometer.common.annotation;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
+import java.lang.reflect.Parameter;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
 
 /**
  * Utility class that can verify whether the method is annotated with the Micrometer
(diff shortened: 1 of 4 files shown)


marcingrzejszczak (approved):

micrometer-core/src/main/java/io/micrometer/core/aop/MeterTags.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * <p>
+ * Can be used natively, declaring several nested {@link MeterTag} annotations. Can also
+ * be used in conjunction with Java 8's support for repeatable annotations, where
+ * {@link MeterTag} can simply be declared several times on the same parameter, implicitly
+ * generating this container annotation.
+ *
+ * @author Maksym Symonov
+ * @author Marcin Grzejszczak
+ * @since 1.13.0
```
jonatan-ivanov, line 36: We usually don't add new features after our RC release.
I think this should go to `1.14.0` instead of `1.13.0` (in less than a week).

micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationUtils.java
```diff
@@ -36,18 +40,16 @@ private AnnotationUtils() {
...
             Object[] args) {
-        Annotation[][] parameters = method.getParameterAnnotations();
-        List<AnnotatedParameter> result = new ArrayList<>();
-        int i = 0;
-        for (Annotation[] parameter : parameters) {
-            for (Annotation parameter2 : parameter) {
-                if (annotationClazz.isAssignableFrom(parameter2.annotationType())) {
-                    result.add(new AnnotatedParameter(i, parameter2, args[i]));
-                }
+        Parameter[] parameters = method.getParameters();
```
jonatan-ivanov, line 43: I appreciate the refactoring effort especially since the previous version bothers me a little:
- `Annotation[][] parameters` should have been `parameterAnnotations`
- `Annotation[] parameter` should have been `annotations`
- `Annotation parameter2` should have been `annotation`
- incrementing the index separately

This refactor fixes these concerns but it also introduces a few more.
I don't think that doing this is a good usage of the Stream API (it might be also less performant), what do you think about using a for loop with an index instead of an `IntStream`?
Something like this I guess:
```java
Parameter[] parameters = method.getParameters();
for (int i = 0; i < parameters.length; i++) {
    Parameter parameter = parameters[i];
    Annotation[] annotations = parameter.getAnnotationsByType(annotationClazz);
    if (annotations.length == 0) {
        return Collections.emptyList();
    }
    else {
        return Arrays.stream(annotations)
            .map(annotation -> new AnnotatedParameter(i, annotation, args[i]))
    }
}
```

micrometer-core/src/main/java/io/micrometer/core/aop/MeterTags.java
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Container annotation that aggregates several {@link MeterTag} annotations.
+ *
+ * <p>
```
jonatan-ivanov, line 28: You can remove `<p>` I think.

micrometer-core/src/test/java/io/micrometer/core/aop/TimedAspectTest.java
```diff
@@ -525,6 +546,11 @@ void getAnnotationForTagValueExpression(
             @Timed
             void getAnnotationForArgumentToString(@MeterTag("test") Long param);
 
+            @Timed
+            void getMultipleAnnotationForTagValueExpression(
```
jonatan-ivanov, line 550: Do you think it would worth adding one more test case that uses `MeterTags` explicitly (instead of Java 8's repeatable annotations)?

jonatan-ivanov (changes requested):

micrometer-core/src/test/java/io/micrometer/core/aop/TimedAspectTest.java
smaxx: good point, added

micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationUtils.java
smaxx: well, honestly i would argue about the approach :) Stream API-based approach could look like an overhead however it is always a matter of doing performance analysis first and then doing the conclusions especially now when we have corresponding tools like JMH. And unfortunately here we don't use extra libraries which would hide the additional complexity like StreamEx(`EntryStream.of(list)` will give you a stream of tuples where list element is associated with its index, or Guava's zipWithIndex). So this approach with IntStream doesn't seem to be breaking the main "rules" if such exist with regrads to how and where to use Stream API and in addition to that i remember i saw benchmark results where it was stated that on modern JVMs performance of Stream API is really close to `for` loops.

Anyway i reworked this place a bit to the suggested approach. Please see the result. Thank you so much for your comments and feedback

jonatan-ivanov (approved):

jonatan-ivanov:
We can merge this next week after we released 1.13.0.

shakuzen:
This is failing on `main` after merging. I'm looking into it now.

shakuzen:
Looks like [this commit](https://github.com/micrometer-metrics/micrometer/commit/2069e9292e3649b98df77d5bfc7f58f475c78e07) wasn't in the builds run for this PR and that is making the difference in the tests failing. Though I think it's actually the implementation of `mergeAnnotatedParameters` in `AnnotationHandler` that should be fixed. In the failing tests, it is resulting in only one of the `MeterTag` annotations (the first one) being used.

shakuzen:
I've reverted this change in 1f208955fa941e5648cbbd71fe8c33d965c155fb on `main`. We need to get it working with the latest code on main as pointed out in my previous comment. I'd reopen this pull request if I had the option. @smaxx would you mind opening a new pull request for this and we can figure out how to get everything working?

smaxx:
> I've reverted this change in [1f20895](https://github.com/micrometer-metrics/micrometer/commit/1f208955fa941e5648cbbd71fe8c33d965c155fb) on `main`. We need to get it working with the latest code on main as pointed out in my previous comment. I'd reopen this pull request if I had the option. @smaxx would you mind opening a new pull request for this and we can figure out how to get everything working?

sure, will integrate the latest `main` and open a new PR today

PR #5057, opened 2024-05-08
Add counter of failed attempts to retrieve a connection from the pool

codesimplicity:


shakuzen (commented):
Thanks for the pull request. It would be good to have a test demonstrating the counter gets incremented when expected.

codesimplicity:
@shakuzen I actually have no idea why this GC related test fails, wouldn't know how this could be related to my change in any way. Also have no clue how to run those tests locally on my system or rerun them in CircleCI (without pushing another change). Any guidance would be appreciated.

shakuzen:
@codesimplicity sorry for the long delay in responding. I have re-run the CI manually. You can amend your commit and force push to cause CI to run again for future reference. Unfortunately there isn't a better way for non-maintainers to re-run CI builds with unrelated errors.

shakuzen (approved):

Issue #1149, opened 2019-01-11
Compile-time weaving support for aspects


jefftap:
Hi, I am attempting to make use of `TimedAspect` via AspectJ compile-time weaving using the `aspectj-maven-plugin`. When the aspect code is executed, it throws one of the two errors, depending on what is being executed:

```
java.lang.NoSuchMethodError: io.micrometer.core.aop.TimedAspect.aspectOf()Lio/micrometer/core/aop/TimedAspect
```

```
java.lang.NoSuchMethodException: io.micrometer.core.aop.TimedAspect.aspectOf()
```

This usually indicates that the AspectJ weaver has not properly processed the aspect code. I am not sure if this is a problem with my own code/configuration, or if it's an issue in Micrometer's jar. The aspect works perfectly fine using Spring AOP, but I would prefer to use AspectJ weaving with `@Timed` to allow it access to private/protected methods and the other various benefits one gets from AspectJ integration.

Relevant information:
* micrometer-core version: 1.1.1
* Spring Boot version: 2.1.1.RELEASE
* AspectJ verison: 1.9.2
* Java version: 1.8
* aspectj-maven-plugin version: 1.11

aspectj-maven-plugin configuration:
```xml
            <plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>aspectj-maven-plugin</artifactId>
                <version>1.11</version>
                <configuration>
                    <complianceLevel>1.8</complianceLevel>
                    <source>1.8</source>
                    <target>1.8</target>
                    <showWeaveInfo>true</showWeaveInfo>
                    <Xlint>ignore</Xlint>
                    <aspectLibraries>
                        <aspectLibrary>
                            <groupId>org.springframework</groupId>
                            <artifactId>spring-aspects</artifactId>
                        </aspectLibrary>
                        <aspectLibrary>
                            <groupId>io.micrometer</groupId>
                            <artifactId>micrometer-core</artifactId>
                        </aspectLibrary>
                    </aspectLibraries>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>compile</goal>
                            <goal>test-compile</goal>
                        </goals>
                    </execution>
                </executions>
                <dependencies>
                    <dependency>
                        <groupId>org.aspectj</groupId>
                        <artifactId>aspectjtools</artifactId>
                        <version>${aspectj.version}</version>
                    </dependency>
                </dependencies>
            </plugin>

```

[...]

PR #5058, opened 2024-05-09
Fix AspectJ pointcut syntax

mihalyr:
AspectJ pointcuts use the `&&` and `!` operators and the alternative word form syntax with `and` and `not` operators is only allowed for load-time weaving with `aop.xml` which cannot use the symbols in the XML. This fixes compile-time weaving support which failed on the `and not` syntax.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
@@ -167,7 +167,7 @@ public class CountedAspect {
         this.shouldSkip = shouldSkip;
     }
 
-    @Around("@within(io.micrometer.core.annotation.Counted) and not @annotation(io.micrometer.core.annotation.Counted)")
+    @Around("@within(io.micrometer.core.annotation.Counted) && !@annotation(io.micrometer.core.annotation.Counted)")
     @Nullable
     public Object countedClass(ProceedingJoinPoint pjp) throws Throwable {
         if (shouldSkip.test(pjp)) {
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -161,7 +161,7 @@ public class TimedAspect {
         this.shouldSkip = shouldSkip;
     }
 
-    @Around("@within(io.micrometer.core.annotation.Timed) and not @annotation(io.micrometer.core.annotation.Timed)")
+    @Around("@within(io.micrometer.core.annotation.Timed) && !@annotation(io.micrometer.core.annotation.Timed)")
     @Nullable
     public Object timedClass(ProceedingJoinPoint pjp) throws Throwable {
         if (shouldSkip.test(pjp)) {
--- a/micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
@@ -105,7 +105,7 @@ public class ObservedAspect {
         this.shouldSkip = shouldSkip;
     }
 
-    @Around("@within(io.micrometer.observation.annotation.Observed) and not @annotation(io.micrometer.observation.annotation.Observed)")
+    @Around("@within(io.micrometer.observation.annotation.Observed) && !@annotation(io.micrometer.observation.annotation.Observed)")
     @Nullable
     public Object observeClass(ProceedingJoinPoint pjp) throws Throwable {
         if (shouldSkip.test(pjp)) {


mihalyr:
Related discussion in https://github.com/micrometer-metrics/micrometer/issues/1149#issuecomment-2102207826

jonatan-ivanov:
Thank you for fixing this! Do you think we should open an issue for this in AspectJ?
I mean if AspectJ could print out a warning, less people would fall into this trap.

mihalyr:
Hi @jonatan-ivanov there isn't really a problem with AspectJ, more like Micrometer is missing tests for this functionality.

There are two ways basically how you can use the aspects with AspectJ:

**Compile-time weaving**

If your library was compiled (or post-compiled) with the AspectJ compiler (`ajc`) like the discussion we have in #1149, then the aspects could be woven into classes at compile time by others. If your project would be setup this way, you would get compile error and your build would fail with a message like this:

```
io/micrometer/core/aop/TimedAspect.java [error] Invalid pointcut '@within(io.micrometer.core.annotation.Timed) and not @annotation(io.micrometer.core.annotation.Timed)': org.aspectj.weaver.patterns.ParserException: unexpected pointcut element: and@45:47 at position 45
```

However, at the moment you don't have an AspectJ compilation step in your build process (see #1149) so you don't get any error at build time.

**Load-time weaving**

People can also use your aspects with AspectJ at load time. In this case the error happens at runtime when the class is loaded and presented to the AspectJ weaver at execution time. You should probably have tests for this to catch these errors. Otherwise your aspects would throw exceptions for users at runtime like this:

```
[AppClassLoader@4e0e2f2a] error at io/micrometer/core/aop/TimedAspect.java::0 Invalid pointcut '@within(io.micrometer.core.annotation.Timed) and not @annotation(io.micrometer.core.annotation.Timed)': org.aspectj.weaver.patterns.ParserException: unexpected pointcut element: and@45:47 at position 45
```

And some tests should fail as the aspect is not run. Now that I tried to repro it not sure how well these class annotations work, I got some runtime failures when I tried to use them. The method annotations should be OK, but some tests should be perhaps included to make sure users get correct metrics.

mihalyr:
And I forgot the third approach that you use in your tests actually, using Spring AspectJ proxy at runtime, which perhaps is able to deal with the alternative expression syntax. But would be good to add some AspectJ LTW tests to make sure things work as intended also if you don't use Spring proxies.

I've seen some aop.xml added in https://github.com/micrometer-metrics/micrometer/pull/5059/files probably @marcingrzejszczak was adding also some LTW tests to the project there.

mihalyr:
@marcingrzejszczak @jonatan-ivanov it seems that this change revealed some other issues with these aspects. 

Before this PR my local tests with load-time weaving succeeded for `@Counted` and `@Timed` on methods, but didn't work on class-level due to the pointcut syntax error so it was just ignored. But now that I fixed the pointcut the aspect is failing because it matches things like static constructors and fails the program at runtime. 

```
class org.aspectj.runtime.reflect.ConstructorSignatureImpl cannot be cast to class org.aspectj.lang.reflect.MethodSignature (org.aspectj.runtime.reflect.ConstructorSignatureImpl and org.aspectj.lang.reflect.MethodSignature are in unnamed module of loader 'app')
java.lang.ClassCastException: class org.aspectj.runtime.reflect.ConstructorSignatureImpl cannot be cast to class org.aspectj.lang.reflect.MethodSignature (org.aspectj.runtime.reflect.ConstructorSignatureImpl and org.aspectj.lang.reflect.MethodSignature are in unnamed module of loader 'app')
	at io.micrometer.core.aop.TimedAspect.timedClass(TimedAspect.java:171)
	at io.micrometer.MeasuredClass.<init>(MeasuredClass.java:7)
	at io.micrometer.MeasuredClassTest.setUp(MeasuredClassTest.java:25)
```

mihalyr:
@marcingrzejszczak @jonatan-ivanov so there is a more serious bug that was just revealed by this PR. Before this PR the class-level annotation pointcut was just ignored due to the syntax issue with AspectJ LTW. After this PR however, the pointcut started to work and matches also constructors that cause runtime failures of the user programs when calling a constructor on such annotated classes, because the Timed/Counted aspect implementations assume a MethodSignature but they get a ConstructorSignature instance. 

I have a fix in this draft PR including AspectJ LTW tests that will help you find these issues next time. https://github.com/micrometer-metrics/micrometer/pull/5060

However, I don't really have time to help you clean this up and adjust for your project, especially the tests are very problematic there, because they need AspectJ and the current version will not work with JDK11 that you also build for.

mihalyr:
I think, I'll just extract the pointcut fixes from https://github.com/micrometer-metrics/micrometer/pull/5060 and send it separately, that should work and then you can later add some tests according to your priorities and in a way that suits the project.

mihalyr:
Sent a fix in https://github.com/micrometer-metrics/micrometer/pull/5061

jonatan-ivanov:
>Hi @jonatan-ivanov there isn't really a problem with AspectJ

From my perspective (user of AspectJ), there is: if AspectJ knows (I assume) that I'm doing something that I should not, it could warn me but I would not call this a "problem" but it would be a nice gesture towards the users.

Based on the other things you wrote and since we are post-RC and our GA release is on Monday, I think we should revert your fix in gh-5058 and apply it with your other fixes after the release. Please don't get me wrong, I think we should fix these things and I really appreciate your help on this but I think we should do this in the next release so that we have more time to find issues and write tests. /cc @marcingrzejszczak @shakuzen

mihalyr:
Hi @jonatan-ivanov the AspectJ compiler/weaver really tells you what is wrong, it is showing you errors and the code doesn't work. But you don't have tests for that, because you use Spring AOP proxies and not AspectJ CTW/LTW in your tests, and the proxies seem to work differently from the actual compiler/weaver. That's why I suggested that you should add also AspectJ tests to have these things covered and then you would see the errors and get test failures too. 

Regarding the revert, yes, that's the safe decision. I didn't know what stage you are in right now and I wanted to suggest that if you can't merge the other fixes, then please revert my change from `main` as an ignored aspect does not harm that much as a failing one.

Thank you for your quick response and really appreciate taking the time to help out with this.

jonatan-ivanov:
The changes in this PR were reverted in https://github.com/mihalyr/micrometer/commit/8a39c518393f1d9935961e45c089153a2b0ab397 we are planning to apply AspectJ fixes after our release (2024-05-13).

mihalyr:
@jonatan-ivanov I have all the fixes now in https://github.com/micrometer-metrics/micrometer/pull/5064 that goes into Marcin's branch with all LTW/CTW stuff together. The binary compatibility issue is also fixed in that PR, so once we are able to get the Spring fix from 6.1.7-SNAPSHOT for the build, it should all be good for the next release hopefully.

frankjkelly:
Hi Folks - the above has me a little confused - after upgrading from Spring boot `3.2.2` to `3.3.1` I am getting the following
```
io/micrometer/core/aop/TimedAspect.java [error] Invalid pointcut '@within(io.micrometer.core.annotation.Timed) and not @annotation(io.micrometer.core.annotation.Timed)': org.aspectj.weaver.patterns.ParserException: unexpected pointcut element: and@45:47 at position 45
```

Right now is there anything I can do to work-around this?
Should I pin the `aspectj` version or `micrometer` version or do I need to wait for a new Spring release? 
Thanks in advance!

mihalyr:
@frankjkelly the changes from this PR were originally reverted from the 1.13.1 release, but a more complete implementation including this patch and proper AspectJ CTW/LTW support has been already pushed to `main` in https://github.com/micrometer-metrics/micrometer/commit/763e1f74e60f5c7cf08a369ecac4c7e71f813ee6 which is now available as `1.14.0-SNAPSHOT` and probably will be released soon if everything goes fine. 

Not sure why do you see this error only now, it was there all along if you used AspectJ LTW, it was showing the error at startup also on Spring Boot 3.2.x. But if you don't use `@Timed` annotations on class-level, you can probably safely just ignore it. The method-level `@Timed` annotations should still work, those don't have the invalid syntax in the pointcuts.

frankjkelly:
@mihalyr Thanks for getting back to me - maybe my problem is similar but not this one as I am getting a compilation at the compile step with the following. So it does appear to be AspectJ post compile weaving complaining about something.
Seems to be the same message? 

```
Note: /runner/_work/platform2-***/platform2-***/platform-***-service/src/main/java/com/cogito/platform/***/dao/ApiClientFactory.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: /runner/_work/platform2-***/platform2-***/platform-***-service/src/main/java/com/cogito/platform/***/dao/RoutingConfigDaoEntityImpl.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
5 warnings
Error: ometer/core/aop/TimedAspect.java [error] Invalid pointcut '@within(io.micrometer.core.annotation.Timed) and not @annotation(io.micrometer.core.annotation.Timed)': org.aspectj.weaver.patterns.ParserException: unexpected pointcut element: and@45:47 at position 45
	

1 error

> Task :platform-***-service:compileJava FAILED
```

mihalyr:
@frankjkelly yes, that's the same issue, Micrometer doesn't work with compile-time weaving, the fix is in the commit waiting to be released I mentioned before. With load-time weaving you would get the error during runtime and could ignore, but CTW will fail the build on the syntax error. I had the same issue when I first switched to Micrometer from Dropwizard which led to this patch.

See also #1149 which was the main issue tracking AspectJ compile-time weaving support.

mihalyr:
@frankjkelly 
I forgot to mention that this pointcut syntax issue was introduced recently with this commit https://github.com/micrometer-metrics/micrometer/commit/a7a0274c7b6617895753a697051635c67a8c422e

So you can revert to a Micrometer version before that commit until the fix from 1.14.0-SNAPSHOT is released.

mihalyr:
@frankjkelly now I realized why you saw this when the Spring upgrade. That bad commit was added in Micrometer 1.13.0 which is what came with Spring Boot 3.3.  Now, it makes sense. So, yeah, your choice for now is to stay with Micrometer 1.12.x, not sure how it works with SB 3.3, though. Alternatively, you can switch to load-time weaving to ignore the error, that's what I did as a workaround until waiting for the fix to be released.

jonatan-ivanov:
fyi: I'm backporting that change to 1.13.x: https://github.com/micrometer-metrics/micrometer/issues/5285

mihalyr:
@jonatan-ivanov please do not backport **this**, it was reverted for a reason: https://github.com/micrometer-metrics/micrometer/pull/5058#issuecomment-2103276853

My complete fix is in https://github.com/micrometer-metrics/micrometer/pull/5064 that was merged to main in https://github.com/micrometer-metrics/micrometer/commit/763e1f74e60f5c7cf08a369ecac4c7e71f813ee6

Please use the changes from https://github.com/micrometer-metrics/micrometer/commit/763e1f74e60f5c7cf08a369ecac4c7e71f813ee6 instead for the Aspect classes. And consider the LTW tests at least.

jonatan-ivanov:
I reverted the change, sorry for jumping too quickly: https://github.com/micrometer-metrics/micrometer/issues/5285#issuecomment-2215894314

PR #5059, opened 2024-05-09
Compile-time and load-time weaving support for aspects

marcingrzejszczak:
Adds support for CTW (compile time weaving) and LTW (load time weaving) support

fixes https://github.com/micrometer-metrics/micrometer/issues/1149


mihalyr:
What would be the best way to merge this with my changes from #5060 and https://github.com/mihalyr/micrometer/tree/fix-aspectj-ctw ?

There are a few things I noticed missing from this PR:
* there are no load-time weaving tests with AspectJ
* there are no class-level annotation tests
* aspects will not work properly without the fixes in #5060 

I have all this addressed in #5060 and added also CTW in https://github.com/mihalyr/micrometer/tree/fix-aspectj-ctw. I named the module slightly differently `micrometer-test-aspectj-ltw` and `micrometer-test-aspectj-ctw`, but that's just a small detail. We could either merge #5060 to `main` and I will send another PR for your branch with the missing tests for CTW. Or I can change #5060 to be merged into your branch. But I would prefer merging #5060 to main first since that way your change would be more focused and limited to one issue at a time without the pointcut problems fixed in my PR. 

But I didn't address the `@Observed` with the global ObservationRegistry, so that would need to be added for the LTW part too, which I can address in a PR also for your branch.

There is a thread-safety issue with the new `Observations` class that should be addressed too.

marcingrzejszczak:
I think we can do everything in this PR. So please create a PR to the [issues_#1149_ctw](https://github.com/micrometer-metrics/micrometer/tree/issues_%231149_ctw) branch so that I can merge your changes.

mihalyr:
Ok, no problem, just wanted to point out that my change in #5060 is a bugfix for the broken pointcut with AspectJ load-time weaving rather an enhancement as the compile-time weaving is.

marcingrzejszczak:
I understand. Regardless, let's go with everything in that PR. I'll merge it to my branch and then will refine the branch.

mihalyr:
@marcingrzejszczak here we go https://github.com/micrometer-metrics/micrometer/pull/5064

This works for me locally the only issue is that Spring aspect check that we know about - it should work with JDK17 and Spring 6.1.7-SNAPSHOT. And of course the japicmp that was mentioned in my other PR.

build.gradle
```diff
@@ -408,6 +408,11 @@ subprojects {
     description = 'Application monitoring instrumentation facade'
 
     repositories {
+        // mavenLocal()
```
marcingrzejszczak, line 411: TODO: To be removed when 5.3.35 is out

gradle/libs.versions.toml
```diff
@@ -3,7 +3,7 @@ activemq-artemis = "2.33.0"
 application-insights = "2.6.4"
 archunit = "1.3.0"
 asmForPlugins = "7.3.1"
-aspectjweaver = "1.9.22.1"
+aspectjweaver = "1.9.20.1"
```
marcingrzejszczak, line 6: 1.9.20.1 is the last release that accepts jdk 11 for building https://github.com/eclipse-aspectj/aspectj/blob/master/docs/release/JavaVersionCompatibility.adoc

gradle/libs.versions.toml
```diff
@@ -75,7 +75,7 @@ rest-assured = "5.4.0"
 signalfx = "1.0.41"
 slf4j = "1.7.36"
 spectator-atlas = "1.7.12"
-spring = "5.3.34"
+spring = "5.3.35-SNAPSHOT"
```
marcingrzejszczak, line 78: Change to 5.3.35 when released

micrometer-observation/build.gradle
```diff
@@ -1,3 +1,7 @@
+plugins {
+    id 'io.freefair.aspectj.post-compile-weaving' version '8.6'
```
mihalyr, line 2: this might be replaced with a reference declared in libs to be consistent with `micrometer-core`


```suggestion
    alias(libs.plugins.aspectj)
```

micrometer-observation/build.gradle
```diff
@@ -21,6 +25,7 @@ dependencies {
     optionalApi 'javax.servlet:javax.servlet-api'
 
     // Aspects
+    implementation "org.aspectj:aspectjrt:${libs.versions.aspectjweaver.get()}"
```
mihalyr, line 28: we can also replace this with a reference like in `micrometer-core`

```suggestion
    implementation libs.aspectjrt   
```

micrometer-observation/src/test/resources/META-INF/aop.xml
```diff
@@ -0,0 +1,34 @@
...
+         defined in. The load-time aspect weaver must be able to weave-in additional methods each
+         Spring requires each aspect to have at run-time. If you exclude the aspects, you'll get
+         an error about aspectOf() not being defined.
+      -->
+    <include within="io.micrometer.observation..*"/>
+  </weaver>
+  <aspects>
+    <!-- These are the two aspects we want to switch on for now. -->
+    <aspect name="io.micrometer.observation.aop.ObservedAspect"/>
+  </aspects>
```
mihalyr, line 33: I don't think this `aop.xml` file is being used, it should be only needed for LTW with AspectJ which is tested in the `micrometer-test-aspectj-ltw` module.

mihalyr (approved):
Thank you! Looks good, just minor comments.

micrometer-core/build.gradle
```diff
@@ -81,6 +82,8 @@ dependencies {
     }
 
     // Aspects
+    implementation libs.aspectjrt
```
shakuzen, line 85: I don't think we want to add a transitive dependency for micrometer-core users (same for micrometer-observation). When/how is this dependency used?

marcingrzejszczak: Oh good catch, @mihalyr @kriegaex can you try to shed more light on this?

kriegaex:
This is a known fact, and of course it is a dependency. I already explained long ago in https://github.com/micrometer-metrics/micrometer/issues/1149#issuecomment-1867109246. Using annotations like `@Aspect`, `@Before` or classes like `JoinPoint` are from _aspectjrt_, so of course it ius a dependency. So far, so obvious. Instead, you can remove the `aspectjweaver` dependency, because the lkatter is a signifiucantly bigger superset of AspectJ runtime and weraving classes.

Why would you want to avoid the dependency? It is like using Log4J and trying to avoid the dependency. IMO, it does make a lot of sense to have this dependency. You always had it. If you did not advertise it to users before, it was wrong to begin with and a bug.

marcingrzejszczak:
Yeah so that being optional would be the way to go. E.g. in Spring users need to explicitly say that they want aspect support by adding `spring-boot-starter-aop` dependency

marcingrzejszczak:
[This commit](https://github.com/micrometer-metrics/micrometer/pull/5059/commits/1d06cc4a4cbdd5164ad656c915a8ad190750eb66) makes the aspectj deps optional

mihalyr:
I mentioned this also previously in https://github.com/micrometer-metrics/micrometer/issues/1149#issuecomment-2100851850

> I also had to include aspectjrt as an implementation (compile-time) dependency, I think previously aspectjweaver was an optional dep (because you decided if you want to weave or not during load time, but it seems to be needed for CTW at build-time, I wonder if we could get away with compileOnly dep for this and users can include the runtime if they use weaving).

micrometer-core/build.gradle
```diff
@@ -82,9 +82,8 @@ dependencies {
     }
 
     // Aspects
-    implementation libs.aspectjrt
-    java11Implementation libs.aspectjrt
-    optionalApi 'org.aspectj:aspectjweaver'
+    optionalApi libs.aspectjrt
+    java11RuntimeOnly libs.aspectjrt
```
mihalyr, line 86: this still adds this as a transient dependency for the Java 11 build, is that OK?

shakuzen: I don't think we need the dependency at all for the java11 sources in micrometer-core. There is no aspect-related code [there](https://github.com/micrometer-metrics/micrometer/tree/d162bddb4310df2136c5371e3be63cecc515b377/micrometer-core/src/main/java11/io/micrometer/core/instrument/binder/jdk), and the code that is there is deprecated in favor of the separate module [micrometer-java11](https://github.com/micrometer-metrics/micrometer/tree/2b5181cb384811e3f86d85685d6572f13f3ca481/micrometer-java11).

marcingrzejszczak: If I do that then we get :disappointed: 

```
* What went wrong:
Execution failed for task ':micrometer-core:compileJava11Java'.
> Cannot infer AspectJ class path because no AspectJ Jar was found on class path: configuration ':micrometer-core:java11RuntimeClasspath'
```

marcingrzejszczak: Maybe we should remove the whole multi jar config from micrometer-core ?

kriegaex:
There is `TimedAspect` in Core and `ObservedAspect` in Observation. Why are you tring to avoid a dependency that clearly exists? Why are you trying to eat the cake and keep it at the same time?Chances are, there is something I do not understand, but to me this does not make much sense.

marcingrzejszczak:
I mean we do have those classes but you don't need to use them. E.g. in Spring Boot not a single aspect will be registered if the aop dependencies are not on the classpath.

marcingrzejszczak:
I've added a commit about not deploying the aspectj test modules. Hopefully now it's ok?

shakuzen:
I rebased and resolved the conflicts. I think this is ready to be merged.

> Why are you tring to avoid a dependency that clearly exists? Why are you trying to eat the cake and keep it at the same time? Chances are, there is something I do not understand, but to me this does not make much sense.

I think your points are reasonable. Historically, Micrometer has not taken such an approach except where we have no choice due to a conflict (see micrometer-jakarta9, micrometer-jetty12, etc). Almost all of the optional dependencies in micrometer-core are libraries that Micrometer maintains instrumentation for in micrometer-core. The aspects are a bit different in nature, and perhaps a different approach would have been better for them, but TimedAspect at least has been there since 1.0.0. We could consider making a separate aspect module and migrating users to there from the aspects in micrometer-core and micrometer-observation now, but that will be a long process and it should be considered separate of this work since this can still provide benefit without otherwise changing the current arrangement.

gradle/libs.versions.toml
shakuzen: If we want to avoid bumping this version in the future, we should leave a comment here so we know why. I'll add it as a polish commit.

PR #5060, opened 2024-05-09
Fix AspectJ load-time weaving and class-level annotations

mihalyr:
The pointcut syntax used in the aspects is not valid for the AspectJ
weaver/compiler. Replaced `and not` with `&& !` to fix it and also had
to add a condition to match only methods and not constructors which the
implementation assumed it gets.

One problem was that the aspects were only tested with Spring AOP
proxies, but AspectJ weaver/compiler bugs were going unnoticed.  The
added LTW test module will run with the `aspectjweaver` java agent to
weave the classes at load time, then tests if the aspects are applied
and metrics collected as expected.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
@@ -167,7 +167,7 @@ public class CountedAspect {
         this.shouldSkip = shouldSkip;
     }
 
-    @Around("@within(io.micrometer.core.annotation.Counted) and not @annotation(io.micrometer.core.annotation.Counted)")
+    @Around("@within(io.micrometer.core.annotation.Counted) && !@annotation(io.micrometer.core.annotation.Counted) && execution(* *(..))")
     @Nullable
     public Object countedClass(ProceedingJoinPoint pjp) throws Throwable {
         if (shouldSkip.test(pjp)) {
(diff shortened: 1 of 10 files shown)


mihalyr:
I sent the pointcut syntax separately to be able to merge it ASAP to fix the crashes https://github.com/micrometer-metrics/micrometer/pull/5061

mihalyr:
@marcingrzejszczak I think you have this also covered in your PR draft, feel free to cancel this one if you are adding the tests there.

jonatan-ivanov:
The build is failing because of checkstyle, running `./gradlew format` + commit-push the changes should fix it.

mihalyr:
@jonatan-ivanov @marcingrzejszczak I've updated this PR now with the reverted changes from main and the change from my other LTW PR that I merged into one here. 

This is now passing every check locally with `./gradlew -x japicmp check`

The only remaining issue is with `japicmp` because I changed a method in CountedAspect.

```
***! MODIFIED CLASS: PUBLIC io.micrometer.core.aop.CountedAspect  (not serializable)
	---! REMOVED METHOD: PUBLIC(-) java.lang.Object interceptAndRecord(org.aspectj.lang.ProceedingJoinPoint, io.micrometer.core.annotation.Counted)
		---  REMOVED EXCEPTION: java.lang.Throwable
***  MODIFIED CLASS: PUBLIC io.micrometer.core.instrument.config.InvalidConfigurationException  (default serialVersionUID changed)
	===  UNCHANGED SUPERCLASS: java.lang.IllegalStateException (<- java.lang.IllegalStateException)
```

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -167,7 +167,7 @@ public CountedAspect(MeterRegistry registry, Function<ProceedingJoinPoint, Itera
         this.shouldSkip = shouldSkip;
     }
 
-    @Around("@within(io.micrometer.core.annotation.Counted) and not @annotation(io.micrometer.core.annotation.Counted)")
+    @Around("@within(io.micrometer.core.annotation.Counted) && !@annotation(io.micrometer.core.annotation.Counted) && execution(* *(..))")
```
marcingrzejszczak, line 170: why is there `&& execution(* *(..))` at the end?

mihalyr: Because the implementation has an unchecked class-cast to MethodSignature that causes runtime crashes if the pointcut matches a constructor, which it does without this addition that restricts it only to method signatures (with a return type).

* `@within(io.micrometer.core.annotation.Counted)` - will match the methods of class annotated with `@Counted` 
* `&& !@annotation(io.micrometer.core.annotation.Counted)` - excludes methods that are already annotated separately with `@Counted` to avoid double instrumentation
* `&& execution(* *(..))` - matches method signatures only excluding constructors

What happens if you don't include the last part is that if you annotate a class with `@Counted` it will crash at runtime when you try to create a new instance of the class because the pointcut would match the constructor which gives a ConstructorSignature and fails to be cast to MethodSignature in the implementation. The tests cover this, so if you remove that last part you'll see the error when running the tests.

mihalyr:
Agree with @marcingrzejszczak that we will do this together in his PR over at https://github.com/micrometer-metrics/micrometer/pull/5059#issuecomment-2104639855

PR #5061, opened 2024-05-09
Fix AspectJ load-time weaving and class-level annotations

mihalyr:
After fixing the pointcut syntax in the previous commit it revealed that the implementation has further errors. The pointcuts matched also constructors where the code assumed it gets a MethodSignature at runtime it got a ConstructorSignature and failed the unchecked class-cast.

Extended the pointucts to only match method signatures. Tested with AspectJ load-time weaving locally and with the existing Spring AOP proxy tests.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
@@ -167,7 +167,7 @@ public class CountedAspect {
         this.shouldSkip = shouldSkip;
     }
 
-    @Around("@within(io.micrometer.core.annotation.Counted) && !@annotation(io.micrometer.core.annotation.Counted)")
+    @Around("@within(io.micrometer.core.annotation.Counted) && !@annotation(io.micrometer.core.annotation.Counted) && execution(* *(..))")
     @Nullable
     public Object countedClass(ProceedingJoinPoint pjp) throws Throwable {
         if (shouldSkip.test(pjp)) {
@@ -198,15 +198,21 @@ public class CountedAspect {
      * {@link Counted#recordFailuresOnly()} is set to {@code false}, a success is
      * recorded.
      * @param pjp Encapsulates some information about the intercepted area.
-     * @param counted The annotation.
      * @return Whatever the intercepted method returns.
      * @throws Throwable When the intercepted method throws one.
      */
-    @Around(value = "@annotation(counted)", argNames = "pjp,counted")
+    @Around("execution (@io.micrometer.core.annotation.Counted * *.*(..))")
     @Nullable
-    public Object interceptAndRecord(ProceedingJoinPoint pjp, Counted counted) throws Throwable {
+    public Object interceptAndRecord(ProceedingJoinPoint pjp) throws Throwable {
         if (shouldSkip.test(pjp)) {
             return pjp.proceed();
+        }
+
+        Method method = ((MethodSignature) pjp.getSignature()).getMethod();
+        Counted counted = method.getAnnotation(Counted.class);
+        if (counted == null) {
+            method = pjp.getTarget().getClass().getMethod(method.getName(), method.getParameterTypes());
+            counted = method.getAnnotation(Counted.class);
         }
 
         return perform(pjp, counted);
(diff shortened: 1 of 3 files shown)


mihalyr:
The load-time weaving test are not part of the PR as they were causing unrelated build failures. They should be added by someone familiar with this project structure, the code can be found in this draft: https://github.com/micrometer-metrics/micrometer/pull/5060

micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
```diff
@@ -161,7 +161,7 @@ public TimedAspect(MeterRegistry registry, Function<ProceedingJoinPoint, Iterabl
         this.shouldSkip = shouldSkip;
     }
 
-    @Around("@within(io.micrometer.core.annotation.Timed) && !@annotation(io.micrometer.core.annotation.Timed)")
+    @Around("@within(io.micrometer.core.annotation.Timed) && !@annotation(io.micrometer.core.annotation.Timed) && execution(* *(..))")
```
mihalyr, line 164: The added part matches only method signatures (with a return type) so it will not apply to constructors that caused the runtime failures.

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -198,17 +199,23 @@ public Object countedClass(ProceedingJoinPoint pjp) throws Throwable {
...
 
+        Method method = ((MethodSignature) pjp.getSignature()).getMethod();
+        Counted counted = method.getAnnotation(Counted.class);
+        if (counted == null) {
+            method = pjp.getTarget().getClass().getMethod(method.getName(), method.getParameterTypes());
+            counted = method.getAnnotation(Counted.class);
+        }
+
         return perform(pjp, counted);
     }
```
mihalyr, line 220: I aligned this method with `TimedAspect` not sure why were there two different approaches for the same, but the previous pointcut declaration wasn't working properly for me and this is also more consistent with `ObservedAspect` and `TimedAspect` too.

mihalyr:
@marcingrzejszczak @jonatan-ivanov Could you please merge this? My previous changed revealed a more serious issue that is actually causing runtime failures in client-code under certain conditions. 

There is some japi failure that I have no clue what it is and how to fix it. But the whole build and tests are working othewise.

jonatan-ivanov:
japicmp is a tool we use to have binary compatibility reporting for our builds.
It downloads an earlier minor release and it compares it with the current binary if there are breaking changes in them (minor versions should not contain source/binary incompatible changes).

The first issue in the report is the problem that broke the build, changing this:
```java
public Object interceptAndRecord(ProceedingJoinPoint pjp, Counted counted) throws Throwable {
```
to this:
```java
public Object interceptAndRecord(ProceedingJoinPoint pjp) throws Throwable {
```
breaks source and binary compatibility so we should not do it in a minor release. On the other hand, the whole aspect class with this method in it should not be used from user code and AspectJ will be able to deal with this change I assume so I think we can add this class to the ignore list (I can help ignoring this). Btw can these methods have `default` visibility or do we need to keep them `public`? That would fix the binary incompatibility issue but it might break AspectJ.

The second issue in the report is something we can look into but it does not fail the build, you can handle it as a warning.

Locally you can disable this with `./gradlew build -x japicmp`.

mihalyr:
Thank you for the details, it makes sense now. I'll take a look if default visibility works, but as you said, it's an aspect and not something a user should depend on directly...hopefully 🙂

Let me rework these PRs a little bit, first let's revert my commit from `main`, then I'll close these and create a PR that comes as a whole, with the pointcut fixes and LTW tests. Then on top of it we can add the compile-time weaving as the next step to have fully working aspects.

There is one thing I noticed, Metrics includes a global MeterRegistry, which can be used during weaving since it is done automatically by AspectJ without user involvement. However, for the `@Observed` annotation we need also a way to use a default ObservationRegistry with and add a default constructor to ObservedAspect. I think @marcingrzejszczak has something for this already in his CTW PR draft, so I could use the same approach perhaps also for the LTW. Or we can do all this together on Marcin's PR and cover LTW and CTW and the pointcut fixes too in one go of you would prefer that.

mihalyr:
Cool, let's do that!

Could you please revert my commit from main https://github.com/micrometer-metrics/micrometer/commit/5e16809f7986bb0d63540737da92f403082492c1 ? It's not safe without some other changes which should be addressed together and don't want to risk your release.

marcingrzejszczak:
Done via https://github.com/micrometer-metrics/micrometer/commit/af7093ff21c29946e95fcc31d1c89c4a6c6590c1

PR #5064, opened 2024-05-10
Fix AspectJ LTW and CTW

mihalyr:
Fix AspectJ pointcut syntax to use `&& !` instead `and not` which is
invalid for the AspectJ compiler/weaver and only works with the Spring
AOP implementation. Also add `&& execution(* *.*(..))` to match only
methods, because the implementation assumes it gets only
MethodSignatures and crashes on ConstructorSignature at runtime.

Fixed the thread-safety and mutability issues with the singleton
Observations class, so changes are propagated to aspects that are
initialized only once.

Added AspectJ load-time weaving tests to make sure that the further
issues with pointcuts and aspects are noticed at build time.

Added more class-level annotation tests and moved the module to
'micrometer-test-aspectj-ctw' to align with
'micrometer-test-aspectj-ltw'.


micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -198,17 +198,23 @@ public Object countedClass(ProceedingJoinPoint pjp) throws Throwable {
...
      * @param pjp Encapsulates some information about the intercepted area.
-     * @param counted The annotation.
      * @return Whatever the intercepted method returns.
      * @throws Throwable When the intercepted method throws one.
      */
-    @Around(value = "@annotation(counted)", argNames = "pjp,counted")
+    @Around("execution (@io.micrometer.core.annotation.Counted * *.*(..))")
     @Nullable
-    public Object interceptAndRecord(ProceedingJoinPoint pjp, Counted counted) throws Throwable {
+    public Object interceptAndRecord(ProceedingJoinPoint pjp) throws Throwable {
```
mihalyr, line 206: Alexander kindly provided a fix for this that doesn't need to change the method signature https://github.com/micrometer-metrics/micrometer/issues/1149#issuecomment-2104899044

mihalyr: Fixed :tada: no more binary incompatibility issues

cigaly:
What is the reason why this commit has been reverted? Now CTW is failing due to bad expressions in Around annotation

marcingrzejszczak:
We are working on proper CTW support in this PR https://github.com/micrometer-metrics/micrometer/pull/5059

cigaly:
OK, but those expressions were fixed, then reverted to invalid and published as 1.13.0. That is confusing to me :-(

marcingrzejszczak:
We never said that we supported LTW or CTW in the first place. So adding LTW and CTW support is a new feature which we wanted to add for 1.14.0. Those changes were made just before a GA so we didn't want to take chances

mihalyr:
@cigaly sorry that I left this here without proper explanation, this PR was meant for Marcin to be merged into his branch. Let me add a bit of clarification here. 

First, the commits in this PR (https://github.com/micrometer-metrics/micrometer/pull/5064) were not reverted, they were merged into Marcin's new PR which will add complete AspectJ load-time and compile-time weaving support: https://github.com/micrometer-metrics/micrometer/pull/5059

I think what you may be referring to as reverted was the commit from my earlier PR https://github.com/micrometer-metrics/micrometer/pull/5058 which fixed some of the pointcuts for use with AspectJ. However, this change was incomplete as it revealed some other issues in the implementation that would have caused more harm if released unfixed and due to the timing of the new release it was considered too risky, so we rather reverted to preserve the status quo and prepare a complete change for the next release (see link to Marcin's PR above). 

You can read the comments on my earlier PR about the issues with the initial pointcut fix e.g. https://github.com/micrometer-metrics/micrometer/pull/5058#issuecomment-2103276853 or https://github.com/micrometer-metrics/micrometer/pull/5058#issuecomment-2103313929 while they did fix the annotations on the method level, they led to runtime errors when the annotations were used on class level.

PR #5067, opened 2024-05-12
Virtual threads metrics

ArtyomGabeev:


shakuzen:
This looks generally good. Have you tried this out in an application yet? What is remaining for this to be ready to merge, you think? I couldn't see what the build failure was perhaps because it is too old at this point. Could you rebase on the latest `main` and update this branch so a new build runs?

It'd be good to have a test for this also. You can make the test disabled on Java versions before 21.

Copying from the issue to keep the discussion here where changes would be made:

> I think this instrumentation should be specific to java-21. Should I go and create a java-21 source set or separate gradle module?

Is there any API used that isn't available in Java 8? If it is only a runtime issue of the specific JFR events won't be available except on Java 21 and later, I don't think we need to do anything special with the source or making a separate module. It will be up to users to not bind the MeterBinder if they are not on Java 21 or later.

> I'm not sure we need to process jdk.VirtualThreadStart and jdk.VirtualThreadEnd events.

Agreed. I think we can ignore those because users can instrument run time of virtual threads via instrumenting the executor as you said or they can instrument the execution of a single virtual thread created ad hoc.

> For [VirtualThreadPinnedEvent], things to consider:
> 1. Add a tag with stackframe where pinning happened? e.g.: com.test.ThreadPinningExample.sleep(Duration)
> 1. Additionally, I would like to be able to perform log once with extended strack trace, but I'm not sure this is a responsibility of micrometer library.

Logging a stack trace can be enabled with the system property `jdk.tracePinnedThreads` - see [this documentation](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-04C03FFC-066D-4857-85B9-E5A27A875AF9). So I don't think we should/need to do anything in Micrometer for that.

As for tagging the method signature where the pinning happened, I'm not sure. I think I would err on the side of not including it initially. The cardinality of it is hard to know and application-specific. Without that tag, users will know how much pinning is happening and for how long, at least. Maybe that provides enough value. We can consider adding it after getting some feedback on the metrics without it. Thoughts?

ArtyomGabeev:
Hi @shakuzen ,

I've been using this meter binder for a couple of months already.

I think the JFR streaming API is available only from JDK 14; that's why I've asked about a separate module for jdk21. Additionally, covering it with unit tests is easier on jdk21 since we have access to virtual threads builders.

I agree that neither logging pinned stack trace nor tagging with class caused pinning is incorrect for meter binder implementation.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmVirtualThreadEventMetrics.java
```diff
@@ -0,0 +1,70 @@
...
+    public JvmVirtualThreadEventMetrics() {
+        this(emptyList());
+    }
+
+    public JvmVirtualThreadEventMetrics(Iterable<Tag> tags) {
+        this.tags = tags;
+        this.recordingStream = new RecordingStream();
+
+        recordingStream.enable("jdk.VirtualThreadPinnedEvent");
+        recordingStream.enable("jdk.VirtualThreadSubmitFailedEvent");
```
shakuzen, line 42: I'm not familiar with the right `RecordingStream` usage, but I was looking at [this blog](https://mikemybytes.com/2024/04/17/continuous-monitoring-of-pinned-threads-with-spring-boot-and-jfr/), and it made me wonder: do we not need to start the stream? And should we not set a max age so we don't accumulate events indefinitely?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmVirtualThreadEventMetrics.java
```diff
@@ -0,0 +1,70 @@
...
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.binder.MeterBinder;
+import jdk.jfr.consumer.RecordingStream;
+
+import static java.util.Collections.emptyList;
+
+public class JvmVirtualThreadEventMetrics implements MeterBinder, AutoCloseable {
```
shakuzen, line 27: Unless we expect this to be used in a try-with-resources block (I don't think we generally expect that), I think it should be `Closeable` rather than `AutoCloseable`. Typically users or an application framework will need to take care of the lifecycle of MeterBinders that need to be closed. Making it `AutoCloseable` has the effect of causing IDEs to warn about it not being closed but it is code outside of where a MeterBinder is instantiated that will usually take care of closing it.

shakuzen (commented):
Left some review on the implementation. I'll discuss with @jonatan-ivanov about making a new java 21+ module for this and get back to you.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmVirtualThreadEventMetrics.java
ArtyomGabeev: hey @shakuzen,

Do you think we need to expose maxAge duration as a constructor parameter?

shakuzen: Good question. I had the same thought, but I'm not sure due to my lack of familiarity with JFR and in particular `RecordingStream`. I don't understand what this max age means in terms of the event listeners we register. I assume the event listeners are run nearly instantly after an event is recorded. So I'm not sure the max age could affect them. Or will max age effectively filter out events with a duration longer than the max age duration?

shakuzen:
I talked to @jonatan-ivanov about where to put this code. I think a new micrometer-java21 module is probably the least bad solution unless someone has other suggestions. @ArtyomGabeev would you update the PR to move this to such a module? You can see the micrometer-java11 module for something similar that already exists.

ArtyomGabeev:
> I talked to @jonatan-ivanov about where to put this code. I think a new micrometer-java21 module is probably the least bad solution unless someone has other suggestions. @ArtyomGabeev would you update the PR to move this to such a module? You can see the micrometer-java11 module for something similar that already exists.

I've extracted micrometer-java21 module and have a couple of questions:

1. I've removed jdk11 and jdk17 CI executors since we are building and publishing under jdk21.
However, strategically we may still want to execute unit tests under jdk11/17 and conditionally add new module in gradle settings. Note: docker tests will be run only on jdk21.
2. What is the minimum supported java version? If it is 8, we are not running any builds right now on jdk8.
3. If the minimum supported java version is 11, we can do some clean up in separate PR and remove java11 specific sourcesets from micrometer-core module, clean up the docs and gradle scripts. This is unrelated to this PR, but I can take a look later.

shakuzen (changes requested):
Rather than remove any CI executors, see what is done in the `micrometer-jetty12` module:

https://github.com/micrometer-metrics/micrometer/blob/9917c67363f4507291b407edea57aebd2508b950/micrometer-jetty12/build.gradle#L3-L6

> * What is the minimum supported java version? If it is 8, we are not running any builds right now on jdk8.

It is 8. It would be nice if we still have a JDK 8 build but it became a pain to maintain that due to test dependencies. The intention is that using the `-release` option set to 8 should catch compatibility issues.

ArtyomGabeev:
> Rather than remove any CI executors, see what is done in the `micrometer-jetty12` module:
> 
> https://github.com/micrometer-metrics/micrometer/blob/9917c67363f4507291b407edea57aebd2508b950/micrometer-jetty12/build.gradle#L3-L6
> 
> > * What is the minimum supported java version? If it is 8, we are not running any builds right now on jdk8.
> 
> It is 8. It would be nice if we still have a JDK 8 build but it became a pain to maintain that due to test dependencies. The intention is that using the `-release` option set to 8 should catch compatibility issues.

Done.

.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ executors:
   machine-executor:
     working_directory: ~/micrometer
     machine:
-      image: ubuntu-2204:2024.05.1
+      image: ubuntu-2004:2024.08.1
```
shakuzen, line 28: This looks like an unrelated change. Was this needed for some reason? Note it is downgrading the ubuntu major version from 22.04 to 20.04. If it isn't needed here, let's revert it and upgrade the machine image separately.

shakuzen (changes requested):

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jfr/JfrVirtualThreadEventMetrics.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public JfrVirtualThreadEventMetrics(RecordingSettings settings, Iterable<Tag> tags) {
+        this.settings = settings;
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        if (started) {
+            return;
```
shakuzen, line 62: This would effectively prevent binding an instance of this MeterBinder to multiple registries. We have a recent related (albeit somewhat different) question (https://github.com/micrometer-metrics/micrometer/issues/5461) on what the expected behavior is that we still haven't answered yet.
/cc @jonatan-ivanov

shakuzen (commented):
I think overall this looks good. I'll try to take a look at some prior art on instrumenting metrics with JFR to make sure we aren't missing any best practices. I left a review comment more for the maintainers to consider.
@jonatan-ivanov any concerns with this PR?

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jfr/JfrVirtualThreadEventMetrics.java
```diff
@@ -0,0 +1,113 @@
...
+        final Counter submitFailedCounter = Counter.builder("jvm.virtual.thread.submit.failed")
+            .tags(tags)
+            .description("The number of virtual thread submissions that failed")
+            .register(registry);
+
+        recordingStream.onEvent(PINNED_EVENT, event -> pinnedTimer.record(event.getDuration()));
+        recordingStream.onEvent(SUBMIT_FAILED_EVENT, event -> submitFailedCounter.increment());
+    }
+
+    protected RecordingStream createRecordingStream(RecordingSettings settings) {
```
jonatan-ivanov, line 82: Can we make this `private` and make `protected` as needed (based on user feedback)?
Do you need this to be `protected`?

micrometer-java21/src/test/java/io/micrometer/java21/instrument/binder/jfr/JfrVirtualThreadEventMetricsTests.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @BeforeEach
+    void setup() {
+        registry = new SimpleMeterRegistry();
+        jfrMetrics = new JfrVirtualThreadEventMetrics(USER_TAGS);
+        jfrMetrics.bindTo(registry);
+    }
+
+    @Test
+    void registerPinnedEvent() throws Exception {
```
jonatan-ivanov, line 46: I'm going to check how easy is to add a test for counting the failed submissions.

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jfr/JfrVirtualThreadEventMetrics.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public void close() throws IOException {
+        if (started) {
+            recordingStream.close();
+        }
+    }
+
+    public record RecordingSettings(Duration maxAge, long maxSizeBytes, Duration pinnedThreshold) {
```
jonatan-ivanov, line 99: We usually call these config, i.e. `RecordingConfig`, I would make this private in the first round and open up as necessary.

jonatan-ivanov:
@ArtyomGabeev @shakuzen I fixed the comments I mage on a branch except the package and class rename, see first commit here: https://github.com/ArtyomGabeev/micrometer/compare/virtual-threads-metrics...jonatan-ivanov:micrometer:virtual-threads-metrics

I also found a way to test starting and unparking failures and receive JFR events for them. Unfortunately this is a hack with reflection. Even more unfortunate that the JDK tests are doing the same, see the comments of the test.

ArtyomGabeev:
> @ArtyomGabeev @shakuzen I fixed the comments I mage on a branch except the package and class rename, see first commit here: https://github.com/ArtyomGabeev/micrometer/compare/virtual-threads-metrics...jonatan-ivanov:micrometer:virtual-threads-metrics
> 
> 
> 
> I also found a way to test starting and unparking failures and receive JFR events for them. Unfortunately this is a hack with reflection. Even more unfortunate that the JDK tests are doing the same, see the comments of the test.

Hi, @jonatan-ivanov 
Do you want me to cherry pick your changes to this pr?

jonatan-ivanov:
Thank you!
Fyi: I added the test case for the failed events, let's see if maintenance will be a pain and if we can improve it in the future.

jonatan-ivanov (approved):

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
```diff
@@ -0,0 +1,107 @@
...
+        recordingStream.onEvent(SUBMIT_FAILED_EVENT, event -> submitFailedCounter.increment());
+    }
+
+    private RecordingStream createRecordingStream(RecordingConfig config) {
+        RecordingStream recordingStream = new RecordingStream();
+        recordingStream.enable(PINNED_EVENT).withThreshold(config.pinnedThreshold);
+        recordingStream.enable(SUBMIT_FAILED_EVENT);
+        recordingStream.setMaxAge(config.maxAge);
+        recordingStream.setMaxSize(config.maxSizeBytes);
+        recordingStream.startAsync();
```
shakuzen, line 83: It's probably not a big deal and in most cases bind will be called very soon after this class is instantiated, but it would make more sense to start the stream only when the meters are going to be bound (when `bindTo` is called). To do that, though, we would have to bring back the `started` flag and worry about concurrent calls to `bindTo` to be safest. Maybe it's not worth it to avoid a stream being created and started even though `bindTo` may be called later (or never).

If we had more instrumentation that used JFR, it would be a good idea to have some centralized infrastructure around it so we aren't making many RecordingStream instances where one could be re-used, but it's not apparently worth doing that when we only have one instrumentation that uses it, I think.

micrometer-java21/build.gradle
```diff
@@ -0,0 +1,29 @@
...
+
+tasks.withType(JavaCompile).configureEach {
+    sourceCompatibility = JavaVersion.VERSION_21
+    targetCompatibility = JavaVersion.VERSION_21
+    options.release = 21
+}
+
+test {
+    // This hack is needed since VirtualThreadMetricsTests utilizes reflection against java.lang, see its javadoc
+    jvmArgs += ['--add-opens', 'java.base/java.lang=ALL-UNNAMED']
```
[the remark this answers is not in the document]
jonatan-ivanov: I can make this "worse": we only have one test that needs this. :)
I think a separate test task is a good idea, I can add it to the PR.

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
jonatan-ivanov: I think once we add more instrumentation to JFR, we can create a component that manages starting the recording stream and the registration in a thread-safe way.

jonatan-ivanov:
@ArtyomGabeev @shakuzen I separated the reflective tests, see the last commit and let me know what you think: https://github.com/micrometer-metrics/micrometer/pull/5067/commits/11f101649ea00bfe64de25b49e261e247431b43d

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
ArtyomGabeev: I think JFR may be interesting source for observability, lets park it as is until we have another metrics derived from JFR and rethink record stream lifecycle management.

shakuzen: Thanks to JFR being an implementation detail of this instrumentation, we should hopefully be able to accomplish that in a way that causes the least disruption later.

shakuzen (approved):

PR #5110, opened 2024-05-17
Refactored data clumps with the help of LLMs (research project)

compf:
Hello maintainers,

I am conducting a master thesis project focused on enhancing code quality through automated refactoring of data clumps, assisted by Large Language Models (LLMs).

<details>
  <summary>Data clump definition</summary>
  
A data clump exists if
1. two methods (in the same or in different classes) have at least 3 common parameters and one of those methods does not override the other,  or 
2. At least three fields in a class are common with the parameters of a method (in the same or in a different class), or
3. Two different classes have at least three common fields
  
See also the following UML diagram as an example
Example data clump
</details>

I believe these refactoring can contribute to the project by reducing complexity and enhancing readability of your source code.

Pursuant to the EU AI Act, I fully disclose the use of LLMs in generating these refactorings, emphasizing that all changes have undergone human review for quality assurance. 

Even if you decide not to integrate my changes to your codebase (which is perfectly fine), I ask you to fill out a feedback survey, which will be scientifically evaluated to determine the acceptance of AI-supported refactorings. You can find the feedback survey under https://campus.lamapoll.de/Data-clump-refactoring/en

Thank you for considering my contribution. I look forward to your feedback. If you have any other questions or comments, feel free to write a comment, or email me under tschoemaker@uni-osnabrueck.de .

Best regards,
Timo Schoemaker
Department of Computer Science
University of Osnabrück

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationContextInfo.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationContextInfo.java
@@ -20,89 +20,107 @@ import io.grpc.Metadata;
 import io.grpc.MethodDescriptor.MethodType;
 import io.grpc.Status.Code;
 import io.micrometer.common.lang.Nullable;
-import io.micrometer.observation.Observation;
-import io.micrometer.observation.transport.Propagator.Setter;
-import io.micrometer.observation.transport.RequestReplySenderContext;
 
 /**
- * {@link Observation.Context} for gRPC client.
- *
- * @author Tadaya Tsuyukubo
- * @since 1.10.0
+ * Holds information relevant to gRPC client and server observation contexts.
  */
-public class GrpcClientObservationContext extends RequestReplySenderContext<Metadata, Object> {
+public class GrpcObservationContextInfo {
 
-    private GrpcObservationContextInfo grpcInfo;
+    private String serviceName;
 
-    public GrpcClientObservationContext(Setter<Metadata> setter) {
-        super(setter);
-        this.grpcInfo = new GrpcObservationContextInfo();
+    private String methodName;
+
+    private String fullMethodName;
+
+    private MethodType methodType;
+
+    private Code statusCode;
+
+    private String authority;
+
+    private Metadata headers;
+
+    private Metadata trailers;
+
+    public GrpcObservationContextInfo(String serviceName, String methodName, String fullMethodName,
+            MethodType methodType, Code statusCode, String authority, Metadata headers, Metadata trailers) {
+        this.serviceName = serviceName;
+        this.methodName = methodName;
+        this.fullMethodName = fullMethodName;
+        this.methodType = methodType;
+        this.statusCode = statusCode;
+        this.authority = authority;
+        this.headers = headers;
+        this.trailers = trailers;
     }
 
-    // Delegates to grpcInfo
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcClientObservationContext.java
```diff
@@ -32,110 +32,77 @@
...
-
-    @Nullable
-    private Code statusCode;
-
-    private String authority;
-
-    private Metadata headers;
-
-    private Metadata trailers;
+    private GrpcObservationContextInfo grpcInfo;
```
jonatan-ivanov, line 35: I think in this particular case having a common `GrpcObservationContext` might be a better solution.

Also, I don't thin the automated refactoring too does what you claim:
>reducing complexity and enhancing readability of your source code.

I think this code is more complex and less readable than it was before. _In some cases_, if we need to change things, this could be an improvement.

Also now that there is an extra object inside of this class that is heavily used for very short-lived object, introducing this could also mean worse performance (didn't measure).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationContextInfo.java
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright 2022 the original author or authors.
```
jonatan-ivanov, line 2: Was this really reviewed by a human before opening the PR?
These mistakes are easy to make even for humans (or miss them during reviews) but it does not help me trusting the PR more.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationContextInfo.java
```diff
@@ -0,0 +1,126 @@
...
+
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor.MethodType;
+import io.grpc.Status.Code;
+import io.micrometer.common.lang.Nullable;
+
+/**
+ * Holds information relevant to gRPC client and server observation contexts.
+ */
+public class GrpcObservationContextInfo {
```
jonatan-ivanov, line 27: I guess this is a completely internal detail and it should not be public.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcClientObservationContext.java
```diff
@@ -32,110 +32,77 @@
...
 
+    // Delegates to grpcInfo
     public String getServiceName() {
-        return this.serviceName;
+        return grpcInfo.getServiceName();
     }
 
-    public void setServiceName(String serviceName) {
-        this.serviceName = serviceName;
+    public String getMethodName() {
```
jonatan-ivanov, line 47: Reordering the methods makes this PR hard to review and also, it is not what the PR claims as its goal. If you want to increase readability by reordering the methods, I think that should be a separate PR or it should not do such thing.

Also, removing javadoc is something that the automation should not do, it is removing valuable information.

compf:
Thank you for the constructive feedback. The wrong header year is my fault. the LLM didn't generate it and I had to copy. On the other points you are absolutely right and it is also true that not every data clump should be refactored like the LLM suggested :)

PR #5115, opened 2024-05-20
Fix quantile value for DistributionSummary for OtlpMeterRegistry

izeye:
This PR fixes quantile value for `DistributionSummary` for `OtlpMeterRegistry`.
Before this change values were always converted from nanos to the base time unit of the registry even if their unit was not time-based.


jonatan-ivanov:
I think we should fix this first in `1.12.x`: https://github.com/micrometer-metrics/micrometer/blob/6dbaa9dc7d6050024965805065488e770c53e358/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java#L389-L391

Could you please fix it there (we already have an `isTimeBased` indicator in-place) then I will do what you did in this PR during forward-merge (and add you as co-author).

jonatan-ivanov:
Oops I meant 1.9.x, I think this will not be as simple as I originally thought (because of the tests), let me push this through all the branches.

PR #5137, opened 2024-05-22
Add "cancelled" event to the gRPC observation instrumentation

### LINKED ISSUE #5109: Cancelled status code not reported in the gRPC server metrics
Hi,

cancelled status codes are not reported in the gRPC server metrics, metrics in these cases are reported without the tag status code because `GrpcServerObservationContext` is not updated with the cancelled. For us it's important the gRPC server metric give us  info of the status codes, and this could be improved for example modifying the `onCancel` method of the `ObservationGrpcServerCallListener` before running the stop operation, something like this:

```
  @Override
  public void onCancel() {
    try (final Scope scope = this.observation.openScope()) {
      super.onCancel();
    } finally {
      if (this.observation.getContext() instanceof final GrpcServerObservationContext obsCtx) {
        obsCtx.setStatusCode(Status.Code.CANCELLED);
      }
      this.observation.stop();
    }
  }
```

I have created a [demo](https://github.com/micrometer-metrics/micrometer/files/15345863/demo-grpc.zip) to reproduce easily cancelleds, the service has a delay of 5 seconds to give time to the client to cancel the request, you could make the call using postman or a simple grpc client with a deadline lower than 5 seconds.

Thanks for your time.

ttddyy:
When the gRPC server detects a cancelation of the client request, add `cancelled` event to the observation.

Closes #5109

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationDocumentation.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationDocumentation.java
@@ -121,6 +121,12 @@ public enum GrpcObservationDocumentation implements ObservationDocumentation {
                 return "sent";
             }
 
+        },
+        CANCELLED {
+            @Override
+            public String getName() {
+                return "cancelled";
+            }
         }
 
     }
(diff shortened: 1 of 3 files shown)


jonatan-ivanov (approved):

juliojgd:
Hello @jonatan-ivanov any estimated date to merge this PR? Thanks in advance

PR #5141, opened 2024-05-23
Fix log to include stack trace in OtlpMeterRegistry.publish()

izeye:
This PR fixes log to include stack trace in `OtlpMeterRegistry.publish()`.

See gh-4829


jonatan-ivanov:
Thank you!

Issue #2413, opened 2021-01-22
Description of process_start_time_seconds metric needs to be fixed in PushGateway as well as in the Prometheus client


vlaguduva:
As  per ["Writing libs from Prometheus website"](https://prometheus.io/docs/instrumenting/writing_clientlibs),  Help string of the metric "process_start_time_seconds" needs to be "Start time of the process since unix epoch in seconds." but Micrometer has "Start time of the process since unix epoch."; I understand that value kept in this metric is of unit milli seconds, nevertheless, the metric still named as seconds. Please change the help string as required. If required, please add additional metric process_start_time_milliseconds; This will also fix the long pending push gateway specific issue related to this very same metric.

shakuzen:
My understanding of that document is not that the help string MUST be exactly as shown there. What problem does it cause you for the help string to not match that exactly?

> I understand that value kept in this metric is of unit milli seconds

The value is not reported in milliseconds; it is scaled to seconds because that is what Prometheus expects. Other registries may have different preferences. We don't specify the final unit in the description because it differs by meter registry.

> This will also fix the long pending push gateway specific issue related to this very same metric.

I don't know what issue this is referencing.

shakuzen:
If you really wanted to change the description of that meter for the Prometheus registry, you could register a `MeterFilter` like the following on the PrometheusMeterRegistry (again note that you would not want to register this filter for other registries that do not use seconds as their base time unit):

```java
new MeterFilter() {
    @Override
    public Meter.Id map(Meter.Id id) {
        if (id.getName().equals("process.start.time")) {
            return new Meter.Id(id.getName(), Tags.of(id.getTagsAsIterable()), id.getBaseUnit(), "Start time of the process since unix epoch in seconds.", id.getType());
        }
        return id;
    }
}
```

Until there is some demonstrated problem to fix, I'm going to close this issue for now. Please let us know if I've missed something.

vlaguduva:
Thanks for providing more information.

While the metric description is not pressing for the regular client, it is breaking push gateway for sure. 
Example log :
> >> Metric families 'name:\"process_start_time_seconds\" help:\"Start time of the process since unix epoch.\" and name:\"process_start_time_seconds\" help:\"Start time of the process since unix epoch in seconds.\"....have inconsistent help strings. The latter will have priority. This is bad. Fix your pushed metrics!"

Fixing the metric description will make metric as required by Prometheus + fixes the push gateway issue.

[...]

PR #5144, opened 2024-05-24
fix: process_start_time_seconds help string

atompi:
We should refer to the Prometheus [documentation](https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics) to define the metrics.

According to Prometheus [documentation](https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics), the metric name to collect process start time is `process_start_time_seconds`, and help string is `Start time of the process since unix epoch in seconds.`.

This change not only can solve the problems encountered in the usage scenario of Pushgateway (like #2413), but also reflects respect for the official specifications, in order to reduce the complexity of common code.


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/UptimeMetrics.java
```diff
@@ -64,7 +64,7 @@ public void bindTo(MeterRegistry registry) {
 
         TimeGauge.builder("process.start.time", runtimeMXBean, TimeUnit.MILLISECONDS, RuntimeMXBean::getStartTime)
             .tags(tags)
-            .description("Start time of the process since unix epoch.")
+            .description("Start time of the process since unix epoch in seconds.")
```
lenin-jaganathan, line 67: I think adding seconds to the description is redundant and also may be inaccurate behavior. `TimeGauge` is reported in the base unit of registries choice. So, it may be seconds for Prometheus but it is not the case for other registries.

Either this has to be done within the scope of PrometheusMeterRegistry or should be done via a meter filter in the applications.

PR #5149, opened 2024-05-27
Do not register GC metrics when GC notifications are unavailable

codesimplicity:
This addresses the issue that GraalVM currently does not support GC notifications which means no useful GC metrics with Micrometer. Micrometer does not provide any hint and even registers a few useless metrics which only have initial values but never get updates as this relies on GC notifications.

Current GraalVM GCs show the following behaviour:
- Serial GC (default): two `GarbageCollectorMXBean` exposed but those don't support any notification types
- G1 GC (Enterprise Edition only): no `GarbageCollectorMXBean` exposed

Verified that the GCs actually work by enabling GC logging via `-XX:+PrintGC` - just to be sure.

See the [GraalVM issue](https://github.com/oracle/graal/issues/8237) for this missing feature for reference.


shakuzen (approved):
Thank you for the pull request and linking to the issue in GraalVM. I've subscribed to notifications (pun intended) on it. When it is resolved, we could consider adding some native image tests for GC metrics.

Issue #2141, opened 2020-06-06
Buffered statsd lines are not published on close


WillBrindle:
This may well be misunderstanding on my part but is there any way to ensure all counters are updated on close?
Here is a minimal example of the problem:
```
public static void main(String args[]) throws InterruptedException {
        StatsdConfig config = new StatsdConfig() {
            @Override
            public String get(String k) {
                return null;
            }
        };
        MeterRegistry registry = new StatsdMeterRegistry(config, Clock.SYSTEM);

        String METRIC = "test.metric";

        registry.counter(METRIC);
        registry.close();
}
```
If you spin up a simple udp server you'll see nothing is sent. This extends to if I add a sleep for a value more than `step()` , before or after closing. If I add a loop before closing where I increment the counter then sleep I start to get buffered messages across, but if not enough messages have been sent to clear the buffer by the last metrics are lost.

Setting buffered in my config to false resolves the problem but the natural behave feels like flushing should occur on close.

mickaeltr:
Isn't it the same issue as https://github.com/micrometer-metrics/micrometer/issues/1882 ?

jonatan-ivanov:
#1882 fixed this for `StepMeterRegistry` implementations while `StatsdMeterRegistry extends MeterRegistry`.

jgrigg:
I'm still seeing this behaviour when running any of the aforementioned versions. i.e doing `nc -ul 8125` and running @WillBrindle's original example yields nothing. The test looks right though :/

shakuzen:
@jgrigg could you open a new issue with the relevant details and refer to this issue? That will be better for tracking. Please try to reproduce the issue with a minimal sample.

izeye:
I could reproduce @jgrigg's comment with `nc -ulp 8125` command and [this sample](https://github.com/izeye/hello-micrometer/blob/statsd/src/main/java/com/izeye/helloworld/micrometer/Main.java).

`counter.increment()` seems to have been lost due to a missing UDP connection. If I add a second of sleep before it, it seems to work.

I created https://github.com/micrometer-metrics/micrometer/pull/6371 to demonstrate the status quo.

The reason why it's okay in the `StatsdMeterRegistryPublishTest` is that the tests wait for a UDP connection to be available as follows:

[...]

PR #5195, opened 2024-06-07
StatsdMeterRegistry - flush buffer on close

BJMg:
This PR should resolve [this issue](https://github.com/micrometer-metrics/micrometer/issues/2141).

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
@@ -26,6 +26,7 @@ import io.micrometer.core.instrument.util.HierarchicalNameMapper;
 import io.micrometer.core.lang.Nullable;
 import io.micrometer.core.util.internal.logging.WarnThenDebugLogger;
 import io.micrometer.statsd.internal.*;
+import io.netty.channel.Channel;
 import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.util.AttributeKey;
 import org.reactivestreams.Publisher;
@@ -92,7 +93,12 @@ public class StatsdMeterRegistry extends MeterRegistry {
 
     FluxSink<String> sink = new NoopFluxSink();
 
+    Flux<String> flux = Flux.empty();
+
     Disposable.Swap statsdConnection = Disposables.swap();
+
+    @Nullable
+    private Channel flushableChannel;
 
     private Disposable.Swap meterPoller = Disposables.swap();
 
@@ -219,6 +225,7 @@ public class StatsdMeterRegistry extends MeterRegistry {
             else {
                 final Publisher<String> publisher;
                 if (statsdConfig.buffered()) {
+                    flux = Flux.from(this.processor);
                     publisher = BufferingFlux
                         .create(Flux.from(this.processor), "\n", statsdConfig.maxPacketLength(),
                                 statsdConfig.pollingFrequency().toMillis())
@@ -295,6 +302,7 @@ public class StatsdMeterRegistry extends MeterRegistry {
     private void retryReplaceClient(Mono<? extends Connection> connectMono) {
         connectMono.retryWhen(Retry.backoff(Long.MAX_VALUE, Duration.ofSeconds(1)).maxBackoff(Duration.ofMinutes(1)))
             .subscribe(connection -> {
+                this.flushableChannel = connection.channel();
                 this.statsdConnection.replace(connection);
 
                 // now that we're connected, start polling gauges and other pollable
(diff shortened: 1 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -64,7 +64,7 @@
  * @author Tommy Ludwig
  * @author Marcin Grzejszczak
  */
-public abstract class MeterRegistry {
+public abstract class MeterRegistry implements AutoCloseable {
```
shakuzen, line 67: This change appears unnecessary to include with the rest of the changes. We would want to consider such a change separately.

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
```diff
@@ -92,8 +93,13 @@ public class StatsdMeterRegistry extends MeterRegistry {
 
     FluxSink<String> sink = new NoopFluxSink();
 
+    Flux<String> flux = Flux.empty();
```
shakuzen, line 96: It doesn't look like this is used. Am I missing something?

shakuzen (commented):
Thanks for the pull request. I'm surprised we would need to use Netty API to achieve this instead of using Reactor Netty API, but I am no expert in this area. @violetagg would you have any spare time to review this?

violetagg:
> Thanks for the pull request. I'm surprised we would need to use Netty API to achieve this instead of using Reactor Netty API, but I am no expert in this area. @violetagg would you have any spare time to review this?

Let me take a look

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java
BJMg: Yes, you are right. I just kind of forgot it there from some previous attempt.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
BJMg: Sure if you insist, but it would be nice to have it there, because IMO it makes it easier to use, especially with some java frameworks such spring, which then automatically closes its beans on shutdown.

shakuzen: Spring already calls `close` on Bean types that have it whether they are `AutoCloseable` or not. See [the documentation](https://docs.spring.io/spring-framework/reference/core/beans/factory-nature.html#beans-factory-lifecycle-disposablebean):

> Note that Spring also supports inference of destroy methods, detecting a public `close` or `shutdown` method. This is the default behavior for `@Bean` methods in Java configuration classes and automatically matches `java.lang.AutoCloseable` or `java.io.Closeable` implementations, not coupling the destruction logic to Spring either.

AutoCloseable makes sense for things expected to be used with try-with-resources, but I don't think `MeterRegistry` falls into that category for most uses. By making it `AutoCloseable`, IDEs will start warning users they aren't closing `MeterRegistry`, but in most cases, the framework in use will take care of it, so users should rather not call close themselves.

BJMg: I deleted it, but to be honest I think it would be nice to let users know that the resource need to be closed otherwise they might lose some unsent metrics. Similar to some buffering streams that need to be closed so they would flush all data.

violetagg:
> Thanks for the pull request. I'm surprised we would need to use Netty API to achieve this instead of using Reactor Netty API, but I am no expert in this area. @violetagg would you have any spare time to review this?

@shakuzen By default Reactor Netty tries to flush batches, what happens here is that between write and actual flush operation, close operation happens that discards the data that is not flushed. We either proceed with this patch, or you can switch to flush after every element (which might have some overhead).
https://projectreactor.io/docs/netty/release/api/reactor/netty/NettyOutbound.html#send-org.reactivestreams.Publisher-java.util.function.Predicate-

shakuzen:
@violetagg thank you for taking a look. I'm surprised specifically because we're already doing `statsdConnection.get().dispose()` in our `stop`. The issue is that calling `dispose` on the `Connection` will not flush the channel, right? In that case, I think this change makes sense.

shakuzen (approved):
@BJMg thank you for the pull request and quick action on review. I think we're good to merge this. I've rebased the changes onto the `1.9.x` branch so we can include this in upcoming patch releases.

violetagg:
> @violetagg thank you for taking a look. I'm surprised specifically because we're already doing `statsdConnection.get().dispose()` in our `stop`. The issue is that calling `dispose` on the `Connection` will not flush the channel, right? In that case, I think this change makes sense.

Yes you invoke `dispose`, which triggers `close` before our `flush`, as a result the data that is not flushed is discarded.

PR #5222, opened 2024-06-12
Add JvmThreadDeadlockMetrics

rkurniawati:
Hello,

I was wondering if you would consider accepting the changes in this PR so that we will have these two deadlock-related metrics in a future version of Micrometer?

This PR adds the following JVM thread deadlock-related metrics:
- `jvm.threads.deadlocked`: The current number of threads that are deadlocked
- `jvm.threads.deadlocked.monitor`: The current number of threads that are deadlocked on object monitors

Thanks,
Ruth

Ruth Kurniawati

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2024 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.jvm;
+
+import io.micrometer.common.lang.NonNullApi;
+import io.micrometer.common.lang.NonNullFields;
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.BaseUnits;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadMXBean;
+
+import static java.util.Collections.emptyList;
+
+/**
+ * {@link MeterBinder} for JVM deadlocked threads. These metrics are somewhat expensive to
+ * collect, so they should not be enabled unless necessary. To enable these metrics, add
+ * the following to your application's code ({@code registry} is the meter registry being
+ * used by your application):
+ * <pre>{@code new JvmDeadlockedThreadMetrics().bindTo(registry);}</pre>
+ *
+ * @author Ruth Kurniawati
+ */
+@NonNullApi
+@NonNullFields
+public class JvmThreadDeadlockMetrics implements MeterBinder {
+
+    private static final InternalLogger log = InternalLoggerFactory.getInstance(JvmThreadDeadlockMetrics.class);
+
+    private final Iterable<Tag> tags;
+
+    public JvmThreadDeadlockMetrics() {
+        this(emptyList());
+    }
+
+    public JvmThreadDeadlockMetrics(Iterable<Tag> tags) {
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        ThreadMXBean threadBean = ManagementFactory.getThreadMXBean();
+
+        if (threadBean.isObjectMonitorUsageSupported()) {
+            Gauge.builder("jvm.threads.deadlocked", threadBean, JvmThreadDeadlockMetrics::getDeadlockedThreadCount)
+                .tags(tags)
+                .description("The current number of threads that are deadlocked")
+                .baseUnit(BaseUnits.THREADS)
+                .register(registry);
+        }
+        else {
+            log.warn("jvm.threads.deadlocked is not available on this JVM");
+        }
+
+        Gauge
+            .builder("jvm.threads.deadlocked.monitor", threadBean,
+                    JvmThreadDeadlockMetrics::getDeadlockedMonitorThreadCount)
+            .tags(tags)
+            .description("The current number of threads that are deadlocked on object monitors")
+            .baseUnit(BaseUnits.THREADS)
+            .register(registry);
+    }
+
+    // VisibleForTesting
+    static long getDeadlockedThreadCount(ThreadMXBean threadBean) {
+        final long[] deadlockedThreads = threadBean.findDeadlockedThreads();
+        return deadlockedThreads == null ? 0 : deadlockedThreads.length;
+    }
+
+    static long getDeadlockedMonitorThreadCount(ThreadMXBean threadBean) {
+        final long[] monitorDeadlockedThreads = threadBean.findMonitorDeadlockedThreads();
+        return monitorDeadlockedThreads == null ? 0 : monitorDeadlockedThreads.length;
+    }
+
+}
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetricsTest.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2024 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


shakuzen (commented):
Thank you for the pull request. Those sound like interesting additions. I am concerned though that the JavaDoc for both have the following warning (emphasis is mine):

> This method is designed for troubleshooting use, but not for synchronization control. **It might be an expensive operation.**

Due to that it might make more sense to put these in a separate class so users can more easily separately choose to enable these potentially expensive metrics. What do you think? Do you have any practical experience using these in production that you can share in terms of the expense of regularly calling these methods?

rkurniawati:
Thank you for your feedback, @shakuzen! I agree it's a good idea to let the users enable these metrics when they need them. The APIs used by the deadlock metrics have some overhead in the JVM implementation that I use, calling the `findDeadlockedThreads` triggers a JVM `safepoint` which will cause a slowdown that may be significant to some applications. 

When you suggest putting these in a separate class, do you mean put them in a separate metric binder?  It would be great if you have an example of a metric that's disabled by default. 

Thanks,
Ruth

shakuzen:
> When you suggest putting these in a separate class, do you mean put them in a separate metric binder?

Yes. By doing that, users can bind `JvmThreadMetrics` without these potentially heavy metrics being registered. When users want these heavy metrics, they should make the choice to bind the new binder containing them. It's up to users. Everything is disabled by default in the sense that no meters are registered until you bind a MeterBinder or register a meter.

rkurniawati:
Thanks! I refactored the code and put the new metrics in a class called `JvmThreadDeadlockMetrics`. I've verified that these metrics are not enabled by Spring Boot's auto-configuration, but can be enabled by binding the `JvmDeadlockedThreadMetrics` binder to the registry.

Let me know if there's anything else that I could do to improve this PR 😃

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
```diff
@@ -0,0 +1,83 @@
...
+                .baseUnit(BaseUnits.THREADS)
+                .register(registry);
+        }
+        catch (UnsupportedOperationException ignored) {
+        }
+    }
+
+    // VisibleForTesting
+    static long getDeadlockedThreadCount(ThreadMXBean threadBean) {
+        return threadBean.findDeadlockedThreads() == null ? 0 : threadBean.findDeadlockedThreads().length;
```
shakuzen, line 75: Given these calls can be costly, it'd be better to avoid calling it twice. It could also return null on the second call. Same for the code in the other method.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
```diff
@@ -0,0 +1,83 @@
...
+
+            Gauge
+                .builder("jvm.threads.deadlocked.monitor", threadBean,
+                        JvmThreadDeadlockMetrics::getDeadlockedMonitorThreadCount)
+                .tags(tags)
+                .description("The current number of threads that are deadlocked on object monitors")
+                .baseUnit(BaseUnits.THREADS)
+                .register(registry);
+        }
+        catch (UnsupportedOperationException ignored) {
```
shakuzen, line 69: I don't think there's anything in the try block that throws this. The methods on the ThreadMXBean may throw this but they are not called in this try block. They will be called at publish time by the thread doing publishing.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
```diff
@@ -0,0 +1,83 @@
...
+import io.micrometer.core.instrument.binder.BaseUnits;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadMXBean;
+
+import static java.util.Collections.emptyList;
+
+/**
+ * {@link MeterBinder} for JVM deadlocked threads.
```
shakuzen, line 32: We should warn in the JavaDoc that these metrics may be expensive to collect and thus care should be taken in deciding to bind this meter binder or not.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetricsTest.java
```diff
@@ -0,0 +1,94 @@
...
+    private void sleep(int seconds) {
+        try {
+            TimeUnit.SECONDS.sleep(seconds);
+        }
+        catch (InterruptedException ignored) {
+        }
+    }
+
+    private void createTemporarilyDeadlockedThread() {
+        // create two threads that are deadlocked for about 5 seconds
```
shakuzen, line 68: Could we control this with something like CountDownLatch so we can have the same effect without the test always taking multiple seconds?

shakuzen (changes requested):
Thank you for the changes. I left some review comments.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
rkurniawati: Good point, I've changed the code so that the `find*Deadlockedthread` methods are only called once in those methods.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
rkurniawati: Ah - thanks for pointing that out. I've removed the try/catch

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
rkurniawati: I've added a warning in the JavaDoc. Let me know if it needs to be more ominous

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetricsTest.java
rkurniawati: I changed the test to use `CountDownLatch`. Unfortunately I have to use `ReentrantLock` instead of the monitor lock, so the the `jvm.threads.deadlocked` is non-zero but `jvm.threads.deadlocked.monitor` is zero

rkurniawati:
It looks like the CI job failed due to a flaky test (##5296). I'm going to try to rebase so that it will trigger CI (and hopefully pass the test next time)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public JvmThreadDeadlockMetrics(Iterable<Tag> tags) {
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        ThreadMXBean threadBean = ManagementFactory.getThreadMXBean();
+
+        Gauge.builder("jvm.threads.deadlocked", threadBean, JvmThreadDeadlockMetrics::getDeadlockedThreadCount)
```
shakuzen, line 58: While the `UnsupportedOperationException` won't be thrown at this time, it doesn't make sense to register a Gauge that will knowingly throw an exception every time the value function is called. We should call ThreadMXBean::findDeadlockedThreads in a try catch block so this Gauge is only registered if the operation is supported. We can log a message instead if it isn't supported.

shakuzen (commented):
Thank you for the quick updates. One more thing and then this is ready to go I think.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
rkurniawati: Makes sense - thanks for the suggestion! I've made the changes and added a unit test for it.

shakuzen: Thanks again for the quick turnaround. I think ThreadMXBean::findMonitorDeadlockedThreads does not throw `UnsupportedOperationException` which I take to mean it should always be available. Was it intentional that you included both gauges in the try-catch block, meaning both won't be registered? Would it not be useful to users to have the monitor deadlock gauge even if the other is unavailable?

shakuzen: I also noticed we can call ThreadMXBean::isSynchronizerUsageSupported rather than using a try-catch block with ThreadMXBean::findDeadlockedThreads.

rkurniawati: Thanks @shakuzen! I've changed the code to use `ThreadMXBean::isSynchronizerUsageSupported` instead of using a try/catch for `UnsupportedOperationException`. Since`ThreadMXBean::findMonitorDeadlockedThreads` is always available, I've made the registration of `jvm.threads.deadlocked.monitor` unconditional.

shakuzen (approved):
Thank you for the pull request and the patience on review.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadDeadlockMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public JvmThreadDeadlockMetrics(Iterable<Tag> tags) {
+        this.tags = tags;
+    }
+
+    @Override
+    public void bindTo(MeterRegistry registry) {
+        ThreadMXBean threadBean = ManagementFactory.getThreadMXBean();
+
+        if (threadBean.isObjectMonitorUsageSupported()) {
```
shakuzen, line 62: This should be `isSynchronizerUsageSupported`. I'll fix it in a polish commit.

rkurniawati: Oops! Thanks for taking care of it!

PR #5226, opened 2024-06-19
Upgrade CI machine image to ubuntu-2204:2024.05.1

izeye:
This PR upgrades CI machine image to ubuntu-2204:2024.05.1.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ executors:
   machine-executor:
     working_directory: ~/micrometer
     machine:
-      image: ubuntu-2204:2024.01.2
+      image: ubuntu-2204:2024.05.1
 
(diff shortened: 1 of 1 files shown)


shakuzen:
Looking at the test failure, I wonder if the OS version upgrade results in a version of Java used that systemstubs is not compatible with.

shakuzen:
I think maybe we can avoid the issue by applying this to `1.12.x` and forward. The discrepancy between the branches will be unfortunate, but it's probably not worth the effort to figure out how to otherwise work around this. Unless anyone has a quick, easy fix in mind.

izeye:
@shakuzen Thanks for the feedback! I switched the base branch to the 1.12.x as you suggested.

Issue #5192, opened 2024-06-04
Metric will be skipped when producing scrape output for PrometheusMeterRegistry


danielksb:
**Describe the bug**
If two metrics have the same name but different tags, they will be registered in the PrometheusMeterRegistry as two distinct meters.
After calling `PrometheusMeterRegistry::scrape` one of the two meters will not be part of the output. See below for an example.

**Environment**
 - Micrometer Registry Prometheus: 1.12.2, 1.13.0
 - OS: tested on Linux and Windows
 - Java version: tested on OpenJDK 11.0.21, OpenJDK 21.0.2

**To Reproduce**
How to reproduce the bug:
```java
package org.example;

import io.micrometer.core.instrument.Counter;
import io.micrometer.core.instrument.Meter;
import io.micrometer.prometheusmetrics.PrometheusConfig;
import io.micrometer.prometheusmetrics.PrometheusMeterRegistry;

import java.util.stream.Collectors;

public class Main {
    public static void main(String[] args) {
        PrometheusMeterRegistry meterRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
        Counter.builder("test")
               .tag("status", "aaa")
               .register(meterRegistry)
               .increment();

        Counter.builder("test")
               .tag("status", "bbb")
               .tag("tag_for_status_bbb", "xyz") // adding an additional tag seems to be the cause of the problem
               .register(meterRegistry)
               .increment();


        // We check the content of the registry by simply printing the IDs of the registered meters.
        System.out.println(
                meterRegistry.getMeters().stream()
                             .map(Meter::getId)
                             .map(Meter.Id::toString)
                             .collect(Collectors.joining("\n")));
        // Output will be:
        // MeterId{name='test', tags=[tag(status=aaa)]}
        // MeterId{name='test', tags=[tag(status=bbb),tag(tag_for_status_bbb=xyz)]}

        // The registry contains two meters, this is the expected behaviour.
        // Now we create the output for the Prometheus scrape API.

        System.out.println(meterRegistry.scrape());
        // Output will be:
        // # HELP test_total
        // # TYPE test_total counter
        // test_total{status="aaa"} 1.0

        // The entry for the second meter is missing.
    }
}
```

**Expected behavior**
The function call to  `PrometheusMeterRegistry::scrape` should contain entries for all registered meters:
```
# HELP test_total  
# TYPE test_total counter
test_total{status="aaa"} 1.0
[...]

Issue #2068, opened 2020-05-08
Add onMeterRegistrationFailed listener configuration to MeterRegistry


jkschneider:
This is mainly driven by the significant burden Prometheus' requirement places on everyone for the same metric name to have the same set of tags.

Previously, `PrometheusMeterRegistry` threw an `IllegalArgumentException`. Now, we'll _not_ throw an exception but instead silently continue, but inform any callbacks registered that this `Meter.Id` could not be registered.

In this way, the user can do whatever they'd like with these problems, including logging or propagating a hard failure if they choose.

See issues like https://github.com/micrometer-metrics/micrometer/issues/1253 for motivation.

checketts:
The tag mismatch sometimes happens due to two different libraries creating the same meter name and not matching on tags; inetty and httpclient I believe hit this at one time.

Would a user be able to use a MeterFilter to clean up those tags if they got into that situation?

Sure, the best solution is to have the libraries not step on each other, but I just wanted to think through what options would be available to users when hitting this exception.

jkschneider:
> Would a user be able to use a MeterFilter to clean up those tags if they got into that situation?

`MeterFilter` is already an option. Once you see the implementation, I think you'll see this listener is called as a last resort after filters have been applied and there's still a problem.

jkschneider:
Also included here is a new method on `PrometheusMeterRegistry` called `throwExceptionOnRegistrationFailure()` which sets up such a listener that throws the same unchecked exception it did before. But now this behavior is _opt-in_.

jkschneider:
cc / @izeye because I know you've been squashing all the tag problems lately (and many thanks for this)! Just want you to be aware of the different behavior.

PR #5228, opened 2024-06-23
Warn about Prometheus meter registration failure

izeye:
A meter registration failure seems to be an important event for users, but it's not easy for the users to be aware of the failure unless they register a callback for it.

It would be better to make the users easily aware of the failures by default.

See gh-5192

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
@@ -16,6 +16,7 @@
 package io.micrometer.prometheusmetrics;
 
 import io.micrometer.common.lang.Nullable;
+import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.cumulative.CumulativeFunctionCounter;
 import io.micrometer.core.instrument.cumulative.CumulativeFunctionTimer;
@@ -67,6 +68,8 @@ import static java.util.stream.StreamSupport.stream;
  * @since 1.13.0
  */
 public class PrometheusMeterRegistry extends MeterRegistry {
+
+    private static final WarnThenDebugLogger log = new WarnThenDebugLogger(PrometheusMeterRegistry.class);
 
     private final PrometheusConfig prometheusConfig;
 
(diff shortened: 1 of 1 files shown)


shakuzen:
I don't think we should do this by default. https://github.com/micrometer-metrics/micrometer/issues/2068 has some history. I understand the usefulness of being able to identify these issues with a log message, but it's important to balance this with the principle that observability impact to your application should be minimal - I think this includes avoiding logging too much. It's better to lose some observability info than cause issues for the application. Logging this is far more reasonable if we're only talking about meter registration failures caused by instrumentation issues in code that users control in their application code, but it's a problem when we're talking about instrumentation issues in code that users don't control (third-party libraries). I believe that's essentially why we changed the default to doing nothing in Prometheus on meter registration failure due to the known limitation but we added a way (`MeterRegistry#onMeterRegistrationFailed`) to configure logging if that's what users want (possibly locally or in a test environment rather than production). That's still configurable for users without this change, but I can appreciate the point that it's hard for users to know they should configure this because it's hard to realize the problem without logging. Perhaps as a compromise, we could use a WarnThenDebugLogger. Though this will result in double logging if a user also configures logging via `onMeterRegistrationFailed`. @jonatan-ivanov thoughts? You mentioned a concern elsewhere about duplicate logging when a composite meter registry is used.

izeye:
@shakuzen Thanks for the explanation!

Using a `WarnThenDebugLogger` seems to be a good compromise for both.

jonatan-ivanov:
I like `WarnThenDebugLogger`, I would still keep the option open of making the feature opt-in (off by default). Similar to Prometheus' `registry.throwExceptionOnRegistrationFailure()` (maybe by adding a listener that does this instead of modifying the `meterRegistrationFailed` method).

Also, I think if you have a composite registry, and all the registries will report meter registration failures by default, you will get this message multiple times, e.g.: composite(prometheus, simple) will report this 3 times. With an extra config option you can control which registries are ok to report such an issue and if the feature is off by default it won't disrupt anything, then you can turn it on for the registry of your choice if you want.

What do you think?

shakuzen:
I don't want to keep this PR in a state of limbo. I think given the above discussion, it maybe is best to not do something in MeterRegistry by default that can't be overridden due to duplicate logging with multiple registries. I know this could be a generic issue and the reasoning behind wanting to do this potentially applies to any registry - it's hard to know you have meter registration failures in the first place if nothing is logged. But since the original issue is about the known registration failure that happens with the Prometheus registry, maybe a compromise is to use a WarnThenDebugLogger in the Prometheus registry only for now. Thoughts?

izeye:
@jonatan-ivanov @shakuzen Thanks for the feedback!

> a compromise is to use a WarnThenDebugLogger in the Prometheus registry only for now.

I updated this PR in this direction.

shakuzen:
Thanks for quickly updating the PR. I've polished things a bit in https://github.com/micrometer-metrics/micrometer/commit/c799962c13aaf35d23982fce6b581bb0bfa6fb1f

PR #5231, opened 2024-06-24
Log warning message with WARN level in logWarningAboutLateFilter() when DEBUG level is enabled

izeye:
Lowering log level for the warning message when DEBUG level is enabled seems to be inconsistent.

See gh-4917


shakuzen:
Having everything in one log message was intentional so there was no chance of the context being split across log messages that possibly have other log messages in between them when viewing the logs. I can see your point about there being no warn log in the case of debug logging being enabled, but I think I imagined most users will not have debug enabled initially until they see the warning and then they will enable debug logging. In that sense, they will have likely seen the warn level log already and be looking for the stacktrace in the debug level log. That's not guaranteed, though. I suppose it's a tradeoff one way or the other. Does that make sense? Am I missing anything?
@jonatan-ivanov thoughts?

izeye:

Is it okay if the split messages were merged into one as before but with WARN level?

izeye:
> Is it okay if the split messages were merged into one as before but with WARN level?

Spring Framework seems to have a similar handling as follows: https://github.com/spring-projects/spring-framework/blob/0ea7af746521118f9f562ed3ea7da5c6966a99bd/spring-beans/src/main/java/org/springframework/beans/factory/annotation/InitDestroyAnnotationBeanPostProcessor.java#L243-L248

shakuzen:
I suppose that's a decent compromise. Let's go with that. I do worry a little if it will confuse users to see a stacktrace at WARN level when they have DEBUG enabled, but Spring Framework doing so is a decent precedent to follow.

shakuzen (approved):

PR #5236, opened 2024-06-28
Expose ForkJoinPool parallelism and pool size metrics

genuss:
Hello, contributors,

I recently switched to using virtual threads, so became a big user of ForkJoinPool. When I looked at metrics, exposed by micrometer, I saw that parallelism isn't exposed. Is there a reason for this? Well, if not, I believe this could be a nice addition to already exposed ones.

I hesitated to add [ForkJoinPool::getPoolSize](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ForkJoinPool.html#getPoolSize()), as I'm not sure this one is important, but I could add it too. What do you think about it?

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
@@ -59,6 +59,8 @@ public class ExecutorServiceMetrics implements MeterBinder {
 
     private static final String DEFAULT_EXECUTOR_METRIC_PREFIX = "";
 
+    public static final String DESCRIPTION_POOL_SIZE = "The current number of threads in the pool";
+
     @Nullable
     private final ExecutorService executorService;
 
@@ -362,7 +364,7 @@ public class ExecutorServiceMetrics implements MeterBinder {
 
         Gauge.builder(metricPrefix + "executor.pool.size", tp, ThreadPoolExecutor::getPoolSize)
             .tags(tags)
-            .description("The current number of threads in the pool")
+            .description(DESCRIPTION_POOL_SIZE)
             .baseUnit(BaseUnits.THREADS)
             .register(registry);
 
@@ -385,32 +387,38 @@ public class ExecutorServiceMetrics implements MeterBinder {
             .description("Estimate of the total number of tasks stolen from "
                     + "one thread's work queue by another. The reported value "
                     + "underestimates the actual total number of steals when the pool " + "is not quiescent")
+            .baseUnit(BaseUnits.TASKS)
             .register(registry);
 
         Gauge.builder(metricPrefix + "executor.queued", fj, ForkJoinPool::getQueuedTaskCount)
             .tags(tags)
             .description("An estimate of the total number of tasks currently held in queues by worker threads")
+            .baseUnit(BaseUnits.TASKS)
             .register(registry);
 
         Gauge.builder(metricPrefix + "executor.active", fj, ForkJoinPool::getActiveThreadCount)
             .tags(tags)
             .description("An estimate of the number of threads that are currently stealing or executing tasks")
+            .baseUnit(BaseUnits.THREADS)
             .register(registry);
 
         Gauge.builder(metricPrefix + "executor.running", fj, ForkJoinPool::getRunningThreadCount)
             .tags(tags)
             .description(
                     "An estimate of the number of worker threads that are not blocked waiting to join tasks or for other managed synchronization threads")
+            .baseUnit(BaseUnits.THREADS)
             .register(registry);
 
         Gauge.builder(metricPrefix + "executor.parallelism", fj, ForkJoinPool::getParallelism)
             .tags(tags)
             .description("The targeted parallelism level of this pool")
+            .baseUnit(BaseUnits.THREADS)
             .register(registry);
 
         Gauge.builder(metricPrefix + "executor.pool.size", fj, ForkJoinPool::getPoolSize)
             .tags(tags)
-            .description("The number of worker threads that have started but not yet terminated")
+            .description(DESCRIPTION_POOL_SIZE)
+            .baseUnit(BaseUnits.THREADS)
             .register(registry);
     }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
@@ -407,6 +407,11 @@ public class ExecutorServiceMetrics implements MeterBinder {
             .tags(tags)
             .description("The targeted parallelism level of this pool")
             .register(registry);
+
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
>I saw that parallelism isn't exposed. Is there a reason for this? 

I think no one asked for it so far. What is your use-case behind parallelism? Are you setting it explicitly? By default, it uses the value of `Runtime.getRuntime().availableProcessors()` (cpu count) which Micrometer already reports but if you set it explicitly, it can have other values.

I think based on the use-case, the pool size can also be useful since its value will be the same as parallelism if enough tasks are submitted but if not, it can be lower, please feel free to add a `Gauge` for that too.

I also wanted to ask you to write some tests but then I realized there are no tests for `ForkJoinPool` in `ExecutorServiceMetricsTest`. :(
Let me know if you want to add some, otherwise I will take a look at them.

/cc @shakuzen

genuss:
> I think no one asked for it so far. What is your use-case behind parallelism? Are you setting it explicitly? By default, it uses the value of `Runtime.getRuntime().availableProcessors()` (cpu count) which Micrometer already reports but if you set it explicitly, it can have other values.

Well, I don't do it now, but I might in future as in container environment JVM's heuristics don't always work correct. Anyway, it's possible to create a new `ForkJoinPool` apart from the common one. In this case `availableProcessors` won't work too.

> I also wanted to ask you to write some tests but then I realized there are no tests for ForkJoinPool in ExecutorServiceMetricsTest. :(

I added a simple one. Please tell me if you need more.

genuss:
The failed test is this one:
```
JvmGcMetricsTest > gcTimingIsCorrectForPauseCycleCollectors() FAILED
    expected: 10L
     but was: 9L
        at app//io.micrometer.core.instrument.binder.jvm.JvmGcMetricsTest.lambda$checkPhaseCount$3(JvmGcMetricsTest.java:236)

6 tests completed, 1 failed

> Task :micrometer-core:zgcGenerationalTest FAILED
```

Doesn't look like a related failure, does it?

jonatan-ivanov:
>I might in future as in container environment JVM's heuristics don't always work correct

Yepp, that might be a valid scenario, though in that case upgrading Java might be a simpler task than modifying every place you use a pool. I think there could also be a valid (but very rare) use-case even if ergonomics sets the cpu size and the pool sizes correctly. In containerized environments, you can modify cgroup limits while the process is running. This can mean that the cpu count can increase or decrease over time and your pools might not. I haven't tried if `ForkJoinPool` does or not but even if it does, bugs can arise so knowing the parallelism level and the current pool size seems like a good idea to me.

>I added a simple one. Please tell me if you need more.

Thank you! I also forgot to say thank you for opening the PR in the first place so double thank you! :)
I think we should add more but that should not effect this PR, I will add those tests separately. I think I would add some basic verifications for the values in the test of this PR but I can do those too during merging it.

>Doesn't look like a related failure, does it?

No, that test is flaky which we should fix, I re-triggered the failed job and now it's green.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -402,6 +402,16 @@ private void monitor(MeterRegistry registry, ForkJoinPool fj) {
...
             .register(registry);
+
+        Gauge.builder(metricPrefix + "executor.parallelism", fj, ForkJoinPool::getParallelism)
+            .tags(tags)
+            .description("The targeted parallelism level of this pool")
+            .register(registry);
+
+        Gauge.builder(metricPrefix + "executor.pool.size", fj, ForkJoinPool::getPoolSize)
+            .tags(tags)
+            .description("The number of worker threads that have started but not yet terminated")
```
shakuzen, line 413: If we're using the same metric name, we should keep the description the same as the one for ThreadPoolExecutor to avoid issues with backends that don't accept different descriptions (see e.g. https://github.com/micrometer-metrics/micrometer/pull/3201)

> The current number of threads in the pool

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -402,6 +402,16 @@ private void monitor(MeterRegistry registry, ForkJoinPool fj) {
...
+            .register(registry);
```
shakuzen, line 414: The unit should also be `BaseUnits.THREADS` to match the corresponding meter for ThreadPoolExecutor.

shakuzen (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
genuss: Wow, that's unexpected. FIxed

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -59,6 +59,8 @@ public class ExecutorServiceMetrics implements MeterBinder {
 
     private static final String DEFAULT_EXECUTOR_METRIC_PREFIX = "";
 
+    public static final String DESCRIPTION_POOL_SIZE = "The current number of threads in the pool";
```
shakuzen, line 62: Let's keep it private unless there's a good reason for it to be public.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -385,22 +387,38 @@ private void monitor(MeterRegistry registry, ForkJoinPool fj) {
...
         Gauge.builder(metricPrefix + "executor.queued", fj, ForkJoinPool::getQueuedTaskCount)
             .tags(tags)
             .description("An estimate of the total number of tasks currently held in queues by worker threads")
+            .baseUnit(BaseUnits.TASKS)
             .register(registry);
 
         Gauge.builder(metricPrefix + "executor.active", fj, ForkJoinPool::getActiveThreadCount)
             .tags(tags)
             .description("An estimate of the number of threads that are currently stealing or executing tasks")
+            .baseUnit(BaseUnits.THREADS)
```
shakuzen, line 402: Adding units to existing metrics may be a breaking change depending on the backend. At the least, let's keep that change separate from this PR so one doesn't block the other.

shakuzen (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
genuss: Oh, sorry, something wrong with my autocomplete. Fixed

jonatan-ivanov (approved):

shakuzen (approved):

genuss:
Thank you, folks!

shakuzen:
Thank you for the pull request and the quick action on reviews.

PR #5271, opened 2024-07-03
Improve performance of NamingConvention#toSnakeCase

michaldo:


micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
```diff
@@ -57,7 +57,7 @@ public String tagKey(String key) {
         }
 
         private String toSnakeCase(String value) {
-            return Arrays.stream(value.split("\\.")).filter(Objects::nonNull).collect(Collectors.joining("_"));
+            return value.replace(".", "_");
```
shakuzen, line 60: It's subtle, but this does not have the same behavior as the previous code. See, for example, the following test on the `main` branch (i.e. without this code change):

```java
@Test
void snakeCaseReplace() {
    String name = "a.Name.with.Words.";
    String expectedConventionName = "a_Name_with_Words";
    String actualConventionName;

    actualConventionName = NamingConvention.snakeCase.name(name, Meter.Type.COUNTER);
    assertThat(actualConventionName).isEqualTo(expectedConventionName);

    actualConventionName = name.replace(".", "_");
    assertThat(actualConventionName).isEqualTo(expectedConventionName);
}
```

The second assertion fails:

```
expected: "a_Name_with_Words"
 but was: "a_Name_with_Words_"
```

It's unfortunate we don't have existing tests to cover a more complete set of expected input and output, but it would be risky to change this behavior and cause potential regressions for users.

shakuzen (commented):
Improving the performance of this would be a nice enhancement for many users, but we need to be careful as the impact of issues here could be large.

micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
michaldo: If https://github.com/micrometer-metrics/micrometer/pull/5288 increases performance and is more consistent with module design, I'm fine and this PR can be closed.

According to "subtle behavior change", it should be documented how metric name is changed for Prometheus. Then it will be possible to say if removing leading and trailing docs is part of contract or random side effect

jonatan-ivanov: I think some benchmarks (JMH) would be also nice to prove that this is in fact an improvement. Regarding #5288, this might not be the root cause since this line is there for 7 years and as far as I understand the issue is new.

michaldo: Here you are:

```
@Benchmark
 public void snakeStream(Blackhole blackhole) {
     String value = "http.client.requests.active.seconds.max";
     String snak = Arrays.stream(value.split("\\.")).filter(Objects::nonNull).collect(Collectors.joining("_"));
     blackhole.consume(snak);
 }

 @Benchmark
 public void snakeSimple(Blackhole blackhole) {
     String value = "http.client.requests.active.seconds.max";
     String snak = value.replace(".", "_");
     blackhole.consume(snak);
 }
```

```
Benchmark               Mode  Cnt         Score         Error  Units
SnakeCase.snakeSimple  thrpt   25  49690145,331 ± 3281631,270  ops/s
SnakeCase.snakeStream  thrpt   25   3717968,300 ±  101394,596  ops/s
```

Issue #5229, opened 2024-06-24
Prometheus Java client 1.x consumes more memory than 0.x


michaldo:
I observed that micrometer registry consumes more memory when I switch Spring Boot from 3.2.6 to 3.3.1

**Environment**

 - Micrometer version 1.13.1
 - Micrometer registry prometheus
 - Java version: Java 21

I attached profiler report collected by async profiler.

Green is memory consumption for micrometer-registry-prometheus: `scrape()` takes 446 MB
Red is memory consumption for micrometer-registry-prometheus-simpleclient (I keep Spring Boot 3.3.1 but switch to simple client): 121 MB

![image](https://github.com/micrometer-metrics/micrometer/assets/4304236/90e45a72-b6d2-45b9-8d6a-839e4cc335e0)

jonatan-ivanov:
Could you please verify if the two outputs are the same and you are not having four times more time series in the case of Prometheus 1.x?

How many time series (or lines) you have in your Prometheus output approximately?

Also, will the second and subsequent scrapes consume equal amount of memory or are they more lightweight?

What tool you used for the flame graph?

Will you get similar results if you look at a heap dump?

Also, Micrometer's overhead on the flame graphs seems pretty minimal to me, depending on the answer of the questions above, maybe this should be a Prometheus Client question at the end?

michaldo:
Hi Jonatan,
I check memory usage on test cloud cluster when only micrometer-registry-prometheus is changed.
So answer for question 1 and 3 is that environment is stable and memory consumption is stable.

It may be important that my application heavy use Kafka and prometheous output contains 4703 Kafka rows out of 4850 total
Flame graph is generated with Intellij

After profiler output inspection, suspected fragments of code are `io.micrometer.core.instrument.config.NamingConvention#toSnakeCase` and `io.prometheus.metrics.model.snapshots.PrometheusNaming#replaceIllegalCharsInMetricName` . They can be refactor with less memory requirements

jonatan-ivanov:
>After profiler output inspection, suspected fragments of code are `io.micrometer.core.instrument.config.NamingConvention#toSnakeCase` and `io.prometheus.metrics.model.snapshots.PrometheusNaming#replaceIllegalCharsInMetricName`. They can be refactor with less memory requirements

This conflicts with your flame graphs, I don't see any of these classes on it.
[...]

PR #5288, opened 2024-07-08
Avoid calling naming convention on scrape

shakuzen:
The naming convention was being called on each scrape for each metric to create the MetricMetadata. This is unnecessary because we already have the computed convention name specifically to avoid needing to call the convention again.

See #5229

The same test has been added to the micrometer-registry-prometheus module and the simpleclient module as well. The test is passing without changes to the simpleclient module, but the changes made in main code were needed for this to pass in micrometer-registry-prometheus, demonstrating the difference in behavior we want to ensure does not regress in the future.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
@@ -213,8 +213,8 @@ public class PrometheusMeterRegistry extends MeterRegistry {
                     (conventionName,
                             tagKeys) -> Stream.of(new MicrometerCollector.Family<>(conventionName,
                                     family -> new CounterSnapshot(family.metadata, family.dataPointSnapshots),
-                                    getMetadata(id), new CounterDataPointSnapshot(counter.count(),
-                                            Labels.of(tagKeys, tagValues), counter.exemplar(), 0))));
+                                    getMetadata(conventionName, id.getDescription()), new CounterDataPointSnapshot(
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -549,13 +558,8 @@ private void onMeterRemoved(Meter meter) {
...
 
-    private MetricMetadata getMetadata(Meter.Id id) {
-        return getMetadata(id, "");
-    }
-
-    private MetricMetadata getMetadata(Meter.Id id, String suffix) {
-        String name = config().namingConvention().name(id.getName(), id.getType(), id.getBaseUnit()) + suffix;
-        String help = prometheusConfig.descriptions() ? Optional.ofNullable(id.getDescription()).orElse(" ") : " ";
+    private MetricMetadata getMetadata(String name, @Nullable String description) {
+        String help = prometheusConfig.descriptions() ? Optional.ofNullable(description).orElse(" ") : " ";
```
jonatan-ivanov, line 562: This was a copy-paste form the 0.x client code but eliminating the `Optional` might also have some positive impact on performance.

jonatan-ivanov (approved):

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
shakuzen: Agreed and it stood out to me too when making these changes, but I wanted to keep the changes here focused to one issue. Let's take it up in a separate change.

PR #5292, opened 2024-07-09
Improve MeterTag to allow extracting multiple keys/value tags

### LINKED ISSUE #4081: Improve MeterTag to allow extracting multiple keys/value tags

smaxx:
Closes #4081


micrometer-commons/src/main/java/io/micrometer/common/util/StreamUtils.java
```diff
@@ -0,0 +1,31 @@
...
+package io.micrometer.common.util;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class StreamUtils {
+
+    public static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor) {
```
shakuzen, line 25: If this is only used in the one place, I'd rather not add it as public API. We can keep it private in AnnotationHandler, I think.

micrometer-core/src/test/java/io/micrometer/core/aop/TimedAspectTest.java
```diff
@@ -518,7 +562,8 @@ void meterTagOnSuperClass() {
 
         enum AnnotatedTestClass {
 
-            CLASS_WITHOUT_INTERFACE(MeterTagClass.class), CLASS_WITH_INTERFACE(MeterTagClassChild.class);
+            // CLASS_WITHOUT_INTERFACE(MeterTagClass.class),
```
shakuzen, line 565: This being commented out will result in tests for it not running. Actually it looks like all the tests under `MeterTagsTests` weren't being run because it wasn't annotated with `@Nested`.

smaxx: i've reenabled the enum value, must be an accidental commenting out((
All tests are executed as far as i can see

micrometer-commons/src/main/java/io/micrometer/common/util/StreamUtils.java
smaxx: honestly i would not agree. It's purely independent code which can and should be extracted into own utility class. I would even imagine that it can be later widely reused elsewhere.
Regarding leaving it in AnnotationHandler - technically it would also work fine, however that one has quite enough code dealing with annotations and  reflection, so messing it with stream API utility methods would make purpose of the class even less obvious.
In any case if you think that it is necessary then i will move the method to AnnotationHandler. Just let me know it's a final "go|no go" unblocker :)

micrometer-core/src/test/java/io/micrometer/core/aop/TimedAspectTest.java
shakuzen: > All tests are executed as far as i can see

It looks like it may just be an issue with running tests from my IDE (IntelliJ). I've confirmed when running the Gradle build, all the tests are being executed.

micrometer-commons/src/main/java/io/micrometer/common/util/StreamUtils.java
shakuzen: I think your approach would make sense in application code that is not shared like this library with many applications. We need to support public API we expose and we do not want to pollute users' classpaths with utility classes they don't expect from an observability library (we already unfortunately do this with yet another StringUtils on everyone's classpath) for something we only want to use internally.

> I would even imagine that it can be later widely reused elsewhere.

We have not had a need for it elsewhere so far, and if we eventually do have such a need, we can revisit the decision. It's easy to move private code and make it public, but it is not easy to remove or change public API that may be used by many people.

I'll take care of moving this with merging so we can get this enhancement in the soon to be released milestone.

shakuzen (approved):
Thank you for the pull request and quick action to review. We'll get this in the 1.14.0-M2 milestone.

micrometer-commons/src/main/java/io/micrometer/common/util/StreamUtils.java
smaxx: oh, i haven't looked at it from this perspective, thanks for the detailed explanation. `utility classes they don't expect from an observability library` - that is more than super-descriptive statement.

Issue #5239, opened 2024-07-01
Test duplicate stops on observations


bclozel:
While working on a Spring Framework instrumentation problem, I have noticed that my instrumentation could in some cases call `observation.stop()` twice. This is not an intended usage and is considered as a bug in the instrumentation.

In production, calling `stop()` multiple times on an observation will result in calling observation handlers multiple times. In case of metrics, this will create duplicate records for the same observation (it will be counted twice). I have tried to replicate this case and test against it in my test suite, but failed to do that. Right now "micrometer-observation-test" allows to check whether an observation has been closed. In case of multiple `stop()` calls, I couldn't find a way to detect this case with the `TestObservationRegistry`.

I understand that introducing additional state checks in observations would likely create a performance degradation - testing this case would help instrumentations.

jonatan-ivanov:
@bclozel What do you think about something like this: https://github.com/micrometer-metrics/micrometer/pull/5300?

PR #5300, opened 2024-07-11
Add ObservationValidator

jonatan-ivanov:
Closes gh-5239

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2024 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.observation.tck;
+
+import io.micrometer.common.lang.Nullable;
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
+import io.micrometer.observation.Observation.Context;
+import io.micrometer.observation.Observation.Event;
+import io.micrometer.observation.ObservationHandler;
+
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+/**
+ * An {@link ObservationHandler} that validates the order of events of an Observation (for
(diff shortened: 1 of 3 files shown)


shakuzen (approved):
Looks good. I think it'd be good to have some test cases of expected instrumentation patterns and assert that there is no invalid message - we want to make sure we don't introduce false positive validation issues.

jonatan-ivanov:
Added some tests to cover those cases too.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
```diff
@@ -0,0 +1,159 @@
...
+class ObservationValidator implements ObservationHandler<Context> {
+
+    private static final InternalLogger LOGGER = InternalLoggerFactory.getInstance(ObservationValidator.class);
+
+    private final Consumer<ValidationResult> consumer;
+
+    private final Predicate<Context> supportsContextPredicate;
+
+    ObservationValidator() {
+        this(validationResult -> LOGGER.warn(validationResult.toString()));
```
shakuzen, line 45: I approved and merged, but I wonder if it would be better to throw an exception by default (what gets used with TestObservationRegistry). I worry that a log message in tests will very easily be missed. The things we're checking for here should plainly be instrumentation bugs and so unless there's a bug in this validator, I'm not sure it should be a problem that an exception is thrown. Thoughts?

bclozel: I agree. The validator is installed by default, but I can't think of a case where duplicate stops calls is not considered as a bug in the instrumentation. If somehow this is expected, people should overrride the consumer and log things.

shakuzen: > If somehow this is expected, people should overrride the consumer and log things.

The issue with this is that the validator was intentionally made package private so it is all kept as an implementation detail. As things are, there's no option for users to override the consumer. We could make that public API so it is possible, but we'll have to make sure the API is right before going GA.

jonatan-ivanov: I think we have a few options to make this more flexible:
1. We can make the API public and add an overload to the test registry to inject handlers
1. We can make only the consumer part public (+ the class that is received by the consumer) and add an overload to the test registry to inject the consumer (I don't think there is a huge difference between this and the previous one) 
1. We can also add an overload to the test registry with a flag to not register the validator

shakuzen: I think my preference would be to switch it to throw an exception and not make anything configurable until we get feedback from a user why the validator's assumptions about correct instrumentation are wrong for their use case. Catching and ignoring the exception thrown in a test is always a workaround on a case-by-case basis. The challenge, as always, is getting enough feedback during the milestone phase before releasing GA, but it's also good for us to understand such use cases since I think we are lacking imagination to come up with them.

jonatan-ivanov: 👍🏼  In order to do that, we need to surface at least an exception so that users can handle it as needed: https://github.com/micrometer-metrics/micrometer/pull/5307

Issue #3112, opened 2022-04-01
Support EMF LoggingMeterRegistry


chenqi0805:
**Please describe the feature request.**
We need a metric log output file that uses [Embedded Metrics Format](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html), e.g.

```
{
  "_aws": {
    "Timestamp": 1574109732004,
    "CloudWatchMetrics": [
      {
        "Dimensions": [["..."]],
         ...
      }
    ] 
  },
  "ClientId": "012345678910",
  ...
}
```

**Rationale**
Please help us understand the rationale (why) behind the enhancement request.

**Additional context**
Add any other context about the feature request here, e.g. related issues, prior art.

jonatan-ivanov:
Micrometer supports CloudWatch, you don't need to log your metrics out in EMF if you just want to use CloudWatch. Would the CloudWatchMeterRegistry work for you?

chenqi0805:
> Micrometer supports CloudWatch, you don't need to log your metrics out in EMF if you just want to use CloudWatch. Would the CloudWatchMeterRegistry work for you?

@jonatan-ivanov Thanks for the quick feedback! In our use case, stream processing is needed including filtering, update, transformation, and aggregation on the metrics before publishing to CloudWatch while CloudWatchMeterRegistry directly uses the cloudwatch client to publish data. Follow up questions:

1. How much efforts (estimation of days) it will take to build such an EMF logging meter?
2. As an alternative, is there any logging meter registry that output metrics to logs in a particular format?

checketts:
@chenqi0805 I'm not aware of anyone actively working on an AWS EMF implementation. So there is no estimate available. Unless you were asking how hard it would be for you to contribute, the effort to extend an existing `SimpleMeterRegistry` probably wouldn't be too much.

As an alternative, instead of enhancing the existing logging registry, you could extend the `LoggingMeterRegistry` or `StepMeterRegistry` and hook into the `publish` method to write out any EMF specific data.

Sorry I can be more helpful.

chenqi0805:
@checketts Thanks for the response!

>As an alternative, instead of enhancing the existing logging registry, you could extend the LoggingMeterRegistry or StepMeterRegistry and hook into the publish method to write out any EMF specific data.

This seems to be the right strategy for our usecase. I did a little diving, looks like we just need to override this part

[...]

PR #5304, opened 2024-07-12
Implement a CloudWatch EmbeddedMetricFormat (EMF) registry

ksletmoe-aws:
I based this heavily off of the existing CloudWatch2 registry and tried to maintain attribution comments where it made sense.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -343,7 +343,7 @@ subprojects {
 
             check.dependsOn("testModules")
 
-            if (!(project.name in ['micrometer-test-aspectj-ltw', 'micrometer-test-aspectj-ctw'])) { // add projects here that do not exist in the previous minor so should be excluded from japicmp
+            if (!(project.name in ['micrometer-test-aspectj-ltw', 'micrometer-test-aspectj-ctw', 'micrometer-registry-cloudwatch-embedded-metrics'])) { // add projects here that do not exist in the previous minor so should be excluded from japicmp
                 apply plugin: 'me.champeau.gradle.japicmp'
                 apply plugin: 'de.undercouch.download'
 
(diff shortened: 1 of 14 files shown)


ksletmoe-aws:
There should already be a corporate CLA signed by my employer (Amazon).

jonatan-ivanov:
Thank you for the PR!

CLA:
Could you please check what happens if you click on the first link in the comment of the cla-bot? Do you get an error or does the page says that the CLA is signed?
If the latter, could you please click on sync (second link in the comment).

EMF:
Micrometer supports CloudWatch, you don't need to log your metrics out in EMF if you just want to use CloudWatch. Would the CloudWatchMeterRegistry work for you?

ksletmoe-aws:
CLA:
I clicked the sync -- it seems it's not finding our agreement. Is it possible for you to check if it's been signed for @amazon.com domains? Our legal team seems to think that it is.

EMF:
In addition to the concerns Qi Chen raised in [#3112](https://github.com/micrometer-metrics/micrometer/issues/3112), EMF is useful if you are already publishing logs and/or metrics with a CloudWatch Agent or Firelens sidecar in ECS, as there's less configuration changes needed. You also don't need to inject a null object CloudWatch client or modify your code to skip setting up a metric registry in a dev environment with EMF -- you can simply just not run an EMF publisher and metrics will not be sent anywhere. Lastly, EMF gives more flexibility for how to handle failed batches because you are offloading the actual publishing to another process, which can be configured to retry batches, keep metrics on disk, etc in the event CloudWatch is unreachable.

ksletmoe-aws:
I spoke with @jonatan-ivanov on Slack, and they were able to help me determine that the CLA has been signed by somebody in the amzn github organization. I joined that organization. Trying to sync the PR via the above link but still no dice on the CLA approval workflow. Suspect there might be some propagation delay

ksletmoe-aws:
Wanted to check in and see if this review has any feedback or will be accepted

ksletmoe-aws:
It was decided that this should be hosted in a different repository.

jonatan-ivanov:
Could you please add a link to that repo here so that the next person who might need it will find it?

jonathannaguin:
Does this repository exist? @ksletmoe-aws

PR #5307, opened 2024-07-16
Throw exception in case of invalid Observations with TestObservationRegistry

jonatan-ivanov:


micrometer-observation-test/src/test/java/io/micrometer/observation/tck/TestObservationRegistryAssertTests.java
```diff
@@ -64,7 +64,7 @@ void should_not_break_on_multiple_threads() {
 
     @Test
     void should_fail_when_observation_not_started() {
-        Observation.createNotStarted("foo", registry).stop();
+        Observation.createNotStarted("foo", registry);
```
jonatan-ivanov, line 67: I needed to "fix" this test: it simulates an invalid scenario but it does not need to call stop on an `Observation` that hasn't been started to verify the expected behavior.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
```diff
@@ -35,14 +33,12 @@
...
 
-    private static final InternalLogger LOGGER = InternalLoggerFactory.getInstance(ObservationValidator.class);
-
     private final Consumer<ValidationResult> consumer;
 
     private final Predicate<Context> supportsContextPredicate;
 
     ObservationValidator() {
-        this(validationResult -> LOGGER.warn(validationResult.toString()));
+        this(ObservationValidator::throwInvalidObservationException);
```
shakuzen, line 41: How hard would it be to update the tests to assert on the exception (or lack thereof) rather than using a custom consumer that users can't themselves use?

shakuzen (approved):
Looks good to me. I wonder if we should update our tests (ObservationValidatorTests) to not use a custom consumer now so we're testing the same way our users will be affected by the ObservationValidator.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
jonatan-ivanov: That's a great idea and I have no clue why I did not do that way originally. :)

PR #5309, opened 2024-07-19
Added constructor to MeterRegistry to enable percentile publish

nacho270:
Added a handy constructor that takes `publishPercentileHistogram` as parameter given the default value is `false`. This provides an easy way to enable a useful feature with little code.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -483,6 +483,29 @@ public abstract class MeterRegistry {
     }
 
     /**
+     * Measures the time taken for short tasks and the count of these tasks.
+     * @param name The base metric name
+     * @param tags Sequence of dimensions for breaking down the name.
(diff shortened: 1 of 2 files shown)


shakuzen:
Thanks for the pull request. We already have the Builder for all options including `publishPercentileHistogram`. I don't think it is worth increasing the API surface area to have this option outside of using the Builder. I get how this is shorter to write but I think it doesn't read as well (without IDE assistance) and it makes a special case for Timer compared to the pattern of the API for other meter types. Unless @jonatan-ivanov disagrees, I think we will say no to adding this at this time.

jonatan-ivanov:
I agree with Tommy. I also think that if we would do this we should introduce this to `DistributionSummary` too and I think it would also be a fair argument to add client-side percentiles and SLOs too so we would end-up with a lot of overloads which I think hurts discoverability.

Depending on the situation, using [`MeterProvider`](https://docs.micrometer.io/micrometer/reference/concepts/meter-provider.html) could help to write less code.

Let me close this issue, @nacho270 please let us know if you disagree.

PR #5316, opened 2024-07-28
Add default factoryType tag in CommonsObjectPool2Metrics

HYEONSEOK1:
## Overview

~~This PR addresses the issue of Prometheus tag collision by introducing distinct metric name prefixes for different pool types in the `CommonsObjectPool2Metrics` class. This change ensures that metrics from `GenericObjectPool` and `GenericKeyedObjectPool` have unique names, preventing conflicts caused by differing tag sets.~~

This PR addresses the issue of Prometheus tag collision by ensuring that all metrics have a consistent set of tags. For `GenericObjectPool`, a `factoryType` tag is added, and for other pool types, a default value of `factoryType=none` is used. This approach resolves the conflict caused by differing tag sets between `GenericObjectPool` and other pool types.

## Changes

~~- Introduced distinct metric name prefixes: `objectpool.` for `GenericObjectPool` and `keyedobjectpool.` for `GenericKeyedObjectPool`.~~
~~- Updated the `registerMetricsEventually` method to apply the appropriate prefix based on the pool type.~~
~~- Adjusted metric registration methods (`registerGaugeForObject`, `registerFunctionCounterForObject`, `registerTimeGaugeForObject`) to use the new prefixes.~~

- Added a default `factoryType=none` tag for all pool types except `GenericObjectPool`.

## Reasoning

Prometheus requires that all meters with the same name have the same set of tag keys. The previous implementation violated this rule by adding a `factoryType` tag for `GenericObjectPool`, which was not present in other pool types. This caused registration failures in `PrometheusMeterRegistry` at [this line](https://github.com/micrometer-metrics/micrometer/blob/main/implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L594). By ensuring that all metrics include a consistent set of tags, we eliminate tag collisions and comply with Prometheus' requirements.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/commonspool2/CommonsObjectPool2Metrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/commonspool2/CommonsObjectPool2Metrics.java
@@ -129,12 +129,16 @@ public class CommonsObjectPool2Metrics implements MeterBinder, AutoCloseable {
 
     private Iterable<Tag> nameTag(ObjectName name, String type)
             throws AttributeNotFoundException, MBeanException, ReflectionException, InstanceNotFoundException {
+
         Tags tags = Tags.of("name", name.getKeyProperty("name"), "type", type);
+        String factoryType = "none"; // Default value for factoryType if not available
+
         if (Objects.equals(type, "GenericObjectPool")) {
             // for GenericObjectPool, we want to include the name and factoryType as tags
-            String factoryType = mBeanServer.getAttribute(name, "FactoryType").toString();
-            tags = Tags.concat(tags, "factoryType", factoryType);
+            factoryType = mBeanServer.getAttribute(name, "FactoryType").toString();
         }
+
+        tags = Tags.concat(tags, "factoryType", factoryType);
         return tags;
     }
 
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
Since in Micrometer we consider changing existing meter names and tags a breaking change, I don't think we should do this outside of a new major version.
What we can do though is assigning a value to `factoryType` if it is not available, e.g.: `none`. What do you think about doing that instead?

jonatan-ivanov:
Also, for everyone who might bump into this issue: you can work this around by adding a `MeterFilter` that adds a `factoryType=none` tag if it does not exists for these metrics.

HYEONSEOK1:
Thank you for your feedback and suggestions!

I’ve reverted the previous changes and updated the code as suggested. The nameTag method now defaults to factoryType=none for all pool types except GenericObjectPool, which gets its specific factoryType tag.

logan-28:
> Thank you for your feedback and suggestions!
> 
> I’ve reverted the previous changes and updated the code as suggested. The nameTag method now defaults to factoryType=none for all pool types except GenericObjectPool, which gets its specific factoryType tag.
LGTM

logan-28:
> Thank you for your feedback and suggestions!
> 
> I’ve reverted the previous changes and updated the code as suggested. The nameTag method now defaults to factoryType=none for all pool types except GenericObjectPool, which gets its specific factoryType tag.

LGTM

jonatan-ivanov:
I polished your changes a bit and added tests so that we verify that all the tags are there all the time and rebased it on 1.9.x so that older versions of Micrometer will also get the fix.

PR #5365, opened 2024-08-02
Make more explicit that recording a metric is safe to call

cbrachem:
I've added a `TIP` block to the Registry's docs to make it more explicit what happens when operations on `Meter`s like `incrementCount` are called, which will make it clear that such operations are safe to call from regular code.

This PR is the result of a [discussion in Slack](https://micrometer-metrics.slack.com/archives/C662HUJC9/p1721900777113779).

Of course, feel free to challenge the general usefulness of the added paragraph, its accuracy and wording.


docs/modules/ROOT/pages/concepts/registry.adoc
```diff
@@ -3,6 +3,8 @@
 
 A `Meter` is the interface for collecting a set of measurements (which we individually call metrics) about your application. Meters in Micrometer are created from and held in a `MeterRegistry`. Each supported monitoring system has an implementation of `MeterRegistry`. How a registry is created varies for each implementation.
 
+TIP: Recoding a measurement for a `Meter` is expected to be a simple math/tallying operation that will not consume any significant amount of computing time or throw exceptions. Transferring metrics to an external monitoring system is a scheduled operation on a separate thread and will not interfere with meter recordings.
```
shakuzen, line 6: I would simplify the first part to something along the lines "recording a measurement for a `Meter` is expected to be a relatively cheap operation and it should not throw any exception."
The second sentence is about publishing which is a bit implementation dependent. Perhaps I would word it as something like "if the registry supports publishing metrics to a metrics backend, this should be done on a separate thread from recording metrics"

There's also a bit of a question of whether both these belong in this page (Registry) of the docs. Maybe we should move the Meters page to before the Registry page in the table of contents and have the first part about recording measurements there?

shakuzen (commented):
Thank you for the pull request. This is useful info to add to the docs for sure. I left some review.

docs/modules/ROOT/pages/concepts/registry.adoc
cbrachem: > I would simplify the first part to something along the lines "recording a measurement for a Meter is expected to be a relatively cheap operation and it should not throw any exception."
The second sentence is about publishing which is a bit implementation dependent. Perhaps I would word it as something like "if the registry supports publishing metrics to a metrics backend, this should be done on a separate thread from recording metrics"

Sounds good, I will rephrase according to your suggestion.

> There's also a bit of a question of whether both these belong in this page (Registry) of the docs. Maybe we should move the Meters page to before the Registry page in the table of contents and have the first part about recording measurements there?

Interesting. That might be nice. Let me try and move things around a bit and then we can see how it looks.

cbrachem: @shakuzen Is this what you had in mind?

docs/modules/ROOT/pages/concepts/meters.adoc
```diff
@@ -1,6 +1,11 @@
 [[meters]]
 = Meters
 
+A `Meter` is the interface for collecting a set of measurements (which we individually call metrics) about your application.
+
 Micrometer supports a set of `Meter` primitives, including `Timer`, `Counter`, `Gauge`, `DistributionSummary`, `LongTaskTimer`, `FunctionCounter`, `FunctionTimer`, and `TimeGauge`. Different meter types result in a different number of time series metrics. For example, while there is a single metric that represents a `Gauge`, a `Timer` measures both the count of timed events and the total time of all timed events.
 
+TIP: Recoding a measurement for a `Meter` is expected to be a relatively cheap operation and should not throw any exception.
+If the registry supports publishing metrics to a monitoring system, this is done in a separate thread snd should not affect recording metrics.
```
shakuzen, line 9: It would probably be good to link to the registry page when mentioning "registry" here since we haven't introduced the concept yet now that this page comes before the registry one in the contents.

docs/modules/ROOT/pages/concepts/registry.adoc
shakuzen: Yes, that looks good to me. I left a minor review comment.

docs/modules/ROOT/pages/concepts/meters.adoc
cbrachem: Good point. I've added a link.

shakuzen (approved):
Thank you for the documentation improvement.

shakuzen:
I've rebased the pull request on the `1.12.x` branch so we can add apply this change which is applicable to the oldest version of the docs maintained here and merge forward.

Issue #5366, opened 2024-08-02
Metrics not collected after ExecutorService recreation


ajax-surovskyi-y:
**Describe the bug**
When using Micrometer to collect metrics from an `ExecutorService`, I wrap it with `ExecutorServiceMetrics.monitor`. In the framework logic that I use, the original ExecutorService shuts down if it is not used for some time, and then it is recreated as needed. After it is recreated, metrics are not collected from it because it is created with the exact same name, and all methods like `ThreadPoolExecutor::getCorePoolSize` are called on the `ThreadPoolExecutor` that has already been shut down. I have created a [PR](https://github.com/micrometer-metrics/micrometer/pull/5367) with a test for this problem and a potential fix.
**Environment**

 - Micrometer version 1.13.1
 - Micrometer registry statsd
 - OS: Linux
 - Java version: 21.0.1

**To Reproduce**
How to reproduce the bug:

1. Wrap an `ExecutorService` with `ExecutorServiceMetrics.monitor.`
2. Allow the shutdown.
3. Recreate the ExecutorService.
4. Observe that metrics are not being collected.

**Expected behavior**
Metrics should continue to be collected from the recreated ExecutorService.

**Additional context**

slovdahl:
Nice that this was fixed 👍

I think the `TimedScheduledExecutorService` variant still has the same problem? I could maybe look into it following a similar pattern if no one else feels like it.

jonatan-ivanov:
@slovdahl I think it has, please feel free to open a PR to fix it.

slovdahl:
https://github.com/micrometer-metrics/micrometer/pull/6920

PR #5367, opened 2024-08-02
Metrics not collected after ExecutorService recreation

ajax-surovskyi-y:
https://github.com/micrometer-metrics/micrometer/issues/5366

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
@@ -29,9 +29,12 @@ import io.micrometer.core.instrument.internal.TimedExecutorService;
 import io.micrometer.core.instrument.internal.TimedScheduledExecutorService;
 
 import java.lang.reflect.Field;
+import java.util.List;
+import java.util.Set;
 import java.util.concurrent.*;
 
 import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toSet;
 
 /**
  * Monitors the status of executor service pools. Does not record timings on operations
@@ -61,6 +64,7 @@ public class ExecutorServiceMetrics implements MeterBinder {
 
     private static final String DESCRIPTION_POOL_SIZE = "The current number of threads in the pool";
 
+    private final Set<Meter.Id> registeredMeterIds = ConcurrentHashMap.newKeySet();
     @Nullable
     private final ExecutorService executorService;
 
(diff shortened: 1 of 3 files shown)


shakuzen:
Thank you for the pull request. Could you take a look at the test failures?

ajax-surovskyi-y:
@shakuzen kind reminder. Could you take a look this PR?

shakuzen (commented):
`ExecutorServiceMetrics` and `TimedExecutorService` can be used together on the same `ExecutorService` (e.g. through one of the `monitor` method on `ExecutorServiceMetrics`) or separately through the available constructor(s) on each. It looks like right now the metrics will only be deregistered if the `ExecutorService` is wrapped with `monitor`. If `ExecutorServiceMetrics` are bound to an `ExecutorService` without wrapping and it is shutdown, there is no way to unregister the metrics still. Maybe we can leave this as existing behavior, but it did stand out to me as seeming a bit inconsistent. We could make `ExecutorServiceMetrics` `Closeable` and handle unregistering its metrics through that, but it would be up to users to call that close method when shutting down the executor. Thoughts?

ajax-surovskyi-y:
You're right about slight inconsistency. What threw me off was the comment above the class: "An {@link java.util.concurrent.ExecutorService} that is timed. This class is for internal use." But since it's public, users might have already relied on it. I would prefer not to require users to close ExecutorServiceMetrics, especially since it wouldn't work in my case where I pass the ExecutorService to the Akka framework and it handles the shutdown itself.

shakuzen:
Thanks for reminding me that `TimedExecutorService` is intended to be an internal class. Given that, the current PR makes more sense. It's still a little inconsistent because `ExecutorServiceMetrics` is not internal and it is valid to instrument an executor with the metrics it provides without wanting to time executions. I think for now we can wait until we get feedback from users that they want to unregister metrics in that case also before handling it.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -61,6 +64,8 @@ public class ExecutorServiceMetrics implements MeterBinder {
 
     private static final String DESCRIPTION_POOL_SIZE = "The current number of threads in the pool";
 
+    private final Set<Meter.Id> registeredMeterIds = ConcurrentHashMap.newKeySet();
```
shakuzen, line 67: I don't think we need to use the Concurrent variant here but it also shouldn't matter much if we do.

shakuzen (approved):

PR #5370, opened 2024-08-08
Add history-tracking to ObservationValidator

jonatan-ivanov:
This change lets you peek into the history of certain interactions with your `Observation`s if an `InvalidObservationException` is thrown. This includes which method was called on your `Observation`s (`start`, `stop`, `error`, etc) and also the relevant part of the stack trace. This is helpful when the instrumentation is complex and it is not apparent where a previous call to Observation happened that needs to be fixed.

`InvalidObservationException` can provide you the full history through its `getHistory` method and it also gives you a summary through its `toString`, something like this:
```
io.micrometer.observation.tck.InvalidObservationException: Invalid error signal: Observation has already been stopped
START: app//io.micrometer.observation.tck.ObservationValidatorTests.errorAfterStopShouldBeInvalid(ObservationValidatorTests.java:98)
STOP: app//io.micrometer.observation.tck.ObservationValidatorTests.errorAfterStopShouldBeInvalid(ObservationValidatorTests.java:99)
ERROR: app//io.micrometer.observation.tck.ObservationValidatorTests.errorAfterStopShouldBeInvalid(ObservationValidatorTests.java:100)
```

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/InvalidObservationException.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/InvalidObservationException.java
@@ -18,6 +18,10 @@ package io.micrometer.observation.tck;
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.Observation.Context;
 
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
 /**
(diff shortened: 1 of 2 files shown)


shakuzen (approved):
Looks great. This will be very helpful in more quickly tracking down instrumentation issues, especially if the IDE makes the class name + line number a link to the exact line of code where the Observation calls were made.

PR #5391, opened 2024-08-15
change method signature of `MultiGauge.register`

### LINKED ISSUE #5390: MultiGauge.register should accept more types
**Describe the bug**
The typical way to use a `MultiGauge` is converting some list to a list of `Row`s using a Stream. This is also done in the`MultiGaugeSample` example. Java's type inference is tricky here. In the example, it works because Java knows what is expected by `MultiGauge.register` and infers `List<Row<?>>`. But if you separate the conversion and assign the list to a variable, the inferred type is `List<Row<Number>>`. The same is true for some reason if you use the newer `Stream.toList()` instead of `Stream.collect(Collectors.toList())`.

So
```java
            temperatures.register(
                fetchTemperatures().stream()
                                   .map(record -> Row.of(Tags.of("room", record.getRoom().name()), record.getTemperature()))
                                   .collect(Collectors.toList()));
```
will compile but
```java
            temperatures.register(
                fetchTemperatures().stream()
                                   .map(record -> Row.of(Tags.of("room", record.getRoom().name()), record.getTemperature()))
                                   .toList());
```
and
```java
            var rows = fetchTemperatures().stream()
                    .map(record -> Row.of(Tags.of("room", record.getRoom().name()), record.getTemperature()))
            temperatures.register(rows);
```
will not.

I wonder if the type parameter of `Row` adds any value at all. But removing it would probably break existing code, so I suggest to change the signature of `MultiGauge.register` to accept an `Iterable<? extends Row<?>>` instead. Then it could also be called with a `List<Row<Number>>`.

I will create a PR for this small change.

**Environment**

 - Micrometer version [e.g. 1.7.1]: 1.13.2, also on the main branch at the time of writing
 - Micrometer registry [e.g. prometheus]: prometheus, but not relevant
 - OS: [e.g. macOS] macOS, but not relevant
  ```
openjdk version "21.0.2" 2024-01-16
OpenJDK Runtime Environment GraalVM CE 21.0.2+13.1 (build 21.0.2+13-jvmci-23.1-b30)
OpenJDK 64-Bit Server VM GraalVM CE 21.0.2+13.1 (build 21.0.2+13-jvmci-23.1-b30, mixed mode, sharing)
```

**To Reproduce**
See description

**Expected behavior**
`List<Row<Number>>` is accepted and no ugly casts are necessary.

lbilger:
change method signature of `MultiGauge.register` to also allow e.g. List<Row<Number>> as a parameter

fixes #5390

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java
@@ -54,18 +54,18 @@ public class MultiGauge {
         return new Builder(name);
     }
 
-    public void register(Iterable<Row<?>> rows) {
+    public void register(Iterable<? extends Row<?>> rows) {
         register(rows, false);
     }
 
     @SuppressWarnings("unchecked")
-    public void register(Iterable<Row<?>> rows, boolean overwrite) {
+    public void register(Iterable<? extends Row<?>> rows, boolean overwrite) {
         registeredRows.getAndUpdate(oldRows -> {
             // for some reason the compiler needs type assistance by creating this
             // intermediate variable.
             Stream<Meter.Id> idStream = StreamSupport.stream(rows.spliterator(), false).map(row -> {
                 Row r = row;
-                Meter.Id rowId = commonId.withTags(row.uniqueTags);
+                Meter.Id rowId = commonId.withTags(r.uniqueTags);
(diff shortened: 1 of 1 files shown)


shakuzen (approved):
Thanks for the pull request. This sounds reasonable to me. The only concern is backward compatibility, but this looks like it should be backward compatible.

PR #5394, opened 2024-08-16
Include minimumExpectedValue in exponential histogram

lenin-jaganathan:


shakuzen (approved):
Thanks for the pull request. I was trying to think of another way we can do this. If it's just about where to record a value (positive buckets or zero count), we can just keep track of the `minimumExpectedValue` and change the conditional to be `<` rather than `<=`. But since we need to include the `zero_threshold` in the snapshot, this may be the best way to do it.

Issue #5066, opened 2024-05-11
Log a warning when instrumenting a cache that is not recording stats


keith-turner:
**Please describe the feature request.**
It would be nice if a warning were logged (or the user was informed in some way) when a cache is instrumented that is not recording stats. 

**Rationale**
For some Cache implementations like Caffeine recording stats is optional.  When instrumenting a cache that is not recording stats the meters will constantly report 0 which could be misleading.  This is probably not what the user desires.

**Additional context**

apache/accumulo#4552

shakuzen:
That sounds like a good improvement to me. Would you or anyone else be interested in making a pull request for it?

izeye:
I created https://github.com/micrometer-metrics/micrometer/pull/5402 to try to resolve this for the `CaffeineCacheMetrics`.

shakuzen:
With #7568 now merged, I think every cache metrics implementation we maintain is covered. Closing this as complete.

PR #5402, opened 2024-08-25
Log a warning when instrumenting a cache that is not recording stats in CaffeineCacheMetrics

izeye:
This PR changes to log a warning when instrumenting a cache that is not recording statistics in the `CaffeineCacheMetrics`.

See gh-5066

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
@@ -23,6 +23,8 @@ import com.github.benmanes.caffeine.cache.stats.CacheStats;
 import io.micrometer.common.lang.NonNullApi;
 import io.micrometer.common.lang.NonNullFields;
 import io.micrometer.common.lang.Nullable;
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
 import io.micrometer.core.instrument.*;
 
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
```diff
@@ -58,6 +62,10 @@ public class CaffeineCacheMetrics<K, V, C extends Cache<K, V>> extends CacheMete
      */
     public CaffeineCacheMetrics(C cache, String cacheName, Iterable<Tag> tags) {
         super(cache, cacheName, tags);
+
+        if (!cache.policy().isRecordingStats()) {
+            log.warn("The cache '{}' is not recording statistics.", cacheName);
```
shakuzen, line 67: I think we should be more explicit about the consequences of this. Does it mean metrics will never be recorded unless `Caffeine#recordStats` is called and metrics are re-bound?

izeye: @shakuzen Their values will be zero without it being called. I tried to refine the warning in d43ddb3.

izeye:
The Guava cache also has the same concept, but it doesn't seem to have an API corresponding to `Policy#isRecordingStats()` from the Caffeine cache.

izeye:
In addition to logging the warning, we can try not to register meters that depend on the `Caffeine.recordStats()` call at all instead of registering zero-valued meters.

shakuzen:
> In addition to logging the warning, we can try not to register meters that depend on the `Caffeine.recordStats()` call at all instead of registering zero-valued meters.

That sounds like a good idea to me. It will be more clear we don't have metrics for those rather than saying the metrics for those are 0. Do you want to update this pull request with that or make a separate PR for it?

izeye:

Looking into its implementation details, it seems a bit tricky, so it would be better to be discussed in a separate issue. I can try another PR if it's proven to be okay in the issue.

izeye:
> The Guava cache also has the same concept, but it doesn't seem to have an API corresponding to `Policy.isRecordingStats()` from the Caffeine cache.

I created https://github.com/google/guava/issues/7381 to see if it could be improved.

PR #5407, opened 2024-08-29
Include the observation name in validator message

shakuzen:
This will make it easier to read the exception message and match it up to the corresponding code.


jonatan-ivanov (approved):
Great idea!

Issue #5408, opened 2024-08-29
Skip registering Caffeine meters when statistics are not enabled


izeye:
See https://github.com/micrometer-metrics/micrometer/pull/5402#issuecomment-2315806111

izeye:
I created https://github.com/micrometer-metrics/micrometer/pull/5409 to try to resolve this.

PR #5409, opened 2024-08-29
Skip registering Caffeine meters when statistics are not enabled

izeye:
This PR changes to skip registering Caffeine meters when statistics are not enabled.

This PR also fixes and improves `GuavaCacheMetricsTest` and `CaffeineCacheMetricsTest`.

Closes gh-5408


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
```diff
@@ -40,6 +40,8 @@ public abstract class CacheMeterBinder<C> implements MeterBinder {
 
     private static final String DESCRIPTION_CACHE_GETS = "The number of times cache lookup methods have returned a cached (hit) or uncached (newly loaded or null) value (miss).";
 
+    static final long UNSUPPORTED = -1L;
```
izeye, line 43: I'm not happy with this magic number to express an unsupported meter for the `long` primitive type but couldn't find a better way without breaking changes.

shakuzen (approved):

PR #5468, opened 2024-09-07
Use Prometheus BOMs

izeye:
This PR changes to use the Prometheus BOMs.

This PR also polishes dependencies configuration a bit along the way.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -31,11 +31,6 @@ def VERSIONS = [
         libs.grpcTestingProto,
         libs.grpcKotlinStub,
         libs.gmetric4j,
-        libs.prometheusMetrics,
-        libs.prometheusMetricsExpositionFormats,
-        libs.prometheusMetricsTracerCommon,
-        libs.prometheusSimpleClient,
-        libs.prometheusSimpleClientPushgateway,
         libs.restAssured,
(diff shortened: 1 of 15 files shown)


dependencies.gradle
```diff
@@ -87,7 +80,9 @@ def VERSIONS = [
 def PLATFORM_BOMS = [
         libs.reactorBom,
         libs.nettyBom,
-        libs.junitBom
+        libs.junitBom,
+        libs.prometheusMetricsBom,
+        libs.prometheusSimpleClientBom
```
shakuzen, line 85: This will end up applying the Prometheus BOMs to all modules, but the Prometheus BOMs should only be needed in the Prometheus modules. Historically we did this kind of global management of dependencies, but I think for new changes, it will be better to keep it more focused where it makes sense and ultimately use this global `dependencies.gradle` less and less. So, I would remove the Prometheus BOMs from here and apply the right one in the corresponding Prometheus module instead.

shakuzen (changes requested):

dependencies.gradle
izeye: @shakuzen Thanks for the feedback! I updated in 3bea2a1 as suggested.

shakuzen (approved):

PR #5476, opened 2024-09-16
Fixed @Timed for methods raising Error

MarinaMoiseenko:
When a method is annotated with `@Timed` and failed its execution with any Exception exception tag is populated with class name of exception.
BUT
If the same method is failed with any error (StackOverflowError for example) metrics are counted as if method was executed without issues.
At the same time `@Counted` annotation and http_server_requests_seconds metrics are correctly detect error.

Below are examples of /actuator/metrics produced for `@Timed` method in Spring Boot Application
Notice that 

- for `@Timed("timed_with_error")` metric exception="none" and 
- for http_server_requests_seconds metric exception="StackOverflowError"

```
# HELP timed_with_error_seconds  
# TYPE timed_with_error_seconds summary
timed_with_error_seconds_count{class="com.twj.provided.observations.TimedController",exception="none",method="timedMethod"} 1
timed_with_error_seconds_sum{class="com.twj.provided.observations.TimedController",exception="none",method="timedMethod"} 0.0089739
# HELP timed_with_error_seconds_max  
# TYPE timed_with_error_seconds_max gauge
timed_with_error_seconds_max{class="com.twj.provided.observations.TimedController",exception="none",method="timedMethod"} 0.0089739
# HELP http_server_requests_seconds  
# TYPE http_server_requests_seconds summary
http_server_requests_seconds_count{error="StackOverflowError",exception="StackOverflowError",method="GET",outcome="SERVER_ERROR",status="500",uri="/timed"} 1
http_server_requests_seconds_sum{error="StackOverflowError",exception="StackOverflowError",method="GET",outcome="SERVER_ERROR",status="500",uri="/timed"} 0.0532252
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -221,9 +221,9 @@ public class TimedAspect {
                 return ((CompletionStage<?>) pjp.proceed()).whenComplete(
                         (result, throwable) -> record(pjp, timed, metricName, sample, getExceptionTag(throwable)));
             }
-            catch (Exception ex) {
-                record(pjp, timed, metricName, sample, ex.getClass().getSimpleName());
-                throw ex;
+            catch (Throwable e) {
+                record(pjp, timed, metricName, sample, e.getClass().getSimpleName());
+                throw e;
             }
         }
 
@@ -231,9 +231,9 @@ public class TimedAspect {
         try {
             return pjp.proceed();
         }
-        catch (Exception ex) {
-            exceptionClass = ex.getClass().getSimpleName();
-            throw ex;
+        catch (Throwable e) {
+            exceptionClass = e.getClass().getSimpleName();
+            throw e;
         }
         finally {
             record(pjp, timed, metricName, sample, exceptionClass);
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
Thank you for the PR!

As your fix suggests, this is because `TimedAspect` only catches `Exception` and not `Throwable`:
https://github.com/micrometer-metrics/micrometer/blob/b410174db680401423659955581bd03009113908/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java#L224-L227
While `CountedAspect` catches `Throwable`:
https://github.com/micrometer-metrics/micrometer/blob/b410174db680401423659955581bd03009113908/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java#L226-L229

I think they should be consistent and we should probably catch `Throwable` since the app/JVM can recover from `Error`s in some cases.

My only concern is if we should handle this as a bug and apply it on `1.12.x` instead of 1.14.x (`main`).

MarinaMoiseenko:
Hi Jonatan,

Yes, good point. I've created [PR 5479](https://github.com/micrometer-metrics/micrometer/pull/5479) for 1.12.x branch.
You may notice test method is simplified. It appeared current version is not compiled with Java 8 as CompletableFuture.failedFuture() is only introduced in Java 9.

Let me know if you want me to close this PR.

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/5479

PR #5483, opened 2024-09-19
Add Javadoc since for GrpcObservationDocumentation.GrpcServerEvents.CANCELLED

izeye:
This PR adds a Javadoc `@since` tag for the `GrpcObservationDocumentation.GrpcServerEvents.CANCELLED`.


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/GrpcObservationDocumentation.java
```diff
@@ -122,6 +122,10 @@ public String getName() {
             }
 
         },
+        /**
+         * For a canceled event.
+         * @since 1.14.0
```
shakuzen, line 127: I wonder how docs-generator will (or won't) handle `@since` tags in JavaDocs. Probably something to consider separately from this and test in docs-generator if it isn't already tested.

PR #5491, opened 2024-09-26
Pass observation to tasks

michaldo:
As developer, I would like parallelize my code and observe subtasks (https://stackoverflow.com/questions/78746378/spring-boot-3-micrometer-tracing-in-mdc/78765658#78765658 https://stackoverflow.com/questions/78889603/traceid-propagation-to-virtual-thread).

I found solution with `ContextExecutorService`, but according to discussion https://github.com/micrometer-metrics/context-propagation/issues/295, ContextExecutorService is low level API, which should not be used by application code.

`Observation` is enough high level to be used by application code.

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -20,11 +20,14 @@ import io.micrometer.common.KeyValues;
 import io.micrometer.common.lang.NonNull;
 import io.micrometer.common.lang.Nullable;
 import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
+import io.micrometer.context.ContextExecutorService;
+import io.micrometer.context.ContextSnapshotFactory;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


chemicL:
As explained in https://github.com/micrometer-metrics/context-propagation/issues/295#issuecomment-2388988545 `Observation` is a specific user of the context-propagation library and should not delegate to its generic capabilities for the purpose of propagating a specific state. I believe this PR should be discarded.

shakuzen:
At the risk of saying the same thing @chemicL already said but in a different way: `Observation` does not have a hard dependency on context-propagation, so adding a method to it that uses context-propagation is a non-starter, I think. In the micrometer-observation module, it is only the `ObservationThreadLocalAccessor` class that depends on context-propagation.

Maybe we could add something to our Micrometer Observation documentation to talk about using the context-propagation library as a solution for propagating Observation across thread switches, such as when using an ExecutorService.

jonatan-ivanov:
Tests here might help, particularly this one: https://github.com/micrometer-metrics/micrometer/blob/4a29322e598537187a777ce83841dc690629671a/micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java#L57-L70

You should be able to open a scope for your parent Observation inside of the task and then create a child observation (and a child scope). The limiting thing is that the parent observation should be "visible" from the code of the task but in your use-case this might not be an issue.

Something like this:
```java
ExecutorService executor = Executors.newSingleThreadExecutor(); 
  
Observation parent = Observation.start("parent", registry); 
executor.submit(() -> {
    try (Observation.Scope ignored = parent.openScope()) { 
        // create a child observation here and do what you want
    }          
});
// collect futures and wait for every future to finish their job
parent.stop(); 
```
This should work with parallel streams too.

michaldo:
@jonatan-ivanov, let me recap

1. I meet problem with passing observation to tasks and I applied similar solution as you proposed. Let's name in "_manual solution_"
2. Then igor.zh pointed me altermative solution, based on `ContextExecutorService`. Let's name it _"low level solution_"
3.  As _low_level solution_ requires complex setup, not adequtate to simple expected goal, I provided this PR, _high level solution_

So question is simple: what is correct solution to solve my problem: passing observation to tasks.

_Manual solution_ for typical problem looks as reinventing wheel
_Low_level solution_ requires deep knowledge how library works internally
_High level solution_ is probably too rough

PR #5551, opened 2024-10-05
Expose TestObservationRegistry as an AssertJ AssertProvider

filiphr:
This is just an idea which is inspired by the way the Spring Boot [`ApplicationContextAssertProvider`](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/assertj/ApplicationContextAssertProvider.java) works. Some more information from the Spring Boot [Testing your Auto-configuration](https://docs.spring.io/spring-boot/reference/features/developing-auto-configuration.html#features.developing-auto-configuration.testing).

The main idea is basically for people not to need to do `TestObservationRegistryAssert.assertThat`, but instead use (the most likely already exposed) `Assertions.assertThat`.

I tried to look into the issues and PRs if something like this was proposed before, but I could not find anything and I thought to share an idea as a PR. I fully understand that perhaps the idea of the `TestObservationRegistry` is not to depend on AssertJ and / or you do not like the idea at all. In any case if you think that this is not useful for the project then please go ahead and reject the PR.

### DIFF
--- a/docs/src/test/java/io/micrometer/docs/observation/ObservationHandlerTests.java
+++ b/docs/src/test/java/io/micrometer/docs/observation/ObservationHandlerTests.java
@@ -27,13 +27,13 @@ import io.micrometer.observation.annotation.Observed;
 import io.micrometer.observation.aop.ObservedAspect;
 import io.micrometer.observation.docs.ObservationDocumentation;
 import io.micrometer.observation.tck.TestObservationRegistry;
-import io.micrometer.observation.tck.TestObservationRegistryAssert;
 import org.junit.jupiter.api.Test;
 import org.springframework.aop.aspectj.annotation.AspectJProxyFactory;
 import org.springframework.lang.Nullable;
 
 import static io.micrometer.docs.observation.ObservationHandlerTests.TaxObservationDocumentation.TaxHighCardinalityKeyNames.USER_ID;
 import static io.micrometer.docs.observation.ObservationHandlerTests.TaxObservationDocumentation.TaxLowCardinalityKeyNames.TAX_TYPE;
+import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * Sources for observation-components.adoc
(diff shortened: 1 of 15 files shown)


shakuzen (approved):
Thanks for the pull request. It sounds like a nice change to me. We already declare an API dependency on AssertJ in the micrometer-observation-test module (see [here](https://github.com/micrometer-metrics/micrometer/blob/2df569e478ded74aac00a01cd8c09779f1138f3a/micrometer-observation-test/build.gradle#L6)), so I don't think it would be a big deal for the TestObservationRegistry to implement API from AssertJ. Let's see what the rest of the team thinks.

shakuzen:
I'm not sure why CircleCI did not run. Could you try updating the branch to trigger the CI to run again?

filiphr:
Thanks @shakuzen for looking into this. I rebased this on top of main, but it seems like CircleCI didn't run again, it's the same problem as in my other PR.

Is there a reason why you are not using GitHub Actions? Would you be interested in a PR which adds builds for them?

shakuzen:
> Is there a reason why you are not using GitHub Actions?

The short answer is: it's what has historically been used, it's free, and we have very few problems with it. If it ain't broke, don't fix it, as they say. That said, this issue happens rarely when users have setup a CircleCI account before and so it is trying to build the PR on their CircleCI account as opposed to the Micrometer org account, but their account auth is no longer valid for CircleCI. I'm not certain that's what is happening here, but it's a possibility.

filiphr:
You were right, I did have a CircleCI account setup from long time ago for some other Open Source projects I am working on. I've logged in there and now it looks like CircleCI is running. I'll adjust the other PR as well.

>If it ain't broke, don't fix it, as they say.

100% agree with you. Was mostly curious if there was not some other reason in particular.

marcingrzejszczak:
I think this is a great idea!

Issue #5136, opened 2024-05-22
Instrumented Java 11 HttpClient does not re-throw exceptions in sendAsync call


alex-inv:
**Describe the bug**
After instrumenting our `java.net.HttpClient` with newly released `MicrometerHttpClient` decorator, we've noticed our tests started failing. It seems that if any exception (e.g. network error) occurs, it will be caught in `MicrometerHttpClient` and not propagated further to the next `CompletionStage` available to the user, contrary to what happens in the original `HttpClient` implementation.

[Implementation](https://github.com/micrometer-metrics/micrometer/blob/37883fa6fb4a6d3f83d01f6b53101cc9f52b3f78/micrometer-java11/src/main/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClient.java#L274) of `sendAsync` decorator in `MicrometerHttpClient` does not re-throw exception in the `CompletableFuture.handle()` method, but returns only `response`. In case `handle` got an exception as its input, `null` result will be returned instead:
```
return client.sendAsync(request, bodyHandler, pushPromiseHandler).handle((response, throwable) -> {
            if (throwable != null) {
                instrumentation.setThrowable(throwable);
            }
            instrumentation.setResponse(response);
            stopObservationOrTimer(instrumentation, request, response);
            return response;
        });
```

**Environment**
 - Micrometer version: 1.13.0
 - OS: macOS Sonoma 14.4.1
 - Java version: Corretto-17.0.6.10.1

**To Reproduce**
The easiest way to reproduce the bug is to write a test simulating a faulty HTTP server, for example using `WireMock` library. I've [attached](https://github.com/micrometer-metrics/micrometer/files/15402296/micrometer-java-11-client-issue-test.zip) a minimal demo application (built with JDK 17 and Gradle) with a failing test.

Here I show two tests, first passing and second failing, highlighting the reported issue:
```
@Test
public void testError() {
        WireMock.stubFor(WireMock.get(WireMock.urlEqualTo("/test-url"))
                .willReturn(new ResponseDefinitionBuilder().withFault(Fault.CONNECTION_RESET_BY_PEER)));

        var client = HttpClient.newBuilder().build();

        var request = HttpRequest.newBuilder(URI.create(wireMockServer.baseUrl() + "/test-url"))
                .GET().build();

        var response = client.sendAsync(request, HttpResponse.BodyHandlers.ofString());

        assertThrows(CompletionException.class, response::join);
}

@Test
[...]

PR #5553, opened 2024-10-08
Consistent tagging and propagate sendAsync throwable

shakuzen:
When handling the sendAsync call in our instrumentation, we were not propagating the throwable observed to the CompletionFuture returned. This fixes that and adds a test. To keep tagging consistent in the case where response is null or not, this also updates tagging to always include `status` and `outcome`. In case the response is null, they will have value `UNKNOWN` for now. This is a change from before for this case, but the behavior before is problematic for the Prometheus registry that expects a consistent set of tags. The change is adding tags that weren't there before in the case response is null, so it should generally not break users.

Resolves gh-5136

### DIFF
--- a/micrometer-java11/src/main/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClient.java
+++ b/micrometer-java11/src/main/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClient.java
@@ -17,9 +17,7 @@ package io.micrometer.java11.instrument.binder.jdk;
 
 import io.micrometer.common.lang.Nullable;
 import io.micrometer.core.instrument.MeterRegistry;
-import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
-import io.micrometer.core.instrument.binder.http.Outcome;
 import io.micrometer.core.instrument.observation.ObservationOrTimerCompatibleInstrumentation;
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationRegistry;
@@ -36,6 +34,7 @@ import java.net.http.HttpResponse;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.Executor;
 import java.util.function.Function;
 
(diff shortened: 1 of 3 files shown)


micrometer-java11/src/test/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClientTests.java
```diff
@@ -91,6 +99,25 @@ void shouldInstrumentHttpClientWithTimer(WireMockRuntimeInfo wmInfo) throws IOEx
...
+        var client = MicrometerHttpClient.instrumentationBuilder(httpClient, meterRegistry).build();
+
+        var request = HttpRequest.newBuilder(URI.create(wmInfo.getHttpBaseUrl() + "/test-fault")).GET().build();
+
+        var response = client.sendAsync(request, HttpResponse.BodyHandlers.ofString());
+
+        assertThatThrownBy(response::join).isInstanceOf(CompletionException.class);
+        assertThatNoException().isThrownBy(() -> meterRegistry.get("http.client.requests")
+            .tag("method", "GET")
+            // TODO fix status/uri in exceptional cases where response may be null
```
shakuzen, line 114: This is not a TODO we should tackle in this PR. This should probably be fixed on `main` instead since it results in a change in behavior. Ideally, we should add more to our `HttpClientTimingInstrumentationVerificationTests` to ensure consistent behavior across instrumentations. The issue, if I remember correctly, is that some HTTP clients cannot instrument the case where the response is null.

micrometer-java11/src/main/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClient.java
```diff
@@ -277,6 +270,9 @@ public <T> CompletableFuture<HttpResponse<T>> sendAsync(HttpRequest httpRequest,
             }
             instrumentation.setResponse(response);
             stopObservationOrTimer(instrumentation, request, response);
+            if (throwable != null) {
+                throw new CompletionException(throwable);
```
shakuzen, line 274: This is the important change for the original issue.

jonatan-ivanov: Is this the proper way to do this or does `CompletableFuture` let us to signal an error on it instead?
(Though, since this is a "handle" method, I would assume it's ok to throw an exception from it.)

jonatan-ivanov (approved):

micrometer-java11/src/main/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClient.java
shakuzen: I'm not sure there's another way to do it. This results in the same outcome as without the handler, I believe. Exceptions thrown will always be wrapped in `CompletionException` as far as I know.

PR #5582, opened 2024-10-10
Move test code depending on Spring to separate module

shakuzen:
It is difficult to depend on Spring 6 in test code in our published modules because we maintain a Java 8 baseline while Spring 6 requires Java 17. Moving this code to a separate module avoids that issue. There is still one usage of Spring in our normal test code for ModifiedClassPathClassLoader.

This will unblock https://github.com/micrometer-metrics/micrometer/pull/5432 because the same tests do not fail on the latest Spring 6.1.


micrometer-core/build.gradle
```diff
@@ -143,13 +143,15 @@ dependencies {
...
     testImplementation 'javax.inject:javax.inject'
-    testImplementation 'org.springframework:spring-context'
     testImplementation 'com.google.inject:guice'
 
     // Uncomment these if you are interested in testing injection with dagger in MeterRegistryInjectionTest
     // testImplementation 'com.google.dagger:dagger'
     // testAnnotationProcessor 'com.google.dagger:dagger-compiler'
 
+    // Only needed for ModifiedClassPathClassLoader
+    testImplementation(libs.spring5.core)
```
shakuzen, line 153: I'd like to get rid of this test dependency as well, but that's a more involved change than I want to make now to solve the issue of test failures on bumping the Spring version.

micrometer-core/src/test/java/io/micrometer/core/aop/MeterTagSupportTests.java
```diff
@@ -15,74 +15,14 @@
...
 /**
- * Tests for {@link MeterTagSupport}.
+ * Tests for {@link MeterTagSupport} have been moved to the module
+ * micrometer-samples-spring-framework6 to test with the current version of Spring that
+ * requires JDK 17+..
  *
  * @author Marcin Grzejszczak
  * @author Johnny Lim
  */
 class MeterTagSupportTests {
```
shakuzen, line 26: We could just delete this file, but I thought it might help to have a reference in the same project to the test class for the main code.

docs/build.gradle
```diff
@@ -16,10 +16,29 @@ apply plugin: 'org.antora'
 
 description = 'Micrometer Docs'
 
+// Spring 6 requires Java 17+
+// skip this module when building with jdk <17
+if (!javaLanguageVersion.canCompileOrRun(17)) {
+    project.tasks.configureEach { task -> task.enabled = false }
```
shakuzen, line 22: I assume this is okay for the docs module since we run it with JDK 17 (according to [this](https://github.com/micrometer-metrics/micrometer/actions/runs/11266458878/workflow#L25-L29)), but let me know if I'm missing any reason this would be a problem.

docs/build.gradle
```diff
@@ -16,10 +16,29 @@ apply plugin: 'org.antora'
...
 
+java {
+    toolchain {
+        languageVersion = javaLanguageVersion
+    }
+}
+
+compileTestJava {
+    // need to override the config applied in the root build.gradle to all subprojects
+    // TODO can we not do this with Gradle's toolchains instead?
```
shakuzen, line 37: My hope was the above `java` block would take care of setting the Java target/source for the test compile task as well, but it didn't. I'm not sure if there's a better way to do it than I did here, but I ran out of patience looking into ways to configure Gradle.

micrometer-core/src/test/java/io/micrometer/core/aop/MeterTagSupportTests.java
```diff
@@ -15,74 +15,14 @@
...
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Method;
-
-import static org.assertj.core.api.Assertions.assertThat;
-
 /**
- * Tests for {@link MeterTagSupport}.
+ * Tests for {@link MeterTagSupport} have been moved to the module
+ * micrometer-samples-spring-framework6 to test with the current version of Spring that
+ * requires JDK 17+..
```
jonatan-ivanov, line 21:
```suggestion
 * requires JDK 17+.
```

jonatan-ivanov (approved):

PR #5583, opened 2024-10-10
Set user agent header in OTLP registry

### LINKED ISSUE #5577: Set user agent header in OTLP registry
The OTel specification defines that OTLP exporters SHOULD emit a `User-Agent` header identifying the exporter and its version. 

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.37.0/specification/protocol/exporter.md#user-agent

> OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
>
> `OTel-OTLP-Exporter-Python/1.2.3`

Right now, when using the OTLP registry in Micrometer, I get this "default" user agent `Java/21.*`. It makes it difficult to identify error with requests coming from micrometer apps using the OTLP registry.

**Rationale**
The exporter should follow the specification and emit the header with a more clear value. For example `OTel-OTLP-Exporter-Micrometer/1.2.3`

joaopgrassi:
Fixes #5577

I modified the code slightly, so I could mock/test the HTTP request the registry sends. Let me know if that's OK. 

CC @jonatan-ivanov

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
@@ -498,4 +498,5 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
         }
         return "Micrometer-OTLP-Exporter-Java/" + MeterRegistry.class.getPackage().getImplementationVersion();
     }
+
 }
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
@@ -91,6 +91,8 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
 
     private final TimeUnit baseTimeUnit;
 
+    private final String userAgentHeader;
+
     // Time when the last scheduled rollOver has started. Applicable only for delta
     // flavour.
     private volatile long lastMeterRolloverStartTime = -1;
@@ -117,15 +119,17 @@ public class OtlpMeterRegistry extends PushMeterRegistry {
         this(config, clock, threadFactory, new HttpUrlConnectionSender());
     }
 
+    // VisibleForTesting
     // not public until we decide what we want to expose in public API
     // HttpSender may not be a good idea if we will support a non-HTTP transport
-    private OtlpMeterRegistry(OtlpConfig config, Clock clock, ThreadFactory threadFactory, HttpSender httpSender) {
+    OtlpMeterRegistry(OtlpConfig config, Clock clock, ThreadFactory threadFactory, HttpSender httpSender) {
         super(config, clock);
         this.config = config;
         this.baseTimeUnit = config.baseTimeUnit();
         this.httpSender = httpSender;
         this.resource = Resource.newBuilder().addAllAttributes(getResourceAttributes()).build();
         this.aggregationTemporality = config.aggregationTemporality();
+        this.userAgentHeader = getUserAgentHeader();
         config().namingConvention(NamingConvention.dot);
         start(threadFactory);
     }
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -487,4 +492,11 @@ static double[] getSloWithPositiveInf(DistributionStatisticConfig distributionSt
         return sloWithPositiveInf;
     }
 
+    static String getUserAgentHeader() {
+        if (MeterRegistry.class.getPackage().getImplementationVersion() == null) {
```
shakuzen, line 496: In best practice, users should align versions (via our BOM etc.) but `MeterRegistry` is from a different module (micrometer-core) than `OtlpMeterRegistry` (micrometer-registry-otlp), so it is possible for misaligned versions to end up on the classpath. From the issue description, it sounds like it is the version of the OTLP exporter that's expected here (rather than the version of the library used for instrumentation), which I would say is the version of the micrometer-registry-otlp JAR, so probably better to get it from `OtlpMeterRegistry.class` or perhaps even better `this.getClass()` in a non-static context.

shakuzen (commented):
Thanks for the pull request. Overall it looks good. I left a review about how to get which version for this.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
jonatan-ivanov: Also, can we make this method `private`?
So together with Tommy's comment, how about this?
```java
private String getUserAgentHeader() {
    if (this.getClass().getPackage().getImplementationVersion() == null) {
        return "Micrometer-OTLP-Exporter-Java/unknown";
    }
    return "Micrometer-OTLP-Exporter-Java/" + this.getClass().getPackage().getImplementationVersion();
}
```

I'm also thinking if doing this would be better if the version number is not available:
```java
return "Micrometer-OTLP-Exporter-Java";
```

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMeterRegistryTest.java
```diff
@@ -129,6 +144,22 @@ void timeGauge() {
                     + "    time_unix_nano: 1000000\n" + "    as_double: 0.024\n" + "  }\n" + "}\n");
     }
 
+    @Test
+    void httpHeaders() throws Throwable {
```
jonatan-ivanov, line 148: Not crucial but we have an `@Issue` annotation that could be used here to connect the issue number to the test, something like this:

```suggestion
    @Test
    @Issue("#5577")
    void httpHeaders() throws Throwable {
```

jonatan-ivanov:
@shakuzen I think we should apply this on `1.12.x` (as a bug), what do you think?

shakuzen:
> I think we should apply this on `1.12.x` (as a bug), what do you think?

I'm undecided on whether I would consider it a bug or not, but I think it's low risk enough I'm fine with it going to maintenance branches.

jonatan-ivanov:
Fyi: I rebased this on `1.12.x` and also polished it a little:
https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:otlp-registry-emit-user-agen

I cant update the PR since I cannot write `dynatrace-oss-contrib:otlp-registry-emit-user-agent` :( @joaopgrassi Could you please push these changes to the PR. Also, is there a way you can open PRs from your own for not from `dynatrace-oss-contrib` so that we can modify the PRs?

Also, I tested it in a real app `.getPackage().getImplementationVersion()` always return null, we should not merge this PR until we find a way to get the version number.

jonatan-ivanov:
It seems my test was lying to me (I guess the jar was cached), I merged it in: https://github.com/micrometer-metrics/micrometer/commit/d7daaefffe0d74b5334a45273191fa2ad227f88d

joaopgrassi:
Hi all, sorry I was on PTO on Friday. Thanks for taking care of the changes!

Issue #5588, opened 2024-10-12
Map time units to UCUM format for Dynatrace


jonatan-ivanov:
Reported by @pirgeo.

`MeterRegistry` implementations mapping `TimeUnits` to a string representation using its `.toString()` through this cache: https://github.com/micrometer-metrics/micrometer/blob/eff8a6174e422714894ec8857555197493b78ccb/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L70-L78

This is the standard OpenJDK representation for time units but for Dynatrace, we need [UCUM](https://ucum.org/)-compliant units, e.g.: `s` instead of `seconds`, `ms` instead of `milliseconds`, etc.

PR #5589, opened 2024-10-12
Map time units to UCUM format for Dynatrace

jonatan-ivanov:
Closes gh-5588

### DIFF
--- a/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
+++ b/implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
@@ -15,8 +15,11 @@
  */
 package io.micrometer.dynatrace.v2;
 
-import com.dynatrace.metric.util.*;
+import com.dynatrace.metric.util.DynatraceMetricApiConstants;
+import com.dynatrace.metric.util.MetricException;
+import com.dynatrace.metric.util.MetricLineBuilder;
 import com.dynatrace.metric.util.MetricLineBuilder.MetadataStep;
+import com.dynatrace.metric.util.MetricLinePreConfiguration;
 import io.micrometer.common.lang.NonNull;
 import io.micrometer.common.util.StringUtils;
 import io.micrometer.common.util.internal.logging.InternalLogger;
@@ -60,8 +63,10 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
 
     private static final Pattern IS_NULL_ERROR_RESPONSE = Pattern.compile("\"error\":\\s?null");
 
-    private static final Map<String, String> staticDimensions = Collections.singletonMap("dt.metrics.source",
+    private static final Map<String, String> STATIC_DIMENSIONS = Collections.singletonMap("dt.metrics.source",
             "micrometer");
+
+    private static final Map<String, String> UCUM_TIME_UNIT_MAP = ucumTimeUnitMap();
 
     // Loggers must be non-static for MockLoggerFactory.injectLogger() in tests.
     private final InternalLogger logger = InternalLoggerFactory.getInstance(DynatraceExporterV2.class);
@@ -128,7 +133,7 @@ public final class DynatraceExporterV2 extends AbstractDynatraceExporter {
 
     private Map<String, String> enrichWithMetricsSourceDimensions(Map<String, String> defaultDimensions) {
         LinkedHashMap<String, String> orderedDimensions = new LinkedHashMap<>(defaultDimensions);
-        orderedDimensions.putAll(staticDimensions);
+        orderedDimensions.putAll(STATIC_DIMENSIONS);
         return orderedDimensions;
     }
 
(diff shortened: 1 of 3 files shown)


jonatan-ivanov:
@pirgeo Could you please confirm that this is not a breaking change for the users?
I.e.: Can one create a dashboard or any component on the Dynatrace UI / use the Dynatrace API responses and will be broken after this change?

/cc @shakuzen

implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -547,4 +553,33 @@ private String extractMetricKey(String metadataLine) {
...
+     */
+    private static Map<String, String> ucumTimeUnitMap() {
+        Map<String, String> mapping = new HashMap<>();
+        mapping.put("nanoseconds", "ns");
+        mapping.put("nanosecond", "ns");
+        mapping.put("microseconds", "us");
+        mapping.put("microsecond", "us");
+        mapping.put("milliseconds", "ms");
+        mapping.put("millisecond", "ms");
+        mapping.put("seconds", "s");
```
shakuzen, line 579: If the units are going to be coming from `TimeUnit#toString()`, wouldn't it be safer to build the map using e.g. `TimeUnit.SECONDS.toString()` instead of `"seconds"`?

jonatan-ivanov: I was thinking the same (got this from @pirgeo) and I even implemented it using `TimeUnit` but then I thought the intention here was to also handle cases when someone manually sets a time unit for a non-time-based `Meter` so I kept this mapping (not a good example because this is already in UCUM format: [`ProcessorMetrics`](https://github.com/micrometer-metrics/micrometer/blob/eff8a6174e422714894ec8857555197493b78ccb/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java#L124-L128)).

pirgeo: Yes, it would be nice to cover "all" cases. In a perfect world, all data producers would comply with just one standard. That's not very likely to happen anytime soon though, so my attempt was to get started here and at least keep it consistent for these seconds-based units, because these are the ones we have seen make problems in the past. At the end of the day, it shouldn't really matter whether the user sets `milliseconds`, `millisecond` or `ms`, they all refer to the same concept, so I just put them all in to have consistency.
If I recall correctly, using `TimeUnit#toString` yields `milliseconds`, so that is really the most important one to me. I am not opposed to adding them to the map - in fact, that would future-proof us in case the `TimeUnit#toString` ever changes.

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/v2/DynatraceExporterV2Test.java
```diff
@@ -866,6 +867,40 @@ void shouldAddMetadataOnlyWhenUnitOrDescriptionIsPresent() {
...
+                "test.second,dt.metrics.source=micrometer count,delta=100 " + clock.wallTime(),
+                "#test.second count dt.meta.unit=s",
+                "test.seconds,dt.metrics.source=micrometer count,delta=10 " + clock.wallTime(),
+                "#test.seconds count dt.meta.unit=s",
+                "process.cpu.time,dt.metrics.source=micrometer count,delta=1000000 " + clock.wallTime(),
+                "#process.cpu.time count dt.meta.unit=ns",
+                "test.ltt,dt.metrics.source=micrometer gauge,min=62000,max=62000,sum=62000,count=1 " + clock.wallTime(),
+                "#test.ltt gauge dt.meta.unit=ms")));
+    }
+
```
pirgeo, line 903: I wrote another test that ensures that all the TimeGauge instruments are mapped to the same `ms` base unit, but I only now realized that it's basically testing the same thing five times. Feel free to ignore this; I posted it anyway because it showcases how the TimeGauge would work with the Dynatrace exporter. 

```suggestion

    @Test
    void timeGaugeUnitIsMappedToBaseUnit() {
        HttpSender.Request.Builder builder = spy(HttpSender.Request.build(config.uri(), httpClient));
        when(httpClient.post(anyString())).thenReturn(builder);

        TimeGauge.builder("tg1", () -> 1., MILLISECONDS).register(meterRegistry);
        TimeGauge.builder("tg2", () -> 2., MICROSECONDS).register(meterRegistry);
        TimeGauge.builder("tg3", () -> 3., NANOSECONDS).register(meterRegistry);
        TimeGauge.builder("tg4", () -> 4., SECONDS).register(meterRegistry);
        TimeGauge.builder("tg5", () -> 5., MINUTES).register(meterRegistry);

        clock.add(config.step());
        exporter.export(meterRegistry.getMeters());

        // since the base unit of the registry is millis, everything will be exported as millis
        verify(builder).withPlainText(assertArg(body -> {
            assertThat(body.split("\n")).containsExactlyInAnyOrder(
                "tg1,dt.metrics.source=micrometer gauge,1 " + clock.wallTime(),
                "#tg1 gauge dt.meta.unit=ms",
                "tg2,dt.metrics.source=micrometer gauge,0.002 " + clock.wallTime(),
                "#tg2 gauge dt.meta.unit=ms",
                "tg3,dt.metrics.source=micrometer gauge,3.0E-6 " + clock.wallTime(),
                "#tg3 gauge dt.meta.unit=ms",
                "tg4,dt.metrics.source=micrometer gauge,4000 " + clock.wallTime(),
                "#tg4 gauge dt.meta.unit=ms",
                "tg5,dt.metrics.source=micrometer gauge,300000 " + clock.wallTime(),
                "#tg5 gauge dt.meta.unit=ms"
                );
        }));
    }
```

pirgeo (approved):
@jonatan-ivanov @shakuzen 
> Could you please confirm that this is not a breaking change for the users?

Confirmed, this change will not break any user. Existing users must trigger the unit upgrade, but new users will already get the UCUM-compliant ones.

PR #5590, opened 2024-10-14
Log delta count in addition to throughput in LoggingMeterRegistry

### LINKED ISSUE #5548: Log delta count in addition to throughput in LoggingMeterRegistry
In [LoggingMeterRegistry](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java), instruments such as counter, timer and histogram are logged only with a throughput (e.g. `throughput=0.016667/s`).\
I propose to add the count next to the throughput in the logs for those instruments (e.g. `throughput=0.016667/s count=1`).

**Rationale**
For instruments such as counter, timer and histogram, throughput is not always the most useful information to log.\
In some cases, users want to know the count as it is exported by other meter registries (e.g. OTLP).

Deriving the count from the throughput requires to know the step and it is not always obvious from logs (especially when logInactive is disabled).\
Adding the count in the logs next to the throughput would make the logs more clear.

**Additional context**
N/A

I can provide a PR if it is OK for you.

fstaudt:
Closes #5548

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
@@ -128,8 +128,8 @@ public class LoggingMeterRegistry extends StepMeterRegistry {
                     double count = counter.count();
                     if (!config.logInactive() && count == 0)
                         return;
-                    loggingSink.accept(
-                            print.id() + " delta_count=" + wholeOrDecimal(count) + " throughput=" + print.rate(count));
+                    loggingSink.accept(print.id() + " delta_count=" + print.humanReadableBaseUnit(count)
+                            + " throughput=" + print.rate(count));
                 }, timer -> {
                     HistogramSnapshot snapshot = timer.takeSnapshot();
                     long count = snapshot.count();
@@ -161,8 +161,8 @@ public class LoggingMeterRegistry extends StepMeterRegistry {
                     double count = counter.count();
                     if (!config.logInactive() && count == 0)
                         return;
-                    loggingSink.accept(
-                            print.id() + " delta_count=" + wholeOrDecimal(count) + " throughput=" + print.rate(count));
+                    loggingSink.accept(print.id() + " delta_count=" + print.humanReadableBaseUnit(count)
+                            + " throughput=" + print.rate(count));
                 }, timer -> {
                     double count = timer.count();
                     if (!config.logInactive() && count == 0)
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
@@ -49,6 +49,7 @@ import static java.util.stream.Collectors.joining;
  *
  * @author Jon Schneider
  * @author Matthieu Borgraeve
+ * @author Francois Staudt
  * @since 1.1.0
  */
 @Incubating(since = "1.1.0")
@@ -114,6 +115,7 @@ public class LoggingMeterRegistry extends StepMeterRegistry {
     @Override
     protected void publish() {
         if (config.enabled()) {
+            loggingSink.accept("Step: " + config.step().getSeconds() + "s");
             getMeters().stream().sorted((m1, m2) -> {
                 int typeComp = m1.getId().getType().compareTo(m2.getId().getType());
                 if (typeComp == 0) {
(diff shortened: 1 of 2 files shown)


shakuzen (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
```diff
@@ -114,6 +115,7 @@ private Function<Meter, String> defaultMeterIdPrinter() {
     @Override
     protected void publish() {
         if (config.enabled()) {
+            loggingSink.accept("Step: " + config.step().getSeconds() + "s");
```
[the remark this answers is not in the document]
fstaudt: I'll remove it for now.

I thought it would be interesting to log the step duration but it would actually only increase number of logs.\
Step duration can already be derived from delta_count & throughput of any metric.

micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
```diff
@@ -126,22 +128,24 @@ protected void publish() {
                     double count = counter.count();
                     if (!config.logInactive() && count == 0)
                         return;
-                    loggingSink.accept(print.id() + " throughput=" + print.rate(count));
+                    loggingSink
+                        .accept(print.id() + " delta_count=" + ((int) count) + " throughput=" + print.rate(count));
```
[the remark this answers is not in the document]
fstaudt: It was indeed for formatting reasons and I had not taken into account that counters could be increased with decimal values.\
I've found function wholeOrDecimal to format delta counts, I'll use it to avoid the cast.

micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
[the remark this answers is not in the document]
shakuzen: The step is logged on start for LoggingMeterRegistry (and any PushMeterRegistry). While the default behavior is that things are logged, users can customize the `loggingSink` to do whatever it wants with lines sent to it. So, sending a line with the step each time that's a different format than metrics lines is bound to cause issues for custom loggingSinks.

shakuzen (approved):
Thanks for the pull request and quick action on review. This looks good to me. We'll have to wait to merge until we start 1.15 development on `main`.

jonatan-ivanov:
What do you think about calling it `delta` instead of `delta_count`?

fstaudt:
> Thanks for the pull request and quick action on review. This looks good to me. We'll have to wait to merge until we start 1.15 development on `main`.

I was hoping that this change could still be part of 1.14.0 GA.\
What is the reason that it should wait for 1.15 ?

fstaudt:
> What do you think about calling it `delta` instead of `delta_count`?

I woud keep delta_count in the logs.\
The internal field is named count, it feels right to keep count in the logs.\
If we only keep delta, it would make it less clear for users that it is a indeed a count.

shakuzen:
> I was hoping that this change could still be part of 1.14.0 GA.
> What is the reason that it should wait for 1.15 ?

Our general policy is that we do not merge enhancements after a release candidate is released. 1.14.0-RC1 was released yesterday, so the next opportunity for enhancements is 1.15.0-M1.

fstaudt:
I've noticed a discrepancy in logs for counters between delta_count and throughput when baseUnit is set.\
I've added an extra commit with updated unit tests to fix this.

jonatan-ivanov:
We have a 6 months minor release cycle, based on our current release schedule, `1.15.0` should go GA around the second Monday of May (12th) but you might be able to use `1.15.0-M1` in January.

shakuzen:
I've gone ahead and added tentative dates for 1.15 to our support page: https://micrometer.io/support/

micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
```diff
@@ -157,13 +161,14 @@ protected void publish() {
...
+                    loggingSink.accept(print.id() + " delta_count=" + print.humanReadableBaseUnit(count)
+                            + " throughput=" + print.rate(count));
                 }, timer -> {
                     double count = timer.count();
                     if (!config.logInactive() && count == 0)
                         return;
-                    loggingSink.accept(print.id() + " throughput=" + print.rate(count) + " mean="
-                            + print.time(timer.mean(getBaseTimeUnit())));
+                    loggingSink.accept(print.id() + " delta_count=" + wholeOrDecimal(count) + " throughput="
+                            + print.unitlessRate(count) + " mean=" + print.time(timer.mean(getBaseTimeUnit())));
```
jonatan-ivanov, line 171: Was changing
```java
" throughput=" + print.rate(count)
```
to 
```java
" throughput=" + print.unitlessRate(count)
```
intentional?

jonatan-ivanov: @fstaudt Could you please take a look?

fstaudt: @jonatan-ivanov ,

sorry, I've skipped your previous comment.\
Let me check

fstaudt: @jonatan-ivanov ,

change was indeed intentional.

While writing unit tests for LoggingMeterRegistry, I noticed a discrepancy between logs for Timer and FunctionTimer.\
For Timer, I had: `" throughput=" + print.unitlessRate(count) + ...`.\
For FunctionTimer, I had `" throughput=" + print.rate(count) + ...`.

As a result, The logs for those 2 meters were:
```
my.timer{} delta_count=30 throughput=0.5/s mean=1s max=1s
my.function-timer{} delta_count=30 throughput=0.5 milliseconds/s mean=1s
```

The logs for function-timer were incorrect; the throughput for this meter provides the number of measurements by seconds and does not depend on the base unit of the meter.\
Therefore, I used unitlessRate for throughput to align the logging of those 2 meters and get similar logs for both:
```
my.timer{} delta_count=30 throughput=0.5/s mean=1s max=1s
my.function-timer{} delta_count=30 throughput=0.5/s mean=1s
```

Let me know your opinion on this.\
If you need me to create a separate issue and separate PR for this fix, please let me know.

jonatan-ivanov: Makes sense, thank you!

jonatan-ivanov (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
jonatan-ivanov: Good point, I created https://github.com/micrometer-metrics/micrometer/issues/5816 to track it, please let me know if any of you are interested opening a PR against `1.13.x`.

jonatan-ivanov: fyi: I found a few other things to improve with the tests, I'm going to do those + back-port them since it seems in `1.13.x` we are not even test `publish` that were also added here.

PR #5591, opened 2024-10-14
Improve average performance of LongTaskTimer for out-of-order stopping

fogninid:
The current implementation of DefaultLongTaskTimer optimizes for O(1) task starting, but performs poorly when stopping tasks that are not at the beginning of its internal queue (the oldest ones).
At the worst case, when calling stop immediately after starting, the stop call is currently expected to require O(N) operations.
Depending on the distribution of task lifetimes, the average case would be O(1) only for applications that stop tasks in exactly the same order as they were started; applications completing out-of-order and with unbiased lifetime would experience O(N) average.

Task stopping should not have any intrinsic difference to starting: both action are expected to be performed on application threads, and for a well-functioning application (that is not leaking of piling-up tasks) every call to start is matched by exactly one call to stop.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
@@ -98,7 +98,7 @@ public class DefaultLongTaskTimer extends AbstractMeter implements LongTaskTimer
 
     @Override
     public Sample start() {
-        final long startTime = clock.monotonicTime();
+        long startTime = clock.monotonicTime();
         SampleImpl sample = new SampleImpl(startTime);
         if (!activeTasks.add(sample)) {
             sample = new SampleImplCounted(startTime, nextNonZeroCounter());
@@ -132,8 +132,7 @@ public class DefaultLongTaskTimer extends AbstractMeter implements LongTaskTimer
     @Override
     public double max(TimeUnit unit) {
         try {
-            Sample oldest = activeTasks.first();
-            return oldest.duration(unit);
+            return activeTasks.first().duration(unit);
         }
         catch (NoSuchElementException e) {
             return 0.0;
@@ -278,14 +277,17 @@ public class DefaultLongTaskTimer extends AbstractMeter implements LongTaskTimer
         public String toString() {
             double durationInNanoseconds = duration(TimeUnit.NANOSECONDS);
             return "SampleImpl{" + "duration(seconds)=" + TimeUtils.nanosToUnit(durationInNanoseconds, TimeUnit.SECONDS)
-                    + ", " + "duration(nanos)=" + durationInNanoseconds + ", " + "startTimeNanos=" + startTime + '}';
+                    + ", duration(nanos)=" + durationInNanoseconds + ", startTimeNanos=" + startTime + '}';
         }
 
         @Override
-        public int compareTo(DefaultLongTaskTimer.SampleImpl o) {
-            int startCompare = Long.compare(startTime, o.startTime);
+        public int compareTo(DefaultLongTaskTimer.SampleImpl that) {
+            if (this == that) {
+                return 0;
+            }
+            int startCompare = Long.compare(this.startTime, that.startTime);
             if (startCompare == 0) {
-                return Integer.compare(counter(), o.counter());
+                return Integer.compare(this.counter(), that.counter());
             }
             return startCompare;
         }
@@ -304,6 +306,14 @@ public class DefaultLongTaskTimer extends AbstractMeter implements LongTaskTimer
         @Override
         int counter() {
             return counter;
+        }
+
+        @Override
+        public String toString() {
+            double durationInNanoseconds = duration(TimeUnit.NANOSECONDS);
+            return "SampleImplCounted{" + "duration(seconds)="
+                    + TimeUtils.nanosToUnit(durationInNanoseconds, TimeUnit.SECONDS) + ", duration(nanos)="
+                    + durationInNanoseconds + ", startTimeNanos=" + super.startTime + ", counter=" + this.counter + '}';
         }
 
     }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
@@ -27,9 +27,9 @@ import io.micrometer.core.instrument.util.TimeUtils;
 
 import java.util.*;
 import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.ConcurrentLinkedDeque;
-import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
```diff
@@ -92,7 +96,9 @@ public DefaultLongTaskTimer(Id id, Clock clock, TimeUnit baseTimeUnit,
     @Override
     public Sample start() {
         SampleImpl sample = new SampleImpl();
-        activeTasks.add(sample);
+        if (!activeTasks.add(sample)) {
+            throw new IllegalStateException();
```
shakuzen, line 100: We shouldn't throw exceptions when recording metrics, generally. In what case would this happen?

shakuzen (commented):
Thanks for the pull request. What you wrote makes sense. Still, I wanted to verify with some JMH benchmarks so we can put some numbers behind it and have them in place for checking any future changes that would affect performance around this. I made https://github.com/micrometer-metrics/micrometer/pull/5595 to add JMH benchmarks.

shakuzen:
Sharing results from my MacBook Pro M1 with the benchmarks in the linked PR with 10,000 active samples and stopping a random sample. As expected, start is slower but the overall time to start and stop on average (with a random sample) is better.

Before
```
Benchmark                                   Mode  Cnt   Score   Error  Units
DefaultLongTaskTimerBenchmark.start           ss  200   0.495 ± 0.064  us/op
DefaultLongTaskTimerBenchmark.startAndStop    ss  200  15.351 ± 2.508  us/op
DefaultLongTaskTimerBenchmark.stopRandom      ss  200  14.784 ± 2.584  us/op
```

After
```
Benchmark                                   Mode  Cnt  Score   Error  Units
DefaultLongTaskTimerBenchmark.start           ss  200  1.002 ± 0.116  us/op
DefaultLongTaskTimerBenchmark.startAndStop    ss  200  6.338 ± 0.577  us/op
DefaultLongTaskTimerBenchmark.stopRandom      ss  200  5.154 ± 0.608  us/op
```

micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
fogninid: It is a leftover from testing. See the implementation of SampleImpl#compareTo.
I was a bit surprised that the skip-list considers the comparator returning "0" as sufficient test for equality, and does not check Object#equals.
I am intentially leaving equality to be only reference identity (obviously two samples are never the same), but the naive comparator can not distinguish if two samples are created at exactly the same startTime. Putting hashCode in the comparison makes it "almost always" a total order, but it can theoretically still fail for hash collisions.

You can try removing the hashCode check from the comparator, and see some unit tests will fail.

A fail-proof solution would be to use a strictly increasing counter (for example an AtomicLong) instead of the time based one, but it does not look worth to increase the heap memory usage of SampleImpl.

I tend to say it is better to remove the exception that you mention here and accept that some samples can be lost (not accounted) if bothe the monotonic time does not increase and there are hash collisions.

fogninid: @shakuzen how do you suggest me to proceed?

Should I change the `DefaultLongTaskTimer.SampleImpl` class to use more heap memory and store an `AtomicInteger` to be always correct even in those very rare corner cases?

Or do you have another suggestions?

shakuzen: Sorry for the late response. Now that 1.14 GA is released, I'll get back to reviewing this so we can hopefully get it merged. I'll take a look tomorrow.

fogninid: > [..] everyone would have to pay [..]

Good point, what about making only the collisions pay for storing more memory?
https://github.com/micrometer-metrics/micrometer/pull/5591/commits/d8dcd3bed4c7696c2ac64248b3d6b566dbccae00

The additional functions and indirect calls should also almost never happen, so the only disadvantage I see would be in the slightly non-obvious code

fogninid: @shakuzen have you thought about which approach you would like for me to implement?

shakuzen: Your approach looks like the best we've come up with so far. I think we can go with that.

micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java
```diff
@@ -91,11 +98,27 @@ public DefaultLongTaskTimer(Id id, Clock clock, TimeUnit baseTimeUnit,
...
+        if (!activeTasks.add(sample)) {
+            sample = new SampleImplCounted(startTime, nextNonZeroCounter());
+            activeTasks.add(sample);
+        }
         return sample;
     }
 
+    private int nextNonZeroCounter() {
+        int nextCount;
+        while ((nextCount = counter.incrementAndGet()) == 0) {
```
shakuzen, line 112: I'm not sure I understand why this (whole method) is needed. Could you explain?

fogninid: We should try to avoid `0` here because the classes are constructed such that `SampleImplCounted(startTime, 0)` compares equal to `SampleImpl(startTime)`, so the two would collide again.

This will happen on the 2^32-th collision on the same instance of `DefaultLongTaskTimer`, when `counter` has wrapped around the whole int range.

If you feel skipping 1 in 2^32 collisions is acceptable compared to having the strange code, I can just remove this method and the relative test

micrometer-core/src/test/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimerTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertThat(sample3.stop()).isEqualTo(TimeUnit.SECONDS.toNanos(0));
+        assertThat(t.activeTasks()).isEqualTo(2);
+        assertThat(sample4.stop()).isEqualTo(TimeUnit.SECONDS.toNanos(0));
+        assertThat(t.activeTasks()).isEqualTo(1);
+        assertThat(sample1.stop()).isEqualTo(TimeUnit.SECONDS.toNanos(1));
+        assertThat(t.activeTasks()).isEqualTo(0);
+    }
+
+    @Test
+    void counterJumpsZeroAndWraps() {
```
jonatan-ivanov, line 94: Can we split this into two?
1. `counterShouldSurviveOverflow`
2. `counterShouldJumpZero`

micrometer-core/src/test/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimerTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class DefaultLongTaskTimerTest {
+
+    private static final InstanceOfAssertFactory<SampleImpl, AbstractComparableAssert<?, SampleImpl>> SAMPLE_IMPL_ASSERT = new InstanceOfAssertFactory<>(
+            SampleImpl.class, Assertions::assertThat);
+
+    @Test
+    void sampleTimestampCollision() {
```
jonatan-ivanov, line 40: What do you think about naming the tests about the behavior we expect?
```suggestion
    void timestampCollisionShouldBeFine() {
```

micrometer-core/src/test/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimerTest.java
```diff
@@ -0,0 +1,115 @@
...
+        final MockClock clock = new MockClock();
+        MeterRegistry registry = new SimpleMeterRegistry(SimpleConfig.DEFAULT, clock);
+        LongTaskTimer t = LongTaskTimer.builder("my.timer").register(registry);
```
jonatan-ivanov, line 43: Can we eliminate one-letter variables e.g.: use `ltt` or `timer` instead here and later in this test class too?

jonatan-ivanov (commented):
Thank you for the PR, I really like this.
Not necessarily in this PR but I' also wondering if we should add a JCStress tests on top of #5595.

jonatan-ivanov:
Fyi: I polished the PR a bit and also resolved the comments above and rebased it to `main` (otherwise the build failed).

jonatan-ivanov:
@fogninid Thank you for the contribution!

Issue #5296, opened 2024-07-10
zgcGenerationalTest is flaky on CI


shakuzen:
See, for example, https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/6982/workflows/66800576-a887-4030-b36a-5068d498950b/jobs/35044. JvmGcMetricsTest > gcTimingIsCorrectForPauseCycleCollectors() fails fairly often when running the `zgcGenerationalTest` task during the CI build. Unfortunately I think it's a matter of whether a GC happens during the test code running that is collecting numbers to make assertions about the metrics having the right numbers for GCs. I don't have a great idea how to improve this.

```
> Task :micrometer-core:zgcGenerationalTest

JvmGcMetricsTest > gcTimingIsCorrectForPauseCycleCollectors() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 11L
     but was: 10L
        at app//io.micrometer.core.instrument.binder.jvm.JvmGcMetricsTest.lambda$checkPhaseCount$3(JvmGcMetricsTest.java:236)
```

PR #5602, opened 2024-10-23
Improving flaky zgcGenerationalTest (#5296)

the-thing:
Improves https://github.com/micrometer-metrics/micrometer/issues/5296

Moved calculation of expected values into awaitility. This will allow to retry assertion when GC happens between reading from GarbageCollectorMXBean and respective meters.

Comments
 - the original test doesn't check if GC values are greater than initial values before explicit GC.
 - await could be longer


shakuzen (approved):
Thanks for the pull request.

PR #5605, opened 2024-10-24
Support assertions on Observation events

### LINKED ISSUE #5576: micrometer-observation-test support for assertions on events
The `micrometer-observation-test` dependency does not appear to offer a simple way of asserting on observation events in tests

**Rationale**
The `TestObservationRegistryAssert` class handles assertions on what observations have been made, the key values associated with an observation, etc. But it does not appear to provide support for asserting that particular calls to `Observation#event(...)` have been made.

This is necessary to provide test coverage for observations which raise events.

**Additional context**
This is based  on using version `1.13.2` of `io.micrometer:micrometer-observation-test`

wabrit:
Adds the ability to assert that Events have been raised on Observations; matching can be done against event name only, or event name and contextual name.

Resolves #5576

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistry.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistry.java
@@ -19,8 +19,10 @@ import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationHandler;
 import io.micrometer.observation.ObservationRegistry;
 
+import java.util.HashSet;
 import java.util.Objects;
 import java.util.Queue;
+import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
 import org.assertj.core.api.AssertProvider;
@@ -118,6 +120,14 @@ public final class TestObservationRegistry
             return true;
         }
 
+        @Override
+        public void onEvent(final Observation.Event event, final Observation.Context context) {
+            this.contexts.stream()
+                .filter(testContext -> testContext.getContext() == context)
+                .findFirst()
+                .ifPresent(testContext -> testContext.addEvent(event));
+        }
+
     }
 
     static class TestObservationContext {
@@ -127,6 +137,8 @@ public final class TestObservationRegistry
         private boolean observationStarted;
 
         private boolean observationStopped;
+
+        private final Set<Observation.Event> contextEvents = new HashSet<>();
 
         TestObservationContext(Observation.Context context) {
             this.context = context;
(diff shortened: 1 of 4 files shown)


shakuzen (approved):
Looks good to me. Thanks.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistry.java
```diff
@@ -179,6 +191,37 @@ Observation.Context getContext() {
...
+
+        /**
+         * Check if an {@link Observation.Event} with the given name and contextual name
+         * was stored in this context.
+         * @param name name of the event to check
+         * @param contextualName contextual name of the event to check
+         * @return {@code true} if an event was stored under the given name and contextual
+         * name
+         */
+        public boolean hasEvent(String name, String contextualName) {
```
shakuzen, line 220: I think the methods added here to the TestContext do not need to be public. Could we get away with them being package private? We don't want to add unnecessary public API in case we want to change it, and users should be using the API on TestObservationRegistryAssert, right?

shakuzen (changes requested):

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistry.java
wabrit: Good catch - I'll get that fixed.

shakuzen (approved):

wabrit:
Many thanks @shakuzen for approving this PR; is there anything else I need to do to progress it further?

wabrit:
Hi @shakuzen any chance this could get merged into main? I don't have write permissions.

shakuzen:
@wabrit Sorry for the wait. I should have explained. We were waiting for the 1.14.0 GA release when we would create a `1.14.x` branch, which in turn would make `main` for 1.15.0 development. We were already in the release candidate phase for 1.14, and we in general don't merge enhancements after entering the release candidate phase. Now that `main` is for 1.15 development, we can merge this in. Thank you for your contribution.

wabrit:
Thanks @shakuzen - no problem, completely understand. Happy to contribute!

When (roughly) would the 1.15.0 release see the light of day? Just wondering whether its worth holding off some of our own components merges (which rely on a rather hacky workaround to this) until we can do it "the right way".

shakuzen:
> When (roughly) would the 1.15.0 release see the light of day?

The planned release date for 1.15.0 GA is on our support page at https://micrometer.io/support/
We will be releasing milestones before then for testing, and snapshots are published for every commit.

PR #5606, opened 2024-10-24
[DynatraceV2] Log simple counts for exported data points

pirgeo:
This PR adds some basic status reporting on `INFO` log level. We have debug logs for the number of accepted and rejected items for each request. However, we have seen cases where many requests must be sent, and these logs can get very verbose. Additionally, they are DEBUG logs, so by definition, they are not expected to be available in a production system. 

In order to get a general heartbeat on the health of the export, I would like to add this log line. It will be written once per export (so usually once a minute) and looks something like this:

```
:46:17 dynatrace-metrics-publisher INFO i.m.d.v.DynatraceExporterV2 - Export finished. Generated 3 request(s) totaling 1706 metric lines and 1204 metadata lines. Of those, 2910 lines were successfully ingested, 0 lines failed to ingest. Target URI: https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest
```

(The `{your-environment-id}` part would obviously be replaced with the actual URI)


pirgeo:
I identified a potential bug in the exporting (unrelated to this change) that will require a code change in a similar area as this. I think I'll close this issue here, investigate the other bug, and if it turns out to be a bug, I'll raise a bug fix that contains the change proposed in this PR.

PR #5610, opened 2024-10-28
Add docs for VirtualThreadMetrics

jonatan-ivanov:
See https://github.com/micrometer-metrics/micrometer/pull/5067


shakuzen (changes requested):
I think we need to mention somehow that `VirtualThreadMetrics` is in the micrometer-java21 module, so users know they need to add that dependency to use it. Maybe putting it in a different section/paragraph than the other things would be better in this sense.

jonatan-ivanov:
👍🏼 I moved it to a separate section.

shakuzen (approved):

PR #5612, opened 2024-10-29
Enable japicmp task for micrometer-registry-statsd

izeye:
This PR enables `japicmp` task for `micrometer-registry-statsd` as it seems to be disabled unintentionally.


jonatan-ivanov:
Thank you so much! I moved this to `1.12.x`.

PR #5617, opened 2024-10-31
Warn about Gauge re-registration

jonatan-ivanov:


jonatan-ivanov:
Since it is on the hot-path, I'm not 100% sure about the warning but I think we should have the tests and document it.

docs/modules/ROOT/pages/concepts/gauges.adoc
```diff
@@ -42,7 +42,7 @@ Note that, in this form, unlike other meter types, you do not get a reference to
...
 
-WARNING: Attempting to construct a gauge with a primitive number or one of its `java.lang` object forms is always incorrect. These numbers are immutable. Thus, the gauge cannot ever be changed. Attempting to "`re-register`" the gauge with a different number does not work, as the registry maintains only one meter for each unique combination of name and tags.
+WARNING: Attempting to construct a gauge with a primitive number or one of its `java.lang` object forms is always incorrect. These numbers are immutable. Thus, the gauge cannot ever be changed. Attempting to "re-register" the gauge with a different number does not work, as the registry maintains only one meter for each unique combination of name and tags. "Re-registering" a gauge can happen indirectly for example as the result of a `MeterFilter` modifying the name and/or the tags of two different gauges so that they will be the same after the filter is applied.
```
keith-turner, line 45: Not sure if text should be added for this here, but this problem with gauges seems like another reason to user Counter meters instead of gauges if possible.   The docs to nudge twoards using counters instead of gauges another place.

keith-turner (approved):
> Since it is on the hot-path, I'm not 100% sure about the warning but I think we should have the tests and document it.

I like the warning because it makes it easier to see that information is being ignored.    I only noticed information was being lost because I tested it in a system where I know what the gauges data should be.  When meters are silently dropping data and its not know what the data should be, someone will take the data they see and not even know they are not seeing some of the data that was registered with the system.

docs/modules/ROOT/pages/concepts/gauges.adoc
shakuzen: I wouldn't add text to that effect because already if instead of a gauge you could use a counter for the metric, you should. Adding text about this as a reason to use a counter instead of a gauge might cause readers to infer if they don't have this issue it is okay to use a gauge as a counter.

shakuzen:
> Since it is on the hot-path, I'm not 100% sure about the warning

We should see how much it costs with JMH benchmarks. I don't know if we already have a good benchmark that would measure the difference.

shakuzen:
@jonatan-ivanov do you want to separate out the documentation changes from the code changes so we can get the documentation changes into the release and work on benchmarking the code changes?

Issue #5604, opened 2024-10-24
Native Image / Spring Boot / Hazelcast: java.lang.NoSuchMethodError: com.hazelcast.map.IMap.getName()


blaghed:
**Describe the bug**

I've run into the same situation as reported here:
https://stackoverflow.com/questions/76807745/hazelcast-with-spring-boot-3-graalvm-native-nosuchmethoderror

Basically, both Hazelcast and Spring Boot are working with Native Image, but sadly Micrometer's `HazelcastCacheMetrics` breaks this compatibility with: `java.lang.NoSuchMethodError: com.hazelcast.map.IMap.getName()`

There is no workaround available, sadly.

Extended stacktrace:
```
Caused by: java.lang.ExceptionInInitializerError
	at io.micrometer.core.instrument.binder.cache.HazelcastCacheMetrics.<init>(HazelcastCacheMetrics.java:79)
	at org.springframework.boot.actuate.metrics.cache.HazelcastCacheMeterBinderProvider.getMeterBinder(HazelcastCacheMeterBinderProvider.java:47)
	at org.springframework.boot.actuate.metrics.cache.HazelcastCacheMeterBinderProvider.getMeterBinder(HazelcastCacheMeterBinderProvider.java:41)
	at org.springframework.boot.actuate.metrics.cache.CacheMetricsRegistrar.lambda$getMeterBinder$0(CacheMetricsRegistrar.java:76)
	at org.springframework.boot.util.LambdaSafe$Callbacks.lambda$invokeAnd$2(LambdaSafe.java:301)
	at org.springframework.boot.util.LambdaSafe$LambdaSafeCallback.invoke(LambdaSafe.java:159)
	at org.springframework.boot.util.LambdaSafe$Callbacks.lambda$invokeAnd$3(LambdaSafe.java:300)
	at java.base@17.0.9/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base@17.0.9/java.util.Spliterators$IteratorSpliterator.tryAdvance(Spliterators.java:1856)
	at java.base@17.0.9/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129)
	at java.base@17.0.9/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527)
	at java.base@17.0.9/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
[...]

PR #5618, opened 2024-11-01
Add native image support for Hazelcast instrumentation

izeye:
This PR tries to add native image support for Hazelcast.

Closes gh-5604

### DIFF
--- a/micrometer-core/src/main/resources/META-INF/native-image/io.micrometer/micrometer-core/reflect-config.json
+++ b/micrometer-core/src/main/resources/META-INF/native-image/io.micrometer/micrometer-core/reflect-config.json
@@ -27,5 +27,71 @@
   {
     "name":"org.HdrHistogram.Histogram",
     "methods":[{"name":"<init>","parameterTypes":["long","long","int"] }]
+  },
+  {
+    "name":"com.hazelcast.core.DistributedObject",
+    "methods":[{"name":"getName","parameterTypes":[] }]
+  },
+  {
+    "name":"com.hazelcast.map.IMap",
+    "methods":[{"name":"getLocalMapStats","parameterTypes":[] }]
+  },
+  {
+    "name":"com.hazelcast.core.IMap",
+    "methods":[{"name":"getLocalMapStats","parameterTypes":[] }]
+  },
+  {
+    "name":"com.hazelcast.map.LocalMapStats",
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/resources/META-INF/native-image/io.micrometer/micrometer-core/reflect-config.json
```diff
@@ -27,5 +27,71 @@
   {
     "name":"org.HdrHistogram.Histogram",
     "methods":[{"name":"<init>","parameterTypes":["long","long","int"] }]
+  },
+  {
+    "name":"com.hazelcast.core.DistributedObject",
+    "methods":[{"name":"getName","parameterTypes":[] }]
```
shakuzen, line 33: I think it's going to get unwieldy quickly if we try to put reflection config for everything we have instrumentation for here. Since there is already metadata [here](https://github.com/oracle/graalvm-reachability-metadata/tree/master/metadata/com.hazelcast/hazelcast), can users not rely on that? That way we don't have to maintain reflection config for Hazelcast and distribute it in our core jar when most of our users may not even use hazelcast?

izeye: @shakuzen Thanks for the feedback!

I'm not sure if I understand your feedback correctly, but these reflection hints are due to reflection usages on Hazelcast in the `HazelcastIMapAdapter`, so I thought this was a proper place for them.

Am I missing something?

shakuzen: I think I just wasn't paying full attention. That makes sense, and I guess there isn't a good alternative to us including the reflection config here, then.

shakuzen (approved):
Looks good to me. Do you think there's any issue merging this for `1.12.x` so we can resolve the issue in patch releases for Hazelcast users?

We should probably think about adding some native image testing if we're intending to support it as a feature, but I think we can consider/work on that in a separate pull request from this.

izeye:
> Do you think there's any issue merging this for `1.12.x` so we can resolve the issue in patch releases for Hazelcast users?

@shakuzen No, I switched the base branch to `1.12.x`.

Issue #5093, opened 2024-05-13
Breaking Change in 1.13.0


TheSench:
**Describe the bug**
Version 1.13.0 moved `PrometheusMeterRegistry`, resulting in `NoClassDefFoundError` in code referencing the class in its old location in 1.12.5.

**Environment**

 - Micrometer version: `1.13.0`
 - Micrometer registry: `prometheus`
 - OS: `N/A`
 - Java version: `openjdk 17.0.9 2023-10-17`

**Expected behavior**
A new major version, due to the breaking changes to the API surface (e.g. package of various classes).

**Additional context**
This class appears to have been moved to a new package [here](https://github.com/micrometer-metrics/micrometer/commit/d651867b9f19b859534816dea413945562a4c53b) (along with many others).

jonatan-ivanov:
Please take a look at the [Release Notes](https://github.com/micrometer-metrics/micrometer/releases/tag/v1.13.0) and the [1.13 Migration Guide](https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide).

Here's some background:
- After 9 years of being on 0.x, Prometheus release 1.x with breaking changes
- `micrometer-registry-prometheus` was migrated to the new 1.x client
- At the same time the original bits `micrometer-registry-prometheus` (Prometheus 0.x) were moved to `micrometer-registry-prometheus-simpleclient`

The decision we needed to make: move to 1.x and offer a way go back to 0.x or stick with 0.x and let users opt-in to the upgrade. We decided to migrate `micrometer-registry-prometheus` to 1.x and also offer `micrometer-registry-prometheus-simpleclient` which is the exact copy of `micrometer-registry-prometheus` before the migration.

The decision you need to make: migrate to 1.x (use `micrometer-registry-prometheus`) or stay on 0.x (use `micrometer-registry-prometheus-simpleclient`).

If you want to migrate to the new client, this is the change you should make:
```
- implementation 'io.micrometer:micrometer-registry-prometheus:1.12.6'
+ implementation 'io.micrometer:micrometer-registry-prometheus:1.13.0'
```

If you do not want to migrate to the new client, this is the change you should make:
```
- implementation 'io.micrometer:micrometer-registry-prometheus:1.12.6'
+ implementation 'io.micrometer:micrometer-registry-prometheus-simpleclient:1.13.0'
```
(Please notice that you need to change the version number and the name of the artifact in this case.)

[...]

PR #5652, opened 2024-11-10
Use BOM for Micrometer dependency examples in reference docs

izeye:
This PR changes to use Micrometer version for Micrometer dependency examples in reference docs.

See gh-5093


shakuzen (approved):

jonatan-ivanov (approved):

jonatan-ivanov:
Should we do this starting from `1.12.x`?
I can take care of rebasing/back-porting if we decide so.

shakuzen:
Yes, I think we should do it in all the Antora-based docs.

shakuzen:
I think another option that Jonatan mentioned offline is we could go in the direction of either showing how to configure the BOM in the install snippet, or we could link to another section of the docs that talks about setting up the BOM. Then we could remove the version from individual dependency declarations and mention that it's expected to use the BOM so you don't need to declare explicit versions on dependencies. That is probably better since we do expect the version to be managed for the vast majority of users, and we do not want to encourage people to declare the version at the level of individual dependencies. @izeye would you be willing to update the pull request to that effect if you don't see any issues with that approach?

izeye:

Sounds good. I'll try to update this in that direction.

By the way, did you mean to apply this to 1.13.x, not to 1.12.x? 1.12.x also seems to be Antora-based although it seems to be soon to be EOLed for OSS support.

shakuzen:
> By the way, did you mean to apply this to 1.13.x, not to 1.12.x? 1.12.x also seems to be Antora-based although it seems to be soon to be EOLed for OSS support.

I should have left a comment on that. We should target and apply this to `1.12.x`. I made the milestone 1.13 because we don't plan to have another OSS release of 1.12, but we haven't moved the branch to the private repository for enterprise support yet.

izeye:
@shakuzen Thanks for the explanation!

I switched the base branch to `1.12.x` and updated as suggested.

jonatan-ivanov:
I added a polish commit (https://github.com/micrometer-metrics/micrometer/pull/5652/commits/51fe18942a243c3b5041d52cc9fed1803d404de0): moved the templates a level up and found a few other places where it can be used (during forward merge, we should do the same exercise).

izeye:
@jonatan-ivanov Looks great. Thank you!

jonatan-ivanov:
@izeye I talked to Tommy and we were thinking if it is better to have the BOM instructions in the in install section and everywhere else place a link to it. Tommy's guess was if we duplicate it everywhere, people might not understand and have duplicate BOM declarations.

I made these changes on the PR, what do you think?

docs/modules/ROOT/pages/implementations/influx.adoc
```diff
@@ -120,22 +122,18 @@ management.metrics.export.influx:
...
-
-Add the following dependency:
+After the BOM is xref:../installing.adoc[configured], add the following dependency:
 
 [source,groovy]
 ----
 implementation 'io.micrometer:micrometer-registry-statsd'
 ----
 
-The following example shows the Maven configuration:
```
izeye: "NOTE: The version is not needed for this dependency since it is defined by the BOM." seems to be missing.

docs/modules/ROOT/pages/installing.adoc
```diff
@@ -4,24 +4,48 @@
 
 Micrometer contains a core library with the instrumentation SPI and an in-memory implementation that does not export data anywhere, a series of modules with implementations for various monitoring systems, and a test module.
 
-To use Micrometer, add the dependency for your monitoring system.
+To use Micrometer, add the dependency for your monitoring system. It is recommended to use the BOM provided by Micrometer, you can configure it as follows (you only need to declare it once in your project).
 
-The following example adds Prometheus in Gradle:
+NOTE: If you use a framework, it might already comes with a BOM that defines the Micrometer versions or imports the Micrometer BOM. If this is the case (for example if you use Spring Boot), you can skip configuring the Micrometer BOM.
```
izeye, line 9: Typo: "comes" -> "come"

izeye (commented):
Thanks! Looks good to me.

docs/modules/ROOT/pages/implementations/influx.adoc
jonatan-ivanov: Thank you, fixed.

docs/modules/ROOT/pages/installing.adoc
jonatan-ivanov: Thank you, fixed.

docs/modules/ROOT/pages/installing.adoc
```diff
@@ -4,26 +4,61 @@
 
 Micrometer contains a core library with the instrumentation SPI and an in-memory implementation that does not export data anywhere, a series of modules with implementations for various monitoring systems, and a test module.
 
-To use Micrometer, add the dependency for your monitoring system.
+To use Micrometer, add the dependency for your monitoring system. It is recommended to use the BOM provided by Micrometer, you can configure it as follows (you only need to declare it once in your project).
 
-The following example adds Prometheus in Gradle:
+NOTE: If you use a framework, it might already come with a BOM that defines the Micrometer versions or imports the Micrometer BOM. If this is the case (for example if you use Spring Boot), you can skip configuring the Micrometer BOM.
```
shakuzen, line 9:
```suggestion
NOTE: If you use a framework, it might have dependency management that defines Micrometer versions or imports the Micrometer BOM. You can defer to the framework's dependency management instead of declaring the Micrometer BOM directly in that case.
```
Reworded it some. It is possible to use Spring Boot without its dependency management. Initially I tried to rewrite to specifically say if you're using Spring Boot's dependency management, but I didn't want to get into the different ways that might happen (Gradle plugin vs platform BOM vs Maven parent etc), and it felt redundant with the more generic "framework's dependency management"

shakuzen (commented):
Looks good. I just had one suggestion on wording. Let me know what you think.

PR #5663, opened 2024-11-14
Remove obsolete .gitattributes

izeye:
This PR removes the `.gitattributes` file as it only contains an obsolete entry.


jonatan-ivanov:
Thank you!

jonatan-ivanov (approved):

Issue #2528, opened 2021-03-24
Using micrometer-registry-statsd doesn't report TimeUnit to DataDog


ayeletmorris:
Hi, I'm using micrometer-registry-statsd with @Timed to report time series data to DataDog, and all of my metrics have "none" in the time unit section in the metrics summary page.

I am trying to understand what is the TimeUnit that is being used and to even define it manually and I saw in your code that the timers use Nanoseconds to save the data, but when I defined the metrics time unit as nanoseconds the data makes no sense, I am trying to dig deeper and I saw that in the StatsdMeterRegistry class the "getBaseTimeUnit" method returns Milliseconds, so now I'm confused...

Basically my questions are:
1. What time unit does the timed annotation (timers) report to dataDog?
2. Why doesn't it show the supposedly reported time unit int he dataDog metric?

shakuzen:
> I saw in your code that the timers use Nanoseconds to save the data

Nanoseconds is an internal representation. Depending on the registry, timings will be converted to the base unit for that implementation, according to what is idiomatic or required for that backend.

>     1. What time unit does the timed annotation (timers) report to dataDog?

You're using the statsd registry, and [the statsd standard for timing is milliseconds](https://github.com/statsd/statsd/blob/master/docs/metric_types.md#timing). So we report it in milliseconds.

> Why doesn't it show the supposedly reported time unit int he dataDog metric?

That's really a question for the maintainers of DogStatsD. It seems like they could include such metadata but aren't currently, as far as I can tell.
If you were using the DatadogMeterRegistry (and you configure an application key) instead of statsd, we do report the time unit metadata for metrics since we directly report them to Datadog.

PR #5687, opened 2024-11-15
Consistently use baseTimeUnit in StatsD

coreyoconnor:
Draft: Opening for discussion with (incomplete) proposed changes.

There is a `baseTimeUnit` for `Timer`s that appears to be inconsistently used. Given the definition "The base time unit of the timer to which all published metrics will be scaled" I'd expect that a value, eg 2000 sec, reported as the value 2000 with the unit seconds would be emitted to statsd as the value in the base time units,  2000 in this example. However, the actual value is 2,000,000. Which is the corresponding millisecond value for the seconds expected to be recorded.

As that explainer above is a challenge to read: I've included some (untested or even compiled) changes to illustrate.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdTimer.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdTimer.java
@@ -43,9 +43,9 @@ public class StatsdTimer extends AbstractTimer {
 
     StatsdTimer(Id id, StatsdLineBuilder lineBuilder, FluxSink<String> sink, Clock clock,
             DistributionStatisticConfig distributionStatisticConfig, PauseDetector pauseDetector, TimeUnit baseTimeUnit,
-            long stepMillis) {
+            long stepBaseUnits) {
         super(id, clock, distributionStatisticConfig, pauseDetector, baseTimeUnit, false);
-        this.max = new StepDouble(clock, stepMillis);
+        this.max = new StepDouble(clock, stepBaseUnits);
         this.lineBuilder = lineBuilder;
         this.sink = sink;
     }
@@ -55,13 +55,13 @@ public class StatsdTimer extends AbstractTimer {
         if (!shutdown && amount >= 0) {
             count.increment();
 
-            double msAmount = TimeUtils.convert(amount, unit, TimeUnit.MILLISECONDS);
-            totalTime.add(msAmount);
+            double baseUnitAmount = TimeUtils.convert(amount, unit, baseTimeUnit());
+            totalTime.add(baseUnitAmount);
 
             // not necessary to ship max, as most StatsD agents calculate this themselves
-            max.getCurrent().add(Math.max(msAmount - max.getCurrent().doubleValue(), 0));
+            max.getCurrent().add(Math.max(baseUnitAmount - max.getCurrent().doubleValue(), 0));
 
-            sink.next(lineBuilder.timing(msAmount));
+            sink.next(lineBuilder.timing(baseUnitAmount));
         }
     }
 
@@ -72,7 +72,7 @@ public class StatsdTimer extends AbstractTimer {
 
     @Override
     public double totalTime(TimeUnit unit) {
-        return TimeUtils.convert(totalTime.doubleValue(), TimeUnit.MILLISECONDS, unit);
+        return TimeUtils.convert(totalTime.doubleValue(), baseTimeUnit(), unit);
     }
 
     /**
@@ -82,7 +82,7 @@ public class StatsdTimer extends AbstractTimer {
      */
     @Override
     public double max(TimeUnit unit) {
-        return TimeUtils.convert(max.poll(), TimeUnit.MILLISECONDS, unit);
+        return TimeUtils.convert(max.poll(), baseTimeUnit(), unit);
     }
 
     void shutdown() {
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
@@ -258,7 +258,7 @@ public abstract class AbstractTimer extends AbstractMeter implements Timer {
     @Override
     public final void record(long amount, TimeUnit unit) {
         if (amount >= 0) {
-            histogram.recordLong(TimeUnit.NANOSECONDS.convert(amount, unit));
+            histogram.recordLong(baseTimeUnit().convert(amount, unit));
             recordNonNegative(amount, unit);
 
             if (intervalEstimator != null) {


jonatan-ivanov:
I think there might be a misunderstanding since `Timer`s does not really have a time unit from user's perspective.
(They have a unit the hood: always nanoseconds but that is not available for the users, they don't need to know about this internal detail).

The expectation is that a recorded value is converted to the unit of the registry before publication. When you record a value using a timer and a unit, the unit you used for the recording has nothing to do with the unit used for publishing. (Btw we don't recommend measuring and recording duration like, that, but we recommend using `Timer.Sample`, see the [docs](https://docs.micrometer.io/micrometer/reference/concepts/timers.html#_storing_start_state_in_timer_sample).)

This means that if the unit of the registry is milliseconds, it does not matter if you record in seconds, millis or micros, everything will be converted and published in millis.
StatsD happens to use milliseconds by default: https://github.com/micrometer-metrics/micrometer/blob/4d9efae6b7f3635be5e99fa089eb2851f9d9ca78/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java#L464-L467

So using your example, if you do this:
```java
timer.record(2_000, TimeUnit.SECONDS);  // about 30 min
```
It will be (recorded as nanos under the hood and) published in the unit of the registry. Since StatsD uses milliseconds as the unit, you will see `2,000,000ms` when published.

This is the expected behavior which is also guarded by tests and that's why that your change in `AbstractTimer` fails the tests. If you want to override the time unit of a registry, you can override the `getBaseTimeUnit` method I linked above but based on your diff it seems that millis is hardcoded in a few places in case of StatsD which we can fix but is there any backend that can use StatsD with a non-milliseconds unit?

coreyoconnor:
Starting from the end as that will explain things:

> If you want to override the time unit of a registry, you can override the getBaseTimeUnit method I linked above but based on your diff it seems that millis is hardcoded in a few places in case of StatsD which we can fix

indeed! I am using with `getBaseTimeUnit` overriden in the registery. Which I would expect to override (ha!) the time unit used. As you can tell: no such luck.

But what backend is this for? Datadog

There are several other issues where users have encountered the same unexpected behavior from datadog. I can pull these up later. 

Specifically: Datadog does *not* support units for timers. Timers are unit-less gauges. 

So, setting the units has no effect - datadog is going to accept them as unit-less and presumes you've pre-scaled them to whatever you want *or* manually changed the metric config in the UI.

Obviously, changing the metric config in datadog for every timer after they've been submitted is a non-starter.

So what to do? Seems to me like overriding the unit should do what is documented as: Overrides the unit used. "The base time unit of the timer to which all published metrics will be scaled". Which, if it was true, would do what is expected for datadog.

So even if there is no room for adjusting micrometer to scale the values nicely: The documentation is a bit misleading no?

jonatan-ivanov:
As I said above, it seems that "millis" is hardcoded in a few places in case of StatsD which we can fix. Please remove your change from `AbstractTimer` and replace hardcoded `TimeUnit.MILLISECONDS` to `baseTimeUnit()`. I think doing that should make you able to override the base unit. (I can help to write a tests for this use-case.)

Could you please also run `./gradlew format` and `./gradlew build` to see if your changes break the build or not? (Right now they do.)

>The documentation is a bit misleading no?

If I understand this correctly, I don't think the documentation is misleading, I think the docs are ok but we have a bug: I agree with the intention of the docs but the current behavior does not follow that. :) After fixing the issue, and you still think the docs are misleading, please open an issue/PR.

micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
```diff
@@ -258,7 +258,7 @@ public void record(Runnable f) {
     @Override
     public final void record(long amount, TimeUnit unit) {
         if (amount >= 0) {
-            histogram.recordLong(TimeUnit.NANOSECONDS.convert(amount, unit));
+            histogram.recordLong(baseTimeUnit().convert(amount, unit));
```
jonatan-ivanov, line 261: Please remove this change, this should be recorded as nanos and converted to the right unit at publish-time.

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdTimer.java
```diff
@@ -43,9 +43,9 @@ public class StatsdTimer extends AbstractTimer {
 
     StatsdTimer(Id id, StatsdLineBuilder lineBuilder, FluxSink<String> sink, Clock clock,
             DistributionStatisticConfig distributionStatisticConfig, PauseDetector pauseDetector, TimeUnit baseTimeUnit,
-            long stepMillis) {
+            long stepBaseUnits) {
```
jonatan-ivanov, line 46:
```suggestion
            long amount) {
```

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdTimer.java
```diff
@@ -43,9 +43,9 @@ public class StatsdTimer extends AbstractTimer {
...
         super(id, clock, distributionStatisticConfig, pauseDetector, baseTimeUnit, false);
-        this.max = new StepDouble(clock, stepMillis);
+        this.max = new StepDouble(clock, stepBaseUnits);
```
jonatan-ivanov, line 48:
```suggestion
        this.max = new StepDouble(clock, amount);
```

jonatan-ivanov (changes requested):
Could you please also fix `StatsdFunctionTimer` and `StatsdLongTaskTimer`?

coreyoconnor:
Yes, certainly: I will resolve the tests. As well as the other issues. Where would be good to document this aspect of datadog? At least enough to point somebody in the right direction. The statsd registry or config?

As you mentioned statsd works in milliseconds. So, Datadog is an exceptional case then? I'm not familiar with the standards but sounds like it! In that case, should this be a datadog specific config somehow? Or is the override of the base time unit sufficient to capture the exceptional aspect?


for reference, the related motivating tickets. 

These are duplicates - just adding here for completeness

- https://github.com/micrometer-metrics/micrometer/issues/2528
- https://github.com/DataDog/java-dogstatsd-client/issues/145
- https://github.com/statsd/statsd/issues/719

discussion about the the unitless of timers:

- https://github.com/DataDog/datadogpy/issues/229#issuecomment-546872068

jonatan-ivanov:
>Where would be good to document this aspect of datadog? At least enough to point somebody in the right direction. The statsd registry or config?

Mostly in the Datadog docs. Did you find these details on their doc site too? In Micrometer we can document it on our docs site and link the Datadog docs. We support Datadog in two ways:
1. StatsD: https://docs.micrometer.io/micrometer/reference/implementations/statsD.html
2. Datadog's HTTP API: https://docs.micrometer.io/micrometer/reference/implementations/datadog.html
(We need to check if the behavior is the same.)

>Datadog is an exceptional case then? I'm not familiar with the standards but sounds like it! In that case, should this be a datadog specific config somehow? Or is the override of the base time unit sufficient to capture the exceptional aspect?

I think so, see [the statsd standard for timing](https://github.com/statsd/statsd/blob/master/docs/metric_types.md#timing). I think Datadog could include unit metadata but it seems they aren't currently. I would not make a separate config option for this nor encourage the users to use anything other than `ms` but we can probably let users override the base unit by fixing the hardcoded values.

shakuzen:
@coreyoconnor if you have a chance to address the review feedback, I do think there are some changes we could merge here if you are still interested. If that gets addressed we can reopen the pull request and review again.

PR #5699, opened 2024-11-21
Do not leak OTLP types on public-facing API

jonatan-ivanov:
`AggregationTemporality#toOtlpAggregationTemporality` was not intended to be public, users should not use it. This is a breaking change.

See https://github.com/micrometer-metrics/micrometer/pull/5733

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -392,6 +392,7 @@ subprojects {
                     packageExcludes = ['io.micrometer.shaded.*', 'io.micrometer.statsd.internal']
 
                     fieldExcludes = []
+                    methodExcludes = ['io.micrometer.registry.otlp.AggregationTemporality#toOtlpAggregationTemporality(io.micrometer.registry.otlp.AggregationTemporality)']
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -338,7 +338,7 @@ subprojects {
 
             check.dependsOn("testModules")
 
-            if (!(project.name in ['micrometer-jakarta9'])) { // add projects here that do not exist in the previous minor so should be excluded from japicmp
+            if (!(project.name in ['micrometer-jakarta9', 'micrometer-registry-otlp'])) { // add projects here that do not exist in the previous minor so should be excluded from japicmp
```
shakuzen, line 341: In the `japicmp` task configuration, you can make more fine-grained excludes, e.g. a specific method.

shakuzen (commented):
Even though we didn't intend for this to be public in the first place (as far as I remember), it is probably better we do this in `main` so no one is broken upon upgrading a patch version. I think any changes that would follow this will likely not be applied to maintenance branches anyway.

PR #5700, opened 2024-11-22
Global observation registry

cfredri4:
This change adds a static ObservationRegistry.global similar to the Metrics.globalRegistry. For the same reasons, for use in places where dependency injection is not desired.

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/DelegatingObservationRegistry.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/DelegatingObservationRegistry.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2022 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.observation;
+
+/**
+ * Delegating {@link ObservationRegistry}.
+ *
+ * @author Christian Fredriksson
+ * @since 1.14.2
+ */
+class DelegatingObservationRegistry implements ObservationRegistry {
+
+    private ObservationRegistry delegate = ObservationRegistry.NOOP;
+
+    @Override
+    public Observation getCurrentObservation() {
+        return delegate.getCurrentObservation();
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
If you ask me, I'm not a big fan of the global `MeterRegistry`. In case of the `ObservationRegistry`, since
- chances are much higher that you will end up with a multiple registries if libraries start using the global instead of letting the user inject one
- it has components that were designed to be injected (`ObservationConvention`)
- it has more config options and easier to misconfigure something

I think we should not do this.
As far as I remember we discussed the possibility of creating a global but we decided not to.

Also, could you please mention examples where dependency injection is not desired?

cfredri4:
All perfectly valid objections.

The main reason for why dependency injection is not always desired is overhead. It is trivial for any application or library to register metrics (just `Metrics.xxx()`), but having to inject an observation registry everywhere observation is desired makes it less trivial. Especially for classes which are not managed by Spring.

My guess was that this had been considered in the past but wanted to raise this in case it was somehow missed. 😅

jonatan-ivanov:
What kind of "overhead" are you referring to? Is it performance, or possibly more code because you need an extra ctor param or setter?

>It is trivial for any application or library to register metrics (just Metrics.xxx()), but having to inject an observation registry everywhere observation is desired makes it less trivial.

It is also easy to make mistakes and get into trouble with configuring things or the classloader. It is also very hard to test.

Please notice that:
- In order to use dependency injection, you don't need a framework (like Spring)
- You only need a ctor parameter (or setter)
- You can always use a global in your code, you "only" need a `public static final` field (and probably a public wrapper class)

I still don't think it is a good idea to open this can of worms up and possibly encourage libraries to instrument themselves using a global registry instead of letting the users injecting one (in the past, users bumped into issues using the global meter registry).

marcingrzejszczak:
I would discourage doing global instances in the library core. User can manually create one in their code.

cfredri4:
@jonatan-ivanov 
> What kind of "overhead" are you referring to? Is it performance, or possibly more code because you need an extra ctor param or setter?

More code. When creating something to be observed, e.g. a JmsTemplate, instead of just providing "simple properties" like a URL we now also have to provide a bean (the registry). And sometimes this means propagating this bean through several layers.

> You can always use a global in your code, you "only" need a public static final field (and probably a public wrapper class)

Yes this is what we're doing now instead. And it seems so is [Reactor Netty](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/Metrics.java#L42).

I will close this. Thanks for the discussion.

cfredri4:
I just noticed that a global registry [already exists](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-observation/src/main/java/io/micrometer/observation/Observations.java). 🙈

michaldo:
> I just noticed that a global registry [already exists](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-observation/src/main/java/io/micrometer/observation/Observations.java). 🙈

I repeated exactly same mistake 🤭

Issue #5599, opened 2024-10-19
"./gradlew clean check" fails


izeye:
`./gradlew clean check` fails as follows:

```
...
* What went wrong:
Could not determine the dependencies of task ':micrometer-osgi-test:test'.
> Could not create task ':micrometer-osgi-test:testingBundle'.
   > Could not create task of type 'Bundle'.
      > Could not create an instance of type aQute.bnd.gradle.BundleTaskExtension.
         > Cannot change attributes of configuration ':micrometer-osgi-test:compileClasspath' after it has been locked for mutation
...
```

jonatan-ivanov:
Thank you for letting us know!
I was able to repro it even with simply `./gradlew check`. `./gradlew test` is ok though. It seems the order of configuring things is different depending on which task you want to execute.

shakuzen:
I updated the issue title to reflect this is only a problem with JDK 11 with #5735 having been merged.

PR #5735, opened 2024-12-07
Fix `gradle check` with JDK17+

ngocnhan-tran1996:
Related gh-5599

I run `./gradlew :micrometer-osgi-test:check` and the result is success on my local machine with OS is Windows 11. 
I think we should check `./gradlew clean check` before merging (maybe the result will be different)

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -234,6 +234,7 @@ plugin-japicmp = { module = "me.champeau.gradle:japicmp-gradle-plugin", version 
 plugin-downloadTask = { module = "de.undercouch:gradle-download-task", version = "5.6.0" }
 plugin-spotless = { module = "com.diffplug.spotless:spotless-plugin-gradle", version = "6.25.0" }
 plugin-bnd = "biz.aQute.bnd:biz.aQute.bnd.gradle:6.4.0"
+plugin-bndForJava17 = "biz.aQute.bnd:biz.aQute.bnd.gradle:7.1.0"
 
 [plugins]
 kotlin19 = { id = "org.jetbrains.kotlin.jvm", version = "1.9.25" }
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,12 @@ buildscript {
         classpath libs.plugin.japicmp
         classpath libs.plugin.downloadTask
         classpath libs.plugin.spotless
-        classpath libs.plugin.bnd
+
+        if (javaLanguageVersion.asInt() < 17) {
+            classpath libs.plugin.bnd
+        } else {
+            classpath libs.plugin.bndForJava17
+        }
 
         constraints {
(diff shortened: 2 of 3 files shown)


jonatan-ivanov:
Cannot we "just" upgrade to 7.x instead of using 6.x and 7.x simultaneously?

shakuzen:
7.x requires Java 17, I believe.

ngocnhan-tran1996:
@jonatan-ivanov 

Java 11 is incompatiable with version 7.x.
It will raise error when build step run.

If just upgrade version, I think we will wait until we have a plan not support prior java 17

shakuzen:
Just to be clear, this will fix it as long as you are running the build with JDK 17 or higher, but the build will still fail with 11. Is that right?

ngocnhan-tran1996:
I have not yet tested on java 11. Could we can check via ci/cd?

shakuzen:
> Could we can check via ci/cd?

The [CI build](https://github.com/micrometer-metrics/micrometer/blob/f3f5d766df616dd25a8790a4bae05e8d56895298/.circleci/config.yml#L38-L43) uses `./gradlew build`, which is also the [default task](https://github.com/micrometer-metrics/micrometer/blob/f3f5d766df616dd25a8790a4bae05e8d56895298/build.gradle#L439). It'd be good for us to understand what is different when running `./gradlew check` vs `./gradlew build`. When running the `build` task, `check` also gets run as far as I can tell. I really don't get why it fails when the command is `check`.

ngocnhan-tran1996:
I confirm it works on java 17

shakuzen:
Yes, it fixes it when using JDK 17+ with the build, but as far as I could tell `./gradlew check` is still broken when using JDK 11 with the build.

gradle/libs.versions.toml
```diff
@@ -234,6 +234,7 @@ plugin-japicmp = { module = "me.champeau.gradle:japicmp-gradle-plugin", version
 plugin-downloadTask = { module = "de.undercouch:gradle-download-task", version = "5.6.0" }
 plugin-spotless = { module = "com.diffplug.spotless:spotless-plugin-gradle", version = "6.25.0" }
 plugin-bnd = "biz.aQute.bnd:biz.aQute.bnd.gradle:6.4.0"
+plugin-bndForJava17 = "biz.aQute.bnd:biz.aQute.bnd.gradle:7.1.0"
```
shakuzen, line 237: Dependabot struggles with having multiple entries in the version catalog for the same artifact. We may want to consider inlining the older version where it is used and keeping only the newer version here so it gets updated by dependabot. We can do this separate from this PR as I think we have a number of other multiple entries that need to be handled.

shakuzen:
Merging to unblock the `check` command on JDK17+ (which is what I expect most people building the project are using). The original issue will remain open as it is still a problem on JDK 11.

Issue #5741, opened 2024-12-12
NPE occurs when AOP is applied to a method that returns CompletableFuture


laststem:
**Describe the bug**
Returning null from a method that returns CompletableFuture will result in an NPE in AOP.
```kotlin
@Timed
fun doSomething(): CompletableFuture<Int>? {
    return if (condition) {
        CompletableFuture.completedFuture(1)
    } else { 
        null
    }
}
```

Because AOP considers it to be non-null if it is of type CompletionStage. ([code](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java#L238))

As a result, the original method did not cause an NPE, but the AOP caused an NPE, so the calling thread was affected.

If you agree, please let me know if I can create a PR.

**Environment**

 - Micrometer version 1.14.1
 - OS: macOS
 - Java version: 17

**To Reproduce**

```kotlin
@SpringBootApplication
class SpringPlaygroundApplication(
    private val testComponent: TestComponent
) : ApplicationRunner {
    override fun run(args: ApplicationArguments?) {
        val result = testComponent.test()
        println("result: $result") // No output is produced as NPE is thrown.
        Thread.sleep(10000)
    }
}

@Configuration
class TestCfg(
    private val meterRegistry: MeterRegistry,
) {
    @Bean
    fun timedAop() = TimedAspect(meterRegistry)
}

fun main(args: Array<String>) {
    runApplication<SpringPlaygroundApplication>(*args)
}

@Component
class TestComponent {
    @Timed
    fun test(): CompletableFuture<Int>? {
         return null
    }
}

```

**Expected behavior**
AOP needs to be aware that values ​​of type CompletionStage can be null.

Also, exceptions raised in AOP should not affect the original method.

**Additional context**
Add any other context about the problem here, e.g. related issues.

shakuzen:
Thanks for the detailed report. Indeed it sounds like a bug. We'd be happy to review a pull request for a fix. It looks like `CountedAspect` and `ObservedAspect` also suffer from the same bug.

PR #5742, opened 2024-12-12
Fix completion stage returns null in aspects

laststem:
Fixes https://github.com/micrometer-metrics/micrometer/issues/5741


shakuzen (approved):
Thank you very much. This looks good. Would you please rebase it on `1.12.x` so we can merge it there and then forward to other branches, so this fix gets included in upcoming maintenance releases?

laststem:
I rebased it. Please let me know if there's anything wrong. Thanks for the review.

shakuzen:
Thank you for the quick turnaround.

Issue #3561, opened 2022-12-09
Add counter for total loaded classes


lenin-jaganathan:
**Describe the bug**
Currently, the [JVM class unloading metric ](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ClassLoaderMetrics.java#L47-L59) is being registered as a Function counter. The description says "The total number of classes unloaded since the Java virtual machine has started execution" which will not be true when this metric is used in a Step Meter Registry.

**Environment**
 - Micrometer version - Any version
 - Micrometer registry - Any Step Registry can be used
 - OS: Any
 - Java version: Any

**To Reproduce**
How to reproduce the bug:
Start the java application with the micrometer JVM binder enabled. Use any step registry to register these metrics. After one step is completed, the unloaded class becomes zero.

**Expected behavior**
jvm.classes.unloaded - should represent the total number of classes unloaded since the Java virtual machine has started execution

lenin-jaganathan:
I am not sure such behavior was created. But ideally, using the below snippet for class unloading will help solve this issue.

`Gauge.builder("jvm.classes.unloaded", classLoadingBean, ClassLoadingMXBean::getUnloadedClassCount).tags(tags)
                .description("The total number of classes unloaded since the Java virtual machine has started execution")
                .baseUnit(BaseUnits.CLASSES).register(registry);`

patpatpat123:
Upvoting this

marcingrzejszczak:
Some questions:

- `jvm.classes.loaded` is a gauge
- `jvm.classes.unloaded` is a counter not a gauge, right?

So is this issue about adding the "the total number of classes loaded" ? Are you upvoting this @patpatpat123 ?

patpatpat123:
Hello @marcingrzejszczak ,

Thank you for your input.

First of all, you are correct about the two metrics:

```
jvm_classes_loaded	        metric_type=gauge	        statistic=value
jvm_classes_unloaded	metric_type=counter	statistic=count

```

And yes, it would be great to have metrics unifying those, to allow building visuals such as a loaded vs. unloaded counter.

lenin-jaganathan:
I am voting for the below 2 things,
* Update the description of `jvm.classes.unloaded` to make the statement valid for Step registries. Currently, it says "classes loaded from JVM start" but this will not be true for a StepRegistry.
[...]

PR #5745, opened 2024-12-16
Make jvm.classes.unloaded description generic

shakuzen:
The description before could have been confusing for step-based registries where the number would be the classes unloaded since the previous step rather than the cumulative count since the JVM started.

See gh-3561


jonatan-ivanov (approved):

shakuzen:
I think GitHub was having stability issues yesterday when I tried to merge this. It looks like it is already merged, despite what the GH UI shows right now. See https://github.com/micrometer-metrics/micrometer/commit/3383e5cad5644875511b7d9c3cf38d23672fc318

PR #5746, opened 2024-12-17
Handle RuntimeException when getting/setting JMS headers

cfredri4:
Currently JMSException is handled when getting/setting JMS headers. Some JMS providers will throw RuntimeException instead of JMSException when failing to get/set JMS headers. This change adds so that RuntimeException is also handled.

For example, [Qpid JMS throws IllegalArgumentException](https://github.com/apache/qpid-jms/blob/main/qpid-jms-client/src/main/java/org/apache/qpid/jms/message/JmsMessagePropertySupport.java#L101) which leads to similar issue as #4202.

### DIFF
--- a/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsProcessObservationContext.java
+++ b/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsProcessObservationContext.java
@@ -16,8 +16,8 @@
 
 package io.micrometer.jakarta9.instrument.jms;
 
+import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
 import io.micrometer.observation.transport.ReceiverContext;
-import jakarta.jms.JMSException;
 import jakarta.jms.Message;
 
 /**
@@ -33,12 +33,15 @@ import jakarta.jms.Message;
  */
 public class JmsProcessObservationContext extends ReceiverContext<Message> {
 
+    private static final WarnThenDebugLogger logger = new WarnThenDebugLogger(JmsProcessObservationContext.class);
+
     public JmsProcessObservationContext(Message receivedMessage) {
         super((message, key) -> {
             try {
                 return message.getStringProperty(key);
             }
-            catch (JMSException exc) {
+            catch (Exception exc) {
+                logger.log("Failed to get message property.", exc);
                 return null;
             }
         });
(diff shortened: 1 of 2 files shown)


jonatan-ivanov (commented):
Thank you for the PR!

I'm thinking about the consequences of doing this: since this can happen if a propagation format that contains `-` is used, this can hide the underlying problem and propagation still won't work.

I'm wondering if we should log a warning (maybe a warning first then debug using `WarnThenDebugLogger`). What do you think?

/cc @bclozel @shakuzen

micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsProcessObservationContext.java
```diff
@@ -33,12 +34,15 @@
...
+    private static final WarnThenDebugLogger logger = new WarnThenDebugLogger(JmsProcessObservationContext.class);
+
     public JmsProcessObservationContext(Message receivedMessage) {
         super((message, key) -> {
             try {
                 return message.getStringProperty(key);
             }
-            catch (JMSException exc) {
+            catch (JMSException | RuntimeException exc) {
+                logger.log("Failed get message property.", exc);
```
jonatan-ivanov, line 45:
```suggestion
                logger.log("Failed to get message property.", exc);
```

micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsPublishObservationContext.java
```diff
@@ -33,15 +34,17 @@
...
-                if (message != null) {
+            if (message != null) {
+                try {
                     message.setStringProperty(key, value);
                 }
-            }
-            catch (JMSException exc) {
-                // ignore
+                catch (JMSException | RuntimeException exc) {
+                    logger.log("Failed set message property.", exc);
```
jonatan-ivanov, line 46:
```suggestion
                    logger.log("Failed to set message property.", exc);
```

jonatan-ivanov (changes requested):

micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsProcessObservationContext.java
```diff
@@ -33,12 +34,15 @@
...
 
+    private static final WarnThenDebugLogger logger = new WarnThenDebugLogger(JmsProcessObservationContext.class);
+
     public JmsProcessObservationContext(Message receivedMessage) {
         super((message, key) -> {
             try {
                 return message.getStringProperty(key);
             }
-            catch (JMSException exc) {
+            catch (JMSException | RuntimeException exc) {
```
jonatan-ivanov, line 44: Should we do `catch (Exception exc)`?

jonatan-ivanov:
I think we should do this against `1.13.x` but I can back-port this and modify the PR before merging.

cfredri4:
Pushed fixes. 👍

shakuzen (approved):
Looks good to me. It might help our future selves know why we're catching Exception if there's a comment with the reasoning.

PR #5750, opened 2024-12-18
Fix performance regression in MeterRegistry#remove

### LINKED ISSUE #5466: Performance regression in `MeterRegistry#remove` with many meters
**Describe the bug**

We noticed a performance regression related to #4857. Specifically, with a lot of meters, this `preFilterIdToMeterMap` map could be hundreds of thousands of elements or more, and we linearly iterate through them. 

**Environment**

 - Micrometer version [e.g. 1.7.1]
 - Micrometer registry [e.g. prometheus]
 - OS: [e.g. macOS]

Micrometer 1.13.2, able to produce on macOS Sonoma 14.2.

This was discovered after it was bumped in ActiveMQ Artemis.

```
openjdk 17.0.11 2024-04-16 LTS
OpenJDK Runtime Environment Zulu17.50+19-CA (build 17.0.11+9-LTS)
OpenJDK 64-Bit Server VM Zulu17.50+19-CA (build 17.0.11+9-LTS, mixed mode, sharing)
```
**To Reproduce**
How to reproduce the bug:

Unfortunately don't have a MRE to share, however this can likely be seen if creating many meters and gauges and then trying to delete them.

**Expected behavior**

**Additional context**
Add any other context about the problem here, e.g. related issues.

shakuzen:
Adds a reverse look-up for the pre-filter meter ID for use when removing a Meter. This avoids the need to iterate over the meters in the cache (preFilterIdMeterMap), which scales linearly with the number of meters, and is problematic because it does this while holding the meterMap lock needed to add new meters. Also adds benchmarks for measuring the performance of the remove method with different numbers of meters registered.

Resolves #5466

### DIFF
--- a/benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/MeterRemovalBenchmark.java
+++ b/benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/MeterRemovalBenchmark.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2024 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.benchmark.core;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.profile.GCProfiler;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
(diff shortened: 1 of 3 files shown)


mauhiz (approved):

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/MeterRemovalBenchmark.java
```diff
@@ -0,0 +1,65 @@
...
+    public void setup() {
+        for (int i = 0; i < meterCount; i++) {
+            registry.counter("counter", "key", String.valueOf(i));
+        }
+    }
+
+    @Benchmark
+    @Warmup(iterations = 100)
+    @Measurement(iterations = 500)
+    @BenchmarkMode(Mode.SingleShotTime)
```
jonatan-ivanov, line 60: Should we add a line that unlike `Sample` this measures this calls the method once and also measures the "cold" performance?

jonatan-ivanov (approved):

jonatan-ivanov:
I guess since we "only" doubling the references and not the objects, the memory impact is not too bad.

shakuzen:
> I guess since we "only" doubling the references and not the objects, the memory impact is not too bad.

That is my expectation, but we don't track/test this now. If anyone has good ideas on doing that, please share.

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/MeterRemovalBenchmark.java
shakuzen: It's not necessarily the cold performance we care about, but by the mutable nature of a `remove` method, we want to measure the performance of removing a meter that exists in the registry when the registry has a certain number of total meters, and I didn't know a good way to do that other than single shot mode. In other modes, we would either be measuring removing a meter from a registry with progressively less meters in total, or we would be measuring the time to remove a non-existent meter, which isn't what we want. I've added reasoning in a JavaDoc comment now.

PR #5759, opened 2024-12-27
Polish

izeye:
This PR polishes a bit.


micrometer-core/src/test/java/io/micrometer/core/aop/CountedAspectTest.java
```diff
@@ -60,7 +60,7 @@ void countedWithSuccessfulMetrics() {
 
         Counter counter = meterRegistry.get("metric.success")
             .tag("method", "succeedWithMetrics")
-            .tag("class", getClass().getName() + "$CountedService")
+            .tag("class", CountedService.class.getName())
```
shakuzen, line 63: I don't remember exactly, but I think the reason I didn't do this before is because we moved this test to another module and it did not move cleanly with this. It was trying to reference the class from the other module or something. But it's already moved, so I guess there's no harm in making this change now.

PR #5760, opened 2024-12-29
Polish gh-5751

izeye:
This PR polishes the changes made in gh-5751 a bit.

### DIFF
--- a/docs/modules/ROOT/pages/implementations/prometheus.adoc
+++ b/docs/modules/ROOT/pages/implementations/prometheus.adoc
@@ -133,26 +133,25 @@ image::implementations/prometheus-long-task-timer.png[Grafana-rendered Prometheu
 == Limitation on same name with different set of tag keys
 
 The `PrometheusMeterRegistry` doesn't allow to create meters having the same name with a different set of tag keys, so you should guarantee that meters having the same name have the same set of tag keys.
-Otherwise, subsequent meters having the same name with a different set of tag keys will not be registered. This means that you should not do things like:
+Otherwise, subsequent meters having the same name with a different set of tag keys will not be registered silently by default.
+This means that you should not do things like:
+
 [source,java]
 ----
 // Please don't do this
 registry.counter("test", "first", "1").increment();
 registry.counter("test", "second", "2").increment();
 ----
-This will result in the following warning and the second `Meter` will not be registered:
-[source]
-----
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/implementations/prometheus.adoc
```diff
@@ -152,7 +152,7 @@ registry.counter("test", "first", "1", "second", "none").increment();
 registry.counter("test", "first", "none", "second", "2").increment();
 ----
 
-You can change the default warning behavior by registering a meter registration failed listener.
+In addition to the warning, you can register a meter registration failed listener to handle the failure.
```
shakuzen, line 155: The WarnThenDebugLogger was added with https://github.com/micrometer-metrics/micrometer/pull/5228, right? Before 1.14.0, the log message won't be there unless a user explicitly registers a failed listener that logs. /cc @jonatan-ivanov

izeye: @shakuzen Ah, you're right. So https://github.com/micrometer-metrics/micrometer/pull/5751 should have been done against 1.14.x, right?

shakuzen: The part about the log message, yes. The rest is still good to document before, I think.

jonatan-ivanov: Good catch, what do you think about doing something like this here (removing the warning and restoring the original line about the listener):

```markdown
Otherwise, subsequent meters having the same name with a different set of tag keys will not be registered silently by default.
This means that you should not do things like:
[source,java]
----
// Please don't do this
registry.counter("test", "first", "1").increment();
registry.counter("test", "second", "2").increment();
----
Instead you can do something like this:
[source,java]
----
registry.counter("test", "first", "1", "second", "none").increment();
registry.counter("test", "first", "none", "second", "2").increment();
----
You can change the default behavior by registering a meter registration failed listener.
```

and during forward merge, in 1.14.x, we keep the current docs but we do the change currently in this PR so that the last sentence will be:
```markdown
In addition to the warning, you can register a meter registration failed listener to handle the failure.
```

izeye: @shakuzen @jonatan-ivanov Thanks for the feedback!

I updated this PR as @jonatan-ivanov suggested.

jonatan-ivanov (approved):

shakuzen (approved):

PR #5791, opened 2025-01-02
Add a test verifying that a single character meter name is allowed for Prometheus

izeye:
This PR adds a test verifying that a single character meter name is allowed for Prometheus.

There seems to have been a regression that https://github.com/prometheus/client_java/pull/951 has fixed in 1.3.0 or later, and Micrometer 1.13.x uses Prometheus Java client 1.2.1 that doesn't contain the fix, so the test that this PR adds fails.

I'm not sure if it's worth upgrading to Prometheus Java client 1.3.0 or later to fix it as it's a minor upgrade that might cause other compatibility issues.

The migration guide for Micrometer 1.13 contains [an entry for it](https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide#single-character-names-are-no-longer-valid), and it sounds like a feature, so at least we might want to update the guide to rephrase it.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
+++ b/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
@@ -1016,6 +1016,12 @@ class PrometheusMeterRegistryTest {
         assertThat(convention.tagKeyCount.get()).isEqualTo(expectedTagKeyCount);
     }
 
+    @Test
+    void scrapeWhenMeterNameContainsSingleCharacter() {
+        registry.counter("c").increment();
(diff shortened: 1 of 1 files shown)


shakuzen (approved):

shakuzen:
> I'm not sure if it's worth upgrading to Prometheus Java client 1.3.0 or later to fix it as it's a minor upgrade that might cause other compatibility issues.

I think it'll be best for us to leave it up to users to override the version of the Prometheus client to the one they want if they are affected by this particular issue and can't upgrade from Micrometer 1.13.x.

jonatan-ivanov:
Thank you, I fixed the migration guide: https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide/_compare/362088c45addf409808a5a293aec344f0add1f1b...b94c71bc9bb7f1ff6b6440b03c331b0aab29899b

Issue #5756, opened 2024-12-23
Rebind `Log4j2Metrics` when `LoggerContext#reconfigure` is called


ddavidzhang:
**Describe the bug**
Now，log4j2Metrics is registed when server starting up. But when Log4j2 LoggerContexted is reconfigured because of config being changed, the filter will not work. 

**Environment**
 - Micrometer version : 1.9.15 and 1.12.3
 - Micrometer registry : prometheus
 - OS: ALL
 - Java version: jdk 8 and jdk 21

**To Reproduce**
How to reproduce the bug:
After a spring server starts up, we change some log4j2's config, logging level for example and Log4j2's reconfigure will be triggered. After this, the filter in log4j2Metrics will not work.

**Expected behavior**
log4j2Metrics always work

shakuzen:
I'm not familiar with how the details of how reconfiguring works. Could you share a snippet of code demonstrating what is being done? Ideally if it can be put in a unit test that is minimal (i.e. doesn't involve Spring), we can add it to our tests when we address this and ensure it works now and going forward. I wonder if there is a way we can register a callback that is run when reconfiguration happens.

ddavidzhang:
I just change some config (any) in Nacos, and that will trigger a RefreshEvent of Spring.   Then, the logging system of Spring will load loadConfiguration again. 
 I'm also not familiar with how the details of how reconfiguring works. 
The "LoggerContext.setConfiguration" is called, or "LoggerContext.reconfigure" is called.
The following are the stack:

setConfiguration:601, LoggerContext (org.apache.logging.log4j.core)
start:285, LoggerContext (org.apache.logging.log4j.core)
start:87, AsyncLoggerContext (org.apache.logging.log4j.core.async)
loadConfiguration:228, Log4J2LoggingSystem (org.springframework.boot.logging.log4j2)
loadConfiguration:194, Log4J2LoggingSystem (org.springframework.boot.logging.log4j2)
initializeWithSpecificConfig:66, AbstractLoggingSystem (org.springframework.boot.logging)
initialize:57, AbstractLoggingSystem (org.springframework.boot.logging)
initialize:170, Log4J2LoggingSystem (org.springframework.boot.logging.log4j2)
reinitializeLoggingSystem:168, PropertySourceBootstrapConfiguration (org.springframework.cloud.bootstrap.config)
doInitialize:148, PropertySourceBootstrapConfiguration (org.springframework.cloud.bootstrap.config)
[...]

PR #5810, opened 2025-01-17
Rebind log4j2 metrics if configuration is changed

pativa:
*Describe the issue*

This PR is intended to fix https://github.com/micrometer-metrics/micrometer/issues/5756

When log4j2 metrics are reconfigured, metrics will no longer be recorded. For example, if `monitorInterval` ([docs](https://logging.apache.org/log4j/2.x/manual/configuration.html#configuration-attribute-monitorInterval)) is configured to 30 is used and configuration is reloaded after application start, metrics would only be recorded for the first 30 seconds.

---

By using a `PropertyChangeListener`, it looks like we can register the `MetricsFilter` in much the same way, but it will be rebound after reloading the configuration. All the previous tests seem to pass too when using this method of configuration.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
@@ -89,7 +89,9 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
         PropertyChangeListener changeListener = listener -> {
             if (listener.getNewValue() instanceof Configuration && listener.getOldValue() != listener.getNewValue()) {
-                registerMetricsFilter((Configuration) listener.getNewValue(), registry);
+                Configuration newConfiguration = (Configuration) listener.getNewValue();
+                registerMetricsFilter(newConfiguration, registry);
+                loggerContext.updateLoggers(newConfiguration);
                 if (listener.getOldValue() instanceof Configuration) {
                     removeMetricsFilter((Configuration) listener.getOldValue());
                 }
@@ -139,6 +141,7 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
     @Override
     public void close() {
         changeListeners.forEach(loggerContext::removePropertyChangeListener);
+        changeListeners.clear();
 
         if (!metricsFilters.isEmpty()) {
             Configuration configuration = loggerContext.getConfiguration();
@@ -146,6 +149,7 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
             loggerContext.updateLoggers(configuration);
 
             metricsFilters.values().forEach(MetricsFilter::stop);
+            metricsFilters.clear();
         }
     }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
@@ -31,9 +31,12 @@ import org.apache.logging.log4j.core.config.LoggerConfig;
 import org.apache.logging.log4j.core.filter.AbstractFilter;
 import org.apache.logging.log4j.core.filter.CompositeFilter;
 
-import java.util.*;
+import java.beans.PropertyChangeListener;
+import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import static java.util.Collections.emptyList;
 
@@ -63,6 +66,8 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
     private final ConcurrentMap<MeterRegistry, MetricsFilter> metricsFilters = new ConcurrentHashMap<>();
 
+    private final List<PropertyChangeListener> changeListeners = new CopyOnWriteArrayList<>();
+
     public Log4j2Metrics() {
         this(emptyList());
     }
@@ -79,12 +84,28 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
     @Override
     public void bindTo(MeterRegistry registry) {
         Configuration configuration = loggerContext.getConfiguration();
+        registerMetricsFilter(configuration, registry);
+        loggerContext.updateLoggers(configuration);
+
+        PropertyChangeListener changeListener = listener -> {
+            if (listener.getNewValue() instanceof Configuration && listener.getOldValue() != listener.getNewValue()) {
+                registerMetricsFilter((Configuration) listener.getNewValue(), registry);
+                if (listener.getOldValue() instanceof Configuration) {
+                    removeMetricsFilter((Configuration) listener.getOldValue());
+                }
+            }
+        };
+
+        changeListeners.add(changeListener);
+        loggerContext.addPropertyChangeListener(changeListener);
+    }
+
+    private void registerMetricsFilter(Configuration configuration, MeterRegistry registry) {
         LoggerConfig rootLoggerConfig = configuration.getRootLogger();
         MetricsFilter metricsFilter = getOrCreateMetricsFilterAndStart(registry);
         rootLoggerConfig.addFilter(metricsFilter);
 
-        loggerContext.getConfiguration()
-            .getLoggers()
+        configuration.getLoggers()
             .values()
             .stream()
             .filter(loggerConfig -> !loggerConfig.isAdditive())
@@ -105,8 +126,6 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
                 loggerConfig.addFilter(metricsFilter);
             });
-
-        loggerContext.updateLoggers(configuration);
     }
 
     private MetricsFilter getOrCreateMetricsFilterAndStart(MeterRegistry registry) {
@@ -119,26 +138,30 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
     @Override
     public void close() {
+        changeListeners.forEach(loggerContext::removePropertyChangeListener);
+
         if (!metricsFilters.isEmpty()) {
             Configuration configuration = loggerContext.getConfiguration();
-            LoggerConfig rootLoggerConfig = configuration.getRootLogger();
-            metricsFilters.values().forEach(rootLoggerConfig::removeFilter);
-
-            loggerContext.getConfiguration()
-                .getLoggers()
-                .values()
-                .stream()
-                .filter(loggerConfig -> !loggerConfig.isAdditive())
-                .forEach(loggerConfig -> {
-                    if (loggerConfig != rootLoggerConfig) {
-                        metricsFilters.values().forEach(loggerConfig::removeFilter);
-                    }
-                });
-
+            removeMetricsFilter(configuration);
             loggerContext.updateLoggers(configuration);
+
             metricsFilters.values().forEach(MetricsFilter::stop);
-            metricsFilters.clear();
         }
+    }
+
+    private void removeMetricsFilter(Configuration configuration) {
+        LoggerConfig rootLoggerConfig = configuration.getRootLogger();
+        metricsFilters.values().forEach(rootLoggerConfig::removeFilter);
+
+        configuration.getLoggers()
+            .values()
+            .stream()
+            .filter(loggerConfig -> !loggerConfig.isAdditive())
+            .forEach(loggerConfig -> {
+                if (loggerConfig != rootLoggerConfig) {
+                    metricsFilters.values().forEach(rootLoggerConfig::removeFilter);
+                }
+            });
     }
 
     @NonNullApi
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
@@ -17,6 +17,8 @@ package io.micrometer.core.instrument.binder.logging;
 
 import io.micrometer.common.lang.NonNullApi;
 import io.micrometer.common.lang.NonNullFields;
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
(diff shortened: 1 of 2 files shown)


shakuzen (commented):
Thank you for the quick turnaround time on submitting a pull request for this. It generally looks good to me. Will metrics filters still be configured on the old Configuration instance? Should that be the case? Does that not result in a memory leak of sorts?

pativa:
> Thank you for the quick turnaround time on submitting a pull request for this. It generally looks good to me. Will metrics filters still be configured on the old Configuration instance? Should that be the case? Does that not result in a memory leak of sorts?

Thanks for the review!

That is a good point! I don't think it would be any big issue, as the old configuration is already stopped here, but I do think it would be cleaner to clean up any old references to `MetricsFilter` when we rebind it.

I made a new commit here where I also do a cleanup of the old configuration instance. I also changed the changeListener to only do an update if the configuration instance is not the same object as the old instance (as no rebinding or cleanup should be needed then).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
```diff
@@ -61,7 +64,9 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
     private final LoggerContext loggerContext;
 
-    private final List<MetricsFilter> metricsFilters = new ArrayList<>();
+    private final Map<String, MetricsFilter> metricsFilters = new HashMap<>();
```
shakuzen, line 67: I don't want to block progress on this pull request, but in reviewing it, I've noticed something I think should be fixed perhaps first. We used to just have a single `MetricsFilter` instance. That was changed in https://github.com/micrometer-metrics/micrometer/pull/2183 when `MetricsFilter` was changed to take a `LoggerConfig`. That was subsequently changed in https://github.com/micrometer-metrics/micrometer/commit/66f59e175022fe84f5ef699f4d9285e4096270a7 when the reason it was needed was fixed in log4j itself. I think we should have at that time changed back to using a single instance of `MetricsFilter` since its input is always the same for a given instance of `Log4j2Metrics` - the `MeterRegistry` and tags. We're needlessly making multiple instances of equivalent `MetricsFilter`s. Let me know if I'm missing something. I'll prepare a change for maintenance branches to stop making extraneous instances of `MetricsFilter`. If we get that merged, you can rebase your pull request on top of it and it will hopefully simplify things a bit.
/cc @jonatan-ivanov

shakuzen: The case where it wouldn't always be the same `MeterRegistry` is if the same instance of `Log4j2Metrics` is bound to different registries. This came up before recently on whether that's expected/should be supported in all cases. I'm not sure we came to a conclusion as it depends on the implementation of the `MeterBinder`. We'll have to discuss this.

pativa: I was wondering about why multiple `MetricsFilter` were needed, but as it was that way I left that implementation. But it would be considerably easier to handle this without having multiple `MetricsFilter`, yes (no reason to store it in a map and keeping track of which filter was used for which logger).

If you change it to having a single filter, I can rebase this branch onto that and simplify this implementation a bit.

Having a single `MetricsFilter` wouldn't support binding the same `Log4j2Metrics` to different registries though, as you say.

shakuzen: We discussed this. While it would simplify things more for us to only have one `MetricsFilter`, I think it would be better for compatibility and abiding by the apparent contract of `MeterBinder` to retain support here for binding to multiple registries. That said, we only need one `MetricsFilter` per bound registry. I've opened https://github.com/micrometer-metrics/micrometer/pull/5818 to apply that fix to maintenance branches. I'll merge it once the build passes and then you can rebase your PR on top of the changes once they are merged into `main`.

pativa:
Kind of unrelated to this PR, so I haven't changed it, but the call to

```java
loggerContext.updateLoggers(configuration);
```

Seems to be completely unnecessary. I don't know how expensive this operation is, but it seems to be unnecessary to trigger a reload as this class modifies the configuration in-place?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
pativa: I'll have an updated PR ready soon!

pativa: I have rebased it on `main` branch now. The diff is much smaller in this version when we only have one `MetricsFilter` to keep track of for each `MeterRegistry`! 👍

jonatan-ivanov:
`updateLoggers` was removed in the past and was quickly restored it, see:
- https://github.com/micrometer-metrics/micrometer/pull/867
- https://github.com/micrometer-metrics/micrometer/pull/872

Basically it does two things:
1. It updates the internal "copy" of the config, since references are used in some cases, this might not be an issue but things can change (in Log4J2 or in Micrometer) so it's safer to call `updateLoggers` too
2. It emits a `PropertyChangeEvent` to notify components about the change

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
```diff
@@ -79,11 +84,27 @@ public Log4j2Metrics(Iterable<Tag> tags, LoggerContext loggerContext) {
     @Override
     public void bindTo(MeterRegistry registry) {
         Configuration configuration = loggerContext.getConfiguration();
+        registerMetricsFilter(configuration, registry);
+        loggerContext.updateLoggers(configuration);
+
+        PropertyChangeListener changeListener = listener -> {
+            if (listener.getNewValue() instanceof Configuration && listener.getOldValue() != listener.getNewValue()) {
+                registerMetricsFilter((Configuration) listener.getNewValue(), registry);
```
jonatan-ivanov, line 92: I guess after doing this, we should call `updateLoggers` but that's another change event so we have an infinite loop. Can we break that loop another way and still call `updateLoggers` in `registerMetricsFilter` and in `removeMetricsFilter`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
```diff
@@ -118,28 +137,32 @@ private MetricsFilter getOrCreateMetricsFilterAndStart(MeterRegistry registry) {
...
-                    if (loggerConfig != rootLoggerConfig) {
-                        metricsFilters.values().forEach(loggerConfig::removeFilter);
-                    }
-                });
-
+            removeMetricsFilter(configuration);
             loggerContext.updateLoggers(configuration);
+
             metricsFilters.values().forEach(MetricsFilter::stop);
-            metricsFilters.clear();
```
jonatan-ivanov, line 139: Should not we keep this and clear `metricsFilters` and also `changeListeners`?

pativa: Ah, of course, that was a mistake 😓 I'll clear both of these out!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
pativa: It's actually no problem as we only react on changes where `listener.getOldValue() != listener.getNewValue()`, so triggering `updateLoggers` with the same configuration will not register a new metrics filter (as it should already be registered when its not a new `Configuration` instance).

So I'll just add a call to `updateLoggers`!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
```diff
@@ -118,28 +139,34 @@ private MetricsFilter getOrCreateMetricsFilterAndStart(MeterRegistry registry) {
...
+        LoggerConfig rootLoggerConfig = configuration.getRootLogger();
+        metricsFilters.values().forEach(rootLoggerConfig::removeFilter);
+
+        configuration.getLoggers()
+            .values()
+            .stream()
+            .filter(loggerConfig -> !loggerConfig.isAdditive())
+            .forEach(loggerConfig -> {
+                if (loggerConfig != rootLoggerConfig) {
+                    metricsFilters.values().forEach(rootLoggerConfig::removeFilter);
```
shakuzen, line 165: Shouldn't this be `loggerConfig::removeFilter`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
```diff
@@ -79,11 +84,29 @@ public Log4j2Metrics(Iterable<Tag> tags, LoggerContext loggerContext) {
...
+                Configuration newConfiguration = (Configuration) listener.getNewValue();
+                registerMetricsFilter(newConfiguration, registry);
+                loggerContext.updateLoggers(newConfiguration);
+                if (listener.getOldValue() instanceof Configuration) {
+                    removeMetricsFilter((Configuration) listener.getOldValue());
+                }
+            }
+        };
+
+        changeListeners.add(changeListener);
```
shakuzen, line 101: I'm curious about the choice of `changeListeners` being a `List`. Would it be better to use a `ConcurrentMap` keyed on the `MeterRegistry` passed to `bindTo` like the `MetricsFilter` map? In most expected usage, it won't matter either way, but maybe it's safer and more consistent to have only one listener per registry unless I'm missing some reason that may not be what we want.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
```diff
@@ -229,4 +229,56 @@ void multipleRegistriesCanBeBound() {
...
+            logger.error("first");
+        }
+
+        // This will reload the configuration to default
+        context.reconfigure();
+
+        // This event should not be counted as the metrics binder is already closed
+        logger.error("second");
+
+        assertThat(registry.get("log4j2.events").tags("level", "error").counter().count()).isEqualTo(1);
```
shakuzen, line 281: Let's copy this assert after the `first` log so we confirm it was counted then. As things are now it's possible the first wasn't counted and the second was and the assertion will still pass.

shakuzen (approved):
Thank you for all the quick turnaround on reviews and sorry for all the back-and-forth. Things look good to me overall. I left a couple more minor review comments.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
pativa: Ok, so this will be kind of a weird case, but using a `ConcurrentMap` the following test (where we bind the `Log42Metrics` instance to the same `MeterRegistry` multiple times) would fail:

```java
    @Test
    void bindingTwice() {
        LoggerContext context = new LoggerContext("test");
        Logger logger = context.getLogger("com.test");

        try (Log4j2Metrics metrics = new Log4j2Metrics(emptyList(), context)) {
            // binding twice
            metrics.bindTo(registry);
            metrics.bindTo(registry);

            logger.error("first");

            assertThat(registry.get("log4j2.events").tags("level", "error").counter().count()).isEqualTo(2);

            context.reconfigure();

            logger.error("second");
            // this succeeds, i was thinking this might fail, but the next check does not
            assertThat(registry.get("log4j2.events").tags("level", "error").counter().count()).isEqualTo(4);
        }

        context.reconfigure();

        logger.error("third");
        // this fails, as the reconfigure() hasn't properly removed the listeners and the events are equal to 5
        assertThat(registry.get("log4j2.events").tags("level", "error").counter().count()).isEqualTo(4);
    }
```

If we use a 
```java
private final ConcurrentMap<MeterRegistry, PropertyChangeListener> changeListeners = new ConcurrentHashMap<>();
```

the following may not be enough to clean the `loggerContext` anymore:

```java
changeListeners.values().forEach(loggerContext::removePropertyChangeListener);
```

As the same changeListener may have been registered multiple times, and only one would be removed by log4j2.

Using the current setup with a `List` does not fail the test.

shakuzen: While we don't document it anywhere as far as I know, binding multiple times to the same registry I think we should generally classify the behavior as unspecified, meaning no guarantee what the outcome will be for any given binder implementation. I'm not sure why someone would intentionally bind multiple times to the same registry, and I think we can wait to hear from a user before we go to the effort of trying to support that. Thus I think doing nothing in particular would be an option here, but if we want to be extra cautious we can decide what the behavior should be in this case. I would go with ignoring the second call to `bindTo` and logging a warning if we want to handle this case at all. Pinging @jonatan-ivanov to see what he thinks on this.

pativa: Yeah, I would be fine with any handling for the case that `bindTo` is called twice, it's just important that the cleanup works regardless.

Ignoring the second call / logging a warning would work fine, as long as we don't register the `PropertyChangeListener` twice (as the second listener also has to be removed).

shakuzen: I've rebased on `main` and added a commit to check for binding to the same registry multiple times. It will log a warning and do nothing now. I think we can go ahead and merge this now. Let me know if anyone has any remaining concerns.

shakuzen (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
pativa: Thanks, it looks good to me now!

shakuzen:
@pativa Thank you for the great work on this PR and your patience in getting it reviewed and merged through all the side tasks that came out of it. This instrumentation is now better for everyone thanks to your contributions.

Issue #5812, opened 2025-01-20
Remove special handling of 404 status in HTTP client instrumentations


shakuzen:
This is a tracking issue for all the specific instrumentations that need to be updated. Existing HTTP client instrumentations should be updated to remove the special handling of 404 status responses. This has come up a few times in issues such as https://github.com/micrometer-metrics/micrometer/issues/2410. The reasoning for the special handling of 404 and some redirection statuses makes sense for server-side instrumentation but the same reasoning does not apply to client-side instrumentation where the user is in control and responsible for what requests are being made and configuring an appropriate URL mapper. The default URL mapper should provide a safe default (typically returning a constant value for the URL tag), so this change does not affect the default safety, but users who wrote URL mappers relying on the special handling (outside the URL mapper) of 404/etc statuses will need to update their implementation if that handling is what they want.

HTTP client instrumentation to update:

- [x] OkHttp https://github.com/micrometer-metrics/micrometer/pull/5814
- [x] Jetty https://github.com/micrometer-metrics/micrometer/pull/5825
- [x] Apache HTTP client https://github.com/micrometer-metrics/micrometer/pull/5837
- [x] Apache Async HTTP client https://github.com/micrometer-metrics/micrometer/pull/5837
- [x] JDK HTTP client https://github.com/micrometer-metrics/micrometer/pull/5838

izeye:
I created https://github.com/micrometer-metrics/micrometer/pull/5837 for Apache (Async) HTTP clients and https://github.com/micrometer-metrics/micrometer/pull/5838 for JDK HTTP client.

izeye:
@shakuzen https://github.com/micrometer-metrics/micrometer/pull/5837 also tried to cover the "Apache HTTP client" entry. Is there anything left for it?

shakuzen:
I don't think so. Thanks for the reminder.

shakuzen:
Actually, we could add something to `HttpClientTimingInstrumentationVerificationTests`. That fails for some deprecated instrumentation. I wonder if we should remove the tests for deprecated instrumentation since we don't plan to update this for them.

PR #5814, opened 2025-01-20
Remove special handling of 404/301 from OkHttp instrumentation

shakuzen:
See gh-5812


jonatan-ivanov (approved):

jonatan-ivanov:
I set the labels and milestone on this PR instead of the bigger tracking issue in case not every sub-item will be released within the same milestone release.

PR #5817, opened 2025-01-22
📊 Add @Gauged Annotation for Active Method Invocation Tracking

sandeepkv93:
## 🎯 Overview
Added a new @Gauged annotation and corresponding aspect to track active invocations at both method and class levels in real-time. This implementation mirrors Micrometer's @Timed aspect pattern but focuses on concurrent execution tracking.

## ✨ Features
- 🆕 New `@Gauged` annotation for method and class-level metrics
- 🔄 Automatic concurrent execution tracking
- 🧵 Thread-safe implementation with ConcurrentHashMap
- 🎨 Customizable metric names, descriptions, and tags
- 🔌 Support for CompletionStage return types
- 🎯 Programmatic skip predicates for selective monitoring
- 🏷️ Dynamic tag generation based on join points
- ⚡ Zero-overhead when skipped via predicate

### Method-Level Annotation
```java
@RestController
public class UserController {
    @Gauged("user.requests.active")  // Basic usage
    public User getUser(String id) {
        return userService.fetch(id);
    }

    @Gauged(
        value = "user.update.active",
        description = "Active user update operations",
        extraTags = {"operation", "update", "priority", "high"}
    )
    public User updateUser(User user) {  // Advanced usage
        return userService.update(user);
    }
}
```

### Class-Level Annotation
```java
@Gauged(
    value = "batch.jobs.active",
    extraTags = {"component", "batch", "tier", "processing"}
)
@Service
public class BatchProcessingService {
    // All methods automatically tracked
    public void processItem() { ... }
    public void validateItem() { ... }
}
```

## 🔧 Implementation Details

🎯 Follows established patterns from @Timed aspect
🧮 Uses AtomicInteger for thread-safe counters
🏭 Efficient gauge creation with lazy initialization
🧹 Automatic cleanup on method completion
🎭 Proper handling of both sync and async methods
🔍 Detailed exception tracking with custom tags

## 📚 References

This implementation was inspired by and follows patterns from Micrometer's:
- [@Timed annotation](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/annotation/Timed.java)
- [TimedAspect class](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java)

## 🧪 Test Coverage

🧵 Extensive concurrent execution testing
🌳 Inheritance and override scenarios
⚡ Performance impact validation
🎭 Exception handling verification
🔄 Async completion testing

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/annotation/Gauged.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/annotation/Gauged.java
@@ -1,0 +1,37 @@
+package io.micrometer.core.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotated methods are instrumented using a {@code Gauge}
+ *
+ * @author Sandeep Vishnu
+ * @see io.micrometer.core.aop.GaugedAspect
+ */
+@Target({ ElementType.ANNOTATION_TYPE, ElementType.METHOD, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Gauged {
+    /**
+     * Name of the gauge metric.
+     *
+     * @return gauge name
+     */
+    String value() default "";
+
+    /**
+     * Description of the gauge.
+     *
+     * @return gauge description
+     */
+    String description() default "";
+
+    /**
+     * Additional tags for the gauge.
+     *
+     * @return array of tag strings in the form key:value
+     */
+    String[] extraTags() default {};
+}
--- a/micrometer-core/src/main/java/io/micrometer/core/annotation/Gauged.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/annotation/Gauged.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2025 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


jonatan-ivanov:
Hi,
Thank you for the PR. You might want to take a look at [`LongTaskTimer`](https://docs.micrometer.io/micrometer/reference/concepts/long-task-timers.html). Maybe its name should be `ActiveTaskTimer` but I think its use-case is the same. We have an annotation for it: `@Timed(value = "test.active", longTask = true)`. Also, if you use the Observation API (or `@Observed`) you can also get a `LongTaskTimer` out of the box. This is the default behavior of Spring Boot, don't you have those `.active` metrics?

Also, was this PR generated using an AI tool?

sandeepkv93:
> Hi, Thank you for the PR. You might want to take a look at [`LongTaskTimer`](https://docs.micrometer.io/micrometer/reference/concepts/long-task-timers.html). Maybe its name should be `ActiveTaskTimer` but I think its use-case is the same. We have an annotation for it: `@Timed(value = "test.active", longTask = true)`. Also, if you use the Observation API (or `@Observed`) you can also get a `LongTaskTimer` out of the box. This is the default behavior of Spring Boot, don't you have those `.active` metrics?
> 
> Also, was this PR generated using an AI tool?

Hi @jonatan-ivanov, thanks for the suggestion. I experimented with `@Timed(value = "test.active", longTask = true)` and it does provide an active count :-) I can go ahead and abandon the PR since there is not much value for this PR anymore. 

This PR description & the javadoc for each functions are generated with the help of github copilot, not the code though. It is heavily influenced by `@TimedAspect` code.

jonatan-ivanov:
I'm glad it works for you, thank you anyways!
Fyi: we have https://slack.micrometer.io where you can ask questions like this (or you can open an issue) before implementing something bigger so that you don't re-implement things unnecessarily.

PR #5818, opened 2025-01-23
`Log4j2Metrics` creates more `MetricsFilter` instances than needed

shakuzen:
Before, we were needlessly making multiple instances of `MetricsFilter` that were functionally equivalent. This reduces the MetricsFilter instances to one per registry to which Log4j2Metrics is bound. This will reduce allocations and memory usage slightly.


shakuzen:
Merging before review so we don't keep blocking https://github.com/micrometer-metrics/micrometer/pull/5810, but anyone feel free to leave review post merge and we can follow-up with anything that comes up.

shakuzen:
@pativa thank you for taking a look and for providing a failing unit test. I'll take a closer look tomorrow.

PR #5822, opened 2025-01-24
Add test verifying log4j2 updateLoggers is called

pativa:
As a questioned whether the call to `updateLoggers` is necessary in https://github.com/micrometer-metrics/micrometer/pull/5810 I made a quick test verifying that its being called so it can't accidentally be removed 🙂

(made as a separate PR as this test is unrelated to the other change)

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
@@ -36,6 +36,7 @@ import org.junit.jupiter.api.Test;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static java.util.Collections.emptyList;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


shakuzen (commented):
Thanks. Could you rebase on `1.13.x` and update the target branch to there?

pativa:
Sure, it's updated to target `1.13.x` now 👍

shakuzen (approved):
I worry the test may be relying on internal details too much which could make it fragile to later changes, but I'm not sure how else to write the test. It would have caught removing `updateLoggers` before, so perhaps it's still worth adding and see if fragility becomes an issue.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
```diff
@@ -186,6 +187,27 @@ void asyncLogShouldNotBeDuplicated() throws IOException {
             .until(() -> registry.get("log4j2.events").tags("level", "info").counter().count() == 3);
     }
 
+    // see https://github.com/micrometer-metrics/micrometer/pull/872
```
jonatan-ivanov, line 190: Should we also add `@Issue("#867")`?

jonatan-ivanov (approved):

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
shakuzen: We can, though I guess 867 is linked to from 872. Obviously out of scope for this PR, but since you reminded me, I wonder if we should get rid of that `@Issue` annotation entirely since we have no tooling that uses it and so it doesn't seem to add value over a comment without an annotation.

PR #5835, opened 2025-01-29
Make Timer and LongTaskTimer output similar in LoggingMeterRegistry

jonatan-ivanov:
In case of the LoggingMeterRegistry, the output of a LongTaskTimer looks like this:
```
"my.ltt{} active=30 duration=30m"
```
while the output of a Timer is something like this:
```
"my.timer{} delta_count=30 throughput=0.5/s mean=1s max=1s"
```

We can add the missing fields to the LongTaskTimer output:
```
"my.ltt{} active=30 duration=30m mean=1m max=1m"
```

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/logging/LoggingMeterRegistry.java
@@ -150,8 +150,11 @@ public class LoggingMeterRegistry extends StepMeterRegistry {
                     int activeTasks = longTaskTimer.activeTasks();
                     if (!config.logInactive() && activeTasks == 0)
                         return;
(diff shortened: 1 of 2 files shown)


shakuzen (commented):
I'm not sure how useful/meaningful throughput is when we're talking about active (unfinished) measurements. Other parts I can understand but what does throughput mean in this context?

* `active=30` there are 30 unfinished measurements now
* `duration=30m` the sum of time taken so far for each of the measurements is 30 minutes
* `mean=1m` the average elapsed time of all the active tasks is one minute
* `max=1m` the longest active task has been running for one minute so far
* `throughput=0.5/s` ???

jonatan-ivanov:
I thought the rate you tasks are starting but looking into the calculation behind it, mathematically it does not really mean anything (active tasks/step duration), I'm removing it.

shakuzen (approved):

Issue #4868, opened 2024-03-22
Distribution value has |bucket_counts| fields that sum to 4 which does not equal the |count| field value of 2


altunkan:
**Describe the bug**
Hello,

Recently we have migrated our Spring Boot application from 2.7 to 3.2 and we've started to observe following error.

```
com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: One or more TimeSeries could not be written: Field timeSeries[31].points[0].distributionValue had an invalid value: Distribution value has |bucket_counts| fields that sum to 4 which does not equal the |count| field value of 2.
```

**Environment**

 - Micrometer version: 1.12.2
 - Micrometer registry: Stackdriver
 - OS: Linux
 - Java version: 17

**To Reproduce**
How to reproduce the bug:
```
  private Timer buildDependencyTimer(DependencyTimer dependencyTimer) {
    return Timer.builder(LATENCY_DEPENDENCY_PREFIX)
        .tags(getDependencyTags(dependencyTimer))
        .publishPercentiles(percentiles)
        .publishPercentileHistogram()
        .minimumExpectedValue(Duration.ofNanos(1L))
        .register(meterRegistry);
  }

    dependencyLatencyTimers.putIfAbsent(dependencyTimer, buildDependencyTimer(dependencyTimer));
    dependencyLatencyTimers.get(dependencyTimer).record(responseTime, TimeUnit.MILLISECONDS);
```

**Expected behavior**
Metrics should be reported to stackdriver.

**Additional context**
Add any other context about the problem here, e.g. related issues.

hudson155:
Seeing the same thing on a new project @altunkan is this new after a version upgrade? I might try older verisons...

hudson155:
I got it working by downgrading to 1.11.6.

altunkan:
Yeah, I just tried with `1.11.6` by adding `set('micrometer.version', '1.11.6')` and looks like error is gone. Thank you! This will keep us good until micrometer team addresses the issue.

rafal-dudek:
I believe this is the same issue as that: https://github.com/micrometer-metrics/micrometer/issues/998 
The problem is there since 2018 but looks like on Spring Boot 3.2 it is much more common.
We had a sporadic problem with it in our company since 2021, but after migration to Spring Boot 3.2 we have already dozens of applications experiencing it regularly.

altunkan:
[...]

PR #5836, opened 2025-01-29
Fix Stackdriver Distribution count and bucket count sum mismatch

shakuzen:
The count passed to the `HistogramSnapshot` is from the StepTimer and is step-based, while the histogram buckets have a separate time window. This mismatch in time windows caused the two to generally not match which causes a validation error when publishing Distribution metrics to Stackdriver.

This solves this by tracking an infinity bucket in the histogram so we can get a consistent count in the same time window as the rest of the histogram.

Closes gh-4868

### DIFF
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2025 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.stackdriver;
+
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.distribution.*;
+
+final class StackdriverHistogramUtil {
+
+    private StackdriverHistogramUtil() {
+    }
+
+    // copied and modified from AbstractDistributionSummary/AbstractTimer
+    static Histogram stackdriverHistogram(Clock clock, DistributionStatisticConfig distributionStatisticConfig) {
+        if (distributionStatisticConfig.isPublishingPercentiles()) {
+            return new StackdriverClientSidePercentilesHistogram(clock, distributionStatisticConfig);
+        }
+        if (distributionStatisticConfig.isPublishingHistogram()) {
+            return new StackdriverFixedBoundaryHistogram(clock, distributionStatisticConfig);
+        }
+        return NoopHistogram.INSTANCE;
+    }
+
+    static class StackdriverClientSidePercentilesHistogram extends TimeWindowPercentileHistogram {
+
+        StackdriverClientSidePercentilesHistogram(Clock clock,
+                DistributionStatisticConfig distributionStatisticConfig) {
+            super(clock, distributionStatisticConfig, true, false, true);
+        }
+
+    }
+
+    static class StackdriverFixedBoundaryHistogram extends TimeWindowFixedBoundaryHistogram {
+
+        StackdriverFixedBoundaryHistogram(Clock clock, DistributionStatisticConfig config) {
+            super(clock, config, true, false, true);
+        }
+
+    }
+
+}
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
@@ -32,9 +32,7 @@ import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
 import io.micrometer.core.instrument.distribution.HistogramSnapshot;
 import io.micrometer.core.instrument.distribution.HistogramSupport;
 import io.micrometer.core.instrument.distribution.pause.PauseDetector;
-import io.micrometer.core.instrument.step.StepDistributionSummary;
 import io.micrometer.core.instrument.step.StepMeterRegistry;
-import io.micrometer.core.instrument.step.StepTimer;
 import io.micrometer.core.instrument.util.DoubleFormat;
 import io.micrometer.core.instrument.util.NamedThreadFactory;
 import org.slf4j.Logger;
@@ -46,7 +44,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -289,15 +286,15 @@ public class StackdriverMeterRegistry extends StepMeterRegistry {
     @Override
     protected DistributionSummary newDistributionSummary(Meter.Id id,
             DistributionStatisticConfig distributionStatisticConfig, double scale) {
-        return new StepDistributionSummary(id, clock, distributionStatisticConfig, scale, config.step().toMillis(),
-                true);
+        return new StackdriverDistributionSummary(id, clock, distributionStatisticConfig, scale,
+                config.step().toMillis());
     }
 
     @Override
     protected Timer newTimer(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig,
             PauseDetector pauseDetector) {
-        return new StepTimer(id, clock, distributionStatisticConfig, pauseDetector, getBaseTimeUnit(),
-                this.config.step().toMillis(), true);
+        return new StackdriverTimer(id, clock, distributionStatisticConfig, pauseDetector, getBaseTimeUnit(),
+                this.config.step().toMillis());
     }
 
     @Override
@@ -508,18 +505,17 @@ public class StackdriverMeterRegistry extends StepMeterRegistry {
         Distribution distribution(HistogramSnapshot snapshot, boolean timeDomain) {
             CountAtBucket[] histogram = snapshot.histogramCounts();
 
-            // selected finite buckets (represented as a normal histogram)
-            AtomicLong truncatedSum = new AtomicLong();
-            AtomicReference<Double> last = new AtomicReference<>(0.0);
-            List<Long> bucketCounts = Arrays.stream(histogram).map(countAtBucket -> {
-                double cumulativeCount = countAtBucket.count();
-                long bucketCount = (long) (cumulativeCount - last.getAndSet(cumulativeCount));
-                truncatedSum.addAndGet(bucketCount);
-                return bucketCount;
-            }).collect(toCollection(ArrayList::new));
+            List<Long> bucketCounts = Arrays.stream(histogram)
+                .map(CountAtBucket::count)
+                .map(Double::longValue)
+                .collect(toCollection(ArrayList::new));
+            long cumulativeCount = Arrays.stream(histogram).mapToLong(c -> (long) c.count()).sum();
 
-            if (!bucketCounts.isEmpty()) {
-                int endIndex = bucketCounts.size() - 1;
+            // no-op histogram will have no buckets; other histograms should have at least
+            // the +Inf bucket
+            if (!bucketCounts.isEmpty() && bucketCounts.size() > 1) {
+                // the rightmost bucket should be the infinity bucket; do not trim that
+                int endIndex = bucketCounts.size() - 2;
                 // trim zero-count buckets on the right side of the domain
                 if (bucketCounts.get(endIndex) == 0) {
                     int lastNonZeroIndex = 0;
@@ -529,30 +525,41 @@ public class StackdriverMeterRegistry extends StepMeterRegistry {
                             break;
                         }
                     }
+                    long infCount = bucketCounts.get(bucketCounts.size() - 1);
                     bucketCounts = bucketCounts.subList(0, lastNonZeroIndex + 1);
+                    // infinite bucket count of 0 can be omitted
+                    bucketCounts.add(infCount);
                 }
             }
 
-            // add the "+infinity" bucket, which does NOT have a corresponding bucket
-            // boundary
-            bucketCounts.add(Math.max(0, snapshot.count() - truncatedSum.get()));
+            // no-op histogram
+            if (bucketCounts.isEmpty()) {
+                bucketCounts.add(0L);
+            }
 
             List<Double> bucketBoundaries = Arrays.stream(histogram)
                 .map(countAtBucket -> timeDomain ? countAtBucket.bucket(getBaseTimeUnit()) : countAtBucket.bucket())
                 .collect(toCollection(ArrayList::new));
 
+            if (bucketBoundaries.size() == 1) {
+                bucketBoundaries.remove(Double.POSITIVE_INFINITY);
+            }
+
+            // trim bucket boundaries to match bucket count trimming
             if (bucketBoundaries.size() != bucketCounts.size() - 1) {
                 bucketBoundaries = bucketBoundaries.subList(0, bucketCounts.size() - 1);
             }
 
-            // stackdriver requires at least one finite bucket
+            // Stackdriver requires at least one explicit bucket bound
             if (bucketBoundaries.isEmpty()) {
                 bucketBoundaries.add(0.0);
             }
 
             return Distribution.newBuilder()
+                // is the mean optional? better to not send as it is for a different time
+                // window than the histogram
                 .setMean(timeDomain ? snapshot.mean(getBaseTimeUnit()) : snapshot.mean())
-                .setCount(snapshot.count())
+                .setCount(cumulativeCount)
                 .setBucketOptions(Distribution.BucketOptions.newBuilder()
                     .setExplicitBuckets(
                             Distribution.BucketOptions.Explicit.newBuilder().addAllBounds(bucketBoundaries).build())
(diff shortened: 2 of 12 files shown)


implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
```diff
@@ -508,18 +505,17 @@ private TimeInterval interval(MetricDescriptor.MetricKind metricKind) {
         Distribution distribution(HistogramSnapshot snapshot, boolean timeDomain) {
             CountAtBucket[] histogram = snapshot.histogramCounts();
 
-            // selected finite buckets (represented as a normal histogram)
```
shakuzen, line 511: We no longer need to convert cumulative buckets to non-cumulative because with the change to dedicated Stackdriver types for Timer and DistributionSummary, we pass the flag for non-cumulative histogram.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
```diff
@@ -551,8 +553,10 @@ Distribution distribution(HistogramSnapshot snapshot, boolean timeDomain) {
             }
 
             return Distribution.newBuilder()
+                // is the mean optional? better to not send as it is for a different time
+                // window than the histogram
                 .setMean(timeDomain ? snapshot.mean(getBaseTimeUnit()) : snapshot.mean())
-                .setCount(snapshot.count())
+                .setCount(cumulativeCount)
```
shakuzen, line 559: This is the part that fixes things - `snapshot.count()` is the step-based count, which does not align with the histogram time window, while the `cumulativeCount` is taken from the histogram.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
```diff
@@ -529,18 +525,24 @@ Distribution distribution(HistogramSnapshot snapshot, boolean timeDomain) {
                             break;
                         }
                     }
+                    long infCount = bucketCounts.get(bucketCounts.size() - 1);
                     bucketCounts = bucketCounts.subList(0, lastNonZeroIndex + 1);
+                    // infinite bucket count of 0 can be omitted
+                    bucketCounts.add(infCount);
```
shakuzen, line 531: I left this note here but did not omit 0 count infinity buckets to try to keep the behavior the same as before since this is a change in a maintenance branch.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
```diff
@@ -166,25 +171,6 @@ private ValueAtPercentile[] takeValueSnapshot() {
         return values;
     }
 
-    private CountAtBucket[] takeCountSnapshot() {
-        if (!distributionStatisticConfig.isPublishingHistogram()) {
-            return null;
-        }
-
-        final Set<Double> monitoredValues = distributionStatisticConfig
-            .getHistogramBuckets(supportsAggregablePercentiles);
```
shakuzen, line 175: This logic of getting the buckets again to include in the snapshot was problematic for the goal of always having an infinity bucket (in the case of Stackdriver), since this relies only on the logic in `DistributionStatisticConfig` and we do not know at this level (`AbstractTimeWindowHistogram`) if we want to include the infinity bucket or not. Hence this was removed to avoid trying to keep the buckets in the concrete histogram type (e.g. `TimeWindowFixedBoundaryHistogram`) to match the buckets returned here when `Histogram#takeSnapshot` is called. With the new package-private method `countsAtBuckets` we let the concrete implementation determine what buckets to return rather than from here passing the buckets. We could even get rid of the `countsAtValues` method, I think.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
```diff
@@ -38,28 +38,65 @@ public class TimeWindowFixedBoundaryHistogram extends AbstractTimeWindowHistogra
...
+     * @param config distribution statistic configuration
+     * @param supportsAggregablePercentiles whether the backend supports aggregable
+     * percentiles
+     * @param isCumulativeBucketCounts whether to use cumulative bucket counts
+     * @param includeInfinityBucket whether to always include an infinity bucket
+     * @since 1.13.11
+     * @see PercentileHistogramBuckets
+     */
+    protected TimeWindowFixedBoundaryHistogram(Clock clock, DistributionStatisticConfig config,
+            boolean supportsAggregablePercentiles, boolean isCumulativeBucketCounts, boolean includeInfinityBucket) {
```
shakuzen, line 88: This `protected` constructor is the only (semi-)public API added/changed as part of this pull request. I tried to keep API changes minimal since this is in a maintenance branch. I would have liked to not add any, but I did not come up with a way to achieve that without doing something like copying everything from `TimeWindowFixedBoundaryHistogram` to an implementation of Histogram in the stackdriver module. That would probably be more prone to code rot and bugs.

shakuzen (commented):
Leaving some notes for anyone reviewing and my future self before I forget.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
jonatan-ivanov: Indeed, `Iterator<CountAtBucket> countsAtValues(Iterator<Double> values)` might be unnecessary while we are also having `CountAtBucket[] countsAtBuckets()`. I'm not sure though if there is a use-case of `countsAtValues`, where not all the bucket boundaries are passed.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
jonatan-ivanov, line 88: It's a bit more verbose but might describe the intent better: `alwaysIncludeInfinityBucket` instead of `includeInfinityBucket`?

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
```diff
@@ -38,28 +38,65 @@ public class TimeWindowFixedBoundaryHistogram extends AbstractTimeWindowHistogra
...
+     * @see PercentileHistogramBuckets
+     */
     public TimeWindowFixedBoundaryHistogram(Clock clock, DistributionStatisticConfig config,
             boolean supportsAggregablePercentiles) {
         this(clock, config, supportsAggregablePercentiles, true);
     }
 
     /**
-     * Create a {@code TimeWindowFixedBoundaryHistogram} instance.
+     * Create a {@code TimeWindowFixedBoundaryHistogram} instance with buckets based on
```
jonatan-ivanov, line 57: Should we mention here and in the docs of the previous ctor what are the defaults?
Here: inf bucket is not forced/might be missing.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
shakuzen: It's package-private, so we can check all the usage. AFAIR, it's always `distributionStatisticConfig.getHistogramBuckets` being called, until this PR deviated to add the infinity bucket. We previously had a regression where the buckets didn't match - see https://github.com/micrometer-metrics/micrometer/pull/3922.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
```diff
@@ -529,18 +525,24 @@ Distribution distribution(HistogramSnapshot snapshot, boolean timeDomain) {
...
                 }
             }
 
-            // add the "+infinity" bucket, which does NOT have a corresponding bucket
-            // boundary
-            bucketCounts.add(Math.max(0, snapshot.count() - truncatedSum.get()));
+            // Stackdriver Distribution must have at least one bucket count
+            if (bucketCounts.isEmpty()) {
```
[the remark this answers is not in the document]
shakuzen: This is for the no-op Histogram case, which still will have no buckets.

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverMeterRegistryTest.java
```diff
@@ -53,10 +56,16 @@ public String get(String key) {
...
+        ds.record(1);
+        ds.record(1);
+        ds.record(2);
+        ds.record(2);
+        ds.record(2);
         StackdriverMeterRegistry.Batch batch = meterRegistry.new Batch();
-        // count is 4, but sum of bucket counts is 5 due to inconsistent snapshotting
-        HistogramSnapshot histogramSnapshot = new HistogramSnapshot(4, 14.7, 5, null,
-                new CountAtBucket[] { new CountAtBucket(1.0, 2), new CountAtBucket(2.0, 5) }, null);
+        // count is 0 from previous step, but sum of bucket counts is 5
```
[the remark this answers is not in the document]
shakuzen: We can but we check this in other tests and I was trying to keep this test case focused on the mentioned bug fix.

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverMeterRegistryTest.java
```diff
@@ -53,10 +56,16 @@ public String get(String key) {
...
+        HistogramSnapshot histogramSnapshot = ds.takeSnapshot();
+        assertThat(histogramSnapshot.count()).isEqualTo(0);
         Distribution distribution = batch.distribution(histogramSnapshot, false);
         List<Long> bucketCountsList = distribution.getBucketCountsList();
         assertThat(bucketCountsList.get(bucketCountsList.size() - 1)).isNotNegative();
```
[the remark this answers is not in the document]
shakuzen: We can now with these changes since the infinity bucket is properly tracked - before we could not and the previous fix was only to ensure we didn't report it as less than zero which would cause a validation error on the Stackdriver side. I'm not sure how much we want to change this existing test case or not. In a lot of ways it's probably not needed anymore with the assertions done in other test cases, but I thought it might be good to keep it as a focused unit test for a previously fixed bug.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
```diff
@@ -45,7 +44,7 @@ abstract class AbstractTimeWindowHistogram<T, U> implements Histogram {
 
     private final Clock clock;
 
-    private final boolean supportsAggregablePercentiles;
+    protected final boolean supportsAggregablePercentiles;
```
shakuzen, line 47: I needed to make this `protected` for `TimeWindowPercentileHistogram` to include the logic that was previously here in `takeCountSnapshot()`

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
shakuzen: I was trying to allude to that by saying the buckets are based on the DistributionStatisticConfig and supportsAggregablePercentiles. If those result in an inf bucket, it'll be there. If not, it won't.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
shakuzen: I've removed it now as I refactored things in `TimeWindowPercentileHistogram`

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
shakuzen: If there are no monitored buckets (i.e. no SLOs and no publishPercentileHistogram) I wonder if we should add the infinity bucket. This case doesn't come up in our usage but from the perspective of this constructor/class it is possible.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
shakuzen: In the latest iteration, I was able to remove this field because it is no longer used anywhere but in constructors. This will save a whole byte per histogram potentially (ignoring object layout losses).

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowPercentileHistogram.java
```diff
@@ -37,10 +38,35 @@ public class TimeWindowPercentileHistogram extends AbstractTimeWindowHistogram<D
...
+    private final boolean isCumulativeBucketCounts;
+
     public TimeWindowPercentileHistogram(Clock clock, DistributionStatisticConfig distributionStatisticConfig,
             boolean supportsAggregablePercentiles) {
-        super(clock, distributionStatisticConfig, DoubleRecorder.class, supportsAggregablePercentiles);
+        this(clock, distributionStatisticConfig, supportsAggregablePercentiles, true, false);
+    }
+
+    public TimeWindowPercentileHistogram(Clock clock, DistributionStatisticConfig distributionStatisticConfig,
+            boolean supportsAggregablePercentiles, boolean isCumulativeBucketCounts, boolean includeInfinityBucket) {
```
shakuzen, line 51: I didn't want to add this as a public constructor but see StackdriverHistogramUtil for why I couldn't do the same thing as TimeWindowFixedBoundaryHistogram (it's the `implementation` dependency on HdrHistogram).

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
```diff
@@ -0,0 +1,57 @@
...
+        if (distributionStatisticConfig.isPublishingHistogram()) {
+            return new StackdriverFixedBoundaryHistogram(clock, distributionStatisticConfig);
+        }
+        return NoopHistogram.INSTANCE;
+    }
+
+    // Can't do this because java: cannot access org.HdrHistogram.DoubleRecorder
+    /*
+     * static class StackdriverClientSidePercentilesHistogram extends
+     * TimeWindowPercentileHistogram {
```
shakuzen, line 40: This is what I wanted to do but the compiler won't let me. Adding a dependency to HdrHistogram to the stackdriver module didn't seem like the right thing to do. I wonder if we could make it a compile only dependency and if this would work as expected at runtime or not.

shakuzen (commented):
I've got things working and tested in the StackdriverMeterRegistryTest now for client-side percentiles (TimeWindowPercentileHistogram) and the no-op histogram and TimeWindowFixedBoundaryHistogram implementations. I'm not happy with adding the public constructor to `TimeWindowPercentileHistogram`, though. I should add some more tests for it given the expanded use cases it supports.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
shakuzen: Adding HdrHistogram as a compileOnly dependency allowed this to work as expected.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
shakuzen: The behavior is actually different right now between `TimeWindowFixedBoundaryHistogram` and `TimeWindowPercentileHistogram` where the former will add the infinity bucket even if there are no other buckets but the latter will only add it when there are any buckets in the first place. We should align on one but I'm not sure which. I guess I would lean toward having the infinity bucket even without other buckets, since that does feel like what's being asked by the configuration option.

shakuzen: I updated it to always add the infinite bucket, which broke a test. I think we have an odd inconsistency now - the default `DistributionStatisticConfig` will result in a no-op histogram but configuring client-side percentiles (and no SLOs) will result in a histogram with only the infinity bucket. We should change the least amount necessary in maintenance branches. Next week I will separate out the tests and merge those first so I can confirm the behavior before the main code changes and try to align things accordingly. We can decide on further behavior changes on main instead.

shakuzen: [The documentation](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TypedValue#explicit) has the following statement on the Explicit buckets we use, for the case when there is only one bucket boundary (e.g. when we use `0` only for empty histograms):

>  If `bounds` has only one element, then there are no finite buckets, and that single element is the common boundary of the overflow and underflow buckets.

We should be using the histogram count rather than 0, but with the no-op histogram, the count is always 0. We could stop using the no-op histogram with the Stackdriver registry or we could stop sending DISTRIBUTION type metrics when the histogram is no-op - we aren't adding any info over the separate `count` metric, but maybe it's nicer to consistently get a DISTRIBUTION metric with the count info. That said, I think it's probably better to not change this for no-op histogram cases in maintenance branches. We should only make the other histogram implementations produce correct Stackdriver DISTRIBUTION metrics. I've updated the PR and tests pass. I think this should be ready to merge following review.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java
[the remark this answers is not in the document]
shakuzen: I updated the comment to make clear this is for the no-op histogram case.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
jonatan-ivanov: It is a wider API change but having an `includeInfinityBucket()` config method on `DistributionStatisticConfig` could spare the extra ctors in histograms.

jonatan-ivanov (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowFixedBoundaryHistogram.java
shakuzen: That is an interesting idea, but I'd rather avoid it in a maintenance release because the current change is more internal and if we are going to move toward making including the infinity bucket the only option on main, we'd be immediately deprecating the new public API.

PR #5840, opened 2025-01-30
Use direct equals call instead of Objects.equals wrapper

turbanoff:
We known that `name` and `tags` are non-null values. (E.g. hashCode doesn't have `null` checks). When we know that value is non-null - we can call `equals` directly - it will be a bit faster


shakuzen (approved):
Sounds reasonable to me. The build is failing due to a now unused import. Could you please fix that?

shakuzen (approved):
Sounds reasonable to me. The build is failing due to a now unused import. Could you please fix that?

shakuzen:
Thanks for the pull request. This will be included in the upcoming 1.15.0-M2 release.

Issue #5820, opened 2025-01-23
Log metrics recording failures in CountedAspect and TimedAspect


TheHett:
For Counter/Timer annotation argument `extraTags` defined as array of string, but requires even items becuase it is key->value pair. 
It's not obvious.
I spet half day before until I saw the documentation. 
Then I remembered that I had already encountered this a couple of years ago.

Maybe it's worth adding a warning to log?

jonatan-ivanov:
Thanks for the issue!
My brain is definitely spoiled here and I cannot unlearn the knowledge that tags are key+values so your input on this would be definitely helpful.

Could you please tell us:
- What you tried to do before you looked at the docs?
- What was your first impression about the usage of `extraTags`, what did you think it does?
- Why you spent half a day before looking at the docs?

I think we could improve the javadocs and add this fact to even more places but right now you can see this if you look into the docs:
https://github.com/micrometer-metrics/micrometer/blob/c39bda3dd010fe690fe7b66a77729038e283ae51/micrometer-core/src/main/java/io/micrometer/core/annotation/Counted.java#L55-L61

https://github.com/micrometer-metrics/micrometer/blob/c39bda3dd010fe690fe7b66a77729038e283ae51/micrometer-core/src/main/java/io/micrometer/core/instrument/Counter.java#L77-L84

https://github.com/micrometer-metrics/micrometer/blob/c39bda3dd010fe690fe7b66a77729038e283ae51/micrometer-core/src/main/java/io/micrometer/core/annotation/Timed.java#L46-L51

https://github.com/micrometer-metrics/micrometer/blob/c39bda3dd010fe690fe7b66a77729038e283ae51/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimerBuilder.java#L54-L61

I think we could add a sentence that the array must have an even number of elements in `Counted.java` and `Timed.java` as well but you you should get an exception if you don't:
https://github.com/micrometer-metrics/micrometer/blob/c39bda3dd010fe690fe7b66a77729038e283ae51/micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java#L362-L364

Do you want to open a PR to improve the docs of `Counted.java` and `Timed.java`?

TheHett:
Hi!

> What you tried to do before you looked at the docs?

First I tried to use single value for tags.  It not lead to error or any log, but metric was not presented in actuator.

```kotlin
    @Timed("batcher", extraTags = ["write"])
```

[...]

PR #5841, opened 2025-02-01
Log metrics recording failures in CountedAspect and TimedAspect

izeye:
This PR changes to log metrics recording failures in the `CountedAspect` and the `TimedAspect`.

This PR also prevents metrics recording failures from interfering with application flow in the `CountedAspect`.

See gh-5820

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
@@ -79,7 +79,7 @@ import java.util.function.Predicate;
 @NonNullApi
 public class CountedAspect {
 
-    private static final WarnThenDebugLogger joinPointTagsFunctionLogger = new WarnThenDebugLogger(CountedAspect.class);
+    private static final WarnThenDebugLogger WARN_THEN_DEBUG_LOGGER = new WarnThenDebugLogger(CountedAspect.class);
 
     private static final Predicate<ProceedingJoinPoint> DONT_SKIP_ANYTHING = pjp -> false;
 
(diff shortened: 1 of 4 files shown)


jonatan-ivanov:
Thanks for the PR!
I'm thinking if we should also fix this: https://github.com/micrometer-metrics/micrometer/blob/c39bda3dd010fe690fe7b66a77729038e283ae51/micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java#L362-L364

and emit a log message instead of throwing an exception.

shakuzen:
> I'm thinking if we should also fix this

Not in the same PR, I'd say. That's much more core than the aspect implementations.

micrometer-core/src/main/java/io/micrometer/core/aop/CountedAspect.java
```diff
@@ -180,7 +180,7 @@ private Function<ProceedingJoinPoint, Iterable<Tag>> makeSafe(
                 return function.apply(pjp);
             }
             catch (Throwable t) {
-                joinPointTagsFunctionLogger
+                WARN_THEN_DEBUG_LOGGER
```
shakuzen, line 183: I wonder if we want the two catch cases to share the same `WarnThenDebugLogger`. With them shared, if an application is throwing exceptions from both places only whichever one runs first will be logged at warn, masking the other. Separate loggers would ensure they both get logged at warn the first time.

shakuzen: I don't have a strong opinion either. We can leave it up to @jonatan-ivanov to decide :)

jonatan-ivanov: [pressure intensifies...] I flipped a coin, it's tails! Oh wait, I guess we did not agree what happens when it's tails. 🙂

I like the idea but I think I would probably leave it as-is (two loggers) for now because:
- (It's tails! 🙂)
- It's easier to discover the source of the log event with the fully qualified class name (unlikely but the stacktrace can be missing)
- It can alert the users about the issue with `@Timed` and `@Counted` too in one run so they can fix both and don't need to restart/redeploy the app
- Users can configure different behaviors for the two loggers (unlikely that they will)
- It does not add much overhead (one logger and one extra WARN event)

jonatan-ivanov (approved):

PR #5881, opened 2025-02-03
Cache hash code of immutable values of type Tags & KeyValues

mstyura:
`Tags` objects are often used as keys in `HashMap` causing non-negligible time consumed in repeated computation of hash codes - since objects are immutable it's possible to compute and cache hash code values. 
[Similar caching technique](https://github.com/openjdk/jdk/blob/43979fb1be424157b33268db078d345b4da29119/src/java.base/share/classes/java/lang/String.java#L2434-L2454) is used inside JDK for a `String` objects. The key difference here I've decided to handle zero hash codes in a way which only adds 1 `int` filed to and object.


micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
```diff
@@ -51,6 +51,11 @@ public final class KeyValues implements Iterable<KeyValue> {
      */
     private final int length;
 
+    /**
+     * Cache the hash code of the {@code Tags}, 0 value means hash was not computed.
```
jonatan-ivanov, line 55:
```suggestion
     * Cache the hash code of the {@code KeyValues}, 0 value means hash was not computed.
```

jonatan-ivanov (commented):
>Tags objects are often used as keys in HashMap

Could you please show us where they are used directly?

`Meter.Id` is used as a key and it uses `tags.hashCode()` but I don't think Micrometer uses `Tags` or `KeyValues` as a key directly and I don't think users should do that either. Please show us a use-case if you disagree.

Because `Meter.Id` uses `tags.hashCode()`, I think the optimization might be valid, not sure about the performance impact, maybe we should check this with our JMH tests.

I guess `KeyValue` is not used like this but if we do this with `Tags`, I think it's god to keep them in-sync.

mstyura:
I was solely looking at micrometer thru datadog profiler and seen this (initially there was slightly different filter where I've missed this relationship between `Id` and `Tags`):
image
and I've missed that fact, that `Tags.hashCode` is actually called from `Id.hashCode`.
So to fix this `1%` of total service CPU time spent id calculating hash codes it might make sense to cache the hash codes of `Id` instead of tags.
This high CPU usage might be due to the nature of service I've monitored, but I think micrometer as generic library can provide more optimal performance in such scenario.
The only unfortunate thing I see in implementing caching of hash code inside `Meter.Id` is that this class is not final and could it be the problem in case some tricky implemented ancestors?

shakuzen:
Thanks everyone for the discussion on this proposal. Sorry we haven't followed up on this more since it was opened. It was recently brought up again in #7039 by @belugabehr.

My thoughts on it are that on the surface it seems reasonable given we know Meter.Id will be hashed for lookup in HashMap, but I have concerns. First would be: is this really a performance bottleneck to begin with? When I've done profiling of sample apps under load using async-profiler, this barely shows up in the profiling, from what I remember. So I'm curious if people are seeing this come up in profiling and what's the scenario? Another concern is we're (potentially) increasing the memory footprint of core classes, of which there will be many instances. In reality, it may not increase the footprint for most users because of object alignment/padding (in most cases, we have 4 bytes to spare), so maybe that's effectively a non-issue. Another concern is this change is generally incompatible with Valhalla value types (all fields must be final), which I've been exploring lately. Caching the hash code is an internal implementation detail we could change if Tags/Id ever becomes a value type, so it's not necessarily a blocker, but in a future where Tags is a value type, we can't lazily cache the hash code so we would have to figure out another way to solve this if it is a performance issue we need to address.

I also wonder if this change actually helps. Won't the typical case be using a new Id and Tags instance to lookup a meter? That is, something like the following code is going to make a new Tags instance and pass it to a new Id instance, and that will be used as the key to lookup the existing meter in a Map, which means that will have `hashCode()` called on it and it won't be cached.

```java
Timer.builder("http.server.request").tags(Tags.of("status", status, "method", method, "outcome", outcome, "uri", uri)).register(registry);
```

What does the scenario where the same `Id` or `Tags` is being hashed many times look like? If retrieving an existing meter (should be the most common, performance-critical scenario), I would generally expect only hashing once on usually fresh instances. If the same instance of `Tags` is being used, then it's a case where dynamically retrieving the meter is probably not needed and no lookup needs to happen in the first place. I'm sure I'm missing some scenario where this caching would be helpful, so please help make sure we're talking about and benchmarking the same thing.

belugabehr:
Great questions - I'll take a look, and you are probably correct that we are doing this to ourselves. I'll have to get back to you.

Though please note that I liked my proposed implementation better FWIW

shakuzen:
> Though please note that I liked my proposed implementation better FWIW

We can discuss details of any implementation after we've established the exact nature of the problem to fix.

If we can get the details requested in my previous comment, we can be clear about the scenarios we are improving.

shakuzen:
We can reopen if we get the details to understand the scenario in which the same object's hash code is needed many times.

Issue #5901, opened 2025-02-06
Log4j2Metrics does not work with programatically added LoggerConfig


pativa:
**Describe the bug**

This is a follow-up issue to https://github.com/micrometer-metrics/micrometer/issues/5756 .
In that issue, rebinding when log4j2 `Configuration` is reloaded was fixed.

However, there is also a case when the existing instance of `Configuration` is updated with a new `LoggerConfig` programatically, without actually creating a new `Configuration` instance. This does not work due to the comparison of oldValue with newValue here: https://github.com/micrometer-metrics/micrometer/blob/04f6d8cafe0d8212ccbb72344b33023a044ec675/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java#L99

Currently the logic for rebinding will only reload if a new `Configuration` instance is loaded, so an added logger to the existing `Configuration` will not get a `MetricsFilter` added, even if `updateLoggers()` is called.

For an example of when this might be done in real code, setting log levels in the loggers endpoint of Spring Boot Actuator can create a new `LoggerConfiguration` on an existing `Configuration`.

I will provide a PR together with a test that shows when it does not work.

PR #5902, opened 2025-02-06
Handle programatically added log4j2 LoggerConfig

pativa:
The needed change for being able to bind to programmatically added loggers is simply to remove `listener.getOldValue() != listener.getNewValue()`.

However, when we do that it leads to three things:
- We need to always check if the `MetricsFilter` is already added to avoid adding it again on configuration reload
- We need some way to tell when to call `updateLoggers()` to avoid an endless loop (as the listener will be triggered again when calling `updateLoggers()`
- We need to move the check if its a new `Configuration` to the conditional before `removeMetricsFilter` to avoid removing all filters when we reload with the same `Configuration` instance

I did this by returning whether a `MetricsFilter` was added or not from `registerMetricsFilter()`. I tried out some different ways of doing this, but I ended up refactoring it a bit and I think it ended up quite readable.

I have also included a test that adds a logger programmatically (which would fail without these changes). It also verifies that `updateLoggers()` is called once with the new `MetricsFilter` configured, as expected.

Resolves https://github.com/micrometer-metrics/micrometer/issues/5901

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
@@ -368,4 +368,49 @@ class Log4j2MetricsTest {
         assertThat(registry.get("log4j2.events").tags("level", "error").counter().count()).isEqualTo(2);
     }
 
+    @Issue("#5901")
+    @Test
+    void programmaticallyAddedLoggerConfigShouldBeCounted() {
+        LoggerContext context = new LoggerContext("test");
+
+        AtomicInteger newLoggerUpdatedWithLoggerFilter = new AtomicInteger();
+
+        // checks that updateLoggers() is called after MetricsFilter is added to the
+        // programmatically added logger
+        context.addPropertyChangeListener(event -> {
+            if (event.getNewValue() instanceof Configuration) {
+                Configuration newConfiguration = (Configuration) event.getNewValue();
+                LoggerConfig loggerConfig = newConfiguration.getLoggerConfig("com.test");
+                if (!loggerConfig.isAdditive() && loggerConfig.getFilter() instanceof Log4j2Metrics.MetricsFilter) {
+                    newLoggerUpdatedWithLoggerFilter.incrementAndGet();
+                }
+            }
+        });
+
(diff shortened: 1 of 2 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
```diff
@@ -368,4 +368,49 @@ void bindingTwiceToSameRegistry_doesNotDoubleCount() {
...
+        AtomicInteger newLoggerUpdatedWithLoggerFilter = new AtomicInteger();
+
+        // checks that updateLoggers() is called after MetricsFilter is added to the
+        // programmatically added logger
+        context.addPropertyChangeListener(event -> {
+            if (event.getNewValue() instanceof Configuration) {
+                Configuration newConfiguration = (Configuration) event.getNewValue();
+                LoggerConfig loggerConfig = newConfiguration.getLoggerConfig("com.test");
+                if (!loggerConfig.isAdditive() && loggerConfig.getFilter() instanceof Log4j2Metrics.MetricsFilter) {
+                    newLoggerUpdatedWithLoggerFilter.incrementAndGet();
```
shakuzen, line 385: I wonder if we should assert on this. The only effect we need to ensure is that log events for programmatically added loggers are counted after calling `updateLoggers`, unless I'm missing something. It may be a bit too into the details of the implementation to check this. What do you think?

shakuzen (approved):
Looks good. Thank you for the complete pull request. I just had one comment about the test.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/Log4j2MetricsTest.java
pativa: Sure, I was considering whether to add this check or not. I added it because it is the only thing verifying `updateLoggers()` is called at all (at that it is once).

But I agree that its a bit too into the details. I'll remove it

pativa: We could also remove the configured logger filters checks, as that is also a bit into the details, and replace them with just checking that:
1. It is not counted when added to `Configuration` but `updateLoggers()` is not yet called
2. It is counted after `updateLoggers()` is called
3. It is no longer counted when the `Log4j2Metrics` is closed

PR #5944, opened 2025-02-19
Stackdriver integration test

### LINKED ISSUE #5943: Integration test for Stackdriver registry
I don't know how we can validate we are sending valid time series payloads to Stackdriver other than actually trying to write them to the real Stackdriver backend. It is a pain to set this up and has a cost associated with it, which is why we have not done this before. However, to prevent issues like #5927, we really should have this (or find another way to validate). Originally I started to work on this in #5942. However, it is going to take more time and process to figure out the logistics of the integration test, so I'm separating out the work to not block the fix there.

shakuzen:
Closes #5943

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,6 +100,17 @@ jobs:
           name: Deployment
           command: sh ./gradle/deploy.sh
 
+  stackdriver-it:
+    executor: circle-jdk-executor
+    steps:
+      - run:
+          name: Setup Google Cloud auth
+          command: |
+            echo "$GCLOUD_SERVICE_KEY" >> "$HOME"/gcloud-service-key.json
+            echo 'export GOOGLE_APPLICATION_CREDENTIALS="$HOME"/gcloud-service-key.json' >> "$BASH_ENV"
+      - gradlew-build:
+          command: stackdriverTest
+
 workflows:
   version: 2
   build_prs_deploy_snapshots:
@@ -109,6 +120,15 @@ workflows:
       - build-jdk17
       - concurrency-tests
       - docker-tests
+      - stackdriver-it:
+          context:
+            - gcp-metrics-writer
+          filters:
+            branches:
+              only:
+                - stackdriver-it
+                - main
+                - /\d+\.\d+\.x/
       - deploy:
           context:
             - deploy
@@ -118,6 +138,7 @@ workflows:
             - build-jdk17
             - concurrency-tests
             - docker-tests
+            - stackdriver-it
           filters:
             branches:
               only:
@@ -150,6 +171,14 @@ workflows:
             tags:
               only: /^v\d+\.\d+\.\d+(-(RC|M)\d+)?$/
       - docker-tests:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v\d+\.\d+\.\d+(-(RC|M)\d+)?$/
+      - stackdriver-it:
+          context:
+            - gcp-metrics-writer
           filters:
             branches:
               ignore: /.*/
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -100,6 +100,17 @@ jobs:
...
           command: sh ./gradle/deploy.sh
 
+  stackdriver-it:
+    executor: circle-jdk-executor
+    steps:
+      - run:
+          name: Setup Google Cloud auth
+          command: |
+            echo "$GCLOUD_SERVICE_KEY" >> "$HOME"/gcloud-service-key.json
+            echo 'export GOOGLE_APPLICATION_CREDENTIALS="$HOME"/gcloud-service-key.json' >> "$BASH_ENV"
```
jonatan-ivanov, line 110: This seems to have a default location so that: `GOOGLE_APPLICATION_CREDENTIALS` is not needed (but in that case you need to create the folder with `mkdir -p`):
```
$HOME/.config/gcloud/application_default_credentials.json
```

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
```diff
@@ -0,0 +1,152 @@
...
+    void timerWithSlos() {
+        Timer slos = Timer.builder("micrometer.stackdriver.it.timer.slo")
+            .maximumExpectedValue(Duration.ofMillis(2))
+            .serviceLevelObjectives(Duration.ofMillis(1), Duration.ofMillis(2))
+            .register(registry);
+        slos.record(Duration.ofMillis(1));
+        slos.record(Duration.ofMillis(2));
+        slos.record(Duration.ofMillis(3));
+
+        System.out.println(slos.getId().toString() + batch.distribution(slos.takeSnapshot(), true));
```
jonatan-ivanov, line 79: Should we keep the `System.out.println` statements?

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
```diff
@@ -0,0 +1,152 @@
...
+        summary.record(10d);
+        summary.record(20d);
+        summary.record(30d);
+
+        System.out.println(summary.getId().toString() + batch.distribution(summary.takeSnapshot(), false));
+        publishAndVerify();
+    }
+
+    @Test
+    void summaryWithOneSlo() {
```
jonatan-ivanov, line 140: Should we mirror the Timer tests for DistributionSummary too?

jonatan-ivanov (approved):

.circleci/config.yml
shakuzen: It's probably better to not rely on the default, and since it wouldn't eliminate the need to have a script, I think we can keep it this way. I was aligning with the CircleCI GCP orb defaults because I thought that might make things easier for us, but in the end it didn't so I removed it.

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
shakuzen: I think they're nice for debugging in the IDE when there are issues. They won't show up in the Gradle output generally, so it seems low cost to leave them. Thoughts?

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
shakuzen: We could, but other than unit and conversion of that unit, there isn't really a difference between the two when it comes to the logic or output. I was trying to make sure we had coverage of all the different underlying histograms and edge cases in the logic for producing the Stackdriver Distribution. If I've missed something, we should definitely add it. Otherwise I'm not sure we need to do the same configuration with each (famous last words).

shakuzen:
Merging now but we can still follow-up on any of the feedback as needed.

PR #5974, opened 2025-02-26
Configure histogram flavor/max buckets per meter

### LINKED ISSUE #5459: Support configuring exponential histograms at the meter level
This is a follow up to #3861 and requests from users such as https://github.com/micrometer-metrics/micrometer/pull/3959#issuecomment-2287000458.

Currently, exponential histograms can only be configured at the registry level for the OTLP registry via OtlpConfig. This is fine for users who want to use exponential histograms for everything that has `percentileHistogram` enabled. However, more fine grained control gives more flexibility.

As a workaround, configuring any SLO on a meter's DistributionStatisticConfig will cause exponential histograms to not be used (they don't support arbitrary additional buckets).

Another aspect of configuration would be the (max) bucket count, which again is currently controlled by OtlpConfig for all meters in the registry.

shakuzen:
Adds new configuration methods to override the registry-wide configuration at a per Meter (name) level. This gives more fine-grain control of the configuration.

The registry-wide HistogramFlavor configuration can be overridden at the meter level with the new `histogramFlavorPerMeter` configuration method, which returns a `Map<String, HistogramFlavor>` where the key is the exact Meter name to match.

Likewise, the registry-wide max bucket count can be overridden at the meter level with the new `maxBucketsPerMeter` configuration method, which returns a `Map<String, Integer>` where the key is the exact Meter name to match.

- [x] TODO: Update documentation

Closes #5459

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
@@ -15,7 +15,6 @@
  */
 package io.micrometer.registry.otlp;
 
-import io.micrometer.common.util.StringUtils;
 import io.micrometer.core.instrument.config.InvalidConfigurationException;
 import io.micrometer.core.instrument.config.validate.Validated;
 import io.micrometer.core.instrument.push.PushRegistryConfig;
@@ -24,7 +23,6 @@ import java.time.Duration;
 import java.net.URLDecoder;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static io.micrometer.core.instrument.config.MeterRegistryConfigValidator.*;
@@ -238,8 +236,8 @@ public interface OtlpConfig extends PushRegistryConfig {
      * @see #histogramFlavor()
      */
     default Map<String, HistogramFlavor> histogramFlavorPerMeter() {
-        String flavorPerMeterString = getString(this, "histogramFlavorPerMeter").orElse(null);
-        return propertiesStringToMap(flavorPerMeterString, HistogramFlavor::fromString);
+        return getStringMap(this, "histogramFlavorPerMeter", HistogramFlavor::fromString)
+            .orElse(Collections.emptyMap());
     }
 
     /**
@@ -277,7 +275,7 @@ public interface OtlpConfig extends PushRegistryConfig {
      * @see #maxBucketCount()
      */
     default Map<String, Integer> maxBucketsPerMeter() {
-        return propertiesStringToMap(getString(this, "maxBucketsPerMeter").orElse(null), Integer::parseInt);
+        return getStringMap(this, "maxBucketsPerMeter", Integer::parseInt).orElse(Collections.emptyMap());
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -264,4 +292,14 @@ default TimeUnit baseTimeUnit() {
         return getTimeUnit(this, "baseTimeUnit").orElse(TimeUnit.MILLISECONDS);
     }
 
+    static <V> Map<String, V> propertiesStringToMap(String propertiesString,
```
shakuzen, line 295: I don't like that this is a public method. We should probably move it somewhere internal instead.

shakuzen:
Pinging some folks who expressed interest in or provided input on this feature before to get any feedback they may have on this proposal: @lenin-jaganathan @pirgeo @sfc-gh-rscott @sfc-gh-dguy. Of course anyone else is welcome to give feedback too.

In https://github.com/micrometer-metrics/micrometer/issues/5459 I expressed a desire to expose the configuration at the `DistributionStatisticConfig` level. In this proposal (at the time of writing this comment), I've abandoned that and went for what feels more straightforward: making the configuration specific to `OtlpConfig` and the `OtlpMeterRegistry`. If anyone has thoughts on that, please share.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
shakuzen: Moved to `PropertyValidator` in 04616d8

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
```diff
@@ -276,4 +277,25 @@ void histogramPreferenceUseEnvVarWhenConfigNotSet() throws Exception {
                 .isEqualTo(HistogramFlavor.BASE2_EXPONENTIAL_BUCKET_HISTOGRAM));
     }
 
+    @Test
+    void histogramFlavorPerMeter() {
+        Map<String, String> properties = new HashMap<>();
+        properties.put("otlp.histogramFlavorPerMeter",
+                "a.b.c=explicit_bucket_histogram ,expo =base2_exponential_bucket_histogram");
```
jonatan-ivanov, line 284: Is spacing intentional?
```suggestion
                "a.b.c=explicit_bucket_histogram, expo=base2_exponential_bucket_histogram");
```

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
```diff
@@ -276,4 +277,25 @@ void histogramPreferenceUseEnvVarWhenConfigNotSet() throws Exception {
...
+        assertThat(otlpConfig.validate().isValid()).isTrue();
+        assertThat(otlpConfig.histogramFlavorPerMeter()).containsExactly(
+                entry("a.b.c", HistogramFlavor.EXPLICIT_BUCKET_HISTOGRAM),
+                entry("expo", HistogramFlavor.BASE2_EXPONENTIAL_BUCKET_HISTOGRAM));
+    }
+
+    @Test
+    void maxBucketsPerMeter() {
+        Map<String, String> properties = new HashMap<>();
+        properties.put("otlp.maxBucketsPerMeter", "a.b.c = 10");
```
jonatan-ivanov, line 295: Is spacing intentional?
```suggestion
        properties.put("otlp.maxBucketsPerMeter", "a.b.c=10");
```

micrometer-core/src/test/java/io/micrometer/core/instrument/config/validate/PropertyValidatorTest.java
```diff
@@ -0,0 +1,66 @@
...
+import io.micrometer.core.instrument.config.MeterRegistryConfig;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Properties;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class PropertyValidatorTest {
```
jonatan-ivanov, line 27: I would put tests here with tricky spacing (like above):
```
a=b ,c =d, e = f
```

jonatan-ivanov (approved):

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpConfigTest.java
shakuzen: Yes, this was intentionally testing that it can handle weird spacing. That could be made more explicit in a separate test perhaps.

micrometer-core/src/test/java/io/micrometer/core/instrument/config/validate/PropertyValidatorTest.java
shakuzen: Done in 8db4d6e5c84b1674a406fd7926bac5574606a9b5

PR #5988, opened 2025-03-02
Clarify JavaDoc for JMS `DESTINATION_NAME`

### LINKED ISSUE #5929: Why is JMS `destination.name` a high-cardinality key?
Why is JMS `destination.name` a _high_-cardinality key?
The related code - https://github.com/micrometer-metrics/micrometer/blob/ffda31402e0c745e32c6f6e27a0589ee27501b7d/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsObservationDocumentation.java#L137

At first sense it looks like a low-cardinality key because it is static at most times.
For example, in Spring `@JmsListener` it's specified from config
So, it's similar to `@RequestMapping` HTTP URI template which is low-cardinality key (see https://github.com/spring-projects/spring-framework/blob/fe41cd6d20842b8c8b34e530d1d93c9d83d617da/spring-web/src/main/java/org/springframework/http/server/observation/ServerHttpObservationDocumentation.java#L84)

This question relates to changes from https://github.com/micrometer-metrics/micrometer/pull/4007

AlexElin:
Closes #5929

### DIFF
--- a/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsObservationDocumentation.java
+++ b/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsObservationDocumentation.java
@@ -133,6 +133,10 @@ public enum JmsObservationDocumentation implements ObservationDocumentation {
 
         /**
          * The name of destination the current message was sent to.
+         * <p>
+         * The rationale for classifying it as a high-cardinality key is that
+         * {@link jakarta.jms.TemporaryQueue} has a great number of possible values of
+         * names.
          */
         DESTINATION_NAME {
             @Override


shakuzen (approved):
Thank you for the pull request. It looks good to me. We should update the JavaDoc back as far as we can. It looks like this was added in 1.12.0, so we should target the `1.12.x` branch and merge forward to include it in all of the upcoming maintenance releases.

shakuzen:
Another thought I had was would it be worth documenting this in [the documentation](https://docs.micrometer.io/micrometer/reference/reference.html)? We don't have a page for the JMS instrumentation currently, but if we did, it could include this explanation as well as an example of how users could customize the instrumentation to make it a low-cardinality key if they know that their usage of JMS that will be instrumented will no produce high-cardinality.

AlexElin:
@shakuzen thanks for proposal. I'm going to prepare a first version of the documenttaion on this or next week

shakuzen:
I'll go ahead and merge this so we get this update into the releases going out today. Feel free to send another pull request with any additional changes.

Issue #5690, opened 2024-11-19
Send metrics via any protocol in the OTLP Registry


kasparkivistik:
**Please describe the feature request.**
The idea is to make it available for any consumer of the micrometer-otlp-registry to provide their own metrics sending implementation. The OTLP Registry can still only support HTTP/protobuf as it does today but, for example, its implementation can be described by a public interface.

**Rationale**
The purpose of the change is to allow for metrics sending to be more flexible on the consumer side. I'm not advocating any changed to the OTLP protocols, rather my intention would be to open metrics sending up.
And example on my end would be that I desire to push all metrics to a Kafka topic before they are sent via HTTP/protobuf to OTLP collector. This way I have greater control over my metrics that might be coming from a number of microservices.

**Additional context**
Slack discussion: https://micrometer-metrics.slack.com/archives/C01C4Q8FHC2/p1731676707725909

jonas-xyb:
+1

jonatan-ivanov:
People who give this issue a 👍🏼 or 🚀, what is your use-case?

Right now, the use-case seems to be sending the data through Kafka which is not supported by the specs if I remember correctly. But I'm curious if people have other use-cases or you are co-workers and has the same use-case.

kasparkivistik:
@jonatan-ivanov they are co-workers 😅

For the time being I did a local workaround for the issue. Let's see what the community might think of it

schewara:
Personally I am not a big fan of adding too much complexity at this level, when it might be better placed on the collector level, but switching from the prometheus-registry to the opentelemetry-registry without losing any functionality by allowing to configure a prometheus exporter would allow for a much simpler migration path.

We are currently migrating away from prometheus metrics to otlp metrics and being able to switch between prometheus pull to otlp push on a switchover would be extremly helpful and would allow for fast fallback in case of problems.

jonatan-ivanov:
@schewara I'm not sure how your use-case is connected. You can use the Prometheus and the OTLP registry simultaneously, you can use as many registries as you want. If you use Spring Boot just add both dependencies, if you do things manually, register both prometheus and otlp registries into one `CompositeMeterRegistry` and use the composite to record data.

schewara:
@jonatan-ivanov You are absolutely correct. 
[...]

PR #5994, opened 2025-03-04
Improve OtlpMetricsSender API

shakuzen:
Remove the possible inconsistency where the sender could be given an instance of OtlpConfig that differs from the one passed to OtlpMeterRegistry. It is not clear which would be used or why. Now it is clear that the config on the registry will be used and the same sender can be used with different registry instances, potentially configured with different addresses (OtlpConfig#url).

See #5691 and specifically https://github.com/micrometer-metrics/micrometer/pull/5691#discussion_r1948505743. Breaks the API introduced in the previous milestone for #5690.

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/MetricsSender.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/MetricsSender.java
@@ -24,9 +24,12 @@ interface MetricsSender {
     /**
      * Send encoded metrics data from a {@link io.micrometer.core.instrument.MeterRegistry
      * MeterRegistry}.
+     * @param address where to send the metrics
      * @param metricsData encoded batch of metrics
      * @param headers metadata to send as headers with the metrics data
+     * @throws Throwable when there is an exception in sending the metrics; the caller
+     * should handle this in some way such as logging the exception
      */
-    void send(byte[] metricsData, Map<String, String> headers);
+    void send(String address, byte[] metricsData, Map<String, String> headers) throws Throwable;
 
 }
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpHttpMetricsSender.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpHttpMetricsSender.java
@@ -32,32 +32,31 @@ public class OtlpHttpMetricsSender implements OtlpMetricsSender {
 
     private final HttpSender httpSender;
 
-    private final OtlpConfig config;
-
     private final String userAgentHeader;
 
-    public OtlpHttpMetricsSender(HttpSender httpSender, OtlpConfig config) {
+    public OtlpHttpMetricsSender(HttpSender httpSender) {
         this.httpSender = httpSender;
-        this.config = config;
         this.userAgentHeader = getUserAgentHeader();
     }
 
+    /**
+     * Send a batch of OTLP Protobuf format metrics to an OTLP HTTP receiver.
+     * @param address address of the OTLP HTTP receiver to which metrics will be sent
+     * @param metricsData OTLP protobuf encoded batch of metrics
+     * @param headers metadata to send as headers with the metrics data
+     * @throws Throwable when there is an exception in sending the metrics; the caller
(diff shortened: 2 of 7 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
```diff
@@ -26,11 +26,14 @@
...
+     * Send a batch of OTLP Protobuf format metrics to an OTLP receiver.
+     * @param address address of the OTLP receiver to which metrics will be sent
      * @param metricsData OTLP protobuf encoded batch of metrics
      * @param headers metadata to send as headers with the metrics data
+     * @throws Throwable when there is an exception in sending the metrics; the caller
+     * should handle this in some way such as logging the exception
      */
     @Override
-    void send(byte[] metricsData, Map<String, String> headers);
+    void send(String address, byte[] metricsData, Map<String, String> headers) throws Throwable;
```
shakuzen, line 37: I think taking the address as a String is generic enough to work for other implementations than the HTTP one, but let me know if anyone has a concern about this. The implementation is responsible for parsing the address String. The OtlpMeterRegistry will always send the OtlpConfig#url as the address, so while it may semantically be weird if the address isn't actually a URL but the configured sender implementation can handle it (since it is a String that isn't necessarily a URL), this should work nonetheless.

shakuzen: Thanks for the review. We could do something like that, and there's precedent in the codebase with HttpSender taking a Request object. I can take a closer look at it tomorrow.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/MetricsSender.java
```diff
@@ -24,9 +24,12 @@ interface MetricsSender {
...
      * Send encoded metrics data from a {@link io.micrometer.core.instrument.MeterRegistry
      * MeterRegistry}.
+     * @param address where to send the metrics
      * @param metricsData encoded batch of metrics
      * @param headers metadata to send as headers with the metrics data
+     * @throws Throwable when there is an exception in sending the metrics; the caller
+     * should handle this in some way such as logging the exception
      */
-    void send(byte[] metricsData, Map<String, String> headers);
+    void send(String address, byte[] metricsData, Map<String, String> headers) throws Throwable;
```
shakuzen, line 33: I didn't mention it in the PR description but `throws Throwable` was added here and elsewhere. The point was to ensure that the caller (MeterRegistry publish implementation) handles the exception case, usually with logging. `HttpSender` similarly declares `throws Throwable`. That said, it may be worth looking at doing things differently here because catching Throwable includes things like OOME, as pointed out by @marcingrzejszczak in an offline review.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpHttpMetricsSender.java
```diff
@@ -70,14 +69,12 @@ private String getUserAgentHeader() {
...
 
-    /**
-     * Get the configuration context.
-     * @return A message containing enough information for the log reader to figure out
-     * what configuration details may have contributed to the failure.
-     */
-    private String getConfigurationContext() {
-        // While other values may contribute to failures, these two are most common
-        return "url=" + config.url() + ", resource-attributes=" + config.resourceAttributes();
+    private static class OtlpHttpMetricsSendUnsuccessfulException extends RuntimeException {
```
jonatan-ivanov, line 72: I'm thinking if we should make this more general and public so that other implementations can reuse it and users can get the same exception class for the different implementation.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
jonatan-ivanov: I'm totally fine calling it `address` but if you are interested in an alternative: `endpoint`.

jonatan-ivanov (approved):

shakuzen:
I've opened #6025 as a follow-up

Issue #5488, opened 2024-09-24
Metrics for Executors.newVirtualThreadPerTaskExecutor()


shakuzen:
See https://github.com/micrometer-metrics/micrometer/issues/3956#issuecomment-2326269549 and other comments in that issue related to this.

izeye:
I created https://github.com/micrometer-metrics/micrometer/pull/6008 to try to resolve this.

PR #6008, opened 2025-03-11
Support instrumentation for Executors.newVirtualThreadPerTaskExecutor()

izeye:
This PR tries to support instrumentation for the `Executors.newVirtualThreadPerTaskExecutor()`.

Closes gh-5488

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
@@ -29,6 +29,7 @@ import io.micrometer.core.instrument.internal.TimedExecutorService;
 import io.micrometer.core.instrument.internal.TimedScheduledExecutorService;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.*;
@@ -55,6 +56,11 @@ import static java.util.stream.Collectors.toSet;
 @NonNullApi
 @NonNullFields
 public class ExecutorServiceMetrics implements MeterBinder {
+
+    private static final String CLASS_NAME_THREAD_PER_TASK_EXECUTOR = "java.util.concurrent.ThreadPerTaskExecutor";
+
+    @Nullable
+    private static final Method METHOD_THREAD_COUNT_FROM_THREAD_PER_TASK_EXECUTOR = getMethodForThreadCountFromThreadPerTaskExecutor();
 
     private static boolean allowIllegalReflectiveAccess = true;
 
@@ -315,6 +321,9 @@ public class ExecutorServiceMetrics implements MeterBinder {
                 monitor(registry,
                         unwrapThreadPoolExecutor(executorService, executorService.getClass().getSuperclass()));
             }
+            else if (className.equals(CLASS_NAME_THREAD_PER_TASK_EXECUTOR)) {
+                monitorThreadPerTaskExecutor(registry, executorService);
+            }
             else {
                 log.warn("Failed to bind as {} is unsupported.", className);
             }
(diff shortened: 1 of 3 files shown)


micrometer-java11/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     testImplementation 'ru.lanwen.wiremock:wiremock-junit5'
     testImplementation 'com.github.tomakehurst:wiremock-jre8-standalone'
     testImplementation project(":micrometer-observation-test")
+    testImplementation libs.awaitility
```
shakuzen, line 10: This isn't needed, is it?

shakuzen (commented):
Thanks for the pull request. I wonder how many users will want to run with `--add-opens` and if we should in some sense be encouraging that. At a minimum, we should document the limitation. I suppose there isn't another option now. I did mean to re-start conversations with the JDK team about providing some way to do this monitoring without illegal reflection access.

micrometer-java11/build.gradle
izeye: @shakuzen Sorry, you're right. I removed it in d99f984.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -439,6 +448,39 @@ private void monitor(MeterRegistry registry, ForkJoinPool fj) {
...
+        try {
+            return (long) METHOD_THREAD_COUNT_FROM_THREAD_PER_TASK_EXECUTOR.invoke(executorService);
+        }
+        catch (Throwable e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Nullable
+    private static Method getMethodForThreadCountFromThreadPerTaskExecutor() {
```
juliojgd, line 472: Wouldn't it be better to generate a `java.lang.invoke.MethodHandle` once (maybe static) and use it (the same one, cached) in every call? Maybe it would be better for performance reasons

izeye: @juliojgd Thanks for the feedback!

I'm not sure how easy implementing this with the `MethodHandle` would be, but it's not a hot spot, so I think it's okay as is.

juliojgd: IMHO, the gauge will call the method every N seconds, right? And the method in turn will make the reflective call each time, so maybe it could be a hotspot (due to the number of calls made during the JVM instance life).

izeye: @juliojgd I'm not familiar with the `MethodHandle`, so I'm not sure how effective it is here, but I attempted to switch to it in e5d159b.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -461,20 +463,20 @@ private void monitorThreadPerTaskExecutor(MeterRegistry registry, ExecutorServic
...
         catch (Throwable e) {
             throw new RuntimeException(e);
         }
     }
 
     @Nullable
-    private static Method getMethodForThreadCountFromThreadPerTaskExecutor() {
+    private static MethodHandle getMethodHandleForThreadCountFromThreadPerTaskExecutor() {
         try {
             Class<?> clazz = Class.forName(CLASS_NAME_THREAD_PER_TASK_EXECUTOR);
```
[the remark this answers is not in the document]
juliojgd: with methodhandles in theory the accesscheck occurs only once (during the lookup) but with plain reflection the access check occurs in every invocation. At least this is my understanding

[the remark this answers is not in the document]
izeye: @juliojgd `MethodHandles.privateLookupIn()` requires JDK 9+, but Micrometer needs to work on JDK 8, so we can't call it directly.

Issue #5950, opened 2025-02-20
Track unfinished virtual threads


ceremo:
In order to have more details about how VT's are being created/destroyed, it seems reasonable to listen also the [start/end](https://openjdk.org/jeps/425#JDK-Flight-Recorder-JFR) events, and create a metric that measures all the VT started but not finished, this could be useful to detect if the application is being overwhelmed, and is creating VT's quicker than closing them.

Additionally, it could be useful in the class [`VirtualThreadMetrics`](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java) to customize what events are going to generate metrics, and the recording settings.

Thanks for your time.

jonatan-ivanov:
Hi,

Thank you for the issue. I think this can be useful but I also have two concerns/questions:
1. This can mean lots of events and I'm not sure about the performance implications (should be negligible but it might also be redundant, see the next point).
2. Aren't your operations that are executed via virtual threads already instrumented? E.g.: if `LongTaskTimer` is used, you get the "active count" which should be the same as the number of virtual threads that are "active" at that moment for that operation.

Or do you have a different use-case?

ceremo:
Hi @jonatan-ivanov,

I've been running some performance tests in my application, and I cannot see any difference in performance, even when including the VT start/end events. However, as you mentioned, using a task timer should allow you to obtain the same VT "active" value, so it could be an alternative.

Thanks.

ceremo:
Thanks @jonatan-ivanov, I'll come up with a PR. In my opinion, it would be useful to have a global metric to report the number of active VT's.

shakuzen:
I made a note for myself to look into [JDK-8338890](https://bugs.openjdk.org/browse/JDK-8338890) before, but I realize now it seems to provide a solution for the underlying ask, and it doesn't require us to deal with JFR events (meaning we could add support now), although it only works on Java 24+.

Would that work for your purposes, @ceremo and anyone else looking for this feature? See the methods on [VirtualThreadSchedulerMXBean](https://docs.oracle.com/en/java/javase/24/docs/api/jdk.management/jdk/management/VirtualThreadSchedulerMXBean.html).

Unfinished virtual thread count would be `getQueuedVirtualThreadCount()` + `getMountedVirtualThreadCount()`.

ceremo:
[...]

PR #6009, opened 2025-03-11
Add metric to check the current number of VT's

ceremo:
Related to #5950

### DIFF
--- a/micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
+++ b/micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
@@ -55,20 +55,37 @@ public class VirtualThreadMetrics implements MeterBinder, Closeable {
 
     private final Iterable<Tag> tags;
 
-    private boolean activeMetricEnabled;
-
+    /**
+     * Instantiates a new Virtual thread metrics.
+     */
     public VirtualThreadMetrics() {
         this(new RecordingConfig(), emptyList());
     }
 
+    /**
+     * Instantiates a new Virtual thread metrics.
+     *
+     * @param config the config
+     */
     public VirtualThreadMetrics(RecordingConfig config) {
         this(config, emptyList());
     }
 
+    /**
+     * Instantiates a new Virtual thread metrics.
+     *
+     * @param tags the tags
+     */
     public VirtualThreadMetrics(Iterable<Tag> tags) {
         this(new RecordingConfig(), tags);
     }
 
+    /**
+     * Instantiates a new Virtual thread metrics.
+     *
+     * @param config the config
+     * @param tags   the tags
+     */
     public VirtualThreadMetrics(RecordingConfig config, Iterable<Tag> tags) {
         this.recordingCfg = config;
         this.tags = tags;
@@ -79,20 +96,26 @@ public class VirtualThreadMetrics implements MeterBinder, Closeable {
         if (this.recordingStream == null) {
             this.recordingStream = createRecordingStream(this.recordingCfg);
         }
-        Timer pinnedTimer = Timer.builder(VT_PINNED_METRIC_NAME)
-            .description("The duration while the virtual thread was pinned without releasing its platform thread")
-            .tags(tags)
-            .register(registry);
 
-        Counter submitFailedCounter = Counter.builder(SUBMIT_FAILED_METRIC_NAME)
-            .description("The number of events when starting or unparking a virtual thread failed")
-            .tags(tags)
-            .register(registry);
+        if (recordingCfg.pinnedMetricEnabled) {
+            Timer pinnedTimer = Timer.builder(VT_PINNED_METRIC_NAME)
+                .description("The duration while the virtual thread was pinned without releasing its platform thread")
+                .tags(tags)
+                .register(registry);
+            
+            recordingStream.onEvent(PINNED_EVENT, event -> pinnedTimer.record(event.getDuration()));
+        }
 
-        recordingStream.onEvent(PINNED_EVENT, event -> pinnedTimer.record(event.getDuration()));
-        recordingStream.onEvent(SUBMIT_FAILED_EVENT, event -> submitFailedCounter.increment());
+        if(recordingCfg.submitFailedMetricEnabled) {
+            Counter submitFailedCounter = Counter.builder(SUBMIT_FAILED_METRIC_NAME)
+                .description("The number of events when starting or unparking a virtual thread failed")
+                .tags(tags)
+                .register(registry);
 
-        if (activeMetricEnabled) {
+            recordingStream.onEvent(SUBMIT_FAILED_EVENT, event -> submitFailedCounter.increment());
+        }
+
+        if (recordingCfg.activeMetricEnabled) {
             final LongAdder activeCounter = new LongAdder();
             this.recordingStream.onEvent(START_EVENT, event -> activeCounter.increment());
             this.recordingStream.onEvent(END_EVENT, event -> activeCounter.decrement());
@@ -106,14 +129,18 @@ public class VirtualThreadMetrics implements MeterBinder, Closeable {
 
     private RecordingStream createRecordingStream(RecordingConfig config) {
         RecordingStream recordingStream = new RecordingStream();
-        recordingStream.enable(PINNED_EVENT).withThreshold(config.pinnedThreshold);
-        recordingStream.enable(SUBMIT_FAILED_EVENT);
-        recordingStream.setMaxAge(config.maxAge);
-        recordingStream.setMaxSize(config.maxSizeBytes);
-        if (activeMetricEnabled) {
+        if(config.pinnedMetricEnabled) {
+            recordingStream.enable(PINNED_EVENT).withThreshold(config.pinnedThreshold);
+        }
+        if(config.submitFailedMetricEnabled) {
+            recordingStream.enable(SUBMIT_FAILED_EVENT);
+        }
+        if (config.activeMetricEnabled) {
             recordingStream.enable(START_EVENT);
             recordingStream.enable(END_EVENT);
         }
+        recordingStream.setMaxAge(config.maxAge);
+        recordingStream.setMaxSize(config.maxSizeBytes);
         recordingStream.startAsync();
 
         return recordingStream;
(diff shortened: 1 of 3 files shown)


micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
```diff
@@ -90,12 +124,16 @@ public void close() {
         recordingStream.close();
     }
 
-    private record RecordingConfig(Duration maxAge, long maxSizeBytes, Duration pinnedThreshold) {
+    public void setActiveMetricEnabled(boolean activeMetricEnabled) {
+        this.activeMetricEnabled = activeMetricEnabled;
+    }
+
+    public record RecordingConfig(Duration maxAge, long maxSizeBytes, Duration pinnedThreshold) {
```
shakuzen, line 131: If making this public, it should have a JavaDoc. We should probably also add some tests verifying usage we expect from users.

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
```diff
@@ -90,12 +124,16 @@ public void close() {
         recordingStream.close();
     }
 
-    private record RecordingConfig(Duration maxAge, long maxSizeBytes, Duration pinnedThreshold) {
+    public void setActiveMetricEnabled(boolean activeMetricEnabled) {
+        this.activeMetricEnabled = activeMetricEnabled;
```
shakuzen, line 128: Having a setter creates an implicit invariant that this needs to be called before `bindTo` or it will not have any effect. That may surprise some users, and I also don't want to establish it has a precedent where we'd similarly have to add a setter for any future additions. What about making this an option on `RecordingConfig` instead? It was somewhat intentional it wasn't public so that if things like this came up we'd be able to more easily change it. Once we make it public, we need to take care to not break backward compatibility.

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
```diff
@@ -41,37 +39,69 @@ public class VirtualThreadMetrics implements MeterBinder, Closeable {
...
         recordingStream.onEvent(PINNED_EVENT, event -> pinnedTimer.record(event.getDuration()));
         recordingStream.onEvent(SUBMIT_FAILED_EVENT, event -> submitFailedCounter.increment());
+
+        if (activeMetricEnabled) {
+            final LongAdder activeCounter = new LongAdder();
+            this.recordingStream.onEvent(START_EVENT, event -> activeCounter.increment());
+            this.recordingStream.onEvent(END_EVENT, event -> activeCounter.decrement());
+
+            Gauge.builder(VT_ACTIVE_METRIC_NAME, activeCounter::doubleValue)
+                .description("The number of active virtual threads")
```
shakuzen, line 101: Rather than a gauge, I wonder if it might be more insightful to have two counters, one for started, one for ended. The active could be derived by taking the difference. And this would allow tracking the rate of virtual threads being started/stopped. Thoughts? Maybe in some use cases we would have to worry about overflow (assuming cumulative counters)?

shakuzen (commented):
Thanks for the pull request.

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
ceremo: It would be better to have both counters. However, these types of events (start/end) are numerous, so they would increase rapidly, and overflow could become a common scenario.

Indresh2410: @ceremo @shakuzen  IMO, if we have the ability to track rate of virtual threads being started, we could have an option to autoscale JVM apps based on rate

Indresh2410:
@shakuzen / @ceremo any help on confirmation if we are going with two metrics for creation and deletion ? Having them will be really helpful

jonatan-ivanov:
I was wondering if we should make `RecordingConfig` public and if we do that, will these additional config + enable/disable parameters get out of hand over time. I came up with a few ideas:

1. Abstract `JfrMeterBinder`
I think this is something we could do regardless of this PR to support JFR in general. We could move the common logic of `RecordingStream` handling into a generic class and implementations can enable/disable what they want and register their meters and event handlers, see this draft: https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:jfr-binder. This also means that we would make `RecordingStream` public on our public API surface.

2. Generic `JfrMeterBinder`
Similarly to the abstract class, we can let the users inject their own "registrar" (`BiConsumer<MeterRegistry, RecordingStream>`) and "configurer" (`Consumer<RecordingStream>`). See the previous draft: https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:jfr-binder but instead of the method definitions we would have injectable functional interfaces.

3. In addition to either `#1` or `#2` or none of them, we could have an enum for these events and users can define in the ctor which ones should be enabled (or disabled) (something similar to [`IgnoredMeters`](https://github.com/micrometer-metrics/micrometer/blob/0dc1b259dcab2bacc8e3758b972198eefccdf628/micrometer-core/src/main/java/io/micrometer/core/instrument/observation/DefaultMeterObservationHandler.java#L65) in `DefaultMeterObservationHandler`).

What do you think?

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
shakuzen: I think for autoscaling, the important metric is whether the active (virtual) thread count is continually increasing. If you just look at the rate of threads starting without looking at the rate of threads finishing, I'm not sure it tells much that is important for autoscaling because if you have a high rate of starting threads but finishing is keeping up, there is no problem and no autoscaling needed. Given that, maybe the rate of starting or stopping on its own isn't that interesting and the gauge of active virtual threads is most useful.

Indresh2410: Yes @shakuzen . The above makes sense. But lets say we have active virtual threads count. Based on the traffic, we'll have to finetune our thresholds accordingly for active virtual threads

Whereas if we calculate % of virtual threads being used like (active threads/total threads created)*100, we can do 1 time setup where if threshold is greater than 60% (An example), autoscaling should kick in, which generally means active threads are increased (Rate of Finishing threads has slowed down)

ceremo:
> I was wondering if we should make `RecordingConfig` public and if we do that, will these additional config + enable/disable parameters get out of hand over time. I came up with a few ideas:
> ...
> What do you think?

The way I see it, if we want to create an abstract solution that's not only useful for listening to Virtual Threads JFR events, we should allow reusing the `RecordingStream` in the different implementations of the `JfrMeterBinder`, having a different `RecordingStream` for each type sounds not optimal. Related to this, the abstract class should not call the start of the `RecordingStream`, if this is shared, and executes more than once, it will fail:

```
Recording can only be started once.
java.lang.IllegalStateException: Recording can only be started once.
```

micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
shakuzen: > Based on the traffic, we'll have to finetune our thresholds accordingly for active virtual threads

I don't think so. You might want to check for an absolute threshold of active virtual threads that is cause for concern, and you may want to check over multiple step intervals that the active virtual thread count is increasing significantly.

> Whereas if we calculate % of virtual threads being used like (active threads/total threads created)*100, we can do 1 time setup where if threshold is greater than 60%

Say normally you have between 10-100 active threads, but when there is an issue it climbs to 300, 500, 2000, etc active threads. If your application just started and there is some issue, checking the ratio of active threads to total created *may* catch the issue. But if the issue happens long after the application started, the ratio no longer catches it because the total threads created is monotonic; it increases over the lifetime of the application. For this reason, using such a ratio is not a good way to catch irregularities.

Indresh2410: Thanks @shakuzen . One final doubt. Apologies for dragging the conversation

> But if the issue happens long after the application started, the ratio no longer catches it because the total threads created is monotonic; it increases over the lifetime of the application. For this reason, using such a ratio is not a good way to catch irregularities.

Since virtual threads created will be a counter, won't the below expression help for us?

sum(active_thread_count)/sum(rate(virtual_threads_created[1m]))

Because taking rate of a counter will give the number of threads created over a period of interval right(1m)? Even if application runs for days/months? (Assuming active_thread_count is a gauge)

shakuzen:
> having a different `RecordingStream` for each type sounds not optimal.

Is it costly to have multiple RecordingStream instances? I wondered about the same thing in the original PR adding the virtual thread metrics but I didn't look into it to get to an answer. While being more efficient is a goal, the benefits from the isolation of not sharing it (which may introduce the ability of one binder to cause issues for another binder, depending on how things are exposed) feel more important if the only efficiency is allocating several bytes for one more RecordingStream object per binder. Presumably different binders will be listening to different JFR events anyway.

As for the "enabled" booleans, I don't think we need these for the existing metrics - those should have near negligible overhead and users don't need to bind this MeterBinder if they don't want that. The reason why it is important to separately be able to decide about enabling the start/stop virtual thread metrics is that the overhead can be more significant with high throughput using virtual threads. We have MeterFilter which lets users disable metrics they don't want, but there may be instrumentation overhead involved even if the metric is disabled. We could avoid calling `RecordingStream#onEvent` if the meter is disabled (a no-op). There may be overhead in enabling the event as well. If we could lazily enable the event only when the meter that needs it is not a no-op, then I think we can eliminate the overhead via a MeterFilter without needing an explicit "enabled" boolean. But it would be better for instrumentation with a potentially significant overhead to be opt-in separate from the rest of the instrumentation, so maybe an "enabled" boolean is still the best for this, unless we want to make a separate binder for instrumenting virtual threads start/stop.

ceremo:
> Is it costly to have multiple RecordingStream instances? I wondered about the same thing in the original PR adding the virtual thread metrics but I didn't look into it to get to an answer. While being more efficient is a goal, the benefits from the isolation of not sharing it (which may introduce the ability of one binder to cause issues for another binder, depending on how things are exposed) feel more important if the only efficiency is allocating several bytes for one more RecordingStream object per binder. Presumably different binders will be listening to different JFR events anyway.

Each RecordingStream requires its own memory allocation and appears to need at least 3 threads ("JFR Periodic Tasks", 
"JFR Event Stream" and "JFR Recorder Thread") to process the events. What I'm suggesting is to support both scenarios:

- the default one, RecordingStream created internally for each binder, not reusable
- the custom one, allowing to set a RecordingStream, enabling reuse across different binders

shakuzen:
There are some valid points of feedback being discussed, but it feels like we're getting farther from delivering on the original feature request rather than closer. That may be for the better in the long-term, but I want to lay out the options to proceed as I see them. We can 1) deliver the original feature (an additional opt-in meter for current number of virtual threads) without delivering another feature (generic JFR event-based meter binder abstraction). However, if we intend to introduce the generic JFR MeterBinder later and apply it to VirtualThreadMetrics, we need to keep that in mind when considering API changes now. Alternatively, we can 2) put this new meter feature on hold while we figure out the generic JFR MeterBinder. Or 3) try to tackle both features at once. There is also the release schedule to consider. A feature like generic JFR MeterBinder probably would be better to deliver in a milestone so there is more opportunity for feedback - right now our next feature release is a Release Candidate. So, delivering a generic JFR MeterBinder probably is best to delay until Micrometer 1.16.

I think my preference is for 2, and if we do that we should focus in a separate PR on reviewing what the API for a generic JFR MeterBinder looks like before deciding how this PR ends up - unless we can deliver this PR in a way that would not cause trouble for adapting to the generic JFR MeterBinder later in a backward compatible way.

@ceremo what are your thoughts? Are you okay putting this on hold or would you like to try to get this PR merged in time for 1.15 (without tackling a generic JFR MeterBinder feature)?

ceremo:
> @ceremo what are your thoughts? Are you okay putting this on hold or would you like to try to get this PR merged in time for 1.15 (without tackling a generic JFR MeterBinder feature)?

No problem keeping this on hold. Defining a new API for a generic JFR MeterBinder makes sense, so I would also choose option the second alternative.

shakuzen:
I've marked this as blocked and opened https://github.com/micrometer-metrics/micrometer/issues/6065, targeted at 1.16.0-M1.

shakuzen:
Superseded by #6104, which is based on a new MBean in Java 24 rather than JFR events.

Issue #6012, opened 2025-03-12
micrometer-observation-test brings unnecessary JUnit dependencies, leading to conflicts


snicoll:
Courtesy of @wilkinsona 

The `micrometer-observation-test` has a too broad (api) dependency on JUnit: https://github.com/micrometer-metrics/micrometer/blob/74e113935b3b8a4d3a546e33eb41189b0b35c907/micrometer-observation-test/build.gradle#L7

This leads to a number of unnecessary dependencies brought to consumers

```
+--- io.micrometer:micrometer-observation-test -> 1.15.0-SNAPSHOT
|    +--- io.micrometer:micrometer-observation:1.15.0-SNAPSHOT (*)
|    +--- org.assertj:assertj-core:3.27.3 -> 3.25.3
|    |    \--- net.bytebuddy:byte-buddy:1.14.11
|    +--- org.junit.jupiter:junit-jupiter:5.12.0
|    |    +--- org.junit:junit-bom:5.12.0 -> 5.11.0 (*)
|    |    +--- org.junit.jupiter:junit-jupiter-api:5.12.0
|    |    |    +--- org.junit:junit-bom:5.12.0 -> 5.11.0 (*)
|    |    |    +--- org.opentest4j:opentest4j:1.3.0
|    |    |    \--- org.junit.platform:junit-platform-commons:1.12.0
|    |    |         \--- org.junit:junit-bom:5.12.0 -> 5.11.0 (*)
|    |    +--- org.junit.jupiter:junit-jupiter-params:5.12.0
|    |    |    +--- org.junit:junit-bom:5.12.0 -> 5.11.0 (*)
|    |    |    \--- org.junit.jupiter:junit-jupiter-api:5.12.0 (*)
|    |    \--- org.junit.jupiter:junit-jupiter-engine:5.12.0
|    |         +--- org.junit:junit-bom:5.12.0 -> 5.11.0 (*)
|    |         +--- org.junit.platform:junit-platform-engine:1.12.0
|    |         |    +--- org.junit:junit-bom:5.12.0 -> 5.11.0 (*)
|    |         |    +--- org.opentest4j:opentest4j:1.3.0
|    |         |    \--- org.junit.platform:junit-platform-commons:1.12.0 (*)
|    |         \--- org.junit.jupiter:junit-jupiter-api:5.12.0 (*)
```

[...]

PR #6017, opened 2025-03-13
JUnit should not be an `api` scope dependency

shakuzen:
API scope can cause issues for consumers due to runtime version mismatches.

Resolves gh-6012


shakuzen:
If we could re-do things, maybe it would make more sense to split into separate modules for the TCK code (that needs JUnit) and other code like the TestObservationRegistry and AssertJ-based assertions (that don't have a JUnit dependency).

PR #6025, opened 2025-03-14
Further enhancement to OtlpMetricsSender

shakuzen:
Introduces an immutable Request object and builder for convenience.

See review comments on #5994 and original PR #5691.


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/MetricsSender.java
shakuzen, line 1: Having this (as package-private) made things more complicated when introducing the Request class that we need to be public, so I removed it. It was only there to demonstrate that the OtlpMetricsSender could be made more generic for use in other registries, but until we verify that need more, let's remove this.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpHttpMetricsSender.java
```diff
@@ -41,19 +35,23 @@ public OtlpHttpMetricsSender(HttpSender httpSender) {
...
-        headers.forEach(httpRequest::withHeader);
-        HttpSender.Response response = httpRequest.send();
+            .withContent("application/x-protobuf", request.getMetricsData());
+        request.getHeaders().forEach(httpRequest::withHeader);
+        HttpSender.Response response;
+        try {
+            response = httpRequest.send();
+        }
+        catch (Throwable e) {
+            throw new Exception(e);
```
shakuzen, line 53: Since the `HttpSender` throws Throwable, we need to catch it here. Wrapping in `Exception` feels weird but I'm not sure what's better to do.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
```diff
@@ -23,17 +27,125 @@
...
+            }
+
+            /**
+             * Build this request and send it to the associated sender. Throws an
+             * exception if there is no sender associated with this builder.
+             * @throws Exception when sending the request fails, an exception is thrown
+             */
+            public void send() throws Exception {
+                if (sender == null) {
+                    throw new IllegalStateException("A sender has not been set");
```
shakuzen, line 142: I'm not sure I like this arrangement - there's basically two paths to using the builder. One where you start with passing a sender and end with calling `send` (although you could still call `build` without issue and manually pass the Request to a sender). The other where you don't pass a sender in the beginning and you should only call `build` at the end and not `send`. I was trying to mirror some of the convenience we have with the HttpSender and make the API more fluent.

I'm not sure how worthwhile it is to do all this engineering on this part of the API, which even users making their own `OtlpMetricsSender` won't interact with, since this API should be used within the `OtlpMeterRegistry`'s `publish` method, and we generally expect users should not need to implement that themselves.

shakuzen (commented):
I may tweak things more, but I wanted to open this for review.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
```diff
@@ -23,17 +27,125 @@
...
+
+        private final byte[] metricsData;
+
+        /**
+         * Represents a payload of metrics to be sent.
+         * @param address where to send the metrics
+         * @param metricsData encoded batch of metrics
+         * @param headers metadata to send as headers with the metrics data
+         */
+        public Request(@Nullable String address, Map<String, String> headers, byte[] metricsData) {
```
kasparkivistik, line 71: When using a builder consider making the constructor private. This way there's only one way of using the API

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
```diff
@@ -23,17 +27,125 @@
...
+                return this;
+            }
+
+            public Builder headers(Map<String, String> headers) {
+                this.headers = headers;
+                return this;
+            }
+
+            public Request build() {
+                return new Request(address, headers, metricsData);
```
kasparkivistik, line 132: should we nullcheck `metricsData` as it's the only non-optional field?

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
kasparkivistik: I'm questioning the same thing. Wouldn't it be easier to allow for the `Request` to be created and then the `OtlpMetricsSender` interface will just be used to pass the `Request` to. It seems a bit too complex at the moment.

But I do like the fluidness there... If we want to make sure the sender is always present, `Request.builder()` should be made private, so we have a guarantee the interface is always present

shakuzen: I've updated the PR so the only way to construct a `Request` object is via the sender's `newRequest` method. I left the option to `build` the `Request`, though that too seems superfluous since the request builder is made from a sender instance.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
shakuzen: I've made it private now. We can always make the scope more public later if it makes sense, so it's probably best to keep it more closed now, especially since this is API that most people won't touch (only those making their own OtlpMeterRegistry implementation of `publish` etc will need to use this)

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
shakuzen: Sure. I've added a null check in the Builder constructor.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
```diff
@@ -15,25 +15,125 @@
...
+        public String toString() {
+            return "OtlpMetricsSender.Request for address: " + address + ", headers: " + headers + ", metricsData: "
+                    + new String(metricsData, StandardCharsets.UTF_8);
+        }
+
+        public static class Builder {
+
+            private final byte[] metricsData;
+
+            private final OtlpMetricsSender sender;
```
jonatan-ivanov, line 101: I think removing this (the responsibility of the builder to send the request) as well as `newRequest()` from the sender would simplify things (the builder only builds and the sender only sends).

So the registry would do this:
```java
metricsSender.send(Request.builder(bytes).address(url).headers(headers).build());
```
instead of
```java
metricsSender.newRequest(request.toByteArray()).address(config.url()).headers(config.headers()).send();
```

shakuzen: Sure, we can do that. I was trying to align somewhat with what is there in HttpSender and the fluent API has its advantages, but it's not worth distracting ourselves with this API so much when its usage will be largely internal. I'll update as you've suggested.

Issue #2625, opened 2021-05-28
Prometheus/OpenMetrics `_created` timestamp


shakuzen:
See the [OpenMetrics specification](https://github.com/OpenObservability/OpenMetrics/blob/master/specification/OpenMetrics.md). This is a follow-up to the initial tracking issue #2419.

We don't currently track or publish the created timestamp for each MetricFamily.

alshain:
This seems to landing soon in Prometheus, so it'd be great to see it supported in Micrometer.

https://github.com/prometheus/prometheus/pull/13506

shakuzen:
Thanks for the heads up. I've subscribed to updates on that PR. I'm not sure if we need to do anything in Micrometer for this anymore since we've upgrade to the Prometheus Java client 1.x (see #4406). The new client has an exporter configuration for created timestamps - see https://prometheus.github.io/client_java/config/config/#exporter-properties. I suspect that would work with Micrometer 1.13.0 (currently 1.13.0-RC1 is available from the milestone repository). Does that do what you need?

stefan-fast:
I looked into this and tried using it. Indeed, it seems to be implemented in the lastest versions of the Prometheus Java Client and Micrometer is using these new versions with support for `includeCreatedTimestamps`. Unfortunately, the `createdTimestampMillis` parameter is always set to `0L` in micrometer for supported metrics types, e.g. counters: https://github.com/micrometer-metrics/micrometer/blob/be70882669c690b28f8ec1c6143edc2fb33a7cb2/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java#L221
(last parameter in that line). Therefore, the `hasCreatedTimestamp()` check in the Prometheus Java Client is always `false`:
https://github.com/prometheus/client_java/blob/e17c0a10d5b55afc107b2e0114a433742b96a0ab/prometheus-metrics-exposition-formats/src/main/java/io/prometheus/metrics/expositionformats/PrometheusTextFormatWriter.java#L102
Because of that, the `_created` timestamp is never actually written yet. In summary, if I didn't miss anything, micrometer still needs to implment `_created` timestamp support.

jonatan-ivanov:
@stefan-fast Yes, that's what this issue is about: we don't track the created timestamp right now.
We are going to do some planning for the next feature release, I will drop this into the set of topics we will discuss but it seems right now there is not a lot of users who needs this

[...]

PR #6032, opened 2025-03-17
Support enabling Prometheus _created timestamps

shakuzen:
Resolves gh-2625


jonatan-ivanov (approved):

jonatan-ivanov:
Looks good but the build is failing because of the formatter.

PR #6033, opened 2025-03-17
Fix a possible typo in metric name.

dmitry-workato:
'httpcomponents.httpclient.pool.total.connections' was set by two different functions before. This PR creates different metrics for 'leased' connections.


dmitry-workato:
Actually I see how it works now. The 'state' tag should separate the metrics. Sorry for the noise :)

Issue #5923, opened 2025-02-13
Make Prometheus Metric and Label naming conventions consistent


dpozinen:
Since 1.13 the behaviour of has been changed.

For metrics the `_counter` prefix has been removed, as is well documented and the **snake case convention remains**:
https://github.com/micrometer-metrics/micrometer/blob/39e977d23013be276292c11a485e5dd39a20349d/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusNamingConvention.java#L47-L48

**But** for _labels_ snake case has been removed in #4866: 
https://github.com/micrometer-metrics/micrometer/blob/39e977d23013be276292c11a485e5dd39a20349d/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusNamingConvention.java#L87-L89

But even though Prometheus client supports dots in metrics and labels since https://github.com/prometheus/client_java/commit/1a955b1d99b3a31f2c48db146832bf4466f9e12a , it still replaces dots with underscores via https://github.com/prometheus/client_java/blob/5e93f4b2af0321061865bad1f39f531f3d19bd0e/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/Labels.java#L119-L130 

For people using `PrometheusNamingConvention` in their own code the change to label handling is undocumented in the [migration guide](https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide) and actually the java doc for the method has an incorrect regex (without `.`):

https://github.com/micrometer-metrics/micrometer/blob/39e977d23013be276292c11a485e5dd39a20349d/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusNamingConvention.java#L82-L87

Now, I am generally confused with the `.` and `_` changes in Prometheus where they are allowed and yet... not allowed really? Maybe you know more about this.

I think it is a good idea to either mention the change regarding labels in `PrometheusNamingConvention` in the guide and fix the javadoc, or revert the snake case change to align with the way metrics are evaluated. To me the latter is preferable.

Let me know what you think, thanks!

shakuzen:
Thanks for bringing this to our attention. At a minimum, I think we should update the migration guide to mention the change since that is what may affect users in released versions regardless of what we may change from this point.

[...]

PR #6049, opened 2025-03-21
Handle prometheus names and labels consistently

jonatan-ivanov:
Closes gh-5923

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
@@ -20,8 +20,8 @@ import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.cumulative.CumulativeFunctionCounter;
 import io.micrometer.core.instrument.cumulative.CumulativeFunctionTimer;
-import io.micrometer.core.instrument.distribution.HistogramSnapshot;
 import io.micrometer.core.instrument.distribution.*;
+import io.micrometer.core.instrument.distribution.HistogramSnapshot;
 import io.micrometer.core.instrument.distribution.pause.PauseDetector;
 import io.micrometer.core.instrument.internal.DefaultGauge;
 import io.micrometer.core.instrument.internal.DefaultLongTaskTimer;
@@ -54,7 +54,6 @@ import java.util.stream.StreamSupport;
 
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
-import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.StreamSupport.stream;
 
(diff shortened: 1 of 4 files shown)


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -591,11 +590,11 @@ private void applyToCollector(Meter.Id id, Consumer<MicrometerCollector> consume
...
             meterRegistrationFailed(id,
                     "Prometheus requires that all meters with the same name have the same"
-                            + " set of tag keys. There is already an existing meter named '" + id.getName()
+                            + " set of tag keys. There is already an existing meter named '" + getConventionName(id)
                             + "' containing tag keys ["
                             + String.join(", ", collectorMap.get(getConventionName(id)).getTagKeys())
-                            + "]. The meter you are attempting to register" + " has keys ["
-                            + getConventionTags(id).stream().map(Tag::getKey).collect(joining(", ")) + "].");
+                            + "]. The meter you are attempting to register" + " has keys [" + String.join(", ", tagKeys)
+                            + "].");
```
jonatan-ivanov, line 597: This makes the message to use `_`, previously it was using `.`; with the 0.x client we are using them mixed (name uses `.`, tags use `_`).

So this will be something like:
```
[...] There is already an existing meter named 'my_counter' containing tag keys [test_k1]. The meter you are attempting to register has keys [test_k2]."
```

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusNamingConvention.java
```diff
@@ -85,7 +85,7 @@ else if (!conventionName.endsWith("_seconds")) {
      */
     @Override
     public String tagKey(String key) {
-        return PrometheusNaming.sanitizeLabelName(key);
+        return PrometheusNaming.sanitizeLabelName(PrometheusNaming.prometheusName(key));
```
jonatan-ivanov, line 88: This is the change that makes label names consistent to metric names inside of the registry and for users who directly want to use this `NamingConvention`.

jonatan-ivanov (commented):
I'm thinking if we should fix this starting from `1.13.x` or just in `main` (starting from `1.15.x`).
The behavior is changing (well, it's a bug) for those who use `PrometheusNamingConvention` directly, but otherwise the change should be transparent for others except the error message.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusNamingConvention.java
shakuzen: I wonder why we need to do this since we're already calling the Prometheus library's `sanitizeLabelName`. I would expect that to do any needed transformation of a label name. I wonder if its current behavior is intentional or not. It seems possible that there could be subtle intended differences in the rules about metric names vs labels, so using both here could be problematic in the future if it isn't already.

jonatan-ivanov: For the record: we asked @fstab, for now we should do `prometheusName` + `sanitizeXyz`.

PR #6068, opened 2025-03-28
Avoid needless re-registering Kafka metrics

### LINKED ISSUE #5757: Gauge double registration warning for Kafka consumer metrics
Since https://github.com/micrometer-metrics/micrometer/pull/5688 we observe warnings being logged for the default kafka consumer gauge metrics such as 'kafka.consumer.fetch.manager.bytes.consumed.rate' in the form of "This Gauge has been already registered".

**Stack trace of when a gauge metric is registered**
[1stTime_on_startup.txt](https://github.com/user-attachments/files/18238005/1stTime_on_startup.txt)
[2ndTime_after_startup.txt](https://github.com/user-attachments/files/18238007/2ndTime_after_startup.txt)

Looking at this function, it seems like this behavior is actually intended:
https://github.com/micrometer-metrics/micrometer/blob/9d650ca3634ce11b70c18abf700252a7ab4a27c1/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java#L138-L147

shakuzen:
Functionally this does not cause any issue because calls to register after the first will not do anything, but it results in a warning in the log for Gauge re-registration, which can mask other issues.

Resolves #5757

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
@@ -218,11 +218,10 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
                             registeredMeterIds.remove(otherId);
                         }
                         // Check if already exists
-                        else if (tags.size() == meterTagsWithCommonTags.size())
+                        else if (tags.size() == meterTagsWithCommonTags.size()) {
                             if (tags.containsAll(meterTagsWithCommonTags))
                                 return;
-                            else
-                                break;
+                        }
                         else
                             hasLessTags = true;
                     }


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
```diff
@@ -218,11 +218,10 @@ void checkAndBindMetrics(MeterRegistry registry) {
                             registeredMeterIds.remove(otherId);
                         }
                         // Check if already exists
-                        else if (tags.size() == meterTagsWithCommonTags.size())
+                        else if (tags.size() == meterTagsWithCommonTags.size()) {
                             if (tags.containsAll(meterTagsWithCommonTags))
                                 return;
-                            else
-                                break;
```
shakuzen, line 225: This early break would result in re-registering a meter that may already be registered, depending on the order of iteration.

shakuzen (commented):
I didn't add a test. One of the existing tests was actually already demonstrating the issue: `KafkaMetricsTest.shouldRemoveMeterWithLessTagsWithMultipleClients` had the warning log. After the fix, it is gone. We could add a test specifically for it, but there's nothing functional to assert, only logging.

jonatan-ivanov (approved):

PR #6072, opened 2025-03-31
Configure Error Prone in build

shakuzen:
Also fixes ERROR severity bug patterns. There are many more WARNINGs, but I thought it best to keep the PR minimal for easier review.

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -283,6 +283,11 @@ java11Test {
 }
 
 compileJava11TestJava {
+    options.errorprone.enabled = false
     sourceCompatibility = JavaVersion.VERSION_11
     targetCompatibility = JavaVersion.VERSION_11
+}
+
+compileJava11Java {
+    options.errorprone.enabled = false
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -274,6 +274,7 @@ public class TimedAspect {
 
     private Timer.Builder recordBuilder(ProceedingJoinPoint pjp, Timed timed, String metricName,
             String exceptionClass) {
+        @SuppressWarnings("NullTernary")
         Timer.Builder builder = Timer.builder(metricName)
             .description(timed.description().isEmpty() ? null : timed.description())
             .tags(timed.extraTags())
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
@@ -89,6 +89,7 @@ public interface Timer extends Meter, HistogramSupport {
      * empty.
      * @return This builder.
      */
+    @SuppressWarnings("NullTernary")
     static Builder builder(Timed timed, String defaultName) {
         if (timed.longTask() && timed.value().isEmpty()) {
             // the user MUST name long task timers, we don't lump them in with regular
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryVictoriaMetricsHistogram.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryVictoriaMetricsHistogram.java
@@ -251,7 +251,7 @@ public class FixedBoundaryVictoriaMetricsHistogram implements Histogram {
         printStream.format("%14s %10s\n\n", "Bucket", "TotalCount");
 
         for (CountAtBucket bucket : nonZeroBuckets()) {
-            printStream.format(Locale.US, "%14.1f %10d\n", bucket.bucket() / bucketScaling, bucket.count());
+            printStream.format(Locale.US, "%14.1f %10f\n", bucket.bucket() / bucketScaling, bucket.count());
         }
 
         printStream.write('\n');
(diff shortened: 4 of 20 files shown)


micrometer-core/build.gradle
```diff
@@ -283,6 +283,11 @@ java11Test {
 }
 
 compileJava11TestJava {
+    options.errorprone.enabled = false
     sourceCompatibility = JavaVersion.VERSION_11
     targetCompatibility = JavaVersion.VERSION_11
 }
+
+compileJava11Java {
+    options.errorprone.enabled = false
```
shakuzen, line 292: Disables errorprone for the multi-release JAR source. There was an error trying to compile it, all of the classes in the source are deprecated and duplicated in micrometer-java11, and errorprone runs on the micrometer-java11 module. So I don't think there's any point trying to get errorprone to work here at this time.

micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
```diff
@@ -423,7 +424,7 @@ public Builder pauseDetector(PauseDetector pauseDetector) {
         }
 
         @Override
-        public Builder description(String description) {
+        public Builder description(@Nullable String description) {
```
shakuzen: Adding `@Nullable` here and the other place in this file was an attempt to fix the `NullTernary` error, but it did not help. I can revert since it isn't really part of enabling errorprone.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistry.java
```diff
@@ -122,7 +122,7 @@ public boolean equals(Object o) {
             if (this == o) {
                 return true;
             }
-            if (o == null || getClass() != o.getClass()) {
+            if (!(o instanceof TestObservationContext)) {
```
shakuzen, line 125: I'm not sure if the previous logic was intentional. The `TestObservationContext` class is package private but non-final, so an argument could be made either way, but instanceof seems just as good even if it's probably not needed.

micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
```diff
@@ -274,6 +274,7 @@ private void record(ProceedingJoinPoint pjp, Timed timed, String metricName, Tim
 
     private Timer.Builder recordBuilder(ProceedingJoinPoint pjp, Timed timed, String metricName,
             String exceptionClass) {
+        @SuppressWarnings("NullTernary")
```
shakuzen, line 277: This one was a bit odd. I'm not sure it's actually a problem. We're passing `null` to a primitive varargs, which I don't think has the problem as a non-varargs primitive like shown in [the bug pattern](https://errorprone.info/bugpattern/NullTernary).

shakuzen (commented):
I'm not sure what we want to do about the JDK 11 build. The latest errorprone does not support compiling with JDK 11. We could use an old version of errorprone when running the build with JDK 11 or we could disable errorprone when compiling with JDK 11. Neither are great, and I'm happy to hear other suggestions.

> Running Error Prone on earlier JDK versions
> * Error Prone 2.10.0 is the latest version to support running on JDK 8.
> * Error Prone 2.31.0 is the latest version to support running on JDK 11.

I forget but maybe that's why we didn't introduce errorprone to our build before.

micrometer-core/build.gradle
shakuzen: I guess it's the same issue as the JDK 11 build, since the multi-release JAR plugin will download JDK 11 even if we're using a JDK that could compile to 11 via the `--release` flag.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistry.java
marcingrzejszczak: I'm pretty sure that that was generated by Intellij

jonatan-ivanov (approved):

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistry.java
shakuzen: In that case, no harm in changing it, I think. Thanks for the confirmation.

shakuzen:
I've disabled errorprone on <17 JDK builds. On our CI, this is only the JDK 11 build. I've noticed the JDK 17 build fail twice now with the following error. A re-run is successful, but I worry this is caused by configuring errorprone and our builds are going to be more flaky...

```
* What went wrong:
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryVictoriaMetricsHistogram.java
```diff
@@ -251,7 +251,7 @@ private void outputSummary(PrintStream printStream, double bucketScaling) {
         printStream.format("%14s %10s\n\n", "Bucket", "TotalCount");
 
         for (CountAtBucket bucket : nonZeroBuckets()) {
-            printStream.format(Locale.US, "%14.1f %10d\n", bucket.bucket() / bucketScaling, bucket.count());
+            printStream.format(Locale.US, "%14.1f %10f\n", bucket.bucket() / bucketScaling, bucket.count());
```
shakuzen, line 254: Perhaps worth noting that this is an actual bug Error Prone caught, which I believe no one has reported because no one uses this outputSummary API on HistogramSnapshots. We should probably deprecate it unless I'm missing somewhere it is used.

shakuzen:
I decreased the max Gradle heap from 2 gigs to 1, which seems to have helped. Memory usage is still significantly higher with Error Prone, so we'll have to keep an eye on that and see if we can make improvements some way. Disabling warnings didn't seem to have a noticeable effect.

Issue #4752, opened 2024-02-14
Document Java HttpClient instrumentation


shakuzen:
See #4728

PR #6075, opened 2025-04-02
Delete deprecated JDK 11 classes

shakuzen:
These were deprecated and moved to the micrometer-java11 module with Micrometer 1.13.0. The multi-release JAR arrangement that is used for these in micrometer-core has some issues, which is what caused the move to micrometer-java11. See gh-4728. It would help simplify our build to remove the multi-release JAR plugin, which ends up downloading a JDK for the target release on CI builds even if the configured toolchain can target that release.


shakuzen:
We have [this documentation](https://docs.micrometer.io/micrometer/reference/reference/java-httpclient.html) for the instrumentation and java11 module, which was added with https://github.com/micrometer-metrics/micrometer/issues/4752 since 1.13.0. We did not have any documentation of it at the time it was added via https://github.com/micrometer-metrics/micrometer/pull/3358 with Micrometer 1.10. I'm not sure we need anything in the documentation about migrating, but I'll add something to the wiki.

shakuzen:
Wiki has been updated: https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide

jonatan-ivanov (approved):

jonatan-ivanov:
Shouldn't this be in a new `1.15 Migration Guide` page instead of the existing `1.13 Migration Guide`?

shakuzen:
I'd argue it doesn't belong in a 1.15 migration guide. Users should have migrated when upgrading to 1.13 where the classes were deprecated. We just didn't put migration instructions in the migration guide then. We can still link to that section of the migration guide from the 1.15 release notes, I think. If the argument is the migration guide is only for breaking changes, then I guess there's a point to be made, but it seems better to have given the instructions people would eventually need to follow as early as possible. Since we can't go back in time, adding them to the migration guide for the earliest version where they could have done the migration made sense to me. Let me know what you think.

PR #6080, opened 2025-04-03
Replace global dependency management with version catalog

shakuzen:
Completely removes dependencies.gradle which acted as a global dependency management file for the whole build.

We previously introduced a version catalog and these changes make full use of it. Individual dependency versions are managed in the catalog and the dependencies are referenced using the type-safe API where they are declared. This should alleviate confusion contributors have about where to put versions and how to declare dependencies. It also avoids needless dependency management in all sub-projects even for dependencies that are only used in one place.

This also removes the global dependency management via the BOMs that were declared in dependencies.gradle. Instead, the BOMs are declared in the projects that use them at the appropriate scope. An exception to this is the JUnit BOM because each 'java-library' project should have tests and use JUnit, so it is added to all of them from the root build.gradle.


micrometer-jetty12/build.gradle
```diff
@@ -11,17 +11,13 @@ dependencies {
     optionalApi libs.jetty12Server
     optionalApi libs.jetty12Client
 
-    testRuntimeOnly(libs.logbackLatest) {
-        version {
-            strictly libs.logbackLatest.get().version
-        }
-    }
+    testRuntimeOnly(libs.logbackLatest)
```
shakuzen, line 14: A benefit of getting rid of build-wide dependency management is we don't have to fight it in cases like this where we want to use a different version in a project.

micrometer-observation/build.gradle
```diff
@@ -39,19 +39,15 @@ jar {
...
 
     // Aspects
     optionalApi libs.aspectjrt
 
-    // log monitoring
-    testImplementation 'ch.qos.logback:logback-classic'
-    testImplementation 'org.apache.logging.log4j:log4j-core'
-
-    testImplementation project(':micrometer-observation-test')
+    testRuntimeOnly libs.logback12
```
shakuzen, line 50: This was perhaps only needed because archunit seems to have a runtime dependency on slf4j. I didn't look into it too deeply. But log4j was not used or needed at all in this project so I got rid of that dependency while reducing the scope of the logback dependency.

micrometer-observation/build.gradle
```diff
@@ -39,19 +39,15 @@ jar {
...
+    optionalApi libs.javax.servletApi
 
     // Aspects
     optionalApi libs.aspectjrt
 
-    // log monitoring
-    testImplementation 'ch.qos.logback:logback-classic'
-    testImplementation 'org.apache.logging.log4j:log4j-core'
-
-    testImplementation project(':micrometer-observation-test')
```
shakuzen, line 54: I'm surprised the build didn't fail with this. It's a circular dependency because micrometer-observation-test depends on this project (micrometer-observation).

samples/micrometer-samples-kotlin/build.gradle
```diff
@@ -23,6 +23,8 @@ dependencies {
 
     testImplementation project(":micrometer-test")
 
+    testImplementation platform(libs.junitBom)
+    testRuntimeOnly platform(libs.junitBom)
```
shakuzen, line 27: We shouldn't need to declare the bom at testRuntimeOnly scope but the `downloadDependencies` task failed if I didn't here. I suspect it must have something to do with Kotlin since the other projects didn't need the same.

jonatan-ivanov (approved):
Awesome!

jonatan-ivanov:
I was thinking about two things in connection to this:
1. Should we do this starting from `1.13` (or even earlier for commercial branches)? I think probably not since we are not expecting a lot of changes around this that would cause lots of merge conflicts.
2. Since the upcoming releases will live long, I think we should do this for Tracing `1.5` too (before the RC, the rest of the projects can wait since we are not planning to release them in the upcoming cycle).

shakuzen:
On 1, we probably should have done this when we introduced the version catalog initially, but between merge conflicts and potential for subtle differences in resolved versions between the previous global dependency management and this, I think it's better to limit the risk to a new minor version.
On 2, agreed.

shakuzen:
Deploy failed after merging this. I've fixed it with https://github.com/micrometer-metrics/micrometer/commit/c6395005b3178bccf758ce027df87b532f0c4ff2

PR #6081, opened 2025-04-03
[WIP] Disable all Error Prone checks

shakuzen:
Seeing if this alleviates the memory consumption problem.


shakuzen:
It's not clear whether this helped or not. I'll close for now and we'll revisit if we see issues with the memory consumption causing builds to fail more.

PR #6102, opened 2025-04-09
Make per-meter OTLP configuration more flexible

### LINKED ISSUE #6099: More flexible OTLP per meter configuration
https://github.com/micrometer-metrics/micrometer/pull/5974 introduced the support for configuring HistogramFlavour/buckets per meter for OTLPMeterRegistry. This was done as part of a configuration that was supplied during the registry creation and does not allow overriding via APIs. This configuration does a full match of the meter name and requires the fully qualified meter name to be provided and this can get longer when we need to supply configuration to a family of meters.  

[Spring boot has a solution](https://docs.spring.io/spring-boot/reference/actuator/metrics.html#actuator.metrics.customizing.per-meter-properties) to address this issue by supporting the prefix and if there are more values the longest match wins. If users can supply a Function to be applied for determining these values also be an option to try out.

shakuzen:
The previous logic for matching required the exact meter name, which was straightforward but cumbersome for configuring groups of meters. This changes the default matching for the per-meter config to use the longest dot-separated segments match.

Resolves #6099


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -229,15 +231,16 @@ default HistogramFlavor histogramFlavor() {
...
+     * {@link #histogramFlavor()} configuration for matching Meters. By default, the key
+     * is used to do an exact match on the Meter's name.
+     * @return mapping of meter ID to histogram flavor
      * @since 1.15.0
      * @see #histogramFlavor()
      */
-    default Map<String, HistogramFlavor> histogramFlavorPerMeter() {
-        return getStringMap(this, "histogramFlavorPerMeter", HistogramFlavor::fromString)
-            .orElse(Collections.emptyMap());
+    default Function<Meter.Id, HistogramFlavor> histogramFlavorPerMeter() {
```
shakuzen, line 240: I made the signature `Function<Meter.Id, HistogramFlavor>` for the most flexibility, but perhaps `<String, HistogramFlavor>` is better where the `String` is the meter name. While programmatically the implementation of the `Function` could do anything, we usually try to keep it reasonable to be configured from a Properties/Map source. I don't see a reasonable implementation from properties that can be deserialized into a Function accepting a Meter.Id. I also wonder if users in practice would want to make this configuration on some basis other than name anyway.

jonatan-ivanov (approved):

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -228,18 +229,40 @@ default HistogramFlavor histogramFlavor() {
...
+     * {@link #histogramFlavorPerMeter()} provides the data while this method provides the
+     * logic for the lookup, and you can override them independently.
+     * @param id the {@link Meter.Id} the {@link HistogramFlavor} is configured for
+     * @return the histogram flavor mapped to the {@link Meter.Id}
+     * @since 1.15.0
+     * @see #histogramFlavorPerMeter()
+     * @see #histogramFlavor()
+     */
+    default HistogramFlavor histogramFlavorPerMeter(Meter.Id id) {
+        return histogramFlavorPerMeter().getOrDefault(id.getName(), histogramFlavor());
```
shakuzen, line 263: As mentioned offline, I think we should make a type for the function that is this method's signature and make it configurable on the registry Builder rather than in the OtlpConfig. The reason is to keep logic in the registry and config in the config interface - this feels a bit too much like logic for how the config is used. Probably it would be most usable for most people for the default implementation to copy the longest starts-with match wins behavior they have in Spring Boot for some similar types of configuration properties. If that were the case, I suspect the vast majority of users would not want to customize the logic anyway, so the fact it needs to be done via the builder is not much of an issue. Thoughts?

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -497,6 +503,76 @@ static double[] getSloWithPositiveInf(DistributionStatisticConfig distributionSt
...
+         * name in the map that {@link OtlpConfig#histogramFlavorPerMeter()} returns. This
+         * means that {@link OtlpConfig#histogramFlavorPerMeter()} provides the data while
+         * this method provides the logic for the lookup, and you can override them
+         * independently.
+         * @param id the {@link Meter.Id} the {@link HistogramFlavor} is configured for
+         * @return the histogram flavor mapped to the {@link Meter.Id}
+         * @see OtlpConfig#histogramFlavorPerMeter()
+         * @see OtlpConfig#histogramFlavor()
+         */
+        HistogramFlavor getHistogramFlavor(OtlpConfig config, Meter.Id id);
```
jonatan-ivanov, line 545: Should we make this and `MaxBucketsPerMeterLookup#getMaxBuckets` `@Nullable`?
In the current case the lookup should provide a fallback, if we make it nullable we can do that for the user.

shakuzen: I think allowing nulls and falling back to the default value makes sense.

shakuzen: I've updated the signature along with incorporating this change such that the mapping function should not deal with the default value other than to return `null`. This feels cleaner than the expectation of implementations calling specific methods on OtlpConfig

jonatan-ivanov (approved):

shakuzen:
Final change: remove the API to customize the lookup logic and just change the logic so it is more usable, which solves the original ask without adding more API that doesn't seem necessary (yet).

shakuzen:
Thanks to @lenin-jaganathan for the offline review of these changes.

PR #6119, opened 2025-04-15
Add assertion for no re-registering Kafka metrics

izeye:
This PR adds an assertion for no re-registering Kafka metrics.

This PR also copies the `OutputCaptureExtension` and its support classes from Spring Boot to assert it. While copying it, I dropped the `AnsiOutput` support as it requires more classes that don't require for this.

See gh-6068


izeye:
@jonatan-ivanov Sure, I applied your suggestion in ffb5881.

jonatan-ivanov:
Thank you!

PR #6125, opened 2025-04-16
Add KeyName.withNoneValue()

izeye:
This PR adds `KeyName.withNoneValue()`.

This PR also changes to use it where possible.

### DIFF
--- a/micrometer-commons/src/main/java/io/micrometer/common/docs/KeyName.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/docs/KeyName.java
@@ -49,6 +49,7 @@ public interface KeyName {
     /**
      * Creates a key value for the given key name with {@value KeyValue#NONE_VALUE}.
      * @return key value
+     * @since 1.16.0
      */
     default KeyValue withoutValue() {
         return KeyValue.of(this, KeyValue.NONE_VALUE);
--- a/micrometer-commons/src/main/java/io/micrometer/common/docs/KeyName.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/docs/KeyName.java
@@ -47,6 +47,14 @@ public interface KeyName {
     }
 
     /**
+     * Creates a key value for the given key name with {@value KeyValue#NONE_VALUE}.
+     * @return key value
+     */
+    default KeyValue withoutValue() {
+        return KeyValue.of(this, KeyValue.NONE_VALUE);
+    }
+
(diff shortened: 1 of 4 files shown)


micrometer-commons/src/main/java/io/micrometer/common/docs/KeyName.java
```diff
@@ -46,6 +46,14 @@ default KeyValue withValue(String value) {
         return KeyValue.of(this, value);
     }
 
+    /**
+     * Creates a key value for the given key name with {@value KeyValue#NONE_VALUE}.
+     * @return key value
+     */
```
jonatan-ivanov, line 52: Could you please add `@since 1.16.0`?

micrometer-commons/src/main/java/io/micrometer/common/docs/KeyName.java
```diff
@@ -46,6 +46,14 @@ default KeyValue withValue(String value) {
...
+    default KeyValue withoutValue() {
```
jonatan-ivanov, line 53: I'm thinking about the name of this, what do you think about `withNoneValue`?
Not sure if it's better though and maybe `withoutValue` is the more user-friendly choice.
/cc @shakuzen

micrometer-commons/src/main/java/io/micrometer/common/docs/KeyName.java
izeye: @jonatan-ivanov Sure, I added it in d633e82.

micrometer-commons/src/main/java/io/micrometer/common/docs/KeyName.java
izeye: @jonatan-ivanov Thanks for the feedback!

Actually, I considered it initially, and I'm okay with either way, so just let me know which one you prefer.

shakuzen: I think `withNoneValue` is more clear. `withoutValue` to me sounds like a KeyValue with a `null` or otherwise missing value, which isn't possible, but that's what it sounds like it means.

izeye: @shakuzen Thanks for the feedback!

I renamed it in a4e4bfe.

jonatan-ivanov (approved):

PR #6138, opened 2025-04-18
(docs): add document for JMS instrumentation

AlexElin:
Relates to https://github.com/micrometer-metrics/micrometer/pull/5988#issuecomment-2693696191


shakuzen (approved):
Thanks a lot for the thorough documentation. I'll rebase onto `1.13.x` so we can get this added to the older docs as well.

PR #6141, opened 2025-04-19
Remove redundant Error Prone disabling for JDK 11 in micrometer-core

izeye:
This PR removes redundant Error Prone disabling for JDK 11 in the `micrometer-core`.


izeye:
Sorry for the noise.

I thought it's already covered by its global disabling, but it seems to be intentional as mentioned in https://github.com/micrometer-metrics/micrometer/pull/6072#discussion_r2020613696.

PR #6143, opened 2025-04-19
Resolve MissingOverride warnings from Error Prone

izeye:
This PR tries to resolve "MissingOverride" warnings from the Error Prone.


implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticConfig.java
```diff
@@ -44,13 +44,15 @@ public interface ElasticConfig extends StepRegistryConfig {
      * @param key Key to look up in the config.
      * @return Value for the key or null if no key is present.
      */
+    @Override
     @Nullable
     String get(String key);
```
izeye, line 49: I wanted to remove these, but I couldn't due to binary compatibility check failures.

I created https://github.com/micrometer-metrics/micrometer/pull/6145 instead.

PR #6145, opened 2025-04-19
Remove redundant get() methods from "Config" classes

izeye:
This PR removes redundant `get()` methods from "Config" classes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -399,7 +399,11 @@ subprojects {
                     ignoreMissingClasses = true
                     includeSynthetic = true
 
-                    classExcludes = []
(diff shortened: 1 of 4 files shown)


shakuzen (commented):
I agree, from a source code perspective, it's nicer to not have these overrides, but I worry if the binary incompatibility will be an issue for anyone.

shakuzen:
I think it'll be best to hold off on this change until 2.0 due to the binary incompatibility. I don't know if it's worth keeping the PR open as we don't yet know when we will do a 2.0. Perhaps it's better to open an issue to track this for 2.0.

PR #6070, opened 2025-03-28
Log warning about "function" meter re-registration

jonatan-ivanov:
>I also realized that while we added logging for gauge double registration, any of the "functional" meters are affected - FunctionCounter, FunctionTimer. That's relevant to KafkaMetrics because it only registers Gauges and FunctionCounters. /cc @jonatan-ivanov

found by @shakuzen, see: https://github.com/micrometer-metrics/micrometer/issues/5757#issuecomment-2760837057

See gh-5616
See gh-5688
See gh-5617

PR #6146, opened 2025-04-19
MultiGauge doesn't work with MeterFilter.map()

izeye:
The `MultiGauge` doesn't seem to work with the `MeterFilter.ignoreTags()`.

See https://github.com/micrometer-metrics/micrometer/pull/6070/files#r2051508975

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -34,7 +34,6 @@ import io.micrometer.core.instrument.util.TimeUtils;
 
 import java.time.Duration;
 import java.util.*;
-import java.util.Locale;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
@@ -623,7 +622,15 @@ public abstract class MeterRegistry {
         return meterClass.cast(m);
     }
 
-    private Meter.Id getMappedId(Meter.Id id) {
+    Meter.Id getMappedId(Meter.Id id) {
+        Meter m = preFilterIdToMeterMap.get(id);
+        if (m != null && !isStaleId(id)) {
+            return m.getId();
+        }
+        return mapId(id);
+    }
+
+    private Meter.Id mapId(Meter.Id id) {
         if (id.syntheticAssociation() != null) {
             return id;
         }
@@ -644,7 +651,7 @@ public abstract class MeterRegistry {
             return m;
         }
 
-        Meter.Id mappedId = getMappedId(originalId);
+        Meter.Id mappedId = mapId(originalId);
         m = meterMap.get(mappedId);
 
         if (m != null) {
@@ -771,7 +778,7 @@ public abstract class MeterRegistry {
     public Meter removeByPreFilterId(Meter.Id preFilterId) {
         final Meter meterToRemove = preFilterIdToMeterMap.get(preFilterId);
         if (meterToRemove == null)
-            return remove(getMappedId(preFilterId));
+            return remove(mapId(preFilterId));
         return remove(meterToRemove);
     }
 
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/MultiGaugeTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/MultiGaugeTest.java
@@ -162,6 +162,25 @@ class MultiGaugeTest {
         testOverwrite("prefix.my.multi.gauge");
     }
 
+    @Test
+    void withMeterFilterIgnoreTags() {
+        registry.config().meterFilter(MeterFilter.ignoreTags("ignored"));
+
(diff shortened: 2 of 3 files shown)


jonatan-ivanov:
Thank you for finding this bug and proving it with a test.
I pushed a hack that I don't particularly like because:
1. It extends the visibility of `getMappedId` from `private` to `package-private` (less of an issue I think)
2. It adds an extra execution of `MeterFilter`s, also see the commit message

I'm continue working on it to see if I can find a better solution (how to integrate this with the `MeterFilter` caching mechanism of the `MeterRegistry`).

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -769,7 +777,7 @@ public Meter remove(Meter meter) {
     public Meter removeByPreFilterId(Meter.Id preFilterId) {
         final Meter meterToRemove = preFilterIdToMeterMap.get(preFilterId);
         if (meterToRemove == null)
-            return remove(getMappedId(preFilterId));
+            return remove(mapId(preFilterId));
```
jonatan-ivanov, line 780: This was the result of just blind refactor but don't we need the `isStaleId` check here?
If so, we should call the new `getMappedId` method which also does that check.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -622,7 +622,15 @@ private <M extends Meter> M registerMeterIfNecessary(Class<M> meterClass, Meter.
         return meterClass.cast(m);
     }
 
-    private Id getMappedId(Id id) {
+    Id getMappedId(Id id) {
```
jonatan-ivanov, line 625: This is the same as https://github.com/micrometer-metrics/micrometer/pull/6146/files#diff-8dca70a0fea53228353bc7f6497be0f243f2370d8301f435c121336eabc2e8dfR648-R652
but without the double registration warning.

I'm thinking if we should have a version of this that does that (depending on a boolean flag).

jonatan-ivanov: The link does not jump, it's line 648-652 in the changed code.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: Maybe. It'd take some time and effort to get back up to speed on this part of the code. I vaguely remember fixing some issues around removing and stale things, though. I've got a few other things distracting me at the moment. It's probably easiest to make a unit test that fails than to reason the other way around.

jonatan-ivanov: TL;DR: ignore me. :)

I dug deeper and I think we should not add the stale id check.

If I add a new filter to make the id stale, the next meter registration will handle the stale ids so `removeByPreFilterId` will not see any stale ids.

If there is no meter registration between adding the filter and the `removeByPreFilterId` call, this method should carry on without checking the stale ids since nothing added a new meter since then. In fact, if it does check if the id is stale, it will not remove meters that it otherwise could have.

micrometer-core/src/test/java/io/micrometer/core/instrument/MultiGaugeTest.java
```diff
@@ -162,6 +162,52 @@ public Meter.Id map(Meter.Id id) {
...
+        multiGauge.register(List.of(Row.of(Tags.of("key", "1", "ignored", "3"), 3d)), true);
+        assertThat(registry.get("mg").tag("key", "1").gauges()).hasSize(1);
+        assertThat(registry.get("mg").tag("key", "1").gauge().value()).isEqualTo(3d);
+    }
+
+    @Test
+    void withMeterFilterMapTags() {
+        registry.config()
+            .meterFilter(
+                    MeterFilter.replaceTagValues("key", value -> Integer.parseInt(value) % 2 == 0 ? "even" : "odd"));
```
shakuzen, line 188: Is this second test added testing a different code path? As far as I can tell it is testing the same code path as the above test. I don't think we need both. They both test the case where ID `a` which maps to `c` is registered followed by ID `b` which also maps to `c` being subsequently registered.

shakuzen (approved):
The changes look good to me. I may follow up post merge with some general polish to the MultiGauge code to try to name things in a way that hopefully makes the logic easier to read. It took a bit of re-reading and taking notes to make sure I understood what was happening before and what's happening now.

micrometer-core/src/test/java/io/micrometer/core/instrument/MultiGaugeTest.java
jonatan-ivanov: The code path in `MultiGauge` and `MeterRegistry` is the same, the use-case/transformation is different: the previous test removes a tag, this one modifies its value so the result is the same (same set of tags), how we get there is different.
I think it's ok to remove it.

jonatan-ivanov:
Thank you @izeye!

PR #6147, opened 2025-04-20
Replace Meter.Id.getTags() with cheaper alternatives

izeye:
This PR replaces `Meter.Id.getTags()` with cheaper alternatives.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
@@ -239,7 +239,7 @@ public interface Meter {
          * @since 1.1.0
          */
         public Id withTags(Iterable<Tag> tags) {
-            return new Id(name, Tags.concat(getTags(), tags), baseUnit, description, type);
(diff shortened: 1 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
```diff
@@ -239,7 +239,7 @@ public Id withTag(Tag tag) {
          * @since 1.1.0
          */
         public Id withTags(Iterable<Tag> tags) {
-            return new Id(name, Tags.concat(getTags(), tags), baseUnit, description, type);
+            return new Id(name, Tags.concat(this.tags, tags), baseUnit, description, type);
```
izeye, line 242: I wonder if it's worth exposing `tags` as `Tags` to save `Tags` creation.

shakuzen: I had a similar thought when looking into something else. We're working on planning for 1.16 work now, and I'm hoping I'll have some time to holistically look at this type of performance related issue where we're doing conversion in many places perhaps unnecessarily.

PR #6149, opened 2025-04-21
Remove Spring Boot 2 samples

shakuzen:
Spring Boot 2 is out of OSS support at this point, and the latest versions of Micrometer are used with Spring Boot 3, so it is odd to have samples with the latest version of Micrometer using Spring Boot 2. Therefore, I don't think they add much value and can be removed. There is no code run as part of the build (only compilation), and we should try to eliminate dependency cycles: these samples depend on Spring but Spring depends on Micrometer. If we still want Spring-based samples, a separate repository would be a better setup.


jonatan-ivanov (approved):
I was thinking about two things while reviewing this:
1. Should we migrate some of these to the micrometer-samples repo?
2. I think we should add a link to the readme about the samples repo

shakuzen:
> 1. Should we migrate some of these to the micrometer-samples repo?

I'm not sure we should since Boot 2 is out of OSS support, so it doesn't make sense to maintain a sample with that to me. Let me know if I'm missing something.

> 2\. I think we should add a link to the readme about the samples repo

Good idea

jonatan-ivanov:
>I'm not sure we should since Boot 2 is out of OSS support, so it doesn't make sense to maintain a sample with that to me. Let me know if I'm missing something.

By migrating I was thinking creating a Boot 3 based example, since here we had example apps for almost every registry while in the sample repo we only have prometheus.

>Good idea

👍🏼, done in https://github.com/micrometer-metrics/micrometer/issues/6155

shakuzen:
> since here we had example apps for almost every registry while in the sample repo we only have prometheus.

I'm not sure what it would look like in the samples repo. I think it depends what the intended purpose is. I think "samples" keep ending up having a nebulous purpose, which confuses both the maintainers and the users. Is it supposed to be something that users should look to as a model of what they should do in their applications? Is it for running automated integration tests? Is it for ad hoc local testing? This sample was for ad hoc local testing, which I suppose users could have used, but it was really for maintainers since it runs with the checked out Micrometer source rather than a released version. The setup is not how a typical user should configure their application, with every registry on the classpath and all disabled by default.

What did you have in mind for the samples repo? We already have many Boot 3 based samples there. It feels overly redundant to copy any of them and only change the registry, but it'd also make them less like a sample (to me) if they had a complicated setup to switch registries that's not a natural thing users would typically have in their applications.

jonatan-ivanov:
I was thinking about having an app for each registry but also, to make them work we need a backend so probably it does not worth the hassle.

PR #6151, opened 2025-04-21
Polish

izeye:
This PR polishes a bit.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
@@ -245,8 +245,7 @@ public interface Meter {
         /**
          * Generate a new id replacing all tags with new ones.
          * @param tags The tags to add.
-         * @return A new id with the only the provided tags. The source id remains
-         * unchanged.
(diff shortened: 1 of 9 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
```diff
@@ -245,8 +245,7 @@ public Id withTags(Iterable<Tag> tags) {
         /**
          * Generate a new id replacing all tags with new ones.
          * @param tags The tags to add.
-         * @return A new id with the only the provided tags. The source id remains
-         * unchanged.
+         * @return A new id with the only provided tags. The source id remains unchanged.
```
jonatan-ivanov, line 248: Should not this be:
```suggestion
         * @return A new id with only the provided tags. The source id remains unchanged.
```

or:
```suggestion
         * @return A new id with the provided tags. The source id remains unchanged.
```

izeye: @jonatan-ivanov Thanks for the suggestion! I applied it in 5539fac.

Issue #5547, opened 2024-10-02
Migrate to JSpecify annotations for nullability constraints


shakuzen:
Deprecate our internal nullability annotations that are meta-annotated with JSR-305 annotations, add a dependency on JSpecify, update nullability imports and annotation usage to match correct use of JSpecify. See https://jspecify.dev/docs/using/

Breaking the work down into parts to make review more manageable:
- https://github.com/micrometer-metrics/micrometer/pull/6158
- https://github.com/micrometer-metrics/micrometer/pull/6270
- https://github.com/micrometer-metrics/micrometer/pull/6296
- https://github.com/micrometer-metrics/micrometer/pull/6385
- https://github.com/micrometer-metrics/micrometer/pull/6420
- https://github.com/micrometer-metrics/micrometer/pull/6422

shakuzen:
The migration should be complete now with all of the linked PRs, although there are some follow-up issues we have opened.

PR #6158, opened 2025-04-23
Migrate commons/observation modules to jSpecify

shakuzen:
Replaces our JSR-305-based nullability annotations with the jSpecify annotations and adds NullAway compile-time analysis of nullability issues.

This PR migrates micrometer-commons and micrometer-observation. In order to keep review manageable, it may be best to merge this before proceeding with the migration of the other modules.

### Important!
- [x] branch for `1.15.x` before merging this to `main`.

See https://github.com/micrometer-metrics/micrometer/issues/5547

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,6 +73,11 @@ subprojects {
 
     if (project.name != 'micrometer-bom') {
         tasks.withType(JavaCompile).configureEach {
+            if (it.name == "compileJava" && !(it.project.name in ["micrometer-java11"])) {
+                options.errorprone.disable(
+                        "JavaDurationGetSecondsToToSeconds" // Requires JDK 9+
+                )
+            }
             options.errorprone {
                 disableWarningsInGeneratedCode = true
                 excludedPaths = ".*/build/generated/.*"
@@ -80,11 +85,6 @@ subprojects {
                 disable(
                         "StringConcatToTextBlock" // Requires JDK 15+
                 )
-                if (it.name == "compileJava" && !(it.project.name in ["micrometer-java11"])) {
-                    disable(
-                            "JavaDurationGetSecondsToToSeconds" // Requires JDK 9+
-                    )
-                }
 
                 error("BadImport",
                         "DefaultCharset",
--- a/build.gradle
+++ b/build.gradle
@@ -97,6 +97,7 @@ subprojects {
                 option("NullAway:OnlyNullMarked", "true")
                 option("NullAway:CustomContractAnnotations", "io.micrometer.common.lang.internal.Contract")
                 option("NullAway:CheckContracts", "true")
+                option("NullAway:HandleTestAssertionLibraries", "true")
                 if (!javaLanguageVersion.canCompileOrRun(17)) {
                     // Error Prone does not work with JDK <17
                     enabled = false
--- a/build.gradle
+++ b/build.gradle
@@ -73,31 +73,37 @@ subprojects {
 
     if (project.name != 'micrometer-bom') {
         tasks.withType(JavaCompile).configureEach {
-            options.errorprone.disableWarningsInGeneratedCode = true
-            options.errorprone.excludedPaths = ".*/build/generated/.*"
-            options.errorprone.error(
-                    "BadImport",
-                    "DefaultCharset",
-                    "LongDoubleConversion",
-                    "MissingOverride",
-                    "StringCaseLocaleUsage"
-            )
-            options.errorprone.disable(
-                    "StringConcatToTextBlock" // Requires JDK 15+
-            )
+            options.errorprone {
+                disableWarningsInGeneratedCode = true
+                excludedPaths = ".*/build/generated/.*"
 
-            if (it.name == "compileJava" && !(it.project.name in ["micrometer-java11"])) {
-                options.errorprone.disable(
-                        "JavaDurationGetSecondsToToSeconds" // Requires JDK 9+
+                disable(
+                        "StringConcatToTextBlock" // Requires JDK 15+
                 )
-            }
+                if (it.name == "compileJava" && !(it.project.name in ["micrometer-java11"])) {
+                    disable(
+                            "JavaDurationGetSecondsToToSeconds" // Requires JDK 9+
+                    )
+                }
 
-            if (!javaLanguageVersion.canCompileOrRun(17)) {
-                // Error Prone does not work with JDK <17
-                options.errorprone.enabled = false
-            }
-            if (System.env.CI != null) {
-                options.errorprone.disableAllWarnings = true
+                error("BadImport",
+                        "DefaultCharset",
+                        "LongDoubleConversion",
+                        "MissingOverride",
+                        "NullAway",
+                        "StringCaseLocaleUsage"
+                )
+
+                option("NullAway:OnlyNullMarked", "true")
+                option("NullAway:CustomContractAnnotations", "io.micrometer.common.lang.internal.Contract")
+                option("NullAway:CheckContracts", "true")
+                if (!javaLanguageVersion.canCompileOrRun(17)) {
+                    // Error Prone does not work with JDK <17
+                    enabled = false
+                }
+                if (System.env.CI != null) {
+                    disableAllWarnings = true
+                }
             }
         }
         if ((project.name.contains('samples') && !project.name.contains('kotlin')) || project.name.contains('benchmarks')) {
@@ -105,6 +111,8 @@ subprojects {
         } else {
             apply plugin: 'java-library'
             dependencies {
+                api(libs.jspecify)
+
                 testImplementation platform(libs.junitBom)
                 testImplementation libs.junitJupiter
                 testRuntimeOnly libs.junitPlatformLauncher
@@ -139,6 +147,7 @@ subprojects {
             optionalApi libs.jsr305
             checkstyle libs.spring.javaformatCheckstyle
             errorprone(libs.errorprone)
+            errorprone(libs.nullAway)
         }
 
         tasks {
--- a/build.gradle
+++ b/build.gradle
@@ -98,6 +98,11 @@ subprojects {
                 option("NullAway:CustomContractAnnotations", "io.micrometer.common.lang.internal.Contract")
                 option("NullAway:CheckContracts", "true")
                 option("NullAway:HandleTestAssertionLibraries", "true")
+                if (javaLanguageVersion.canCompileOrRun(22)) {
+                    // see https://bugs.openjdk.org/browse/JDK-8346471
+                    // see https://github.com/uber/NullAway/wiki/JSpecify-Support
+                    option("NullAway:JSpecifyMode", "true")
+                }
                 if (!javaLanguageVersion.canCompileOrRun(17)) {
                     // Error Prone does not work with JDK <17
                     enabled = false
--- a/micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
@@ -15,7 +15,8 @@
  */
 package io.micrometer.common;
 
-import io.micrometer.common.lang.Nullable;
+import io.micrometer.common.lang.internal.Contract;
+import org.jspecify.annotations.Nullable;
 
 import java.util.*;
 import java.util.function.Function;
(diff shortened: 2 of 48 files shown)


micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
```diff
@@ -409,7 +410,8 @@ public static KeyValues of(@Nullable String... keyValues) {
         return toKeyValues(keyValueArray);
     }
 
-    private static boolean blankVarargs(@Nullable Object[] args) {
+    @Contract("null -> true")
```
shakuzen, line 413: Without this `@Contract`, NullAway fails to recognize that a Nullable parameter has been null-checked and is safe to pass as a parameter to NonNullable marked methods.

micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
```diff
@@ -16,7 +16,7 @@
 package io.micrometer.common.annotation;
 
 import io.micrometer.common.KeyValue;
-import io.micrometer.common.lang.Nullable;
+import org.jspecify.annotations.Nullable;
```
shakuzen, line 19: Import order should be fixed. I did a find and replace, so it came out like this.

micrometer-commons/src/main/java/io/micrometer/common/lang/internal/Contract.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.common.lang.internal;
+
+import java.lang.annotation.*;
+
+/**
+ * This is for internal use only.
```
shakuzen, line 21: Maybe worth writing more here, but I wanted to be very clear this is not supported and should not be used outside of Micrometer. My hope is that we can remove it in the future, and so I want to be clear this is not subject to normal compatibility guarantees. Maybe we can even mark is deprecated if that doesn't make it annoying for us to use.

micrometer-commons/src/main/java/io/micrometer/common/lang/package-info.java
```diff
@@ -13,6 +13,4 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@NonNullApi
-@NonNullFields
```
shakuzen, line 17: I suppose it may be better to make this `@NullMarked` even though we don't have any nullability in the annotations in this package.

shakuzen:
The build failure right now is a bug in NullAway. See https://github.com/uber/NullAway/issues/1184. It has been fixed but a new version hasn't been released yet. The build failure will go away once we can upgrade to a new version of NullAway or we migrate the modules containing code NullAway is tripping on.

micrometer-observation/src/test/java/io/micrometer/observation/docs/ObservationDocumentationTests.java
```diff
@@ -65,6 +67,17 @@ void iaeShouldBeThrownWhenDocumentedObservationHasOverriddenDefaultConventionBut
...
     }
 
+    @Test
+    void whenGetNameIsNotOverridden_thenNameShouldBeTakenFromDefaultConvention() {
+        ObservationRegistry registry = observationRegistry();
+        Observation.Context context = new Observation.Context();
+
+        TestConventionObservation.OVERRIDDEN.observation(registry, () -> context).start().stop();
+
+        then(context.getName()).isEqualTo("one");
```
shakuzen, line 77: Moving this to a separate issue to consider to not block progress on the nullability migration.

micrometer-observation/src/main/java/io/micrometer/observation/NoopObservationConfig.java
```diff
@@ -48,7 +50,7 @@ public ObservationRegistry.ObservationConfig observationPredicate(ObservationPre
     }
 
     @Override
-    public boolean isObservationEnabled(String name, Observation.Context context) {
+    public boolean isObservationEnabled(@Nullable String name, Observation.@Nullable Context context) {
```
shakuzen, line 53: These are `@Nullable` in the method this overrides.

micrometer-observation/src/main/java/io/micrometer/observation/NoopObservationRegistry.java
shakuzen, line 1: This is aligning the nullability here with the method signatures being overridden.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -754,7 +754,7 @@ default <T> T scoped(Supplier<T> supplier) {
      * @return the result from {@link CheckedCallable#call()}
      */
     @SuppressWarnings("unused")
-    default <T, E extends Throwable> T scopedChecked(CheckedCallable<T, E> checkedCallable) throws E {
+    default <T, E extends Throwable> @Nullable T scopedChecked(CheckedCallable<T, E> checkedCallable) throws E {
```
shakuzen, line 757: Hmm I'm not sure how this should be declared. I guess the nullability on the generic type `T` on `CheckedCallable` should match the return type nullability.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -142,7 +140,7 @@ public Object observeMethod(ProceedingJoinPoint pjp) throws Throwable {
         return observe(pjp, method, observed);
     }
 
-    private Object observe(ProceedingJoinPoint pjp, Method method, Observed observed) throws Throwable {
+    private @Nullable Object observe(ProceedingJoinPoint pjp, Method method, Observed observed) throws Throwable {
```
shakuzen, line 143: I suppose we don't know the nullability of the instrumented method's return value. Should we make it `@NullUnmarked`?

shakuzen (commented):
Adding some notes that hopefully makes review easier.

micrometer-commons/src/main/java/io/micrometer/common/lang/internal/Contract.java
jonatan-ivanov: I think this is fine. I'm thinking if we should add a link in a non-javadoc comment to nullaways' doc about this in case someone is working on Micrometer's codebase and wanders here.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
jonatan-ivanov: I think this is the way. :)

jonatan-ivanov: Should we do the same for `observe`, `observeChecked`, etc?

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -933,15 +933,15 @@ class Context implements ContextView {
...
+        public @Nullable String getName() {
             return this.name;
         }
 
         /**
          * Sets the observation name.
          * @param name observation name
          */
-        public void setName(String name) {
+        public void setName(@Nullable String name) {
```
jonatan-ivanov, line 944: Is it valid? I mean if you set the name should you set it as `null`?
(I guess somewhere we use this and pass `null`?)

But now I think it's symmetric with `setContextualName`.

micrometer-observation/src/main/java/io/micrometer/observation/Observations.java
```diff
@@ -61,6 +61,8 @@ public static ObservationRegistry getGlobalRegistry() {
         return globalRegistry;
     }
 
+    // We know delegate.get() cannot be null but NullAway does not.
+    @SuppressWarnings("NullAway")
```
jonatan-ivanov, line 65: Is this a bug we should report?
Also, can `@SuppressWarnings` go on the `delegate`?

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
jonatan-ivanov: Semantically, the difference would be this?
`@Nullable`: we know that it can return `null`.
`@NullUnmarked`: we have no idea (it can but there will be no warning)

Let's discuss this, I feel we have a lot of similar use cases (`Observation.observe`, `Timer.record`, etc. where we accept something from the user we call and return its value and it's up to the user if that something can return `null` or not).

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspectObservationDocumentation.java
```diff
@@ -17,14 +17,15 @@
 
 import io.micrometer.common.KeyValues;
 import io.micrometer.common.docs.KeyName;
-import io.micrometer.common.lang.Nullable;
+
```
jonatan-ivanov, line 20: new line?

micrometer-observation/src/main/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessor.java
```diff
@@ -49,6 +50,7 @@ public class ObservationThreadLocalAccessor implements ThreadLocalAccessor<Obser
      */
     private ObservationRegistry observationRegistry = ObservationRegistry.create();
 
+    @SuppressWarnings("NullAway.Init")
```
jonatan-ivanov, line 53: Shouldn't we mark this as `@Nullable`?

micrometer-commons/src/main/java/io/micrometer/common/util/StringUtils.java
```diff
@@ -48,6 +49,7 @@ public static boolean isBlank(@Nullable String string) {
      * @param string String to check
      * @return {@code true} if the String has any non-whitespace character
      */
+    @Contract("null -> false")
```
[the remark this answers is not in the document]
shakuzen: Not yet at least 😁

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
shakuzen: I'm willing to be convinced otherwise, but semantically it feels like using `@NullUnmarked` on the return type is the most accurate. I can imagine a user complaining if we mark it as `@Nullable` but they know in their usage it isn't nullable because the Callable/Supplier/Function they pass does not have a nullable return value.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
shakuzen: I think you're right it shouldn't be `@Nullable`. We need to change `SimpleObservation`'s package-private constructor arg for name to be not nullable for that, which is fine since it is package private and all call sites don't pass nullable values. I'll update the PR.

micrometer-observation/src/main/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessor.java
shakuzen: It's a bit of an unusual setup, but I think the idea is that it's lazily initialized and non-null after initialization. We could defensively mark it as `@Nullable` but I thought this better expresses the intent.

micrometer-observation/src/main/java/io/micrometer/observation/Observations.java
shakuzen: AtomicReference is a JDK class, so we can't directly control its specified internal nullability. NullAway is considering that `AtomicReference.get` returns a nullable value, it seems. Ideally, by not marking the generic type as `@Nullable` we'd hope it is treated as `@NonNull` due to the package containing the field being annotated `@NullMarked`. It might be related to https://github.com/uber/NullAway/issues/1161, but I'm not sure.

Moving the suppression to the field doesn't help because the compiler error is thrown from the `delegate.get` call sites.

build.gradle
```diff
@@ -73,38 +73,52 @@ subprojects {
...
+                )
+
+                option("NullAway:OnlyNullMarked", "true")
+                option("NullAway:CustomContractAnnotations", "io.micrometer.common.lang.internal.Contract")
+                option("NullAway:CheckContracts", "true")
+                option("NullAway:HandleTestAssertionLibraries", "true")
+                if (javaLanguageVersion.canCompileOrRun(22)) {
+                    // see https://bugs.openjdk.org/browse/JDK-8346471
+                    // see https://github.com/uber/NullAway/wiki/JSpecify-Support
+                    option("NullAway:JSpecifyMode", "true")
```
shakuzen, line 104: I added enabling JSpecifyMode on JDK 22+. It didn't result in any more compilation errors.

Issue #3144, opened 2022-04-28
Support min/max in OTLP histograms


shakuzen:
We do not currently ship min/max (though we do ship an infinite bucket which is the max) in OTLP histograms. At this time, we cannot add them because we are waiting for a release of opentelemetry-proto (see https://github.com/open-telemetry/opentelemetry-proto/issues/387). When we do add them, we will need to figure out the right semantics of what min and max will mean. See https://github.com/micrometer-metrics/micrometer/pull/3129#discussion_r850688499_ for some initial thoughts.

_Originally posted by @twicksell in https://github.com/micrometer-metrics/micrometer/pull/3129#discussion_r850688499_

shakuzen:
This should be unblocked now with this release: https://github.com/open-telemetry/opentelemetry-proto-java/releases/tag/v0.17.0

sspieker:
> This should be unblocked now with this release: https://github.com/open-telemetry/opentelemetry-proto-java/releases/tag/v0.17.0

Hi @shakuzen , I agree.
I just recently took interest in OpenTelemetry, and the way I see it min/max support in histograms would be a valuable addition.

shakuzen:
The open question is what will the min/max mean. We currently only support cumulative temporality (https://github.com/micrometer-metrics/micrometer/issues/3145 to support delta), which means min/max should be the min/max of all values observed in the interval, which will be since the meter was first registered (generally about when the application started). For long-running applications, this generally becomes less useful over time. Max values in Micrometer in other registries use a time-windowed maximum for this reason.

The [OTLP spec on histograms](https://opentelemetry.io/docs/reference/specification/metrics/datamodel/#histogram) mentions

> The max of all values in the histogram.

> The aggregation temporality also has implications on the min and max fields. Min and max are more useful for Delta temporality, since the values represented by Cumulative min and max will stabilize as more events are recorded. Additionally, it is possible to convert min and max from Delta to Cumulative, but not from Cumulative to Delta. When converting from Cumulative to Delta, min and max can be dropped, or captured in an alternative representation such as a gauge.

I think from a usefulness perspective, when using cumulative temporality, we should publish our time-windowed max as a separate gauge and not set max on the histogram.
If/when we support delta temporality, we could set max on the histogram, but our time-windowed max intentionally does not match the same interval as other metrics published in the step (see https://github.com/micrometer-metrics/micrometer/issues/1993).

[...]

PR #6159, opened 2025-04-23
Support max in OTLP histogram 

BenEfrati:
Create gauge for Timer and DistributionSummary

Closes https://github.com/micrometer-metrics/micrometer/issues/3144


BenEfrati:
@shakuzen , Can you pleases review?

BenEfrati:
Ok, let me know if you want me to add a condition to prevent publishing the separate max Gauge when delta temporality is active.

shakuzen:
@BenEfrati as a user of OtlpMeterRegistry, what do you think makes the most sense? Are you using cumulative temporality and that's why you'd like this change?

I do think it's a little unfortunate that the way to query the max on your metrics will change depending on whether you're using delta or cumulative aggregation, but it also makes sense because the max has different semantics in the different places. I am interested in hearing the perspective of users on this.

BenEfrati:
@shakuzen Yes, that's correct: I am currently using cumulative temporality because our target backend is Prometheus. 
For Prometheus, publishing the time-windowed max as a separate Gauge will work perfectly.

I have not personally used the [deltatocumulativeprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/deltatocumulativeprocessor/README.md) when pushing to Prometheus, so I cannot comment on its specific behavior.

Maybe this is useful for other backends that supports delta temporality, not sure if there is an option to query the max value.

We can introduce a configuration option. This would allow users to explicitly choose whether to create the max Gauge, decoupling that choice from the `AggregationTemporality` setting.

shakuzen:
> We can introduce a configuration option. This would allow users to explicitly choose whether to create the max Gauge, decoupling that choice from the `AggregationTemporality` setting.

That's an interesting idea. It would eliminate the issue of doing or not doing something people do or don't expect, but it adds some complexity. We try to avoid adding configuration options if we can avoid it, but this may be a case where it is worth it. Something to potentially consider is using a `Boolean` where `null` can mean "default behavior" which is a separate gauge for cumulative aggregation and no separate gauge for delta aggregation. I'll discuss with @jonatan-ivanov but feedback from users is still appreciated.

jonatan-ivanov:
I have a hard time to get to a conclusion about what would be the best (since cumulative max is not very useful in OTLP).
But I think so far, the best option is what @shakuzen mentioned above:
- default behavior: separate gauge for cumulative only
- the user can explicitly enable/disable the separate gauge

We should also write docs to explain the OTLP behavior, the problems with its cumulative max and the default behavior above.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
```diff
@@ -126,6 +126,8 @@ private void writeHistogramSupport(HistogramSupport histogramSupport) {
         double max = isTimeBased ? histogramSnapshot.max(baseTimeUnit) : histogramSnapshot.max();
         long count = histogramSnapshot.count();
 
+        addMaxGaugeForTimer(id, tags, max);
```
jonatan-ivanov: `addMaxGaugeForHistogramSupport`?
Since this is for both `Timer` and `DistributionSummary`.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
```diff
@@ -154,6 +156,20 @@ private static Optional<ExponentialHistogramSnapShot> getExponentialHistogramSna
...
 
+    private void addMaxGaugeForTimer(Meter.Id id, Iterable<KeyValue> tags, double max) {
+        String metricName = id.getName() + ".max";
+        Metric.Builder metricBuilder = getOrCreateMetricBuilder(id.withName(metricName), DataCase.GAUGE);
+        if (!metricBuilder.hasGauge()) {
+            metricBuilder.setGauge(io.opentelemetry.proto.metrics.v1.Gauge.newBuilder());
+        }
+
+        metricBuilder.getGaugeBuilder()
+            .addDataPoints(NumberDataPoint.newBuilder()
```
jonatan-ivanov, line 173: It seems `addDataPoints` has an overload for `NumberDataPointBuilder` and `NumberDataPoint` too?
It seems `writeGauge` (which is quite similar to this) uses `NumberDataPoint` (calls `.build()`) while this one does not. Should not be the two consistent?

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
```diff
@@ -107,11 +107,13 @@ void collectorShouldExportMetrics() throws Exception {
             // see: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/20519
             matchesPattern("(?s)^.*test_timer_milliseconds_sum\\{.+} 123\\.0\\n.*$"),
             matchesPattern("(?s)^.*test_timer_milliseconds_bucket\\{.+,le=\"\\+Inf\"} 1\\n.*$"),
+            matchesPattern("(?s)^.*test_timer_max_milliseconds\\{.+} 123\\.0\n.*$"),
```
jonatan-ivanov, line 116: Since this is a separate gauge, there should be two more lines, something like this should cover them:
```java
containsString("# HELP test_timer_max_milliseconds \n"),
containsString("# TYPE test_timer_max_milliseconds gauge\n"),
```

shakuzen:
> We should also write docs to explain the OTLP behavior, the problems with its cumulative max and the default behavior above.

I think we can just link to the OTLP docs on it. See https://github.com/micrometer-metrics/micrometer/issues/3144#issuecomment-1165274503 where I link to it and quote the relevant section.

shakuzen:
The real bikeshed issue is going to be what to name the configuration method on `OtlpConfig`. `enableSeparateMaxGauge`?

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
BenEfrati: Done

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
BenEfrati: Fixed.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
BenEfrati: Done

BenEfrati:
> The real bikeshed issue is going to be what to name the configuration method on `OtlpConfig`. `enableSeparateMaxGauge`?

`publishMaxGaugeForHistograms` ?
https://github.com/micrometer-metrics/micrometer/blob/465a98faf85b3c206e4ddb822ea70bc5829ff944/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java#L286

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -275,6 +275,18 @@ default Map<String, HistogramFlavor> histogramFlavorPerMeter() {
...
+    /**
+     * Whether to publish a separate gauge for the max value of histogram-based meters
+     * (Timer, DistributionSummary, LongTaskTimer). When enabled, a gauge with the suffix
+     * {@code .max} will be created for each histogram meter, reporting the maximum value
+     * from the histogram snapshot.
+     * @return true if max gauge should be published; default is true
+     * @since 1.16.3
+     */
+    default boolean publishMaxGaugeForHistograms() {
+        return getBoolean(this, "publishMaxGaugeForHistograms").orElse(true);
```
shakuzen: I'm still not sure on the name - I'll think on it more. However, this doesn't incorporate the feedback discussed in https://github.com/micrometer-metrics/micrometer/pull/6159#issuecomment-3384788856 and https://github.com/micrometer-metrics/micrometer/pull/6159#issuecomment-3387742190. The key part being the default behavior depends on the aggregation temporality but it can be explicitly overridden by users. The logic could be in a lambda in a `orElseGet` call here with `boolean` return type. Though there may be benefits to users being able to explicitly configure the default behavior by setting this as `null` with a `Boolean` return type.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -275,6 +275,21 @@ default Map<String, HistogramFlavor> histogramFlavorPerMeter() {
...
 
+    /**
+     * Whether to publish a separate gauge for the max value of histogram-based meters
+     * (Timer, DistributionSummary, LongTaskTimer). When enabled, a gauge with the suffix
+     * {@code .max} will be created for each histogram meter, reporting the maximum value
+     * from the histogram snapshot.
+     * @return true if max gauge should be published; default is true for
+     * {@link AggregationTemporality#CUMULATIVE} and false for
+     * {@link AggregationTemporality#DELTA}
+     * @since 1.16.3
```
jonatan-ivanov: This should be `1.17.0`:
```suggestion
     * @since 1.17.0
```

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
```diff
@@ -167,6 +172,49 @@ void collectorShouldExportMetricsWithGzipCompression() throws Exception {
...
+        whenPrometheusScraped().then().body(
+            containsString("# HELP test_ds_nomax \n"),
+            containsString("# TYPE test_ds_nomax histogram\n"),
+            matchesPattern("(?s)^.*test_ds_nomax_count\\{.+} 1\\n.*$"),
+            matchesPattern("(?s)^.*test_ds_nomax_sum\\{.+} 24\\.0\\n.*$"),
+            matchesPattern("(?s)^.*test_ds_nomax_bucket\\{.+,le=\"\\+Inf\"} 1\\n.*$")
+        );
+
+        // Verify .max gauges are NOT exported
+        org.assertj.core.api.Assertions.assertThat(body)
```
jonatan-ivanov: Could you please static import this?

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMeterRegistryTest.java
```diff
@@ -206,14 +206,24 @@ void distributionWithPercentileShouldWriteSummary() {
...
+        });
+        List<Metric> timerExpoMetrics = writeToMetrics(timer.register(registryWithExponentialHistogram));
+        assertThat(timerExpoMetrics).filteredOn(Metric::hasSummary).singleElement().satisfies(summary -> {
+            assertThat(summary.getDataCase().getNumber()).isEqualTo(Metric.DataCase.SUMMARY.getNumber());
+        });
+        List<Metric> dsExpoMetrics = writeToMetrics(ds.register(registryWithExponentialHistogram));
+        assertThat(dsExpoMetrics).filteredOn(Metric::hasSummary).singleElement().satisfies(summary -> {
+            assertThat(summary.getDataCase().getNumber()).isEqualTo(Metric.DataCase.SUMMARY.getNumber());
+        });
+        assertMaxGaugeMetrics(timerMetrics, dsMetrics, timerExpoMetrics, dsExpoMetrics);
```
jonatan-ivanov: If this fails it will not as simple figuring out what the problem was since there are four things checked at the same time. I would remove the varargs from the "assert" method so it would verify a single list instead of multiple ones:
```java
assertMetricsHaveMaxGauge(timerMetrics);
```

shakuzen (approved):
I will follow-up with some polishing, but this looks good. Thanks for all the updates and work on it.

PR #6162, opened 2025-04-24
Resolve StringCaseLocaleUsage warnings from Error Prone

izeye:
This PR tries to resolve the "StringCaseLocaleUsage" warnings from the Error Prone.

#5711 has made the same changes on the `main` branch already.


jonatan-ivanov:
Thank you!

PR #6163, opened 2025-04-24
Remove redundant Checkstyle checks in favor of StringCaseLocaleUsage check from Error Prone

izeye:
This PR removes redundant Checkstyle checks in favor of the "StringCaseLocaleUsage" check from the Error Prone that provides the following warning:

```
/Users/izeye/IdeaProjects/micrometer/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java:75: warning: [StringCaseLocaleUsage] Specify a `Locale` when calling `String#to{Lower,Upper}Case`. (Note: there are multiple suggested fixes; the third may be most appropriate if you're dealing with ASCII Strings.)
                (timeUnit) -> timeUnit.toString().toLowerCase(),
                                                             ^
    (see https://errorprone.info/bugpattern/StringCaseLocaleUsage)
  Did you mean '(timeUnit) -> timeUnit.toString().toLowerCase(Locale.ROOT),' or '(timeUnit) -> timeUnit.toString().toLowerCase(Locale.getDefault()),' or '(timeUnit) -> Ascii.toLowerCase(timeUnit.toString()),'?
```

See gh-5711

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -46,22 +46,6 @@
             <property name="message" value="Please use AssertJ imports." />
             <property name="ignoreComments" value="true" />
         </module>
-        <module name="com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineJavaCheck">
-            <property name="id" value="toLowerCaseWithoutLocale"/>
-            <property name="format" value="\.toLowerCase\(\)"/>
-            <property name="maximum" value="0"/>
-            <property name="message"
-                      value="String.toLowerCase(Locale.ROOT) should be String.toLowerCase(Locale.ROOT)"/>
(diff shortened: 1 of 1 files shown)


jonatan-ivanov:
Thank you!

shakuzen:
Did Checkstyle cause the build to fail before? I see the Error Prone pattern is a compiler warning by default, which means it won't fail our build.

izeye:
@shakuzen Right. I thought the warning is enough for this.

shakuzen:
If we've fixed all the places where this would have failed, I'd lean toward making it a compiler error, because if the String comes from an untrusted source, it is potentially a security issue to not specify the Locale.

izeye:

I created https://github.com/micrometer-metrics/micrometer/pull/6175 to try to apply your suggestion.

Issue #6113, opened 2025-04-13
Tags-related performance improvements


etki:
**Please describe the feature request.**
Hi. As you surely know, all observability instruments have more demands for performance than anything else, as they lie on the critical path and are called more often than anything. Currently, Tags and its surroundings are major performance drainers due to many unoptimized places (i bet you've faced it, and it was the reason for `Tags#merge()`), and due to its API i can't solve many problems myself without overriding the class and its siblings in my own codebase. My suggestions are the following:

- Add `Tags#length()` and `Tags#isEmpty()` methods. Tags is already an Iterable and streamable type, there's not much sense in not doing it. With it, consumers would be able to size their ArrayList's and arrays properly, should they need to do some kind of processing.
- Add `Tags#getTag(int i)` method. This would allow zero-allocation iterations.
- Add `Tags#getTag(String key)` method and reuse it in `Meter#getTag(String key)`. It would be also beneficial to test if binary search would be more performant on common tag collection sizes.
- Add a `Tags#asList()` method, which will simply return `Collections.unmodifiableList(Arrays.asList(content))`. This will provide a common interface for consumers with almost minimal possible allocation.
- Add a `Tags#toArray()` method that would simply copy the array. For functionality such as replacing single tag this would result in much faster operations.
- Add a `Tags#copyTo(Tags[] target, int at, int skip, int amount)` for other operations that preserve full segments during their processing.
- Use it in `Meter.Id#getTags()`. This will significantly reduce allocation (currently a new ArrayList is allocated there, without a capacity hint, which also adds penalty) and avoid a potential call in currently existing forEach() call (if it won't get inlined).
- Expose `Tags Meter.Id#getRawTags()` or a similar method. There is not much sense in hiding it, as class is immutable and is already returned through getTagsAsIterable().
- Expose methods of kind `Tags.Unsafe#ofSorted(Tag... tags)` and `Tags.Unsafe#ofSorted(Tag[] tags, int length)` to avoid sorting altogether for the people who know what they are doing. I merge different sorted sources of tags in my case, by knowing that i can merge this sources in an already sorted way; for the tag value replacing filter it would also be a blessing to just copy an array and modify a single item there without any re-sorting.
- Add `Meter.Id#replaceTags(Tags)`, which would allow to skip excessive calls (at least .iterator().next() is currently on the code path, unnecessary allocation that could have been avoided).

[...]

PR #6174, opened 2025-04-25
Add benchmarks for basic prebuilt MeterFilter implementations

etki:
# What is this?

This PR contains only benchmarks for some basic API, MeterFilter implementations and a couple of Tags methods - no new symbols visible for users, no changes in production code. It goes a bit extra by exploiting different inputs so that different usage patterns would be evaluated and "better at this count, yet worse at that count" situations had less chance of getting unnoticed (assuming that whoever works on improving performance of specific code part runs these benchmarks prior to and after the changes).

# Is it really needed?

I'm prepping other PRs for some non-dramatic updates in filters and tags code (also no new symbols or new functionality, but changing the existing one preserving the present functionality), so these benchmarks just had to emerge. It's not up to me to decide if they are needed in the project, but they came up naturally and they shouldn't take any resources from anyone except for the time from people explicitly running them, so i don't see an obstacle here.

But at the same time i'm just having fun, so i don't have a problem with any requested changes or removals.

# There are already benchmarks for tags!

Yes, but: 1) i've noticed them just when i was preparing this PR, 2) they use static input (which also comes with a fixed size). I don't have any energy to check the assembly, but JIT theoretically might infer wrong assumptions from that; i also think that dynamic input size is a must here, as the whole sort-the-array story is non-linear from the very beginning. Not sure what do with these, but you can guide me.

# Aren't these benchmarks too sophisticated?

They are sophisticated, but not "too" for me; they aren't entirely necessary, but since they are coming for free and doing the more thorough matrix evaluation, i see this as a benefit. But i'm farming my XP here, so if there's a concern against checking these in, i'm OK with it.

# Any interesting results?

These are just the baseline measurements, i.e. they exist to compare two different revisions. There are some hints on the situations that can be improved. Selecting the biggest offender (using Intel N100 @ 2GHz) for a dramatic effect:

| name | ignored tags | tags in identifier | result | unit|
|:--|:--|:--|:--|:---|
| config.filter.MeterFilterIgnoreTagsBenchmark.baseline | 64 | 64 | 8439.085 ±  7.119 | ns/op |
| config.filter.MeterFilterIgnoreTagsBenchmark.baseline:instructions:u | 64 | 64 | 38896.080 | #/op |

Thing's happening here is just taking ~64 elements (there are many samples, _most_ of which have 64 tags) and filter out the ones that match a set of 64 keys. While here it comes at 8.4usec, this could be just an O(n) task executing in a matter of a couple of hundreds of nanoseconds. My forthcoming PRs would somewhat address things like that (being honest: this is the biggest offender, other places are not so impressive), but getting to the best position would require #6113.

### DIFF
--- a/benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/BenchmarkSupport.java
+++ b/benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/BenchmarkSupport.java
@@ -15,7 +15,7 @@
  */
 package io.micrometer.benchmark;
 
-import io.micrometer.common.lang.Nullable;
+import org.jspecify.annotations.Nullable;
 import org.openjdk.jmh.annotations.Mode;
 import org.openjdk.jmh.profile.GCProfiler;
 import org.openjdk.jmh.profile.LinuxPerfAsmProfiler;
@@ -204,7 +204,7 @@ public class BenchmarkSupport {
             .warmupTime(new TimeValue(5L, TimeUnit.SECONDS))
             // 9.5m benchmarks
             .measurementIterations(57)
-            .measurementTime(new TimeValue(5L, TimeUnit.SECONDS))
+            .measurementTime(new TimeValue(10L, TimeUnit.SECONDS))
             .mode(Mode.AverageTime)
             .timeUnit(TimeUnit.NANOSECONDS)
             .shouldDoGC(true)
--- a/benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/BenchmarkSupport.java
+++ b/benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/BenchmarkSupport.java
@@ -1,0 +1,298 @@
+/*
+ * Copyright 2025 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.benchmark;
+
+import io.micrometer.common.lang.Nullable;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.profile.GCProfiler;
+import org.openjdk.jmh.profile.LinuxPerfAsmProfiler;
+import org.openjdk.jmh.profile.LinuxPerfNormProfiler;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.ChainedOptionsBuilder;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.TimeValue;
+
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.Locale;
+import java.util.Random;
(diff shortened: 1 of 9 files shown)


benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/BenchmarkSupport.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    public static ChainedOptionsBuilder defaults() {
+        ChainedOptionsBuilder options = new OptionsBuilder().forks(1)
+            // 0.5m warmups
+            .warmupIterations(6)
+            .warmupTime(new TimeValue(5L, TimeUnit.SECONDS))
+            // 9.5m benchmarks
+            .measurementIterations(57)
+            .measurementTime(new TimeValue(5L, TimeUnit.SECONDS))
```
shakuzen: I'm curious if there was any particular motivation to use 5 seconds instead of the JMH default of 10 seconds per iteration. Also, the comment says 9.5 minutes of measurement, but 9.5 minutes would be 57 iterations * 10 seconds.

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/instrument/config/filter/MeterFilterCommonTagsBenchmark.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private static final int MASK = BenchmarkSupport.DEFAULT_MASK;
+
+    private static final int SAMPLE_STEP = BenchmarkSupport.SAMPLE_STEP;
+
+    @Param({ "0", "1", "2", "4", "8", "16", "32", "64" })
+    public int supplied;
+
+    @Param({ "0", "1", "2", "4", "8", "16", "32", "64" })
+    public int ignored;
```
shakuzen: Probably should be something like `commonTagCount`. I'd guess this was copied from the ignore tags benchmark.

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/core/instrument/config/filter/MeterFilterIgnoreTagsBenchmark.java
```diff
@@ -0,0 +1,91 @@
...
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.config.MeterFilter;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.runner.RunnerException;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Evaluates performance of {@link MeterFilter#commonTags(Iterable)}. To simulate at least
```
shakuzen:
```suggestion
 * Evaluates performance of {@link MeterFilter#ignoreTags(Iterable)}. To simulate at least
```

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/BenchmarkSupport.java
```diff
@@ -0,0 +1,298 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.benchmark;
+
+import io.micrometer.common.lang.Nullable;
```
shakuzen: Please rebase on the latest `main` and update any nullability annotations to use the jSpecify annotations.

shakuzen (commented):
Thank you for all the work on this. It's far more sophisticated than our existing benchmarks. It generally looks good to me, but I've left some review comments. Could you rebase on `main` when you have a chance and look at the review comments? Then I think we're ready to merge, if @jonatan-ivanov doesn't have any objections.

shakuzen:
@etki following up to see if you've had a chance to check my review.

shakuzen:
I've addressed the review comments and rebased on the latest `main`.

PR #6180, opened 2025-04-26
Resolve LongDoubleConversion from Error Prone

izeye:
This PR resolves LongDoubleConversion from Error Prone.

This PR also changes to handle LongDoubleConversion as errors.


jonatan-ivanov:
Thanks for the PR!
I'm halfway in reviewing this, there are a few things I want to verify because they might be trickier that they seem first.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpStepTimer.java
```diff
@@ -53,10 +53,10 @@ class OtlpStepTimer extends AbstractTimer {
 
     @Override
     protected void recordNonNegative(final long amount, final TimeUnit unit) {
-        final long nanoAmount = (long) TimeUtils.convert(amount, unit, TimeUnit.NANOSECONDS);
+        final long nanoAmount = (long) TimeUtils.convert((double) amount, unit, TimeUnit.NANOSECONDS);
         count.add(1L);
         total.add(nanoAmount);
-        max.record(nanoAmount);
+        max.record((double) nanoAmount);
```
jonatan-ivanov, line 59: I don't really like the chain of conversions here (`long -> double -> integer conversion -> long -> double`) but this PR just makes it explicit so would not change t here. I guess first we need to add `long` overrides to `TimeUtils` to fix this.

@shakuzen What do you think?

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you!

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpStepTimer.java
shakuzen: That's probably a good idea.

jonatan-ivanov: 👍🏼 https://github.com/micrometer-metrics/micrometer/issues/6215

PR #6187, opened 2025-04-27
TagRenamingFilter

etki:
This is the last of performance-related PRs. As with previous one, this description will be updated once i have more time to run all the benchmarks. Nothing super super cool, done mostly to keep consistency, but these are the latest results (requires additional verification). Please keep in mind that X is not X supplied tags in _every_ input, but only in 90% of them (other 10% contain 1-64 tags, done to simulate a somewhat real situation).

| processed tags | version | nanoseconds / operation | instructions |
|--:|:--|--:|--:|
| 0 | main | ` 91.356 ±  0.353` | 374.790 |
| 0 | pr | ` 81.980 ±  0.859` | 335.794 |
| 1 | main | `126.592 ±  0.345` | 473.020 |
| 1 | pr | ` 92.874 ±  0.277` | 350.152 |
| 4 | main | `146.673 ±  0.369` | 568.136 |
| 4 | pr | `106.541 ±  0.284` | 409.278 |
| 8 | main | `185.319 ±  0.457` | 725.593 |
| 8 | pr | `127.259 ±  0.235` | 500.279 |
| 16 | main | `274.408 ±  0.560` | 1162.592 |
| 16 | pr | `191.191 ±  0.331` | 800.652 |
| 32 | main | `503.440 ±  0.745` | 2126.366  |
| 32 | pr | `395.133 ±  0.781` | 1817.026 |
| 64 | main |  `821.375 ±  1.231` | 4549.852 |
| 64 | pr | `865.062 ±  1.071` | 5404.027 |

The results are getting worse in 32+ range (compared to the single-iteration version as well), because the chance of actually seeing a replacement gets to 50% and higher. This, i believe, is where we are actually leaving the real situation. The double-iteration version is in fact the last minute experiment, there might be a bit of an improvement, will hopefully have some time to investigate. As in all other PRs, #6113 would greatly help to make another CPU relief step.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
@@ -17,10 +17,10 @@ package io.micrometer.core.instrument.config;
 
 import io.micrometer.common.lang.Nullable;
 import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.config.filter.TagRenamingFilter;
 import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
 
 import java.time.Duration;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


jonatan-ivanov:
Hi,

Instead of replying to all of your PRs separately, let me reply here to all of them first.
First of all: thank you very much!

I looked through them quickly, here are a few updates/things to consider:
- We need to discuss them internally (they are somewhat connected)
- Timeline-wise we can consider them for 1.16.0 (which means the earliest we can merge them is mid next month)
- I know @shakuzen wanted to create JMH benchmarks that are connected to these areas, I think we should do that first
- Last week I worked on making it easier to run perf tests in a more stable infrastructure than our laptops, I need to discuss this with @shakuzen, it can help merging your PRs if we can use it to run stable JMH benchmarks

We will get back to you soon with more update after we discussed the above/reviewed your PRs.

PR #6194, opened 2025-04-30
TimedHandler shutdown hanging indefinitely

RafeArnold:
the future returned by `io.micrometer.jetty12.server.TimedHandler.shutdown()` would never complete if `shutdown()` is called while requests are still active, even after all active requests are completed.

this change fixes the issue by calling `shutdown.check()` after completing each request.

an `if (shutdown.isShutdown())` could be introduced before `shutdown.check()` for clarity, but since the current implementation of `check()` effectively does this already, i felt it wasn't worth it.

### DIFF
--- a/micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
+++ b/micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
@@ -25,7 +25,6 @@ import org.eclipse.jetty.http.HttpTester;
 import org.eclipse.jetty.server.*;
 import org.eclipse.jetty.util.BufferUtil;
 import org.eclipse.jetty.util.Callback;
-import org.eclipse.jetty.util.component.Graceful;
 import org.eclipse.jetty.util.component.LifeCycle;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
@@ -166,13 +165,11 @@ class TimedHandlerTest {
 
             // initiate a shutdown
             Future<Void> shutdownFuture = timedHandler.shutdown();
-            Graceful.Shutdown shutdown = timedHandler.getShutdown();
             assertThat(shutdownFuture.isDone()).isFalse();
 
             // delay half what the handler is sleeping
             Thread.sleep(delay / 2);
             // response is still active, so don't shutdown.
-            shutdown.check();
             assertThat(shutdownFuture.isDone()).isFalse();
 
             // Read response to ensure it is done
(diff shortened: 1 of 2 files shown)


micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
```diff
@@ -166,13 +166,11 @@ public boolean handle(Request request, Response response, Callback callback) {
 
             // initiate a shutdown
             Future<Void> shutdownFuture = timedHandler.shutdown();
-            Graceful.Shutdown shutdown = timedHandler.getShutdown();
             assertThat(shutdownFuture.isDone()).isFalse();
 
             // delay half what the handler is sleeping
             Thread.sleep(delay / 2);
             // response is still active, so don't shutdown.
-            shutdown.check();
```
RafeArnold, line 175: i believe the test should keep this check, as this is testing that if the check is run at this point, it will not change its state to "done".

RafeArnold: the comment above the line is implying this assertion.

shakuzen: Thanks for the fast review on the changes I made. I see the intention of the test before, but it also seems odd. TimedHandler should be controlling when `shutdown.check()` is called, and I don't think it should be called externally, although it shouldn't hurt any behavior to call it externally. If the purpose of this part of the test is for the scenario when another request finishes while one is still in progress, I think it would be better to make a test that does that rather than externally call `shutdown.check()`. Thoughts?

shakuzen: I'm merging before we resolve this feedback because I want to get the fix included in our round of releases today. We can continue the discussion and follow-up later.

RafeArnold: thanks for merging the pr. it's much appreciated.

i think a test to assert that another request finishing doesn't mark the shutdown as complete would be a good test, like you suggest.

i agree that `shutdown.check()` being called externally is strange, but given that the handler's `shutdown` field is exposed externally via `TimedHandler.getShutdown()`, it is possible that library consumers will call `shutdown.check()` out of band (in fact, that's exactly how we fixed this issue before this pr was merged), so a test for this scenario would make sense to me. either that or encapsulating the `shutdown` field completely, which would be a breaking change. you may feel this is not worth the effort, which i understand.

shakuzen: I've opened https://github.com/micrometer-metrics/micrometer/issues/6227 for the work of adding more tests around this.

PR #6202, opened 2025-05-05
Disable StringConcatToTextBlock warnings from Error Prone

izeye:
This PR changes to disable `StringConcatToTextBlock` warnings from Error Prone as there are 43 occurences that are not applicable.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,6 +78,9 @@ subprojects {
                     "MissingOverride",
                     "StringCaseLocaleUsage"
             )
+            options.errorprone.disable(
+                    "StringConcatToTextBlock"
+            )
 
             if (!javaLanguageVersion.canCompileOrRun(17)) {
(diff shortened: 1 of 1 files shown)


jonatan-ivanov:
Thanks!
I think we cannot fix [StringConcatToTextBlock](https://errorprone.info/bugpattern/StringConcatToTextBlock) without raising the Java baseline.

shakuzen:
Perhaps it is worth a comment where the check is disabled to say if we ever raise the baseline to 17 or higher, we should re-enable the check.

izeye:
@shakuzen I added a similar comment in https://github.com/micrometer-metrics/micrometer/pull/6207/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R82.

PR #6206, opened 2025-05-06
Custom refresh interval for Kafka metrics

spryslmatej:
- Makes refresh interval of Kafka metrics customizable.
- If no interval is supplied, the binders will use previously hard-coded 60 seconds.


jonatan-ivanov:
Could you please tell us your use-case first?
Why do you need to modify it?

jonatan-ivanov:
I understand what you are doing I don't get though what is the connection with the scheduling frequency. I guess this part:
>60 seconds of delay is not suitable for me anymore.

Why it is not suitable anymore?
What value is better for your use-case and why?
Which registry you use and if it is a step registry what is your `step`?

(I think I'm ok doing this but I would like to understand the reason behind it.)

spryslmatej:
> Why it is not suitable anymore? What value is better for your use-case and why?

It simply takes too long for me to see that everything works as expected. (mainly in graphs, but sometimes in the `/metrics` endpoint output)

I would probably set it to 15 seconds for my use-case.

> Which registry you use and if it is a step registry what is your step?

I use the `PrometheusMeterRegistry`.

Now that I think about it, there are two solutions to my problem. 
1. Allow custom refresh intervals (this PR).
2. Allow on-demand refresh invocation. That could work because the app receives an HTTP request containing the topic routing setting and sets it up in Kafka Streams. Once I'd know that a setting has been updated, I could invoke the refresh method. That would make the original periodical `scheduler.scheduleAtFixedRate` refresh superfluous for me.

jonatan-ivanov:
Thank you!

Just to double check, the values of existing metrics are refreshed in your use-case whenever you scrape (since that's on-demand), right? (Micrometer asks the underlying Kafka objects for the latest values every time you scrape.)
What causes issues for you is new metrics are not being registered fast enough.

Let us discuss this internally, I think we should be able to make this configurable.

jonatan-ivanov:
Before I merge this in, could you please verify two things?
- Refreshing can be costly. Could you please verify that more frequently doing this does not cause any issues to you?
- Metrics can appear and disappear without being scraped if the interval is not right. Could you please check that the metrics refresh and the scraping interval are right?

spryslmatej:
Sorry for the late reply, had a busy schedule :)

>    Refreshing can be costly. Could you please verify that more frequently doing this does not cause any issues to you?

Yes, I had no issues with more frequent refreshing, even when reaching high throughputs.

 >   Metrics can appear and disappear without being scraped if the interval is not right. Could you please check that the metrics refresh and the scraping interval are right?

You are correct, if the scrape interval happens to be larger than two refresh intervals, it might appear as if there were no metrics added/removed.
I use Prometheus for scraping with an interval of 15 seconds, so this is not an issue for me.
I'm quite confident users tweaking this interval will be aware about this effect. We could add it to the javadocs just to be sure.

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/6319

Issue #6215, opened 2025-05-07
Add long overloads for methods of TimeUtils


jonatan-ivanov:
Methods that convert time units in `TimeUtils` accept and return `double`.
This can lead to unnecessary conversions so we might want to add methods that accept and return `long` to avoid this. See: https://github.com/micrometer-metrics/micrometer/pull/6180#discussion_r2070778831

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/6224

PR #6224, opened 2025-05-11
Replace TimeUtils usage to TimeUnit where applicable

ngocnhan-tran1996:
Closes gh-6215

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpStepTimer.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpStepTimer.java
@@ -53,7 +53,7 @@ class OtlpStepTimer extends AbstractTimer {
 
     @Override
     protected void recordNonNegative(final long amount, final TimeUnit unit) {
-        final long nanoAmount = (long) TimeUtils.convert((double) amount, unit, TimeUnit.NANOSECONDS);
+        final long nanoAmount = TimeUtils.convert(amount, unit, TimeUnit.NANOSECONDS);
         count.add(1L);
         total.add(nanoAmount);
         max.record((double) nanoAmount);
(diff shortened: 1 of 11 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpStepTimer.java
```diff
@@ -53,7 +53,7 @@ class OtlpStepTimer extends AbstractTimer implements OtlpHistogramSupport {
 
     @Override
     protected void recordNonNegative(final long amount, final TimeUnit unit) {
-        final long nanoAmount = (long) TimeUtils.convert((double) amount, unit, TimeUnit.NANOSECONDS);
+        final long nanoAmount = TimeUtils.convert(amount, unit, TimeUnit.NANOSECONDS);
```
jonatan-ivanov, line 56: Looking through the whole PR, now I remember why `TimeUtils` does not have overloads with `long`: instead of using `TimeUtils`, this PR can be done by simply using `TimeUnit`. For example, instead of:
```java
final long nanoAmount = TimeUtils.convert(amount, unit, TimeUnit.NANOSECONDS);
```
we can do
```java
final long nanoAmount = unit.toNanos(amount);
```
Or instead of
```java
private long timeNanos = TimeUtils.millisToUnit(1, TimeUnit.NANOSECONDS);
```
we can do
```java
private long timeNanos = TimeUnit.MILLISECONDS.toNanos(1);
```

In both cases, using purely `TimeUnit` seems more intuitive to me: it's more apparent/obvious what is the source and destination unit, hence more readable.
What do you think?

ngocnhan-tran1996: When I saw the issue, I've just thought we'd like to manage convert long type method in `TimeUtils`. And you're right, we should use directly `TimeUnit` for long type. Let me change to use `TimeUnit`

shakuzen (approved):

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you!

PR #6262, opened 2025-05-16
Do not register error on gRPC client cancel

### LINKED ISSUE #6261: gRPC client interceptor incorrectly registers status CANCELLED as error
**Describe the bug**
[ObservationGrpcClientCallListener registers all closures with a cause as an error](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcClientCallListener.java#L54). This is wrong because CANCELLED is not an error (and in theory even status OK could have a cause, _but no one should do that_).

_In general_ when one cancels an RPC you don't supply a cause which means it will not be registered as error, but gRPC will set a cause when it automaticall cancels outgoing RPCs due to cancellation propogation from an incoming RPC. Then you get ⬇ which has a cause and thus gets registered as error:
`Status{code=CANCELLED, description=RPC cancelled, cause=io.grpc.InternalStatusRuntimeException: CANCELLED: RPC cancelled`

For reference, [ObservationGrpcServerCallListener never treats cancellations as errors](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcServerCallListener.java#L51).

cfredri4:
Currently a gRPC client cancellation will be registered as an error but a cancellation is not an error.
This change makes it so that OK and CANCELLED statuses are not registered as errors.
Closes #6261

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcClientCallListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/grpc/ObservationGrpcClientCallListener.java
@@ -51,7 +51,8 @@ class ObservationGrpcClientCallListener<RespT> extends SimpleForwardingClientCal
         GrpcClientObservationContext context = (GrpcClientObservationContext) this.observation.getContext();
         context.setStatusCode(status.getCode());
         context.setTrailers(trailers);
-        if (status.getCause() != null) {
+        if (!Status.Code.OK.equals(status.getCode()) && !Status.Code.CANCELLED.equals(status.getCode())
+                && status.getCause() != null) {
             observation.error(status.getCause());
         }
         this.observation.stop();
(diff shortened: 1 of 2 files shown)


shakuzen (commented):
Thanks for the pull request. This was probably just missed. Could you add some test cases for the cases this covers so we make sure this stays fixed through any future changes? Also, since this sounds like a bug, we should fix it for `1.13.x` and we can merge it forward.

cfredri4:
Sure. But can I get some pointers on a suitable test class to change or copy from? I'm not very familiar with the code base.

shakuzen:
It looks like existing tests are in `GrpcObservationTest` and `GrpcAsyncTest`. It would probably make sense to add tests to those.

cfredri4:
@shakuzen Added test now. I'm in deep waters here so please check thoroughly. 😅

shakuzen:
Thanks for adding the test @cfredri4. I've asked @ttddyy who wrote the instrumentation to review, if he has spare time to do so.

shakuzen:
I've rebased onto `1.13.x` so we can include this in maintenance releases.

PR #6270, opened 2025-05-20
Migration to jSpecify (part 2)

shakuzen:
Migrates more modules (but not all) to jSpecify.

I believe (though I can't be confident) this is everything except for the following modules that remain to be migrated:
- micrometer-core
- micrometer-observation
- micrometer-observation-test
- micrometer-registry-prometheus
- micrometer-registry-prometheus-simpleclient

See #5547


gradle/libs.versions.toml
```diff
@@ -71,7 +71,7 @@ slf4j = "1.7.36"
 spectator-atlas = "1.8.13"
 spring5 = "5.3.39"
 spring6 = "6.2.7"
-spring-javaformat = "0.0.43"
+spring-javaformat = "0.0.45"
```
shakuzen, line 74: Upgrades to the latest formatter with fixes for the jSpecify annotations.

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/DatadogStatsdLineBuilder.java
```diff
@@ -38,16 +38,15 @@ public class DatadogStatsdLineBuilder extends FlavorStatsdLineBuilder {
 
     private final Object conventionTagsLock = new Object();
 
-    @SuppressWarnings({ "NullableProblems", "unused" })
+    @SuppressWarnings({ "NullAway.Init", "unused" })
```
jonatan-ivanov, line 41: Can we create an issue to fix these?
I think:
- `NullAway.Init` might not be needed on `namingConvention`
- `name` and `tagsNoStat` can be initialized with `""` and can be `@NonNull`

Same for the other *LineBuilders.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryTest.java
```diff
@@ -57,6 +57,7 @@ class StatsdMeterRegistryTest {
 
     private MockClock clock = new MockClock();
 
+    @SuppressWarnings("NullAway.Init")
```
jonatan-ivanov, line 60: Can we create an issue to fix this or fix it here?
Since every test case creates it's own registry anyways, we can remove this field with `cleanUp` and replace `registry =` to `StatsdMeterRegistry registry =`.

I guess the same for `StatsdMeterRegistryPublishTest` (`meterRegistry`, `server`, `serverLatch`).

gradle/libs.versions.toml
jonatan-ivanov: fyi: https://github.com/micrometer-metrics/micrometer/pull/6273

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/compare/CompareOTLPHistograms.java
```diff
@@ -29,8 +29,6 @@
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 //CHECKSTYLE:OFF
 import com.google.common.collect.Iterators;
-////CHECKSTYLE:ON
```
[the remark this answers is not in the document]
shakuzen: I'm not sure what made this change. It wasn't intentional. Although four slashes look like it was incorrect anyway. I'll restore it with two slashes.

implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/AtlasUtils.java
```diff
@@ -15,15 +15,14 @@
...
 import io.micrometer.core.instrument.Statistic;
+import org.jspecify.annotations.Nullable;
 
 import static com.netflix.spectator.api.Statistic.*;
 
 public class AtlasUtils {
 
-    @Nullable
-    static com.netflix.spectator.api.Statistic toSpectatorStatistic(Statistic stat) {
+    static com.netflix.spectator.api.@Nullable Statistic toSpectatorStatistic(Statistic stat) {
```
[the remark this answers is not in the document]
shakuzen: No, unfortunately not - it doesn't compile. This is how type-use annotations work, both with inner classes referenced through their parent (`Meter.Id`) and fully qualified types like this.

implementations/micrometer-registry-cloudwatch2/src/test/java/io/micrometer/cloudwatch2/CloudWatchConfigTest.java
```diff
@@ -78,12 +79,13 @@ public String get(String key) {
     void invalidOverrideNamespace() {
         CloudWatchConfig config = new CloudWatchConfig() {
             @Override
+            @SuppressWarnings("NullAway")
```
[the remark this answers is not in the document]
shakuzen: The test is testing a CloudWatchConfig instance with an invalid `namespace()`, so we need to suppress NullAway here to test what we want. I suppose an alternative would be to make it something invalid other than `null`, but then we're testing something different than we have been.

micrometer-java11/src/main/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClient.java
```diff
@@ -259,7 +255,7 @@ public <T> CompletableFuture<HttpResponse<T>> sendAsync(HttpRequest httpRequest,
 
     @Override
     public <T> CompletableFuture<HttpResponse<T>> sendAsync(HttpRequest httpRequest,
-            HttpResponse.BodyHandler<T> bodyHandler, @Nullable HttpResponse.PushPromiseHandler<T> pushPromiseHandler) {
+            HttpResponse.BodyHandler<T> bodyHandler, HttpResponse.@Nullable PushPromiseHandler<T> pushPromiseHandler) {
```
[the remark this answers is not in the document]
shakuzen: Yes. We could either suppress the ErrorProne rule about inner class references or configure it back to a warning and ignore it here. We can't do anything about where Java requires type-use annotations to be declared if we have any nesting in references, though.

micrometer-java11/src/test/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClientTests.java
```diff
@@ -203,6 +203,7 @@ public void onStart(HttpClientContext context) {
 
     static class StoreContextObservationHandler implements ObservationHandler<HttpClientContext> {
 
+        @SuppressWarnings("NullAway.Init")
```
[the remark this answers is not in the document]
shakuzen: I don't think we need to be so careful in tests. If the test throws a NPE, it will fail and we will investigate. The suppression is saying we know the instance variable won't be initialized at construction, but we expect based on the order in which methods will be called, it will be initialized before it is used. If that assumption doesn't hold, a test will fail with a NPE, which seems as good as throwing an IllegalStateException instead.

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/DatadogStatsdLineBuilder.java
shakuzen: Perhaps. I didn't want to mess with the logic to get the migration done when we have no reported issues for these. It's probably best to try to improve that in a separate change.

implementations/micrometer-registry-dynatrace/src/test/java/io/micrometer/dynatrace/DynatraceMeterRegistryTest.java
```diff
@@ -67,7 +69,7 @@ void shouldSendProperRequest() throws Throwable {
         when(httpClient.send(isA(HttpSender.Request.class)))
             .thenReturn(new HttpSender.Response(202, "{ \"linesOk\": 3, \"linesInvalid\": 0, \"error\": null }"));
 
-        Double gauge = meterRegistry.gauge("my.gauge", 42d);
+        Double gauge = Objects.requireNonNull(meterRegistry.gauge("my.gauge", 42d));
```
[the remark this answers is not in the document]
shakuzen: In some sense it can't; in some sense it can. This is an issue of inconsistent nullability in micrometer-core API (which I did not have the sanity to tackle in this PR). Some `gauge` methods allow a nullable state object, some don't, but all have a nullable return value.

implementations/micrometer-registry-new-relic/src/test/java/io/micrometer/newrelic/NewRelicMeterRegistryTest.java
```diff
@@ -1001,6 +1003,7 @@ void builderBuildWhenBothHttpClientAndClientProviderAreSetShouldThrowIllegalStat
 
     static class MockHttpSender implements HttpSender {
 
+        @SuppressWarnings("NullAway.Init")
```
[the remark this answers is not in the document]
shakuzen: For tests, I don't think it's worth it.

implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsConfig.java
```diff
@@ -54,8 +54,7 @@ default String apiToken() {
      * @return The tag that will be mapped to {@literal @host} when shipping metrics to
      * AppOptics.
      */
-    @Nullable
-    default String hostTag() {
+    default @Nullable String hostTag() {
```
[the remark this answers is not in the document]
shakuzen: I'm not sure I understand the question. It's the same semantics as before (the return value is nullable), but by convention, type-use annotations are next to the type they're annotating (the return type of this method) rather than annotating the method like the JSR305 annotations did (because they aren't type-use annotations).

implementations/micrometer-registry-influx/src/test/java/io/micrometer/influx/InfluxNamingConventionTest.java
```diff
@@ -86,7 +87,7 @@ void newlineCharReplacedInTagValues() {
     private static class CustomNamingConvention implements NamingConvention {
 
         @Override
-        public String name(String name, Meter.Type type, String baseUnit) {
+        public String name(String name, Meter.Type type, @Nullable String baseUnit) {
```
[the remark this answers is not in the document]
shakuzen: We could, but it's already a gigantic PR. I'd vote for doing it in a separate PR when we do micrometer-core.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryTest.java
shakuzen: I'd vote for a separate issue so it isn't buried in all these changes.

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
```diff
@@ -56,13 +56,11 @@ default String uri() {
      * datasource defined by these tags.
      * @return Tags which uniquely determine the datasource to store metrics in.
      */
-    @Nullable
-    default Map<String, String> tags() {
+    @Nullable default Map<String, String> tags() {
```
shakuzen, line 59: I wonder why this didn't end up after `default` but it did other places.

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioMeterRegistry.java
```diff
@@ -128,8 +128,7 @@ protected void publish() {
     }
 
     @Override
-    @NonNull
-    protected TimeUnit getBaseTimeUnit() {
+    protected @NonNull TimeUnit getBaseTimeUnit() {
```
shakuzen, line 131: `@NonNull` is redundant here since the package-info is annotated with `@NullMarked`. I'll remove it.

micrometer-commons/src/main/java/io/micrometer/common/annotation/NoOpValueResolver.java
```diff
@@ -25,7 +25,7 @@ public class NoOpValueResolver implements ValueResolver {
 
     @Override
     public String resolve(Object parameter) {
-        return null;
+        return "";
```
shakuzen, line 28: This is a significant change we should probably give more consideration. @jonatan-ivanov

micrometer-jetty12/src/main/java/io/micrometer/jetty12/client/package-info.java
```diff
@@ -17,7 +17,4 @@
 /**
  * Instrumentation for Jetty 12 client.
  */
-@NonNullApi
 package io.micrometer.jetty12.client;
```
shakuzen, line 20: I'm not sure why this wasn't changed to `@NullMarked`. I'll fix it.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -1029,8 +1028,7 @@ public <T> Context put(Object key, T object) {
          * @return entry or {@code null} if not present
          */
         @Override
-        @Nullable
-        public <T> T get(Object key) {
+        @Nullable public <T> T get(Object key) {
```
shakuzen, line 1031: The annotation is in the wrong place (although effectively it will work anyway). This is because I ignored the micrometer-observation module when running the recipe to move the annotations, but the formatter ran here anyway. I think we can leave it like this for now and fix it in a separate change when we do the micrometer-observation module.

micrometer-test/src/main/java/io/micrometer/common/util/internal/logging/LogEvent.java
```diff
@@ -15,6 +15,8 @@
  */
 package io.micrometer.common.util.internal.logging;
 
+import io.micrometer.common.lang.Nullable;
```
shakuzen, line 18: Wrong import

implementations/micrometer-registry-appoptics/src/main/java/io/micrometer/appoptics/AppOpticsConfig.java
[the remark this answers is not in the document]
jonatan-ivanov: Sorry, I meant this should not be null because of `.orElse("instance")` but now I guess `@Nullable` is there since it can be overridden.

micrometer-commons/src/main/java/io/micrometer/common/annotation/NoOpValueResolver.java
jonatan-ivanov: I agree. Since `ValueResolver` has a pair: `ValueExpressionResolver` I think we should keep the two in-sync but since they are different now, we should change one of them:

1. Mark `ValueResolver#resolve` `@Nullable`, like `ValueExpressionResolver#resolve` is also `@Nullable`. We already do a null-check here: https://github.com/micrometer-metrics/micrometer/blob/6c7cf24a1415f11dc3ec904fb916d60a316c9f31/micrometer-core/src/main/java/io/micrometer/core/aop/MeterTagSupport.java#L52 but we need to verify this across the codebase.
2. Keep `ValueResolver#resolve` `@NonNull` but maybe we should also mark `ValueExpressionResolver#resolve` `@NonNull` (right now it's @Nullable).

If we decide to keep them out of sync as they are now returning `""` instead of `null` should not cause an issue in Micrometer's codebase since we are already mapping `null` to `""` (see above) and `NoOpValueResolver` is more like a "marker", it should not be called but we need to verify this across the codebase.

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
jonatan-ivanov: I think the formatter does this and the recipe you used uses the other order?

micrometer-commons/src/main/java/io/micrometer/common/annotation/NoOpValueResolver.java
jonatan-ivanov: I think if we decide to change this, we should take care about this in a separate PR.

implementations/micrometer-registry-humio/src/main/java/io/micrometer/humio/HumioConfig.java
shakuzen: But in this same file, the other change place (`apiToken`) ended up the other way

micrometer-commons/src/main/java/io/micrometer/common/annotation/NoOpValueResolver.java
shakuzen: I opened https://github.com/micrometer-metrics/micrometer/issues/6280

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryTest.java
jonatan-ivanov: Done in https://github.com/micrometer-metrics/micrometer/issues/6284

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/DatadogStatsdLineBuilder.java
jonatan-ivanov: I created a PR: https://github.com/micrometer-metrics/micrometer/pull/6285

PR #6271, opened 2025-05-20
Resolve StringSplitter from Error Prone

izeye:
This PR resolves [StringSplitter](https://errorprone.info/bugpattern/StringSplitter) from Error Prone.

This PR also changes to handle StringSplitter as errors.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
@@ -73,7 +73,7 @@ public interface NamingConvention {
         }
 
         private String toCamelCase(String value) {
-            String[] parts = value.split("\\.", -1);
+            String[] parts = value.split("\\.", 0);
             StringBuilder conventionName = new StringBuilder(value.length());
             for (int i = 0; i < parts.length; i++) {
                 String str = parts[i];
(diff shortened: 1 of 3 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
```diff
@@ -73,7 +73,7 @@ public String tagKey(String key) {
         }
 
         private String toCamelCase(String value) {
-            String[] parts = value.split("\\.");
+            String[] parts = value.split("\\.", -1);
```
jonatan-ivanov, line 76: I think there is a behavioral difference here.
`.split("\\.")` is the same as `.split("\\.", 0)` and
>If the limit is zero then the pattern will be applied as many times as possible, the array can have any length, and trailing empty strings will be discarded.
If the limit is negative then the pattern will be applied as many times as possible and the array can have any length.

Though, it seems empty parts will be dropped later on anyways.

jonatan-ivanov (approved):

micrometer-core/src/main/java/io/micrometer/core/instrument/config/NamingConvention.java
jonatan-ivanov: @izeye What do you think?
I think it would be better to keep the behavior (limit: 0).

izeye: @jonatan-ivanov Thanks for the feedback!

I chose "-1" as it seemed more consistent, and as you mentioned, its backward compatibility seemed to be guaranteed by later empty parts filtering.

If you think keeping its own original behavior is more important, I'm okay with it. I restored its original behavior in 
2554100.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you!

PR #6319, opened 2025-05-28
Make KafkaMetrics refresh interval configurable

jonatan-ivanov:
See https://github.com/micrometer-metrics/micrometer/pull/6206


shakuzen (approved):

shakuzen (commented):
I wonder if we couldn't use the MeterRegistry's step by default if a refresh interval isn't provided. We could only do that for PushMeterRegistry with PushRegistryConfig, but it might alleviate some need to pass a custom refresh interval.

jonatan-ivanov:
I created https://github.com/micrometer-metrics/micrometer/issues/6342, in the meantime, we can go with the refresh interval parameter in this PR.

jonatan-ivanov:
@spryslmatej A new SNAPSHOT release should be created in the next few minutes, could you please check if it works for you?

spryslmatej:
@jonatan-ivanov Tested with version `1.16.0-SNAPSHOT`, works as expected :+1: 

Thank you very much!

Issue #3563, opened 2022-12-09
Micrometer client not reconnecting to statsd after server is restarted


dunnk2022:
**Describe the bug**
The micrometer client does not reconnect to the statsd server (Datadog agent) when the agent is restarted.

**Environment**

 - Micrometer version - 1.9.5
 - Micrometer registry - statsd, Datadog flavor 
 - OS: Linux
 - Java version: 11.0.14

**To Reproduce**
How to reproduce the bug:
Restart the statsd server (datadog-agent)

**Expected behavior**

The micrometer client re-establishes a connection to the statsd server on port 8125 when the daemon is back up.

**Additional context**
Add any other context about the problem here, e.g. related issues.

dunnk2022:
Thanks for the response @shakuzen

dunnk2022:
We really just do a rolling restart of the agent and the issue occurs. Perhaps turning DEBUG logging on for the client will help to gather more info?

shakuzen:
If you turn on debug logging for classes under the `io.micrometer.shaded.reactor.netty` package, we can probably get some more information about what's happening. Or turn on debug logging for everything under `io.micrometer`.

shakuzen:
@raymondchen625 thanks for the additional datapoint. Unfortunately, since I haven't been able to reproduce it locally or get logs from anyone with more info about what is happening, we haven't made any progress on this. If you would be able to provide steps to reproduce this (ideally locally or with as minimal other things involved as possible) or some debug logs, it may help get to the bottom of this.

raymondchen625:
> @raymondchen625 thanks for the additional datapoint. Unfortunately, since I haven't been able to reproduce it locally or get logs from anyone with more info about what is happening, we haven't made any progress on this. If you would be able to provide steps to reproduce this (ideally locally or with as minimal other things involved as possible) or some debug logs, it may help get to the bottom of this.

I could not reproduce this locally. Unfortunately, I didn't try to capture IP packets before restarting the pods to see if Micrometer was sending traffic somewhere. I did check the application logs and didn't see any error. For custom metrics being sent to UDP 8125, I believe there is no error logs if the agent is down. But we also had APM metrics sent to TCP 8126, which should print error logs if the service is down. That makes me think Micrometer might have stopped sending out metrics. I might have captured nothing even if I had done the tcpdump on it.

shakuzen:
[...]

Issue #1252, opened 2019-02-27
StatsdMeterRegistry doesn't renew resolved ip address on DNS change


denzelby:
Hello everyone.

I've faced with the problem that StatsdMeterRegistry doesn't aware of DNS changes after it was `start()`ed.
As a result, metrics aren't delivered to Telegraf and there is no warning or so in logs.

How to reproduce:
1) Create record in /etc/hosts file, e.g.: `127.0.0.1         telegraf.kube-system.svc.cluster.local`.
2) Start application with StatsdMeterRegistry enabled and use host described above.
3) Change record in /etc/hosts file to another ip (`127.0.0.2` for example)

Actual result:
Using wireshark/tcpdump you can see that all metrics are being sent to 127.0.0.1 even after step 3.

Expected result:
StatsdMeterRegistry respect DNS name change.

Can be reproduced on real network/DNS server. Security manager disabled.
Micrometer version: 1.1.3, reproducible on 1.0.7 as well.

mweirauch:
Could this actually be related to the default JVM DNS caching behaviour? It's quite conservative (never expire) at times: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-jvm-ttl.html
Usually the host operating system network stack should do quite a good job at managing the DNS cache.

denzelby:
@mweirauch no. As I mentioned earlier the security manager is disabled and according to the docs: 
```
# default value is forever (FOREVER). For security reasons, this
# caching is made forever when a security manager is set. When a security
# manager is not set, the default behavior in this implementation
# is to cache for 30 seconds.
#
#networkaddress.cache.ttl=-1
```

On OS level everything set to defauls. Tested on desktop Ubuntu 18.04 LTS with Oracle jvm and debian-stretch in a Docker container (openjdk:8-jdk-stretch).

shakuzen:
I looked into this a bit. reactor-netty's `UdpClient`, which we use to send StatsD lines over UDP, only resolves the hostname on initial connection creation. It seems we would have to detect the change in resolution and then swap out the `UdpClient` for a new one which would resolve to the new IP address. Thus, this is not as simple as just changing a configuration option on the `UdpClient` unfortunately.

mootpt:
@shakuzen knowing very little about this issue, would setting the connection duration to hosts ttl be an option?

uchandroth:
[...]

PR #6330, opened 2025-06-02
No Fast Retry option to avoid retrying on the same IP indefinitely

arthurosncf:
In Kubernetes UDP servers (typically datadog), pods are not closed very well and IP are still reachable after autoscaling applied. This cause the statsd module to retry indefinitely on the same IP, skipping name resolution with host setup based on the hostname, even when the pod is killed (typically behind clusterIP load-balancing).

This setup is just to skip retrying and rebuild the whole configuration, each time Netty lose the UDP server port.

Issue with Kubernetes autoscaled services is mentioned in : #3563

To reproduce : 
- add two UDP servers on 127.0.0.2 and 127.0.03, with proper loopbacks listening on 8125, and DNS local cache "mylocalhost" pointing to 127.0.0.2
- Start Java service with StatsD module sending to host: mylocalhost:8125
- Check the 127.0.02 UDP server is receiving metrics
- Terminate 127.0.0.2 UDP server -> Netty Client is logging PortUnreachableException
- Update the local DNS cache to point to 127.0.0.3
- The Netty client is still logging PortUnreachableException indefinitely until the loopback is closed (IOException then)
- Only after IOException (one time), the client resolves the hostname and is redirected to 127.0.0.3

With "FastRetry" setup to False : 
- The previous last step is skipped and PortUnreachableException is causing resolution and redirection every time


shakuzen:
Thank you for the pull request. Are you sure this would solve your problem? We still have #1252 open. Ideally we should add tests demonstrating the issue and that it is solved with this change.

arthuroparis:
Thank you for considering the request, here is my analysis : 
The issue is related to the "connection" of the Reactor Netty UDP client : due to intermittent disruptions in the Kubernetes Service and lacks of Readyness probes, some pods are terminated prematurely and this results in the Kubernetes Service returning ICMP errors indefinitely, which in turn trigger PortUnreachableException.

I'm generally okay with the current connect() behavior of the UDP client and would prefer not to introduce any disruptive changes to Micrometer’s default Reactor Netty configuration. That’s why simply skipping retries in this context seems like a sufficient workaround to me.

Alternative workarounds could include : 
- Implementing a custom StatsSender that avoids calling connect(), thereby eliminating the ICMP port unreachable responses entirely.
- Adding a side-car to the Java Mircrometer container that also listens on UDP 8125 : that will drops the port unreachable packets for the Java container on the ICMP layer -and thus, avoiding PortUnreachableException

To properly unit test a “RetryOrIgnorePortUnreachableException” setup, we would need to:
- Mock the DNS resolution of the JVM (InetAddress)
- Mock the Netty Client

However, given the inherently network-related nature of the issue, I’m not convinced that writing such tests would be worthwhile.

arthuroparis:
I just reworded the setup to make it clearer.

shakuzen:
I haven't had a chance yet to look deeply into this, but I'm still missing something. I could try testing more later when I have time, but in your reproduction steps, you have:

> Only after IOException (one time), the client resolves the hostname and is redirected to 127.0.0.3

My understanding of #1252 is that that won't happen because DNS resolution only happens with Reactor Netty when the client is created. But are you saying DNS resolution is happening when the client disconnects and reconnects?
/cc @violetagg in case I'm misunderstanding anything or something has changed in Reactor Netty.

arthuroparis:
Yes i do,

IMHO, you are right, the Netty client will not resolve the host twice, and the non-resolved inetSocketAddress doesn't do the trick either.
However, when the flag is 'false' micrometer will loop over "prepareUdpClient" as long as the server sends port unreachable exceptions.
We did some tests extensively, and it seems it fixes the issue well, but only because we have setup an additional IP resolution within the configuration host() property.

The related code in the StatsDMeterRegistry context focus on building the "remoteAddress" supplier, when a "InetSocketAddress.createUnresolved" is used, this call was added with commit 24b19c7594a476ae69ef34e920a0850fb55173d4 (UDS datagram support in StatsD (#2722)), in replacement of the host directly transfered to the UdpClient builder before.

So, maybe should we replace : 
```
line 236 :
	prepareUdpClient(publisher,
       () -> InetSocketAddress.createUnresolved(statsdConfig.host(), statsdConfig.port()));
```
by 
```
line 236 :
	prepareUdpClient(publisher,
       () -> new InetSocketAddress(InetAddress.getByName(statsdConfig.host()), statsdConfig.port()));
```

What do you think ?

shakuzen:
> What do you think ?

I will defer to @violetagg on this. I don't remember why exactly I used `InetSocketAddress.createUnresolved` then. Will the UdpClient resolve the address when connecting with the proposed change above?

arthuroparis:
Hello,

I've just added a new update on the fork with two additions : 

- as mentioned early, we need to resolve the IP address on restarts. So i've just added the InetAddress resolution within the SocketAddress resolution, plus fallback to the "createUnresolved" in case the hostname cannot be resolved. Please note, we are using, in our setup, a local cache or the IP/Hostname pair, so the behavior regarding this may differs.
- The workaround configuration was initially put in place to fix a bug when the server is shutoff but still returning ICMP port unreachable packets. And yes, that makes the micrometer module works better. But we still have other issues, typically on cluster restarts, when, it seems, our UDP datadog pods are loaded afterward the Micrometer client or even OOM killed. In that case, the "loopback close" step applies. So after a little lecture of the Netty Reactor documentation i've figures out to access the underlying channel pipeline and add an additional name resolution step there too.

After some others unitary local tests, the fallback on cutting first UDP server loopback directly + change of DNS mapping to a second UDP server is passing.

I also send my test scripts attached for anyone wants to reproduces testing steps (on MacOs) : 
- Simple UDP server logging received metrics lines : [udp-server-127-0-0-2.py.txt](https://github.com/user-attachments/files/20836543/udp-server-127-0-0-2.py.txt)

- Open/Close loopbacks : [close-loopback-127-0-0-2.sh.txt](https://github.com/user-attachments/files/20836525/close-loopback-127-0-0-2.sh.txt) - [create-loopback-127-0-0-2.sh.txt](https://github.com/user-attachments/files/20836526/create-loopback-127-0-0-2.sh.txt)

- Local update of the DNS "mylocalhost" mappings: [mylocalhost.sh.txt](https://github.com/user-attachments/files/20836510/mylocalhost.sh.txt)

Finally please notice, as i have no time to make full tracing of the network exchanges here, my way to do is just to fix local near end-to-end unitary tests when i can reproduce falling use case, simply put.

arthuroparis:
Hello,

Following additional testing, the update as it is now works very well as soon as the DNS balancing is done **_before_** the previous pods are killed.
Unfortunately, if that's not the case the actual UDPClient cannot address the issue as it cannot loop or wait until the DNS is redirecting to a living UDP server. Worst, during HPA, if some of the pods are addressed by the DNS before the datadog service is up, they are killed automatically by lacking of Readyness Probes : in fact the container is receiving UDP packet and starts connections and then, the datadog service cannot starts as the port is allready in use.

So, maybe should we rename/update this Pull Request and aims to implement a full "Load Balancing over CoreDNS" UDPClient, relaying on the already done commits (don't retry on PortUnreachableException and Resolve IP address on connecting). That could be done by refactoring the prepareUdpClient method and callees : use instead a reactor "Flux.interval" that recreates a fresh UDPClient periodically that would be in relation with the BufferedFlux, or refactor the instanciation of this one too.

What do you think ?
Maybe an additional Pull-Request to be done or following on this one ?

PS : I'm aware of the failing unit test cases, i'm ok to work on that as soon as the PR will be almost finalized

Issue #6227, opened 2025-05-12
Add more tests for TimedHandler


shakuzen:
See https://github.com/micrometer-metrics/micrometer/pull/6194#discussion_r2081159987

PR #6343, opened 2025-06-03
Add more tests for TimedHandler

kwondh5217:
This pull request addresses gh-6227.

This test verifies that TimedHandler.shutdown() does not complete while any in-flight requests are still processing.

Specifically, it covers the following scenario:

- Two requests (/a and /b) are initiated concurrently

- TimedHandler.shutdown() is called while both are still in progress

- One request (/b) completes

- shutdownFuture should remain incomplete until the other request (/a) also completes

This test verifies the shutdown behavior using only the internal request/response flow, without externally calling shutdown.check(), in line with the discussion in https://github.com/micrometer-metrics/micrometer/pull/6194#discussion_r2081159987

### DIFF
--- a/micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
+++ b/micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
@@ -15,6 +15,7 @@
  */
 package io.micrometer.jetty12.server;
 
+import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.MockClock;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.binder.http.Outcome;
@@ -44,7 +45,7 @@ class TimedHandlerTest {
 
     private SimpleMeterRegistry registry;
 
-    private TimedHandler timedHandler;
+    private TestableTimedHandler timedHandler;
 
     private Server server;
 
@@ -55,7 +56,7 @@ class TimedHandlerTest {
     @BeforeEach
     void setup() {
         this.registry = new SimpleMeterRegistry(SimpleConfig.DEFAULT, new MockClock());
-        this.timedHandler = new TimedHandler(registry, Tags.empty());
+        this.timedHandler = new TestableTimedHandler(registry, Tags.empty());
 
         this.server = new Server();
         this.connector = new LocalConnector(server);
@@ -240,6 +241,8 @@ class TimedHandlerTest {
             HttpTester.Response responseB = HttpTester.parseResponse(endpoint2.getResponse());
             assertThat(responseB.getStatus()).isEqualTo(200);
 
+            assertThat(timedHandler.awaitOnComplete(5, TimeUnit.SECONDS)).isTrue();
+
             assertThat(shutdownFuture.isDone()).isFalse();
 
             requestAFinish.countDown();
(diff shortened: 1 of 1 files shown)


micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
```diff
@@ -182,6 +182,64 @@ public boolean handle(Request request, Response response, Callback callback) {
...
+
+        try (LocalConnector.LocalEndPoint endpointA = connector.connect();
+                LocalConnector.LocalEndPoint endpointB = connector.connect()) {
+            // Request A: delayed
+            endpointA.addInputAndExecute("GET /slow HTTP/1.1\r\nHost: localhost\r\n\r\n");
+
+            // Request B: quick response
+            endpointB.addInputAndExecute("GET /fast HTTP/1.1\r\nHost: localhost\r\n\r\n");
+            HttpTester.Response responseB = HttpTester.parseResponse(endpointB.getResponse());
+            assertThat(responseB.getStatus()).isEqualTo(HttpStatus.OK_200);
```
shakuzen: This is a bit different than what we wanted to test because request B starts and finishes before shutdown starts. I think what we want is to start two requests, start shutdown, finish one request, assert shutdown isn't finished, finish the other request, assert shutdown completes.

kwondh5217: Thanks for the review!
I’ve updated the test as suggested
please take another look when you have time 🙇

micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
```diff
@@ -182,6 +182,68 @@ public boolean handle(Request request, Response response, Callback callback) {
...
+            endpoint2.addInputAndExecute("GET /b HTTP/1.1\r\nHost: localhost\r\n\r\n");
+
+            assertThat(requestAStarted.await(5, TimeUnit.SECONDS)).isTrue();
+            assertThat(requestBStarted.await(5, TimeUnit.SECONDS)).isTrue();
+
+            Future<Void> shutdownFuture = timedHandler.shutdown();
+            assertThat(shutdownFuture.isDone()).isFalse();
+
+            requestBFinish.countDown();
+            assertThat(shutdownFuture.isDone()).isFalse();
```
shakuzen, line 246: Before this we should assert that we've received the response to request B. We want to make sure that `shutdown.check()` has been called before we assert on the shutdownFuture. I'm not even sure if that's sufficient to ensure that's happened or this will be a race condition.

shakuzen (commented):
Thanks for the quick update. I've left another review.

micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
kwondh5217: Thanks for the clarification!
I added the response check for request B as suggested.
Please take another look when you have time 🙇

shakuzen: Thanks for making that change. I've done some debugging and unfortunately it looks like my suspicion about the race was correct. Just because we've parsed the response does not mean that the `onComplete` callback in `TimedHandler` has finished (which is where `shutdown.check()` is called). The `onComplete` callback is called after the response is written. I'm not immediately sure how best to change the test to verify what we want. Maybe we need to extend TimedHandler to override `onComplete` and call another latch after `super.onComplete` so we will be able to wait until that is done to make the assertion. I'd be happy to consider other ideas for how to work around this.

kwondh5217: Sounds good, thanks for confirming the race.
I’ll go ahead and extend TimedHandler, override onComplete, and add a latch as you described.
I’ll push an updated commit shortly. 👍

kwondh5217:
@shakuzen Thanks for the review!

I updated the test to wait for the `onComplete()` callback by overriding it in a custom `TestableTimedHandler`, as you suggested.

Please take another look when you have time 🙇

shakuzen (approved):
Looks good to me. Thanks for the pull request and all the quick updates in response to feedback. I'll rebase it on `1.13.x` and merge forward from there.

Issue #6243, opened 2025-05-14
IndexProviderFactory throws ConcurrentModificationException


sfc-gh-dguy:
**Describe the bug**
While creating a DistributionSummary we get a ConcurrentModificationException.
When looking at the stack trace, 2 questions come to mind:
1. Why does the registration process seems to be nested in itself (see, for example, `io.micrometer.core.instrument.DistributionSummary$Builder.register(DistributionSummary.java:413)` that is shown twice in the stack trace)
2. Why does IndexProviderFactory uses a HashMap which is not thread safe

This is the stack trace:
```
[java.util.HashMap.computeIfAbsent(HashMap.java:1221)
io.micrometer.registry.otlp.internal.IndexProviderFactory.getIndexProviderForScale(IndexProviderFactory.java:41)
io.micrometer.registry.otlp.internal.Base2ExponentialHistogram.<init>(Base2ExponentialHistogram.java:91)
io.micrometer.registry.otlp.internal.CumulativeBase2ExponentialHistogram.<init>(CumulativeBase2ExponentialHistogram.java:52)
io.micrometer.registry.otlp.OtlpMeterRegistry.getHistogram(OtlpMeterRegistry.java:417)
io.micrometer.registry.otlp.OtlpMeterRegistry.getHistogram(OtlpMeterRegistry.java:394)
io.micrometer.registry.otlp.OtlpCumulativeDistributionSummary.<init>(OtlpCumulativeDistributionSummary.java:37)
io.micrometer.registry.otlp.OtlpMeterRegistry.newDistributionSummary(OtlpMeterRegistry.java:233)
io.micrometer.core.instrument.MeterRegistry.lambda$summary$6(MeterRegistry.java:368)
io.micrometer.core.instrument.MeterRegistry.getOrCreateMeter(MeterRegistry.java:668)
io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:606)
io.micrometer.core.instrument.MeterRegistry.summary(MeterRegistry.java:367)
io.micrometer.core.instrument.DistributionSummary$Builder.register(DistributionSummary.java:417)
io.micrometer.core.instrument.DistributionSummary$Builder.register(DistributionSummary.java:413)
io.micrometer.core.instrument.composite.CompositeDistributionSummary.registerNewMeter(CompositeDistributionSummary.java:85)
[...]

PR #6363, opened 2025-06-05
Fix Concurrent Modification Exception in Exponential Histogram

lenin-jaganathan:
Fixes https://github.com/micrometer-metrics/micrometer/issues/6243

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/internal/DefaultExponentialHistogramSnapShot.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/internal/DefaultExponentialHistogramSnapShot.java
@@ -81,8 +81,10 @@ class DefaultExponentialHistogramSnapShot implements ExponentialHistogramSnapSho
     }
 
     static ExponentialHistogramSnapShot getEmptySnapshotForScale(int scale) {
-        return emptySnapshotCache.computeIfAbsent(scale, key -> new DefaultExponentialHistogramSnapShot(key, 0, 0,
-                EMPTY_EXPONENTIAL_BUCKET, EMPTY_EXPONENTIAL_BUCKET));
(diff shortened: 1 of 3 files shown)


shakuzen (approved):
Thanks!

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/internal/DefaultExponentialHistogramSnapShot.java
```diff
@@ -81,8 +81,10 @@ public boolean isEmpty() {
     }
 
     static ExponentialHistogramSnapShot getEmptySnapshotForScale(int scale) {
-        return emptySnapshotCache.computeIfAbsent(scale, key -> new DefaultExponentialHistogramSnapShot(key, 0, 0,
-                EMPTY_EXPONENTIAL_BUCKET, EMPTY_EXPONENTIAL_BUCKET));
+        synchronized (emptySnapshotCache) {
+            return emptySnapshotCache.computeIfAbsent(scale, key -> new DefaultExponentialHistogramSnapShot(key, 0, 0,
+                    EMPTY_EXPONENTIAL_BUCKET, EMPTY_EXPONENTIAL_BUCKET));
```
shakuzen, line 86: I approved and merged, but I do wonder if it's worth having a LRU cache for empty snapshots. We don't expire anything from the IndexProvider cache, and it seems like that will have the same number of entries as the empty snapshot cache. Maybe it's simpler to have an unbounded cache (with ConcurrentHashMap) for empty snapshots too. What do you think?

lenin-jaganathan: Yeah, in almost 99.999999% (any 9's) of cases, this will be in the range of about 10-20 entries here. A simple ConcurrentHashMap would do fine.

PR #6371, opened 2025-06-06
Tests showing Statsd metrics dropped after start before connection established

izeye:
See https://github.com/micrometer-metrics/micrometer/issues/2141#issuecomment-2926471786

### DIFF
--- a/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
+++ b/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
@@ -329,6 +329,45 @@ class StatsdMeterRegistryPublishTest {
         server.disposeNow();
     }
 
+    @Test
+    void counterIncrementDoesNotWorkWithoutSleepAfterRegistryStart() throws InterruptedException {
+        StatsdConfig config = StatsdConfig.DEFAULT;
+
+        CountDownLatch serverLatch = new CountDownLatch(1);
+        DisposableChannel server = startServer(StatsdProtocol.UDP, config.port(), serverLatch);
+
+        StatsdMeterRegistry registry = new StatsdMeterRegistry(config, Clock.SYSTEM);
+        Thread.sleep(1000);
+
+        Counter counter = Counter.builder("my.counter").register(registry);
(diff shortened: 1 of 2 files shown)


shakuzen:
Thanks for the tests. I'm not sure what we should do about this. It's mostly expected, even if it may surprising. The statsd registry is unique in that it sends metrics as they're recorded (though potentially buffered). We don't generally block creation of registries on connecting to the backend, and the statsd registry is no different. A question is: should it be? In the normal case where we can quickly connect to the backend, it seems reasonable to block on connecting before finishing starting the registry, so users know that recorded metrics can be sent. However, in other cases where connecting to the backend is slow or even unavailable initially, it would be bad to block, which potentially blocks the whole application's start-up. Another option might be to try to buffer some metrics waiting for initial connection, so they are flushed once a connection is established, avoiding the loss demonstrated by the test. I'll discuss with @jonatan-ivanov, but if others have feedback on this, feel free to share here.

shakuzen:
@jgrigg is this capturing what you're seeing? Could you explain your use case so we better understand how this is impacting users?

shakuzen:
Without user feedback on this, I'd lean toward documenting this behavior and leaving it for now.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
```diff
@@ -329,6 +329,45 @@ void receiveParallelMetricsSuccessfully(StatsdProtocol protocol) throws Interrup
         server.disposeNow();
     }
 
+    @Test
+    void counterIncrementDoesNotWorkWithoutSleepAfterRegistryStart() throws InterruptedException {
```
shakuzen: Are the names of the tests switched? It looks like this is showing it works with sleeping.

shakuzen (commented):
Given the lack of user feedback in weeks on this, would you mind updating the documentation to document this behavior? Then we can merge this. We could also do it separately if you don't have the time/desire to do the documentation update in this PR.

izeye:

I fixed the test names you pointed out in 8da76de and added a note about this as you suggested in 4d57048.

shakuzen (approved):

PR #6385, opened 2025-06-11
Migrate micrometer-core to jSpecify

shakuzen:
Migrates from our custom nullability annotations to the jSpecify nullability annotations.

See #5547

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -30,7 +30,6 @@ import io.micrometer.core.instrument.search.MeterNotFoundException;
 import io.micrometer.core.instrument.search.RequiredSearch;
 import io.micrometer.core.instrument.search.Search;
 import io.micrometer.core.instrument.util.TimeUtils;
-import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
 import java.time.Duration;
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
@@ -22,7 +22,6 @@ import io.micrometer.core.instrument.distribution.HistogramSupport;
 import io.micrometer.core.instrument.distribution.ValueAtPercentile;
 import io.micrometer.core.instrument.distribution.pause.PauseDetector;
 import io.micrometer.core.instrument.util.TimeUtils;
-import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.NullUnmarked;
 import org.jspecify.annotations.Nullable;
 


micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
```diff
@@ -27,7 +27,7 @@
  */
 public interface FunctionCounter extends Meter {
 
-    static <T> Builder<T> builder(String name, @Nullable T obj, ToDoubleFunction<T> f) {
+    static <T> Builder<T> builder(String name, T obj, ToDoubleFunction<T> f) {
```
shakuzen, line 30: Changing this broke a number of callsites with NullAway errors. I was trying to align this with FunctionTimer and also it wasn't apparent how passing a null `obj` would actually be handled eventually (I didn't check the implementations to find out).

micrometer-core/src/main/java/io/micrometer/core/lang/NonNullFields.java
shakuzen, line 1: Separate from this PR, we can probably consider removing these. They've been deprecated for a while and generally are only intended for internal use.

micrometer-core/src/main/java/io/micrometer/core/instrument/util/package-info.java
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 @NonNullApi
-@NonNullFields
+@NullMarked
```
shakuzen, line 16: I haven't done the migration for `@NonNullApi` yet and can't do it automatically until the code compiles. It doesn't currently due to the NullAway errors.

micrometer-core/src/main/java/io/micrometer/core/instrument/search/MeterNotFoundException.java
```diff
@@ -88,8 +87,7 @@ private String nameDetail() {
             return NOT_OK + " No meter that matches the name predicate was found.";
         }
 
-        @Nullable
-        private List<String> tagDetail() {
+        @Nullable private List<String> tagDetail() {
```
shakuzen: The annotation didn't get moved to the right place for some reason.

micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
jonatan-ivanov: I'm thinking if `@Nullable` makes sense here when `ToDoubleFunction` can accept null but since this is not marked `@Nullable` all the way down to the registry maybe we should do this first and see if there will be a use-case.

micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
```diff
@@ -296,25 +298,25 @@ public Iterable<Tag> getTagsAsIterable() {
             return tags;
         }
 
+
```
jonatan-ivanov: I guess this was accidental, the formatter should remove it (as well as the others above and below).

micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
```diff
@@ -89,6 +90,7 @@ enum Type {
      * @return return value from the applied function
      * @since 1.1.0
      */
+    @NullUnmarked // I can't figure out the magical combination that makes NullAway happy here
```
jonatan-ivanov: Let's get back to this once others are resolved so that there are no other moving parts?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
```diff
@@ -35,7 +35,7 @@
  * @author Jon Schneider
  */
 @NonNullApi
```
jonatan-ivanov: Can we remove this (here and in other classes)?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
```diff
@@ -35,7 +35,7 @@
...
-@NonNullFields
+@NullMarked
```
jonatan-ivanov, line 36: Shouldn't we do this in `package-info.java` and remove this (here and in other classes)?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
jonatan-ivanov: Ah, ignore me: https://github.com/micrometer-metrics/micrometer/pull/6385#discussion_r2140484247

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CacheMeterBinder.java
shakuzen: Yes, but that's adding nullability where it wasn't, which it'll probably be better (for sanity, if nothing else) to do in a separate change.

shakuzen: We should make sure all the package-info's have nullability annotations, but I was trying to avoid adding it to ones that didn't have it so we could focus the first round on just migrating the existing annotations.

micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
shakuzen: I've got past this with marking the object as non-nullable. I think that makes sense from the perspective: why register a FunctionCounter with a null backing object? Might as well not call register in that case.

micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
```diff
@@ -67,8 +67,7 @@ static Builder builder(Timed timed) {
      * @return The return value of {@code f}.
      * @throws Exception Any exception bubbling up from the callable.
      */
-    @Nullable
-    default <T> T recordCallable(Callable<T> f) throws Exception {
+    default <T> @Nullable T recordCallable(Callable<T> f) throws Exception {
```
shakuzen: I'm not sure how to properly mark this. The return value nullability should be the same as the nullability of the `Callable` passed here, which may or may not be nullable.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -69,7 +70,7 @@ public abstract class MeterRegistry {
     private static final WarnThenDebugLogger doubleRegistrationLogger = new WarnThenDebugLogger(MeterRegistry.class);
 
     // @formatter:off
-    private static final EnumMap<TimeUnit, String> BASE_TIME_UNIT_STRING_CACHE = Arrays.stream(TimeUnit.values())
+    private static final EnumMap<TimeUnit, @NonNull String> BASE_TIME_UNIT_STRING_CACHE = Arrays.stream(TimeUnit.values())
```
shakuzen: This was an attempt to get NullAway to understand that `get` on `EnumMap` does not return null, but it doesn't work - hence I had to suppress NullAway on `getBaseTimeUnitStr()`

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -517,8 +519,8 @@ public More more() {
      * @return The state object that was passed in so the registration can be done as part
      * of an assignment statement.
      */
-    @Nullable
-    public <T> T gauge(String name, Iterable<Tag> tags, @Nullable T stateObject, ToDoubleFunction<T> valueFunction) {
+    public <T> @Nullable T gauge(String name, Iterable<Tag> tags, @Nullable T stateObject,
```
shakuzen, line 524: Probably should be consistent with `Gauge` and `Function*` types. I'd vote for marking the state object as non-nullable for the reasons stated elsewhere: if the user knows the state object is nullable, they may as well not register a meter with it.

micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
```diff
@@ -131,17 +132,18 @@ default void record(Duration duration) {
...
      * Executes the Supplier {@code f} and records the time taken.
      * @param f Function to execute and measure the execution time.
      * @return The return value of {@code f}.
      * @since 1.10.0
      */
+    // boolean cast to Boolean cast back to boolean can't be null
+    @SuppressWarnings("NullAway")
     default boolean record(BooleanSupplier f) {
-        return record((Supplier<Boolean>) f::getAsBoolean);
+        return record((Supplier<@NonNull Boolean>) f::getAsBoolean);
```
shakuzen: Adding `@NonNull` to the generic type here didn't make a difference, IIRC.

micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
```diff
@@ -181,8 +189,7 @@ default double record(DoubleSupplier f) {
      * @return The return value of {@code f}.
      * @throws Exception Any exception bubbling up from the callable.
      */
-    @Nullable
-    <T> T recordCallable(Callable<T> f) throws Exception;
+    @Nullable <T> T recordCallable(Callable<T> f) throws Exception;
```
shakuzen: Wrong place for the annotation.

micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterRegistryConfigValidator.java
```diff
@@ -67,8 +68,8 @@ public static <M extends MeterRegistryConfig, T> Function<M, Validated<T>> check
         };
     }
 
-    public static <M extends MeterRegistryConfig, T> Function<M, Validated<T>> checkRequired(String property,
-            Function<M, T> getter) {
+    public static <M extends MeterRegistryConfig, T extends @Nullable Object> Function<M, Validated<T>> checkRequired(
```
shakuzen: This was an attempt to get rid of errors like the following:

```
implementations/micrometer-registry-azure-monitor/src/main/java/io/micrometer/azuremonitor/AzureMonitorConfig.java:72: error: [NullAway] referenced method returns @Nullable, but functional interface method java.util.function.Function.apply(T) returns @NonNull
                check("connectionString", AzureMonitorConfig::connectionString));
                                          ^
    (see http://t.uber.com/nullaway )
```

Some config methods will have a nullable return value; some won't. How do we mark that accordingly, or is this a limitation of what NullAway can understand?

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
```diff
@@ -54,9 +55,11 @@ abstract class AbstractTimeWindowHistogram<T, U> implements Histogram {
     private volatile long lastRotateTimestampMillis;
 
     @SuppressWarnings({ "unused", "FieldCanBeLocal" })
-    private volatile int rotating; // 0 - not rotating, 1 - rotating
+    private volatile int rotating;
 
-    @Nullable
+    // 0 - not rotating, 1 - rotating
```
shakuzen: I don't know why this comment got moved to another line, but we should keep it the same as before.

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/TimeWindowMax.java
```diff
@@ -49,7 +50,8 @@ public class TimeWindowMax {
 
     @SuppressWarnings("ConstantConditions")
     public TimeWindowMax(Clock clock, DistributionStatisticConfig config) {
-        this(clock, config.getExpiry().toMillis(), config.getBufferLength());
+        this(clock, Objects.requireNonNull(config.getExpiry()).toMillis(),
+                Objects.requireNonNull(config.getBufferLength()));
```
shakuzen, line 54: It feels a bit silly to sprinkle these `Objects.requireNonNull` calls everywhere, which will throw a NPE like would have been thrown without it... I guess it makes it more clear from reading the code where that could happen, at least, but the result is the same.

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/DefaultMeterObservationHandler.java
```diff
@@ -83,6 +83,8 @@ public void onStart(Observation.Context context) {
     }
 
     @Override
+    // TODO decide what to do about context.getName being Nullable
+    @SuppressWarnings("NullAway")
```
shakuzen, line 87: This is a problem where we're doing `Timer.builder(context.getName())` where the builder does not accept a null name - meters must have a non-null name. Ideally, a Context should always have a name. Perhaps we should have a default name when one isn't set?

shakuzen (commented):
Another review in the meantime. I still haven't managed to resolve all the NullAway errors to apply the migration of `@NonNullApi`

micrometer-core/src/test/java/io/micrometer/core/instrument/push/PushMeterRegistryTest.java
```diff
@@ -123,7 +127,9 @@ void doNotPublishAgainOnClose_whenScheduledPublishInProgress() throws Interrupte
         onClosePublishThread.join();
 
         assertThat(overlappingStepMeterRegistry.publishes).as("only one publish happened").hasSize(1);
-        Deque<Double> firstPublishValues = overlappingStepMeterRegistry.publishes.get(0);
+        // TODO: remove requireNonNull: https://github.com/uber/NullAway/issues/1219
+        Deque<Double> firstPublishValues = Objects.requireNonNull(overlappingStepMeterRegistry.publishes.get(0));
+        assertThat(firstPublishValues).isNotNull();
```
jonatan-ivanov, line 132: See https://github.com/micrometer-metrics/micrometer/issues/6399

micrometer-core/src/test/java/io/micrometer/core/instrument/push/PushMeterRegistryTest.java
```diff
@@ -185,7 +191,9 @@ void waitForScheduledPublishToFinish_whenClosedWhilePublishIsInProgress()
         closeThread.join();
 
         assertThat(registry.publishes).as("only one publish happened").hasSize(1);
-        Deque<Double> firstPublishValues = registry.publishes.get(0);
+        // TODO: remove requireNonNull: https://github.com/uber/NullAway/issues/1219
+        Deque<Double> firstPublishValues = Objects.requireNonNull(registry.publishes.get(0));
+        assertThat(firstPublishValues).isNotNull();
```
jonatan-ivanov, line 196: See https://github.com/micrometer-metrics/micrometer/issues/6399

micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
jonatan-ivanov: Not sure if this is a good example:
```java
FunctionCounter.builder("test", Sensor.class, ignored -> Sensor.getCount());
```
vs.
```java
FunctionCounter.builder("test", null, ignored -> Sensor.getCount());
```
Nor this, does it make sense if the null-check is in the `ToDoubleFunction`?
```java
@Nullable String value = null; // let's pretend this is "monothonic" with howMany

Integer howMany(@Nullable String value) {
    return value != null ? Integer.valueOf(value) : 0;
}

FunctionCounter.builder("test", value, this::howMany);
```
Since `FunctionCounter.builder` does not have an overload when the object is missing like `Gauge.builder` has.


We have `NullGaugeSample` and `SimpleMeterRegistryTest` that are misusing this contract, maybe we should remove the first, I fixed the second.

Also, it's a bit weird that `MeterRegistry::newGauge(Meter.Id id, @Nullable T obj, ...)` is nullable but the builders are not. Though maybe it should be `@Nullable` since the `Gauge.builder` has an overload with only a supplier.
But regardless, we should make these Gauge-like components consistent, let's talk about this on our next meeting.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -325,6 +325,8 @@ protected String getConventionName(Meter.Id id) {
      */
     protected abstract DistributionStatisticConfig defaultHistogramConfig();
 
+    // I don't know how to make NullAway understand the EnumMap can't return null
```
jonatan-ivanov: Should we create an issue?

shakuzen: Maybe. It's hard to tell sometimes what is intended behavior vs a bug vs a misunderstanding on my part with these things.

micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterRegistryConfigValidator.java
shakuzen: Perhaps this is the limitation described by https://github.com/uber/NullAway/issues/1075?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
```diff
@@ -158,8 +154,7 @@ private Method detectMethod(String name) {
         }
     }
 
-    @Nullable
-    private Class<?> getFirstClassFound(List<String> classNames) {
+    @Nullable private Class<?> getFirstClassFound(List<String> classNames) {
```
jonatan-ivanov:
```suggestion
    private @Nullable Class<?> getFirstClassFound(List<String> classNames) {
```

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/NullGaugeSample.java
```diff
@@ -26,6 +26,7 @@
  */
 public class NullGaugeSample {
 
+    @SuppressWarnings("NullAway")
     public static void main(String[] args) {
```
jonatan-ivanov, line 30: Should we delete this whole class?

micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterRegistryConfigValidator.java
```diff
@@ -54,7 +56,7 @@ public static <M extends MeterRegistryConfig> Validated<?> checkAll(M config,
      * @return A function which, given a configuration instance, validates the property.
      */
     @SuppressWarnings("unchecked")
-    public static <M extends MeterRegistryConfig, T> Function<M, Validated<T>> check(String property,
+    @NullUnmarked public static <M extends @NonNull MeterRegistryConfig, T> Function<M, Validated<T>> check(@NonNull String property,
```
jonatan-ivanov, line 59: I found this as a workaround for the error we had.
I also tied to help the compiler with explicit types (refactoring lambdas to `Function<...>` variables) without much luck.

Also opened https://github.com/micrometer-metrics/micrometer/issues/6404

(Inlining `@NullUnmarked` seems to be an issue with the formatter.)

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/NullGaugeSample.java
shakuzen: I don't have a strong opinion either way. I can see how it could be useful, but it's also a fact I've never once run the sample.

micrometer-core/src/main/java/io/micrometer/core/lang/NonNullFields.java
jonatan-ivanov: https://github.com/micrometer-metrics/micrometer/issues/6407

micrometer-core/src/main/java/io/micrometer/core/instrument/Meter.java
jonatan-ivanov: https://github.com/micrometer-metrics/micrometer/issues/6408

micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterRegistryConfigValidator.java
jonatan-ivanov: Also see: https://github.com/spring-io/spring-javaformat/issues/446

micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
jonatan-ivanov: We used `NullUnmarked` for `Observation`, I guess `Timer` has the same issue.

jonatan-ivanov: Maybe we can do this:
```java
default <T> @NullUnmarked T recordCallable(Callable<T> f) throws Exception { ... }
```
or maybe this? :|
```java
default <T extends @NullUnmarked Object> T recordCallable(Callable<T> f) throws Exception { ... }
```

micrometer-core/src/main/java/io/micrometer/core/instrument/Timer.java
jonatan-ivanov: Let's remove this because of `@SuppressWarnings("NullAway")` and file an issue with NullAway.

micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
shakuzen: I've update it to mark the return type as `@NullUnmarked`

micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
shakuzen: Let's tackle that in a separate issue before M1 but after merging this PR. It will be more manageable and trackable to separate out these kinds of decisions rather than have them buried in this massive PR.

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/DefaultMeterObservationHandler.java
shakuzen: Let's figure this out in a separate issue before M1, possibly when we migrate the observation modules

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: I've updated the comment. Maybe in the future there will be a way to specify what we want.

micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
shakuzen: https://github.com/micrometer-metrics/micrometer/issues/6416

micrometer-core/src/main/java/io/micrometer/core/instrument/observation/DefaultMeterObservationHandler.java
shakuzen: https://github.com/micrometer-metrics/micrometer/issues/6417

micrometer-core/src/main/java/io/micrometer/core/instrument/docs/MeterDocumentation.java
```diff
@@ -105,7 +104,8 @@ default KeyName[] getAdditionalKeyNames() {
      * @return {@link MeterDocumentation} for which you don't want to create a default
      * metric documentation
      */
-    default Enum<?> overridesDefaultMetricFrom() {
+    // TODO should this be nullable?
+    default @Nullable Enum<?> overridesDefaultMetricFrom() {
         return null;
```
shakuzen, line 109: I wasn't really sure how this is used and marking it `@Nullable` is probably fine. We can follow-up if we want to change it.

PR #6386, opened 2025-06-11
Return NaN for null KafkaMetrics values

ArneKlein:
* Adds a guard for null metric values in toDouble() to default to NaN instead.


jonatan-ivanov:
Thank you for the PR!
I rebased it on `1.13.x` and polished/fixed the tests.

PR #6400, opened 2025-06-13
Add JDK 24 build job

shakuzen:


shakuzen:
The Gradle wrapper fails with the following error when trying to run on the selected JDK 24 image (`bellsoft/liberica-openjdk-alpine:24.0.1`)

```
Starting a Gradle Daemon (subsequent builds will be faster)
Did not find udev library in operating system. Some features may not work.
Disk Store information requires libudev, which is not present.
```

That's annoying. I guess we don't need to use Alpine, but I figured why use something bulkier if we don't have to.

shakuzen:
And now when switching the jdk 24 image to debian, we get the following error...

```
> Configure project :
Git repository not found at /root/micrometer -- nebula-release tasks will not be available. Use the git.root Gradle property to specify a different directory.

FAILURE: Build failed with an exception.

* Where:
Build file '/root/micrometer/build.gradle' line: 57
```

We haven't been updating the nebula-release-plugin because it has released major versions since the one we are on and dependabot does not make PRs for major versions. We should probably adopt some strategy to avoid this kind of issue in the future where we are unaware of new major versions of things we want to track.

shakuzen:
Updating the nebula-release-plugin didn't solve the issue. Looking at the other build logs, this may be the cause:

```
Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.
```

shakuzen:
I've switched to the [gradle images](https://hub.docker.com/_/gradle), which is closer to the CircleCI convenience image we were using before. The images are quite big, which feels wasteful, but other JDK images tend to be missing things we need (git, udev: see previous comments), which means we need to customize the image, and I was trying to avoid doing that. Also, I don't think we need or benefit from Gradle being included in the image since we're using the wrapper.

Now the jdk 24 build is failing with the following error:

```
java.lang.UnsupportedOperationException: Setting a system-wide Policy object is not supported
        at java.base/java.security.Policy.setPolicy(Policy.java:114)
        at java.base/java.security.Policy.setPolicy(Policy.java:114)
        at aQute.launcher.Launcher.activate(Launcher.java:493)
        at aQute.launcher.Launcher.activate(Launcher.java:493)
        at aQute.launcher.Launcher.launch(Launcher.java:403)
        at aQute.launcher.Launcher.launch(Launcher.java:403)
        at aQute.launcher.Launcher.run(Launcher.java:185)
        at aQute.launcher.Launcher.run(Launcher.java:185)
        at aQute.launcher.Launcher.main(Launcher.java:161)
        at aQute.launcher.Launcher.main(Launcher.java:161)
        at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:169)
        at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:169)
        at aQute.launcher.pre.EmbeddedLauncher.findAndExecute(EmbeddedLauncher.java:134)
        at aQute.launcher.pre.EmbeddedLauncher.findAndExecute(EmbeddedLauncher.java:134)
        at aQute.launcher.pre.EmbeddedLauncher.main(EmbeddedLauncher.java:51)
        at aQute.launcher.pre.EmbeddedLauncher.main(EmbeddedLauncher.java:51)

> Task :micrometer-osgi-test:testOSGi FAILED
error  : Exit code remote process 127
```

This has already been fixed in the BND launcher with https://github.com/bndtools/bnd/pull/6371, but a new release containing the fix is not available yet.

gradle/libs.versions.toml
```diff
@@ -218,7 +218,7 @@ wiremockJunit5 = { module = "ru.lanwen.wiremock:wiremock-junit5", version.ref =
 # plugin dependencies
 plugin-errorprone = { module = "net.ltgt.errorprone:net.ltgt.errorprone.gradle.plugin", version = "4.2.0" }
 plugin-license = { module = "gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin", version = "0.16.1" }
-plugin-nebulaRelease = { module = "com.netflix.nebula:nebula-release-plugin", version = "18.0.8" }
+plugin-nebulaRelease = { module = "com.netflix.nebula:nebula-release-plugin", version = "20.2.0" }
```
shakuzen: Since this didn't end up fixing anything, we can revert it in this PR and look to upgrade these plugins separately, since dependabot doesn't do it for us for new major versions.

shakuzen:
@jonatan-ivanov This passes now and can be merged. We'll have to see if this can be backported to the oldest supported branch. I don't know if there will be some incompatibility with things in the older branches. And I figured it'd be best to include the Java 24 specific changes here (virtual thread tests) that won't be in the oldest branch.

.circleci/config.yml
```diff
@@ -1,13 +1,13 @@
...
 executors:
-  circle-jdk23-executor:
+  jdk24-executor:
     working_directory: ~/micrometer
     environment:
       GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
     resource_class: medium+
     docker:
-      - image: cimg/openjdk:23.0.2
+      - image: gradle:8-jdk24
```
jonatan-ivanov, line 10: I'm thinking if this works, we should migrate all of our images so that they are consistent.

micrometer-osgi-test/build.gradle
```diff
@@ -2,6 +2,12 @@ import aQute.bnd.gradle.Bundle
 import aQute.bnd.gradle.Resolve
 import aQute.bnd.gradle.TestOSGi
 
+// skip this module when building with jdk >=24
+// waiting for a release with the fix https://github.com/bndtools/bnd/pull/6371
+if (javaLanguageVersion.canCompileOrRun(24)) {
+    project.tasks.configureEach { task -> task.enabled = false }
+}
+
```
jonatan-ivanov, line 10: I created https://github.com/micrometer-metrics/micrometer/issues/6402

.circleci/config.yml
shakuzen: Consistency is a good goal. I was trying to change as little as possible and try out the new image without potentially breaking everything and needing to revert everything. If the trend continues of the CircleCI JDK convenience images not doing new releases, we'll also need to move off them to keep testing with the latest JDK patch releases.

chrisrueger:
FYI : bnd 7.2.0 Release is close: https://github.com/bndtools/bnd/issues/6974 Any help testing Release Candidate is appreciated.

PR #6415, opened 2025-06-17
Remove io.micrometer.core.lang annotations

### LINKED ISSUE #6407: Remove io.micrometer.core.lang annotations
Annotations in `io.micrometer.core.lang` have been deprecated for a while and intended for internal use.

From:
>Separate from this PR, we can probably consider removing these. They've been deprecated for a while and generally are only intended for internal use.

(shakuzen https://github.com/micrometer-metrics/micrometer/pull/6385#discussion_r2140482311)

See [`micrometer-core/src/main/java/io/micrometer/core/lang`](https://github.com/micrometer-metrics/micrometer/tree/af78e7d033ce879611395754819e23de60fc947e/micrometer-core/src/main/java/io/micrometer/core/lang)

MemencioPerez:
These annotations are not used in the core module code and have been deprecated for a long time. Closes #6407.

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -79,7 +79,6 @@ jar {
             rx.*;resolution:=dynamic;version="${@}",\
             javax.persistence.*;resolution:=dynamic;version="${@}",\
             io.netty.*;resolution:=dynamic;version="${@}",\
-            javax.annotation.*;resolution:=optional;version="${@}",\
             *
         '''.stripIndent()
     }


jonatan-ivanov:
Thanks for the PR. Right now we won't be able to merge it since:
- There are merge conflicts (please rebase on latest `main`)
- The build is failing (it should fail for you locally too, run `./gradlew build`), the japicpm check needs to be disabled for these annotations to make the build green again

MemencioPerez:
I have seen the following warning coming from the jar task:

````
> Task :micrometer-core:jar
warning: Unused Import-Package instructions: [javax.annotation.*] 
warning: Unused Import-Package instructions: [javax.annotation.*] 
````

Is it out of the scope of this PR to remove that unused Import-Package instruction or can I make another commit with the change?

File: micrometer-core/build.gradle [hunk 1]
```diff
@@ -282,3 +282,5 @@ tasks.register("openj9ConcurrentScavengeTest", Test) {
 
     jvmArgs '-Xgc:concurrentScavenge'
 }
+
+japicmp.enabled = false
```
jonatan-ivanov: When I said
> the japicpm check needs to be disabled for these annotations to make the build green again

I really meant only the annotations, not the whole module. :)
This completely disables checking everything in micrometer-core which is definitely something we don't want.

jonatan-ivanov (changes requested):

micrometer-core/build.gradle
MemencioPerez: Sorry for the misunderstanding. I have reverted that commit and added ``io.micrometer.core.lang`` to the package exclusions in the main build.gradle file.

MemencioPerez:
> I have seen the following warning coming from the jar task:
> 
> ```
> > Task :micrometer-core:jar
> warning: Unused Import-Package instructions: [javax.annotation.*] 
> warning: Unused Import-Package instructions: [javax.annotation.*] 
> ```
> 
> Is it out of the scope of this PR to remove that unused Import-Package instruction or can I make another commit with the change?

Any opinion on this matter?

shakuzen:
> Is it out of the scope of this PR to remove that unused Import-Package instruction or can I make another commit with the change?

Is this caused by the BND plugin and the fact micrometer-core no longer uses the `javax.annotation` package anymore? If so, I think it makes sense to remove that line from the BND config to get rid of the warning in this PR.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you very much!

PR #6420, opened 2025-06-17
Migrate micrometer-observation and micrometer-observation-test to JSpecify

jonatan-ivanov:
See gh-5547

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
@@ -447,14 +447,14 @@ public class ObservationContextAssert<SELF extends ObservationContextAssert<SELF
         return (SELF) this;
     }
 
-    @SuppressWarnings("NullAway") // see: https://github.com/uber/NullAway/issues/1222
     private ObservationView checkedParentObservation() {
         isNotNull();
         ObservationView p = this.actual.getParentObservation();
         if (p == null) {
             failWithMessage("Observation should have a parent");
         }
-        return p;
+        // see: https://github.com/uber/NullAway/issues/1222
+        return Objects.requireNonNull(p);
     }
 
     /**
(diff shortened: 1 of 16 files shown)


micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
```diff
@@ -467,7 +469,7 @@ public SELF hasParentObservationEqualTo(Observation expectedParent) {
         if (realParent == null) {
             failWithMessage("Observation should have parent <%s> but has none", expectedParent);
         }
-        if (!realParent.equals(expectedParent)) {
+        if (!Objects.equals(realParent, expectedParent)) {
```
jonatan-ivanov, line 472: This is because of the same issue as above (https://github.com/uber/NullAway/issues/1222).

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
```diff
@@ -446,6 +447,7 @@ public SELF hasParentObservation() {
         return (SELF) this;
     }
 
+    @SuppressWarnings("NullAway") // see: https://github.com/uber/NullAway/issues/1222
```
jonatan-ivanov: Maybe `return Objects.requireNonNull(p);` is better than `@SuppressWarnings`.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationRegistryCompatibilityKit.java
```diff
@@ -1004,15 +1004,13 @@ public boolean supportsContext(Observation.Context context) {
 
     static class TestObservationConventionWithNameOverrides implements GlobalObservationConvention<TestContext> {
 
-        @Nullable
         @Override
-        public String getName() {
+        public @Nullable String getName() {
             return "conventionOverriddenName";
```
shakuzen, line 1009: I wonder if we shouldn't remove the `@Nullable` here. I guess it complains about overriding a `@Nullable` method then? We actually know this override is not nullable, although I doubt it matters either way for this class.

shakuzen (approved):

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationRegistryCompatibilityKit.java
jonatan-ivanov: Did it in the next PR (docs + formatting).

PR #6433, opened 2025-06-19
Missing @NullMarked

jonatan-ivanov:


implementations/micrometer-registry-dynatrace/src/main/java/io/micrometer/dynatrace/v2/DynatraceExporterV2.java
```diff
@@ -152,7 +153,7 @@ public void export(@NonNull List<Meter> meters) {
             return;
         }
 
-        Map<String, String> seenMetadata = null;
+        Map<String, @Nullable String> seenMetadata = null;
```
shakuzen, line 154: I can't comment on the line, but I noticed the signature for this method is `public void export(@NonNull List<Meter> meters) {`. The `@NonNull` could be removed now that the package is `@NullMarked`.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpHttpMetricsSender.java
```diff
@@ -45,7 +47,8 @@ public OtlpHttpMetricsSender(HttpSender httpSender) {
      */
     @Override
     public void send(Request request) throws Exception {
-        HttpSender.Request.Builder httpRequest = this.httpSender.post(request.getAddress())
+        String address = request.getAddress() != null ? request.getAddress() : DEFAULT_ADDRESS;
+        HttpSender.Request.Builder httpRequest = this.httpSender.post(address)
```
shakuzen: I wonder if throwing an exception would be better. If someone is using the OtlpHttpMetricsSender but not giving an address, silently trying to send to `http://localhost` is a bit surprising and unlikely to work anyway.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpConnectionPoolMetrics.java
```diff
@@ -96,8 +96,8 @@ public OkHttpConnectionPoolMetrics(ConnectionPool connectionPool, String namePre
      * exposed by this instance. Therefore this binder allows to pass it, to be able to
      * monitor it.
      */
-    public OkHttpConnectionPoolMetrics(ConnectionPool connectionPool, String namePrefix, Iterable<Tag> tags,
-            Integer maxIdleConnections) {
+    public OkHttpConnectionPoolMetrics(@Nullable ConnectionPool connectionPool, @Nullable String namePrefix,
+            @Nullable Iterable<Tag> tags, @Nullable Integer maxIdleConnections) {
```
shakuzen: It looks like we're throwing exceptions immediately if connectionPool, namePrefix, or tags are null. It seems like we're making things worse by marking them nullable when we don't actually want people to pass null.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmHeapPressureMetrics.java
```diff
@@ -125,10 +126,15 @@ private void monitor() {
...
+                        .mapToLong(pool -> getUsed(after.get(pool)))
                         .sum();
-                    double maxAfter = longLivedPoolNames.stream().mapToLong(pool -> after.get(pool).getMax()).sum();
-                    lastLongLivedPoolUsageAfterGc.set(usedAfter / maxAfter);
+                    double maxAfter = longLivedPoolNames.stream().mapToLong(pool -> getMax(after.get(pool))).sum();
+                    if (usedAfter > 0 && maxAfter > 0) {
+                        lastLongLivedPoolUsageAfterGc.set(usedAfter / maxAfter);
+                    }
+                    else {
+                        lastLongLivedPoolUsageAfterGc.set(0.0);
```
shakuzen: I suppose in the scenarios now covered, an exception would have been thrown before, and this would not have been updated. Maybe it's better to preserve that behavior? Although there's no way for a user to notice the issue.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpConnectionPoolMetrics.java
jonatan-ivanov: I remember I was doing different things on it and I think this might be a remnant of it otherwise I have zero idea why have I done this.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmHeapPressureMetrics.java
jonatan-ivanov: I think this scenario is already somewhat invalid since `Set<String> longLivedPoolNames` is initialized this way in the ctor:
```java
longLivedPoolNames = JvmMemory.getLongLivedHeapPools()
    .map(MemoryPoolMXBean::getName)
    .collect(Collectors.toSet());
```
This means that this can only happen if the JVM is somehow inconsistent and `gcInfo.getMemoryUsageAfterGc()` does not contain usage info for any of the pools names in `longLivedPoolNames`.

If I simulate this by initializing made-up `longLivedPoolNames`, no exception is thrown. Setting `NaN` might be better (with an extra log entry?) or we can also `@SuppressWarnings` and ignore this case (NPE if it happens).

shakuzen: I'd err on the side of not changing this logic unless we have a good reason to (e.g. a bug report). We could wrap the get calls in `Objects.requireNonNull` to please NullAway, though it doesn't change the result (NPE thrown). What do you think?

jonatan-ivanov: I think that's a good idea. Fixed it and also did a fresh rebase.

shakuzen (approved):

PR #6450, opened 2025-06-21
Remove replace call that does nothing in TelegrafStatsdLineBuilder.telegrafEscape()

izeye:
While looking into https://github.com/micrometer-metrics/micrometer/pull/6449, I noticed that the `TelegrafStatsdLineBuilder.telegrafEscape()` doesn't escape '=' properly.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilder.java
+++ b/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilder.java
@@ -109,7 +109,7 @@ public class TelegrafStatsdLineBuilder extends FlavorStatsdLineBuilder {
     // backslash escape =
     // trying to escape spaces and comma drops everything after that
     private String telegrafEscape(String value) {
-        return value.replaceAll("=", "\\=").replaceAll("[\\s,:]", "_");
+        return value.replaceAll("[\\s,:]", "_");
     }
 
 }


izeye:
I'm not familiar with Telegraf StatsD, but my testing with Telegraf StatsD and InfluxDB showed that escaping '`=`' with '`\`' (backslash) doesn't seem to work.

It might be better to sanitize it to '`_`' (underscore) like other special characters (whitespaces, '`,`', and '`:`').

izeye:
Another observation is that '`=`' without '`\`' seems okay for meter names and tag values, but not for tag keys.

shakuzen:
I wonder if something changed since that code was written. It'd be good to understand if there's a difference in behavior between different versions of telegraf.

shakuzen:
I tried to look into this but cannot find specific documentation on it. [Here](https://www.influxdata.com/blog/getting-started-with-sending-statsd-metrics-to-telegraf-influxdb/#heading5) it says (emphasis added):

> Since InfluxDB supports tags, our StatsD implementation does too! Adding tags to a StatsD metric is **similar** to how they appear in [line-protocol.](https://influxdb.com/docs/v0.9/write_protocols/line.html)

It says similar but that leaves the possibility it isn't exactly the same. Going to the documentation on Influx's line-protocol there is [this section](https://docs.influxdata.com/influxdb/v2/reference/syntax/line-protocol/#special-characters), which describes needing to escape commas, spaces, and equals signs depending on the place.

Maybe this pull request isn't the place to solve this mystery, though. Instead, perhaps it's better to just remove the equals part of the replacement, since it wasn't doing anything this whole time and we have no reports of bugs and no clear documentation on what it should be doing.

shakuzen:
@izeye could we remove the replacement for `=` in this change, since that eliminates pointless code while achieving the same result we had before? Then in a separate change, let's try to figure out the right escaping and see if we can integration test it so we aren't guessing about what's accepted or not. What do you think?

izeye:

Sounds good. I applied your suggestion and switched the base branch to the 1.13.x branch.

shakuzen (approved):

PR #6451, opened 2025-06-21
Declare default @Retention on @Incubating explicitly

izeye:
This PR removes explicit default `@Retention` declarations.

See https://github.com/spring-projects/spring-framework/issues/35027

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/annotation/Incubating.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/annotation/Incubating.java
@@ -15,15 +15,14 @@
  */
 package io.micrometer.core.annotation;
 
-import java.lang.annotation.Documented;
-import java.lang.annotation.ElementType;
-import java.lang.annotation.Target;
+import java.lang.annotation.*;
 
 /**
  * This is a feature that is experimental and may yield a breaking change in a minor
  * release.
  */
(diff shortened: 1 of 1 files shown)


shakuzen (commented):
I'm not sure how common knowledge the default retention policy is. I don't know that (or have forgotten that), for example. In that sense, it might actually be nice it is explicitly there when reading the code, even if it doesn't change anything programmatically. @jonatan-ivanov thoughts?

izeye:

I'm okay with either way.

If we want to keep the default, the following needs to be updated instead for consistency:

https://github.com/micrometer-metrics/micrometer/blob/b6269081003b74415e138b9abf430b5db862d077/micrometer-core/src/main/java/io/micrometer/core/annotation/Incubating.java#L22-L32

jonatan-ivanov:
I don't have a strong opinion but I'm leaning towards spelling this out explicitly.
@izeye, if it's ok with you, would you mind do the  change on `@Incubating` instead?

izeye:
@jonatan-ivanov Sure, I updated this as suggested.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you!

jonatan-ivanov (approved):

Issue #6401, opened 2025-06-13
Distribution without percentiles always has value 0 in Stackdriver


rafal-dudek:
**Describe the bug**
StackdriverMeterRegistry is always sending 0 for count and sum for distribution metrics when percentiles are not defined.
E.g. "jvm.gc.pause"
https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java#L210
```
Timer.builder("jvm.gc.pause")
    .tags(gcTags)
    .description("Time spent in GC pause")
    .register(registry)
    .record(duration, TimeUnit.MILLISECONDS);
```

Since 1.14.4, we are observing only 0 in Stackdriver for count and sum:

![Image](https://github.com/user-attachments/assets/1cc0d68e-3df2-48ff-a8dd-7dab16c9571a)

**Environment**

 - Micrometer version: 1.14.4 - 1.15.1 (It was working correctly in 1.14.3)
 - Micrometer registry: stackdriver
 - OS: Windows 11
 - Java version: 21.0.2

**To Reproduce**
How to reproduce the bug:

Configure application to use Stackdriver. Invoke `System.gc();`.
Observer metrics in Cloud Monitoring for jvm.gc.pause e.g. in PromQL:
custom_googleapis_com:jvm_gc_pause_sum{monitored_resource="global"}

Alternatively, create other Timer metric without percentiles and record values in code e.g.:
```
Timer timer = Timer.builder("test.metric")
                .description("Test metric")
                .register(meterRegistry);
...
timer.record(100, TimeUnit.MILLISECONDS);
```

**Expected behavior**
Count and sum from Distribution metrics in Stackdriver should show real values when percentiles/buckets are disabled, not zeroes.

**Additional context**
"jvm.gc.pause.max" is working correctly, as this is a separate metric created next to "jvm.gc.pause".
Chart for max for the same time range:
![Image](https://github.com/user-attachments/assets/484f7bcb-df62-41f4-888d-de8f73c39705)

shakuzen:
Thank you for the report. The fix for https://github.com/micrometer-metrics/micrometer/issues/4868 in 1.14.4 is to blame. I was able to reproduce it with the following test:

```java
@Test
@Issue("#6401")
void distributionWithNoBucketsUsesStepCount() {
    StackdriverMeterRegistry.Batch batch = meterRegistry.new Batch();
[...]

PR #6452, opened 2025-06-23
Always configure histogram for Stackdriver

shakuzen:
A distribution is always published to Stackdriver for Timer and DistributionSummary meters. We want this Distribution to have an accurate count in the time window of the Distribution (histogram). A no-op Histogram was being used when no buckets and no percentiles were configured, and that always has a zero count. We can't use the step count because that would not be the count for the same period of time. This will add some overhead to Timer and DistributionSummary meters that did not have any buckets or percentiles configured.

Fixes gh-6401

### DIFF
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
@@ -31,7 +31,7 @@ final class StackdriverHistogramUtil {
         if (distributionStatisticConfig.isPublishingHistogram()) {
             return new StackdriverFixedBoundaryHistogram(clock, distributionStatisticConfig);
         }
-        return NoopHistogram.INSTANCE;
+        return new StackdriverInfinityBucketHistogram(clock, distributionStatisticConfig);
     }
 
     static class StackdriverClientSidePercentilesHistogram extends TimeWindowPercentileHistogram {
@@ -47,6 +47,14 @@ final class StackdriverHistogramUtil {
 
         StackdriverFixedBoundaryHistogram(Clock clock, DistributionStatisticConfig config) {
             super(clock, config, true, false, true);
+        }
+
+    }
+
+    static class StackdriverInfinityBucketHistogram extends TimeWindowFixedBoundaryHistogram {
+
+        StackdriverInfinityBucketHistogram(Clock clock, DistributionStatisticConfig config) {
+            super(clock, config, false, false, true);
         }
 
     }
--- a/implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverMeterRegistryTest.java
+++ b/implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverMeterRegistryTest.java
@@ -21,6 +21,7 @@ import io.micrometer.core.Issue;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.MockClock;
 import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.distribution.CountAtBucket;
 import io.micrometer.core.instrument.distribution.HistogramSnapshot;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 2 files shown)


implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
```diff
@@ -51,4 +51,11 @@ static class StackdriverFixedBoundaryHistogram extends TimeWindowFixedBoundaryHi
 
     }
 
+    static class StackdriverInfinityBucketHistogram extends TimeWindowFixedBoundaryHistogram {
+
+        StackdriverInfinityBucketHistogram(Clock clock, DistributionStatisticConfig config) {
+            super(clock, config, false, false, true);
```
shakuzen, line 57: While technically this config could have buckets or percentiles in it, we know at the one place where it is called from it doesn't.
We could write a specialized implementation that is simpler and more efficient than `TimeWindowFixedBoundaryHistogram` for this use case (a time-window histogram with only the infinity bucket), but it was easiest to extend `TimeWindowFixedBoundaryHistogram`.

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverMeterRegistryTest.java
```diff
@@ -196,4 +196,19 @@ void meanIsZeroWhenCountIsZero() {
...
+        StackdriverMeterRegistry.Batch batch = meterRegistry.new Batch();
+        // halfway through first step
+        clock.add(config.step().dividedBy(2));
+        DistributionSummary ds = DistributionSummary.builder("ds").register(meterRegistry);
+        ds.record(3);
+        // 1/4 through the second step
+        clock.add(config.step().dividedBy(4).multipliedBy(3));
+        Distribution distribution = batch.distribution(ds.takeSnapshot(), false);
+        System.out.println(distribution);
+        assertThat(distribution.getCount()).isEqualTo(1);
```
shakuzen: Coincidentally the distribution count is equal to the step count here. It may be better to test this in a setup when they wouldn't be equal.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
```diff
@@ -51,4 +51,12 @@ static class StackdriverFixedBoundaryHistogram extends TimeWindowFixedBoundaryHi
...
```
shakuzen, line 57: To explain the parameters here: 

* supportsAggregablePercentiles = false, so there will be no buckets from the percentile histogram buckets
* isCumulativeBucketCounts = false, like the other histograms
* includeInfinityBucket = true, to include the one bucket we want

If there were SLOs configured, those would be added too, but as stated above, we know this doesn't get called unless there are no buckets and no client percentiles.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
shakuzen: I ran some benchmarks. There is more allocation when registering a new Timer/DistributionSummary that would have used a NoopHistogram but instead uses StackdriverInfinityBucketHistogram with this change. In a benchmark that registers and removes a new Timer with the different versions, it is 1136 bytes per operation (Micrometer 1.14.8 using NoopHistogram) vs 1992 bytes per op (this PR using StackdriverInfinityBucketHistogram).

Recording a value is also a bit slower but still quite fast (0.214 ± 0.008 us/op VS 0.240 ± 0.008 us/op on my laptop).

@rafal-dudek heads up on the added overhead in case you have any concerns.

I did try a custom Histogram implementation extending `AbstractTimeWindowHistogram`, and it gets the allocation down to 1296 bytes per op, but because AbstractTimeWindowHistogram is package private, we cannot easily use it or we need to make a new public histogram implementation for Stackdriver to use or copy a lot of code to the Stackdriver module. I'd rather avoid this in a maintenance release. Of course I'd also like to avoid increased allocation in a maintenance release too, but that's unavoidable moving from a no-op implementation.

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverMeterRegistryTest.java
shakuzen: I've updated the test so this isn't the case anymore.

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
jonatan-ivanov: Would it make sense to make `StackdriverInfinityBucketHistogram` implement `Histogram`  and copy a simplified version of `AbstractTimeWindowHistogram` into it?
If I understand it correctly we need a single counter in a ring buffer?

jonatan-ivanov (approved):

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverHistogramUtil.java
shakuzen: That would work, but it comes with the risk that in a patch release we introduce some new bug in the implementation of a Histogram that's never been tested before. Avoiding another bug is top priority in a patch release. So I thought this is the least risk for a functional bug.

The "custom Histogram implementation extending `AbstractTimeWindowHistogram`" that I mentioned in my comment uses `AbstractTimeWindowHistogram` with an `AtomicLong` in the ring buffer. I needed to make it public in micrometer-core since `AbstractTimeWindowHistogram` is package-private.

Making `AbstractTimeWindowHistogram` public would be easy, but I'd like to do a proper review of it and maybe change some things before making it public. We could consider doing that in 1.16, resources permitting.

So we're left with the tradeoff between more memory usage and less functional bug risk. I think we'll go with this for now until further feedback from users.

shakuzen:
I rebased on `1.13.x` because I realized the original change was in that branch.

Issue #6434, opened 2025-06-20
Scrape failure due to ClassCastException when conflicting meter types are registered


PaddyDrury:
I raised an issue on Spring Boot relating to scrape failures I observed when I had caches managed by RedisCacheManager and JCacheCacheManager in a Spring Boot application: https://github.com/spring-projects/spring-boot/issues/46017

I found that an incompatibility with meter types registered by RedisCacheMetrics and by micrometer's JCacheMetrics causes a ClassCastException to occur when calling my application's Prometheus scrape endpoint.

It looks like RedisCacheMetrics registers a counter for "cache.removals" whereas JCacheMetrics registers a gauge:

https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/cache/RedisCacheMetrics.java#L71

https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java#L131

Here is a sample application to reproduce the error: https://github.com/PaddyDrury/redis-jcache-prometheus

Here is a stack trace:

```
java.lang.ClassCastException: class io.prometheus.metrics.model.snapshots.CounterSnapshot$CounterDataPointSnapshot cannot be cast to class io.prometheus.metrics.model.snapshots.GaugeSnapshot$GaugeDataPointSnapshot (io.prometheus.metrics.model.snapshots.CounterSnapshot$CounterDataPointSnapshot and io.prometheus.metrics.model.snapshots.GaugeSnapshot$GaugeDataPointSnapshot are in unnamed module of loader 'app')
	at io.prometheus.metrics.expositionformats.PrometheusTextFormatWriter.writeGauge(PrometheusTextFormatWriter.java:130) ~[prometheus-metrics-exposition-textformats-1.3.6.jar:na]
	at io.prometheus.metrics.expositionformats.PrometheusTextFormatWriter.write(PrometheusTextFormatWriter.java:69) ~[prometheus-metrics-exposition-textformats-1.3.6.jar:na]
	at org.springframework.boot.actuate.metrics.export.prometheus.PrometheusOutputFormat$1.write(PrometheusOutputFormat.java:48) ~[spring-boot-actuator-3.5.0.jar:3.5.0]
[...]

PR #6470, opened 2025-06-26
Do not register the same Prometheus Metric Family with different types

shakuzen:
See gh-6434

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
@@ -44,12 +44,15 @@ class MicrometerCollector implements MultiCollector {
 
     private final String conventionName;
 
+    private final Meter.Type meterType;
+
     private final List<String> tagKeys;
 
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -44,12 +44,15 @@ class MicrometerCollector implements MultiCollector {
...
 
+    private final Meter.Id id;
+
     private final List<String> tagKeys;
 
     // take name to avoid calling NamingConvention#name after the call-site has already
     // done it
     MicrometerCollector(String name, Meter.Id id, NamingConvention convention) {
         this.conventionName = name;
+        this.id = id;
```
shakuzen: If we only want this for the `Type`, maybe we should just store that instead of the `Id`? It's a bit confusing what a MicrometerCollector's `Id` is since it is a collector for potentially multiple IDs.

jonatan-ivanov: 👍🏼 I changed it and rebased it on top of `1.13.x`.

PR #6487, opened 2025-06-30
Remove redundant @NonNull annotations

izeye:
This PR removes `@NonNull` annotations that seem to be redundant.


shakuzen (approved):
Thanks

PR #6508, opened 2025-07-04
Use JoinPoint in Annotation Handler

doumdoum:
It'll be needed in another PR that uses before aspect.

It breaks current API but is more general since ProceedingJoinPoint inherits directly from JoinPoint

### DIFF
--- a/micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
@@ -85,7 +85,7 @@ public class AnnotationHandler<T> {
     /**
      * Modifies the object with {@link KeyValue} related information.
      * @param objectToModify object to modify
-     * @param jp proceeding join point
+     * @param jp join point
      */
     public void addAnnotatedParameters(T objectToModify, JoinPoint jp) {
         try {
@@ -104,7 +104,7 @@ public class AnnotationHandler<T> {
     /**
      * Modifies the object with {@link KeyValue} information based on the method result.
      * @param objectToModify object to modify
-     * @param jp proceeding join point
+     * @param jp join point
      * @param result method return value
      * @since 1.15.0
      */
--- a/micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
@@ -18,7 +18,7 @@ package io.micrometer.common.annotation;
 import io.micrometer.common.KeyValue;
 import io.micrometer.common.util.internal.logging.InternalLogger;
 import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
-import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.JoinPoint;
 import org.aspectj.lang.reflect.MethodSignature;
 import org.jspecify.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


doumdoum:
Hey, anyone to review/ comment? Thx

jonatan-ivanov:
The build is failing because you introduced a breaking change.
I'm also not sure I see why this is needed.

doumdoum:
@jonatan-ivanov it's not a real breaking change:
We replace methods that take a ProceedingJoinPoint parameter by the very same method that takes a JoinPoint parameter
and ProceedingJoinPoint inherits from JoinPoint so that shouldn't change calls unless it in involved in reflection.

See replacement of
`public void addAnnotatedParameters(T objectToModify, ProceedingJoinPoint pjp) {`
by
`public void addAnnotatedParameters(T objectToModify, JoinPoint jp) {`

jonatan-ivanov:
I still think it is (that's also why the build fails, check the japicmp output).
You can also try this to "test" if it is a breaking change:
```java
class DemoAnnotationHandler<T> extends AnnotationHandler<T> {

    public DemoAnnotationHandler(BiConsumer<KeyValue, T> keyValueConsumer, Function<Class<? extends ValueResolver>, ? extends ValueResolver> resolverProvider, Function<Class<? extends ValueExpressionResolver>, ? extends ValueExpressionResolver> expressionResolverProvider, Class<? extends Annotation> annotation, BiFunction<Annotation, Object, KeyValue> toKeyValue) {
        super(keyValueConsumer, resolverProvider, expressionResolverProvider, annotation, toKeyValue);
    }

    @Override
    public void addAnnotatedParameters(T objectToModify, ProceedingJoinPoint pjp) {
        super.addAnnotatedParameters(objectToModify, pjp);
    }
}
```

I think there is a way to do this in a non-breaking way but first, would you share why this is needed?

jonatan-ivanov:
Just to reiterate on this: in order to consider this for merging, could you please share why this is needed?

doumdoum:
I need to write a custom meter annotation that is used only when the annotated method fails.
For that i want to write an afterThrowing aspect and therefore i don't want to wrap the method call.
But that also applies to the `@Counter` annotation where you don't need to  wrap the method call and you could just use an after advice.

jonatan-ivanov:
>I need to write a custom meter annotation that is used only when the annotated method fails.

So you want `@Counted`/`@Timed` but only when they has an exception?
If so, can you use `@Counted`/`@Timed` and drop when there is no exception?
```java
registry.config().meterFilter(MeterFilter.deny(startsWithPrefixAndHasNoException("business.logic")));

Predicate<Meter.Id> startsWithPrefixAndHasNoException(String prefix) {
    return id -> {
      if (id.getName().startsWith(prefix)) {
          String exception = id.getTag("exception");
          return exception != null && "none".equals(exception);
      }
      else return false;
    };
}
```

or I think you can also do this:
```java
@Counted(value = "business.logic", extraTags = {"onlyKeepErrors", "true"})
```
With a meter filter you can drop meters that have `onlyKeepErrors=true` and no errors (`MeterFilter.deny(...)`, like above), and you can also remove the `onlyKeepErrors` tag if it bothers you (`MeterFilter.ignoreTags("onlyKeepErrors")`).

>For that i want to write an afterThrowing aspect and therefore i don't want to wrap the method call.

I think this means that you only want to use `@Counter` but not `@Timer`, right?

Are you trying to re-use `AnnotationHandler` to look up annotated parameters/result (`@MeterTag`) in your custom aspect?

PR #6510, opened 2025-07-04
Provide meter count in HighCardinalityTagsDetector

izeye:
This PR changes to provide meter count in the `HighCardinalityTagsDetector`.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
@@ -16,7 +16,7 @@
 package io.micrometer.core.instrument;
 
 import java.time.Duration;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.Executors;
@@ -65,11 +65,13 @@ public class HighCardinalityTagsDetector implements AutoCloseable {
 
     private final long threshold;
 
-    private final Consumer<String> meterNameConsumer;
+    private final Duration delay;
+
+    @Nullable private final Consumer<String> meterNameConsumer;
+
+    @Nullable private Consumer<HighCardinalityMeterInfo> meterInfoConsumer;
 
     private final ScheduledExecutorService scheduledExecutorService;
-
-    private final Duration delay;
 
     /**
      * @param registry The registry to use to check the Meters in it
@@ -105,7 +107,7 @@ public class HighCardinalityTagsDetector implements AutoCloseable {
         this.registry = registry;
         this.threshold = threshold;
         this.delay = delay;
-        this.meterNameConsumer = meterNameConsumer != null ? meterNameConsumer : this::logWarning;
+        this.meterNameConsumer = meterNameConsumer;
         this.scheduledExecutorService = Executors
             .newSingleThreadScheduledExecutor(new NamedThreadFactory("high-cardinality-tags-detector"));
     }
@@ -135,7 +137,19 @@ public class HighCardinalityTagsDetector implements AutoCloseable {
 
     private void detectHighCardinalityTags() {
         try {
-            findFirst().ifPresent(this.meterNameConsumer);
+            findFirst().ifPresent((meterInfo) -> {
+                if (this.meterNameConsumer != null) {
+                    this.meterNameConsumer.accept(meterInfo.getName());
+                }
+                else {
+                    if (this.meterInfoConsumer != null) {
+                        this.meterInfoConsumer.accept(meterInfo);
+                    }
+                    else {
+                        logWarning(meterInfo);
+                    }
+                }
+            });
         }
         catch (Exception exception) {
             LOGGER.warn("Something went wrong during high cardinality tag detection", exception);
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -65,11 +65,13 @@ public class HighCardinalityTagsDetector implements AutoCloseable {
...
     private final long threshold;
 
-    private final Consumer<String> meterNameConsumer;
+    private final Duration delay;
 
-    private final ScheduledExecutorService scheduledExecutorService;
+    @Nullable private final Consumer<String> meterNameConsumer;
 
-    private final Duration delay;
+    @Nullable private Consumer<HighCardinalityMeterInfo> meterInfoConsumer;
```
shakuzen: We unfortunately don't have any automated formatting available for this (let us know if you're aware of some), but we'd like to have the nullability annotation next to the type being marked. In this case, that would mean `private @Nullable Consumer`.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -135,44 +137,51 @@ public void close() {
...
-     * @return the name of the first Meter that potentially has high cardinality tags, an
-     * empty Optional if none found.
+     * Finds the {@code HighCardinalityMeterInfo} of the first Meter that potentially has
+     * high cardinality tags.
+     * @return the {@code HighCardinalityMeterInfo} of the first Meter that potentially
+     * has high cardinality tags, an empty Optional if none found.
      */
-    public Optional<String> findFirst() {
-        Map<String, Long> meterNameFrequencies = new HashMap<>();
+    public Optional<HighCardinalityMeterInfo> findFirst() {
```
shakuzen: I suppose japicmp doesn't complain because of the erased type. But effectively, I supposed this is a breaking change for any users who expected a string type when calling `get` on the returned Optional. This is unfortunate because I would like to change things as has been done here, but I'm not sure what to do about the backward incompatibility.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -135,44 +137,51 @@ public void close() {
 
     private void detectHighCardinalityTags() {
         try {
-            findFirst().ifPresent(this.meterNameConsumer);
+            findFirst().ifPresent((meterInfo) -> {
+                if (this.meterNameConsumer != null) {
+                    this.meterNameConsumer.accept(meterInfo.getName());
```
shakuzen: Rather than keep two consumers, could we just keep one since it is an internal detail. For backward compatibility, if a user is configuring a `Consumer<String>` we can convert that to a `Consumer<HighCardinalityMeterInfo>` and then always run that. I would probably even go so far as to mark the `Consumer<String>` API deprecated.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
```diff
@@ -183,4 +192,120 @@ private static long calculateThreshold() {
...
+        }
+
+        /**
+         * Build {@code HighCardinalityTagsDetector}.
+         * @return {@code HighCardinalityTagsDetector}
+         */
+        public HighCardinalityTagsDetector build() {
+            HighCardinalityTagsDetector highCardinalityTagsDetector = new HighCardinalityTagsDetector(this.registry,
+                    this.threshold, this.delay, this.highCardinalityMeterNameConsumer);
+            highCardinalityTagsDetector.meterInfoConsumer = this.highCardinalityMeterInfoConsumer;
```
shakuzen: It's probably worth adding a constructor that takes the `Consumer<HighCardinalityMeterInfo>` and deprecating the one that takes `Consumer<String>`. Then the consumer field can be marked final. Was there any reason in particular to set it this way instead of a constructor?

shakuzen (commented):
Thanks for the pull request. This looks like a useful enhancement. I've left some review.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
izeye: @shakuzen I wanted to do so, but I couldn't do so as their signatures are the same after type erasure.

izeye:
@shakuzen Thanks for the feedbacks!

I tried to apply your suggestions in a27b799.

Please let me know what you think.

micrometer-core/src/main/java/io/micrometer/core/instrument/HighCardinalityTagsDetector.java
shakuzen: What about introducing a new type like `HighCardinalityMeterInfoConsumer`?

izeye: @shakuzen It would break existing lambda expressions from users and require casting to use lambda expressions for both.

shakuzen: You're right. I'll think some more if there's anything else we can do, but we can merge for now since this is mostly implementation details we can change later.

Issue #6402, opened 2025-06-13
Re-enable OSGi tests for Java 24


jonatan-ivanov:
Right now we need to skip the micrometer-osgi-test module when building with jdk >=24 because of https://github.com/bndtools/bnd/pull/6371. Once the fix is released we can upgrade the plugin and re-enable the module for jdk >=24.
See [this change](https://github.com/micrometer-metrics/micrometer/pull/6400/files#diff-9c6341b5a61a7eef0275b796dfe322b19c026ab01a77d5e76feafecbaa0ab74aR5-R10).

shakuzen:
Since we need to upgrade the BND plugin for #6829, this issue will be resolved by that.

shakuzen:
Resolved by #6829

PR #6512, opened 2025-07-07
Support slf4j2 in OSGi

### LINKED ISSUE #6406: Incompatible with slf4j 2 in OSGi
**Describe the bug**
I am trying to deploy micrometer-commons and micrometer-observations in an OSGi environment that includes slf4j 2. Micrometer is compatible with slf4j 2 (#3415) but the OSGi Import-Package version range on the `org.slf4j.spi` package in micrometer-commons is not.

**Environment**

 - Micrometer version: 1.15.1
 - Java version: 17

**To Reproduce**
How to reproduce the bug:

Deploy micrometer-commons in an OSGi container along with slf4j 2. The micrometer bundle cannot be started:

```
g! ss micro

id	State       Bundle
330	INSTALLED   micrometer-commons_1.15.1

g! diag 330
micrometer-commons [330]
  Unresolved requirement: Import-Package: javax.annotation; version="3.0.2"; resolution:="optional"
  Unresolved requirement: Import-Package: javax.annotation.meta; version="3.0.2"; resolution:="optional"
  Unresolved requirement: Import-Package: org.slf4j.spi; version="[1.7.0,2.0.0)"

g! packages org.slf4j.spi
osgi.wiring.package; bundle-symbolic-name="slf4j.api"; bundle-version:Version="2.0.17"; version:Version="2.0.17"; osgi.wiring.package="org.slf4j.spi"; uses:="org.slf4j,org.slf4j.event,org.slf4j.helpers"<slf4j.api_2.0.17 [309]>
  jcl.over.slf4j_2.0.17 [276] imports
  io.netty.common_4.2.1.Final [158] imports
  org.jboss.logging.jboss-logging_3.1.1.GA [308] imports
  jul.to.slf4j_2.0.17 [236] imports
  org.apache.logging.log4j.slf4j2.impl_2.24.3 [266] imports
g! b 309
slf4j.api_2.0.17 [309]
  Id=309, Status=ACTIVE      Data Root=/var/tmp/org.eclipse.osgi/309/data
  "No registered services."
  Services in use:
    {org.apache.logging.log4j.spi.Provider}={service.id=29, service.bundleid=198, service.scope=singleton, APIVersion=2.6.0}
  Exported packages
    org.slf4j; version="2.0.17"[exported]
    org.slf4j.event; version="2.0.17"[exported]
    org.slf4j.helpers; version="2.0.17"[exported]
    org.slf4j.spi; version="2.0.17"[exported]
    org.slf4j; version="1.7.36"[exported]
    org.slf4j.helpers; version="1.7.36"[exported]
  Imported packages
    org.apache.aries.spifly; version="1.3.7" <org.apache.aries.spifly.dynamic.bundle_1.3.7 [224]>
  No fragment bundles
  No required bundles
```

Here slf4j 2 exports `1.7.36` versioned packages for all but the `org.slf4j.spi` package, which causes Micrometer to fail to resolve because of its  required version range `[1.7,2.0)`.

**Expected behavior**
That Micrometer could still resolve and work with version 2 of the `org.sfl4j.spi` package. This could be achieved by omitting the upper bounds on the `org.slf4j.spi` package import version range, or increasing it to `3`, or omitting the version range completely. 

**Additional context**
For my use case, I was able to get Micrometer deployed by changing the `org.slf4j` imports to be dynamic, as in [this commit](https://github.com/SolarNetwork/micrometer/commit/4a5ac2d95a976a8178d273561206e35a21c2bf87).

msqr:
Address #6406 by allowing slf4j2 to be used at runtime in OSGi. Adds a new OSGi + slf4j2 sub-project for testing purposes.

Fixes #6406


msqr:
Some tests were failing for me in the 1.14.x branch, but before applying this PR. Additionally the `micrometer-osgi-test-slf4j2` tests seem to take a long time to complete, whereas these [same changes applied to the `main` branch](https://github.com/micrometer-metrics/micrometer/compare/main...SolarNetwork:micrometer:feature/slf4j-osgi) do not.

gradle/libs.versions.toml
```diff
@@ -200,6 +201,8 @@ retrofit2 = { module = "com.squareup.retrofit2:retrofit", version = "2.11.0" }
 signalfx = { module = "com.signalfx.public:signalfx-java", version.ref = "signalfx" }
 slf4jApi = { module = "org.slf4j:slf4j-api", version.ref = "slf4j" }
 slfj4Simple = { module = "org.slf4j:slf4j-simple", version.ref = "slf4j" }
+slf4j2Api = { module = "org.slf4j:slf4j-api", version.ref = "slf4j2" }
+slfj42Simple = { module = "org.slf4j:slf4j-simple", version.ref = "slf4j2" }
```
shakuzen: Dependabot struggles to handle upgrading of dependencies in a Gradle version catalog when there are two different versions of a same dependency declared. For that reason, it is probably better to hard-code the version in the osgi-test-slf4j2 module. That should be sufficient for our purposes of testing osgi with slf4j 2.x

shakuzen:
> the `micrometer-osgi-test-slf4j2` tests seem to take a long time to complete, whereas these [same changes applied to the `main` branch](https://github.com/micrometer-metrics/micrometer/compare/main...SolarNetwork:micrometer:feature/slf4j-osgi) do not.

My guess would be a difference in the bnd plugin version that is used between the branches. We could consider backporting the bnd plugin version upgrade if it does not cause any other problems to do so. I don't think the latest version of the plugin works with JDK versions less than 17, though.

gradle/libs.versions.toml
msqr: Sure, I pushed up a new commit that does this now.

msqr:
> > the `micrometer-osgi-test-slf4j2` tests seem to take a long time to complete, whereas these [same changes applied to the `main` branch](https://github.com/micrometer-metrics/micrometer/compare/main...SolarNetwork:micrometer:feature/slf4j-osgi) do not.
> 
> My guess would be a difference in the bnd plugin version that is used between the branches. We could consider backporting the bnd plugin version upgrade if it does not cause any other problems to do so. I don't think the latest version of the plugin works with JDK versions less than 17, though.

To test it out I cherry-picked 826166a081fb847d4649a48f83651511ab3c4a2e into `feature/slf4j2-osgi`, which brought in the newer version of bnd for Java 17+, and indeed the test completes quickly (running in Java 21). I didn't push up that commit though... would you like me to?

shakuzen:
I opened https://github.com/micrometer-metrics/micrometer/issues/6515. You can rebase on `1.14.x` to get that now. Probably should also add the following to the `build.gradle` for the new osgi-test module:

```groovy
// skip this module when building with jdk >=24
// waiting for a release with the fix https://github.com/bndtools/bnd/pull/6371
if (javaLanguageVersion.canCompileOrRun(24)) {
    project.tasks.configureEach { task -> task.enabled = false }
}
```

See #6402

msqr:
Ok, rebased on 1.14.x and added the work-around as indicated.

micrometer-osgi-test-slf4j2/build.gradle
```diff
@@ -0,0 +1,70 @@
...
+    // osgi test fails when the new prometheus client is used
+    //testImplementation project(':micrometer-registry-prometheus')
+    testImplementation project(':micrometer-registry-prometheus-simpleclient')
+
+    testImplementation libs.felixFramework
+
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+    testRuntimeOnly libs.felixScr
+    testRuntimeOnly 'org.slf4j:slf4j-simple:2.0.17'
+    testRuntimeOnly 'org.apache.aries.spifly:org.apache.aries.spifly.dynamic.bundle:1.3.7'
```
shakuzen, line 29: This looks new compared to the existing osgi test module. What is this for?

msqr: The `org.apache.aries.spifly:org.apache.aries.spifly.dynamic.bundle` I presume you are referring to? That is required by slf4j2, as it uses the [non-OSGi ServiceLoader](https://blog.osgi.org/2013/02/javautilserviceloader-in-osgi.html), which [SPI Fly](https://aries.apache.org/documentation/modules/spi-fly.html) provides support for in OSGi, so that is required at runtime in addition to slf4j2.

SPI Fly is also referenced in the [test.bndrun](https://github.com/micrometer-metrics/micrometer/blob/52167d2602509fbf1d7ffb20af1c7f7467006e8b/micrometer-osgi-test-slf4j2/test.bndrun#L10) file.

shakuzen: Thanks for the explanation.

shakuzen (approved):

Issue #2620, opened 2021-05-26
Silently dropping negative durations gives bad developer experience


jonmcewen:
Having just lost a few hours debugging a missing metric, where the calling code mistakenly calculated a negative duration, I would suggest throwing an exception in the `else` case here: https://github.com/micrometer-metrics/micrometer/blob/67c41db552cac463e44cd45545e5a7b8f3513f1b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java#L169

jonatan-ivanov:
I think a log entry would make more sense, we have a `WarnThenDebugLogger`, do you want to make the change?
Btw, how did you manage to measure a negative amount? Did you use `System.currentTimeMillis()`?

jonmcewen:
Yes, after submitting I realised that it would be rather intrusive to throw an exception from a metrics library.  I'll have a look at the logger.

The negative value was a simple develop error: 
`duration = start - now`
hence my description as bad DX.  Functionally, dropping the negative is fine, as it is meaningless, but there was nothing to help the developer see their mistake.
Of course, if they had unit tested the metrics...

jonmcewen:
@jonatan-ivanov I can't find a `WarnThenDebugLogger`, and when I try to `./gradlew pTML` I get a 401 on hystrix-core

jonatan-ivanov:
Here's the class: [WarnThenDebugLogger](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/util/internal/logging/WarnThenDebugLogger.java)

Here's an usage example: https://github.com/micrometer-metrics/micrometer/blob/67c41db552cac463e44cd45545e5a7b8f3513f1b/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultGauge.java#L38

Based on the gitter messages, I guess your build issue is resolved, right?

PR #6530, opened 2025-07-14
Enhance logging of negative amount in AbstractTimer.record() by printing stack trace

thonra:
In PR #4573 , we added a warning log to `AbstractTimer.record()` when the amount is negative.

In production, we sometimes measure the elapsed time between two code points, and for various reasons the calculated value can occasionally be negative. When that happens, we need to trace the root cause and fix it.

Because this timing pattern is used in multiple places or projects, the current “negative amount” warning alone does not reveal where the problem originates.

Therefore, this PR appends a stack trace to the warning so we can quickly pinpoint the call site that triggered the invalid record() invocation.

See 
- #4573
- #2620

#### Verification with unit test:
image

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
@@ -15,6 +15,8 @@
  */
 package io.micrometer.core.instrument;
 
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
 import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
 import io.micrometer.core.instrument.distribution.*;
 import io.micrometer.core.instrument.distribution.pause.ClockDriftPauseDetector;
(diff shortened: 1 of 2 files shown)


shakuzen:
> we sometimes measure the elapsed time between two code points, and for various reasons the calculated value can occasionally be negative.

Note that best practice to avoid this is to use the methods to measure a block of code as [documented here](https://docs.micrometer.io/micrometer/reference/concepts/timers.html#_recording_blocks_of_code) or use the Timer.Sample API as [documented here](https://docs.micrometer.io/micrometer/reference/concepts/timers.html#_storing_start_state_in_timer_sample) so that you do not need to do the time measurement yourself and such problems are avoided. The Timer API accepting an elapsed time is meant as a last resort for when the other methods of recording a Timer measurement cannot be used.

micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
```diff
@@ -264,7 +264,7 @@ public final void record(long amount, TimeUnit unit) {
             }
         }
         else {
-            log.log(() -> "'amount' should not be negative but was: " + amount);
+            log.log(() -> "'amount' should not be negative but was: " + amount, new IllegalArgumentException());
```
shakuzen: Instantiating an Exception is not exactly cheap. The logger is a `WarnThenDebugLogger` so unless debug logging is enabled, nothing will be logged after the first time, but the cost of instantiating the exception will be paid nonetheless. We should try to avoid that, but I'm not immediately sure what the best way to do that will be.

thonra: Thank you for your valuable insights.

A possible improvement is to log only the first non-micrometer stack frame using `StackWalker`; however, since `StackWalker` is available only in Java 9+, this approach is incompatible with projects that still target JDK 8. I haven’t been able to come up with a reasonable alternative yet.

```
    String caller = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE)
        .walk(frames -> frames
            .dropWhile(f -> f.getClassName().startsWith("io.micrometer"))
            .findFirst())
        .map(Object::toString)
        .orElse("N/A");

    log.log(() -> String.format(
        "'amount' should not be negative but was %d, caller=%s",
        amount, caller));
```

Issue #4259, opened 2023-10-20
Support generational Shenandoah GC


shakuzen:
Similar to #4258, there is work to provide a generational option for Shenandoah GC in [JEP 404](https://openjdk.org/jeps/404). This was at a point scheduled for Java 21 but was subsequently pulled from the release. We should keep an eye on it and add support when the collector names and memory pool names are stabilized.

shakuzen:
[JEP 404](https://openjdk.org/jeps/404) is included in JDK 24 as an experimental feature. We could offer preliminary support for it.

shakuzen:
With [JEP 521](https://openjdk.org/jeps/521), Generational Shenandoah has been integrated into Java 25 as a non-experimental feature.

shakuzen:
MemoryPool names (vs the single pool named `Shenandoah` in non-generational mode):
* `Shenandoah Young Gen`
* `Shenandoah Old Gen`

GarbageCollectorMXBean names (same as non-generational):
* `Shenandoah Pauses`
* `Shenandoah Cycles`

PR #6545, opened 2025-07-17
Support Generational Shenandoah GC metrics

shakuzen:
Resolves gh-4259

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
@@ -46,8 +46,8 @@ class JvmMemory {
 
     static boolean isAllocationPool(@Nullable String name) {
         return name != null && (name.endsWith("Eden Space") //
-                || "Shenandoah".equals(name) // non-generational ZGC
-                || "Shenandoah Young Gen".equals(name) // generational ZGC
+                || "Shenandoah".equals(name) // non-generational Shenandoah
+                || "Shenandoah Young Gen".equals(name) // generational Shenandoah
                 || "ZHeap".equals(name) // ZGC non-generational
                 || "ZGC Young Generation".equals(name) // generational ZGC
                 || name.endsWith("New Gen") // Zing GPGC


micrometer-core/build.gradle
```diff
@@ -239,6 +239,17 @@ tasks.register("shenandoahTest", Test) {
...
 
+tasks.register("generationalShenandoahTest", Test) {
+    // set heap size for the test JVM(s)
+    maxHeapSize = "1500m"
+
+    useJUnitPlatform {
+        includeTags 'gc'
+    }
+
+    jvmArgs '-XX:+UseShenandoahGC', '-XX:+UnlockExperimentalVMOptions', '-XX:ShenandoahGCMode=generational'
```
shakuzen, line 250: We can remove `-XX:+UnlockExperimentalVMOptions` and run this test task only on our Java 25+ builds when we have any. In the meantime, we can run it on Java 24 with this option.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
```diff
@@ -45,7 +45,9 @@ static boolean isConcurrentPhase(String cause, String name) {
     }
 
     static boolean isAllocationPool(@Nullable String name) {
-        return name != null && (name.endsWith("Eden Space") || "Shenandoah".equals(name) //
+        return name != null && (name.endsWith("Eden Space") //
+                || "Shenandoah".equals(name) // non-generational ZGC
+                || "Shenandoah Young Gen".equals(name) // generational ZGC
```
jonatan-ivanov:
```suggestion
        return name != null && (name.endsWith("Eden Space")
                || "Shenandoah".equals(name) // non-generational Shenandoah
                || "Shenandoah Young Gen".equals(name) // generational Shenandoah
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
```diff
@@ -55,7 +57,9 @@ static boolean isAllocationPool(@Nullable String name) {
     }
 
     static boolean isLongLivedPool(@Nullable String name) {
-        return name != null && (name.endsWith("Old Gen") || name.endsWith("Tenured Gen") || "Shenandoah".equals(name)
+        return name != null && (name.endsWith("Old Gen") || name.endsWith("Tenured Gen") //
```
jonatan-ivanov, line 60:
```suggestion
        return name != null && (name.endsWith("Old Gen") || name.endsWith("Tenured Gen")
```

jonatan-ivanov (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
shakuzen: The blank comment is there to prevent the formatter from causing weird formatting. It will move the Shenandoah condition up to this line and then break the comment up over multiple lines.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemory.java
shakuzen: Same as the other one: the blank comment is needed to prevent the formatter from resulting in bad formatting. I'll fix the wrong name in the comments. Thanks for catching that.

Issue #6416, opened 2025-06-17
State objects should not be nullable for function-based meter registration


shakuzen:
In https://github.com/micrometer-metrics/micrometer/pull/6385#discussion_r2140474625 it came up that there are some inconsistencies in how we mark nullability of at least the backing object for function-based meter registration - Gauge, TimeGauge, FunctionCounter, FunctionTimer. We should holistically review this and decide what changes make the most sense.

ragulshanmugam:
Hi,

I reviewed the current code on `main` branch to see where things stand across all function-based meters. The inconsistency is still present, and it's limited to the annotation/contract mismatch - the implementations already behave identically.

In `MeterRegistry`:

```
protected abstract <T> Gauge newGauge(Meter.Id id, @Nullable T obj, ToDoubleFunction<T> valueFunction);
protected abstract <T> TimeGauge newTimeGauge(Meter.Id id, @Nullable T obj, TimeUnit timeFunctionUnit, ToDoubleFunction<T> timeFunction);
protected abstract <T> FunctionTimer newFunctionTimer(Meter.Id id, T obj, ToLongFunction<T> countFunction, ...);
protected abstract <T> FunctionCounter newFunctionCounter(Meter.Id id, T obj, ToDoubleFunction<T> countFunction);
```

Also, same split in the builders: Gauge/TimeGauge take @Nullable T obj, FunctionCounter/FunctionTimer take a bare T obj. But DefaultGauge and CumulativeFunctionCounter both call new WeakReference<>(obj) and both already handle ref.get() returning null, since the object can be collected at any point regardless.

Did a quick test change to confirm this. Adding this to FunctionCounterTest fails to compile:

```
FunctionCounterTest.java:35: error: [NullAway] passing @Nullable parameter
    '(AtomicLong) null' where @NonNull is required
```

while the equivalent call on Gauge compiles fine. MeterRegistryCompatibilityKit also has garbageCollectedSourceObject(), which passes a literal null to a gauge as supported behaviour and there's no equivalent test for FunctionCounter or FunctionTimer.

I'm thinking of two ways to make it consistent:

1. Add @Nullable to the function-based meters. It is backward compatible, matches what the implementations already do.
2. Remove @Nullable from the gauges. It is arguably more correct, since a gauge on a null object silently reports NaN forever, but breaking for anyone passing null today.

[...]

PR #6546, opened 2025-07-17
Improve nullability for gauges

izeye:
This PR changes to use `@Contract` to improve nullability for gauges.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -15,6 +15,7 @@
  */
 package io.micrometer.core.instrument;
 
+import io.micrometer.common.lang.internal.Contract;
 import io.micrometer.common.util.internal.logging.InternalLogger;
 import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
 import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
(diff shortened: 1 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -537,6 +539,7 @@ public More more() {
      * @return The number that was passed in so the registration can be done as part of an
      * assignment statement.
      */
+    @Contract("_, _, !null -> !null")
```
izeye: There are some inconsistencies including this. If the state object (`number`) being non-null is intentional, dropping `@Nullable` seems to be better.

micrometer-core/src/main/java/io/micrometer/core/instrument/Metrics.java
```diff
@@ -145,6 +146,7 @@ public static More more() {
      * @return The number that was passed in so the registration can be done as part of an
      * assignment statement.
      */
+    @Contract("_, _, !null, _ -> !null")
     public static <T> @Nullable T gauge(String name, Iterable<Tag> tags, T obj, ToDoubleFunction<T> valueFunction) {
```
izeye: There is another inconsistency: the state object in the corresponding method in the `MeterRegistry` is nullable while this state object (`obj`) is not nullable.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
izeye: I just dropped `@Nullable` annotations for these cases instead.

jonatan-ivanov (approved):

jonatan-ivanov:
I think this is definitely an improvement over what we have now but we might need to revert parts of this:
- As you mentioned there are inconsistencies, we should review and fix them
- Maybe we should relax contracts and make states `@Nullable` (for `Number` and `Collection` too)?
- Other function-based meters have the same problem, we should make them consistent, see https://github.com/micrometer-metrics/micrometer/issues/6416 and https://github.com/micrometer-metrics/micrometer/pull/6385#discussion_r2140474625

/cc @shakuzen

PR #6554, opened 2025-07-20
Polish gh-5997

izeye:
This PR polishes the changes made in gh-5997 a bit.

### DIFF
--- a/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/mail/DefaultMailSendObservationConvention.java
+++ b/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/mail/DefaultMailSendObservationConvention.java
@@ -40,6 +40,36 @@ public class DefaultMailSendObservationConvention implements MailSendObservation
 
     private static final String UNKNOWN = "unknown";
 
+    private static final KeyValue NETWORK_PROTOCOL_NAME_UNKNOWN = NETWORK_PROTOCOL_NAME.withValue(UNKNOWN);
+
+    private static final KeyValue SERVER_PORT_UNKNOWN = SERVER_PORT.withValue(UNKNOWN);
+
+    private static final KeyValue SERVER_ADDRESS_UNKNOWN = SERVER_ADDRESS.withValue(UNKNOWN);
+
+    private static final KeyValue SMTP_MESSAGE_SUBJECT_UNKNOWN = SMTP_MESSAGE_SUBJECT.withValue(UNKNOWN);
+
+    private static final KeyValue SMTP_MESSAGE_FROM_UNKNOWN = SMTP_MESSAGE_FROM.withValue(UNKNOWN);
+
+    private static final KeyValue SMTP_MESSAGE_ID_UNKNOWN = SMTP_MESSAGE_ID.withValue(UNKNOWN);
+
+    private static final Map<RecipientType, KeyName> RECIPIENT_TYPE_KEY_NAME_MAP;
+    static {
+        Map<RecipientType, KeyName> map = new IdentityHashMap<>();
+        map.put(RecipientType.TO, SMTP_MESSAGE_TO);
+        map.put(RecipientType.CC, SMTP_MESSAGE_CC);
+        map.put(RecipientType.BCC, SMTP_MESSAGE_BCC);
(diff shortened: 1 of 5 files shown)


micrometer-jakarta9/src/test/java/io/micrometer/jakarta9/instrument/mail/DefaultMailSendObservationConventionTests.java
```diff
@@ -80,34 +80,20 @@ void shouldHaveLowCardinalityKeyValues() throws MessagingException {
         message.setSubject("Test Subject");
 
         MailSendObservationContext context = new MailSendObservationContext(message, "smtp", "example.com", 587);
-        assertThat(convention.getLowCardinalityKeyValues(context)).contains(
-                KeyValue.of("server.address", "example.com"), KeyValue.of("server.port", "587"),
-                KeyValue.of("network.protocol.name", "smtp"));
-    }
-
-    @Test
-    void shouldHandleMessagingException() {
```
izeye, line 89: I dropped this as it seems to be a duplicate of the `subjectShouldBeUnknownWhenFetchingFails()`.

micrometer-jakarta9/src/test/java/io/micrometer/jakarta9/instrument/mail/DefaultMailSendObservationConventionTests.java
```diff
@@ -221,26 +207,19 @@ void subjectShouldBeThereWhenSet() throws MessagingException {
     }
 
     @Test
-    void protocolAddressAndPortShouldBeThereWhenSet() {
```
izeye, line 224: I dropped this as it seems to be a duplicate of the `shouldHaveLowCardinalityKeyValues()`.

micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/mail/DefaultMailSendObservationConvention.java
```diff
@@ -58,68 +88,62 @@ public KeyValues getLowCardinalityKeyValues(MailSendObservationContext context)
...
+        return safeExtractValue(SMTP_MESSAGE_FROM, () -> addressesToValue(message.getFrom()),
+                SMTP_MESSAGE_FROM_UNKNOWN);
     }
 
-    private Optional<KeyValue> smtpMessageRecipients(@Nullable Message message, RecipientType recipientType) {
-        if (message == null) {
-            return Optional.empty();
-        }
-        MailObservationDocumentation.HighCardinalityKeyNames key = MailObservationDocumentation.HighCardinalityKeyNames
-            .valueOf("SMTP_MESSAGE_" + recipientType.toString().toUpperCase(Locale.ROOT));
```
izeye, line 114: I'm not sure if the `MimeMessage.RecipientType.NEWSGROUPS` or a custom one should be handled here.

If they need to be considered, the previous code had a problem, and the polished one also has a problem.

We can ignore them defensively or handle them properly if necessary.

jonatan-ivanov: I think we can and should add `MimeMessage.RecipientType.NEWSGROUPS`; it seems like a straightforward change. Thanks for catching it! Do you want to do it in a separate PR?

As for custom ones, I think it would be nice to support every possible values but I'm not sure there is an easy way to do it. It seems `Message` implementations do not store the type and the address(es) together and they hardcode the possible values instead like our implementation does (see `getAllRecipients`). Please feel free to look into this if you want to, I would be happy to add this capability if there is a not so complicated way to do it but otherwise I think we can just ignore them and let the users override `getHighCardinalityKeyValues` if they want to add their custom `RecipientType`s.

izeye: @jonatan-ivanov Thanks for the feedback!

I created https://github.com/micrometer-metrics/micrometer/pull/6558 to try to resolve it.

> If they need to be considered, the previous code had a problem, and the polished one also has a problem.
> 
> We can ignore them defensively or handle them properly if necessary.

Sorry, I misread the code. They are ignored already, so it's not a problem, just not supported.

PR #6563, opened 2025-07-23
Handle ArrayIndexOutOfBoundsException from DoubleHistogram in TimeWindowPercentileHistogram.accumulate() defensively

izeye:
This PR changes to handle `ArrayIndexOutOfBoundsException` from `DoubleHistogram` in `TimeWindowPercentileHistogram.accumulate()` defensively as it could happen due to `DoubleHistogram`'s limitation.

For example, if it happens with Prometheus, its scape would fail.

See https://github.com/HdrHistogram/HdrHistogram/issues/190


micrometer-core/src/test/java/io/micrometer/core/instrument/DistributionSummaryTest.java
```diff
@@ -68,4 +69,17 @@ public CountingMode mode() {
         assertThat(summary.takeSnapshot().histogramCounts()).containsExactly(new CountAtBucket(1.0, 0));
     }
 
+    @Test
+    void takeSnapshotShouldNotThrowExceptionEvenIfDoubleHistogramThrowsException() {
```
jonatan-ivanov: Should we use `CapturedOutput` here and check if the log output is in place?

jonatan-ivanov:
Thanks for catching and fixing this!
I'm thinking if we should qualify this as a bug and do it in `1.14.x`?

izeye:

I applied the `CapturedOutput` and changed the base branch to the 1.13.x which seems to be a preferred one for a bug fix. Please let me know if the 1.14.x branch is the right one for this.

jonatan-ivanov (approved):

jonatan-ivanov:
`1.13.x` is ok at the moment, even if it is out of OSS support, we haven't closed it yet. :)
Thank you!

PR #6568, opened 2025-07-25
Require explicit constraints in @Contract

izeye:
This PR simply applies https://github.com/spring-projects/spring-framework/issues/35157 to Micrometer's one.


jonatan-ivanov (approved):

jonatan-ivanov:
Thank you!

izeye:
@jonatan-ivanov Thanks! By the way, it's for internal use only, so from the user's perspective, the "type: task" label seems more appropriate.

PR #6571, opened 2025-07-27
Add SmartKeyName API for consistent KeyValues with fallback support

### LINKED ISSUE #6564: Add API support for providing fallback for KeyValues
Applying `KeyValue`s conditionally can lead into issues for example Prometheus require that all time-series (observation) with the same name have the same set of labels (low cardinality key names).

For example this can cause an error on if both branches (`color = null` and `color = "red"`) are executed.
```java
if (color != null) {
    observation.lowCardinalityKeyValue("color", color);
}
```
Instead of it, users should do something like this:
```java
observation.lowCardinalityKeyValue("color", color != null ? color : "unknown");
```

We should make this easier for the users and harder to make this mistake.

See:
- https://github.com/spring-projects/spring-data-mongodb/pull/5020#discussion_r2212509260
- https://gist.github.com/mp911de/6729ca327326d072820a1a0eb225c4a6
- https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:key-values-with-fallback

Validation could be also added: https://github.com/micrometer-metrics/micrometer/pull/6713

yybmion:
### Summary

Adds `SmartKeyName` API to provide automatic fallback handling for KeyValues, ensuring consistent key sets across observations as required by Prometheus.

### Changes 
- `SmartKeyName<C>` class with `required()`, `optional()`, and `withFallback()` factory methods
- `KeyValues.of()` and `KeyValues.and()` methods to support SmartKeyName integration
- Comprehensive test coverage for all SmartKeyName scenarios

Closes #6564


yybmion:
Hi @jonatan-ivanov , Just a gentle reminder — PTAL when you have time.

jonatan-ivanov:
Hi,

Thank you for the PR!

I guess this is more or less the copy of this gist: https://gist.github.com/mp911de/6729ca327326d072820a1a0eb225c4a6
We had some discussions around this internally and decided that most probably we would like to follow a simpler approach, see the first two methods added to `KeyName` here: https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:key-values-with-fallback
And if some more advanced use-cases are needed check the other two methods and the `KeyValueProvider` interface and the two demo classes for usage examples.

If this is something you would use in production, please feel free to give feedback on the original issue.

Issue #6313, opened 2025-05-28
Endless loop in SimpleObservation.getLastScope


tgrzej-avaloq:
**Describe the bug**
Our client hits an endless loop in
        at io.micrometer.observation.SimpleObservation$SimpleScope.getLastScope(SimpleObservation.java:327)
which is [here](https://github.com/micrometer-metrics/micrometer/blob/e7b81e2a8a6ae7f6a0779220a9a76eb7bd542e78/micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservation.java#L327)

Currently this only happens in the production environment of our client and we cannot reproduce it anywhere else (yet).
A related stacktrace showing the context of the call (from the Hibernate saveAll() method) is attached.

**Environment**
 - Micrometer version: 1.12.11
 - Hibernate version: 6.4.10.Final
 - Spring Boot version: 3.2.11
 - Micrometer registry: prometheus
 - OS: Linux
 - Java version: 17

**To Reproduce**
The only "reproducer" we have is in the client's production environment.

**Expected behavior**
I can only suspect that because of a memory visibility issue in a concurrent environment the thread that executes the while loop does not ever see the end of the scope list. One would expect that the scope list does have a proper end and the while loop should stop there.

**Additional context**
The related stacktrace from a thread dump showing the context of the call which is the Hibernate saveAll() method.
As can be seen the DataSource is wrapped into a net.ttddyy proxy.
[20250430-stacktrace-SimpleObservation-327-full.txt](https://github.com/user-attachments/files/20518715/20250430-stacktrace-SimpleObservation-327-full.txt)

tgrzej-avaloq:
Hi @ttddyy,
Thank you for looking at this ticket and for your response.

> does the app use the `Open In View`(`Open Session in View`) pattern?

I don't think it does. I have now attached a full stacktrace that I have available (I wanted to ask first if I am allowed). I am not an expert, but I believe that the Session gets opened together with a DB transaction through a method annotated with `@Transactional`. Otherwise in this component (nor in the use case) there isn't any view layer. To provide some context: the data is read from a source DB and written to a target DB (with a rather peculiar multithreading setup).

> I see `DeferredResultSetAccess` in the stack. I don’t know what it does, but is it possible that it(or the app) is switching threads to access the `ResultSet` and could be interfering with the current observation context propagation??

[...]

Issue #6329, opened 2025-06-02
Validate `Observation.Scope` closing when using TestObservationRegistry


shakuzen:
**Please describe the feature request.**
TestObservationRegistry uses ObservationValidator to catch issues related to the order of calling APIs on Observation. It would be nice to check the closing of Scopes as well. This could include things like the order of closing nested scopes and the Thread on which Scopes are opened and closed.

**Rationale**
See #6313. 

**Additional context**
Similar feature to ObservationValidator.

seonghyeoklee:
Hi, I'd like to work on this issue. I'll add scope closing validation to `ObservationValidator` - checking LIFO ordering of nested scopes and verifying that scopes are opened and closed on the same thread. Will submit a PR soon.

jonatan-ivanov:
Thank you for the offer, I assigned the issue to you.

PR #6575, opened 2025-07-29
#6313 handle out-of-order scope closure

tgrzej-avaloq:
As the issue mentions: there is more that one DB connection handled by a particular thread and the connections are not closed in the expected order, causing residual scopes to be left in the `SimpleObservationRegistry` (this is a memory leak plus it leads to CPU saturation with time). In the issue it is because of the specific Spring transaction handling workflow coupled with the fact that one connection is handled through the `SimpleJdbc` and the other one through the `EntityManager`, however otherwise it should be allowed for any code to close DB connections according to its specific requirements.

This PR does not fix the underlying problem which I think is handling disjoint scope hierarchies. It just attempts to prevent the long term corruption of the scope list in the `SimpleObservationRegistry`.


shakuzen:
@tgrzej-avaloq I see the PR is still in draft state. Is there some more work you're planning to do on this? Could you add a test that demonstrates the issue being fixed?

tgrzej-avaloq:
Hi @shakuzen,
First, thank you for having a look at my PR. I cannot see our Slack conversation anymore (it only allows me to go 90 days back), but I recall that after you had helped me to properly submit this PR you mentioned you would need a more thorough look at it. I believe you yourself know best what you meant by that and what were your plans and outcomes in that regard.

When it comes to the PR state being draft, it is possibly my general lack of experience in collaborating on github, I don't quite know what that practically means and if something was/is further expected from me (for example, to move the PR to some other state at some point). The PR itself is in a "works for our use case" state and me and the product team we moved on, and no plans were made to further work on it (please also see the last statement in the next paragraph).

Regarding the PR's content it fixed the scope list corruption issue in our use case. The use case is complex and not well-designed, but possibly a valid one: one thread opening and closing two DB connections in an order that is not as expected by the micrometer scope accounting rules. Having said that, the team owning the product chose to completely disable the affected DataSource proxy so the product is not actually running with that patch.

Now, when it comes to providing a test case... well... I don't want to say that it is impossible, but as often the facts of the corporate life are, I would need to find a booking code to put those hours on, talk with my/project's manager, get that task to the top of my priority list, refresh my coding skills and understanding in that area... that could be fun... but it is always some effort. I guess I want to say this is probably not going to happen next month.

Hope that answers your questions, to some extent at least
and regards

tgrzej-avaloq:
...it occurred to me that possibly setting up an in-memory db (or two) with a datasource wrapped in a ttddyy proxy and then opening and closing two connections (in the "wrong" order) might not be so complicated... I may try to squeeze it in, in a week or two.

tgrzej-avaloq:
@shakuzen it turned out to be possible and not so complicated to come up with a test (after the initial trial-and-error phase...).

Below is a demonstration of the root cause of https://github.com/micrometer-metrics/micrometer/issues/6313 with  datasource-micrometer-1.1.2.
It turns out it is improved in datasource-micrometer-1.2.0 by https://github.com/jdbc-observations/datasource-micrometer/issues/82.
```
public class OutOfOrderConnectionClosure {

    public static void main(String[] args) throws SQLException {
        final DataSource proxiedDs = ProxyDataSourceBuilder.create(createDs())
                .methodListener(new DataSourceObservationListener(ObservationRegistry.create()))
                .buildProxy();

        final Connection conn1 = proxiedDs.getConnection();
        final Connection conn2 = proxiedDs.getConnection();
        /* do some work with the connections */

        // With datasource-micrometer-1.1.2 this closing order leaks the SimpleScope related to the "conn1" in
        // the SimpleObservationRegistry. That order is not always up to the code author to decide when
        // working with JDBC/ORM frameworks.
        // In the newer datasource-micrometer the connection related scope is closed already when acquiring a
        // connection and therefore the connection closing order is no longer relevant.
        conn1.close();
        conn2.close();
    }

    public static DataSource createDs() {
        SQLiteDataSource ds = new SQLiteDataSource();
        ds.setUrl("jdbc:sqlite:file:myDb?mode=memory&cache=shared");
        return ds;
    }
}
```
I used the following libraries to run this code
* datasource-micrometer-1.1.2.jar
* datasource-proxy-1.10.1.jar
* micrometer-commons-1.15.6.jar
* micrometer-core-1.15.6.jar
* micrometer-observation-1.15.6.jar
* sqlite-jdbc-3.51.2.0.jar

I believe that due to the awesome work from @ttddyy and his improvements in datasource-micrometer my naive patch no longer serves any purpose. Let me state that I am in awe of the quality of work in the open source world at times, thank you!

tgrzej-avaloq:
I guess I can close this PR as the related issue is fixed in datasource-micrometer-1.2.0.jar by https://github.com/jdbc-observations/datasource-micrometer/issues/82.

shakuzen:
Thanks for the additional work on this. I further distilled your test to not depend on datasource-micrometer:

```java
@Test
void outOfOrderScopeClosing() {
    Observation obs1 = Observation.start("obs1", registry);
    Observation.Scope scope1 = obs1.openScope(); // previous = null
    assertThat(registry.getCurrentObservation()).isEqualTo(obs1);
    Observation obs2 = Observation.start("obs2", registry);
    Observation.Scope scope2 = obs2.openScope(); // previous = scope1
    assertThat(registry.getCurrentObservation()).isEqualTo(obs2);

    scope1.close(); // previous = null
    //assertThat(registry.getCurrentObservation()).isEqualTo(obs2);
    obs1.stop();
    scope2.close(); // previous = scope1
    //assertThat(registry.getCurrentObservation()).isNull();
    obs2.stop();
}
```

I've added assertions, with the last two commented out. With the changes in this PR, the outcome of the first scope closing remains the same while the second will leave the current observation as `null` (without the changes in this PR, current will be set to `obs1`). In some sense it is an improvement to not end up with scope1 as the current scope after it has been closed. But the state after closing `scope1` seems wrong either way. I'm not sure what, if anything, we should do about that though.

Related to this and the linked issue, in the meantime, we have added validation of scope closing order as a capability to the `TestObservationRegistry` in https://github.com/micrometer-metrics/micrometer/issues/6329. And we are planning to remove the `getLastScope` method in https://github.com/micrometer-metrics/micrometer/pull/7273.

tgrzej-avaloq:
Maybe I can still add my 5 cents here, I think that the underlying problem is that two (or more) logically disjoint scope hierarchies are managed on a single list. When a thread opens and works with two (or more) DB connections their metrics are still not related to each other and I believe they should be managed independently. Having a single list in the `SimpleObservationRegistry` to hold independent scopes introduces an artificial coupling between them which is then the source of the observed issue(s).

PR #6584, opened 2025-08-01
Catch IllegalArgumentException in VirtualThreadMetrics

genuss:
According to javadoc `getPlatformMXBean` may throw `IllegalArgumentException` in case if `mxbeanInterface` is not a platform management interface or not a singleton platform MXBean. This is currently the case for graalvm native image (at least for 24.0.2).

I hesitated to add a test, but it looks a bit excessive. I also thought, maybe it's better to catch all `j.l.Exception`s for simplicity?


jonatan-ivanov (approved):

jonatan-ivanov:
Thanks for the PR!
I think this should be qualified as a bug and since the JMX support for this was introduced in `1.15.x`, I rebased your changes based on that branch so this will be fixed in 1.15 and above.

>I also thought, maybe it's better to catch all j.l.Exceptions for simplicity?

Since the exception is swallowed I think it's fine to keep them narrower.

(Catching Exception and log it out might make sense though, but I think that could be in another PR on `main` as an enhancement.)

Issue #6578, opened 2025-07-30
Synchronize access of current connections in JettyConnectionMetrics


deadok22:
**Describe the bug**
The 2 lambda expressions [here](https://github.com/micrometer-metrics/micrometer/blame/ea8d2c45f5c5accea7bc8ede8a976af6f84060f2/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java#L117C65-L130) access the `maxConnections` and `connectionSamples` fields without synchronization. They are run in Micrometer threads. 
These same fields are updated in Jetty threads and the fields are protected by the `connectionSamplesLock` monitor otherwise.
The lambda expressions must acquire `connectionSamplesLock` to ensure correctness.

jonatan-ivanov:
Thanks for the issue!

Those fields are `final` so they will not be overwritten and the type of synchronization you _wrote_ is not needed. These objects are non immutable though and I think that's the kind of issue you _meant_.

`maxConnections` is a `TimeWindowMax` and it is synchronized.
`connectionSamples` is a `HashMap` which is not thread-safe in itself but if you check how it is used, you should see that writing/reading is synchronized except `connectionSamples.size()` which if you look at the code, is reading a single `int` variable which is atomic.

So I'm not sure why extra locking would be needed, if you think it is still needed, could you please explain what could go wrong, a test could be also helpful.

deadok22:
Hey @jonatan-ivanov,

thanks for taking a look!

> These objects are non immutable though and I think that's the kind of issue you meant.

Precisely!

> writing/reading is synchronized except `connectionSamples.size()` which if you look at the code, is reading a single int variable which is atomic.

You are correct that `int` variable reads are _atomic_: that means the value read will always be _a_ value that was written to the variable, as opposed to some sort of intermediate state. 
The intention of the `connectionSamples.size()` call is to retrieve the size of the collection at the time the call is made. Without synchronization, there is no guarantee the calling thread will observe changes to the size variable made by other threads.

Here's a test program showing effects of failing to synchronize access to an `int` variable:
```java
import java.time.Duration;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;

class Scratch {
    int aConcurrentlyAccessedInt; // adding the `volatile` modifier or otherwise synchronizing access will yield very different results
    public static void main(String[] args) throws InterruptedException {
[...]

PR #6587, opened 2025-08-03
JettyConnectionMetrics: synchronize access to connectionSamples (#6578)

deadok22:
Address [#6578](https://github.com/micrometer-metrics/micrometer/issues/6578) and shorten the duration the lock is held for, as a drive-by

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
@@ -121,7 +121,7 @@ public class JettyConnectionMetrics extends AbstractLifeCycle implements Connect
             .tags(tags)
             .register(registry);
 
-        Gauge.builder("jetty.connections.current", this, jcm -> jcm.connectionSamples.size())
+        Gauge.builder("jetty.connections.current", this, JettyConnectionMetrics::currentConnections)
             .strongReference(true)
             .baseUnit(BaseUnits.CONNECTIONS)
             .description("The current number of open Jetty connections")
@@ -258,6 +258,12 @@ public class JettyConnectionMetrics extends AbstractLifeCycle implements Connect
         catch (NoSuchMethodException ignore) {
         }
         return method;
+    }
+
+    private int currentConnections() {
+        synchronized (connectionSamplesLock) {
+            return connectionSamples.size();
+        }
     }
 
 }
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
```diff
@@ -160,10 +160,12 @@ private static Tags getConnectorNameTag(Connector connector) {
     @Override
     public void onOpened(Connection connection) {
         Timer.Sample started = Timer.start(registry);
+        int connections;
         synchronized (connectionSamplesLock) {
             connectionSamples.put(connection, started);
-            maxConnections.record(connectionSamples.size());
+            connections = connectionSamples.size();
         }
+        maxConnections.record(connections);
```
jonatan-ivanov, line 168: I assume this was intentional before so that `put` and `record` are atomic. Reading the two values are happening separately though (separate gauges) so we can still end up with ~"inconsistent values".

I don't think this is a big problem though but I think if we don't want to ensure that that `put` and `record` are atomic, maybe we can use a `ConcurrentMap` instead of a lock (should be better because of lock striping).

deadok22: > Reading the two values are happening separately though

That's what led me to believe there's no need to make the 2 updates in a single transaction. The `maxConnections` readings still capture the correct maximum number of simultaneous connections observed over time (correct in the sense that it's a value `connectionSamples.size()` call would produce at a point in time).

We need an operation that both updates the map and returns its size transactionally here - that is not something `ConcurrentMap` offers

deadok22:
Hi @jonatan-ivanov, how can I help move this forward?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyConnectionMetrics.java
jonatan-ivanov: It does not need to be atomic (both update the map and returns its size transactionally here), it is totally fine if it let's other threads also update it and return that size for this use-case I think (it needs to be visible though which I think it fulfills).

But that aside, as I mentioned this seems ok.

jonatan-ivanov:
Sorry, I was on vacation and needed some time to get back to this.
I rebased it on top of `1.14.x` to ensure it's ok with those changes and I'm going to merge this as soon as the build turns green.

Thank you very much for all of your efforts fixing this!

PR #6589, opened 2025-08-04
Remove redundant requireNonNull checks

scordio:
This commit also adds `org.assertj.core.internal.annotation.Contract` as a custom contract annotation.

Relates to:
* https://github.com/uber/NullAway/issues/1222
* https://github.com/assertj/assertj/pull/3882

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,7 +117,7 @@ subprojects {
                 )
 
                 option("NullAway:OnlyNullMarked", "true")
-                option("NullAway:CustomContractAnnotations", "io.micrometer.common.lang.internal.Contract")
+                option("NullAway:CustomContractAnnotations", "io.micrometer.common.lang.internal.Contract,org.assertj.core.internal.annotation.Contract")
                 option("NullAway:CheckContracts", "true")
                 option("NullAway:HandleTestAssertionLibraries", "true")
                 if (javaLanguageVersion.canCompileOrRun(22)) {
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
@@ -453,8 +453,7 @@ public class ObservationContextAssert<SELF extends ObservationContextAssert<SELF
         if (p == null) {
             failWithMessage("Observation should have a parent");
         }
-        // see: https://github.com/uber/NullAway/issues/1222
-        return Objects.requireNonNull(p);
+        return p;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
```diff
@@ -453,8 +453,7 @@ private ObservationView checkedParentObservation() {
         if (p == null) {
             failWithMessage("Observation should have a parent");
         }
-        // see: https://github.com/uber/NullAway/issues/1222
-        return Objects.requireNonNull(p);
+        return p;
```
scordio, line 456: Although NullAway is now happy, I still see a warning from IntelliJ:

Image

I'm not sure if there is anything else that I can do here.

Is IntelliJ detecting only `org.jetbrains.annotations.@Contract`?

build.gradle
```diff
@@ -117,7 +117,7 @@ subprojects {
                 )
 
                 option("NullAway:OnlyNullMarked", "true")
-                option("NullAway:CustomContractAnnotations", "io.micrometer.common.lang.internal.Contract")
+                option("NullAway:CustomContractAnnotations", "io.micrometer.common.lang.internal.Contract,org.assertj.core.internal.annotation.Contract")
```
scordio, line 120: Initially, I thought the custom `@Contract` annotation in AssertJ should be an internal detail, hence the `org.assertj.core.internal` package.

However, user projects would need to configure it like here,  and it might feel weird to configure something from an internal package.

I'll discuss this point internally.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
shakuzen: It may be addressed by IntelliJ 2025.2. According to [this](https://www.jetbrains.com/idea/whatsnew/#page__content-jspecify-support--a-major-step-toward-safer-code), support for jSpecify was added.

scordio: I just upgraded but it's still the same, unfortunately, which kind of makes sense because custom `@Contract` annotations are unrelated to JSpecify, [at least for now](https://github.com/uber/NullAway/issues/1222#issuecomment-3113055812).

build.gradle
jonatan-ivanov: It should be but it seems right now NullAway has no discovery mechanism so we need to do this for every `@Contract` annotation.

Even if it is weird to do this, I would consider keeping it internal in case NullAway will have a discovery mechanism or JSpecify adds it making this option obsolete.

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationContextAssert.java
jonatan-ivanov: Usually it's the other way :) but I guess this is an issue we need to live with for now.

jonatan-ivanov (approved):

build.gradle
scordio: We'll keep it internal in version 3 and revise it in version 4, based on user feedback and how JSpecify and NullAway will evolve.

jonatan-ivanov:
Thank you very much for fixing this!

We got automated (dependabot) version upgrades for AssertJ and NullAway so I rebased this PR on top of `main` and removed a few other check we needed to do because of NullAway, those are fixed now.

scordio:
Thanks!

PR #6594, opened 2025-08-07
Add MeterFilter.forPrefix(..) utility method

peterjeschke:
I have noticed that I often want to configure MeterFilters only for a selection of meters that start with a given prefix. For example, I might want to ignore tags for a selection of metrics, but not for all of them. The "normal" approach would be to write a custom MeterFilter that checks the name before filtering. The new utility method helps with this task.


jonatan-ivanov:
Thank you for the PR!

I'm thinking if it would make sense making this a bit more general:
```java
static MeterFilter forMeters(Predicate<Meter.Id> predicate, MeterFilter delegate) { ... }
```

Using `forPrefix` looks like this:
```java
private static final String MAGIC_PREFIX = "test2;

MeterFilter.forPrefix("test", MeterFilter.ignoreTags("ignored"));
MeterFilter.forPrefix(MAGIC_PREFIX, MeterFilter.ignoreTags("ignored"));
```
While `forMeters` could look like:
```java
private static final Predicate<Meter.Id> MAGIC_PREFIX_SELECTOR = id -> id.getName().startsWith("test2");

MeterFilter.forMeters(id -> id.getName().startsWith("test"), MeterFilter.ignoreTags("ignored"));
MeterFilter.forMeters(MAGIC_PREFIX_SELECTOR, MeterFilter.ignoreTags("ignored"));
MeterFilter.forMeters(startsWith("test"), MeterFilter.ignoreTags("ignored")); // you implement startsWith it but it's a one-liner
```

jonatan-ivanov:
I introduced `forMeters` instead of `forPrefix` and polished the PR a little bit.
@shakuzen What do you think?

PR #6608, opened 2025-08-09
Sync OutputCapture from Spring Boot

izeye:
This PR syncs the `OutputCapture` from Spring Boot.

See https://github.com/spring-projects/spring-boot/pull/46685


jonatan-ivanov:
Thank you!

izeye:
@jonatan-ivanov Thanks!

It only fixes tests, so the "type: task" label seems to be more appropriate.

izeye:
@jonatan-ivanov The release notes also need to be updated: https://github.com/micrometer-metrics/micrometer/releases/tag/v1.14.10 and https://github.com/micrometer-metrics/micrometer/releases/tag/v1.15.3

PR #6612, opened 2025-08-10
Add eventexecutor workers metric for netty

### LINKED ISSUE #6375: Add eventexecutor.workers metrics for Netty
We added a few Netty event executor related metrics at #522. This is a request to add another metrics `eventexecutor.workers` to the list of Netty metrics. This new metric represents the total number of Netty's event loops.

**Rationale**
A similar metric was added to Armeria with line/armeria#4675. This is a useful metric while debugging to determine the number of workers without having to explicitly look at the configuration.

**Additional context**
We could choose to use the reference implementation (Apache v2 license) at https://github.com/line/armeria/blob/024f86dedccc7897a6206770dd95852ca2ed66cc/core/src/main/java/com/linecorp/armeria/common/metric/EventLoopMetrics.java#L88

ttaehee:
### Problem
It is difficult to monitor the state of Netty event loop workers in real-time during production, and having this visibility would be helpful for debugging and root cause analysis of network issues.
For example, in our Netty-based WebSocket + STOMP + RabbitMQ environment, we encountered ConnectTimeoutException errors under load. 
Being able to immediately check the status of workers would help us identify and resolve these problems more easily.

### Solution
Add the eventexecutor.workers metric to expose the number of Netty event loop workers in real-time.

### Rationale
This metric allows operators to quickly monitor the count and status of workers, enabling faster diagnosis and response to issues. 
Ultimately, it improves system stability and operational efficiency.

### Local Verification 
I've confirmed locally that the new metric is registered correctly after applying the changes. 
As you can see below, netty.eventexecutor.workers is successfully created with the expected value.

image

<br/>

Resolves #6375

### DIFF
--- a/docs/src/test/java/io/micrometer/docs/netty/NettyMetricsTests.java
+++ b/docs/src/test/java/io/micrometer/docs/netty/NettyMetricsTests.java
@@ -121,7 +121,7 @@ class NettyMetricsTests {
             names.forEach(name -> assertThat(this.registry.get(NettyMeters.EVENT_EXECUTOR_TASKS_PENDING.getName())
                 .tags(Tags.of("name", name))
                 .gauge()
-                .value()).isNaN());
+                .value()).isEqualTo(0.0));
         });
 
         Tags tags = Tags.of("id", String.valueOf(unpooledByteBufAllocator.hashCode()), "allocator.type",
--- a/docs/src/test/java/io/micrometer/docs/netty/NettyMetricsTests.java
+++ b/docs/src/test/java/io/micrometer/docs/netty/NettyMetricsTests.java
@@ -121,7 +121,7 @@ class NettyMetricsTests {
             names.forEach(name -> assertThat(this.registry.get(NettyMeters.EVENT_EXECUTOR_TASKS_PENDING.getName())
                 .tags(Tags.of("name", name))
                 .gauge()
-                .value()).isEqualTo(0.0));
+                .value()).isNaN());
         });
 
         Tags tags = Tags.of("id", String.valueOf(unpooledByteBufAllocator.hashCode()), "allocator.type",
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
@@ -25,6 +25,8 @@ import io.netty.channel.MultithreadEventLoopGroup;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.SingleThreadEventExecutor;
 
+import java.lang.ref.WeakReference;
+
 /**
  * {@link MeterBinder} for Netty event executors.
  * <p>
@@ -78,10 +80,24 @@ public class NettyEventExecutorMetrics implements MeterBinder {
 
     @Override
     public void bindTo(MeterRegistry registry) {
-        Gauge.builder(NettyMeters.EVENT_EXECUTOR_WORKERS.getName(), this::getWorkerCount)
-            .description("The total number of event loop workers.")
-            .tags(tags)
-            .register(registry);
+        WeakReference<Iterable<EventExecutor>> weakExecutors = new WeakReference<>(this.eventExecutors);
+        Gauge.builder(NettyMeters.EVENT_EXECUTOR_WORKERS.getName(), () -> {
+            Iterable<EventExecutor> executors = weakExecutors.get();
+
+            if (executors == null) {
+                return 0.0;
+            }
+
+            if (executors instanceof MultithreadEventLoopGroup) {
+                return (double) ((MultithreadEventLoopGroup) executors).executorCount();
+            }
+
+            int count = 0;
+            for (EventExecutor ignored : executors) {
+                count++;
+            }
+            return (double) count;
+        }).description("The total number of event loop workers.").tags(tags).register(registry);
 
         this.eventExecutors.forEach(eventExecutor -> {
             if (eventExecutor instanceof SingleThreadEventExecutor) {
@@ -95,18 +111,6 @@ public class NettyEventExecutorMetrics implements MeterBinder {
                     .register(registry);
             }
         });
-    }
-
-    private int getWorkerCount() {
-        if (this.eventExecutors instanceof MultithreadEventLoopGroup) {
-            return ((MultithreadEventLoopGroup) this.eventExecutors).executorCount();
-        }
-
-        int count = 0;
-        for (EventExecutor ignored : this.eventExecutors) {
-            count++;
-        }
-        return count;
     }
 
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
@@ -21,6 +21,7 @@ import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.binder.MeterBinder;
 import io.netty.channel.EventLoop;
+import io.netty.channel.MultithreadEventLoopGroup;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.SingleThreadEventExecutor;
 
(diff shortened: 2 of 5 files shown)


shakuzen:
@divijvaidya would this implementation work for what you had in mind when opening #6375?

shakuzen (approved):
Looks good to me. Thank you for a very complete pull request. I'll wait a bit to see if we can get feedback from the original author of the issue requesting this metric.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
```diff
@@ -77,6 +78,16 @@ public NettyEventExecutorMetrics(Iterable<EventExecutor> eventExecutors, Iterabl
 
     @Override
     public void bindTo(MeterRegistry registry) {
+        if (this.eventExecutors instanceof MultithreadEventLoopGroup) {
```
divijvaidya: Is MultithreadEventLoopGroup the right class for all sorts of event loop groups? I am curious to understand why we didn't choose EventLoopGroup here.

ttaehee: Hi,
That's a great question. I chose MultithreadEventLoopGroup for two main reasons:

1. It aligns with the metric's purpose. The eventexecutor.workers metric is designed to measure multi-threaded worker pools. For single-threaded event loops, the value is always 1 and thus less meaningful for monitoring. Targeting MultithreadEventLoopGroup is the most logical approach as it’s the base for all widely-used, multi-threaded implementations.

2. Practicality: We need the executorCount() method to get the worker count. MultithreadEventLoopGroup provides this, but the general EventLoopGroup interface does not.

This approach ensures we're measuring what's actually meaningful while keeping the implementation straightforward. Thanks :)

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetricsTests.java
```diff
@@ -93,4 +98,40 @@ void shouldHaveCustomTags() throws Exception {
         eventExecutors.shutdownGracefully().get(5, TimeUnit.SECONDS);
     }
 
+    @Test
+    void shouldHaveWorkersMetric() throws Exception {
```
divijvaidya, line 103: please add test for other sub-classes of `MultiThreadIoEventLoopGroup` for example, https://netty.io/4.1/api/io/netty/channel/epoll/EpollEventLoopGroup.html

The test should validate that the code works for subclasses as well as direct implementations

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
divijvaidya: > For single-threaded event loops, the value is always 1 and thus less meaningful for monitoring.

You are assuming that the user of the metric knows whether the underlying implementation is single threaded or multi-threaded. Micrometer library can be used by all sorts of use cases, also in situations perhaps where the user of the metric may not know the implementation. 

For the sake of ensuring that this metric doesn't require a conditional clause based on concrete implementation of event loop, I think there is value in representing single threaded loop here as well. From implementation perspective, you can choose to copy https://github.com/line/armeria/blob/024f86dedccc7897a6206770dd95852ca2ed66cc/core/src/main/java/com/linecorp/armeria/common/metric/EventLoopMetrics.java#L88 or have a special case for single event loop here.

ttaehee: Hi @divijvaidya 

Thanks for the great feedback!
I've updated the PR to make the metric more robust and expanded the tests to cover all the cases you mentioned.
- The implementation now supports all EventLoopGroup types by adding a generic fallback, while still keeping the fast-path for MultithreadEventLoopGroup. 
- The new tests verify this works for subclasses and other Iterable types as well.

Let me know what you think.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
```diff
@@ -91,4 +97,16 @@ public void bindTo(MeterRegistry registry) {
...
+    private int getWorkerCount() {
+        if (this.eventExecutors instanceof MultithreadEventLoopGroup) {
+            return ((MultithreadEventLoopGroup) this.eventExecutors).executorCount();
+        }
+
+        int count = 0;
+        for (EventExecutor ignored : this.eventExecutors) {
+            count++;
+        }
+        return count;
```
divijvaidya: There is no test which tests this part of the code. can you please add one

divijvaidya (commented):
Thank you for making the change. Left one final comment

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
ttaehee: Actually, there was already a test covering the fallback-path: shouldHaveWorkersMetricForNonMultithreadEventLoopGroup. 

That said, I agree the intent wasn’t very clear.
To clarify, I replaced it with a simpler mock-based test: shouldCountWorkersForGenericIterableViaFallbackPath.

Thanks for flagging this!

docs/src/test/java/io/micrometer/docs/netty/NettyMetricsTests.java
```diff
@@ -121,7 +121,7 @@ void shouldNotPreventCollectingExecutors() throws Exception {
             names.forEach(name -> assertThat(this.registry.get(NettyMeters.EVENT_EXECUTOR_TASKS_PENDING.getName())
                 .tags(Tags.of("name", name))
                 .gauge()
-                .value()).isNaN());
+                .value()).isEqualTo(0.0));
```
divijvaidya: Let's move this into a separate PR since it is unrelated.

ttaehee: My apologies for the extra notification! I accidentally re-requested the review. 

You're right to point out the test change. It was a necessary update, as my changes caused the original test to fail. 
This simply aligns it with the new expected behavior.

shakuzen: I see what's happening now, I think. However, this change to the test makes the test meaningless. The point of the test is to ensure that the metrics do not prevent the underlying object (executor) from being garbage collected. It's doing that by asserting that eventually after garbage collection is initiated, the underlying object will be garbage collected, and therefore the metric will return NaN.

See https://github.com/micrometer-metrics/micrometer/pull/6641. If the changes here are preventing it from being garbage collected, this should be changed so we aren't reintroducing that bug. Perhaps `getWorkerCount` can take `Iterable<EventExecutor>` as a parameter and the Gauge registration can pass it.

ttaehee: @shakuzen You're right about the strong reference issue—thanks for pointing that out.
I've switched to using a WeakReference to resolve it, and I’ve reverted the test to assert isNaN() accordingly.
Let me know if this approach looks good to you.

shakuzen: I've updated the code in a new commit with what I suggested. By default, we keep weak references of objects passed to Gauge builders as the backing object. That helps eliminate some boilerplate code that would otherwise be common to gauge functions in a lot of places.

shakuzen (approved):
Thank you everyone who worked on this. This is a great example of open source working well. We ended up with something improved with each person's input, and now all users will be able to benefit from that.

injae-kim:
Nice work @ttaehee :+1:

docs/src/test/java/io/micrometer/docs/netty/NettyMetricsTests.java
ttaehee: Great to see this merged! 
Thank you both, @divijvaidya and @shakuzen, for all the help and guidance on this.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetrics.java
```diff
@@ -91,4 +99,16 @@ public void bindTo(MeterRegistry registry) {
         });
     }
 
+    private static double getExecutorCount(Iterable<EventExecutor> executors) {
+        if (executors instanceof MultithreadEventLoopGroup) {
+            return ((MultithreadEventLoopGroup) executors).executorCount();
+        }
+
+        int count = 0;
```
jonatan-ivanov, line 107: I know I'm late to the party but would it make sense to do this before iterating over the executors?
```java
if (executors instanceof Collection) {
    return ((Collection<EventExecutor>) executors).size();
}
```
I mean is there a chance that `executors` will be a `Collection` or there is no implementation in Netty that would conform this?

ttaehee: Hi @jonatan-ivanov
Thanks for the great feedback!
I think it might be more efficient to first check if executors is a Collection, which would avoid an unnecessary loop and also serve as a defensive check for any potential custom implementations.
Should I go ahead and open a new PR for this?

shakuzen: Yes, feel free to send a new PR.

PR #6642, opened 2025-08-17
Add convenience methods to match Meter.Ids

peterjeschke:
A follow-up to #6594, this adds the most likely predicates for usage with MeterFilter.forMeters

### DIFF
--- a/docs/modules/ROOT/pages/concepts/meter-filters.adoc
+++ b/docs/modules/ROOT/pages/concepts/meter-filters.adoc
@@ -160,10 +160,12 @@ See the following example:
 [source, java]
 ----
 registry.config()
-    .meterFilter(MeterFilter.forMeters(startsWith("prefix"), MeterFilter.ignoreTags("extra")));
-
-Predicate<Meter.Id> startsWith(String prefix) {
-    return id -> id.getName().startsWith(prefix);
-}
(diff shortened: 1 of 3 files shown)


peterjeschke:
Failing builds look to be timeout-related, most likely unrelated to the new methods. I can't re-run the builds though

jonatan-ivanov:
Thanks for the PR!
What's wrong with these?
```java
MeterFilter.forMeters(id -> id.getName().equals("test"), ...);
MeterFilter.forMeters(id -> id.getName().startsWith("test"), ...);
MeterFilter.forMeters(id -> id.getType() == GAUGE, ...);
```

Are you really using all three in production (especially `hasType`)?
I'm not convinced this is really needed, also, this can be a slippery slope for adding lots of predicates that are very easy to implement on the user-side. But if more users will ask for this we can consider adding it.

peterjeschke:
I have used a type filter to activate DistributionStatistics for all timers, yes. But if you think this shouldn't be part of the project, that's fine! I understand, that not everything should be included :)

shakuzen:
Thanks for the PR. I'm in agreement with what Jonatan said. While these seem nice for convenience and more direct syntax, I think the benefit is a little too small in this case. We can revisit the decision if something changes in the future. Sorry we didn't accept this PR but we appreciate the effort. Keep the good work coming - thank you.

Issue #6583, opened 2025-08-01
[JOOQ] MetricsDSLContext - fetchExists doesn't report provided tags


eye-dee:
**Describe the bug**
When using `dsl.tag("tag", "value").fetchExists(SOME_TABLE, SOME_CONDITION);`, the expected metric is not reported. The issue arises because the tag value is not being passed correctly to the query listener. Internally, the following code is executed:

```java
fetchExists(selectOne().from(table).where(condition));
```

This causes query tags to be removed from the thread-local context prematurely. Specifically, in `MetricsDSLContext`, the `time()` method clears the tags before they're available to the `JooqExecuteListener`:

```java
public Configuration time(Configuration c) {
    Iterable<Tag> queryTags = contextTags.get();
    contextTags.remove();
    return derive(c, () -> new JooqExecuteListener(registry, tags, () -> queryTags));
}
```

As a result, no tags are attached to the query metrics when `fetchExists` is used.

---

**Environment**

* Micrometer version: 1.15.2
* Micrometer registry: prometheus
* OS: macOS
* Java version: OpenJDK 21

---

**To Reproduce**
Use the following code snippet:

```java
dsl.tag("tag", "value")
   .fetchExists(SOME_TABLE, SOME_CONDITION);
```

Observe that no query metric with the tag `"tag" = "value"` is reported.

---

**Expected behavior**
The tag provided via `dsl.tag(...)` should be passed to the query listener, and metrics should be reported with the specified tag.

---

**Additional context**
It appears that `fetchExists` internally rewrites the query and performs a new one via `selectOne().from(...)`, causing the tags to be cleared before metric recording. This issue may also affect other similar high-level shortcut methods if they internally trigger a new execution context.

HeeChanN:
Hi! I’ve been looking into this issue (#6583) and have an idea for a fix. Before I start implementing, is it okay if I proceed and open a PR?

injae-kim:
Fix PR: https://github.com/micrometer-metrics/micrometer/pull/6643 PTAL 👍

eye-dee:
This issue is not limited to single-table `fetchExists` calls. The same problem occurs when using joins, for example:

```java
dsl.tag("tag", "value")
   .fetchExists(
       MY_TABLE
           .join(OTHER_TABLE).on(MY_TABLE.OTHER_ID.eq(OTHER_TABLE.ID)),
       MY_TABLE.ID.eq(123)
   );
```

Internally this is rewritten to:

[...]

PR #6643, opened 2025-08-18
ensure tags are preserved for fetchExists(Table, Condition)

HeeChanN:
Override MetricsDSLContext#fetchExists(Table, Condition) to delegate to
super.fetchExists(DSL.selectOne().from(table).where(condition)), forcing the
standard SELECT path so that JooqExecuteListener receives the tags and metrics
are reported with the provided tags.

Tests:
- fetchExistsIsTimedWithProvidedTags: verifies that calling
  dsl.tag("name","checkAuthorExists").fetchExists(...) records "jooq.query"
  with tags name=checkAuthorExists and type=read, and timer count = 1 on H2.

related #6583


shakuzen (approved):
Thank you for the pull request with a fix and a test for the issue. I'll rebase it on the `1.14.x` branch so we can easily include it in upcoming maintenance releases.

PR #6667, opened 2025-08-27
Add KeyValues to Observations using annotations

### LINKED ISSUE #5826: Adding metadata to an observation
I am trying to capture some metadata for my spans, specifically the param keys and values.

I know I can use the @Observed annotation but it looks like it only captures lowCardinalityValues and I'm not sure if you can add in param values to them.

The only way I figured out how to capture the param values is to manually wrap the core logic in my service method in an Observation like this:
public List<Recipes> findRecipes(
            String type,
            List<UUID> ownerIds,
            Boolean includeOwner,
            Boolean includeTested,
            String sort,
            Integer limit
) {
    
        KeyValues lowCardinalityKeyValues = KeyValues.of(
                "type", String.valueOf(type),
                "includeOwner", String.valueOf(includeOwner),
                "includeTested", String.valueOf(includeTested),
                "sort", String.valueOf(sort),
                "limit", String.valueOf(limit)
        );

        return Observation.createNotStarted("test", observationRegistry)
                .contextualName("findRecipes")
                .lowCardinalityKeyValues(lowCardinalityKeyValues)
                .highCardinalityKeyValue("ownerIds", String.valueOf(ownerIds))
                .observe(() -> recipeDao.findRecipes(
                        type,
                        ownerIds,
                        includeTested,
                        sort,
                        limit
                ));
    }
I am looking for a better way to do this. Previously with Sleuth I was using @SpanTag and that used to work, but not any more.
**Rationale**
Wrapping all of the logic with an Observation works, but it looks messy and adds quite a bit of code.

**Additional context**
Add any other context about the feature request here, e.g. related issues, prior art.

isanghaessi:
## Summary
Add parameter-based tagging via @ObservedKeyValueTag(s) and handler support (#5826) with AOP.
Because there's inconvenience making span with tags like below.

```java
public List<Recipes> findRecipes(
            String type,
            List<UUID> ownerIds,
            Boolean includeOwner,
            Boolean includeTested,
            String sort,
            Integer limit
) {
        KeyValues lowCardinalityKeyValues = KeyValues.of(
                "type", String.valueOf(type),
                "includeOwner", String.valueOf(includeOwner),
                "includeTested", String.valueOf(includeTested),
                "sort", String.valueOf(sort),
                "limit", String.valueOf(limit)
        );

        return Observation.createNotStarted("test", observationRegistry)
                .contextualName("findRecipes")
                .lowCardinalityKeyValues(lowCardinalityKeyValues)
                .highCardinalityKeyValue("ownerIds", String.valueOf(ownerIds))
                .observe(() -> recipeDao.findRecipes(
                        type,
                        ownerIds,
                        includeTested,
                        sort,
                        limit
                ));
    }
```

This PR introduces parameter-based tag (key-value) support for @Observed by adding:
- New annotations: @ObservedKeyValueTag and @ObservedKeyValueTags.
- AOP support: ObservedKeyValueTagAnnotationHandler and related support classes.
- ObservedAspect extension: setter to inject the tag annotation handler.
- Common utilities: ExtendedImmutableKeyValue for KeyValue class.
- Tests

And, changing some build.gradle.
- Add preferring local module setting, because of conflicts in remote and local module in developing.
- Remove unnecessary dependency in test module.

This brings parity with the developer experience of @MeterTag used in Counted/Timed.
It's just like cherrypick parameter-tag-binding feature of  @MeterTag.

## Changes
- API
    - Add @ObservedKeyValueTag and @ObservedKeyValueTags (container) annotations.
    - Support static values, ValueResolver-based evaluation, and ValueExpressionResolver-based evaluation.
    - Allow multiple annotations on the same parameter for multiple tags(no duplicate key).

- AOP
    - Add ObservedKeyValueTagAnnotationHandler to scan method parameters and attach key-values to Observation.Context
    - Extend ObservedAspect with setObservedKeyValueTagAnnotationHandler(...) to enable the feature when configured.

- Common
    - Enhance key-value construction logic to streamline propagation and immutability where appropriate.

- Tests
    - Update tests to cover static, resolver, and expression-based tagging scenarios.

How to use(Same as @TImed/@MeterTag)
1. Wire the handler into ObservedAspect

- Enable the feature by setting ObservedKeyValueTagAnnotationHandler on your ObservedAspect. Provide factories for ValueResolver and ValueExpressionResolver if you need dynamic values.
``` java
// Java
@Configuration
class ObservationConfig {

    @Bean
    ObservedAspect observedAspect(ObservationRegistry registry) {
        ObservedAspect aspect = new ObservedAspect(registry);
        aspect.setObservedKeyValueTagAnnotationHandler(
            new ObservedKeyValueTagAnnotationHandler(
                clazz -> new MyValueResolver(),          // ValueResolver factory
                clazz -> new MyExpressionResolver()      // ValueExpressionResolver factory
            )
        );
        return aspect;
    }
}

// Example resolvers
final class MyValueResolver implements io.micrometer.common.annotation.ValueResolver {
    @Override
    public String resolve(Object parameter) {
        return parameter == null ? "null" : "vr:" + parameter;
    }
}

final class MyExpressionResolver implements io.micrometer.common.annotation.ValueExpressionResolver {
    @Override
    public String resolve(String expression, Object parameter) {
        // Evaluate expression with your engine of choice; this is a simple demo
        return "expr:" + expression;
    }
}
```
1. Annotate method parameters

- You can derive the tag value directly from the parameter, via a custom ValueResolver, or via a ValueExpressionResolver-based expression. Multiple tags can be attached to the same parameter.
``` java
// Java
class MyService {

    @Observed(name = "my.operation")
    public void process(
        // Key-only: derives value from parameter (toString / null-safe)
        @ObservedKeyValueTag(key = "customerId") String customerId,

        // Static value (if supported) or expression-based via ValueExpressionResolver
        @ObservedKeyValueTag(key = "kind", expression = "'type:' + dtoType") RequestDto dto,

        // Custom resolver-based
        @ObservedKeyValueTag(key = "status", resolver = MyStatusValueResolver.class) int statusCode
    ) {
        // ...
    }

    // Multiple tags from the same parameter (repeatable)
    @Observed(name = "my.batch")
    public void batch(
        @ObservedKeyValueTag(key = "itemId")
        @ObservedKeyValueTag(key = "itemType", expression = "'type:' + type")
        Item item
    ) {
        // ...
    }

    // Container annotation also supported
    @Observed(name = "my.container")
    public void container(
        @ObservedKeyValueTags({
            @ObservedKeyValueTag(key = "a"),
            @ObservedKeyValueTag(key = "b", expression = "'b:' + value")
        }) Data data
    ) {
        // ...
    }
}
```
Evaluation order and fallbacks(Same as @MeterTag)
- If a resolver class is specified, it takes precedence and is used to compute the value.
- Else if an expression is specified, the ValueExpressionResolver is used to compute the value.
- Else the value falls back to the parameter’s toString(); if null, it becomes "null".
- If resolver/expression evaluation throws, the implementation logs the error and falls back to the parameter-based value to keep Observation running.

Support for async/CompletionStage
- The aspect handles methods returning CompletionStage by starting the Observation before proceeding and stopping it upon completion (or error). Parameter-based tags are attached at Observation start and remain unaffected by the async boundary.

Compatibility
- Existing @Observed-based instrumentation continues to work without modification.

## Review Points
- API naming consistency (@ObservedKeyValueTag, @ObservedKeyValueTags) and attribute names.
- Appropriateness of adding ImmutableExtendedKeyValue.
- Test coverage depth and representative cases.

## Reference
Closes #5826.

### DIFF
--- a/docs/modules/ROOT/pages/observation/components.adoc
+++ b/docs/modules/ROOT/pages/observation/components.adoc
@@ -8,6 +8,7 @@ In this section we will describe main components related to Micrometer Observati
 * <<micrometer-observation-events, Signaling Errors and Arbitrary Events>>
 * <<micrometer-observation-convention-example, Observation Convention>>
 * <<micrometer-observation-predicates-filters, Observation Predicates and Filters>>
+* <<micrometer-observation-annotations, Using Annotations With @Observed and @ObservationKeyValue>>
 
 *Micrometer Observation basic flow*
 
@@ -164,7 +165,7 @@ include::{include-java}/observation/ObservationConfiguringTests.java[tags=predic
 -----
 
 [[micrometer-observation-annotations]]
-== Using Annotations With @Observed
+== Using Annotations With @Observed and @ObservationKeyValue
 
 If you have turned on Aspect Oriented Programming (for example, by using `org.aspectj:aspectjweaver`), you can use the `@Observed` annotation to create observations. You can put that annotation either on a method to observe it or on a class to observe all the methods in it.
 
@@ -180,4 +181,20 @@ The following test asserts whether the proper observation gets created when a pr
 [source,java,subs=+attributes]
 -----
 include::{include-java}/observation/ObservationHandlerTests.java[tags=observed_aop,indent=0]
+-----
+
+Also, you can use `@ObservationKeyValue` annotation to add tags via method parameters.
+
+The following example shows an `ObservedServiceWithParameter` that has an annotation on a method:
+
+[source,java,subs=+attributes]
+-----
+include::{include-java}/observation/ObservationHandlerTests.java[tags=observed_service_with_parameter,indent=0]
+-----
+
+The following test asserts whether the proper observation gets created when a proxied `ObservedServiceWithParameter` instance gets called:
+
+[source,java,subs=+attributes]
+-----
+include::{include-java}/observation/ObservationHandlerTests.java[tags=observed_aop_with_parameter,indent=0]
 -----
--- a/docs/src/test/java/io/micrometer/docs/SpelValueExpressionResolver.java
+++ b/docs/src/test/java/io/micrometer/docs/SpelValueExpressionResolver.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micrometer.docs.metrics;
+package io.micrometer.docs;
 
 import io.micrometer.common.annotation.ValueExpressionResolver;
 import io.micrometer.common.util.internal.logging.InternalLogger;
@@ -25,7 +25,7 @@ import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.expression.spel.support.SimpleEvaluationContext;
 
-class SpelValueExpressionResolver implements ValueExpressionResolver {
+public class SpelValueExpressionResolver implements ValueExpressionResolver {
 
     private static final InternalLogger log = InternalLoggerFactory.getInstance(SpelValueExpressionResolver.class);
 
--- a/micrometer-observation/src/main/java/io/micrometer/observation/aop/Cardinality.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/aop/Cardinality.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2025 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.observation.aop;
+
+/**
+ * Represents the cardinality of a key-value. There are two types of cardinality and
+ * treated in different ways.
+ *
+ * @author Seungyong Hong
+ * @author Jonatan Ivanov
+ */
+public enum Cardinality {
+
+    HIGH, LOW
+
+}
--- a/micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservationKeyValueSupport.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservationKeyValueSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 VMware, Inc.
+ * Copyright 2025 VMware, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,37 +13,37 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micrometer.core.aop;
+package io.micrometer.observation.aop;
 
 import io.micrometer.common.annotation.NoOpValueResolver;
 import io.micrometer.common.annotation.ValueExpressionResolver;
 import io.micrometer.common.annotation.ValueResolver;
 import io.micrometer.common.util.StringUtils;
+import io.micrometer.observation.annotation.ObservationKeyValue;
 import org.jspecify.annotations.Nullable;
 
 import java.util.function.Function;
 
 /**
- * Support for {@link MeterTag}.
+ * Support for {@link ObservationKeyValue}.
  *
- * @author Marcin Grzejszczak
- * @author Johnny Lim
  * @author Seungyong Hong
  */
-final class MeterTagSupport {
+class ObservationKeyValueSupport {
 
-    private MeterTagSupport() {
+    private ObservationKeyValueSupport() {
     }
 
-    static String resolveTagKey(MeterTag annotation) {
-        return StringUtils.isNotBlank(annotation.value()) ? annotation.value() : annotation.key();
+    static String resolveKey(ObservationKeyValue observationKeyValue) {
+        return StringUtils.isNotBlank(observationKeyValue.value()) ? observationKeyValue.value()
+                : observationKeyValue.key();
     }
 
     /**
-     * Similar to {@code ObservationKeyValueSupport.resolveTagValue}. The two logics are
-     * similar, so if one is modified, probably the other one should be modified too.
+     * Similar to {@code MeterTagSupport.resolveTagValue}. The two logics are similar, so
+     * if one is modified, probably the other one should be modified too.
      */
-    static String resolveTagValue(MeterTag annotation, @Nullable Object argument,
+    static String resolveValue(ObservationKeyValue annotation, @Nullable Object argument,
             Function<Class<? extends ValueResolver>, ? extends ValueResolver> resolverProvider,
             Function<Class<? extends ValueExpressionResolver>, ? extends ValueExpressionResolver> expressionResolverProvider) {
         String value = null;
--- a/micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
@@ -20,6 +20,7 @@ import io.micrometer.observation.ObservationRegistry;
 import io.micrometer.observation.Observations;
 import io.micrometer.observation.annotation.Observed;
 import io.micrometer.observation.ObservationConvention;
+import io.micrometer.observation.annotation.ObservationKeyValue;
 import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.annotation.Around;
 import org.aspectj.lang.annotation.Aspect;
@@ -68,10 +69,15 @@ import java.util.function.Predicate;
  * }
  * </pre>
  *
+ * To add support for {@link ObservationKeyValue} annotations, set the
+ * {@link ObservationKeyValueAnnotationHandler} via
+ * {@link ObservedAspect#setObservationKeyValueAnnotationHandler(ObservationKeyValueAnnotationHandler)}
+ *
  * @author Jonatan Ivanov
  * @author Yanming Zhou
  * @author Jeonggi Kim
  * @author Maksim Petelin
+ * @author Seungyong Hong
  * @since 1.10.0
  */
 @Aspect
@@ -84,6 +90,8 @@ public class ObservedAspect {
     private final @Nullable ObservationConvention<ObservedAspectContext> observationConvention;
 
     private final Predicate<ProceedingJoinPoint> shouldSkip;
+
+    private @Nullable ObservationKeyValueAnnotationHandler observationKeyValueAnnotationHandler;
 
     /**
      * Create an {@code ObservedAspect} with {@link Observations#getGlobalRegistry()}.
@@ -142,22 +150,26 @@ public class ObservedAspect {
     private @NullUnmarked Object observe(ProceedingJoinPoint pjp, Method method, Observed observed) throws Throwable {
         Observation observation = ObservedAspectObservationDocumentation.of(pjp, observed, this.registry,
                 this.observationConvention);
+        if (observationKeyValueAnnotationHandler != null) {
+            observationKeyValueAnnotationHandler.addAnnotatedParameters(observation.getContext(), pjp);
+        }
+
         if (CompletionStage.class.isAssignableFrom(method.getReturnType())) {
             observation.start();
             Observation.Scope scope = observation.openScope();
             try {
                 Object result = pjp.proceed();
                 if (result == null) {
-                    stopObservation(observation, null);
-                    return result;
+                    stopObservation(observation, pjp, null, null);
+                    return null;
                 }
                 else {
                     CompletionStage<?> stage = (CompletionStage<?>) result;
-                    return stage.whenComplete((res, error) -> stopObservation(observation, error));
+                    return stage.whenComplete((res, error) -> stopObservation(observation, pjp, res, error));
                 }
             }
             catch (Throwable error) {
-                stopObservation(observation, error);
+                stopObservation(observation, pjp, null, error);
                 throw error;
             }
             finally {
@@ -165,7 +177,20 @@ public class ObservedAspect {
             }
         }
         else {
-            return observation.observeChecked(() -> pjp.proceed());
+            observation.start();
+            Observation.Scope scope = observation.openScope();
+            try {
+                Object result = pjp.proceed();
+                stopObservation(observation, pjp, result, null);
+                return result;
+            }
+            catch (Throwable error) {
+                stopObservation(observation, pjp, null, error);
+                throw error;
+            }
+            finally {
+                scope.close();
+            }
         }
     }
 
@@ -188,7 +213,11 @@ public class ObservedAspect {
         return method;
     }
 
-    private void stopObservation(Observation observation, @Nullable Throwable error) {
+    private void stopObservation(Observation observation, ProceedingJoinPoint pjp, @Nullable Object result,
+            @Nullable Throwable error) {
+        if (observationKeyValueAnnotationHandler != null) {
+            observationKeyValueAnnotationHandler.addAnnotatedMethodResult(observation.getContext(), pjp, result);
+        }
         if (error != null) {
             observation.error(error);
         }
@@ -207,6 +236,15 @@ public class ObservedAspect {
             return this.proceedingJoinPoint;
         }
 
+    }
+
+    /**
+     * Setting this enables support for {@link ObservationKeyValue}.
+     * @param observationKeyValueAnnotationHandler annotation handler
+     */
+    public void setObservationKeyValueAnnotationHandler(
+            ObservationKeyValueAnnotationHandler observationKeyValueAnnotationHandler) {
+        this.observationKeyValueAnnotationHandler = observationKeyValueAnnotationHandler;
     }
 
 }
--- a/samples/micrometer-samples-spring-framework6/build.gradle
+++ b/samples/micrometer-samples-spring-framework6/build.gradle
@@ -10,7 +10,6 @@ if (!javaLanguageVersion.canCompileOrRun(17)) {
 
 dependencies {
     implementation project(":micrometer-core")
-    implementation project(":micrometer-observation")
 
     testImplementation project(":micrometer-observation-test")
     testImplementation project(":micrometer-test")
--- a/docs/src/test/java/io/micrometer/docs/metrics/CountedAspectTest.java
+++ b/docs/src/test/java/io/micrometer/docs/metrics/CountedAspectTest.java
@@ -23,6 +23,8 @@ import io.micrometer.core.aop.CountedMeterTagAnnotationHandler;
 import io.micrometer.core.aop.MeterTag;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import io.micrometer.docs.SpelValueExpressionResolver;
+
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 import org.springframework.aop.aspectj.annotation.AspectJProxyFactory;
--- a/docs/src/test/java/io/micrometer/docs/metrics/TimedAspectTest.java
+++ b/docs/src/test/java/io/micrometer/docs/metrics/TimedAspectTest.java
@@ -23,6 +23,8 @@ import io.micrometer.core.aop.MeterTag;
 import io.micrometer.core.aop.TimedAspect;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import io.micrometer.docs.SpelValueExpressionResolver;
+
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 import org.springframework.aop.aspectj.annotation.AspectJProxyFactory;
--- a/docs/src/test/java/io/micrometer/docs/observation/ObservationHandlerTests.java
+++ b/docs/src/test/java/io/micrometer/docs/observation/ObservationHandlerTests.java
@@ -17,14 +17,21 @@ package io.micrometer.docs.observation;
 
 import io.micrometer.common.KeyValue;
 import io.micrometer.common.KeyValues;
+import io.micrometer.common.annotation.ValueExpressionResolver;
+import io.micrometer.common.annotation.ValueResolver;
 import io.micrometer.common.docs.KeyName;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.observation.DefaultMeterObservationHandler;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import io.micrometer.docs.SpelValueExpressionResolver;
 import io.micrometer.observation.*;
 import io.micrometer.observation.annotation.Observed;
+import io.micrometer.observation.annotation.ObservationKeyValue;
+import io.micrometer.observation.annotation.ObservationKeyValues;
+import io.micrometer.observation.aop.Cardinality;
 import io.micrometer.observation.aop.ObservedAspect;
+import io.micrometer.observation.aop.ObservationKeyValueAnnotationHandler;
 import io.micrometer.observation.docs.ObservationDocumentation;
 import io.micrometer.observation.tck.TestObservationRegistry;
 import org.jspecify.annotations.Nullable;
@@ -168,6 +175,46 @@ class ObservationHandlerTests {
                 .hasLowCardinalityKeyValue("class", ObservedService.class.getName())
                 .hasLowCardinalityKeyValue("method", "call").doesNotHaveError();
         // end::observed_aop[]
+        // @formatter:on
+    }
+
+    @Test
+    void annotatedCallShouldBeObservedWithParameter() {
+        // @formatter:off
+        // tag::observed_aop_with_parameter[]
+        // create a test registry
+        TestObservationRegistry registry = TestObservationRegistry.create();
+        // add a system out printing handler
+        registry.observationConfig().observationHandler(new ObservationTextPublisher());
+
+        // create a proxy around the observed service
+        AspectJProxyFactory pf = new AspectJProxyFactory(new ObservedServiceWithParameter());
+        ObservedAspect observedAspect = new ObservedAspect(registry);
+        ValueResolver valueResolver = parameter -> "Value from myCustomTagValueResolver [" + parameter + "]";
+        ValueExpressionResolver valueExpressionResolver = new SpelValueExpressionResolver();
+        observedAspect.setObservationKeyValueAnnotationHandler(
+            new ObservationKeyValueAnnotationHandler(
+                aClass -> valueResolver, aClass -> valueExpressionResolver)
+        );
+
+        pf.addAspect(observedAspect);
+
+        // make a call
+        ObservedServiceWithParameter service = pf.getProxy();
+        service.call("foo");
+
+        // assert that observation has been properly created
+        assertThat(registry)
+                .hasSingleObservationThat()
+                .hasBeenStopped()
+                .hasNameEqualTo("test.call")
+                .hasHighCardinalityKeyValue("key0", "foo")
+                .hasHighCardinalityKeyValue("key1", "foo")
+                .hasHighCardinalityKeyValue("key2", "key2: FOO")
+                .hasHighCardinalityKeyValue("key3", "Value from myCustomTagValueResolver [foo]")
+                .hasLowCardinalityKeyValue("key4", "foo")
+                .doesNotHaveError();
+        // end::observed_aop_with_parameter[]
         // @formatter:on
     }
 
@@ -458,5 +505,20 @@ class ObservationHandlerTests {
 
     }
     // end::observed_service[]
+
+    // tag::observed_service_with_parameter[]
+    static class ObservedServiceWithParameter {
+
+        @Observed(name = "test.call")
+        @ObservationKeyValue(key = "key4", cardinality = Cardinality.LOW)
+        String call(@ObservationKeyValues({ @ObservationKeyValue(key = "key0", cardinality = Cardinality.HIGH),
+                @ObservationKeyValue(key = "key1"),
+                @ObservationKeyValue(key = "key2", expression = "'key2: ' + toUpperCase"),
+                @ObservationKeyValue(key = "key3", resolver = ValueResolver.class) }) String param) {
+            return param;
+        }
+
+    }
+    // end::observed_service_with_parameter[]
 
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/MeterTag.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/MeterTag.java
@@ -24,7 +24,7 @@ import java.lang.annotation.*;
 /**
  * There are 3 different ways to add tags to a meter. All of them are controlled by the
  * annotation values. Precedence is to first try with the {@link ValueResolver}. If the
- * value of the resolver wasn't set, try to evaluate an expression. If there’s no
+ * value of the resolver wasn't set, try to evaluate an expression. If there's no
  * expression just return a {@code toString()} value of the parameter.
  *
  * IMPORTANT: Provided tag values MUST BE of LOW-CARDINALITY. If you fail to provide
--- a/micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservationKeyValue.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservationKeyValue.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2025 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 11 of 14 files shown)


shakuzen:
Thank you for the detailed pull request.

> * Appropriateness of adding ImmutableExtendedKeyValue.

I haven't had a chance to look closely at the changes and consider alternatives, but my initial reaction is I would like to avoid adding this if it is feasible to avoid it.

We'll take a closer look later and give a proper review, but I wanted to give that feedback in case you already have ideas of another good way to implement this.

isanghaessi:
Thanks for opinion @shakuzen.
I changed the code as an alternative and removed ImmutableExtendedKeyValue.
And add some docs.
Please, take a look leter! I'm waiting.

micopiira:
Do you think it would be possible to add support for deriving tags from the returned object using annotations?

isanghaessi:
Yes, I think it's possible to add.
I made it with reference to the logic of @MeterTag, but it was not included in this issue, so I excluded it and applied it.
Do you think this is necessary logic?
If it is needed to add, I can add it.
@shakuzen, @jonatan-ivanov.

build.gradle
```diff
@@ -2,6 +2,12 @@ buildscript {
     ext.javaLanguageVersion = JavaLanguageVersion.of(JavaVersion.current().isJava11Compatible() ? JavaVersion.current().getMajorVersion() : 17)
     ext.javaTargetVersion = JavaVersion.VERSION_1_8
 
+    allprojects {
+        configurations.configureEach {
+            resolutionStrategy.preferProjectModules()
```
shakuzen: What's the reason for this change? I see in the PR description the following:

> Add preferring local module setting, because of conflicts in remote and local module in developing.

Could you explain more the issue? I'd like to make sure I understand.

micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservedKeyValueTag.java
```diff
@@ -0,0 +1,77 @@
...
+ * also can be set by {@link CardinalityType}. default value is
+ * {@link CardinalityType#LOW}.
+ *
+ * @author Seungyong Hong
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Target({ ElementType.PARAMETER })
+@Repeatable(ObservedKeyValueTags.class)
+public @interface ObservedKeyValueTag {
```
shakuzen: I don't think we need "Tag" as part of the name. Unless I'm missing some reason it is there, it'd be better to remove it here and other places in this PR.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedKeyValueTagSupport.java
```diff
@@ -0,0 +1,57 @@
...
+import io.micrometer.common.annotation.ValueResolver;
+import io.micrometer.common.util.StringUtils;
+import io.micrometer.observation.annotation.ObservedKeyValueTag;
+
+/**
+ * Support for {@link ObservedKeyValueTag}.
+ *
+ * @author Seungyong Hong
+ */
+public class ObservedKeyValueTagSupport {
```
shakuzen: Does this need to be public ? It's best practice to create things with the least visibility/access necessary. It gives us the most flexibility to change things without worrying about breaking them when we don't expect users to depend on them directly.

shakuzen:
> Do you think it would be possible to add support for deriving tags from the returned object using annotations?

Sure, I think it's possible. From the perspective of making progress, it may be faster to get this PR merged without that feature and follow up with another pull request to add that functionality, though.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedKeyValueTagSupport.java
isanghaessi: I totally agree with you.
It would be appropriate to change access modifier to 'default'.

micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservedKeyValueTag.java
isanghaessi: It's because There was a opinion with naming in issue.
> This is similar to what we have with @MeterTag, we could have a @...KeyValue... annotation that works the same and also let's you define if the KeyValue is low or high cardinality.

But, it would be great to remove the word 'tag'.

build.gradle
isanghaessi: I faced problem that 'error: cannot find symbol' when executing `./gradlew`.
It was because gradle decided to load 'micrometer-observation'-(what I modified) from 'mavenCentral'. And, there is no changes from local module. Then, there's problem with testing.
So, I added gradle setting that prefers get module from local module changes.

I couldn't build without this setting in my local environment. I don't know if anyone other than me has had this problem.
[My Environment]
- Mac - Ventura 15.6.1
- Intellij 2025.2.1

isanghaessi:
> Thanks for the PR, it looks very complete. I left some comments. I see you changed from the ImmutableExtendedKeyValue approach - thanks for that. I think the current solution is better, but I'm still wondering if somehow there's a better way to handle this. I don't have a good idea right now, but I'll think on it some more.

Very thanks for careful review and opinion @shakuzen!
I left a comments and pushed some changes.
- remove 'tag' from `'...observedkeyvaluetag...`.
- remove unnecessary public access modifier.

docs/src/test/java/io/micrometer/docs/SpelValueExpressionResolver.java
```diff
@@ -25,7 +25,7 @@
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.expression.spel.support.SimpleEvaluationContext;
 
-class SpelValueExpressionResolver implements ValueExpressionResolver {
+public class SpelValueExpressionResolver implements ValueExpressionResolver {
```
jonatan-ivanov, line 28: Since now hat this is used from the `metrics` and from the `observation` package of the docs maybe we should move it out from `metrics`.

micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
```diff
@@ -80,6 +80,33 @@ public AnnotationHandler(BiConsumer<KeyValue, T> keyValueConsumer,
...
+     * Creates a new instance of {@link AnnotationHandler}.
+     * @param keyValueConsumer consumer that takes a {@link KeyValue} and mutates the
+     * {@code <T>} type
+     * @param resolverProvider function converting a class extending a
+     * {@link ValueResolver} to an instance of that class
+     * @param expressionResolverProvider function converting a class extending a
+     * {@link ValueExpressionResolver} to an instance of that class
+     * @param annotation annotation containing {@link KeyValue} related information
+     * @param toKeyValue function converting the annotation and the expression or
+     * @param validToAdd predicate that determines if the annotation and the annotation
```
jonatan-ivanov: I think you inserted it into the wrong line. It should be:
```java
     * @param toKeyValue function converting the annotation and the expression or
     * annotation value to a {@link KeyValue}
```
now it is just:
```java
     * @param toKeyValue function converting the annotation and the expression or
```

Also, not sure if I understand this:
```java
     * @param validToAdd predicate that determines if the annotation and the annotation
     * value to a {@link KeyValue}
```

micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
```diff
@@ -80,6 +80,33 @@ public AnnotationHandler(BiConsumer<KeyValue, T> keyValueConsumer,
         this.expressionResolverProvider = expressionResolverProvider;
         this.annotationClass = annotation;
         this.toKeyValue = toKeyValue;
+        this.validToAdd = (a, o) -> true;
```
jonatan-ivanov: I would replace this whole block to just a call to `this(..., (a, o) -> true)`.

docs/modules/ROOT/pages/observation/components.adoc
```diff
@@ -8,6 +8,7 @@ In this section we will describe main components related to Micrometer Observati
 * <<micrometer-observation-events, Signaling Errors and Arbitrary Events>>
 * <<micrometer-observation-convention-example, Observation Convention>>
 * <<micrometer-observation-predicates-filters, Observation Predicates and Filters>>
+* <<micrometer-observation-annotations, Using Annotations With @Observed and @ObservedKeyValue>>
```
jonatan-ivanov: I'm thinking about the name of `@ObservedKeyValue`. To me, it kind of suggest that a `KeyValue` is observed. I guess the intention was saying that it's the `KeyValue` of `@Observed`.

I'm thinking if `@ObservationKeyValue` would be a better choice (similarly to `@MeterTag`).

/cc @shakuzen?

build.gradle
jonatan-ivanov: This is weird, could you please try what happens if you do these:
- restore this file to its original state
- Close IntelliJ
- stop/kill the gradle daemons: `gradle --stop` (check with pidof or ps if they were really stopped)
- remove the `.gradle` folder in the project folder (where this `build.gradle` file is)
- Remove Micrometer from your local Maven Cache: `$HOME/.m2/repository/io/micrometer`
- Remove Micrometer from your local Gradle File Cache: `$HOME/.gradle/caches/modules-2/files-2.1/io.micrometer`
- Remove your local Gradle Build Cache: `$HOME/.gradle/caches/build-cache-1/`
- Remove the `.idea` folder (also idea files if any like *.ipr, *.iml, *.iws)
- Run `./gradlew build`, wait for it to finish
- Run `idea .` to import the project (make sure the previous step successfully terminated)

build.gradle
```diff
@@ -75,46 +81,46 @@ subprojects {
         tasks.withType(JavaCompile).configureEach {
             if (it.name == "compileJava" && !(it.project.name in ["micrometer-java11"])) {
                 options.errorprone.disable(
-                    "JavaDurationGetSecondsToToSeconds" // Requires JDK 9+
+                        "JavaDurationGetSecondsToToSeconds" // Requires JDK 9+
```
jonatan-ivanov: Could you please rollback these formatting changes?

micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservedKeyValue.java
```diff
@@ -0,0 +1,77 @@
...
+import io.micrometer.common.annotation.NoOpValueResolver;
+import io.micrometer.common.annotation.ValueExpressionResolver;
+import io.micrometer.common.annotation.ValueResolver;
+import io.micrometer.observation.aop.CardinalityType;
+import io.micrometer.observation.aop.ObservedKeyValueAnnotationHandler;
+
+import java.lang.annotation.*;
+
+/**
+ * There are 3 different ways to add tags to an observation. All of them are controlled by
```
jonatan-ivanov: Could you please replace the words tag/tags to key-value/keyvalues everywhere in this file?

micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservedKeyValue.java
```diff
@@ -0,0 +1,77 @@
...
+     * Use this object to resolve the tag value. Has the highest precedence.
+     * @return {@link ValueResolver} class
+     */
+    Class<? extends ValueResolver> resolver() default NoOpValueResolver.class;
+
+    /**
+     * Cardinality type of the tag.
+     * @return {@link CardinalityType} class
+     */
+    CardinalityType cardinality() default CardinalityType.LOW;
```
jonatan-ivanov: I don't think defaulting to low cardinality is a good idea, I would set it to high by default.

micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservedKeyValue.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.lang.annotation.*;
+
+/**
+ * There are 3 different ways to add tags to an observation. All of them are controlled by
+ * the annotation values. Precedence is to first try with the {@link ValueResolver}. If
+ * the value of the resolver wasn't set, try to evaluate an expression. If there’s no
+ * expression just return a {@code toString()} value of the parameter. Cardinality type
+ * also can be set by {@link CardinalityType}. default value is
+ * {@link CardinalityType#LOW}.
```
jonatan-ivanov: If we change the default value below, this line should be updated too.

micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservedKeyValue.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @author Seungyong Hong
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Target({ ElementType.PARAMETER })
```
jonatan-ivanov: `MeterTag` and `MeterTags` also support `ElementType.METHOD` so that return values can be used to attach tags. Should not we support that for Observations too?

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -68,9 +69,14 @@
  * }
  * </pre>
  *
+ * To add support for {@link ObservedKeyValue} annotations set the *
```
jonatan-ivanov: Typo (`*`)?

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -141,6 +149,11 @@ public ObservedAspect(ObservationRegistry registry,
     private @NullUnmarked Object observe(ProceedingJoinPoint pjp, Method method, Observed observed) throws Throwable {
         Observation observation = ObservedAspectObservationDocumentation.of(pjp, observed, this.registry,
                 this.observationConvention);
+
+        if (observedKeyValueAnnotationHandler != null) {
+            observedKeyValueAnnotationHandler.addAnnotatedParameters(observation.getContext(), pjp);
```
jonatan-ivanov: Here's where we could add support for method result: `addAnnotatedMethodResult`.

samples/micrometer-samples-spring-framework6/build.gradle
```diff
@@ -10,7 +10,6 @@ if (!javaLanguageVersion.canCompileOrRun(17)) {
 
 dependencies {
     implementation project(":micrometer-core")
-    implementation project(":micrometer-observation")
```
jonatan-ivanov, line 13: Does this have anything to do with your issue in the root `build.gradle`?

micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
```diff
@@ -59,6 +57,8 @@ public class AnnotationHandler<T> {
 
     private final BiFunction<Annotation, Object, KeyValue> toKeyValue;
 
+    private final BiPredicate<Annotation, Object> validToAdd;
```
jonatan-ivanov: I think we should not do this, see my comment on `ObservedKeyValueAnnotationHandler`.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedKeyValueAnnotationHandler.java
```diff
@@ -0,0 +1,101 @@
...
+                    if (observedKeyValue.cardinality() != CardinalityType.HIGH) {
+                        return false;
+                    }
+
+                    return true;
+                });
+        this.lowCardinalityAnnotationHandler = new AnnotationHandler<>(
+                (keyValue, context) -> context.addLowCardinalityKeyValue(keyValue), resolverProvider,
+                expressionResolverProvider, ObservedKeyValue.class, (annotation, object) -> {
+                    ObservedKeyValue observedKeyValue = (ObservedKeyValue) annotation;
```
jonatan-ivanov: This is not very readable, I think refactoring these two out to separate `private static` inner classes would help.
But before we would try that I think it would make sense to check if this is the best solution for routing low and high cardinality keyvalues.

As far as I can tell the only difference between the two implementations here are 

`addLowCardinalityKeyValue`/`addHighCardinalityKeyValue` in the `keyValueConsumer` (first line) and `CardinalityType.LOW`/`CardinalityType.HIGH` in the `validToAdd` `BiPredicate`. I'm wondering if it would make sense to skip the `BiPredicate` entirely and do this decision in the `keyValueConsumer`: pass the cardinality information and the context in an object that also contains the context.
Does this make sense? Please let me know if not/you want me to help/write some code to show what I'm thinking (at this point I haven't tried but it _seems_ possible).

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedKeyValueSupport.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        else if (StringUtils.isNotBlank(annotation.expression())) {
+            return expressionResolverProvider.apply(ValueExpressionResolver.class)
+                .resolve(annotation.expression(), argument);
+        }
+        else if (argument != null) {
+            return argument.toString();
+        }
+
+        return "";
```
jonatan-ivanov: `MeterTagSupport` did this:
```java
return value == null ? "" : value;
```

Also, maybe it would make sense leaving a note both here and in the Javadoc of `MeterTagSupport` that they are very similar so if one's implementation changes, probably the other should too.

jonatan-ivanov:
This is great, thank you very much!
I left a few comments, please let us know what you think. Also, please let us know if you need any clarification/help.

docs/modules/ROOT/pages/observation/components.adoc
shakuzen: I like your suggestion `@ObservationKeyValue` for the symmetry to `@MeterTag` and for the other reason you mentioned.

micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservedKeyValue.java
shakuzen: That was raised in a previous comment which I responded to here https://github.com/micrometer-metrics/micrometer/pull/6667#issuecomment-3278857646. I didn't look into what additional work is required to support that, but I was trying to get things to a mergeable state faster.

build.gradle
isanghaessi: Thanks!
I solved this problem with fetch all tags from upstream repository, but it was not a gradle cache problem.
I will remove this lines.

The problem is below.
I forked repository only 'main' repository.
But, there's 'nebula' plugin in this project. So, git tags are used to find other project's version.

docs/modules/ROOT/pages/observation/components.adoc
isanghaessi: I will change it.

samples/micrometer-samples-spring-framework6/build.gradle
isanghaessi: Yes. There was an issue with root `build.gradle`.
I fixed that.
But, this is other case that is just optimization that 'micrometer-core' already has 'micrometer-observation' with `api`.
So, this line is unnecessary.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedKeyValueAnnotationHandler.java
isanghaessi: I understood your opinion.
I think that way seems to be risky.
That was my first approach to implement this issue.

There's problem with that way.
keyValueConsumer  has this signature `BiConsumer<KeyValue, T>`. 
And, KeyValue has only String key, String value.
```
public interface KeyValue extends Comparable<KeyValue> {

    /**
     * Use this if you want to indicate that the value is missing.
     */
    String NONE_VALUE = "none";

    String getKey();

    String getValue();

    ...
```
We have to do something(can make side effect) to pass CardinalityType information.
1. Change signature of `keyValueConsumer` in `AnnotationHandler`.
It's likely to cause problems at implementations of `AnnotationHandler`, so it doesn't seem like a good way.
2. Extend KeyValue and pass variable.
I tried this.
Make some `ExtendedKeyValue` like this.
```
class ExtendedKeyValue<T> extends KeyValue {
  T extended;
  ...
}
```
But, this is weired. 
Because we have to cast `KeyValue` to `ExtendedKeyValue`.
The structure in which an interface abstracting motion should be cast and used as a child class does not seem good.

So, I decided to separate two handlers and combine.

However, I accept that it's not easy to read.
I will try to make it better to read.
But, I don't get it how to refactor these two handers to static inner classes.
There's `resolverProvider`, `expressionResolverProvider` that makes these handlers stateful.
I think static inner classes are not stateful.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedKeyValueSupport.java
isanghaessi: I refactored that code. But, I will follow `MeterTagSupport`.
It would be good to note that they are vey similar so if one's implementation changes, probably the other should too.

micrometer-observation/src/main/java/io/micrometer/observation/annotation/ObservedKeyValue.java
isanghaessi: I will add that feature.

isanghaessi:
Really thanks for very detail review @jonatan-ivanov!
I followed most of your comments.

Heres what I followed(left 👍 emojis).
- https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353627272
- https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353634795
- https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353636473
- https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353648232
- https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353671271
- https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353737060

And, heres what still in conversation.
- https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353702949
- https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353732005

build.gradle
shakuzen: Thanks for finding the root cause and sharing. I've added a note to our contributing guide to hopefully save other people the trouble in the future: https://github.com/micrometer-metrics/micrometer/commit/adc44ca7ea3666d6f6b37e2658d490021838de92

isanghaessi:
And, I'm watching why only jdk17 test fails.
I sitll don't know why. In my environment with jdk17(temurin, open jdk) works fine.
If there's any information about that, please let me know.

micrometer-commons/src/main/java/io/micrometer/common/annotation/AnnotationHandler.java
```diff
@@ -75,11 +75,33 @@ public AnnotationHandler(BiConsumer<KeyValue, T> keyValueConsumer,
...
+     * @param keyValueConsumer consumer that takes a {@link KeyValue} and mutates the
+     * {@code <T>} type
+     * @param resolverProvider function converting a class extending a
+     * {@link ValueResolver} to an instance of that class
+     * @param expressionResolverProvider function converting a class extending a
+     * {@link ValueExpressionResolver} to an instance of that class
+     * @param annotation annotation containing {@link KeyValue} related information
+     * @param toKeyValue function converting the annotation and the expression or value to
+     * a {@link KeyValue}
+     * @param validToAdd predicate that determines if the annotation and the annotation
```
jonatan-ivanov: I made a comment about this earlier and it was changed but I'm not sure this was fixed.
The javadoc of the new ctor says this:
```java
     * @param toKeyValue function converting the annotation and the expression or value to
     * a {@link KeyValue}
```
The other (existing) one on the other hand:
```java
     * @param toKeyValue function converting the annotation and the expression or
     * annotation value to a {@link KeyValue}
```
I think the second one is the right one.


Also, I'm not sure I understand this:
```java
* @param validToAdd predicate that determines if the annotation and the annotation
```
It seems the sentence is not complete.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservationKeyValueSupport.java
```diff
@@ -0,0 +1,65 @@
...
+ * Support for {@link ObservationKeyValue}.
+ *
+ * @author Seungyong Hong
+ */
+class ObservationKeyValueSupport {
+
+    private ObservationKeyValueSupport() {
+    }
+
+    public static String resolveTagKey(ObservationKeyValue observationKeyValue) {
```
jonatan-ivanov: Should not this be `resolveKey` (there is no `Tag` here)?
Also, does this need to be `public` given that the class is not?

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservationKeyValueSupport.java
```diff
@@ -0,0 +1,65 @@
...
+        return StringUtils.isNotBlank(observationKeyValue.value()) ? observationKeyValue.value()
+                : observationKeyValue.key();
+    }
+
+    /**
+     * Similar to MeterTagSupport.resolveTagValue. The two logics are similar, so if one
+     * is modified, it looks good to be modified together.
+     */
+    public static String resolveTagValue(ObservationKeyValue annotation, @Nullable Object argument,
```
jonatan-ivanov: Should not this be `resolveValue` (there is no `Tag` here)?
Also, does this need to be `public` given that the class is not?

jonatan-ivanov:
>And, I'm watching why only jdk17 test fails.
I sitll don't know why. In my environment with jdk17(temurin, open jdk) works fine.
If there's any information about that, please let me know.

If you click on "view details" of the failure, CI says this:
>Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)

This means that probably there is nothing wrong with the PR, CI killed the build process for some reason (unfortunately it happens sometime, not sure why). I triggered a re-run, it should be fine soon.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -141,30 +149,45 @@ public ObservedAspect(ObservationRegistry registry,
...
+                stopObservation(observation, scope, error, pjp, null);
                 throw error;
             }
             finally {
                 scope.close();
             }
         }
         else {
-            return observation.observeChecked(() -> pjp.proceed());
+            return observation.observeChecked(() -> {
```
jonatan-ivanov: Adding the method result to an observation changes things here a bit so I'm not sure calling `observeChecked` is a good idea with it mostly because if `pjp.proceed()` throws an exception, the result (`null`) will not be added.

I'm thinking something like this:
```java
Object result = null;
observation.start();
try (Scope ignored = observation.openScope()) {
    result = pjp.proceed();
}
catch (Throwable error) {
    observation.error(error);
    throw error;
}
finally {
    if (observationKeyValueAnnotationHandler != null) {
        observationKeyValueAnnotationHandler.addAnnotatedMethodResult(observation.getContext(), pjp, result);
    }
    observation.stop();
}

return result;
```

Or maybe we should reuse `stopObservation`:
```java
Object result = null;
observation.start();
Observation.Scope scope = observation.openScope();
try {
    result = pjp.proceed();
}
catch (Throwable error) {
    observation.error(error);
    throw error;
}
finally {
    stopObservation(observation, scope, null, pjp, result);
}

return result;
```

What do you think?

isanghaessi: I missed that major point.
Very thank for pointing with example code.
I changed code with 'reusing stopObservation' and added test cases.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
```diff
@@ -141,30 +149,48 @@ public ObservedAspect(ObservationRegistry registry,
...
                 }
             }
             catch (Throwable error) {
-                stopObservation(observation, scope, error);
+                stopObservation(observation, scope, error, pjp, null);
                 throw error;
             }
             finally {
                 scope.close();
             }
```
isanghaessi, line 177: I'm wondering why this code is needed.
I have to know this because it decides me to add same code or not.

This code added via commit(e396f206).
> Fixed issues with leaking scope

I can't see any leaking point.
In the above code, stopObservation is called in all cases, so the scope is closing.

And I can't find any issue or PR with this commit.

jonatan-ivanov: It does not seem necessary to me either but I think we should change this separately from this PR.

jonatan-ivanov:
Hi,

I pushed some changes to the PR, most of the commits are polishing things except one that changes the way low- and high cardinality information is passed.Connected to this discussion: https://github.com/micrometer-metrics/micrometer/pull/6667#discussion_r2353732005

Please notice that this commit rolled back the changes in `AnnotationHandler` so we don't need to introduce any new concepts there (`validToAdd`) and the approach I took did not need surfacing components publicly, everything is a private implementation detail.

Please let us know what you think.

isanghaessi:
Hi @jonatan-ivanov,

First, so thanks for polishing in detail my codes!
But, I think introduce new concept `validToAdd` seems better.

Because of down casting.
We have to do down casting to avoid introducing new concept 'validToAdd'.
I know that 'down casting' is pattern to avoid.

```
    private static void addKeyValue(KeyValue keyValue, Observation.Context context) {
        if (keyValue instanceof KeyValueWithCardinality) {
            if (((KeyValueWithCardinality) keyValue).cardinality == CardinalityType.LOW) {
                context.addLowCardinalityKeyValue(keyValue);
            }
            else {
                context.addHighCardinalityKeyValue(keyValue);
            }
        }
    }
```

Above the code, `KeyValue` is interface to represent `getKey()` and `getValue()` method. But, we use `keyValue.cardinality`. It seems miss-using interface that means not designed properly.
If there's only one way to implement, we have to do down casting. But, we can avoid this pattern to introduce new concept 'vlaidToAdd`.
`AnnotatedHandler` is designed for `KeyValue`.
But, we need `Cardinality` to process `ObservationKeyValue`.
Then, it seems right to add some logic to `AnnotatedHandler`. Changing the existing code that can create a side effect or make code more complicate is a problem, but adding a new code to avoid anti-pattern doesn't seem to have a problem(I think that's why you want to avoid introducing new concept `validToAdd`).

Here is content that shows `down casting` is pattern to avoid.
https://www.reddit.com/r/cpp_questions/comments/az23wc/is_downcasting_a_sign_of_a_bad_code_design/
> In summarize, it tells that downcasting is sign of bad code but there are some excpetions. Because of legacy code or third party dependencies.

jonatan-ivanov:
Thank you for your feedback!
I agree with you that the `KeyValue` -> `KeyValueWithCardinality` cast is unfortunate and that it signals an improper design but let me give you some background about why I did things this way in this change-set:
- When we came up with the `KeyValue` concept, we were considering adding cardinality to it but we decided not to, to keep the class simpler and smaller. This also has some performance benefits. Up until this point, this served us well. I think we cannot really introduce cardinality now without breaking changes but I'm not sure we would have the need to do that, see the next point.
- `AnnotationHandler` lets you customize the type where you set the metadata (type `T`) but it does not let you customize the metadata type itself. It is hard-coded `KeyValue` instead of a type parameter which causes the issue of the cast above and also `KeyValue` is used at places where it is not needed (`CountedMeterTagAnnotationHandler` and `MeterTagAnnotationHandler` both need to use `KeyValue` while they need to set `Tag`). We cannot really fix it without a breaking change.
- I agree that the cast shows an improper design, and I think this boils down to `AnnotationHandler` not letting us defining the metadata type. But this also falls into the "Because of legacy code or third party dependencies" category since this is a piece of code we cannot break to fix it in a minor release.
- In order to work this around, one option is casting which is quite unfortunate but on the other hand, it's hidden from the users, it's a private implementation detail and as such, users should not be confused about it. Also, it does not grow our public API surface. I agree that this is bad but on the other hand, its scope is very narrow and hidden.
Another one is introducing the `validToAdd` concept which I like that it lets us avoiding the cast but I don't think it fixes the root problem in `AnnotationHandler`. On the other hand, it adds a new user-facing concept. It also forces `ObservationKeyValueAnnotationHandler` not being an `AnnotationHandler` (also user-facing) and I think it might be more brittle than the casting approach because of having two handlers in `ObservationKeyValueAnnotationHandler ` and maintaining the "contract" between them and their lambdas. My biggest fear with this approach that if we need another `AnnotationHandler` that has some custom needs, we might end up introducing another user-facing concept just to patch-up the shortcomings of `AnnotationHandler`. So right now I think I would rather bite the bullet and do the cast in a non-user-facing way but also create an issue to fix `AnnotationHandler` in the next major release in Micrometer so that we can eliminate the cast too.

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
jonatan-ivanov: Fyi, someone else also noticed this: https://github.com/micrometer-metrics/micrometer/pull/6728

isanghaessi:
Thank you so much for the detailed explanation, @jonatan-ivanov.
I agree with you 100 percent.

I can understand those.
- There is problem to add function with cardinality without breaking changes to `AnnotationHandler`.
- Introducing `validToAdd` may eliminate the original meaning of `AnnotationHandler`(new concepts may continue to be added to AnnotationHandler whenever features are added in the future).

jonatan-ivanov:
Fyi: I slightly polished the tests.

@isanghaessi Thank you very much again for the PR, we really appreciate your help!

micrometer-observation/src/main/java/io/micrometer/observation/aop/ObservedAspect.java
pema4: > In the above code, stopObservation is called in all cases, so the scope is closing.

This `scope.close()` call is the correct way to close the scope, because `stopObservation` may be called on an arbitrary thread (see #6727)

micrometer-observation/src/main/java/io/micrometer/observation/aop/Cardinality.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.observation.aop;
+
+/**
+ * Represents the cardinality of a key-value. There are two types of cardinality and
+ * treated in different ways.
```
shakuzen, line 20: Probably worth clarifying this is for use with the annotation, specifically. There's no way to use it with `KeyValue` otherwise. We can polish this post-merge.

PR #6670, opened 2025-08-27
Avoid capturing lambda allocation when retrieving existing meters

shakuzen:
In particular focuses on Timer, LongTaskTimer, and Counter as these are the types expected to be retrieved (non-functional) the most on hot code paths. Avoiding this allocation makes the retrieval faster and generates less garbage, causing less GC pressure.

Similar to #6661, this helps with Observation-based instrumentation that's using the `DefaultMeterObservationHandler`, which will retrieve a LongTaskTimer and Timer on start and stop respectively.

Adds new public overloads of timer, longTaskTimer, and counter methods that take `Tags` because the existing methods take `Iterable<Tag>` which means `Tags.of` needs to be called on it, which can result in allocation (to check the size of the Iterable) before JIT optimization.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -456,7 +456,7 @@ public abstract class MeterRegistry {
      * @return A new or existing counter.
      */
     public Counter counter(String name, Iterable<Tag> tags) {
-        return Counter.builder(name).tags(tags).register(this);
+        return counter(name, Tags.of(tags));
     }
 
     /**
@@ -468,6 +468,17 @@ public abstract class MeterRegistry {
      */
     public Counter counter(String name, String... tags) {
         return counter(name, Tags.of(tags));
+    }
+
+    /**
+     * Tracks a monotonically increasing value.
+     * @param name The base metric name
+     * @param tags Sequence of dimensions for breaking down the name.
+     * @return A new or existing counter.
+     * @since 1.16.0
+     */
+    public Counter counter(String name, Tags tags) {
+        return counter(new Meter.Id(name, tags, null, null, Meter.Type.COUNTER));
     }
 
     /**
@@ -498,8 +509,7 @@ public abstract class MeterRegistry {
      * @return A new or existing timer.
      */
     public Timer timer(String name, Iterable<Tag> tags) {
-        return this.timer(new Meter.Id(name, Tags.of(tags), null, null, Meter.Type.TIMER),
-                AbstractTimerBuilder.DEFAULT_DISTRIBUTION_CONFIG, pauseDetector);
+        return this.timer(name, Tags.of(tags));
     }
 
     /**
@@ -511,6 +521,18 @@ public abstract class MeterRegistry {
      */
     public Timer timer(String name, String... tags) {
         return timer(name, Tags.of(tags));
+    }
+
+    /**
+     * Measures the time taken for short tasks and the count of these tasks.
+     * @param name The base metric name
+     * @param tags Sequence of dimensions for breaking down the name.
+     * @return A new or existing timer.
+     * @since 1.16.0
+     */
+    public Timer timer(String name, Tags tags) {
+        return this.timer(new Meter.Id(name, tags, null, null, Meter.Type.TIMER),
+                AbstractTimerBuilder.DEFAULT_DISTRIBUTION_CONFIG, pauseDetector);
     }
 
     /**
@@ -1093,7 +1115,18 @@ public abstract class MeterRegistry {
          * @return A new or existing long task timer.
          */
         public LongTaskTimer longTaskTimer(String name, Iterable<Tag> tags) {
-            return longTaskTimer(new Meter.Id(name, Tags.of(tags), null, null, Meter.Type.LONG_TASK_TIMER),
+            return longTaskTimer(name, Tags.of(tags));
+        }
+
+        /**
+         * Measures the time taken for long tasks.
+         * @param name Name of the long task timer being registered.
+         * @param tags Sequence of dimensions for breaking down the name.
+         * @return A new or existing long task timer.
+         * @since 1.16.0
+         */
+        public LongTaskTimer longTaskTimer(String name, Tags tags) {
+            return longTaskTimer(new Meter.Id(name, tags, null, null, Meter.Type.LONG_TASK_TIMER),
                     LongTaskTimer.Builder.DEFAULT_DISTRIBUTION_CONFIG);
         }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -31,6 +31,8 @@ import io.micrometer.core.instrument.search.MeterNotFoundException;
 import io.micrometer.core.instrument.search.RequiredSearch;
 import io.micrometer.core.instrument.search.Search;
 import io.micrometer.core.instrument.util.TimeUtils;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.NullUnmarked;
 import org.jspecify.annotations.Nullable;
 
 import java.time.Duration;
@@ -106,7 +108,7 @@ public abstract class MeterRegistry {
 
     /**
      * write/remove guarded by meterMapLock, read in
-     * {@link #getOrCreateMeter(DistributionStatisticConfig, BiFunction, Meter.Id, Function)}
+     * {@link #getOrCreateMeter(DistributionStatisticConfig, PauseDetector, NewMeterSupplier, Meter.Id, Function)}
      * is unguarded
      */
     private final Map<Meter.Id, Meter> preFilterIdToMeterMap = new HashMap<>();
@@ -135,7 +137,7 @@ public abstract class MeterRegistry {
 
     private final AtomicBoolean closed = new AtomicBoolean();
 
-    private PauseDetector pauseDetector = new NoPauseDetector();
+    private PauseDetector pauseDetector = NoPauseDetector.INSTANCE;
 
     private @Nullable HighCardinalityTagsDetector highCardinalityTagsDetector;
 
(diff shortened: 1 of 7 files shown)


shakuzen:
What we had before these changes (from the "after" in #6661):

```
       bytes  percent  samples  top
  ----------  -------  -------  ---
224818984183   62.49%   428809  io.micrometer.core.instrument.Meter$Id
134962483827   37.51%   257421  io.micrometer.core.instrument.MeterRegistry$$Lambda.0x00000fc00102da70
     2097148    0.00%        4  java.util.concurrent.locks.AbstractQueuedSynchronizer$ExclusiveNode
```

After these changes, retrieving e.g. a Timer will only allocate the `Meter.Id` (typically 40 bytes) and nothing else, if not customizing the Timer config. This is confirmed by allocation profiling the benchmark:

```
       bytes  percent  samples  top
  ----------  -------  -------  ---
228167605252  100.00%   435196  io.micrometer.core.instrument.Meter$Id
     1572861    0.00%        3  java.util.concurrent.locks.AbstractQueuedSynchronizer$ExclusiveNode

```

And with less allocation, the benchmark results show retrieving an existing Timer is even faster than in #6661. I also realized that having any stale meters affects the performance of retrieving an existing meter because we have to check the stale meter set, whereas we have an early return if the set is empty. This is yet another reason people should heed the warning to avoid setups that result in stale meters. I modified the benchmark and ran it without any stale meters to show the difference. I've summarized the results in the table below.

### MeterRegistrationBenchmark#registerExistingTimer benchmark results

| Version  | Time (per operation) | Allocation |
| ------------- | ------------- | ------------- |
| `1.16.0-M2`  | 25.222 ± 0.419 ns/op  | 176 bytes/op |
| #6661  | 16.465 ± 0.139 ns/op  | 64 bytes/op |
| This PR | 14.737 ±  0.190 ns/op | 40 bytes/op |
| This w/o stale | 12.383 ± 0.074 ns/op | 40 bytes/op |

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -338,9 +340,12 @@ private String getBaseTimeUnitStr() {
      * Only used by {@link Counter#builder(String)}.
      * @param id The identifier for this counter.
      * @return A new or existing counter.
+     * @implNote Avoids allocation in the case the meter is already registered,
+     * particularly capturing lambdas.
```
shakuzen, line 344: I added these implementation notes to the JavaDoc in a few places. It's quite easy to not notice that some code written is a capturing lambda that will likely cause allocation. I suspect we were not consciously aware of that for this code before (I don't recall being aware of that in this part of the code).

We should come up with a way to make sure this doesn't regress. We have some tests that assert there were no allocations in some executed code. Perhaps we could do similar with these. Asserting no allocations is generally easier than asserting a specific number of bytes allocated, since that depends on the JVM and environment.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -352,22 +357,27 @@ Counter counter(Meter.Id id) {
...
+     * from the registry-configured one
      * @return A new or existing timer.
+     * @implNote Avoids allocation in the case the meter is already registered,
+     * particularly capturing lambdas.
      */
     Timer timer(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig,
-            PauseDetector pauseDetectorOverride) {
-        return registerMeterIfNecessary(Timer.class, id, distributionStatisticConfig, (id2, filteredConfig) -> {
-            Meter.Id withUnit = id2.withBaseUnit(getBaseTimeUnitStr());
-            return newTimer(withUnit, filteredConfig.merge(defaultHistogramConfig()), pauseDetectorOverride);
```
shakuzen, line 369: Merging with the default histogram config was moved into `getOrCreateMeter` when the passed `DistributionStatisticConfig` is not null. I think I initially did this in an effort to get rid of references in the lambda body that would be captured. Since things ended up with the registry as a parameter, I'm not sure it matters where the code is; it should be the same either way (famous last words). I suppose it's a little less duplication to have it in `getOrCreateMeter`, but it makes the method body longer in a critical function, which may make it harder to read the code.

Separately, it occurs to me that the method `defaultHistogramConfig()` will most likely always return an equivalent instance but it builds and merges it every time in most implementations. This is wasteful and could likely be improved. It only affects new meters, but it's still wasteful if it's always returning an equivalent config. I did at one point try one naive approach which didn't work, so it can be left as future work to be considered.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -607,14 +619,15 @@ public <T extends Collection<?>> T gaugeCollectionSize(String name, Iterable<Tag
     }
 
     private <M extends Meter> M registerMeterIfNecessary(Class<M> meterClass, Meter.Id id,
-            Function<Meter.Id, M> builder, Function<Meter.Id, M> noopBuilder) {
-        return registerMeterIfNecessary(meterClass, id, null, (id2, conf) -> builder.apply(id2), noopBuilder);
+            BiFunction<MeterRegistry, Meter.Id, M> meterSupplier, Function<Meter.Id, M> noopBuilder) {
+        return registerMeterIfNecessary(meterClass, id, null, null,
+                (registry, mappedId, conf, pd) -> meterSupplier.apply(registry, mappedId), noopBuilder);
```
shakuzen, line 646: This adapting of the `BiFunction` to `NewMeterSupplier` is a capturing lambda that allocates (it did before these changes too), which is why we avoid calling this variant of `registerMeterIfNecessary` from the Counter call stack. It's convenient though, so we still use it for Function meters that should only ever be registered once anyway.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -643,8 +656,29 @@ private Meter.Id mapId(Meter.Id id) {
...
+         * @return a new meter
+         */
+        @NonNull M create(MeterRegistry registry, Meter.Id id, DistributionStatisticConfig distributionStatisticConfig,
+                PauseDetector pauseDetector);
+
+    }
+
     private Meter getOrCreateMeter(@Nullable DistributionStatisticConfig config,
-            BiFunction<Meter.Id, /* Nullable Generic */ DistributionStatisticConfig, ? extends Meter> builder,
+            @Nullable PauseDetector specificPauseDetector, NewMeterSupplier<? extends Meter> meterSupplier,
```
shakuzen, line 682: We take the pause detector as a parameter here so we don't need to capture it in the lambda for the NewMeterSupplier (previously the BiFunction).

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -1058,21 +1115,33 @@ public LongTaskTimer longTaskTimer(String name, String... tags) {
...
+        }
+
+        /**
+         * Measures the time taken for long tasks.
+         * @param name Name of the long task timer being registered.
+         * @param tags Sequence of dimensions for breaking down the name.
+         * @return A new or existing long task timer.
+         * @since 1.16.0
+         */
+        public LongTaskTimer longTaskTimer(String name, Tags tags) {
```
shakuzen, line 1107: This was added to avoid the `Tags.of` call when we already have a `Tags` instance. After JIT warm-up, it should not make any difference, but until then, calls to Tags.of with a Tags instance can allocate an ArrayIterator. This was an issue in the added unit test where JIT will not have eliminated that allocation.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -444,7 +456,7 @@ public RequiredSearch get(String name) {
      * @return A new or existing counter.
      */
     public Counter counter(String name, Iterable<Tag> tags) {
-        return Counter.builder(name).tags(tags).register(this);
+        return counter(name, Tags.of(tags));
```
shakuzen, line 459: Originally I left this with the builder to prove the point that JIT can eliminate the apparent allocation that was here. While that's still true, this change along with the new `counter` method that takes a `Tags` parameter avoids some allocations even before JIT optimizes. This was useful for the added unit test asserting allocated bytes, where JIT won't have optimized the code. It also helps reduce allocation in applications before JIT optimization (e.g. start-up or low traffic applications).

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -501,6 +523,18 @@ public Timer timer(String name, String... tags) {
...
     }
 
+    /**
+     * Measures the time taken for short tasks and the count of these tasks.
+     * @param name The base metric name
+     * @param tags Sequence of dimensions for breaking down the name.
+     * @return A new or existing timer.
+     * @since 1.16.0
+     */
+    public Timer timer(String name, Tags tags) {
```
shakuzen, line 533: We could make this added API (and the equivalent counter/ltt API) non-public if we don't want to add API that's mostly sugar which could overwhelm users with options, but it seems harmless enough, and as mentioned in another comment, there are cases it can improve performance before JIT optimization.

shakuzen (commented):
I added allocation tests, which needed some new public API to avoid calling `Tags.of` with something that's already a Tags instance because it can allocate before JIT optimizes. We looked at reworking the method to avoid that case, but the results were surprising. See https://github.com/micrometer-metrics/micrometer/pull/6148. Perhaps we can revisit that at some point.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: I added tests for this in a subsequent commit. From the non-public methods asserting no allocations worked fine, but to assert only the ID allocation from a public method required adding public API variants of existing API that takes a `Tags` instance to avoid calling `Tags.of` which may allocate before JIT optimizations.

shakuzen:
I'm sure there are ways in which this could be done better, but I'll merge for now. We have some time until the next release, and all of this is implementation details other than the new API that is overloads of existing API. Feedback post merge is welcome.

PR #6682, opened 2025-09-02
Introduce meter conventions concept and apply to some JVM metrics

shakuzen:
Enable specifying a convention to control well-known metrics' names and tags.

The default implementation maintains the existing convention for JVM metrics.
Another implementation is provided based on the current (1.37.0) [OpenTelemetry semantic conventions for JVM metrics](https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/).


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ClassLoaderMetrics.java
```diff
@@ -15,47 +15,72 @@
...
+            .builder(conventions.unloadedConvention().getName(), classLoadingBean,
+                    ClassLoadingMXBean::getUnloadedClassCount)
+            .tags(conventions.unloadedConvention().getTags(null))
             .description("The number of classes unloaded in the Java virtual machine")
             .baseUnit(BaseUnits.CLASSES)
             .register(registry);
+
+        FunctionCounter
+            .builder(conventions.loadedConvention().getName(), classLoadingBean,
+                    ClassLoadingMXBean::getTotalLoadedClassCount)
```
shakuzen: This addresses https://github.com/micrometer-metrics/micrometer/issues/3561 by adding a FunctionCounter for `getTotalLoadedClassCount` with the default name `jvm.classes.loaded.count` which is unfortunately not very symmetrical with the counterpart FunctionCounter for unloaded classes called `jvm.classes.unloaded` but there was always going to be some asymmetry to not break existing convention for the other meters.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
```diff
@@ -36,17 +34,40 @@
  * @see MemoryPoolMXBean
  * @see BufferPoolMXBean
  */
-@NullMarked
 public class JvmMemoryMetrics implements MeterBinder {
 
-    private final Iterable<Tag> tags;
+    private final Iterable<? extends Tag> tags;
```
shakuzen: Still need `tags` for meters not defined in the convention (yet). Namely the metrics based on BufferPoolMXBean

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMetersConventions.java
```diff
@@ -0,0 +1,115 @@
...
+        };
+
+        @Override
+        default Tags getCommonTags(MemoryPoolMXBean memoryPoolBean) {
+            return Tags.of("id", memoryPoolBean.getName(), "area",
+                    MemoryType.HEAP.equals(memoryPoolBean.getType()) ? "heap" : "nonheap");
+        }
+
+        default MeterConvention<MemoryPoolMXBean> getMemoryUsedConvention() {
+            return new MeterConvention<MemoryPoolMXBean>() {
```
shakuzen: It doesn't especially matter for the particular meters where this will be used because they are registered once on binding. For other meters that are retrieved in order to record every time like GC timing metrics, it's wasteful to return a new object like this.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
```diff
@@ -0,0 +1,236 @@
...
+    default JvmCpuMeterConventionGroup jvmCpuMeterConventions(Tags commonTags) {
+        return new DefaultJvmCpuMeterConventionGroup(commonTags);
+    }
+
+    default JvmGcMeterConventionGroup jvmGcMeterConventions() {
+        return new JvmGcMeterConventionGroup() {
+        };
+    }
+
+    abstract class JvmMemoryMeterConventionGroup implements MeterConventionGroup<MemoryPoolMXBean> {
```
shakuzen: Showing what this looks like with an abstract class vs an interface for the purpose of discussion.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetricsTest.java
```diff
@@ -66,6 +68,17 @@ void getThreadStateCountWhenThreadInfoIsNullShouldWork() {
...
     }
 
+    @Test
+    void extraTagsAreApplied() {
+        new JvmThreadMetrics(Tags.of("extra", "tag")).bindTo(registry);
+
+        assertThat(registry.get("jvm.threads.live").tag("extra", "tag").gauge().value()).isPositive();
+        assertThat(registry.get("jvm.threads.daemon").tag("extra", "tag").gauge().value()).isPositive();
+        assertThat(registry.get("jvm.threads.peak").tag("extra", "tag").gauge().value()).isPositive();
+        assertThat(registry.get("jvm.threads.states").tag("extra", "tag").tag("state", "runnable").gauge().value())
```
shakuzen, line 79: I added this test because in a local version it was failing to apply the extra tags. That's still an issue for ClassLoaderMetrics, so we should add a test for that too and fix it.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
```diff
@@ -0,0 +1,236 @@
...
+
+        default MeterConvention<GarbageCollectionNotificationInfo> gcTimeConvention() {
+            return new MeterConvention<GarbageCollectionNotificationInfo>() {
+                @Override
+                public String getName() {
+                    return "jvm.gc.duration";
+                }
+
+                @Override
+                public Tags getTags(GarbageCollectionNotificationInfo gcNotification) {
```
shakuzen: This in public API is a problem, which an ArchTest is causing the build to fail currently. We'll probably have to get clever and use `Notification` or `CompositeData` and cast like is done in the handler now.

```java
CompositeData cd = (CompositeData) notification.getUserData();
GarbageCollectionNotificationInfo notificationInfo = GarbageCollectionNotificationInfo.from(cd);
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
```diff
@@ -0,0 +1,236 @@
...
+
+    }
+
+    interface JvmGcMeterConventionGroup extends MeterConventionGroup<GarbageCollectionNotificationInfo> {
+
+        default MeterConvention<GarbageCollectionNotificationInfo> gcTimeConvention() {
+            return new MeterConvention<GarbageCollectionNotificationInfo>() {
+                @Override
+                public String getName() {
+                    return "jvm.gc.duration";
```
shakuzen: OTel semconv has one metric for GC timing, but the existing Micrometer instrumentation splits it into two: `jvm.gc.concurrent.phase.time` and `jvm.gc.pause`. We'll have to figure out what to do about this in the context of introducing this convention.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
```diff
@@ -0,0 +1,236 @@
...
+                }
+
+                @Override
+                public Tags getTags(GarbageCollectionNotificationInfo gcNotification) {
+                    return getCommonTags(gcNotification).and("jvm.gc.action", gcNotification.getGcAction(),
+                            "jvm.gc.name", gcNotification.getGcName(), "jvm.gc.cause", gcNotification.getGcCause());
```
shakuzen: GC cause in the current OTel semconv spec is marked as opt-in and `development` stability. I'm not sure why, but we'll have to decide what we should do about it.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/MeterConvention.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package io.micrometer.core.instrument.binder;
+
+import io.micrometer.core.instrument.Tags;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.NullUnmarked;
+
+// Work around NullAway. C might be Void, and thus we pass null to getTags.
+@NullUnmarked
+public interface MeterConvention<C> {
```
jonatan-ivanov: With `@NullUnmarked`, `C` in the type parameter (and because of that `C context` in the method) has "unspecified nullness", should we mark the type `@Nullable`?
```java
public interface MeterConvention<C extends @Nullable Object> {...}
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ClassLoaderMetrics.java
```diff
@@ -15,47 +15,71 @@
...
     }
 
     @Override
     public void bindTo(MeterRegistry registry) {
         ClassLoadingMXBean classLoadingBean = ManagementFactory.getClassLoadingMXBean();
 
-        Gauge.builder("jvm.classes.loaded", classLoadingBean, ClassLoadingMXBean::getLoadedClassCount)
-            .tags(tags)
+        Gauge
+            .builder(conventions.currentClassCountConvention().getName(), classLoadingBean,
```
jonatan-ivanov: This makes me thinking if we should have closer integration between `Meter`s and `MeterConvention` by:
- Moving up in the package hierarchy: `core.instrument.binder` -> `core.instrument`
- Add overloads to methods that create a meter to accept a `MeterConvention` so for example this line could look:
```java
Gauge.builder(conventions.currentClassCountConvention(), classLoadingBean, ClassLoadingMXBean::getLoadedClassCount)
    .description("The number of classes that are currently loaded in the Java virtual machine")
    .baseUnit(BaseUnits.CLASSES)
    .register(registry);
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/MeterConventionGroup.java
```diff
@@ -0,0 +1,31 @@
...
+package io.micrometer.core.instrument.binder;
+
+import io.micrometer.core.instrument.Tags;
+
+/**
+ * Provide common tags shared by a group of {@link MeterConvention}.
+ *
+ * @param <C> type of context used to derive common tags
+ */
+public interface MeterConventionGroup<C> {
```
jonatan-ivanov: Should not this look the same from JSpecify/Nullaway perspective as `MeterConvention`?
```java
@NullUnmarked
public interface MeterConventionGroup<C extends @Nullable Object> {

    default @NonNull Tags getCommonTags(C context) {
        return Tags.empty();
    }

}
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ClassLoaderMetrics.java
```diff
@@ -15,47 +15,71 @@
...
     public ClassLoaderMetrics(Iterable<Tag> tags) {
-        this.tags = tags;
+        this(new JvmMetersConventions.JvmClassLoadingMeterConventionGroup() {
+            @Override
+            public Tags getCommonTags(Void context) {
+                return Tags.concat(
+                        JvmMetersConventions.JvmClassLoadingMeterConventionGroup.super.getCommonTags(context), tags);
+            }
+        });
+    }
```
jonatan-ivanov: This might be a common-enough use-case, should not we make it easier to create a `MeterConventionGroup` from a `MeterConventionGroup` and `Iterable<Tags>` so that we can do something like this here:
```java
this(JvmMetersConventions.JvmClassLoadingMeterConventionGroup.DEFAULT.withTags(tags));
```
Or maybe we should keep the two separate and do this for `Meter` creation:
```java
.tags(conventions.currentClassCountConvention().getTags(null))
.tags(conventions.getCommonTags(null))
.tags(tags)
```
Or do what you did with the others:
```java
conventions.jvmThreadMeterConventions(this.tags)
```

I guess an advantage of providing a mechanism for this is having a consistent order for tags from the convention and the user-defined extra ones (which ones are applied first).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ClassLoaderMetrics.java
```diff
@@ -15,47 +15,71 @@
...
     @Override
     public void bindTo(MeterRegistry registry) {
         ClassLoadingMXBean classLoadingBean = ManagementFactory.getClassLoadingMXBean();
 
-        Gauge.builder("jvm.classes.loaded", classLoadingBean, ClassLoadingMXBean::getLoadedClassCount)
-            .tags(tags)
+        Gauge
+            .builder(conventions.currentClassCountConvention().getName(), classLoadingBean,
+                    ClassLoadingMXBean::getLoadedClassCount)
+            .tags(conventions.currentClassCountConvention().getTags(null))
```
jonatan-ivanov: I think this is missing from every `Meter`:
```java
.tags(conventions.getCommonTags(null))
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
```diff
@@ -36,17 +35,30 @@
...
     public JvmMemoryMetrics(Iterable<Tag> tags) {
-        this.tags = tags;
+        this(tags, JvmMetersConventions.DEFAULT);
+    }
+
+    /**
+     * @param convention
+     * @since 1.16.0
+     */
+    public JvmMemoryMetrics(Iterable<? extends Tag> tags, JvmMetersConventions convention) {
```
jonatan-ivanov: I think we should make the ctors (tags, conventions, tags+conventions) consistent across the binders (e.g. `ClassLoaderMetrics` missing the last one).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
```diff
@@ -36,17 +35,30 @@
...
+        this.tags = Tags.of(tags);
+        this.convention = convention.jvmMemoryMeterConventions(this.tags);
```
jonatan-ivanov: See my tags-related comments on `ClassLoaderMetrics`, I think we should make defining extra tags with conventions consistent.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
jonatan-ivanov: I guess one thing we can do is providing the third one (which is redundant).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/MeterConvention.java
shakuzen: I'd have to check again but I think I tried that and it didn't help.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ClassLoaderMetrics.java
shakuzen: The intention (not stated or enforced in any way) is that getTags should return the specific tags for the specific convention, combined with any common tags for the group, combined with any extra tags. So at this point in the code, it shouldn't need to know what to call other than getTags, ideally. That was the idea, anyway.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
shakuzen: This particular inconsistency was intentional, or inherent, due to the fact all Meters produced by ClassLoaderMetrics are covered by its convention, so there is no need to have a constructor that takes extra tags, except for backward compatibility. With the introduction of this convention concept for meters, there's no reason to take a parameter for extra tags separately from the convention, because extra tags can be part of the convention. So if we could break things, and all meters produced are covered by the conventions, we would only need to take the convention in the constructor. Because JvmMemoryMetrics and other binders have additional meters not covered by the convention, we need to take a separate extra tags parameter because the convention doesn't apply to those meters even if the convention produces the desired extra tags.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ClassLoaderMetrics.java
shakuzen: Great idea. I think this helps avoid the peril of how the convention should or shouldn't be called with the builder in the current API. For example, it eliminates the confusion around should you be calling getTags and getCommonTags you pointed out in another comment.
As we discussed offline, it does make me wonder a bit if we should return a builder that doesn't let you call `tags` since those should have been specified by the convention passed to the builder. Maybe something to consider in a future change depending on feedback or additional usage.

shakuzen: Looking at the details here, this is problematic for the tags part of the convention because it requires a context object to be passed, and we may not have one or may not have the right one. For functional meters, if the type of the backing object matches, we could pass that, but in other cases, I don't think we can do anything.

shakuzen:
In the latest update, I've got rid of the `MeterConventionGroup` interface that was there before. I'm not convinced it makes sense as an interface. Let me know if I'm missing some reason it makes sense.

I still have a mix of interfaces and abstract classes for the types returned from `JvmMetersConventions`'s methods as an experiment and to demonstrate the differences. I'm liking the abstract class more because you can make the `getCommonTags` method `protected` which makes it more obvious it is not to be called by the instrumentation but rather within the abstract class and any implementations of it. Let me know if I'm missing some downsides to this approach.

I added a `SimpleMeterConvention` as a convenience to cut down on the number of lines where there was a lot of boilerplate to make anonymous classes implementing `MeterConvention`. It supports passing `Tags` and a function to produce tags.

Another idea I had was to add API to the MeterRegistry to register these groups of conventions, and then instrumentation can check if there are any registered on the MeterRegistry and use them if so. This would avoid needing to pass them to binders and allow a more global configuration. If there isn't a convention group registered, instrumentation can fallback to the default. A concern is that this can be confusing for users if instrumentation they use don't support this new feature, and therefore nothing changes when they register a convention group. It's at least more clear that if a binder takes it in the constructor, it is advertising it supports it.

I changed `Void` type parameters to `Object` so that custom instrumentation could use some custom object passed to their convention implementation that we don't know about. We still pass null and ignore the parameter value in our implementations, but this seemed to offer more flexibility, even if it is ugly. I'm not sure if we can do something better.

Overall, it's quite a lot of public API added for this feature, and I'm concerned about it all being right and how easily we could fix any mess if we need to.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
shakuzen: I fixed this, but I also am leaning toward not tackling GC metrics in this same PR, which means ultimately I'd want to remove any public API related to it until it is ready.

That said, it is a bit of a different use case and that is good for checking the API is sufficient, so there's some motivation to try to implement it there or somewhere similar.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/MeterConvention.java
shakuzen: I checked again and the build fails when changing it to that.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/SimpleMeterConvention.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package io.micrometer.core.instrument.binder;
+
+import io.micrometer.core.instrument.Tags;
+import org.jspecify.annotations.Nullable;
+
+import java.util.Objects;
+import java.util.function.Function;
+
+public class SimpleMeterConvention<T extends @Nullable Object> implements MeterConvention<T> {
```
jonatan-ivanov: It does not matter much but I think the type param should be named the same as `MeterConvention` (`C` vs. `T`).

jonatan-ivanov: Could you please add javadoc with `@since 1.16.0`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ClassLoaderMetrics.java
jonatan-ivanov: Oh, right.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
```diff
@@ -0,0 +1,136 @@
...
+    default JvmMemoryMeterConventions jvmMemoryMeterConventions(Tags extraTags) {
+        return new JvmMemoryMeterConventions(extraTags) {
+        };
+    }
+
+    default JvmClassLoadingMeterConventions jvmClassLoadingMeterConventions() {
+        return JvmClassLoadingMeterConventions.DEFAULT;
+    }
+
+    default JvmThreadMeterConventionGroup jvmThreadMeterConventions(Tags extraTags) {
```
jonatan-ivanov: Is the name intentional? (`JvmThreadMeterConventionGroup ` vs. `JvmThreadMeterConventions`).
Also, should we move these out to a separate `.java` file?

shakuzen: I had not got to this yet as I was experimenting with things yesterday.

shakuzen:
I still need to add JavaDocs, tests, and documentation, but I think the structure is basically done now. Feedback welcome. I've made dedicated interfaces for each conventions grouping and made an implementation of each with a `Micrometer` prefix to the class name for consistency and separation. The OTel implementation extends the `Micrometer` one, which also demonstrates how users could make a customized implementation.

This way instrumentation only depends on the interface rather than an (abstract) class, and hopefully this gives us more flexibility to change things in the future, although the implementations are also public API so users can extend and reuse the logic there.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
```diff
@@ -0,0 +1,41 @@
...
+
+import io.micrometer.core.instrument.Tags;
+
+/**
+ * Defines methods to get the various conventions related to JVM metrics.
+ *
+ * @since 1.16.0
+ * @see #DEFAULT the default implementation
+ */
+public interface JvmMetersConventions {
```
shakuzen: I updated all of the MeterBinders so they aren't referencing this `JvmMetersConventions` anymore. I'm doubtful it is worth keeping it and `MicrometerJvmMetersConventions`. I think I'll remove them and we can add them back if feedback or a clear benefit is demonstrated somehow. They are indirectly used by `OtelJvmMetersConventions` and so I'll need to figure out what makes sense to do there as I add some more tests using that.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
```diff
@@ -40,49 +41,58 @@
 @NullMarked
 public class JvmThreadMetrics implements MeterBinder {
 
-    private final Iterable<Tag> tags;
+    private final Tags extraTags;
+
+    private final JvmThreadMeterConventions convention;
```
jonatan-ivanov: `conventions` instead of `convention`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/ProcessorMetrics.java
```diff
@@ -58,7 +57,9 @@ public class ProcessorMetrics implements MeterBinder {
             "com.sun.management.OperatingSystemMXBean" // HotSpot
     );
 
-    private final Iterable<Tag> tags;
+    private final Tags tags;
+
+    private final JvmCpuMeterConventions convention;
```
jonatan-ivanov: `conventions` instead of `convention`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/JvmMetersConventions.java
shakuzen: I've removed the mentioned interfaces now and I've promoted the Otel conventions classes to top-level classes, but I moved them to their own package, as well as the Micrometer conventions to their own package.

docs/modules/ROOT/pages/concepts/meter-convention.adoc
shakuzen, line 1: This documentation is lacking. We should continue to improve it, but this is a first version to get this new feature into the milestone today.

PR #6700, opened 2025-09-05
Improve no-op behavior of the Observation API

jonatan-ivanov:
There are four cases when an `Observation` can be no-op:
1. The `ObservationRegistry` is `null`
2. The `ObservationRegistry` is `ObservationRegistry.NOOP`
3. The `ObservationRegistry` does not have any handlers
4. `Observation` is disabled by an `ObservationPredicate`

In the first three case, no `Observation`s can be created. In the last case, some `Observation` can be created but some can be ignored. In this scenario we need to support scoping (context-propagation) but in the first three we don't.

This change makes the `NoopObservation` truly no-op to cover the first three cases and introduces `NoopButScopeHandlingObservation` to cover the last one.

This should result in better performance in cases where JIT can eliminate the code because of `NoopObservation` being truly no-op.


jonatan-ivanov:
`ObservationBenchmark#noopObservation` should show the difference. In order to get the behavior before this change, you can change occurrences of this in the `ObservationRegistry`:
```java
if (registry == null || registry.isNoop()) {
	return NoopObservation.INSTANCE;
}
```
to this:
```java
if (registry == null || registry.isNoop()) {
	return NoopButScopeHandlingObservation.INSTANCE;
}
```
the two should produce a significant difference in the no-op benchmark.
Before:
```                                               Mode  Cnt      Score     Error   Units
ObservationBenchmark.noopObservation                     avgt    5      4.701 ±   0.069   ns/op
ObservationBenchmark.noopObservation:gc.alloc.rate       avgt    5  19474.781 ± 287.027  MB/sec
ObservationBenchmark.noopObservation:gc.alloc.rate.norm  avgt    5     24.000 ±   0.001    B/op
ObservationBenchmark.noopObservation:gc.count            avgt    5    997.000            counts
ObservationBenchmark.noopObservation:gc.time             avgt    5    435.000                ms
```
After (benchmarking nothing, JIT optimized the code away):
```                                              Mode  Cnt   Score    Error   Units
ObservationBenchmark.noopObservation                     avgt    5   0.277 ±  0.005   ns/op
ObservationBenchmark.noopObservation:gc.alloc.rate       avgt    5   0.001 ±  0.001  MB/sec
ObservationBenchmark.noopObservation:gc.alloc.rate.norm  avgt    5  ≈ 10⁻⁷             B/op
ObservationBenchmark.noopObservation:gc.count            avgt    5     ≈ 0           counts
```
<details>
<summary>Or you can use this simplified benchmark class (click to reveal)</summary>

```java
@Fork(1)
@Threads(4)
@State(Scope.Benchmark)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class ObservationBenchmark {

    private static final Exception error = new IllegalStateException("error");

    @Benchmark
    public Observation noopObservation() {
        Observation observation = Observation.createNotStarted("test", ObservationRegistry.NOOP).start();
        try (Observation.Scope ignored = observation.openScope()) {
            observation.error(error);
        }
        observation.stop();

        return observation;
    }

    public static void main(String[] args) throws RunnerException {
        new Runner(
            new OptionsBuilder()
                .include(ObservationBenchmark.class.getSimpleName())
                .addProfiler(GCProfiler.class)
                .build()
        ).run();
    }

}
```
</details>

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -309,13 +310,13 @@ static Observation createNotStarted(ObservationConvention<Context> observationCo
      */
     static <T extends Context> Observation createNotStarted(ObservationConvention<T> observationConvention,
             Supplier<T> contextSupplier, ObservationRegistry registry) {
-        if (registry == null || registry.isNoop() || observationConvention == NoopObservationConvention.INSTANCE) {
```
jonatan-ivanov, line 312: `NoopObservationConvention` was never used and it is package-private, since no-one asked for this I removed it to simplify things.

micrometer-observation/src/main/java/io/micrometer/observation/NoopButScopeHandlingObservation.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * No-op implementation of {@link Observation} except scope opening so that we can disable
+ * the instrumentation for certain Observations but keep context-propagation working.
+ *
+ * @author Jonatan Ivanov
+ * @author Tommy Ludwig
+ * @author Marcin Grzejszczak
+ * @since 1.16.0
+ */
+final class NoopButScopeHandlingObservation extends NoopObservation {
```
shakuzen, line 26: If this isn't public, we shouldn't have a `@since` JavaDoc.

shakuzen (commented):
I wonder if something downstream will be broken by this that was relying on scoping working even in some of the no-op cases where it won't now, but I can't think of a specific example. Other than that, it's of course good to eliminate overhead where it isn't needed, so I think the changes are good.

jonatan-ivanov:
>I wonder if something downstream will be broken by this that was relying on scoping working even in some of the no-op cases where it won't now, but I can't think of a specific example.

I was thinking about the same but was not able to come up with a valid example, I have some unlikely ones:
1. Somehow Observations are created with `NoopObservation`, e.g.: users reimplementing the `createNotStarted` methods on their own and they use `NoopObservation` (reflection).
1. NoopObservation is re-used and something depends on the previous behavior (same package/through reflection)
1. `Observation.NOOP` is used in a way where only a subset of Observations go to no-op (reimplementing the functionality of `ObservationPredicate`)

PR #6713, opened 2025-09-09
Validate low cardinality keys

jonatan-ivanov:
Observations with the same name should have the same set of low cardinality keys.
Example 1:
```
observation{name=test, lowCardinalityKeyValues=[color=red]}
observation{name=test, lowCardinalityKeyValues=[color=green]}
observation{name=test, lowCardinalityKeyValues=[color=blue]}
```
Example 1 is valid since all the observations with the same name ("test") has the same set of low cardinality keys ("color").

Example 2:
```
observation{name=test, lowCardinalityKeyValues=[color=red]}
observation{name=test, lowCardinalityKeyValues=[]}
observation{name=test, lowCardinalityKeyValues=[status=ok]}
```
Example 2 is invalid since the second observation is missing a key the first one has ("color") and the third observation is not only missing a key the first one has ("color") but it also has an extra key the first one does not have ("status").

If the validator detects this, it will give you an error like this:
```
Metrics backends may require that all observations with the same name have the same set of low cardinality keys.
There is already an existing observation named 'test' containing keys [color].
The observation you are attempting to register has keys [status].
```

In order to fix it, you should always add the same set of keys to the Observation. If some data is missing, you can signal it with a `"none"`/`"unknown"`/etc. value:
```
observation{name=test, lowCardinalityKeyValues=[color=red, status=unknown]}
observation{name=test, lowCardinalityKeyValues=[color=unknown, status=unknown]}
observation{name=test, lowCardinalityKeyValues=[color=unknown, status=ok]}
```

In your codebase, this means if you attach keyvalues conditionally:
```java
if (color !=null) {
    observation.lowCardinalityKeyValue("color", color);
}
```
you should only apply the condition to the value only:
```java
observation.lowCardinalityKeyValue("color", color !=null ? color : "unknown");
```

If you want to turn this validation off (not recommended), you can do it this way:
```java
TestObservationRegistry registry = TestObservationRegistry.builder()
        .validateObservationsWithTheSameNameHavingTheSameSetOfLowCardinalityKeys(false)
        .build();
```

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/InvalidObservationException.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/InvalidObservationException.java
@@ -19,6 +19,7 @@ import io.micrometer.observation.Observation;
 import io.micrometer.observation.Observation.Context;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -37,6 +38,10 @@ public class InvalidObservationException extends RuntimeException {
 
     private final List<HistoryElement> history;
 
+    InvalidObservationException(String message, Context context) {
+        this(message, context, Collections.emptyList());
+    }
+
     InvalidObservationException(String message, Context context, List<HistoryElement> history) {
         super(message);
         this.context = context;
@@ -53,8 +58,13 @@ public class InvalidObservationException extends RuntimeException {
 
     @Override
     public String toString() {
-        return super.toString() + "\n"
-                + history.stream().map(HistoryElement::toString).collect(Collectors.joining("\n"));
+        if (history.isEmpty()) {
+            return super.toString();
+        }
+        else {
+            return super.toString() + "\n"
+                    + history.stream().map(HistoryElement::toString).collect(Collectors.joining("\n"));
+        }
     }
 
     public static class HistoryElement {
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
@@ -15,6 +15,8 @@
  */
 package io.micrometer.observation.tck;
 
+import io.micrometer.common.KeyValue;
+import io.micrometer.common.KeyValues;
 import io.micrometer.observation.NullObservation.NullContext;
 import io.micrometer.observation.Observation.Context;
 import io.micrometer.observation.Observation.Event;
@@ -23,11 +25,13 @@ import io.micrometer.observation.tck.InvalidObservationException.EventName;
 import io.micrometer.observation.tck.InvalidObservationException.HistoryElement;
 import org.jspecify.annotations.Nullable;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import static io.micrometer.observation.tck.TestObservationRegistry.Capability;
+import static io.micrometer.observation.tck.TestObservationRegistry.Capability.OBSERVATIONS_WITH_THE_SAME_NAME_SHOULD_HAVE_THE_SAME_SET_OF_LOW_CARDINALITY_KEYS;
 
 /**
  * An {@link ObservationHandler} that validates the order of events of an Observation (for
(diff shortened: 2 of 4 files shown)


jonatan-ivanov:
I added a component to be able to turn features on/off.
Right now I disabled this validation by default but now that I just pushed the changes I think maybe it should be enabled by default and if users have issues they can manually disable it by passing an empty array to the `create` method:
```java
TestObservationRegistry.create(new TestObservationRegistry.Capability[]{});
```
or maybe adding an `EMPTY` (or `ALL_OFF`?) capability or array constant so that:
```java
TestObservationRegistry.create(EMPTY);
```

shakuzen:
I'd be fine with it being checked by default if there's a straightforward way to disable it. These kinds of configurations options tend to get complicated easily, though.

jonatan-ivanov:
I enabled by default and added a constant to disable it easily:
```java
TestObservationRegistry.create(ALL_CAPABILITIES_DISABLED);
```

I guess another alternative to make these things configurable is boolean flags with setters or a builder which I think they can get complicated the same way as the enums, not sure what better options we have.

shakuzen (approved):
A builder is most future-proof, I think, but we can try with this for now and switch to a builder later if it makes more sense.

jonatan-ivanov:
Let's use a builder from the beginning since it can prevent making the enum public.
I still kept the enum as an implementation detail to prevent using a bunch of booleans in the future. By default the feature is still enabled but using a builder you can disable it, see the tests.

shakuzen (approved):
LGTM

JoostK:
We've had to disable this validator as it falsely triggers when `MeterRegistry#remove(MeterID)` is being used to clear any prior tags.

jonatan-ivanov:
I would argue that it is a false trigger. The name of the method to enable/disable this is `validateObservationsWithTheSameNameHavingTheSameSetOfLowCardinalityKeys` and its javadoc says:
>Enables/disables validating that Observations with the same name should have the same set of low cardinality keys.

Even though this is most useful from the point of view of metrics but you don't need a `MeterRegistry` to trigger this validation (you can have other outputs using low cardinality key-values).

Also, even if you remove metrics from the registry, you might still have this issue if that data is published to your metrics backend. Unless you also remove the data from your backend too, I would say the validation is not a false trigger, the problem it validates still exists.

shakuzen:
> We've had to disable this validator as it falsely triggers when `MeterRegistry#remove(MeterID)` is being used to clear any prior tags.

I'm lacking context around what you're doing and why. It sounds like you're working around an issue, in which case it probably makes sense you need to disable this validation. If you are using the Prometheus registry, the limitation on registering meters with the same name but a different set of tag keys has been lifted in 1.17.0 (see [the migration guide](https://github.com/micrometer-metrics/micrometer/wiki/1.17-Migration-Guide)). That said, it is still best practice to avoid that. If you have a use case that you think is an exception to that, we'd be interested to learn more about it.

JoostK:
Thanks for the responses! I can see how there's a layering distinction at play here and `MeterRegistry` is just one potential consumer of the general observation registry, so not a source of truth in terms of the captured observations.

This is indeed in the context of a Prometheus registry, so the upcoming changes in 1.17 are interesting. We allow for observation keys to be extended through configuration, such that application refreshes may result in a different bag of observation keys;. These changes are expected to be non-frequent and a refresh listener explicitly purges the meter registry for affected meters to avoid the unsupported change in keys in Prometheus' Java client. The test that verifies this behavior therefore was affected by this validation, as the `TestObservationRegistry` wasn't aware of the purging action (and since only the `MeterRegistry` is being purged, that can be considered WAI as Jonatan mentioned). We do call `TestObservationRegistry#clear()` during the refresh, but I can see why that doesn't also result in clearing the "history" that has been captured in the validator (`#clear()` merely resets the captured observations, not necessarily any facts that have been established for prior observations)

So perhaps it is expected to use the option to disable this validation in this use-case, acknowledging the "risks" that observations with changing keys may have for _any_ handler. So please see this as a data point as to how this validation may affect real world tests and the relevancy of the opt-out flag for future consideration.

Thanks for all your effort in maintaining Micrometer!

jonatan-ivanov:
>We do call TestObservationRegistry#clear() during the refresh, but I can see why that doesn't also result in clearing the "history" that has been captured in the validator (#clear() merely resets the captured observations, not necessarily any facts that have been established for prior observations)

I think this can be a valid feature: when you call `clear` on `TestObservationRegistry`, we could clear the "memory" of the registry and the memory of the "validator" (the latter is not happening today). Do you want to create a pull request for this?

jonatan-ivanov:
@JoostK This might help: https://github.com/micrometer-metrics/micrometer/pull/7337

JoostK:
Thanks, and apologies for the late reply; this was still pending in my inbox. Thanks for #7337, that ought to do it.

PR #6728, opened 2025-09-16
Fix scope leak in ObservedAspect.

### LINKED ISSUE #6727: Scope closed on wrong thread in ObservedAspect with CompletionStage return type
**Describe the bug**
Current `ObservedAspect` implementation for async methods (with return type of `CompletionStage`) leaks parent observation to the future completion thread.
It happens because `Observation.Scope#close()` method is called on the thread that completes the future.

**Environment**
 - Micrometer version: 1.12.2, 1.15.4
 - Micrometer registry: -
 - OS: macOS
 - Java version: openjdk version "24.0.2" 2025-07-15

**To Reproduce**
Here is a code snippet:

```java
class ObservedService {
    @Observed(name = "test.child")
    CompletableFuture<?> child(Supplier<String> supplier) {
        return CompletableFuture.supplyAsync(supplier, executor); // executor is a thread pool
    }

    @Observed(name = "test.parent")
    void parent(Supplier<String> supplier) {
        child(supplier).join();
    }
}
```

When `ObservedService#parent` is called, I expect Micrometer to:
1. open `test.parent` scope in thread `main`
2. open `test.child` scope in thread `main`
3. close `test.child` scope in thread `main`, scope is now `test.parent`
4. close `test.parent` scope in thread `main`, scope is now `null`

But the actual execution is different:
1. open `test.parent` scope in thread `main`
2. open `test.child` scope in thread `main`
3. close `test.child` scope in thread `main`, scope is now `test.parent`
4. close `test.child` scope in thread `pool-1-thread-1`, scope is now `test.parent` — this is wrong
5. close `test.parent` scope in thread `main`, scope is now `null`

Further attempts to call `ObservedService#parent` may trigger [this warning](https://github.com/micrometer-metrics/micrometer/blob/e263348a31a245efd9810d8fcfd9c72a7674183d/micrometer-observation/src/main/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessor.java#L188-L191) — Micrometer detects an attempt to replace an incorrect observation with a new one (that is also incorrect).

**Expected behavior**
`ObservedAspect` should not open any scopes on `CompletionStage` completion thread.

**Additional context**
I will attach a PR with a reproducer and a probable fix.

pema4:
Removed `scope.close` call from a future completion callback

Added new test case `ObservedAspectTests#parentObservationScopeShouldNotLeakToFutureCompletionThread`

Closes #6727


jonatan-ivanov:
Thank you for the PR!
I polished the tests a bit and rebased it on top of `1.14.x` so that older versions will get the fix too.

PR #6822, opened 2025-10-24
Add docs for HighCardinalityTagsDetector

jonatan-ivanov:
The PR is in draft since this needs to be done for `1.14.x`. I made it against `main` since we added a builder and I wanted to get this reviewed including that part. Once this is reviewed and ready to be merged, I will remove the builder parts, rebase on `1.14.x` and put the removed parts back during forward merge.

### DIFF
--- a/docs/modules/ROOT/pages/concepts/high-cardinality-tags-detector.adoc
+++ b/docs/modules/ROOT/pages/concepts/high-cardinality-tags-detector.adoc
@@ -1,0 +1,179 @@
+[[high-cardinality-tags-detector]]
+= High Cardinality Tags Detector
+
+High cardinality tags can cause memory and performance issues in your application and your metrics backend. When tag values are unbounded (for example userID, requestID, traceID), each unique combination creates a new Meter, potentially leading to millions of time series and excessive memory consumption.
+
+`HighCardinalityTagsDetector` helps you to identify Meters that may have high cardinality tags by monitoring the number of Meters that share the same name. When the count exceeds a configurable threshold, it indicates that the `Meter` likely has high cardinality tags.
+
+NOTE: `HighCardinalityTagsDetector` detects potential high cardinality tags by counting Meters with the same name. It does not detect other issues, such as appending random values to Meter names. Its sole purpose is detecting the potential presence of high cardinality tags.
+
+== Understand High Cardinality
+
+High cardinality occurs when a tag has an unbounded or large number of possible values that does not fit in memory. Common examples include:
+
+- UserID, Email, RequestID, SessionID, TraceID
+- Timestamps
+- Full URLs (`/users/123`)
+- Any user input that is not validated/normalized
+
+In contrast, low cardinality tags have a bounded, typically "small" set of values:
+
+- HTTP methods (`method=GET`)
+- HTTP status codes (`status=200`)
+- Application names (`application=payments-app`)
+- Environment names (`env=prod`)
+- Templated URLs (`/users/\{id}`)
+
+TIP: For more information on tag naming best practices, see xref:concepts/naming.adoc#_tag_naming[Tag Naming].
+
+== How It Works
+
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/concepts/high-cardinality-tags-detector.adoc
```diff
@@ -0,0 +1,181 @@
...
+- Collecting metrics about high cardinality detections
+- Triggering any kind of custom actions
+
+== Preventing High Cardinality
+
+When the detector identifies high cardinality tags, consider the following possible solutions:
+
+=== Remove Problematic Tags
+
+If a tag provides little value but high cardinality, remove it:
```
shakuzen: It may be worth specifying to take this approach with tags from instrumentation that you don't control. If you control the implementation, it should be updated rather than using a meter filter.

docs/modules/ROOT/pages/concepts/high-cardinality-tags-detector.adoc
```diff
@@ -0,0 +1,181 @@
...
+- Use templated URLs instead of actual URLs: `/users/\{id}` instead of `/users/123`
+- Group values into ranges: `age.group=20-30` instead of `age=25`
+- Use categories: `outcome=CLIENT_ERROR` (or `status=4xx`) instead of `status=404`
+- Limit unknown/unexpected values to known values: `status=OTHER`
+
+NOTE: `age` and `status` are not necessarily high cardinality data since they are usually fall into a finite set of values (except if input is not validated and `age=12345` and `status=54321` are possible to send to the app). The normalization techniques to reduce cardinality are still valid though.
+
+=== Use High Cardinality data with the Observation API
+
+If you are using Micrometer's Observation API, you can mark certain metadata as high cardinality. These key-values are not used for recording metrics but typically only used in outputs that can handle high cardinality (for example: distributed tracing systems, logs):
```
shakuzen: While we don't use them in the DefaultMeterObservationHandler, they *could* be used there. A minor edit:

```suggestion
If you are using Micrometer's Observation API, you can mark certain metadata as high cardinality. These key-values should not be used for recording metrics. Typically they are only used in outputs that can handle high cardinality (for example: distributed tracing systems, logs):
```

docs/modules/ROOT/pages/concepts/high-cardinality-tags-detector.adoc
```diff
@@ -0,0 +1,181 @@
...
+
+1. **Enable in Production**: Run the detector in production to catch issues that may not appear in testing.
+
+2. **Monitor Regularly**: Use the scheduled monitoring mode rather than one-time checks.
+
+3. **Act on Warnings**: When high cardinality is detected, investigate and fix the root cause rather than just increasing the threshold.
+
+4. **Set Appropriate Thresholds**: The default threshold might work for most applications, but adjust it based on your monitoring system's capabilities and your application's scale as needed.
+
+5. **Combine with MeterFilters**: Use `maximumAllowableTags` and `maximumAllowableMetrics` as preventive measures alongside the detector.
```
shakuzen: I think this advice should be qualified. The max tags filters are for situations in which you cannot fix the cardinality problem. The most common example of this is with HTTP client instrumentation that does not offer a way to tag the URI with a template instead of expanded URI.

I'm not sure what the best succinct wording is, but I would hate for people to use those instead of fixing problems that are fixable.

shakuzen (commented):
Looks good overall. I left a few comments.

docs/modules/ROOT/pages/concepts/high-cardinality-tags-detector.adoc
jonatan-ivanov: I removed it from here to avoid steering people towards it and extended the "Use MeterFilters to Set Bounds" with it.

PR #6825, opened 2025-10-27
Support Postgres 17 in PostgreSQLDatabaseMetrics

### LINKED ISSUE #6657: Support Postgresql 17 in PostgreSQLDatabaseMetrics
**Describe the bug**
After migrating database to Postgresql17 `PostgreSQLDatabaseMetrics` class is still polling old columns `buffers_backend` and `buffers_checkpoint` from view `pg_stat_bgwriter`. But those columns are removed from Postgresql17

**Environment**
 - Micrometer version: 1.15.3
 - Postgresql v17

**To Reproduce**
How to reproduce the bug:
Add meter configuration using PostgreSQLDatabaseMetrics class with datasource postgresql v17;

The-East-Wind:
Resolves #6657 

Stats removed in Pg 17 and their alternatives:

| < Pg17 | Pg17+|
|:---:|:---:|
|`pg_stat_bgwriter.checkpoints_timed`|`pg_stat_checkpointer.num_timed`|
|`pg_stat_bgwriter.checkpoints_req`|`pg_stat_checkpointer.num_requested`|
|`pg_stat_bgwriter.buffers_backend`|`SUM(pg_stat_io.writes)`|
|`pg_stat_bgwriter.buffers_checkpoint`|`pg_stat_checkpointer.buffers_written`|

Reference: https://www.postgresql.org/docs/17/monitoring-stats.html

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
@@ -436,6 +436,8 @@ public class PostgreSQLDatabaseMetrics implements MeterBinder {
             try {
                 final Matcher matcher = VERSION_PATTERN.matcher(versionString);
                 if (!matcher.matches()) {
+                    log.warn("Received Postgres version {} does not match the expected pattern {}", versionString,
+                            VERSION_PATTERN.pattern());
                     return EMPTY;
                 }
                 final String[] versionArr = matcher.group(1).split("\\.", 3);


shakuzen (commented):
The general approach here looks good to me. I'm not sure if @jonatan-ivanov has any thoughts on it or not.

The-East-Wind:
@shakuzen I've polished the changes, review whenever you find time.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
```diff
@@ -366,4 +408,59 @@ private Names() {
...
+
+        final int majorVersion;
+
+        final int minorVersion;
+
+        static Version parse(String versionString) {
+            try {
+                final Matcher matcher = VERSION_PATTERN.matcher(versionString);
+                if (!matcher.matches()) {
+                    return EMPTY;
```
shakuzen, line 441: Are there expected cases in which the version string the server returns won't be parseable by this method? I'm wondering if we should log something when parsing the version fails, or perhaps when `getServerVersion` returns `Version.EMPTY`. Otherwise, users may not notice the issue early on and it will delay feedback about a situation we potentially need to fix this code to handle. What do you think?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
```diff
@@ -272,24 +289,36 @@ private Long getDeadTupleCount() {
         return runQuery(QUERY_DEAD_TUPLE_COUNT);
     }
 
-    private Long getTimedCheckpointsCount() {
-        return runQuery(QUERY_TIMED_CHECKPOINTS_COUNT);
+    private DoubleSupplier getTimedCheckpointsCountSupplier() {
+        if (this.serverVersion.isAbove(Version.V17)) {
```
shakuzen: I thought the new query needs to be run if the version is 17 **or** above, but this looks like it only runs if the version is strictly greater than 17.0. Is that the right logic?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
The-East-Wind: Ideally we don't expect the response to be un parseable. This is added in case a future version of Postgres returns the result in a different format and that doesn't match the regex we use for parsing. Yes, I agree that adding a log will help here. I'll add a warning log where.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
The-East-Wind: The logic implemented returns `true` if the version if 17 or above. Maybe the method naming does not clearly reflect that. I'll change the name to a better suited one.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetrics.java
shakuzen: Thanks for adding that. It looks good. I think it would also be useful to have a warning log here where the versionString doesn't match the expected pattern.

shakuzen (approved):
This looks good to me. I left one follow-up comment. We could also handle this in a polish commit when merging.

The-East-Wind:
> This looks good to me. I left one follow-up comment. We could also handle this in a polish commit when merging.

I've addressed the comment, check whenever you are free and let me know if anything else needs to be changed or added. Thank you.

PR #6829, opened 2025-10-28
Upgrade to Gradle 9.2.1

shakuzen:
Since one of the motivations to upgrade the build to Gradle 9 was support for building with JDK 25, I've also updated the CI config to have a 25 job. Since Gradle 9 needs to be run with a minimum of 17, we need to utilize the toolchain support to run compilation/tests with a different version than Gradle for versions less than 17, such as our 11 build job. See the `toolchainVersion` property that's been added to the build config and the CircleCI config files.

~The BND Gradle plugin appears to not support Gradle 9 yet, so that is a blocker. I've commented out BND related things for the purpose of making progress on other issues upgrading to Gradle 9.~

Skips osgi tests on JDK < 17 because the latest version of BND requires 17.

~We also seem to have an issue with Kotlin and compatibility with JDK 25.~
The Kotlin Gradle plugin (2.2) is not currently compatible with Java 25 so we need to run Gradle with something less than 25, but we can use Gradle toolchain to compile with 25. All build jobs have been updated to use 21 as the Gradle JVM because it is the latest LTS version we can use (due to the Kotlin plugin limitation).

Some CircleCI executor and job names have been changed for clarity/consistency. The required checks need to be updated in GitHub settings.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 version: 2.1
 
 executors:
+# TODO: add and use JDK 25 executor when upgrading to a version of the Kotlin plugin that's compatible
 #  jdk25-executor:
 #    working_directory: ~/micrometer
 #    environment:
--- a/micrometer-bom/build.gradle
+++ b/micrometer-bom/build.gradle
@@ -22,11 +22,3 @@ dependencies {
         api libs.contextPropagation
     }
 }
-
-//publishing {
-//    publications {
-//        nebula(MavenPublication) {
-//            from components.javaPlatform
-//        }
-//    }
-//}


micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterFilter.java
```diff
@@ -414,7 +414,7 @@ public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticC
 
     /**
      * Enables the provided filter for the Meters selected by the predicate.
-     * <code>registry.config().meterFilter(MeterFilter.forMeters(id -> id.getName().startsWith("test"), MeterFilter.ignoreTags("ignored")))</code>
+     * <code>registry.config().meterFilter(MeterFilter.forMeters(id -&gt; id.getName().startsWith("test"), MeterFilter.ignoreTags("ignored")))</code>
```
shakuzen, line 417: I don't understand why this JavaDoc error started happening but it was blocking the build so I fixed it. It looked like the following (although it's for the other JavaDoc with the same issue):

```
> Task :micrometer-jakarta9:javadoc FAILED
/Users/tludwig/repos/micrometer/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/JmsInstrumentation.java:52: error: bad use of '>'
 * consumer.setMessageListener(message -> consumeMessage(message));
                                        ^
1 error
```

Maybe it is worth spending more time to figure out what's different in the JavaDoc configuration that's causing this.

build.gradle
```diff
@@ -20,12 +21,7 @@ buildscript {
         classpath libs.plugin.japicmp
         classpath libs.plugin.downloadTask
         classpath libs.plugin.spotless
-
-        if (javaLanguageVersion.asInt() < 17) {
```
shakuzen, line 24: I assume the condition here is about the version of the JVM used to run Gradle since this is a dependency for the buildscript (Gradle plugin), and since Gradle 9 requires 17, we don't need such conditionals anymore. I think there may be other places we can simplify things as well. I've also added a `gradleJvmVersion` variable we can use if we do need to do something conditional on the version of the JVM that Gradle is using.

build.gradle
```diff
@@ -373,11 +369,9 @@ subprojects {
         apply plugin: 'com.netflix.nebula.maven-apache-license'
         apply plugin: 'com.netflix.nebula.publish-verification'
         apply plugin: 'com.netflix.nebula.contacts'
-        apply plugin: 'com.netflix.nebula.info'
-        apply plugin: 'com.netflix.nebula.project'
```
shakuzen, line 377: I stopped applying these because I would get an error like the following, which occurred at the line applying either of these plugins:

```
A problem occurred evaluating root project 'micrometer'.
> Failed to apply plugin class 'nebula.plugin.info.scm.ScmInfoPlugin'.
   > Extension of type 'ScmInfoExtension' does not exist. Currently registered extension types: [ExtraPropertiesExtension, DevelocityConfiguration, LibrariesForLibs, VersionCatalogsExtension, ReleasePluginExtension, ReleaseExtension, NexusPublishExtension]
```

I could not find anything in the release notes for the plugins about this, but I did find this https://github.com/palantir/gradle-external-publish-plugin/pull/641/files, which seems to indicate we don't need to apply these anymore when applying the maven-publish plugin, which we do before this.

We should confirm we are not missing any functionality we had before.

samples/micrometer-samples-kotlin/build.gradle
```diff
@@ -1,22 +1,29 @@
...
+}
+
 dependencies {
     // libraries that have Kotlin dependencies
     implementation project(":micrometer-core")
     implementation libs.kotlinxCoroutines
-    // force the stdlib versions to match the plugin version for this sample
-    implementation enforcedPlatform('org.jetbrains.kotlin:kotlin-bom')
+    // force the stdlib version
+    implementation enforcedPlatform('org.jetbrains.kotlin:kotlin-bom:1.7.22')
```
shakuzen, line 24: The intention here is to run this sample with Kotlin 1.7, but we need to use a more recent plugin version for Gradle 9 compatibility. I hope this achieves that but I'm not certain.

micrometer-bom/build.gradle
```diff
@@ -23,10 +23,10 @@ dependencies {
...
-        }
-    }
-}
+//publishing {
+//    publications {
+//        nebula(MavenPublication) {
+//            from components.javaPlatform
+//        }
+//    }
+//}
```
shakuzen: I don't remember why I commented this out.

shakuzen:
After upgrading to the BND Gradle plugin 7.2.0-RC1, it looks like everything BND related is working except the `build-jdk11` job. There is still a Kotlin-related issue. I don't have time to look into them now, but at least we've made progress on the BND side of things.

chrisrueger:
> except the `build-jdk11` job

In https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/10913/workflows/fe980361-6468-4404-9964-2c76234f9677/jobs/59468 I noticed the message:

> Could not set unknown property 'sourceSet' for task ':micrometer-osgi-test:testingBundle' of type aQute.bnd.gradle.Bundle.

I am not a gradle expert but it sounds you are using something which was removed gradle 9. In bnd there were recent changes regarding making the bnd gradle plugin compatible with Gradle 9 (https://github.com/bndtools/bnd/pulls?q=is%3Apr+is%3Aclosed+gradle+author%3Abjhargrave+updated%3A2025-11-01..2025-12-29) 

Maybe those PRs can give clues. 


Besides that we identified a Known Issue regarding JUnit versions: https://github.com/bndtools/bnd/wiki/Changes-in-7.2.0#known-issues

micrometer-osgi-test-slf4j2/build.gradle
```diff
@@ -23,7 +22,11 @@ dependencies {
 
     testImplementation libs.felixFramework
 
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+    // BND not compatible with JUnit 5.13+; see https://github.com/bndtools/bnd/issues/6651
+    testImplementation platform('org.junit:junit-bom:5.12.2')
```
shakuzen, line 26: Pins JUnit in the osgi test modules to a compatible version.

build.gradle
```diff
@@ -135,7 +134,7 @@ subprojects {
                 }
             }
         }
-        if ((project.name.contains('samples') && !project.name.contains('kotlin')) || project.name.contains('benchmarks')) {
+        if ((project.name.contains('samples') && !project.name.contains('kotlin')) || project.name.contains('benchmarks') || project.name.contains('osgi-test')) {
```
shakuzen, line 143: osgi test modules are not really java libraries anyway, and this avoids applying the JUnit dependency management we wold need to override in each.

shakuzen:
BND/osgiTest issues have been resolved. We skip the osgiTests in the JDK 11 build now. It passes in the other builds after configuring the right JUnit version.

I've worked around the Kotlin plugin limitation by always using JDK <25 for the Gradle JVM (21 for now). We can still set the toolchain version to 25 for our 25 build.

The remaining failures are the javadoc task failing. This did not happen before and I'm not entirely sure why, but the analysis of the current failures is as follows. For JDK 21:

```
 Task :micrometer-java21:javadoc FAILED
/Users/tludwig/repos/micrometer/micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java:39: error: reference not found
 * @see jdk.management.VirtualThreadSchedulerMXBean
        ^
1 error
```

`VirtualThreadSchedulerMXBean` is only available since 24, so the `javadoc` task fails when trying to link to it with JDK 21.

For JDK 11:

```
> Task :micrometer-test:javadoc
/home/circleci/micrometer/micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryAssert.java:468: error: bad use of '>'
     * Gauge.builder("my.gauge", () -> 42.0)
                                     ^
/home/circleci/micrometer/micrometer-test/src/main/java/io/micrometer/test/assertions/GaugeAssert.java:51: error: bad use of '>'
     * Gauge gauge = Gauge.builder("my.gauge", () -> 42.0).register(registry);
                                                   ^
/home/circleci/micrometer/micrometer-test/src/main/java/io/micrometer/test/assertions/GaugeAssert.java:68: error: bad use of '>'
     * Gauge gauge = Gauge.builder("my.gauge", () -> 42.5).register(registry);
                                                   ^
3 errors

> Task :micrometer-test:javadoc FAILED
```

.circleci/config.yml
```diff
@@ -41,6 +34,9 @@ commands:
       command:
         type: string
         default: 'build'
+      toolchainVersion:
```
jonatan-ivanov, line 38: I'm thinking if we should back-port these (`toolchainVersion`, executor/job naming, etc.) to earlier branches.

docs/build.gradle
```diff
@@ -87,12 +85,6 @@ antora {
     ]
 }
 
-tasks.named('generateAntoraYml') {
```
jonatan-ivanov, line 90: The `antora` task through `generateAntoraResources` depends on this, `./gradlew antora` produces this:
```
* What went wrong:
Could not determine the dependencies of task ':docs:generateAntoraResources'.
> Task with name 'generateAntoraYml' not found in project ':docs'.
```

If I revert the changes in this files, it works.

docs/build.gradle
```diff
@@ -87,12 +85,6 @@ antora {
...
-    asciidocAttributes = project.provider {
-        return ['micrometer-version': project.version.toString()]
```
jonatan-ivanov, line 92: The `micrometer-version` property is used in an `.adoc` file, I think we need the `spring-antora-plugin` and the `generateAntoraYml` task to make it work.

The `generateAntoraResources` task can be removed (the `antora` task and the `antora.yml` need to be modified), though I would keep it as is even if it just "delegates" to `generateAntoraYml` in case we will have other antora-specific resources to generate.

gradle/libs.versions.toml
```diff
@@ -216,21 +216,19 @@ wiremockJunit5 = { module = "ru.lanwen.wiremock:wiremock-junit5", version.ref =
 # plugin dependencies
 plugin-errorprone = { module = "net.ltgt.errorprone:net.ltgt.errorprone.gradle.plugin", version = "4.3.0" }
 plugin-license = { module = "gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin", version = "0.16.1" }
-plugin-nebulaRelease = { module = "com.netflix.nebula:nebula-release-plugin", version = "18.0.8" }
-plugin-nebulaPublishing = { module = "com.netflix.nebula:nebula-publishing-plugin", version = "20.3.0" }
-plugin-nebulaProject = { module = "com.netflix.nebula:nebula-project-plugin", version = "10.1.5" }
-plugin-nebulaInfo = { module = "com.netflix.nebula:gradle-info-plugin", version = "12.1.6" }
+plugin-nebulaRelease = { module = "com.netflix.nebula:nebula-release-plugin", version = "21.0.0" }
+plugin-nebulaPublishing = { module = "com.netflix.nebula:nebula-publishing-plugin", version = "22.0.0" }
```
jonatan-ivanov: A newer version was released very recently:
```suggestion
plugin-nebulaPublishing = { module = "com.netflix.nebula:nebula-publishing-plugin", version = "23.0.0" }
```

gradle/libs.versions.toml
```diff
@@ -216,21 +216,19 @@ wiremockJunit5 = { module = "ru.lanwen.wiremock:wiremock-junit5", version.ref =
...
+plugin-nebulaProject = { module = "com.netflix.nebula:nebula-project-plugin", version = "12.0.0" }
```
jonatan-ivanov: A newer version was released very recently:
```suggestion
plugin-nebulaProject = { module = "com.netflix.nebula:nebula-project-plugin", version = "13.0.0" }
```

gradle/libs.versions.toml
```diff
@@ -216,21 +216,19 @@ wiremockJunit5 = { module = "ru.lanwen.wiremock:wiremock-junit5", version.ref =
...
+plugin-nebulaInfo = { module = "com.netflix.nebula:gradle-info-plugin", version = "15.0.0" }
```
jonatan-ivanov: I'm thinking if we should delete this. IIRC this line wasn't there in the past, we added because there was a bug in the info plugin which was fixed and released but the `nebula-publishing-plugin` and `nebula-project-plugin` were not upgraded immediately and we needed to "force" a newer version of the info plugin this way.

On the other hand, it seems these plugins still disagree the version of info:
```
nebula-publishing-plugin -> gradle-info-plugin:16.0.0
nebula-project-plugin    -> gradle-info-plugin:14.0.0
```
which should resolve to `16.0.0` but maybe it indicates that we should keep forcing the version.

So I think we should either remove this line or upgrade it to `16.0.0`.

implementations/micrometer-registry-statsd/build.gradle
```diff
@@ -1,5 +1,5 @@
 plugins {
-    id 'com.gradleup.shadow' version '8.3.6'
+    id 'com.gradleup.shadow' version '8.3.9'
```
jonatan-ivanov, line 2: Should we create an issue to migrate to 9.x of the plugin (I guess it won't be straightforward because of lots of breaking changes)?

settings.gradle
```diff
@@ -7,7 +7,7 @@ pluginManagement {
 plugins {
     id 'com.gradle.develocity' version '3.19.2'
```
jonatan-ivanov: If it does not make things much more complicated, I would try upgrading this:
```suggestion
    id 'com.gradle.develocity' version '4.2.2'
```

micrometer-osgi-test-slf4j2/build.gradle
jonatan-ivanov: I created https://github.com/micrometer-metrics/micrometer/issues/6935

micrometer-bom/build.gradle
jonatan-ivanov: We will need this if a newer version of the plugin does not auto-configure this, otherwise the bom will not be published (properly or at all).

build.gradle
jonatan-ivanov: This is weird, according to the error message, the plugin's extension is not there (should be different than it is already applied). I'm not sure the nebula maven-publish plugin also applies info; according to its integration tests and behavior (it behaves differently if the info plugin is applied), I would be suspicious that it is not applied.

`./gradlew pTML` also fails.

jonatan-ivanov (commented):
I needed to kill the gradle daemons (killed all of my java processes) otherwise if I did this:
```sh
./gradlew clean build --rerun-tasks --no-build-cache
```
I got:
```
:micrometer-core:compileKotlin FAILED
e: java.lang.IllegalArgumentException: 25
at org.jetbrains.kotlin.com.intellij.util.lang.JavaVersion.parse(JavaVersion.java:307)
at org.jetbrains.kotlin.com.intellij.util.lang.JavaVersion.current(JavaVersion.java:176)
at org.jetbrains.kotlin.cli.jvm.modules.JavaVersionUtilsKt.isAtLeastJava9(javaVersionUtils.kt:11)

[...]

Execution failed for task ':micrometer-core:compileKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
> Internal compiler error. See log for more details
```
See: https://ge.micrometer.io/s/p724iuvb6f5va/console-log/task/:micrometer-core:compileKotlin?page=1#L0

micrometer-bom/build.gradle
shakuzen: I think it caused an error to have it uncommented, but I don't remember the details. We'll have to check.

docs/build.gradle
shakuzen: Reverted

docs/build.gradle
shakuzen: Reverted the change.

implementations/micrometer-registry-statsd/build.gradle
shakuzen: Yes, we should create an issue. I wanted to tackle as few things as needed in this PR since it is already big and taking way too long.

shakuzen:
I've bumped the Kotlin plugin to 2.2.21 and the Kotlin `apiVersion` and `languageVersion` to 1.8 because 1.7 was no longer supported. That allows `./gradlew pTML` to be successful. I haven't checked closely the contents it produces, but at least that seems like progress.

build.gradle
jonatan-ivanov: I looked into this, indeed the plugins are not applied, this is verifiable in the published output: pom.xml is different, manifests in the jars are different, and module properties files are missing.

I fixed it, the root project needs to be included which makes some sense from scm info perspective. I'm still diffing artifacts but so far I only found a non-critical issue:
`nebula_Module_Source` has no value in `pom.xml`
`Module-Source` is missing from manifests and module properties files
(I'm going to open an issue about this.)

implementations/micrometer-registry-statsd/build.gradle
jonatan-ivanov: Created: https://github.com/micrometer-metrics/micrometer/issues/6938

samples/micrometer-samples-kotlin/build.gradle
shakuzen: Still not certain on this and it's interesting to note that we now (are forced to) target Kotlin 1.8 (because that's the lowest the version of the Kotlin plugin that we're using supports) but this (hopefully) runs with 1.7 without issue.

.circleci/config.yml
shakuzen: Maybe. I'm not sure if the added complexity is useful there or would make forward merges better or we'd just be making more work. If we think it's worth it, let's open a separate issue/PR for it.

gradle/libs.versions.toml
```diff
@@ -59,7 +59,7 @@ mongo = "4.11.5"
 netty = "4.2.7.Final"
 newrelic-api = "5.14.0"
 # Kotlin 1.7 sample will fail from OkHttp 4.12.0 due to okio dependency being a Kotlin 1.9 module
-okhttp = "4.11.0"
+okhttp = "4.12.0"
```
shakuzen, line 62: I guess the comment above here is stale? Because we're using a newer Kotlin plugin version now? I don't know.

micrometer-bom/build.gradle
shakuzen: I think we can keep looking into this post-merge. I'll delete it for now since it no longer seems needed.

build.gradle
```diff
@@ -37,6 +36,10 @@ buildscript {
     configurations.classpath.resolutionStrategy.cacheDynamicVersionsFor 0, 'minutes'
 }
 
+plugins {
+    alias(libs.plugins.kotlin) apply false
```
shakuzen, line 40: This needed to be done in the root or a warning was logged about it being declared multiple times.

PR #6845, opened 2025-11-01
Optimize LogbackMetrics ThreadLocal to reduce heap pollution

### LINKED ISSUE #6837: Do Not Repeatedly Clear Logback Metrics ThreadLocal

In some internal testing, I am noticing some performance 'heat' in this area.

https://github.com/micrometer-metrics/micrometer/blob/16084262c63947c2fc61c849dda5abac1f6586ba/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java#L123-L131

Instead 'removing' the record, simply set the value back to 'false'. This will postpone any clean up and removal until the thread dies (if it dies at all).

> Each thread holds an implicit reference to its copy of a thread-local variable as long as the thread is alive and the ThreadLocal instance is accessible; after a thread goes away, all of its copies of thread-local instances are subject to garbage collection (unless other references to these copies exist).

https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html

Also, consider implementing an `initialValue` return return a 'False' value and remove the subsequent `== null` check.

**Rationale**

Better performance and more efficient memory - it pollutes the heap to constantly remove and all of all these Map entry objects.

**Additional context**

A quick JMH benchmark seems to confirm this:

```
Benchmark                           Mode  Cnt   Score   Error  Units
ThreadLocalBenchmark.usingRemove    avgt   30  29.202 ± 0.390  ns/op
ThreadLocalBenchmark.usingSetFalse  avgt   30  15.178 ± 0.232  ns/op
```

And, if I restrict the memory:

```
@Fork(value = 3, jvmArgs = {"-Xms128M", "-Xmx128M", "-XX:+UnlockExperimentalVMOptions", "-XX:+UseEpsilonGC"})
```

The "usingRemove" test fails with an OOM exception, confirming the heap churn.

belugabehr:
Replace ThreadLocal.remove() with set(false) and add initialValue() to eliminate repeated ThreadLocal entry creation/deletion, reducing garbage collection pressure.

Fixes #6837

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/LogbackMetrics.java
@@ -45,7 +45,7 @@ import static java.util.Collections.emptyList;
 @NullMarked
 public class LogbackMetrics implements MeterBinder, AutoCloseable {
 
-    static ThreadLocal<Boolean> ignoreMetrics = new ThreadLocal<>();
+    static ThreadLocal<Boolean> ignoreMetrics = ThreadLocal.withInitial(() -> false);
 
     private final Iterable<Tag> tags;
 
@@ -126,7 +126,7 @@ public class LogbackMetrics implements MeterBinder, AutoCloseable {
             r.run();
         }
         finally {
-            ignoreMetrics.remove();
+            ignoreMetrics.set(false);
         }
     }
 
@@ -209,8 +209,7 @@ class MetricsTurboFilter extends TurboFilter {
             return FilterReply.NEUTRAL;
         }
 
-        Boolean ignored = LogbackMetrics.ignoreMetrics.get();
-        if (ignored != null && ignored) {
+        if (LogbackMetrics.ignoreMetrics.get()) {
             return FilterReply.NEUTRAL;
         }
 


shakuzen:
Thanks for the pull request.

This is similar to https://github.com/micrometer-metrics/micrometer/pull/3952 which we haven't merged for the concerns mentioned there. I do think it'd be best if we solved this more fundamentally instead of trying to do more with ThreadLocal. I don't like the idea of Micrometer polluting threads with this flag. @jonatan-ivanov gave some ideas of other solutions in https://github.com/micrometer-metrics/micrometer/pull/3952#issuecomment-2259388434. Now that Scoped Values is GA in JDK 25, we could look at adding an implementation that uses that, though multi-release JARs have their issues. Alternatively, if Logback tracked the count for each logger and level, we could use that and avoid the need to do this at all.

belugabehr:
@shakuzen Thanks. I knew it felt weird that I was the first person to notice this, sorry I missed the dupes!

I candidly do not understand this notion of "polluting." As I understand the problem, this simple flag gets created for every thread, it's just a question of when the removal happens: after every invocation or once at the end of the life of the thread. For threads that are reused, this is a big benefit (I shared some JMH benchmarks regarding CPU and GC). For short-lived threads that are not reused, the cost is already present and the removal only adds a very small increase in the lifespan of this one static Boolean value.

I understand that this is perhaps not the most ideal solution in a future state, but we will not be moving to JDK 25 in any near future, so designing for that is out of scope of what I can myself work on.

belugabehr:
Alternatively, if you are concerned about backwards compatibility, I can add a feature flag to LogbackMetrics that toggles between the two implementations.

shakuzen:
> I candidly do not understand this notion of "polluting."

Every Thread has a ThreadLocalMap. Currently, we only keep an entry in that map while we are actively ignoring metrics (i.e. the scope of a call to `LogbackMetrics.ignoreMetrics(Runnable)`). The changes here would keep an entry in that map at all times after the first time `ignoreMetrics` is called. That's the "pollution" outside the scope of where we need this mechanism. A fair question is "what practical problem does that cause?" and other than the memory usage for one additional entry per Thread in the ThreadLocalMap, the answer may very well be "no problem." It feels like a code smell to me, but it is encapsulated at least. So maybe my concern about the pollution is not worth paying as much attention to as we have.

All of that said, @jonatan-ivanov has proposed an alternative implementation that completely eliminates the use of `ThreadLocal` in #6882. Please take a look at that and let us know what you think. I think that may make all of us happy and eliminate the source of various bugs/concerns people have about the implementation.

jonatan-ivanov:
Duplicate of https://github.com/micrometer-metrics/micrometer/pull/6882

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/6882

jonatan-ivanov:
@belugabehr Even if we went to another direction, thank you very much for the PR, the issue and your efforts. Please take a look at https://github.com/micrometer-metrics/micrometer/pull/6882 and let us know what you think.

PR #6847, opened 2025-11-03
Remove redundant sorted check before Arrays.sort()

belugabehr:
Removes the isSortedSet() pre-check before calling Arrays.sort() in both KeyValues and Tags classes. Modern JVM implementations make this optimization unnecessary and potentially counterproductive

Arrays.sort() uses TimSort which is already optimized for sorted arrays with O(n) best-case complexity. The pre-sort check adds unnecessary overhead for unsorted data (common case) without significant benefit.

* Simplified toKeyValues() and toTags() by removing isSortedSet() method and relying on TimSort's adaptive sorting behavior.
*  The isSortedSet(array, length) method was only ever called with `length = array.length`. The check if `(length > array.length)` could never be true, making this guard condition pointless.

### DIFF
--- a/micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
@@ -64,27 +64,6 @@ public final class KeyValues implements Iterable<KeyValue> {
     }
 
     /**
-     * Checks if the first {@code length} elements of the {@code keyValues} array form an
-     * ordered set of key-values.
-     * @param keyValues an array of key-values.
-     * @param length the number of elements to check.
-     * @return {@code true} if the first {@code length} elements of {@code keyValues} form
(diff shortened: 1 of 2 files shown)


belugabehr:
Going to close this. My performance results shows that, if the base/common case is a sorted array without dups, then the current `main` code is better, if the base/common case is un-sorted (nearly sorted being the biggest beneficiary) then removing the overhead of the pre-check is better. I don't know what the typical pattern is, but as a rule of thumb for any users of Micrometer, it's better to provided sorted data when possible.

shakuzen:
Thanks for checking and sharing the analysis.

> as a rule of thumb for any users of Micrometer, it's better to provided sorted data when possible.

There is some instrumentation where we take care to provide tags in sorted order, but we've stopped short of documenting this because we want the freedom to change the implementation.

Issue #6763, opened 2025-09-29
Test Support Improvements


etrandafir93:
I tried to use the  [micrometer-test](https://github.com/micrometer-metrics/micrometer/tree/main/micrometer-test) module - the [MeterRegistryAssert](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryAssert.java) to be more specific.

is it meant to be used internally by the other micrometer modules?  
Or is it meant to be publicly used by a wider audience?  

If it's dedicated to public usage, I have a few questions and enhancement requests - if they make sense

### 1. Tests in src/java

There area few tests in src/java, like [DefaultLongTaskTimerTest](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-test/src/main/java/io/micrometer/core/tck/DefaultLongTaskTimerTest.java). 

Was this done on purpose? Or could it be a mistake? 

### 2. Specialized Assertion 

_MeterRegistryAssert_ contains a few basic assertions, such as verifying if a specific meter exists. 
But we can’t really assert what the meter recorded. I was hoping for something more in line with AssertJ-style navigation, like:

``` java
MeterRegistryAssert.assertThat(meterRegistry)
  .hasTimerWithName("foo.time")
  .timer("foo.time")
    .hasCount(5)
    .hasTotalTime(...)
    // etc.
```
Would something like this provide any value?

### 3. Test Extension

In some cases (such as @SpringBootTests), I need to clear the meter registry before each test.

Would it make sense to have a dedicated mechanism for this? (maybe a junit extension or something similar to Spring Security's @WithMockUser?)

shakuzen:
> is it meant to be publicly used by a wider audience?

Yes.

> There area few tests in src/java, like [DefaultLongTaskTimerTest](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-test/src/main/java/io/micrometer/core/tck/DefaultLongTaskTimerTest.java).

DefaultLongTaskTimerTest is a mistake we're aware of. We simply haven't removed it out of an abundance of caution about breaking public API, even though no one should depend on that. I've opened #6779 for this.

Other tests are meant to be there since they are TCKs to be implemented in order to run a suite of tests against an arbitrary implementation of something (e.g. `MeterRegistry`).

> Would something like this provide any value?

[...]

PR #6870, opened 2025-11-09
Add AssertJ assertions for various Meters

etrandafir93:
Dedicated assertion classes for the meter types: CounterAssert, TimerAssert, and GaugeAssert. 

For all the other meter types, we can use _MeterAssert_, which provides a generic assertion functionality (such as checking the type or a specific measurement). 

All classes integrate with _MeterRegistryAssert_ for convenient access and include practical examples via tests and javadoc.        

Closes gh-6763

### DIFF
--- a/micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryAssert.java
+++ b/micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryAssert.java
@@ -16,10 +16,10 @@
 package io.micrometer.core.tck;
 
 import io.micrometer.common.KeyValues;
-import io.micrometer.common.util.assertions.CounterAssert;
-import io.micrometer.common.util.assertions.GaugeAssert;
-import io.micrometer.common.util.assertions.MeterAssert;
-import io.micrometer.common.util.assertions.TimerAssert;
+import io.micrometer.test.assertions.CounterAssert;
+import io.micrometer.test.assertions.GaugeAssert;
+import io.micrometer.test.assertions.MeterAssert;
+import io.micrometer.test.assertions.TimerAssert;
 import io.micrometer.core.instrument.*;
 import org.assertj.core.api.AbstractAssert;
 import org.assertj.core.api.Assertions;
@@ -425,7 +425,7 @@ public class MeterRegistryAssert extends AbstractAssert<MeterRegistryAssert, Met
      * @see CounterAssert
      */
     public CounterAssert counter(String name, Tag... tags) {
-        MeterAssert<?> meter = meter(name, tags).hasType(Meter.Type.COUNTER);
+        MeterAssert<?> meter = meter(name, tags).hasType(Counter.class);
 
         return CounterAssert.assertThat((Counter) meter.actual())
             .as("Counter with name <%s> and tags <%s>", name, tags)
@@ -454,8 +454,7 @@ public class MeterRegistryAssert extends AbstractAssert<MeterRegistryAssert, Met
      * @see TimerAssert
      */
     public TimerAssert timer(String name, Tag... tags) {
-        MeterAssert<?> meter = meter(name, tags).hasType(Meter.Type.TIMER);
-
+        MeterAssert<?> meter = meter(name, tags).hasType(Timer.class);
         return TimerAssert.assertThat((Timer) meter.actual())
             .as("Timer with name <%s> and tags <%s>", name, tags)
             .isNotNull();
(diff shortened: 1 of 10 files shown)


micrometer-test/src/main/java/io/micrometer/common/util/assertions/CounterAssert.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public static CounterAssert assertThat(Counter actual) {
+        return new CounterAssert(actual);
+    }
+
+    /**
+     * Creates a new instance of {@link CounterAssert}.
+     * @param actual the counter to assert on
+     */
+    public CounterAssert(Counter actual) {
```
shakuzen: I'm not familiar with best practice on custom Assert classes, but does it make sense to have this public constructor? Shouldn't users use the static `assertThat` method instead? Same question for other Assert classes in this PR.

micrometer-test/src/main/java/io/micrometer/common/util/assertions/package-info.java
```diff
@@ -13,17 +13,12 @@
...
- * @author Seungyong Hong
- * @author Jonatan Ivanov
+ * AssertJ-style assertion classes for Micrometer
+ * {@link io.micrometer.core.instrument.Meter}s.
  */
-public enum Cardinality {
-
-    HIGH, LOW
+@NullMarked
+package io.micrometer.common.util.assertions;
```
shakuzen: Unfortunately the micrometer-test module packages are a bit of a mess. It creates split package issues with micrometer-core and micrometer-commons. For backward compatibility reasons, we can't easily fix that until Micrometer 2.0. However, for new code like this, we should try to do better. I think it makes most sense for these to go in a package `io.micrometer.test.assert`. /cc @jonatan-ivanov

micrometer-test/src/main/java/io/micrometer/test/assertions/TimerAssert.java
```diff
@@ -0,0 +1,160 @@
...
+import io.micrometer.core.instrument.Timer;
+
+/**
+ * Assertion methods for {@link Timer}s.
+ * <p>
+ * To create a new instance of this class, invoke {@link TimerAssert#assertThat(Timer)} or
+ * use {@link io.micrometer.core.tck.MeterRegistryAssert#timer(String, Tag...)}.
+ *
+ * @author Emanuel Trandafir
+ */
```
shakuzen, line 37: Please add a `@since 1.17.0` JavaDoc to new public types in main code.

micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryAssert.java
```diff
@@ -371,4 +377,115 @@ private String allMetrics() {
...
+     *     .hasCount(0);
+     * </code></pre>
+     * @param name name of the counter
+     * @param tags tags to match
+     * @return a {@link CounterAssert} for the found counter
+     * @throws AssertionError if the counter is not found
+     * @see CounterAssert
+     */
+    public CounterAssert counter(String name, Tag... tags) {
+        MeterAssert<?> meter = meter(name, tags).hasType(Meter.Type.COUNTER);
```
shakuzen: It's better to check the Java type is an `instanceof` `Counter`. A custom `Meter` might have `Meter.Type` of `COUNTER` without implementing the `Counter` interface. The cast to `Counter` after will then fail. Similar feedback for the other methods added in this class.

shakuzen (changes requested):
Thanks for the pull request. This is awesome and I'm sure many users will enjoy the nicer API. I left some comments on things to change, but overall this is looking good.

micrometer-test/src/main/java/io/micrometer/common/util/assertions/CounterAssert.java
marcingrzejszczak: Definitely constructors should not be public and access should happen via statics

marcingrzejszczak:
Public classes should have a @since in their javadocs

micrometer-test/src/main/java/io/micrometer/common/util/assertions/CounterAssert.java
etrandafir93: fixed 👍

micrometer-test/src/main/java/io/micrometer/common/util/assertions/package-info.java
etrandafir93: I moved it to `io.micrometer.test.assertions` (assert is a reserved word)

micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryAssert.java
etrandafir93: fixed now

shakuzen (approved):
Looks good to me. Thank you for this. It looks like a nice addition.

Issue #6869, opened 2025-11-09
Return value nullability is incorrect on function wrapper methods


Mahoney:
**Describe the bug**
On trying to upgrade from 1.15.5 to 1.16.0 in a Kotlin project we are getting compile errors:

```kotlin
fun nullableReturn(): String? = TODO()

val timer: Timer = TODO()

try {
  timer.recordCallable {
    nullableReturn()
  }
} finally {
  timer.close()
}
```
> Argument type mismatch: actual type is 'Callable<String?>', but 'Callable<uninferred T! (of fun <T : Any> recordCallable)>!' was expected.

If I change the return type of `nullableReturn` to `String` from `String?` then it compiles, so it seems the Kotlin compiler thinks `<T>` is not nullable in `Timer.recordCallable`'s signature:
```java
<T> @NullUnmarked T recordCallable(Callable<T> f) throws Exception;
```

**Environment**

 - Micrometer version: 1.16.0
 - Micrometer registry: N/A
 - OS: All
 - Java version: 25

Mahoney:
I get a warning with Java, too:

```java
@Nullable
private String nullableReturn() {
  return null;
}

@Test
void copesWithNullReturn() {
  Timer timer = new MyTimer();
  String x = timer.record(() -> nullableReturn());
}
```
> Expression 'nullableReturn()' might evaluate to null but is returned by the method declared as @NullMarked

shakuzen:
Thanks for the report. I think we need to change the signature to:

```java
<T extends @Nullable Object> T record(Supplier<T> f);

<T extends @Nullable Object> T recordCallable(Callable<T> f) throws Exception;
```

We expected using `@NullUnmarked` had the same effect and semantically felt less hacky, but evidently the tooling does not see it as equivalent.

shakuzen:
1.16.1 snapshots should be published in about 10 minutes. Could you try those out and let us know if the issue is fixed for you?

Mahoney:
Yup, all good with 1.16.1-SNAPSHOT - thanks!

PR #6871, opened 2025-11-10
Fix return type nullability on wrapper methods

shakuzen:
We were using `@NullUnmarked` which we expected to work to mean the nullability is unknown for the return value. However, without changing the type parameter nullability, it was being considered non-null.

This updates all wrapping methods to use `T extends @Nullable Object` for the type parameter of the return value.

Resolves gh-6869

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
@@ -18,7 +18,6 @@ package io.micrometer.core.instrument;
 import io.micrometer.core.annotation.Timed;
 import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
 import io.micrometer.core.instrument.distribution.HistogramSupport;
-import org.jspecify.annotations.NullUnmarked;
 import org.jspecify.annotations.Nullable;
 
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


bendb-instacart:
JFYI this ended up being a breaking change for my Kotlin codebase.  I had an extension function like so:

```kt
fun <T> Timer.track(fn: () -> T): T = recordCallable(fn)!!
```

which, on updating to 1.16, failed compilation with `Argument type mismatch: actual type is 'Function0<T#1 (of fun <T> Timer.track)>', but 'Function0<uninferred T! (of fun <T : Any> recordCallable)>!' was expected.`.  The fix is trivial - changing the generic bounds from `<T>` to `<T : Any>` - but the change is breaking.

(this trivial function only exists to work around the fact that `recordCallable` is marked `@Nullable` - or _was_.  we probably don't need it anymore, with the updated nullability annotations here.)

shakuzen:
Thanks for letting us know. To be clear, is it unbroken with 1.16.1-SNAPSHOT which includes the change in this PR?

bendb-instacart:
Wait maybe I misunderstood the problem.  Was this PR included in 1.16.0?  If not, then I had it wrong, and it was something earlier (the jspecify change maybe?).  Our code seems to compile fine with 1.16.1-SNAPSHOT.

shakuzen:
This PR resolves https://github.com/micrometer-metrics/micrometer/issues/6869 which is included in the 1.16.1 milestone. This was fixing a nullability issue reported against 1.16.0 (#6869). Maybe that's effectively the same issue you were having. If it's fixed by this change, then that's good. This change will be in the upcoming 1.16.1 release.

bendb-instacart:
Wonderful.  Thank you, and sorry for the false alarm!

Issue #5317, opened 2024-07-29
Excessive TLAB memory allocations from MetricsTurboFilter 


drekbour:
**Describe the bug**

In the core loop of an otherwise low-garbage application, majority of heap allocations are reported as coming `MetricsTurboFilter` with the following stack trace linked to a filtered out debug logger

    ThreadLocal.ThreadLocalMap.Entry#Entry
    ThreadLocal$ThreadLocalMap.set(ThreadLocal, Object)
    ThreadLocal$ThreadLocalMap.access$100(ThreadLocal$ThreadLocalMap, ThreadLocal, Object)
    ThreadLocal.setInitialValue()
    ThreadLocal.get()
    MetricsTurboFilter.decide(Marker, Logger, Level, String, Object[], Throwable)
    TurboFilterList.getTurboFilterChainDecision(Marker, Logger, Level, String, Object[], Throwable)
    LoggerContext.getTurboFilterChainDecision_0_3OrMore(Marker, Logger, Level, String, Object[], Throwable)
    Logger.filterAndLog_0_Or3Plus(String, Marker, Level, String, Object[], Throwable)
    Logger.debug(String)

**Environment**
 - Micrometer version 1.9.12
 - Micrometer registry Spring Boot configured
 - OS: RHEL
 - Java version: 8, 21

**To Reproduce**
Profile a contrived hot loop like this in a spring boot application
```java
  while(true) { log.debug(Foo"); }
```
Run briefly a desktop this reported allocations of 
* 500MB/s for the lambda (probably compiled away to zero in real application) 
* 100MB/s for the `setInitialAccess` calls

**Expected behavior**
Filtered call to `log.debug("Foo")` is expected to be lightning fast in java ecosystem. 

Avoid memory allocations where possible - 
```java
ignoreMetrics.set(true);
try {
    r.run();
} finally {
    ignoreMetrics.remove(); // << change this to ignoreMetics.set(Boolean.FALSE)
}
```

**Additional context**
If there's some argument about threadlocals being a memory leak, I would point out that this adds just a few weak-ref bytes per thread. If an app is discarding threads at a rate that _this_ allocation is a problem, it has far larger problems anyway.

See #3891 for related changes to avoid ThreadLocal access where not required

jonatan-ivanov:
Thanks for the issue!
Micrometer 1.9 (and Spring Boot 2.7) is (are) not OSS supported anymore, please upgrade, see: https://micrometer.io/support/

[...]

PR #6882, opened 2025-11-11
Remove ThreadLocal usage from LogbackMetrics

jonatan-ivanov:
Before this change, LogbackMetrics used Counter to keep track of the number of emitted log events by level. This caused issues since there are cases where incrementing a Counter can result in emitting log events. Because of this, the MetricsTurboFilter in LogbackMetrics had a chance to get into an infinite loop since it was triggered by emitting log events and it incremented a Counter that also emitted a log event which triggered MetricsTurboFilter again...

The scenario was the following:
1. Somebody emits a log event (e.g.: LOGGER.info("test");)
2. MetricsTurboFilter gets notified about it and increments a Counter
3. Incrementing the Counter triggers emitting another log event
4. MetricsTurboFilter gets notified about it and increments a Counter ...

To fix this issue, a ThreadLocal was used to ignore some of the log events (the ones that were emitted by incrementing a Counter).

This change removes the ThreadLocal usage and instead of using Counter, it keeps track of the number of emitted log events using a LongAdder. Then it uses this LongAdder in a FunctionCounter. It assumes that LongAdder does not emit any log events that will end up triggering MetricsTurboFilter again.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/LogbackMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/LogbackMetricsTest.java
@@ -19,10 +19,7 @@ import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.LoggerContext;
 import io.micrometer.core.Issue;
-import io.micrometer.core.instrument.Counter;
-import io.micrometer.core.instrument.Meter;
-import io.micrometer.core.instrument.MeterRegistry;
-import io.micrometer.core.instrument.MockClock;
+import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.cumulative.CumulativeCounter;
 import io.micrometer.core.instrument.simple.SimpleConfig;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
@@ -31,8 +28,6 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.slf4j.LoggerFactory;
-
-import java.util.concurrent.atomic.AtomicInteger;
 
 import static java.util.Collections.emptyList;
 import static org.assertj.core.api.Assertions.assertThat;
@@ -62,7 +57,7 @@ class LogbackMetricsTest {
 
     @Test
     void logbackLevelMetrics() {
-        assertThat(registry.get("logback.events").counter().count()).isEqualTo(0.0);
+        assertThat(registry.get("logback.events").functionCounter().count()).isEqualTo(0.0);
 
         // tag::example[]
         logger.setLevel(Level.INFO);
@@ -71,8 +66,8 @@ class LogbackMetricsTest {
         logger.error("error");
         logger.debug("debug"); // shouldn't record a metric
 
-        assertThat(registry.get("logback.events").tags("level", "warn").counter().count()).isEqualTo(1.0);
-        assertThat(registry.get("logback.events").tags("level", "debug").counter().count()).isEqualTo(0.0);
+        assertThat(registry.get("logback.events").tags("level", "warn").functionCounter().count()).isEqualTo(1.0);
+        assertThat(registry.get("logback.events").tags("level", "debug").functionCounter().count()).isEqualTo(0.0);
         // end::example[]
     }
 
@@ -81,7 +76,7 @@ class LogbackMetricsTest {
     void isLevelEnabledDoesntContributeToCounts() {
         logger.isErrorEnabled();
 
-        assertThat(registry.get("logback.events").tags("level", "error").counter().count()).isEqualTo(0.0);
+        assertThat(registry.get("logback.events").tags("level", "error").functionCounter().count()).isEqualTo(0.0);
     }
 
     @Issue("#411, #3623")
@@ -92,7 +87,7 @@ class LogbackMetricsTest {
             logbackMetrics.bindTo(registry);
             LoggerFactory.getLogger("test").info("This should be counted once");
         }
-        assertThat(registry.get("logback.events").tags("level", "info").counter().count()).isOne();
+        assertThat(registry.get("logback.events").tags("level", "info").functionCounter().count()).isOne();
     }
 
     @Issue("#421")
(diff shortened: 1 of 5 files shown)


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/LogbackMetricsTest.java
```diff
@@ -62,7 +57,7 @@ void closeLogbackMetrics() {
 
     @Test
     void logbackLevelMetrics() {
-        assertThat(registry.get("logback.events").counter().count()).isEqualTo(0.0);
+        assertThat(registry.get("logback.events").functionCounter().count()).isEqualTo(0.0);
```
shakuzen, line 60: Is there any difference in what gets published with a function counter vs counter, e.g. with the Prometheus registry? I don't remember off the top of my head.

shakuzen (commented):
I really like getting rid of the ThreadLocal-based implementation. Other than my question about any compatibility issues of the switch from Counter to FunctionCounter, I think the only other thing to check is benchmarking.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/logging/LogbackMetricsTest.java
jonatan-ivanov: I assumed no, just checked it with Prometheus: no difference.
That would be my expectation from publishing perspective with every registry and if they would be different, I would be suspicious that it's a bug. In case the difference is justified we can consider "re-introducing" a "fallback" (and deprecated) version with `ThreadLocal` and `Counter` for the users who need it and we can remove it after a few minor releases but I would only do that if somebody reports this as an issue.

jonatan-ivanov:
Benchmarking:
I should have put it into the description, I ran `LogbackMetricsBenchmark` which I also extended with an extra benchmark that used virtual threads and the "ignore" feature. Performance was ~same/slightly better than the original implementation using `ThreadLocal`.

jonatan-ivanov:
I wanted to collect the connected issues here we accumulated over the years:
- https://github.com/micrometer-metrics/micrometer/pull/6845
- https://github.com/micrometer-metrics/micrometer/issues/6837
- https://github.com/micrometer-metrics/micrometer/pull/3952
- https://github.com/micrometer-metrics/micrometer/issues/3891
- https://github.com/micrometer-metrics/micrometer/issues/5317
- https://github.com/micrometer-metrics/micrometer/issues/3623
- https://github.com/micrometer-metrics/micrometer/pull/3627
- https://github.com/micrometer-metrics/micrometer/issues/411
- https://github.com/micrometer-metrics/micrometer/issues/2270
- https://github.com/micrometer-metrics/micrometer/pull/2271
- https://github.com/spring-projects/spring-boot/issues/33971

I think if we merge this, we should close:
- https://github.com/micrometer-metrics/micrometer/pull/6845
- https://github.com/micrometer-metrics/micrometer/issues/6837
- https://github.com/micrometer-metrics/micrometer/pull/3952

jonatan-ivanov:
@belugabehr, @spmason What do you think about the change?

spmason:
I've moved on from the project where I needed this, but removing the need for the thread local sounds like a much more elegant solution than needing to hack around it!

shakuzen (approved):

jonatan-ivanov:
After fixing `LogbackMetricsBenchmark `(https://github.com/micrometer-metrics/micrometer/pull/6891), it seems that the performance improvement is visible especially in heap allocation rate.

Summary before/after:
```
Benchmark                        (registerLogbackMetrics)    Mode     Cnt       Score   Error   Units
logSomething                                         true  sample  606549      44.339 ± 0.542   ns/op
logSomething:gc.alloc.rate                           true  sample       2     947.088          MB/sec
logSomething:gc.alloc.rate.norm                      true  sample       2      32.001            B/op
logSomething:gc.count                                true  sample       2      42.000          counts
logSomething:gc.time                                 true  sample       2      19.000              ms

logSomething                                         true  sample  687434      39.268 ± 0.333   ns/op
logSomething:gc.alloc.rate                           true  sample       2       0.020          MB/sec
logSomething:gc.alloc.rate.norm                      true  sample       2       0.001            B/op
logSomething:gc.count                                true  sample       2         ≈ 0          counts
(no gc.time)
```
(`947 MB/sec` vs. `0.020 MB/sec`)

<details>
<summary>Click to see detailed reports</summary>

## Sample Benchmark Before the change
See build scan: https://ge.micrometer.io/s/vtncgvjqcytes/console-log?page=1#L319
```
Benchmark                                                (registerLogbackMetrics)    Mode     Cnt       Score   Error   Units
LogbackMetricsBenchmark.logSomething                                        false  sample  636642      49.055 ± 4.184   ns/op
LogbackMetricsBenchmark.logSomething:gc.alloc.rate                          false  sample       2       0.026          MB/sec
LogbackMetricsBenchmark.logSomething:gc.alloc.rate.norm                     false  sample       2       0.001            B/op
LogbackMetricsBenchmark.logSomething:gc.count                               false  sample       2         ≈ 0          counts
LogbackMetricsBenchmark.logSomething:p0.00                                  false  sample                 ≈ 0           ns/op
LogbackMetricsBenchmark.logSomething:p0.50                                  false  sample              42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.90                                  false  sample              42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.95                                  false  sample              42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.99                                  false  sample              84.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.999                                 false  sample             250.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.9999                                false  sample           15013.826           ns/op
LogbackMetricsBenchmark.logSomething:p1.00                                  false  sample          508928.000           ns/op
LogbackMetricsBenchmark.logSomething                                         true  sample  606549      44.339 ± 0.542   ns/op
LogbackMetricsBenchmark.logSomething:gc.alloc.rate                           true  sample       2     947.088          MB/sec
LogbackMetricsBenchmark.logSomething:gc.alloc.rate.norm                      true  sample       2      32.001            B/op
LogbackMetricsBenchmark.logSomething:gc.count                                true  sample       2      42.000          counts
LogbackMetricsBenchmark.logSomething:gc.time                                 true  sample       2      19.000              ms
LogbackMetricsBenchmark.logSomething:p0.00                                   true  sample                 ≈ 0           ns/op
LogbackMetricsBenchmark.logSomething:p0.50                                   true  sample              42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.90                                   true  sample              42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.95                                   true  sample              42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.99                                   true  sample              84.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.999                                  true  sample             167.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.9999                                 true  sample            4496.000           ns/op
LogbackMetricsBenchmark.logSomething:p1.00                                   true  sample           35456.000           ns/op
```

## Sample Benchmark After the change
See build scan: https://ge.micrometer.io/s/p6kwuwsiml5aq/console-log?page=1#L318
```
Benchmark                                                (registerLogbackMetrics)    Mode     Cnt      Score   Error   Units
LogbackMetricsBenchmark.logSomething                                        false  sample  705411     38.842 ± 0.259   ns/op
LogbackMetricsBenchmark.logSomething:gc.alloc.rate                          false  sample       2      0.016          MB/sec
LogbackMetricsBenchmark.logSomething:gc.alloc.rate.norm                     false  sample       2     ≈ 10⁻³            B/op
LogbackMetricsBenchmark.logSomething:gc.count                               false  sample       2        ≈ 0          counts
LogbackMetricsBenchmark.logSomething:p0.00                                  false  sample                ≈ 0           ns/op
LogbackMetricsBenchmark.logSomething:p0.50                                  false  sample             42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.90                                  false  sample             42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.95                                  false  sample             42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.99                                  false  sample             42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.999                                 false  sample             84.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.9999                                false  sample            685.811           ns/op
LogbackMetricsBenchmark.logSomething:p1.00                                  false  sample          19200.000           ns/op
LogbackMetricsBenchmark.logSomething                                         true  sample  687434     39.268 ± 0.333   ns/op
LogbackMetricsBenchmark.logSomething:gc.alloc.rate                           true  sample       2      0.020          MB/sec
LogbackMetricsBenchmark.logSomething:gc.alloc.rate.norm                      true  sample       2      0.001            B/op
LogbackMetricsBenchmark.logSomething:gc.count                                true  sample       2        ≈ 0          counts
LogbackMetricsBenchmark.logSomething:p0.00                                   true  sample                ≈ 0           ns/op
LogbackMetricsBenchmark.logSomething:p0.50                                   true  sample             42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.90                                   true  sample             42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.95                                   true  sample             42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.99                                   true  sample             42.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.999                                  true  sample             84.000           ns/op
LogbackMetricsBenchmark.logSomething:p0.9999                                 true  sample            584.000           ns/op
LogbackMetricsBenchmark.logSomething:p1.00                                   true  sample          35840.000           ns/op
```
</details>

belugabehr:
@jonatan-ivanov 

Assuming the following is true:

> Should not be needed anymore since {@link LogbackMetrics} records logback events asynchronously and should not get into an infinite loop.

Then this is a great solution and is much appreciated!

jonatan-ivanov:
@belugabehr Feel free to try it out, it's available in `1.17.0-SNAPSHOT`.

Issue #6863, opened 2025-11-07
LatencyUtils 2.0.3 is 10+ years old, with a few open issues on Github - problem or not?


the-vj:
LatencyUtils library 2.0.3 is used even by latest Micrometer 1.16.0: 

- https://github.com/LatencyUtils/LatencyUtils 

The last release was 10+ years ago and there are some open issues reported, some of which sound a bit worrying.

Is the Micrometer team aware of those and is it a conscious conclusion the library does not pose any problems to critical, widely used and very much alive Micrometer framework in practice, or is there something that should be checked?

shakuzen:
The vast majority of Micrometer users these days probably have no runtime usage of LatencyUtils because we changed the default from the PauseDetector that uses LatencyUtils to a no-op PauseDetector in version 1.0.10 (see https://github.com/micrometer-metrics/micrometer/issues/1324). So I'm not sure we can speak to the stability of LatencyUtils in general beyond what our users report to us. Unless a user explicitly configures a PauseDetector, LatencyUtils will not be used at runtime. Given that, you can exclude the runtime dependency on LatencyUtils if you are not configuring a PauseDetector with supported versions of Micrometer. See, for example, [this previous comment](https://github.com/micrometer-metrics/micrometer/issues/1599#issuecomment-1174757661) and [this section](https://docs.micrometer.io/micrometer/reference/concepts.html#concepts-dependencies) of the documentation.

We could consider removing the runtime dependency and instead documenting that users who want to use the pause detector that depends on LatencyUtils should add the runtime dependency themselves. In some sense that's a breaking change, but I suspect the number of users affected would be minimal and they can fix it without any changes on the Micrometer side. It would get rid of a transitive dependency most users don't need. In a future with a new major version, we could consider more changes to isolate LatencyUtils dependent code from the core library.

I'm not sure if that answers your question. If you are using the default for PauseDetector in Micrometer, you are not using LatencyUtils and so it's perhaps not a concern for you.

the-vj:
thanks a lot for such a detailed response.

based on your explanation, i think that in our particular case, we could confidently exclude the runtime dependency, as we do not  use those capabilities in our product.

i agree that removal from your side would be a breaking change - btw, the commitment of the Micrometer community to backward compatibility in patch releases is very, very appreciated even if it means that in some cases more time is needed to get certain topics fully resolved.

[...]

PR #6883, opened 2025-11-12
Remove transitive runtime dependency on LatencyUtils

shakuzen:
Before this change, users depending on micrometer-core would get a transitive runtime dependency on LatencyUtils. After this change, that will no longer happen, reducing the number of dependencies brought in by micrometer-core.

For a very long time now, pause detection has been disabled by default, and accordingly users do not need the LatencyUtils dependency at runtime unless they enable pause detection in their application. This change checks if the ClockDriftPauseDetector is configured but LatencyUtils is not available. It will log a warning when this happens so users are aware while avoiding throwing a CNFE. Updates the documentation accordingly.

See #6863

### DIFF
--- a/docs/modules/ROOT/pages/concepts/timers.adoc
+++ b/docs/modules/ROOT/pages/concepts/timers.adoc
@@ -202,7 +202,7 @@ Micrometer supports two pause detector implementations: a clock-drift based dete
 
 The clock-drift based detector has a configurable sleep interval and pause threshold. CPU consumption is inversely proportional to `sleepInterval`, as is pause detection accuracy. 100ms for both values is a reasonable default to offer decent detection of long pause events while consuming a negligible amount of CPU time.
 
-NOTE: You need to add LatencyUtils to your runtime classpath to use the ClockDriftPauseDetector.
+NOTE: You need to add LatencyUtils(`org.latencyutils:LatencyUtils`) to your runtime classpath to use the `ClockDriftPauseDetector`.
 Without it on the runtime classpath, a warning will be logged and no pause detection will happen.
 
 You can customize the pause detector as follows:


docs/modules/ROOT/pages/concepts.adoc
```diff
@@ -13,8 +13,10 @@ For better understanding of the differences among these different types of syste
 [[concepts-dependencies]]
 == Dependencies
 
-The `micrometer-core` module aims to have minimal dependencies. It does not require any third-party (non-Micrometer) dependencies on the classpath at compile time for applications that use Micrometer.
+The `micrometer-core` module aims to have minimal dependencies.
+It does not require any third-party (non-Micrometer) dependencies on the classpath at compile time for applications that use Micrometer, except for https://jspecify.dev/[JSpecify].
```
shakuzen, line 17: This is an unrelated change I noticed at the time. This could and probably should be made separately and back-ported.

micrometer-core/src/test/java/io/micrometer/core/instrument/MissingLatencyUtilsTest.java
```diff
@@ -45,4 +52,16 @@ void createRecordCloseTimerWithoutPauseDetector() {
...
 
+    @Test
+    @Issue("6863")
+    void timerWithPauseDetector_noExceptionThrown(CapturedOutput output) {
+        registry.config().pauseDetector(new ClockDriftPauseDetector(Duration.ofMillis(100), Duration.ofMillis(100)));
+        assertThatCode(() -> {
+            Timer timer = Timer.builder("my.timer").register(registry);
+            timer.record(1, TimeUnit.MILLISECONDS);
+            timer.close();
+        }).doesNotThrowAnyException();
```
shakuzen, line 63: Before the changes in AbstractTimer, this test would throw a CFNE when trying to instantiate `SimplePauseDetector` if LatencyUtils is not on the runtime classpath.

docs/modules/ROOT/pages/concepts/timers.adoc
```diff
@@ -198,10 +198,13 @@ WARNING: This FunctionTimer has been already registered (MeterId{name='my.ft', t
...
 
-Micrometer supports two pause detector implementations: a clock-drift based detector and a no-op detector. Before Micrometer 1.0.10/1.1.4/1.2.0, a clock-drift detector was configured by default to report as-accurate-as-possible metrics without further configuration. Since 1.0.10/1.1.4/1.2.0, the no-op detector is configured by default, but the clock-drift detector can be configured as shown in the next example.
+Micrometer supports two pause detector implementations: a clock-drift based detector and a no-op detector. The no-op detector is configured by default, but the clock-drift detector can be configured as shown in the next example.
 
 The clock-drift based detector has a configurable sleep interval and pause threshold. CPU consumption is inversely proportional to `sleepInterval`, as is pause detection accuracy. 100ms for both values is a reasonable default to offer decent detection of long pause events while consuming a negligible amount of CPU time.
 
+NOTE: You need to add LatencyUtils to your runtime classpath to use the ClockDriftPauseDetector.
```
jonatan-ivanov: I think it would be nice provide the dependency coordinates in the docs now that `micrometer-core` won't bring them in:
```suggestion
NOTE: You need to add LatencyUtils(`org.latencyutils:LatencyUtils`) to your runtime classpath to use the `ClockDriftPauseDetector`.
```

docs/modules/ROOT/pages/concepts.adoc
```diff
@@ -13,8 +13,10 @@ For better understanding of the differences among these different types of syste
...
 
-Use of the xref:concepts/timers.adoc#pause-detection[pause detection] feature requires the https://github.com/LatencyUtils/LatencyUtils[LatencyUtils] dependency to be available on the classpath at runtime. If your application does not use the pause detection feature, you can exclude LatencyUtils from your runtime classpath.
+Use of the xref:concepts/timers.adoc#pause-detection[pause detection] feature requires the https://github.com/LatencyUtils/LatencyUtils[LatencyUtils] dependency to be available on the classpath at runtime.
```
jonatan-ivanov: Dependency coordinates like above:
```suggestion
Use of the xref:concepts/timers.adoc#pause-detection[pause detection] feature requires the https://github.com/LatencyUtils/LatencyUtils[LatencyUtils(`org.latencyutils:LatencyUtils`)] dependency to be available on the classpath at runtime.
```

micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
```diff
@@ -134,6 +136,17 @@ private void initPauseDetector(PauseDetector pauseDetectorType) {
         if (pauseDetectorType instanceof NoPauseDetector) {
             return;
         }
+        if (pauseDetectorType instanceof ClockDriftPauseDetector) {
+            try {
+                Class.forName("org.LatencyUtils.SimplePauseDetector");
+            }
+            catch (ClassNotFoundException e) {
+                pauseDetectorCNFELogger.log(
+                        "LatencyUtils is not on the runtime classpath but a ClockDriftPauseDetector which requires it is configured. Pause detection is disabled.",
```
jonatan-ivanov: How about also explicitly telling the users what to do?
```suggestion
                        "LatencyUtils is not on the runtime classpath but a ClockDriftPauseDetector which requires it is configured. Pause detection is disabled, please add LatencyUtils to the classpath if you want to enable it.",
```

shakuzen:
I addressed the review comments. Let me know if there's anything else before this is ready to merge.

PR #5688, opened 2024-11-16
Warn about gauge re-registration

jonatan-ivanov:
See https://github.com/micrometer-metrics/micrometer/issues/5616 and https://github.com/micrometer-metrics/micrometer/pull/5617

I added the warning log entries and benchmarks since the extra logging is on the hot path.
The benchmarks don't test the case where the `Gauge` re-registration happens indirectly (as the result of a `MeterFilter` transformation) but since the direct and the indirect cases call the same method, I think testing only the direct case is ok.

I ran the benchmarks a couple of times but they were not as stable on my laptop as I liked, the difference between them was sometimes bigger between two identical executions (e.g. without logging) than between the two different ones (with vs. without logging).

So based on this either the benchmarks are wrong or the "noise" is bigger than the difference between the two scenarios I tested. 

## Before logging
```
Benchmark                                                   Mode      Cnt     Score   Error  Units
GaugeBenchmark.baseline                                   sample  1874878     0.065 ± 0.003  us/op
GaugeBenchmark.baseline:p0.00                             sample              0.001          us/op
GaugeBenchmark.baseline:p0.50                             sample              0.041          us/op
GaugeBenchmark.baseline:p0.90                             sample              0.048          us/op
GaugeBenchmark.baseline:p0.95                             sample              0.049          us/op
GaugeBenchmark.baseline:p0.99                             sample              0.084          us/op
GaugeBenchmark.baseline:p0.999                            sample              2.295          us/op
GaugeBenchmark.baseline:p0.9999                           sample             32.624          us/op
GaugeBenchmark.baseline:p1.00                             sample           1249.280          us/op
GaugeBenchmark.gaugeReRegistrationWithBuilder             sample  1314563     0.085 ± 0.007  us/op
GaugeBenchmark.gaugeReRegistrationWithBuilder:p0.00       sample              0.001          us/op
GaugeBenchmark.gaugeReRegistrationWithBuilder:p0.50       sample              0.046          us/op
GaugeBenchmark.gaugeReRegistrationWithBuilder:p0.90       sample              0.054          us/op
GaugeBenchmark.gaugeReRegistrationWithBuilder:p0.95       sample              0.056          us/op
GaugeBenchmark.gaugeReRegistrationWithBuilder:p0.99       sample              0.082          us/op
GaugeBenchmark.gaugeReRegistrationWithBuilder:p0.999      sample             15.680          us/op
GaugeBenchmark.gaugeReRegistrationWithBuilder:p0.9999     sample             51.722          us/op
[...]

PR #6891, opened 2025-11-13
Don't filter log events in LogbackMetricsBenchmark

jonatan-ivanov:
In `LogbackMetricsBenchmark` we test `LogbackMetrics` by emitting log events on `INFO` level. In #5688 a `logback.xml` was added that unfortunately set the lowest log level to `WARN` and broke `LogbackMetricsBenchmark` since the emitted `INFO` events never reached `MetricsTurboFilter` in `LogbackMetrics` and no metrics were recorded.

Another issue with `LogbackMetricsBenchmark` is log events going to `stdout` even when `logback.xml` in not present. To solve this we can use Logback's `NOPAppender` which will not output anything but lets log events reach `MetricsTurboFilter` in `LogbackMetrics` and record metrics.

See #5688


jonatan-ivanov:
Benchmark for filtered log events: #6894

PR #6895, opened 2025-11-15
Polish gh-6682

izeye:
This PR polishes the changes made in gh-6682 a bit.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/micrometer/MicrometerJvmThreadMeterConventions.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/micrometer/MicrometerJvmThreadMeterConventions.java
@@ -32,11 +32,8 @@ public class MicrometerJvmThreadMeterConventions implements JvmThreadMeterConven
 
     private final Tags extraTags;
 
-    private final MeterConvention<Thread.State> threadCountConvention;
-
     public MicrometerJvmThreadMeterConventions(Tags extraTags) {
         this.extraTags = extraTags;
-        this.threadCountConvention = new SimpleMeterConvention<>("jvm.threads.states", this::getStateTag);
     }
 
     private Tags getStateTag(Thread.State state) {
@@ -49,7 +46,7 @@ public class MicrometerJvmThreadMeterConventions implements JvmThreadMeterConven
 
     @Override
     public MeterConvention<Thread.State> threadCountConvention() {
-        return threadCountConvention;
+        return new SimpleMeterConvention<>("jvm.threads.states", this::getStateTag);
     }
 
 }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmMemoryMetrics.java
@@ -39,7 +39,7 @@ import static io.micrometer.core.instrument.binder.jvm.JvmMemory.getUsageValue;
  */
 public class JvmMemoryMetrics implements MeterBinder {
 
-    private final Tags tags;
+    private final Tags extraTags;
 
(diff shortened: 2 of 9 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/micrometer/MicrometerJvmThreadMeterConventions.java
```diff
@@ -49,7 +46,7 @@ protected Tags getCommonTags() {
 
     @Override
     public MeterConvention<Thread.State> threadCountConvention() {
-        return threadCountConvention;
+        return new SimpleMeterConvention<>("jvm.threads.states", this::getStateTag);
```
shakuzen, line 49: If I remember correctly, I was trying to avoid allocating a new convention so this method could be called any number of times without causing allocation every time to make an equivalent convention. What was the motivation to change it to this?

izeye: @shakuzen Thanks for the feedback!

The reason was to try to resolve some inconsistency as other convention methods create convention instances on their invocation as follows:

https://github.com/micrometer-metrics/micrometer/blob/ba6d18fcfc782862bddcc49046b605a557902c82/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/convention/micrometer/MicrometerJvmClassLoadingMeterConventions.java#L45-L58

And there's only one invocation for the `threadCountConvention()` here:

https://github.com/micrometer-metrics/micrometer/blob/ba6d18fcfc782862bddcc49046b605a557902c82/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java#L104-L111

I don't have a strong opinion on this, but I just wanted to make them consistent if possible.

jonatan-ivanov: Should we make it consistent the other way?
(Init everything in the ctor and return the same instance in the accessor)

shakuzen: I had this comment on it in the original PR: https://github.com/micrometer-metrics/micrometer/pull/6682#discussion_r2315417920.
We could make things consistent the other way as Jonatan suggests, but that's wasteful in another way - maintaining a reference to the convention that is only needed once when binding.

There's additional complexity, which is while we know how the conventions are used in our own instrumentation, it's possible they are used in third-party instrumentation as well where we don't know whether they are used once on binding or repeatedly to record metrics. So the optimal way is not necessarily clear.

Given that, I'm not sure what we should do here. Thoughts?

izeye: If a user needs to call it multiple times for some reason, they might not expect it to create a new instance and need to cache it themselves. If that's an expected use case, keeping them in fields might be okay as they are lightweight objects anyway.

shakuzen: > If that's an expected use case

I think this is the crux of the problem - determining whether it's an expected use case in each case or not. However, since this is an implementation detail we can change without changing the API, maybe we only need to worry about not returning a new instance where we know we'll call it multiple/many times in our instrumentation.

As you pointed out, the thread count convention method doesn't get called multiple times (and we don't generally expect a reason to have multiple instances of JvmThreadMetrics binders). So I think your changes there are fine and nothing more needs to be updated. Thanks for taking the long journey to arrive at everything is great how you did it in the first place 😄

Issue #6844, opened 2025-10-31
Make publishing average configurable


jonatan-ivanov:
See https://github.com/micrometer-metrics/micrometer/issues/6843, in `1.x` we can keep publish average but we can let the users disable it with a config option.

jaxxnitt:
I would like to work on this.

Just to confirm the expected behavior for 1.x:

keep the current default (publish averages),

add a new config option in DatadogConfig to enable/disable publishing averages,

and use that flag in DatadogMeterRegistry to control whether .avg is emitted.

Then the default can be flipped in 2.x as discussed in #6843.
Is this the intended direction?

jonatan-ivanov:
@jaxxnitt Thank you for the offer, your description is correct, please feel free to open a PR.

jaxxnitt:
I've opened a PR implementing this: #6903 

It adds the `publishAverage` config option with the current default (enabled), and updates Timer/FunctionTimer/Summary emission accordingly. Happy to adjust anything if needed.

jonatan-ivanov:
Thank you!
In the meantime @shakuzen noticed that I missed one detail in your description: this should work for all of the registries that publish avg not only Datadog. Would you mind updating the PR?

jaxxnitt:
Thanks for the review, @jonatan-ivanov! 

Before I update the PR:
You mentioned this should work for all registries that publish .avg. Since only push-based registries emit average metrics, and those share PushRegistryConfig as the common config interface (whether or not they pass through StepRegistryConfig), should the new publishAverage() flag be added to PushRegistryConfig?
Just want to confirm the correct interface before I refactor the PR.

shakuzen:
That's a good question @jaxxnitt. I haven't given a lot of thought to this yet, but I think it could make sense for the option to be on `DistributionStatisticsConfig` similar to how that is used to control whether client-side percentiles or a histogram is tracked (and published). It could also make sense on the registry config (e.g. `PushRegistryConfig`) as you suggested. One trade-off is that if it is on the registry config, it can only be controlled at the registry level, whereas if it is on the `DistributionStatisticsConfig`, it can be controlled per meter. I imagine most users want it on or off for the whole registry, but it's also logical and in line with other options on `DistributionStatisticsConfig` for it to be there. I would lean toward putting it there, but as I said, I haven't spent enough time thinking about all the implications, so if you have some thoughts on which approach is better and why, please share.

jaxxnitt:
[...]

Issue #6843, opened 2025-10-31
Remove publishing average


jonatan-ivanov:
Remove publishing `avg` for those registries where this is applicable: they currently publish `avg` and it can be removed (the protocol lets us), e.g.: Datadog.

Averages are usually not very useful in terms of latency and can be the source of confusion, they are also not aggregatable. Removing them would also save cost but this is a breaking change.

shakuzen:
We could also consider a step before this which would be to flip the default from publishing average to not publishing it, but that would also be a breaking change, albeit with a fallback to the prior behavior.

jonatan-ivanov:
We can consider doing that in https://github.com/micrometer-metrics/micrometer/issues/6844.

PR #6903, opened 2025-11-18
Add publishAverage config option to Datadog registry 

jaxxnitt:
This PR implements the enhancement discussed in #6844.

**Summary**

Adds a new publishAverage() boolean config option to DatadogConfig (default: true).

Updates DatadogMeterRegistry to conditionally publish .avg metrics for:

Timer

FunctionTimer

DistributionSummary

Ensures metadata for average metrics is also controlled by this flag.

Keeps existing 1.x behavior by default, while allowing users to disable average publishing if desired.

This aligns with the intended direction for 1.x and prepares the way for flipping the default in 2.x as discussed in #6843.

Feel free to suggest changes.


implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
```diff
@@ -76,4 +76,8 @@ default Validated<?> validate() {
                 checkRequired("uri", DatadogConfig::uri));
     }
 
+    default boolean publishAverage() {
```
jonatan-ivanov: See https://github.com/micrometer-metrics/micrometer/issues/6844#issuecomment-3549762243, this should work for all the registries that publish avg.

PR #6914, opened 2025-11-22
Capitalize log message of PushMeterRegistry

rethab:
When spring boot starts up, all log message are capitalized except this one ;-)


jonatan-ivanov:
Thank you for the PR! I think we should change this but not necessarily because of log entries in Spring Boot (Micrometer can be used with or without any framework, not only Spring).

PR #6920, opened 2025-11-25
Metrics not collected after ScheduledExecutorService recreation

slovdahl:
Related to #5366, similar to #5367


jonatan-ivanov:
Thank you for fixing this!

Issue #4602, opened 2024-01-19
MongoMetricsCommandListenerTest.shouldCreateFailedCommandMetricWithCustomSettings() seems flaky


izeye:
It failed on CI as follows:

```
MongoMetricsCommandListenerTest > shouldCreateFailedCommandMetricWithCustomSettings() FAILED
    io.micrometer.core.instrument.search.MeterNotFoundException: Unable to find a meter that matches all the requirements at once. Here's what was found:
       OK: A meter with name 'mongodb.driver.commands' was found.
       OK: A meter with type timer was found.
       FAIL: No meters have a tag 'cluster.id' with value '65aa84157856c446e440099e'. The only value found was '65aa84157856c446e440099f'.
       OK: A meter with name 'mongodb.driver.commands' has a tag 'collection' with value 'testCol'.
       OK: A meter with name 'mongodb.driver.commands' has a tag 'command' with value 'dropIndexes'.
       OK: A meter with name 'mongodb.driver.commands' has a tag 'mongoz' with value '5150'.
       OK: A meter with name 'mongodb.driver.commands' has a tag 'server.address' with value 'localhost:32781'.
       OK: A meter with name 'mongodb.driver.commands' has a tag 'status' with value 'FAILED'.
        at app//io.micrometer.core.instrument.search.MeterNotFoundException$FromRequiredSearch.build(MeterNotFoundException.java:245)
        at app//io.micrometer.core.instrument.search.MeterNotFoundException$FromRequiredSearch.access$100(MeterNotFoundException.java:49)
        at app//io.micrometer.core.instrument.search.MeterNotFoundException.forSearch(MeterNotFoundException.java:46)
        at app//io.micrometer.core.instrument.search.RequiredSearch.lambda$getOne$1(RequiredSearch.java:199)
        at java.base@17.0.8.1/java.util.Optional.orElseThrow(Optional.java:403)
        at app//io.micrometer.core.instrument.search.RequiredSearch.getOne(RequiredSearch.java:199)
        at app//io.micrometer.core.instrument.search.RequiredSearch.timer(RequiredSearch.java:120)
        at app//io.micrometer.core.instrument.binder.mongodb.MongoMetricsCommandListenerTest.shouldCreateFailedCommandMetricWithCustomSettings(MongoMetricsCommandListenerTest.java:140)
```

[...]

PR #6924, opened 2025-11-26
Fixed flaky test for mongo client

etrandafir93:
Closes gh-4602

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
@@ -22,13 +22,13 @@ import com.mongodb.client.MongoClients;
 import com.mongodb.event.ClusterListener;
 import com.mongodb.event.ClusterOpeningEvent;
 import com.mongodb.event.CommandEvent;
+import io.micrometer.common.lang.Nullable;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.search.MeterNotFoundException;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import org.bson.Document;
-import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
@@ -53,7 +53,8 @@ class MongoMetricsCommandListenerTest extends AbstractMongoDbTest {
 
     private MeterRegistry registry = new SimpleMeterRegistry();
 
-    @Nullable private MongoClient mongo;
+    @Nullable
+    private MongoClient mongo;
 
     @Test
     void shouldCreateSuccessCommandMetric() {


micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
```diff
@@ -137,43 +134,47 @@ public Iterable<Tag> commandTags(CommandEvent event) {
...
     void shouldSupportConcurrentCommands() throws InterruptedException {
+        mongo = createMongoClient(new AtomicReference<>());
+        assertThat(mongo).isNotNull();
+
         for (int i = 0; i < 20; i++) {
             Map<String, Thread> commandThreadMap = new HashMap<>();
 
             commandThreadMap.put("insert",
-                    new Thread(() -> mongo.getDatabase("test")
+                    new Thread(() -> requireNonNull(mongo).getDatabase("test")
```
shakuzen, line 162: Are these `requireNonNull` needed? Since it is asserted to not be null earlier, I would expect it shouldn't be needed.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
```diff
@@ -206,7 +207,27 @@ void shouldSupportConcurrentCommands() throws InterruptedException {
...
+
+    private MongoClient createMongoClient(AtomicReference<String> clusterId) {
+        // tag::setup[]
+        MongoClientSettings settings = MongoClientSettings.builder()
+            .addCommandListener(new MongoMetricsCommandListener(registry))
+            .applyToClusterSettings(builder -> builder.hosts(singletonList(new ServerAddress(host, port)))
+                .addClusterListener(new ClusterListener() {
+                    @Override
+                    public void clusterOpening(ClusterOpeningEvent event) {
+                        clusterId.set(event.getClusterId().getValue());
```
shakuzen, line 225: I think we don't want this to end up in the documentation snippet, but since it was there before, this isn't changing anything. We can tackle getting that out of the documentation in another PR.

etrandafir93: I tried removing it from the docs by adding an extra tag `// tag::setup_2[] `, but I couldn't get the docs to build locally and didn't want to risk breaking the indentation.

If we end up creating a separate issue for the docs, it might be worth adding something like a docker-compose to make local doc building easier. (if it works - I'm not familiar with antora). - what do you think?

for eg, Testcontainers has a setup like this for mkdocs: 
https://github.com/testcontainers/testcontainers-java/blob/main/docker-compose.yml

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
etrandafir93: I thought so as well, but, because the field (now _nullable_) is used in the lambda, we still need the null checks.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
shakuzen: Did `./gradlew antora` not work for you? What was the error?

etrandafir93: Sorry for the late reply! I had some time to look into it again today, and I've got it working. 

There were a few different issues, for eg, for some reason, building the `docs` module was not using the project version for `micrometer-observation`. So, I was getting compilation errors because of some new classes used in [ObservationHandlerTests.java](https://github.com/micrometer-metrics/micrometer/blob/main/docs/src/test/java/io/micrometer/docs/observation/ObservationHandlerTests.java). Then, I had some issues with the symbolic links in [here](https://github.com/micrometer-metrics/micrometer/tree/main/docs/modules/ROOT/examples). 

But it worked in the end, and I split the tag into two. Therefore, that whole part with 
`ClusterSettings > ClusterListener > ClusterId` should no longer show up in the documentation

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListenerTest.java
etrandafir93: @shakuzen is it ugly? 
We can have the mongo client as a local variable in each test instead of keeping it as a nullable class field to get rid of this.

But then we won't be able to use `@AfterEach` to close the clients, and we'll use try with resources everywhere

shakuzen: It's fine. Looking closer I see why it's still needed.

PR #6945, opened 2025-12-06
Remove `../` in xref links

ngocnhan-tran1996:
I think we don't need `../` in xref links

>Also, remember that the [file coordinate](https://docs.antora.org/antora/latest/page/resource-id-coordinates/#id-resource) of a target page is calculated from the pages family directory.

Reference
https://docs.antora.org/antora/latest/page/xref/#structure


shakuzen:
Indeed it seems we don't need it. I wonder if it hurts anything to have it, though. I notice that if you're trying to navigate the docs on GitHub's website, the links work as is (with `../`). With these changes to remove `../`, the links when browsing GitHub no longer work. The generated site will work as expected, and browsing in IntelliJ the links still work. I'm not sure how much we care to keep the links working from browsing GitHub, but if it doesn't hurt anything to keep that, I wonder if it's worth it to keep it. I don't feel strongly either way. I'm fine to merge this or not. I'm just curious if there are any specific downsides to the way we have things now that seem to work.

ngocnhan-tran1996:
If the source file is moved, the Antora path usually remains the same, whereas the `../` path breaks immediately. IMO, the generated Antora site is the official output, and GitHub is a secondary concern that can compromise the system's longevity.

jonatan-ivanov:
Is this causing any issues to you? If not, I think we should keep things as-is:

Antora can work with both "~absolute" (starting from `pages`) and "~relative" paths (starting from current file: `../` and `./`), you can see in the commit where this was changed that both were used interchangeably: https://github.com/micrometer-metrics/micrometer/commit/daab96e044ada4b5a95b3bf7fba38d2e569ee595. I think we should make things consistent and only use one of them.

Although "~absolute" paths can be simpler in certain situations for humans, using them actually break things: IDE integration, GitHub, etc. Tools (unlike Antora) do not know that "~absolute" paths start from the `pages` directory. It's very convenient to follow these links in the IDE (or GitHub or in other tools) just by clicking them and breaking this functionality makes contributing to and maintaining the docs harder.

>If the source file is moved, the Antora path usually remains the same, whereas the ../ path breaks immediately.

I'm not sure I get this, if you move an `.adoc` file both "~absolute" and "~relative" paths break. If this happens, we will have two concerns: a broken path will fail the build (Antora checks them) and we should not really move files in a breaking way (without a redirect) since that will break existing links so I'm not sure I see any concern here.

>IMO, the generated Antora site is the official output, and GitHub is a secondary concern that can compromise the system's longevity.

I'm not sure how this "can compromise the system's longevity", also it's not just GitHub, it's also IDEs and all the tools that can follow cross-references in files. I think making the docs contributor- and maintainer friendly is important and I think letting tools follow cross-references plays a somewhat significant role in this.

shakuzen:
I just wanted to share that at least in IntelliJ for me the links in this PR work. I don't remember if I've installed an Antora plugin or something. Actually, the relative links don't work. When I open the high-cardinality-tags-detector.adoc file in IntelliJ with a preview of the adoc file and go to the bottom, the "Meter Filter" link works (`xref:./meter-filters.adoc`) but the "Observation API" link does not (`xref:../observation/introduction.adoc`).

ngocnhan-tran1996:
Thanks for the clarification! There's no issue on my side, this is just a suggestion for consistency. Using Antora-style paths avoids `../`, which tends to break when files are reorganized.

shakuzen:
Thanks for the pull request. For now, we will pass on making the proposed changes. It's not clear to us the benefit, and it makes things worse for some tooling that contributors might be using. We'll keep things how they are for now, but we can revisit this in the future when the benefit is more clear.

PR #6947, opened 2025-12-08
Make JvmThreadMetrics more performant and less resource-intensive

snifferhu:
1. Remove use method call -> threadBean.getAllThreadIds()
2. Performance Issue:
The getThreadStateCountmethod executes both threadBean.getAllThreadIds()and threadBean.getThreadInfo()on each call. Both operations are relatively expensive JMX operations, and invoking them repeatedly within a loop results in unnecessary performance overhead.

3. Resource Waste:
Each time a different thread state is counted, all thread information is retrieved again instead of being fetched once and reused, leading to a waste of system resources.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
@@ -103,7 +103,14 @@ public class JvmThreadMetrics implements MeterBinder {
             .register(registry);
 
         try {
-            Map<Thread.State, Long> stateCountMap = getThreadStateCountMap(threadBean);
+            long[] allThreadIds = threadBean.getAllThreadIds();
+            ThreadInfo[] threadInfos = threadBean.getThreadInfo(allThreadIds);
+            Map<Thread.State, Long> stateCountMap = new EnumMap<>(Thread.State.class);
+
+            for (ThreadInfo threadInfo : threadInfos) {
+                Thread.State state = threadInfo.getThreadState();
+                stateCountMap.merge(state, 1L, Long::sum);
+            }
 
             MeterConvention<Thread.State> threadCountConvention = conventions.threadCountConvention();
             for (Thread.State state : Thread.State.values()) {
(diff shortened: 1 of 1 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmThreadMetrics.java
```diff
@@ -100,10 +103,18 @@ public void bindTo(MeterRegistry registry) {
...
+
+            for (ThreadInfo threadInfo : threadInfos) {
+                Thread.State state = threadInfo.getThreadState();
+                stateCountMap.merge(state, 1L, Long::sum);
+            }
+
             MeterConvention<Thread.State> threadCountConvention = conventions.threadCountConvention();
             for (Thread.State state : Thread.State.values()) {
-                Gauge.builder(threadCountConvention.getName(), threadBean, (bean) -> getThreadStateCount(bean, state))
+                Gauge.builder(threadCountConvention.getName(), threadBean, (bean) -> stateCountMap.get(state))
```
shakuzen, line 117: The `stateCountMap` is never updated after `bindTo` is called, so the thread state metrics will always show the snapshot of the state when bindTo was called. This should be possible to show in a test. Bind JvmThreadMetrics, change the state of a thread, check the metrics, assert that the metrics reflect the changed thread's state.

snifferhu: Oh. I never thought about that.
But Thread. states have 6. Every Time we call JvmThreadMetrics, we need to loop 6 times ThreadIds[].

Wait me, let me try changing another way.

shakuzen: I vaguely remember we may have had a previous pull request trying to improve the efficiency of this. I can check tomorrow when I have time.

MultiGauge is, I think, the closest thing we have to what we want here, which is to do an expensive operation once to update the value for a group of gauges. The challenge is that MultiGauge requires something external to handle when the updating happens. With regular Gauge, "updating" happens when publishing happens for each value (i.e. when the gauge value is read).

shakuzen:
This is the previous pull request I was thinking of: https://github.com/micrometer-metrics/micrometer/pull/3664. We still have some work/design to do there. I'm going to close this because this PR in its current form does not solve the performance issue in a way that preserves the current functionality. Thanks for the effort on this nonetheless.

PR #6974, opened 2025-12-12
Mark CaffeineCacheMetrics generics with nullability

### LINKED ISSUE #6973: 1.16.1 is incompatible with Caffeine 3.2.3
**Describe the bug**
`CaffeineCacheMetrics` is incompatible with newer versions of Caffeine due to differences in nullability markings.

The migration to JSpecify saw `CaffeineCacheMetrics` marked as `@NullMarked`. This works fine in the repository, because Caffeine 2.9 is in use, which doesn't have JSpecify annotations. However, version 3.2.3 of Caffeine also uses JSpecify annotations. `CaffeineCacheMetrics` and Caffeine's `Cache` classes mark their generics differently, resulting in an incompatibility.

```java
public class CaffeineCacheMetrics<K, V, C extends Cache<K, V>> extends CacheMeterBinder<C> {

public interface Cache<K, V extends @Nullable Object> {
```

Note that Caffeine's `Cache` explicitly marks `V` as being nullable.

The impact of this is that application code using both libraries may have a valid `Cache`, that cannot be passed to `CaffeineCacheMetrics` because the `V` type is out of bounds.

Other versions of Caffeine are likely impacted in the same way, but 3.2.3 happens to be the most recent release and what I was using.

**Environment**

 - Micrometer version: 1.16.1
 - Micrometer registry: JMX
 - OS: macOS
 - Java version: 21

**To Reproduce**
How to reproduce the bug:
A short excerpt from my code where the bug was witnessed:

```kotlin
interface ResultsIngester

val resultsIngestersByTaskInstanceSplitId = Caffeine
	.from(CaffeineSpec.parse("maximumSize=1000, expireAfterAccess=PT10M"))
	.recordStats()
	.build<Int, ResultsIngester?>()
val metrics = CaffeineCacheMetrics(
	resultsIngestersByTaskInstanceSplitId,
	"results-ingesters-by-task-instance-split",
	emptyList(),
)
```

I'll provide a PR with the fix. If additional reproduces are requested after that, let me know.

**Expected behavior**
The nullability behavior should match

MariusVolkhart:
This matches the nullability behavior of newer versions of Caffeine: https://github.com/ben-manes/caffeine/blob/fc36e9a1ad4a016b231911e4f27649d2ef8b3b17/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Cache.java#L44

An explanation of the nullability behavior of JSpecify regarding generics is available at https://jspecify.dev/docs/user-guide/#declaring-generics

Fixes #6973

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
@@ -44,7 +44,7 @@ import java.util.function.ToLongFunction;
  * @see CaffeineStatsCounter
  */
 @NullMarked
-public class CaffeineCacheMetrics<K, V, C extends Cache<K, V>> extends CacheMeterBinder<C> {
+public class CaffeineCacheMetrics<K, V extends @Nullable Object, C extends Cache<K, V>> extends CacheMeterBinder<C> {
 
     private static final String DESCRIPTION_CACHE_LOAD = "The number of times cache lookup methods have successfully loaded a new value or failed to load a new value, either because no value was found or an exception was thrown while loading";
 


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetrics.java
```diff
@@ -44,7 +44,7 @@
  * @see CaffeineStatsCounter
  */
 @NullMarked
```
shakuzen, line 46: Not related to this PR and it doesn't need to be changed as part of this PR, but I'm just pointing out that this `@NullMarked` is redundant and can be removed because the package-info.java is annotated with `@NullMarked` already.

shakuzen: I opened https://github.com/micrometer-metrics/micrometer/issues/6980 for this

Issue #7009, opened 2025-12-18
Support compressing metrics payload in OtlpConfig and OtlpHttpMetricsSender


allnightlong:
**Please describe the feature request.**
`OtlpHttpMetricsSender` uses `HttpSender` which already supports `compress()` method. For some big apps, we can significantly improve application's performance by using it.

I believe the following chain of propogation makes sens:

1. Add setting to `OtlpConfig` if the data should be compressed.
2. In `OtlpMeterRegistry.publish()` method pass this setting's value into `OtlpMetricsSender.Request` corresponding field.
3. In `OtlpHttpMetricsSender.send()` method call `compress` on `HttpSender.Request.Builder`, if the setting in ON in the request.

**Rationale**
The amount of data my app generates is too big to fit into my otlp hosting (newrelic's limit is 1MB). 

**Additional context**

jonatan-ivanov:
Thank you for the issue!
It seems you already implemented this, would you mind sharing the difference you see in the content length with- and without gzip?

I quickly checked this with some arbitrary data I generated, I see big enough difference so that I think we should add gzip capability.

I see two options:
- Always do compression
- Do compression by default and let users opt out (config)

According to the OTLP spec, collectors must support gzip. Based on this alone, I would always enable it since this should not affect collectors that comply to the OTLP spec. What drives me to the other way (opt-out config) is:
- There might be non-compliant collectors
- In some cases compressed output it might be bigger (not sure this is interesting though since in those cases the output is usually small)
- Opinions varies a lot within OTel
  - https://github.com/open-telemetry/opentelemetry-collector/issues/4587
  - https://github.com/open-telemetry/opentelemetry-specification/issues/2098
  - https://github.com/open-telemetry/opentelemetry-specification/pull/1895

What you described above (compress flag in `OtlpMetricsSender.Request`) looks good to me, are up to creating a PR for this?

shakuzen:
> The amount of data my app generates is too big to fit into my otlp hosting (newrelic's limit is 1MB).

[...]

PR #7012, opened 2025-12-19
Support compressing metrics payload in OtlpConfig and OtlpHttpMetricsSender

allnightlong:
Resolves https://github.com/micrometer-metrics/micrometer/issues/7009

### DIFF
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
@@ -118,9 +118,62 @@ class OTelCollectorIntegrationTest {
         // @formatter:on
     }
 
+    @Test
+    void collectorShouldExportMetricsWithGzipCompression() throws Exception {
+        MeterRegistry registry = createOtlpMeterRegistryForContainerWithGzipCompression(container);
+        Counter.builder("test.counter.gzip").register(registry).increment(42);
+        Gauge.builder("test.gauge.gzip", () -> 12).register(registry);
+        Timer.builder("test.timer.gzip").register(registry).record(Duration.ofMillis(123));
+        DistributionSummary.builder("test.ds.gzip").register(registry).record(24);
+
+        // @formatter:off
+        await().atMost(Duration.ofSeconds(5))
+            .pollDelay(Duration.ofMillis(100))
+            .pollInterval(Duration.ofMillis(100))
+            .untilAsserted(() -> whenPrometheusScraped().then()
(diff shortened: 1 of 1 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/CompressionMode.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Compression mode for metric data sender.
+ *
+ * @author Tigran Kavanosyan
+ * @since 1.17.0
+ */
+public enum CompressionMode {
+
+    ON, OFF
```
shakuzen: Since other compression algorithms are possible, I think it would be safer to use `GZIP` instead of `ON`. And to match [OTel exporter env var values](https://opentelemetry.io/docs/specs/otel/protocol/exporter/), `OFF` could instead be `NONE`. 

> Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:
> 
> - `none` if compression is disabled.
> - `gzip` is the only specified compression method for now.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -156,6 +156,25 @@ default AggregationTemporality aggregationTemporality() {
...
+     * metrics data should be compressed before being sent to the OTLP endpoint. Default
+     * implementation supports the environment variable
+     * {@code OTEL_EXPORTER_OTLP_METRICS_COMPRESSION} when a value is not provided by
+     * {@link #get(String)}.
+     * @return the compressionMode; default is OFF
+     * @since 1.17.0
+     */
+    default CompressionMode compressionMode() {
+        return getEnum(this, CompressionMode.class, "compressionMode").orElseGet(() -> {
+            String compression = System.getenv().get("OTEL_EXPORTER_OTLP_METRICS_COMPRESSION");
```
shakuzen: We should also support `OTEL_EXPORTER_OTLP_COMPRESSION` but `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` takes precedence over it if it is set. See, for example, a similar situation in the `headers()` method supporting both a generic and metrics specific env var.

shakuzen:
Hi @allnightlong, I wanted to follow-up to see if you have time and interest to update this pull request based on the review I left. Let us know. We'd like to get this feature into the upcoming minor release.

allnightlong:
@shakuzen thanks for follow up
I've updated the PR, as you suggested

allnightlong:
@shakuzen I've added the integration test  - please check it out

Issue #6998, opened 2025-12-17
Add static code analysis to build


shakuzen:
#6946 unveiled some issues that could have been caught before they were released if we had static code analysis for such issues. [SpotBugs](https://spotbugs.github.io/) is one such option we can consider. In a quick local experiment, it did catch the issues mentioned in #6946.

shakuzen:
See #7021. That obviates the initial reason this issue was opened. Perhaps there's still room to consider whether there are other checks that a tool like SpotBugs or something else could add that we don't have with our current build tools like ErrorProne.

shakuzen:
Closing for now until we have more demonstrated reason to prioritize this work.

Issue #6946, opened 2025-12-08
NoopObservationRegistryTests are flaky


shakuzen:
Specifically:

```
java.lang.AssertionError: 
Expecting actual:
  null
and:
  io.micrometer.observation.NoopObservation$NoopScope@2f08c4b
to refer to the same object
	at io.micrometer.observation.NoopObservationRegistryTests.shouldNotRespectScopesIfNoHandlersAreRegistered(NoopObservationRegistryTests.java:95)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

```
java.lang.AssertionError: 
Expecting actual:
  null
and:
  io.micrometer.observation.NoopObservation$NoopScope@2f08c4b
to refer to the same object
	at io.micrometer.observation.NoopObservationRegistryTests.shouldNotRespectScopesIfNoopRegistryIsUsed(NoopObservationRegistryTests.java:71)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

etrandafir93:
Hey @shakuzen, 
Since i already got started with flaky tests, can I take this one, too? 😆

shakuzen:
Of course. Thank you for all the help. I've assigned it to you.

etrandafir93:
@shakuzen It's an interesting issue - and it seems to be more than a flaky test!

In fact, as far as I can see, there are 2 problems here:

### 1. the functionality that the test itself is testing is not really working

Seems as for `NoopObservationRegistry` , the initial currentObservationScope is never set to `Observation.Scope.NOOP`. 
So when the test is failing, it's not flaky - but it's actually showing a real problem. 

I don't think it should be too complicated to fix this, but the problem is definitely there - and we can prove it with just by adding a sysout: https://github.com/micrometer-metrics/micrometer/pull/6960/files

which will consistently break the build 😱 😱: https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/10995/workflows/f7b32be3-fc02-476d-a976-8477ac3b0608/jobs/60041


_But the more interesting question is - **why was this failure hidden?**_

[...]

PR #7021, opened 2025-12-21
add errorprone's ClassInitializationDeadlock

etrandafir93:
@shakuzen, @marcingrzejszczak 

Turns out there was an errorprone rule already that should guard against the exact issue we found in #6946
[https://errorprone.info/bugpattern/ClassInitializationDeadlock](https://errorprone.info/bugpattern/ClassInitializationDeadlock)
They even suggest some fixes, but all of them seem to introduce breaking changes

This successfully found 3 occurrences of this pattern: [see job logs](https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/11104/workflows/88d054d6-9aad-4bba-9778-99d7ee1efde4/jobs/60781)

Should we suppress the 3 errors for now, so that we can merge this? Or are we going to tackle the issue soon? 

See gh-6998

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -52,6 +52,7 @@ public interface Observation extends ObservationView {
      * No-op observation. Do not use it to check if an Observation is no-op, use
      * {@code observation.isNoop()} instead.
      */
+    @SuppressWarnings("ClassInitializationDeadlock")
     Observation NOOP = NoopObservation.INSTANCE;
 
     /**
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
I would vote for suppression unless you fix the issue in the same PR :)

shakuzen:
@etrandafir93 good find on the errorprone check I didn't realize was available (probably a sign we have too many errorprone warnings that people tend to ignore them).
Feel free to suppress the cases now and we can figure out what to do about them in a separate pull request. It still adds value to fail on this check in other cases so we don't introduce more of this same issue with future changes.

etrandafir93:
@shakuzen, @marcingrzejszczak 

The rule existed on errorprone's catalog, but it was not configured for the micrometer project. 

I've added it in my first commit to see the build failing, and then suppress the warning for our 3 known issues in the 2nd commit.

So yes, if we add this anti-pattern again, the build will fail 👍

shakuzen:
Thanks for the quick updates.
Is the PR ready to merge or is there any reason it's still marked as a draft?

etrandafir93:
@shakuzen - yes, ready for merge.

PR #7038, opened 2025-12-28
Apply @CheckReturnValue to Assert implementations

izeye:
This PR applies `@CheckReturnValue` to AssertJ `Assert` implementations similar to https://github.com/spring-projects/spring-boot/pull/46766.

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistryAssert.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistryAssert.java
@@ -18,6 +18,7 @@ package io.micrometer.observation.tck;
 import io.micrometer.common.KeyValue;
 import io.micrometer.common.docs.KeyName;
 import io.micrometer.observation.Observation;
+import org.assertj.core.annotation.CanIgnoreReturnValue;
 import org.assertj.core.annotation.CheckReturnValue;
 import org.assertj.core.api.ThrowingConsumer;
 import org.jspecify.annotations.Nullable;
@@ -93,7 +94,7 @@ public class TestObservationRegistryAssert
      * @return this
      * @throws AssertionError if there is no matching observation
      */
-    @CheckReturnValue
+    @CanIgnoreReturnValue
     public That hasObservationWithNameEqualTo(@Nullable String name) {
         Queue<TestObservationRegistry.TestObservationContext> contexts = this.actual.getContexts();
         if (contexts.isEmpty()) {
@@ -124,7 +125,7 @@ public class TestObservationRegistryAssert
      * @return this
      * @throws AssertionError if there is no matching observation
      */
-    @CheckReturnValue
+    @CanIgnoreReturnValue
     public That hasObservationWithNameEqualToIgnoringCase(String name) {
         Queue<TestObservationRegistry.TestObservationContext> contexts = this.actual.getContexts();
         if (contexts.isEmpty()) {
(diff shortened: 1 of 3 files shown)


shakuzen (approved):
Looks good. Thanks. Should we add an ArchUnit rule as well to make sure we don't miss this going forward?

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistryAssert.java
```diff
@@ -91,6 +93,7 @@ private void failForNoObservations() {
      * @return this
      * @throws AssertionError if there is no matching observation
      */
+    @CheckReturnValue
     public That hasObservationWithNameEqualTo(@Nullable String name) {
```
izeye, line 98: This seems to read okay without `that()` and is also being used internally without `that()` which requires warning suppressions below.

I'm not sure if requiring `that()` is intentional, but it could be a breaking change for someone using it just for its presence check.

micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryAssert.java
```diff
@@ -388,6 +388,7 @@ private String allMetrics() {
      * @return a {@link MeterAssert} for the found meter
      * @throws AssertionError if the meter is not found
      */
+    @CheckReturnValue
     public MeterAssert<?> meter(String meterName, Tag... tags) {
```
izeye, line 392: This one and the following ones are similar cases with the above, but they seem to read incomplete.

izeye (commented):

I tried to reuse the rules from Spring Boot and added more `@CheckReturnValue` annotations to comply with them in ef403a40777d8d8f5d4955b5f4435aaf383b9cb3.

There are some cases that I'm not sure if they are okay, which I commented below.

izeye:
For the build failure, I created https://github.com/micrometer-metrics/micrometer/pull/7044 to try to resolve it.

micrometer-test/src/test/java/io/micrometer/AssertJArchitectureTests.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2025 VMware, Inc.
```
shakuzen: Happy New Year! I guess this should be the first 2026 copyright year file

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistryAssert.java
shakuzen: It does seem a bit awkward compared to other API. Perhaps we can work on migrating to something more consistent. For now, I think it's better to not have `@CheckReturnValue` here, since it can be used with or without using the return value.

micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryAssert.java
shakuzen: I would keep the `@CheckReturnValue` here. If someone just wants to assert a meter with a name and tags exists, there are other ways to do that and they don't need to use this method.

shakuzen (commented):
Happy New Year, Johnny! Thanks for updating things. I took a look and left some comments.

izeye:
@shakuzen Happy New Year, Tommy!

I tried to apply your feedback in 2768cd9.

shakuzen (approved):

PR #7048, opened 2026-01-03
Polish CurrentObservationTest

izeye:
This PR polishes `CurrentObservationTest` a bit.

### DIFF
--- a/micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java
+++ b/micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java
@@ -38,37 +38,43 @@ class CurrentObservationTest {
     @Test
     void nestedSamples_parentChildThreadsInstrumented() throws ExecutionException, InterruptedException {
         ExecutorService executor = Executors.newSingleThreadExecutor();
-
-        Observation observation = Observation.createNotStarted("test.observation", registry);
-        assertThat(registry.getCurrentObservation()).isNull();
-        try (Observation.Scope scope = observation.openScope()) {
-            assertThat(registry.getCurrentObservation()).isSameAs(observation);
-            executor.submit(() -> {
-                assertThat(registry.getCurrentObservation()).isNotEqualTo(observation);
-            }).get();
+        try {
+            Observation observation = Observation.createNotStarted("test.observation", registry);
+            System.out.println("Outside task: " + observation);
+            assertThat(registry.getCurrentObservation()).isNull();
+            try (Observation.Scope scope = observation.openScope()) {
+                assertThat(registry.getCurrentObservation()).isSameAs(observation);
+                executor.submit(() -> {
+                    System.out.println("In task: " + registry.getCurrentObservation());
+                    assertThat(registry.getCurrentObservation()).isNotEqualTo(observation);
+                }).get();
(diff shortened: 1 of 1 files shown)


micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java
```diff
@@ -40,17 +40,17 @@ void nestedSamples_parentChildThreadsInstrumented() throws ExecutionException, I
         ExecutorService executor = Executors.newSingleThreadExecutor();
 
         Observation observation = Observation.createNotStarted("test.observation", registry);
-        System.out.println("Outside task: " + observation);
```
jonatan-ivanov, line 43: These might be helpful for "debugging"/understanding what is happening, what do you think about commenting them out instead?

I also find the other commented out statements (sout and registering ObservationTextPublisher) useful in cases when you want to see what happens under the hood (you just uncomment them). We don't need to do it very often, but whenever we need to, having these in demo/test code can help.

micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java
```diff
@@ -40,17 +40,17 @@ void nestedSamples_parentChildThreadsInstrumented() throws ExecutionException, I
...
             assertThat(registry.getCurrentObservation()).isSameAs(observation);
             executor.submit(() -> {
-                System.out.println("In task: " + registry.getCurrentObservation());
                 assertThat(registry.getCurrentObservation()).isNotEqualTo(observation);
             }).get();
         }
         assertThat(registry.getCurrentObservation()).isNull();
         observation.stop();
+
+        executor.shutdown();
```
jonatan-ivanov: What do you think about using try-with-resource instead so that shutdown will be called even if the test fails?

micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java
izeye: @jonatan-ivanov Sorry, I thought that they were leftovers while developing. I reinstated them in ef790e5.

micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java
izeye: @jonatan-ivanov Actually, I used try-with-resources at first, but removed them in 6c75af2 as they require JDK 19+.

I changed to use try-finally instead which has a similar effect in ef790e5.

micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java
jonatan-ivanov: No worries, maybe we should figure out a way to signal that this was intentional.

micrometer-observation/src/test/java/io/micrometer/observation/CurrentObservationTest.java
jonatan-ivanov: Oh, right. Thank you!

jonatan-ivanov:
Thank you!

PR #7053, opened 2026-01-06
Reduce flakiness of StatsdMeterRegistryPublishTest

shakuzen:
~It's not clear why this await behavior was not applied in the case of UDP protocol, but removing the conditional that excludes that seems to fix flakiness of this test.~

Upgrading the version of Netty has resolved the underlying flakiness seen in the test and at least from running it 15 times or so locally and a few times on the CI it appears stable.

Find an available port instead of using a fixed port that might not be available.

### DIFF
--- a/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
+++ b/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
@@ -141,7 +141,7 @@ class StatsdMeterRegistryPublishTest {
         // For UDP, the first change seems to be lost frequently somehow.
         Counter.builder("another.counter").register(meterRegistry).increment();
 
-        if (protocol == StatsdProtocol.UDS_DATAGRAM) {
+        if (protocol == StatsdProtocol.TCP || protocol == StatsdProtocol.UDS_DATAGRAM) {
             await().until(() -> meterRegistry.statsdConnection.get() != firstClient);
         }
 


jonatan-ivanov (approved):

jonatan-ivanov:
It looks completely reasonable to me too but the build on CI seems to have a bad time.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
```diff
@@ -141,9 +141,7 @@ void resumeSendingMetrics_whenServerIntermittentlyFails(StatsdProtocol protocol)
         // For UDP, the first change seems to be lost frequently somehow.
         Counter.builder("another.counter").register(meterRegistry).increment();
 
-        if (protocol == StatsdProtocol.TCP || protocol == StatsdProtocol.UDS_DATAGRAM) {
-            await().until(() -> meterRegistry.statsdConnection.get() != firstClient);
-        }
+        await().until(() -> meterRegistry.statsdConnection.get() != firstClient);
```
jonatan-ivanov: I guess we can check the OS/protocol combinations and do this conditionally or maybe we can say that sometimes the client will not be replaced:
```suggestion
        try {
            await().until(() -> meterRegistry.statsdConnection.get() != firstClient);
        }
        catch (ConditionTimeoutException ex) {
            // Intentionally ignored: in some cases the client gets replaced
            // and in some cases it does not (os and protocol dependent)
            // Hope Awaitility will have support for this use-case:
            // https://github.com/awaitility/awaitility/issues/48
        }
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
jonatan-ivanov, line 228: Should we change this too?

shakuzen:
Let's wait for https://github.com/netty/netty/pull/16131 which should hopefully make this more uniform across platforms.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
```diff
@@ -351,12 +351,12 @@ void counterIncrementWorksWithSleepAfterRegistryStart() throws InterruptedExcept
 
     @Test
     void counterIncrementDoesNotWorkWithoutSleepAfterRegistryStart() throws InterruptedException {
-        StatsdConfig config = StatsdConfig.DEFAULT;
-
         CountDownLatch serverLatch = new CountDownLatch(1);
-        DisposableChannel server = startServer(StatsdProtocol.UDP, config.port(), serverLatch);
+        DisposableChannel server = startServer(StatsdProtocol.UDP, 0, serverLatch);
```
shakuzen, line 355: I'm not sure why we were using a fixed port before but it doesn't seem important to the test so I updated this to use a random port which avoids port conflicts (especially when running tests locally).

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/StatsdMeterRegistryPublishTest.java
shakuzen: Actually, I reverted the previous changes to the other conditional. I think this is right. The client is not replaced for UDP because of the following code and Violeta's fix in Netty to consistently throw PortUnreachableException with UDP on all platforms.
https://github.com/micrometer-metrics/micrometer/blob/bcbf4fe97b15251c3f09bd9d8ccae796afbc20f9/implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/StatsdMeterRegistry.java#L254

@violetagg let me know if I'm mistaken.

shakuzen:
I've rebased to pick up the Netty upgrade including the changes from the linked pull request. I reverted the change to the conditional and now things pass (seemingly) consistently locally and on CI.

I reworked the pull request to update some places we were using a fixed port instead of getting an available port, which caused some failures running locally.

Issue #7060, opened 2026-01-08
OSGi test isn't reporting failures


jack-berg:
Was working on adding OSGi support for opentelemetry-java and research led me to find that micrometer supports it and tests its support here: https://github.com/micrometer-metrics/micrometer/tree/main/micrometer-osgi-test

In my implementation effort I was seeing weird behavior where I setup a test to blatantly fail and the build was still passing. In an attempt to find the root cause, I tried to compare / contrast with micrometer and found that micrometer-osgi-test has the same problem!

To recreate, change the test to fail, like [here](https://github.com/jack-berg/micrometer/commit/31f46b729287f3f6ad50662a24a0af620a5bca85), and notice that running `./gradlew :micrometer-osgi-test:test` still succeeds.

Didn't have time to track down the exact cause for micrometer, but for opentelemetry-java, the key bit was here: https://github.com/open-telemetry/opentelemetry-java/pull/7964/changes#diff-2a792e194b8d6465c4f4caea2a7f31fc6179a196def02b0dc07f19d97734ba48R41-R43

Changing the micrometer-osgi-test junit dependency from testImplementation to implementation did not fix the issue like it did for me:
```
implementation libs.junitJupiter
```

So there's something weird going. I don't know exactly what it is, but I thought I'd open an issue to let you know your osgi tests aren't to be trusted right now! 😬

jonatan-ivanov:
Thanks for the issue!

I have no idea where/why did this break but I was able to make the tests fail on `main` by using `implementation`:
```diff
-    testImplementation platform('org.junit:junit-bom:5.12.2')
-    testImplementation libs.junitJupiter
+    implementation platform('org.junit:junit-bom:5.12.2')
+    implementation libs.junitJupiter
```
Build scan: https://ge.micrometer.io/s/2qq5x6rogjkkm/console-log?page=1#L92

The same hack does not work on older branches (there we use Gradle 8, a different version of the BND plugin, etc), I need to dig deeper (it seems the tasks are executed but tests are not discovered).

@chrisrueger Do you happen to know if anyone has reported similar issues?

jonatan-ivanov:
It seems the `implementation` hack works on older branches too if I upgrade the bnd plugin
from
```
biz.aQute.bnd:biz.aQute.bnd.gradle:7.1.0
```
to
```
biz.aQute.bnd:biz.aQute.bnd.gradle:7.2.0
```

chrisrueger:
[...]

PR #7063, opened 2026-01-10
Fix OSGi tests

jonatan-ivanov:
They never fail

Closes gh-7060

### DIFF
--- a/micrometer-osgi-test-slf4j2/build.gradle
+++ b/micrometer-osgi-test-slf4j2/build.gradle
@@ -17,6 +17,8 @@ dependencies {
 
     testImplementation libs.felixFramework
 
+    // JUnit needs to be an 'implementation' dependency instead of 'testImplementation' otherwise
+    // tests are not dicovered and executed (so the tests never fail), see: https://github.com/micrometer-metrics/micrometer/issues/7060
     implementation platform(libs.junitBom)
     implementation libs.junitJupiter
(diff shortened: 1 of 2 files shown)


micrometer-osgi-test-slf4j2/build.gradle
```diff
@@ -23,7 +17,11 @@ dependencies {
 
     testImplementation libs.felixFramework
 
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+    // BND not compatible with JUnit 5.13+; see https://github.com/bndtools/bnd/issues/6651
```
shakuzen: BND 7.2.1 doesn't have this limitation. `main` is on 7.2.1 now and I removed the version pinning there.

micrometer-osgi-test-slf4j2/build.gradle
```diff
@@ -23,7 +17,10 @@ dependencies {
 
     testImplementation libs.felixFramework
 
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
+    implementation platform(libs.junitBom)
+    implementation libs.junitJupiter
```
shakuzen, line 23: I think a comment here is warranted why this is not `testImplementation`. I could see myself changing this back to a test-scoped dependency and then we'd be back in the same situation and perhaps not realize it.

shakuzen (approved):

PR #7082, opened 2026-01-16
Sanitize equals signs in Telegraf tag keys

### LINKED ISSUE #6513: Fix escaping for Telegraf StatsD
I tried to look into this but cannot find specific documentation on it. [Here](https://www.influxdata.com/blog/getting-started-with-sending-statsd-metrics-to-telegraf-influxdb/#heading5) it says (emphasis added):

> Since InfluxDB supports tags, our StatsD implementation does too! Adding tags to a StatsD metric is **similar** to how they appear in [line-protocol.](https://influxdb.com/docs/v0.9/write_protocols/line.html)

It says similar but that leaves the possibility it isn't exactly the same. Going to the documentation on Influx's line-protocol there is [this section](https://docs.influxdata.com/influxdb/v2/reference/syntax/line-protocol/#special-characters), which describes needing to escape commas, spaces, and equals signs depending on the place.

_Originally posted by @shakuzen in https://github.com/micrometer-metrics/micrometer/issues/6450#issuecomment-3007637972_

pythaac:
Fixes #6513

- sanitize equals signs in Telegraf tag keys
- Pre-compile sanitization patterns as `static final Pattern`
- Add TelegrafStatsdLineBuilderIntegrationTest

### DIFF
--- a/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
+++ b/implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
@@ -48,6 +48,8 @@ import static org.hamcrest.CoreMatchers.containsString;
 @Testcontainers
 class TelegrafStatsdLineBuilderIntegrationTest {
 
+    // Required to resolve the InfluxDB port,
+    // as the mapped port is unknown during static initialization.
     private static final Network network = Network.newNetwork();
 
     private static final DockerImageName INFLUXDB_IMAGE = DockerImageName


implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilder.java
```diff
@@ -73,7 +78,7 @@ private void updateIfNamingConventionChanged() {
     private @Nullable String createConventionTags(NamingConvention namingConvention) {
         return id.getTagsAsIterable().iterator().hasNext() ? id.getConventionTags(namingConvention)
             .stream()
-            .map(t -> telegrafEscape(t.getKey()) + "=" + telegrafEscape(t.getValue()))
+            .map(t -> telegrafEscapeKey(t.getKey()) + "=" + telegrafEscape(t.getValue()))
```
jonatan-ivanov: In https://github.com/micrometer-metrics/micrometer/issues/6513#issue-3207855253, there is a link to the [docs](https://docs.influxdata.com/influxdb/v2/reference/syntax/line-protocol/#special-characters) which says not only tag keys but also tag values require escaping "Comma, Equals Sign, Space". Here we don't escape `=` for values, only for keys.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
+/*
+ * Copyright 2023 VMware, Inc.
```
jonatan-ivanov:
```suggestion
 * Copyright 2025 VMware, Inc.
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final Network network = Network.newNetwork();
+
+    private static final ExposedPort telegrafExposedPort = new ExposedPort(8125, InternetProtocol.UDP);
+
+    private static final String influxDbName = "metrics_db";
+
+    private static final Integer influxDbPort = 8086;
+
+    @Container
+    static GenericContainer<?> influxDB = new GenericContainer<>(DockerImageName.parse("influxdb:1.8"))
```
jonatan-ivanov: Can we use a recent version?

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+        .withNetwork(network)
+        .withNetworkAliases("influxdb")
+        .withExposedPorts(influxDbPort)
+        .withEnv("INFLUXDB_DB", influxDbName)
+        .waitingFor(Wait.forHttp("/ping").forStatusCode(204));
+
+    @Container
+    static GenericContainer<?> telegraf = new GenericContainer<>(DockerImageName.parse("telegraf:1.30"))
```
jonatan-ivanov: Can we use a recent version?

implementations/micrometer-registry-statsd/build.gradle
```diff
@@ -23,6 +23,8 @@ dependencies {
     testImplementation 'io.projectreactor:reactor-test'
     testImplementation libs.logback12
     testImplementation libs.awaitility
+    testImplementation "org.testcontainers:testcontainers:1.19.3"
+    testImplementation "org.testcontainers:junit-jupiter:1.19.3"
```
jonatan-ivanov: Why do we need to use old versions?
We also manage the version of testcontainers and iirc `org.testcontainers:junit-jupiter` brings in `org.testcontainers:testcontainers`:
```suggestion
    testImplementation libs.testcontainers.junitJupiter
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
+@Tag("docker")
+@Testcontainers
+class TelegrafStatsdLineBuilderIntegrationTest {
+
+    private static final Network network = Network.newNetwork();
```
jonatan-ivanov, line 53: Why is this needed?

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+@Testcontainers
+class TelegrafStatsdLineBuilderIntegrationTest {
+
+    private static final Network network = Network.newNetwork();
+
+    private static final ExposedPort telegrafExposedPort = new ExposedPort(8125, InternetProtocol.UDP);
+
+    private static final String influxDbName = "metrics_db";
+
+    private static final Integer influxDbPort = 8086;
```
jonatan-ivanov: Can we have tests without explicitly defining ports?
Testcontainers lets you assign ports randomly, and it has getters that tells you the port and iirc it let's you inject vaiables to config files.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+    @Issue("#6513")
+    @Test
+    void shouldSanitizeEqualsSignInTagKeys() throws InterruptedException {
+        StatsdMeterRegistry registry = getStatsdMeterRegistry(5000);
+
+        try {
+            Counter counter = Counter.builder("metric").tag("this=is=the", "tag=test").register(registry);
+            counter.increment();
+
+            await().atMost(60, TimeUnit.SECONDS).pollInterval(Duration.ofSeconds(1)).untilAsserted(() -> {
```
jonatan-ivanov: 60s sounds like a lot especially since we have `waitingFor`.
Why does this need to be so long? Can we use a better check for `waitingFor`?

If this is for waiting for the registry, you can close it before this line and it will publish the metrics, you don't need to wait the step size (also configurable) to publish the data.
If we do that, I think we can use default poll interval (100ms).

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+        .waitingFor(Wait.forLogMessage(".*Loaded inputs: statsd.*", 1));
+
+    @Issue("#6513")
+    @Test
+    void shouldSanitizeEqualsSignInTagKeys() throws InterruptedException {
+        StatsdMeterRegistry registry = getStatsdMeterRegistry(5000);
+
+        try {
+            Counter counter = Counter.builder("metric").tag("this=is=the", "tag=test").register(registry);
+            counter.increment();
```
jonatan-ivanov: This can be a one-liner:

```suggestion
            Counter.builder("metric").tag("this=is=the", "tag=test").register(registry)increment();
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+        .dependsOn(influxDB)
+        .waitingFor(Wait.forLogMessage(".*Loaded inputs: statsd.*", 1));
+
+    @Issue("#6513")
+    @Test
+    void shouldSanitizeEqualsSignInTagKeys() throws InterruptedException {
+        StatsdMeterRegistry registry = getStatsdMeterRegistry(5000);
+
+        try {
+            Counter counter = Counter.builder("metric").tag("this=is=the", "tag=test").register(registry);
```
jonatan-ivanov: Can metric name also contain an "escapable character"?
Also, can we test other characters too not only `=`?

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+            counter.increment();
+
+            await().atMost(60, TimeUnit.SECONDS).pollInterval(Duration.ofSeconds(1)).untilAsserted(() -> {
+                String queryResult = queryInfluxDB("SELECT * FROM \"metric\"");
+
+                assertThat(queryResult).contains("\"name\":\"metric\"");
+                assertThat(queryResult).contains("\"this_is_the\"");
+                assertThat(queryResult).contains("\"tag=test\"");
```
jonatan-ivanov: According to the docs, this is invalid (see my previous comment), we should assert

```suggestion
                assertThat(queryResult).contains("\"tag_test\"");
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+            });
+        }
+        finally {
+            registry.close();
```
jonatan-ivanov: I think the try-finally block is unnecessary, closing the registry at the end also. See my previous comment, I think we should close the registry before polling using awaitility.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    private StatsdMeterRegistry getStatsdMeterRegistry(long registryWarmUpMs) throws InterruptedException {
+        StatsdConfig statsdConfig = getStatsdConfig();
+        StatsdMeterRegistry registry = new StatsdMeterRegistry(statsdConfig, Clock.SYSTEM);
+        Thread.sleep(registryWarmUpMs);
```
jonatan-ivanov: Why is this needed?

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+                    .getPorts()
+                    .getBindings()
+                    .get(telegrafExposedPort);
+                Ports.Binding binding = Objects.requireNonNull(bindings)[0];
+                return Integer.parseInt(binding.getHostPortSpec());
+            }
+
+            @Override
+            public @NonNull Duration step() {
+                return Duration.ofSeconds(1);
```
jonatan-ivanov: This is the default.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+
+            @Override
+            public int port() {
+                Ports.Binding[] bindings = telegraf.getContainerInfo()
+                    .getNetworkSettings()
+                    .getPorts()
+                    .getBindings()
+                    .get(telegrafExposedPort);
+                Ports.Binding binding = Objects.requireNonNull(bindings)[0];
+                return Integer.parseInt(binding.getHostPortSpec());
```
jonatan-ivanov: `telegraf.getFirstMappedPort()`?
Or if we use a default mapped port: `telegraf.getMappedPort(1234)`.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+            }
+        };
+    }
+
+    private String queryInfluxDB(String query) {
+        try {
+            String baseUrl = "http://" + influxDB.getHost() + ":" + influxDB.getMappedPort(influxDbPort);
+
+            String fullUrl = baseUrl + "/query?db=" + influxDbName + "&q="
+                    + java.net.URLEncoder.encode(query, StandardCharsets.UTF_8);
```
jonatan-ivanov: Can we import `URLEncoder`? Or static import `encode`?

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,168 @@
...
+
+            HttpURLConnection con = (HttpURLConnection) URI.create(fullUrl).toURL().openConnection();
```
jonatan-ivanov: Please check `PrometheusMeterRegistryIntegrationTest` and let's use `RestAssured` instead.

jonatan-ivanov (changes requested):

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilder.java
pythaac: Yes, as you can see [this comment](https://github.com/micrometer-metrics/micrometer/issues/6513#issuecomment-3739113720), Telegraf is escaping them. This is why the escaping of `=` is removed in [this PR](https://github.com/micrometer-metrics/micrometer/pull/6450).
Even though Telegraf handles escaping, it only replaces `=` with `_` for keys since Telegraf splits the tag key and value based on the first `=`.
Please let me know if I missed anything!

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: The Network makes it easy for both Telegraf and InfluxDB containers to communicate with each other.
See https://java.testcontainers.org/features/networking/

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: I think `.` is missing so I committed myself!

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: As mentioned in my previous comment, `=` in the value works correctly without escaping. Furthermore, we decided not to convert `=` based on [this code](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/statsd/statsd.go#L823-L829) on [this issue](https://github.com/micrometer-metrics/micrometer/issues/6513) as I see🙂, as it appears that `=` is intentionally used in those values.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: In my integration test, the registry takes some time to get ready.
So the metric is not transferred if the `sleep()` is not there.
Please let me know if any other way to check if the registry is ready😀

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: I just wanted to do this test with UDP because it is a default protocol for influxDB.
I changed to TCP and now I can use "telegraf.getFirstMappedPort()"!

pythaac:
Hi @jonatan-ivanov ! I've replied to your comments and updated the code according to your guidance.
Could you please take another look?
Thank you in advance🙂

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilder.java
shakuzen: We should leave a comment in the code (perhaps on `SANITIZE_PATTERN`?) explaining why it doesn't match the docs. Surely in the future we will be confused about this again.

pythaac: @shakuzen , oh yes thank you for the feedback! I've updated as follows:
1. **Added comments** to explain why we sanitize instead of escaping, and why keys are handled differently than values
2. **Renamed** `telegrafEscape()` to `telegrafSanitize()` to better reflect its function
3. **Removed outdated legacy comments** to prevent any future confusion

Please let me know if you have any further feedback🙂

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,190 @@
...
+            }
+
+            @Override
+            public @NonNull String host() {
+                return telegraf.getHost();
+            }
+
+            @Override
+            public int port() {
+                return telegraf.getFirstMappedPort();
```
shakuzen: I think getting the specific mapped port we're looking for would be better. Getting the first mapped port seems more brittle because a future version of the container might have multiple ports mapped.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
shakuzen: Assuming the containers are ready before the tests start (I believe that's the expected behavior), I can only guess that the TCP client is taking some time to connect to the backend. I think this may have come up somewhere else but I can't find the issue/PR now.

I think the only way to directly check the statsd client is connected is to check the package private field `statsdConnection`, which I suppose you could do if the test is moved to the same package as StatsdMeterRegistry.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,190 @@
...
+        .withCopyFileToContainer(MountableFile.forClasspathResource("telegraf-test.conf"),
+                "/etc/telegraf/telegraf.conf")
+        .dependsOn(influxDB)
+        .waitingFor(Wait.forLogMessage(".*Loaded inputs: statsd.*", 1));
+
+    @Issue("#6513")
+    @Test
+    void shouldSanitizeEqualsSignInTagKey() throws InterruptedException {
+        sendMetricWithEqualSign();
+        await().atMost(5, TimeUnit.SECONDS).untilAsserted(() -> verifyEqualsSignMetric());
```
shakuzen: Is the await needed here and elsewhere? Is there some delay between publishing finishing and influx returning the just published metrics?

pythaac: @shakuzen , thank you for your patience and detailed feedback!

Even after the registry is connected, I believe there is some delay after sending the metrics:
1. Delay to reach Telegraf
2. Flushing delay to InfluxDB from Telegraf (I've minimized the interval in the configuration)
3. Delay writing a metric in InfluxDB
4. Latency when querying the API to InfluxDB

In my test, `await()` fails consistently up to 1 second, and becomes flaky around 2 seconds (sometimes passing, sometimes failing). Therefore, I think the current timeout is appropriate to ensure the test stability.

pythaac: ~~I wasn't sure if setting the interval this small was good practice, but setting the Telegraf intervals to 10ms actually made the test pass without `await()`! So I updated the configuration accordingly🙂~~
However, even with the changes above, it still happens that the test fails initially and passes on retry in about 1 out of 10 cases. So I've decided to revert the changes and keep the `await()`.

shakuzen: Thanks for all the details. I wasn't thinking about the delay from Telegraf to InfluxDB when I left my comment (I guess I was thinking direct to InfluxDB ingestion, which of course isn't what a Telegraf specific piece of code is used for). That explains the need for this. To explain this to our future selves looking at this code, a comment nearby or an `alias()` on `await()` would help, I think. Otherwise, it may not be clear to a future reader what expected delay there is.

shakuzen (approved):

implementations/micrometer-registry-statsd/src/main/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilder.java
jonatan-ivanov: The comments added to the PR made things clear, thank you!

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
+/*
+ * Copyright 2025 VMware, Inc.
```
jonatan-ivanov: Sorry, in a previous comment I suggested 2025, this should be 2026:

```suggestion
 * Copyright 2026 VMware, Inc.
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
jonatan-ivanov: I'm not sure that answers my question. I think this works without creating a custom network since containers can communicate through the exposed ports of the host. And you cannot get rid of the exposed ports since you call it through an HTTP client, hence I think the custom network here is unnecessary.

Please check other integration tests (iirc, in the otlp integration tests we have a similar situation: prometheus calls an otlp collector, both are docker containers).

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+import static org.hamcrest.CoreMatchers.containsString;
+
+@Tag("docker")
+@Testcontainers
+class TelegrafStatsdLineBuilderIntegrationTest {
+
+    private static final Network network = Network.newNetwork();
+
+    @Container
+    static GenericContainer<?> influxDB = new GenericContainer<>(DockerImageName.parse("influxdb:latest"))
```
jonatan-ivanov: Using `latest` has the consequence of failing the build without doing anything, not knowing which version we use and making the build non-reproducible.

I think with other tests we made the decision of hardcoding the version but the version comes from a property. Please search for `dockerTest` in all files, you will see that we are setting the `otel-collector-image.version` and the `prometheus.version` for otlp and prometheus integration tests. Maybe we should do the same here.

Instead of `latest` I think we should use:
- `influxdb:2.8.0`
- `telegraf:1.37.1`

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @Test
+    void shouldSanitizeSpaceInTagKeyAndValue() {
+        sendMetricWithSpace();
+        await().alias("Telegraf flushing and InfluxDB ingestion")
+            .atMost(5, TimeUnit.SECONDS)
+            .untilAsserted(() -> verifySpaceMetric());
+    }
+
+    private void sendMetricWithEqualSign() {
```
jonatan-ivanov: I think these `sendMetricWith...` methods are making things less readable to me.
I would either inline them (since they are duplicated code anyways) or maybe I would do something like this:
```java
private void registerMeter(Runnable runnable) {
    StatsdMeterRegistry registry = new StatsdMeterRegistry(getStatsdConfig(), Clock.SYSTEM);
    startRegistryAndWaitForClient(registry);
    runnable.run();
    registry.close();
}

// and in the tests
registerMeter(() -> registry.counter("test,metric,comma", "comma,key", "comma,value").increment());
```

And something similar for the `verify...` methods:
```java
private void verifyMetric(String query, Matcher<?> matcher, Matcher<?>... additionalMatchers) {
        whenGetMetricFromInfluxDb(getFluxQuery(query)).then()
            .statusCode(200)
            .body(matcher, additionalMatchers);
    }
```
What do you think?

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+            .queryParam("org", "my-org")
+            .contentType("application/vnd.flux")
+            .accept("application/csv")
+            .body(fluxQuery)
+            .when()
+            .post("/api/v2/query");
+    }
+
+    private void startRegistryAndWaitForClient(StatsdMeterRegistry meterRegistry) {
+        meterRegistry.start();
```
jonatan-ivanov: Is this needed? I think the registry should start automatically (by its constructor). If so, I would remove this and also rename the method to `waitForClient`.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: Thank you for the guidance🙂 I guess the otlp integration test you mentioned is `OTelCollectorIntegrationTest`.
As you recall, there's a single `otel/opentelemetry-collector-contrib` container in the test, which exposes 2 ports: `4318` and `9090`.
1. Host -> OpenTelemetry Collector (Push Metrics)
2. Host -> OpenTelemetry Collector (Scrape/Verify)

```mermaid
graph TD
    subgraph Host ["Host Machine (Test Runner)"]
        TestCode["Test Code (Micrometer)"]
        RestAssured["RestAssured (Verification)"]
    end

    subgraph Docker ["Docker Environment"]
        subgraph OTelContainer ["OTel Collector Container"]
            Receiver["OTLP Receiver<br>(Port 4318)"]
            Exporter["Prometheus Exporter<br>(Port 9090)"]
        end
    end

    %% Flow 1: Metrics Ingestion
    TestCode -- "1. Push Metrics (HTTP)<br>localhost:RandomPortA -> :4318" --> Receiver

    %% Flow 2: Metrics Scrape
    RestAssured -- "2. Scrape/Verify (HTTP)<br>localhost:RandomPortB -> :9090" --> Exporter

    classDef host fill:#e1f5fe,stroke:#01579b,color:black
    classDef container fill:#fff9c4,stroke:#fbc02d,color:black
    class Host host
    class OTelContainer container
```

In our case, however, we have two containers that must communicate with each other directly.
1. Host -> Telegraf (Push Metric)
3. Telegraf -> InfluxDB (Forward Metrics)
4. Host -> InfluxDB (Query Data)

While the Host knows the random mapped ports for both containers, the Telegraf container needs to know the address of the InfluxDB container to forward metrics. By creating a Network and assigning an alias (influxdb), we allow Telegraf to resolve http://influxdb:8086 internally.

```mermaid
graph TD
    subgraph Host ["Host Machine (Test Runner)"]
        Micrometer["Micrometer Registry"]
        Verifier["RestAssured (Verification)"]
    end

    subgraph Docker ["Docker Environment"]
        subgraph MyNetwork ["Custom Network"]
            subgraph TelegrafContainer ["Telegraf Container"]
                StatsDInput["StatsD Input (Port 8125)"]
                InfluxOutput["InfluxDB Output"]
            end
            
            subgraph InfluxContainer ["InfluxDB Container"]
                InfluxInput["InfluxDB API (Port 8086)"]
            end
        end
    end

    %% Flow 1: Metrics Push
    Micrometer -- "1. Push Metrics (UDP, TCP) <br/> localhost:RandomPort -> :8125" --> StatsDInput

    %% Flow 2: Internal Forwarding
    InfluxOutput -- "2. Forward Metrics (HTTP) <br/> http://influxdb:8086" --> InfluxInput

    %% Flow 3: Verification
    Verifier -- "3. Query Data (HTTP) <br/> localhost:RandomPort -> :8086" --> InfluxInput

    classDef host fill:#e1f5fe,stroke:#01579b,color:black
    classDef container fill:#fff9c4,stroke:#fbc02d,color:black
    classDef network fill:#e0f2f1,stroke:#004d40,stroke-dasharray: 5 5,color:black
    
    class Micrometer,Verifier host
    class TelegrafContainer,InfluxContainer container
    class MyNetwork network
```

Please let me know if this clarifies the necessity of the Network object! 😀

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: Oh my bad. I misunderstood `use a recent version` as using the `latest` tag🥲. I have corrected it!

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: Yes, it would be better that metric information and expected result are visible in the test.
I have updated the code as follows:
(Even the format is changed after applied code formatting using `./gradle format`)

```java
    @Test
    void shouldSanitizeCommaInTagKeyAndValue() {
        registerMeter(registry ->
            Counter.builder("test,metric,comma")
                .tag("comma,key", "comma,value").register(registry).increment()
        );

        await().alias("Telegraf flushing and InfluxDB ingestion")
            .atMost(5, TimeUnit.SECONDS)
            .untilAsserted(() ->
                verifyMetric("test_metric_comma",
                    containsString("comma_key"), containsString("comma_value"))
            );
    }
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/internal/TelegrafStatsdLineBuilderIntegrationTest.java
jonatan-ivanov: In addition, a port check on the container could also help.
(Not sure is the log message of "Loaded inputs..." alone means that the server is listening.)
```
.waitingFor(Wait.forListeningPorts(8125));
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
jonatan-ivanov: Oops, sorry, I misremembered that in the otel tests instead of having Prometheus Server as a separate container we just use the prometheus endpoint of the collector.

As I mentioned above, two containers can communicate with each other without a custom network through the exposed ports of the host (which is already exposed). I think the difficulty without a using a custom network here is telling Telegraf what the randomly exposed Influx port is and since Influx is a `GenericContainer` here, I think an alternative is using the Telegraf support of environment variables in config files so setting `urls = ["${INFLUX_ADDRESS}"]` in `telegraf.conf` and setting `.withEnv("INFLUX_ADDRESS", influxDB.getFirstMappedPort())` should work.

This first seem to be a bit more complicated than just creating a `Network` and using it but the `Network` object also need to be closed and the test is not doing it right now. I'm not sure Testcontainers does it either but I feel it should, maybe ryuk does. I guess `docker network ls` can verify this.

So I guess I would either:
- Use an environment variable to inject the Influx port to the Telegraf config and drop the custom `Network`
- Close the network properly (or verify that it is closed?) and add two comments: 1. We need the network to avoid injecting the influx port 2. The network will be closed by Testcontainers (if this is the case)

To me, using an env var to inject a port seems to be a safer solution right now but if the `Network` is cleaned up by Testcontainers/ryuk and it seems simpler for you, I'm ok using it.

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
```diff
@@ -0,0 +1,195 @@
...
+        .withEnv("DOCKER_INFLUXDB_INIT_BUCKET", "metrics_db")
+        .withEnv("DOCKER_INFLUXDB_INIT_ADMIN_TOKEN", "my-test-token")
+        .waitingFor(Wait.forHttp("/ping").forStatusCode(204));
+
+    @Container
+    static GenericContainer<?> telegraf = new GenericContainer<>(TELEGRAF_IMAGE).withNetwork(network)
+        .withExposedPorts(8125)
+        .withCopyFileToContainer(MountableFile.forClasspathResource("telegraf.conf"), "/etc/telegraf/telegraf.conf")
+        .dependsOn(influxDB)
+        .waitingFor(Wait.forLogMessage(".*Loaded inputs: statsd.*", 1));
```
jonatan-ivanov: Would it make sense to also add this?
```java
.waitingFor(Wait.forListeningPorts(8125));
```

implementations/micrometer-registry-statsd/src/test/java/io/micrometer/statsd/TelegrafStatsdLineBuilderIntegrationTest.java
pythaac: Thank you for the detailed feedback again😀 In conclusion, I have updated the code to explicitly close `Network` object after the test, and added comments as requested.

## Added explicit Network closure
`Network` is automatically closed by the Ryuk as you mentioned(I verified this with my tests with the command you mentioned). However, I agree that explicit cleanup is better, so I have added to close the network after all the test.


## Findings on using mapped ports

If it is not too complicated than using Network, I think using environment variable approach is good as well.
However, it turns out that we cannot retrieve the mapped port of influxDB when initializing Telegraf container.
I attempted to apply the mapped port approach as the follows:

#### telegraf.conf
```
...
[[outputs.influxdb_v2]]
   urls = ["${INFLUX_URL}"]
  token = "my-test-token"
...
```
#### TelegrafStatsdLineBuilderIntegrationTest
```java
// without Network object
    @Container
    static GenericContainer<?> telegraf = new GenericContainer<>(TELEGRAF_IMAGE)
        .withExposedPorts(8125)
        .withCopyFileToContainer(MountableFile.forClasspathResource("telegraf.conf"), "/etc/telegraf/telegraf.conf")
        .dependsOn(influxDB)
        .withEnv("INFLUX_URL", "http://host.testcontainers.internal:" + influxDB.getMappedPort(8086))
        .waitingFor(Wait.forListeningPorts(8125));
```

And It failed with the following exception:
```
        Caused by:
        java.lang.IllegalStateException: Mapped port can only be obtained after the container is started
            at org.testcontainers.shaded.com.google.common.base.Preconditions.checkState(Preconditions.java:513)
            at org.testcontainers.containers.ContainerState.getMappedPort(ContainerState.java:161)
```

Even with `dependOn(influxDB)`, the `.withEnv()` is evaluated at class initialization time. To solve this, I would need to implement more complex code compared to using `Network`.

jonatan-ivanov: Thank you for looking into this! With the mapped port precondition, it indeed seems that using a custom `Network` is a better solution. Thank you for going through all these!

jonatan-ivanov:
Update: I had issues locally that did not encourage me to do this (my docker environment had issues and tests failed and did not want to cause trouble to you by force pushing) but I fixed them so did the rebase.

~~Thank you for the PR, I think it is very near to be ready to get merged in.
Unfortunately we had an issue which resulted docker tests not being executed in our build (including this PR).
This was fixed in https://github.com/micrometer-metrics/micrometer/pull/7107, could you please pull down the latest `main`, rebase your changes on `main`, run `./gradlew dockerTest` to see if they pass (fix them as needed) and force push it?~~

jonatan-ivanov:
@pythaac Thank you very much!

PR #7105, opened 2026-01-22
Untangle circular initialization dependencies

shakuzen:
The public static fields for the no-op singleton on some interfaces (Observation, Scope, ObservationRegistry) could be null due to a circular dependency between the interface and concrete type during class initialization. This backports configuring the ErrorProne check for ClassInitializationDeadlock to help catch these issues. It replaces concrete implementations that could be initialized by reference to their type or fields with anonymous inner classes that can only be initialized as part of initializing the interface type.

See #6946

/cc @etrandafir93

### DIFF
--- a/micrometer-observation/src/test/java/io/micrometer/observation/ObservationNoopTests.java
+++ b/micrometer-observation/src/test/java/io/micrometer/observation/ObservationNoopTests.java
@@ -19,7 +19,7 @@ import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.BDDAssertions.then;
 
-class NoopObservationRegistryTests {
+class ObservationNoopTests {
 
     @Test
     @SuppressWarnings("NullAway")


micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -52,7 +52,72 @@ public interface Observation extends ObservationView {
      * No-op observation. Do not use it to check if an Observation is no-op, use
      * {@code observation.isNoop()} instead.
      */
-    Observation NOOP = NoopObservation.INSTANCE;
+    Observation NOOP = new Observation() {
```
shakuzen, line 56: I did have a version of this that keeps the separate concrete types (e.g. `NoopObservation`) without the singleton field and I think that works too, but I figured this way helps keep us from reintroducing the issue. There's no way to initialize this anonymous inner class so we can't shoot ourselves in the foot. It does make the interface class significantly longer which is unfortunate and I think that's the main reason we had the implementations in package-private separate class files before.

micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservationRegistry.java
```diff
@@ -27,12 +27,16 @@
  */
 class SimpleObservationRegistry implements ObservationRegistry {
 
-    private static final ThreadLocal<Observation.Scope> localObservationScope = new ThreadLocal<>();
+    private static final ThreadLocal<Observation.@Nullable Scope> localObservationScope = new ThreadLocal<>();
```
shakuzen, line 30: `@Nullable` was missing here and on `getCurrentObservationScope` (it's marked nullable on the interface). I can pull the nullable changes out to a separate PR if we want.

micrometer-observation/src/test/java/io/micrometer/observation/NoopObservationRegistryTests.java
```diff
@@ -88,11 +88,11 @@ void shouldNotRespectScopesIfNoHandlersAreRegistered() {
 
         Observation observation = Observation.start("foo", observationRegistry);
         then(observation.isNoop()).isTrue();
-        then(observation).isSameAs(NoopObservation.INSTANCE);
+        then(observation).isSameAs(Observation.NOOP);
         then(observationRegistry.getCurrentObservation()).isNull();
 
         try (Observation.Scope ignored = observation.openScope()) {
-            then(observationRegistry.getCurrentObservationScope()).isSameAs(Observation.Scope.NOOP);
+            then(observationRegistry.getCurrentObservationScope()).isNotNull().isSameAs(Observation.Scope.NOOP);
```
shakuzen: I added a `isNotNull` assertion here and on line 71. Before all the changes in this PR, these tests when run in isolation on my machine would pass *because of* the circular initialization issue that's now fixed with the changes in this PR. That is, `getCurrentObservationScope()` would return `null` and `Observation.Scope.NOOP` was `null` so the assertion would pass. It would fail with the added `isNotNull()` assertion.

They're currently failing because of the `isNotNull()` assertion.

These tests were added in #6700 (see [here](https://github.com/micrometer-metrics/micrometer/pull/6700/changes#diff-47bf9448aee72c0f71925bc09c83a871a570d67f7b3cb8672d74e48d7904a4e7)). @jonatan-ivanov could you take a look and advise what the intention was with these tests?

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -908,6 +990,8 @@ default boolean isNoop() {
     @SuppressWarnings("unchecked")
     class Context implements ContextView {
 
+        static final Context EMPTY = new Context();
```
shakuzen, line 996: I don't know if I like this, since `Context` is mutable, but the field is package-private. This is a replacement of the private static final field that was in `NoopObservation`. We can't have static fields in inner classes because we target Java 8 (it's allowed since Java 16) so I couldn't move that to the anonymous inner class.

micrometer-observation/src/main/java/io/micrometer/observation/NoopButScopeHandlingObservation.java
```diff
@@ -23,13 +26,72 @@
...
+    }
+
+    @Override
+    public void stop() {
+    }
 
     @Override
     public Scope openScope() {
-        return new SimpleObservation.SimpleScope(NoopObservationRegistry.FOR_SCOPES, this);
+        return new SimpleObservation.SimpleScope(ObservationRegistry.NOOP, this);
```
shakuzen, line 94: I got rid of the need for the `FOR_SCOPES` registry in the no-op registry. Since the implementation is based on ThreadLocal, what we need can be achieved with static calls on SimpleObservationRegistry

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
jonatan-ivanov: I'm thinking if we should leave a comment here and the other anonymous classes so no one wants to refactor this to a separate class in the future.

micrometer-observation/src/test/java/io/micrometer/observation/NoopObservationRegistryTests.java
jonatan-ivanov: I think I copied the existing test and messed/mixed up when I modified this:
```java
then(observationRegistry.getCurrentObservationScope()).isSameAs(scope);
```
to this:
```java
then(observationRegistry.getCurrentObservationScope()).isSameAs(Observation.Scope.NOOP);
```
Since the scope in the try-with-resources block (what `observation.openScope()` returns) is `Scope.NOOP`, I think I mistakenly asserted that against `getCurrentObservationScope()` which should return `null` here.

I pushed [aa5b016](https://github.com/micrometer-metrics/micrometer/pull/7105/commits/aa5b016b3cd803abd0fd6bd473b0d7cf9bd2f393) on this PR to fix/improve these assertions, please let me know if it makes sense.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
jonatan-ivanov: I think this isn't worse what we had before and I think we should fix the root issue (maybe in a different PR, also in `1.15x`?).

This can be a source of memory leaks if the user gets a noop Observation and writes the context directly:
```java
Observation observation = Observation.start("test", ObservationRegistry.NOOP);
assertThat(observation.getContext()).isSameAs(Observation.Context.EMPTY);
observation.getContext().put("key", "value"); // this should not do anything
assertThat(observation.getContext().<String>get("key")).isEqualTo("value"); // this should fail
```

We could have an "EmptyContext" that overrides the mutators with empty implementations.

micrometer-observation/src/test/java/io/micrometer/observation/NoopObservationRegistryTests.java
jonatan-ivanov: Should we rename this class from `NoopObservationRegistryTests` to `NoopObservationTests` since `NoopObservationRegistry` is now gone?

jonatan-ivanov:
I executed the benchmarks to make sure JIT still eliminates noop observation and it seems so:
Before (`1.16.x`):
```
Benchmark                                                Mode  Cnt     Score   Error   Units
ObservationBenchmark.baseline                            avgt    2     0.275           ns/op
ObservationBenchmark.baseline:gc.alloc.rate              avgt    2     0.017          MB/sec
ObservationBenchmark.baseline:gc.alloc.rate.norm         avgt    2    ≈ 10⁻⁶            B/op
ObservationBenchmark.baseline:gc.count                   avgt    2       ≈ 0          counts
ObservationBenchmark.noopObservation                     avgt    2     1.064           ns/op
ObservationBenchmark.noopObservation:gc.alloc.rate       avgt    2     0.017          MB/sec
ObservationBenchmark.noopObservation:gc.alloc.rate.norm  avgt    2    ≈ 10⁻⁵            B/op
ObservationBenchmark.noopObservation:gc.count            avgt    2       ≈ 0          counts
ObservationBenchmark.observation                         avgt    2  1147.073           ns/op
ObservationBenchmark.observation:gc.alloc.rate           avgt    2  6067.838          MB/sec
ObservationBenchmark.observation:gc.alloc.rate.norm      avgt    2  1907.258            B/op
ObservationBenchmark.observation:gc.count                avgt    2   192.000          counts
ObservationBenchmark.observation:gc.time                 avgt    2    80.000              ms
```
After (this PR):
```
Benchmark                                                Mode  Cnt     Score   Error   Units
ObservationBenchmark.baseline                            avgt    2     0.274           ns/op
ObservationBenchmark.baseline:gc.alloc.rate              avgt    2     0.017          MB/sec
ObservationBenchmark.baseline:gc.alloc.rate.norm         avgt    2    ≈ 10⁻⁶            B/op
ObservationBenchmark.baseline:gc.count                   avgt    2       ≈ 0          counts
ObservationBenchmark.noopObservation                     avgt    2     0.905           ns/op
ObservationBenchmark.noopObservation:gc.alloc.rate       avgt    2     0.017          MB/sec
ObservationBenchmark.noopObservation:gc.alloc.rate.norm  avgt    2    ≈ 10⁻⁵            B/op
ObservationBenchmark.noopObservation:gc.count            avgt    2       ≈ 0          counts
ObservationBenchmark.observation                         avgt    2  1078.240           ns/op
ObservationBenchmark.observation:gc.alloc.rate           avgt    2  6340.571          MB/sec
ObservationBenchmark.observation:gc.alloc.rate.norm      avgt    2  1866.347            B/op
ObservationBenchmark.observation:gc.count                avgt    2   201.000          counts
ObservationBenchmark.observation:gc.time                 avgt    2    83.000              ms
```
See `noopObservation` (1.064 ns/op and 0.905 ns/op). `baseline` is there to show what does it look like when nothing happens and `observation` is there to show to show what does it look like when an observation is created and used.

micrometer-observation/src/main/java/io/micrometer/observation/NoopButScopeHandlingObservation.java
```diff
@@ -23,13 +26,72 @@
  * @author Tommy Ludwig
  * @author Marcin Grzejszczak
  */
-final class NoopButScopeHandlingObservation extends NoopObservation {
+final class NoopButScopeHandlingObservation implements Observation {
+
+    static final NoopButScopeHandlingObservation INSTANCE = new NoopButScopeHandlingObservation();
```
shakuzen, line 31: This looks like the same problem we had elsewhere, but there is no field on Observation being assigned to this - it is referenced from method bodies and one test, so we won't end up with an unexpected null value in a field on Observation depending on initialization order.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
shakuzen: Yeah, that's better. Let's follow-up with something for that. I've added a TODO comment to the code here.

micrometer-observation/src/test/java/io/micrometer/observation/NoopObservationRegistryTests.java
shakuzen: We could, but NoopObservation is also gone now 😄 
I've renamed to `ObservationNoopTests` to try to capture that this is testing various no-op behavior related to Observation.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
shakuzen: I've added a comment everywhere that was updated to use anonymous classes.

micrometer-observation/src/main/java/io/micrometer/observation/SimpleObservationRegistry.java
shakuzen: I'll leave it as part of this. I think the TL could/should be extracted to its own (package-private) class that encapsulates the logic for getting and setting the current scope/observation. Then SimpleObservationRegistry and other places can use that.

micrometer-observation/src/main/java/io/micrometer/observation/NoopButScopeHandlingObservation.java
shakuzen: Follow up on this, see: https://github.com/micrometer-metrics/micrometer/pull/7105/changes#r2720044344

shakuzen:
> `observation` is there to show to show what does it look like when an observation is created and used.

It's curious that the normalized allocation is reduced in the `observation` benchmark. I wouldn't have expected any change there. I wonder if that's real or run-to-run variance. I'm not sure why the normalized allocation should have run-to-run variance, though.

shakuzen:
We all missed it multiple times. Hopefully it stays fixed now.

micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
jonatan-ivanov: Follow-up PR: https://github.com/micrometer-metrics/micrometer/pull/7133

build.gradle
```diff
@@ -93,6 +93,7 @@ subprojects {
                     "BadImport",
                     "CatchAndPrintStackTrace",
                     "ClassCanBeStatic",
+                    "ClassInitializationDeadlock",
```
etrandafir93: @shakuzen maybe it's a dummy question, but why are the errorprone rules defined here as well - won't they  be inherited from the [root module](https://github.com/micrometer-metrics/micrometer/blob/main/build.gradle#L103)

PR #7132, opened 2026-01-28
Upgrade docker images

jonatan-ivanov:


implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/AbstractElasticsearchMeterRegistryIntegrationTest.java
```diff
@@ -47,7 +47,9 @@ abstract class AbstractElasticsearchMeterRegistryIntegrationTest {
     // https://www.elastic.co/guide/en/elasticsearch/reference/7.8/index-templates.html
     protected static final String VERSION_7 = "7.7.1";
 
-    protected static final String VERSION_8 = "8.6.2";
+    protected static final String VERSION_8 = "8.19.10";
```
jonatan-ivanov, line 50: I'm thinking if we should drop `8.x` tests and only keep `9.x`.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetricsIntegrationTest.java
```diff
@@ -156,7 +156,7 @@ private RequiredSearch get(final String name) {
     }
 
     private static DockerImageName getDockerImageName() {
-        return DockerImageName.parse("postgres:9.6.24");
+        return DockerImageName.parse("postgres:18.1");
```
jonatan-ivanov: We also have [`PostgreSQL17DatabaseMetricsIntegrationTest`](https://github.com/micrometer-metrics/micrometer/blob/d768d58da36a8ed1b3d8acbfc4a855d8023cf9b2/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/db/PostgreSQL17DatabaseMetricsIntegrationTest.java) which is an exact copy of this except the image name.

I think we can drop `17.x` but if we want to keep both, I would consolidate the code like we have for ElasticSearch.

shakuzen: I don't remember what the exact intention was, but maybe we should verify our instrumentation works with whatever the oldest version of Postgres we support (we could document this somewhere) and the latest version. Updating it to work more like the ElasticSearch tests sounds good to me.

implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/AbstractElasticsearchMeterRegistryIntegrationTest.java
shakuzen: I guess it depends on what the support policy is on the Elastic side and what users are typically doing in terms of upgrades. I have no insight into this, but my intuition is that many organizations upgrade their applications and those dependencies more frequently than they typically upgrade their data store versions, which means there might still be a lot of 8.x users.

jonatan-ivanov: According to the [support policy](https://www.elastic.co/support/eol#prior-versions):
- `7.17.x`'s end of support: 2026-01-15 (just ended)
- `8.x`'s end of support: 2027-07-15

We are testing with 7.7.x (a pre-7.8 version) to verify the behavior without index template support.

I would keep all three for now (even if 7.7 was reached EOL probably a while ago) and I think we can drop `7.x` in the next update round.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetricsIntegrationTest.java
jonatan-ivanov: Tests for 17 were added because stats were renamed: https://github.com/micrometer-metrics/micrometer/pull/6825

I upgraded `9.x` to `14.x` (oldest supported version) and `17.x` to `18.x` (latest).

shakuzen: I wonder if we could easily reference these versions in the docs somehow so we can add a statement or table somewhere that shows what versions we test the instrumentation against, so users have an idea of what the expectation is for instrumentation compatibility.

shakuzen (approved):

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/db/PostgreSQLDatabaseMetricsIntegrationTest.java
jonatan-ivanov: The easiest is using the asciidoc tags that we using right now in other places but that means this will be in the docs as code, e.g.:
>Currently our CI runs integration tests with the following versions:
```
protected static final String VERSION_14 = "14.20";

protected static final String VERSION_18 = "18.1";
```
Which has the information but it's not pretty.
(Does not say which image though, but not sure it must.)

We might also be able to use the version catalog for this:
```toml
[docker-images]
postgres14 = "postgres:14.20"
postgres18 = "postgres:18.1"
```
And inject it from Gralde, like we do for other tests (except we can inject the whole image name not only the version):
```groovy
dockerTest {
    systemProperty 'postgres14.image', dockerImages.postgres14
    systemProperty 'postgres18.image', dockerImages.postgres18
}
```
The docs might be nicer:
>Currently our CI runs integration tests with the following versions:
```toml
postgres14 = "postgres:14.20"
postgres18 = "postgres:18.1"
```
The docker images will be simpler to upgrade since everything will be there in the catalog. On the other hand, tests are not as easy to run from IDEs since the property needs to be defined in the run config but we have this issue for some tests already. And also, that part of the version catalog will be a bit uglier since the added asciidoc tags.

jonatan-ivanov: I created https://github.com/micrometer-metrics/micrometer/issues/7140

PR #7133, opened 2026-01-28
Add immutable noop Observation.Context

jonatan-ivanov:
Before this change the `Observation.Context` instance used for noop observations were mutable.
This can be a source of memory leaks if the user gets a noop Observation and writes the context directly:
```java
Observation observation = Observation.start("test", ObservationRegistry.NOOP);
assertThat(observation.getContext()).isSameAs(Observation.Context.EMPTY);
observation.getContext().put("key", "value"); // this should not do anything
assertThat(observation.getContext().<String>get("key")).isEqualTo("value"); // this should fail
```
This change introduces an immutable and noop `Observation.Context` implementation.

See https://github.com/micrometer-metrics/micrometer/pull/7105#discussion_r2715764858

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -1043,6 +1043,7 @@ public interface Observation extends ObservationView {
          * @return the previous value associated with the key, or null if there was no
          * mapping for the key
          */
+        @Nullable
         public Object remove(Object key) {
             return this.map.remove(key);
         }
--- a/micrometer-observation/src/main/java/io/micrometer/observation/NoopContext.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/NoopContext.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2026 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.observation;
+
+import io.micrometer.common.KeyValue;
+import io.micrometer.common.KeyValues;
+import io.micrometer.common.lang.Nullable;
+
+import java.util.function.Function;
+
+/**
+ * A special {@link Observation.Context} that should be used only in special cases where
+ * Observations are disabled/noop. It is an immutable implementation, it will not modify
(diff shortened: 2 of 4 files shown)


micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
```diff
@@ -1121,7 +1118,7 @@ public <T> Context put(Object key, T object) {
          * @return the previous value associated with the key, or null if there was no
          * mapping for the key
          */
-        public Object remove(Object key) {
+        public @Nullable Object remove(Object key) {
```
jonatan-ivanov: This caused an issue in `EmptyContext` (it returns `null`) but also: `Map#remove` can return `null` if there was no entry for the key.

micrometer-observation/src/main/java/io/micrometer/observation/EmptyContext.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.function.Function;
+
+/**
+ * A special {@link Observation.Context} that should be used only in special cases where
+ * Observations are disabled/noop. It is an immutable implementation, it will not modify
+ * the internals of the context.
+ *
+ * @author Jonatan Ivanov
+ */
+class EmptyContext extends Observation.Context {
```
shakuzen: I know I used the "empty" naming before, but on more thought, I think NoopContext is probably better for the class name. It's a bit unfortunate that all the (private) fields from Context will be present even though they aren't needed for a no-op implementation. In that sense, it'd be nice if Context were an interface. Not that I think we can do anything about that now.

micrometer-observation/src/main/java/io/micrometer/observation/EmptyContext.java
```diff
@@ -0,0 +1,114 @@
...
+
+    static final EmptyContext INSTANCE = new EmptyContext();
```
shakuzen: The class is package-private, so it's not critical, but should we make the constructor private so only the one instance gets used? I don't think there's any reason we'd want different instances.

micrometer-observation/src/main/java/io/micrometer/observation/EmptyContext.java
marcingrzejszczak: I think `NoOp` was indeed used throughout the codebase

jonatan-ivanov:
>Thanks for this. We should do this in the maintenance branches, right?

When I started I was not sure but looking at how it ended up, I think so, I moved to `1.15.x`.

micrometer-observation/src/main/java/io/micrometer/observation/EmptyContext.java
jonatan-ivanov: I renamed it to `NoopContext`.

micrometer-observation/src/main/java/io/micrometer/observation/EmptyContext.java
jonatan-ivanov: I made it private.

jonatan-ivanov:
Forward merge will be a bit more adventurous, let me know if it looks ok and I will merge it.

shakuzen (approved):

PR #7134, opened 2026-01-28
Revert version for Logback 1.2.x to 1.2.13

izeye:
This PR reverts the version for Logback 1.2.x to 1.2.13 as it seems that #6770 has changed it to 1.5.26 accidentally.


jonatan-ivanov:
Thank you!

jonatan-ivanov:
@shakuzen Not sure what to do with this:
- We can add a test that fails if the wrong version is upgraded
- We can disable the dependency 😬
- I opened an issue for dependabot: https://github.com/dependabot/dependabot-core/issues/14040

shakuzen:
> We can add a test that fails if the wrong version is upgraded

This sounds interesting. It should prevent accidental merging and the failure message could make clear the issue.

izeye:
I created https://github.com/micrometer-metrics/micrometer/pull/7138 to try to apply the suggestion from @jonatan-ivanov.

jonatan-ivanov:
Reapplied in https://github.com/micrometer-metrics/micrometer/commit/1d8d4a6328c1339d5a88c52a501747f072ef9ed2

PR #7138, opened 2026-01-29
Add dependency version verification task for legacy versions

izeye:
This PR adds a verification task for the legacy Logback version.

See https://github.com/micrometer-metrics/micrometer/pull/7134#issuecomment-3814611122

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -486,6 +486,26 @@ nexusPublishing {
     }
 }
 
+tasks.register('verifyDependencyVersions') {
+    doLast {
+        assertDependencyVersionPrefix(libs.jooq, '3.14.')
+        assertDependencyVersionPrefix(libs.logback12, '1.2.')
+    }
+}
+
+private static void assertDependencyVersionPrefix(def dependencySupplier, def expectedVersionPrefix) {
+    def dependency = dependencySupplier.get()
+    assert dependency.version.startsWith(expectedVersionPrefix)
+}
+
+tasks.register('check') {
+    dependsOn verifyDependencyVersions
+}
+
+tasks.register('build') {
+    dependsOn check
+}
+
 wrapper {
     gradleVersion = '8.14.3'
 }


build.gradle
```diff
@@ -467,9 +467,16 @@ subprojects {
                     onlyIf { compatibleVersion != 'SKIP' }
                 }
 
+                tasks.register("verifyLogbackLegacyVersion") {
+                    doLast {
+                        assert libs.logback12.get().version == "1.2.13"
```
jonatan-ivanov: I'm thinking about three things here:
1. Can we do `.startsWith('1.2.')` instead of equals?

2. I'm also thinking to make this a bit more general, since we have other libs with this problem (they usually fail the build for other reasons), like JOOQ (we can add those here or in a separate PR).
Something like this:
```groovy
tasks.register('verifyVersions') {
    doLast {
        verifyDependencyVersion(libs.logback12, '1.2.')
    }
}

private void verifyDependencyVersion(def dependencySupplier, String expectedVersionPrefix) {
    def dependency = dependencySupplier.get()
    assert dependency.getVersion().startsWith(expectedVersionPrefix)
}
```
(Dependency is a separate variable here to make the error message from assert nicer: it won't print out the supplier, only the dependency, its version and the expected prefix)

3. At this place, this task is registered and executed for many subprojects but since we verify the (global) version catalog, this should be done only once on the root project I think. But this complicates things in another way since the root project does not have the `check` or `build` tasks so we need to add those manually:
```groovy
tasks.register('check') { dependsOn verifyVersions }
tasks.register('build') { dependsOn check }
```

jonatan-ivanov: Oh also, can we do this starting from `1.15.x`?

jonatan-ivanov:
After this is done, we need  to roll back the version again. :(
https://github.com/micrometer-metrics/micrometer/pull/7142

jonatan-ivanov:
Oh, here's the other example: https://github.com/micrometer-metrics/micrometer/pull/7143

izeye:

I applied your suggestions.

> 1. Can we do `.startsWith('1.2.')` instead of equals?

For reference, I checked equality as Logback 1.2.x has been EOLed.

See https://logback.qos.ch/news.html

jonatan-ivanov:
Thank you very much!
Valid point with logback (and probably the older JOOQ version will not be upgraded either), I think but the prefix might be useful elsewhere.

jonatan-ivanov:
After I reverted the version, there was an immediate catch: https://github.com/micrometer-metrics/micrometer/pull/7144

jonatan-ivanov:
That would be a different mechanism, since it should check the resolved version instead of the version catalog. Would not it be simpler to move those dependencies to the version catalog too?

shakuzen:
As discussed offline, the dependencies were intentionally not in the catalog to try to avoid dependabot changing them when we don't want it to. However, that doesn't seem to work (anymore? did it ever?), and so I don't think there's a reason to keep them out of the catalog. So what you suggest is fine for the purpose of this check. For other purposes, it might still be interesting to check the resolved version since that's what matters if we're trying to test a specific version of a dependency and other dependencies can influence resolution. That's somewhat a separate matter to ensuring dependabot doesn't upgrade versions we don't want it to upgrade.

Issue #6502, opened 2025-07-02
Rework KafkaMetrics based on Kafka's MetricsReporter


shakuzen:
See the [MetricsReporter](https://kafka.apache.org/40/javadoc/org/apache/kafka/common/metrics/MetricsReporter.html) interface in Kafka. This looks promising that we could use this to instrument Kafka metrics in a better way than we currently do with our KafkaMetrics.

The-East-Wind:
@shakuzen I'm interested in working on this issue. Can you assign this to me if no one else is working on this ?

jonatan-ivanov:
@The-East-Wind I assigned it you, please let us know if you have any questions.

spryslmatej:
This would be great!

shakuzen:
@The-East-Wind have you had a chance to look at this? Any investigation you can share or questions about implementing this?

parkgeonhu:
@shakuzen @jonatan-ivanov @The-East-Wind 
Hi, I noticed there hasn't been much activity on this issue recently, so I decided to look into it a few days ago. I’ve just submitted a PR(#7146) to address this. Could you please review it? Thanks!

reachvamsikrishnad-lgtm:
Let me know if this is still an issue needs solution fix on this ?

PR #7146, opened 2026-01-31
Rework KafkaMetrics based on Kafka's MetricsReporter

parkgeonhu:
## Summary

This PR introduces `KafkaMetricsReporter`, a new implementation that leverages Kafka's native `MetricsReporter` interface to collect Kafka client metrics into Micrometer.

## Motivation

The existing `KafkaMetrics` class relies on a polling-based approach that requires:
- A `Supplier` to periodically fetch metrics from Kafka clients
- A scheduled executor to run `checkAndBindMetrics()` at regular intervals
- Manual cleanup and meter lifecycle management

By implementing Kafka's `MetricsReporter` interface directly, we can:
- Receive metric changes in real-time via callbacks (`metricChange`, `metricRemoval`)
- Eliminate the need for polling and scheduled executors
- Simplify integration with Kafka clients through the `metric.reporters` configuration

## Usage

```java
// Set the registry before creating Kafka clients
KafkaMetricsReporter.setDefaultMeterRegistry(meterRegistry);

// Configure Kafka client
Properties props = new Properties();
props.put(ConsumerConfig.METRIC_REPORTER_CLASSES_CONFIG, 
          KafkaMetricsReporter.class.getName());
// ... other configs

KafkaConsumer<String, String> consumer = new KafkaConsumer<>(props);
```

## Discussion: Usage Pattern

I'd like to get feedback on the usage pattern for the new `MetricsReporter` implementation.

Unlike the existing `KafkaClientMetrics` binder, which wraps an already instantiated Producer/Consumer, the `MetricsReporter` interface is designed to be instantiated and managed by the Kafka client's lifecycle.

Kafka clients instantiate reporters via reflection inside their constructor based on the `metric.reporters` property. Once the client is initialized, there is no public API to inject or add a reporter dynamically.

As a result, users must configure the reporter class via properties **before** creating the client instance. The "post-construction binding" pattern (e.g., `binder.bindTo(registry)`) is not supported with the `MetricsReporter` architecture.

I'd appreciate any feedback on whether this implementation direction aligns with the project's goals.

If there are concerns or suggestions for a different approach, please let me know—I'm happy to revise the implementation accordingly.

---
See https://github.com/micrometer-metrics/micrometer/issues/6502

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetricsReporter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetricsReporter.java
@@ -1,0 +1,340 @@
+/*
+ * Copyright 2020 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.core.instrument.binder.kafka;
+
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
+import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
+import io.micrometer.core.instrument.*;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.metrics.KafkaMetric;
+import org.apache.kafka.common.metrics.Measurable;
+import org.apache.kafka.common.metrics.MetricsContext;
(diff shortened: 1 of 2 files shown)


shakuzen:
Thanks for all the work on this. I'm distracted by other work at the moment so I haven't looked closely at things, but the usage pattern described is not ideal. I didn't notice this limitation when opening the enhancement issue. Without looking into it more, I don't have an idea of what could/should be changed to improve the situation. It's unfortunate the Kafka client seemingly doesn't support configuring an instance of MetricsReporter. I'll come back to look into this more later, but I wanted to leave some brief feedback in the meantime.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetricsReporter.java
```diff
@@ -0,0 +1,340 @@
...
+    public static @Nullable MeterRegistry getDefaultMeterRegistry() {
+        return DEFAULT_REGISTRY.get();
+    }
+
+    /**
+     * Called by Kafka to configure this reporter with client properties.
+     * @param configs the Kafka client configuration
+     */
+    @Override
+    public void configure(Map<String, ?> configs) {
```
shakuzen, line 101: In working on some other Kafka-related instrumentation, I've used the Kafka config map to pass an `ObservationRegistry` instance. I wonder if we could do the same here for `MeterRegistry` rather than the static calls.

PR #7156, opened 2026-02-04
Remove unused length parameter from Tags/KeyValues#isSortedSet

belugabehr:
## Summary

Remove the unused `length` parameter from `isSortedSet()` in both `Tags` and `KeyValues` classes.

The `length` parameter was never used with a value different from the array's actual length, making it superfluous. This simplification also improves API consistency:

- `Arrays.sort(tags)` - operates on the full array
- `dedup(tags)` - operates on the full array
- `isSortedSet(tags)` - now also operates on the full array (previously accepted a length parameter that was always `tags.length`)

### Changes

- **Tags.java**: Simplified `isSortedSet(Tag[] tags, int length)` to `isSortedSet(Tag[] tags)`
- **KeyValues.java**: Simplified `isSortedSet(KeyValue[] keyValues, int length)` to `isSortedSet(KeyValue[] keyValues)`
- Removed dead code: the check `if (length > tags.length)` could never be true since `length` was always equal to `tags.length`
- Updated `toTags()` and `toKeyValues()` methods accordingly

### DIFF
--- a/micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/KeyValues.java
@@ -64,18 +64,13 @@ public final class KeyValues implements Iterable<KeyValue> {
     }
 
     /**
-     * Checks if the first {@code length} elements of the {@code keyValues} array form an
-     * ordered set of key-values.
+     * Checks if the {@code keyValues} array forms an ordered set of key-values.
      * @param keyValues an array of key-values.
-     * @param length the number of elements to check.
-     * @return {@code true} if the first {@code length} elements of {@code keyValues} form
-     * an ordered set; otherwise {@code false}.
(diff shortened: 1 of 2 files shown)


belugabehr:
@shakuzen - any interest in this one? Thanks!

shakuzen:
Thanks for the pull request. I have been busy with other things but I'll come back to this when I have some time. Generally it looks good. I wonder if something changed that made the length argument superfluous or it always was.

shakuzen:
To answer my own question, it was always superfluous since it was added in https://github.com/micrometer-metrics/micrometer/commit/1482cdf46babc40273d24ebf572b0b15aec7bfd2. Looking at https://github.com/micrometer-metrics/micrometer/pull/4959#discussion_r1611139172, it was used in an assert that could have a different length value than the array length, but the assert was removed. So I think that explains how it was missed that the length argument wasn't needed here.

PR #7168, opened 2026-02-06
Allow Prometheus meters with the same name but a different set of tag keys

### LINKED ISSUE #877: Support meters with the same name but different set of tag keys in PrometheusMeterRegistry

shakuzen:
Prometheus 1.5.0 removes the restriction that prevented this. See https://github.com/prometheus/client_java/issues/696.

Accordingly, no longer fail to register such meters.

Disable by default the ObservationValidator capability that checks for this scenario. It can still be enabled if desired.

Resolves #877

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
+++ b/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
@@ -31,6 +31,7 @@ import io.prometheus.metrics.model.snapshots.*;
 import io.prometheus.metrics.tracer.common.SpanContext;
 import org.assertj.core.api.Condition;
 import org.jspecify.annotations.Nullable;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
@@ -90,6 +91,7 @@ class PrometheusMeterRegistryTest {
     }
 
     @Test
+    @Disabled("We don't detect this situation yet; scrape will fail")
     void differentTypesThatProduceSamePrometheusMetricFamilyFailsToRegister() {
         AtomicBoolean failed = new AtomicBoolean(false);
         registry.config().onMeterRegistrationFailed((name, reason) -> failed.set(true));
@@ -97,7 +99,7 @@ class PrometheusMeterRegistryTest {
         Timer.builder("test").register(registry).record(Duration.ofMillis(10));
         assertThat(failed.get()).isFalse();
         Gauge.builder("test_seconds_max", new AtomicInteger(42), AtomicInteger::get).register(registry);
-        assertThat(failed.get()).isTrue(); // TODO this assertion fails
+        assertThat(failed.get()).isTrue();
     }
 
     @Test


jonatan-ivanov:
Related to this, should not we change the wording here: https://github.com/micrometer-metrics/micrometer/blob/8e953145b12cd88510da25e2578bbf78fc75c0e2/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java#L154-L158

Something similar to the change in the docs.

shakuzen:
> Related to this, should not we change the wording here

Good catch. I forgot about that code. We might need more than a wording change. Now that the Prometheus registry won't fail because of this, I wonder if the capability should be disabled by default.

jonatan-ivanov:
I still find it useful since Prometheus still strongly discourages users from creating meters having the same name with a different set of tag keys so if someone wants to keep the instrumentation tidy, the validator can help. IIRC that option is off by default so people need to explicitly ask for it.

zeitlinger:
Please tell me when you're fine with the change so we can create the release.

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
jonatan-ivanov: I think this can be removed, `metersWithSameNameAndDifferentTagsCanScrape` checks this.

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
```diff
@@ -70,42 +71,45 @@ void metersWithSameNameAndDifferentTagsContinueSilently() {
...
+        registry.counter(meterName, "k2", "v2");
+        registry.counter(meterName, "k3", "v3");
+        assertThat(registry.scrape()).isEqualTo("# HELP my_counter_total  \n" + "# TYPE my_counter_total counter\n"
+                + "my_counter_total{k1=\"v1\"} 0.0\n" + "my_counter_total{k2=\"v2\"} 0.0\n"
+                + "my_counter_total{k3=\"v3\"} 0.0\n");
+    }
+
+    @Test
+    void sameNameDifferentTagsWithEmptyTagsDoesNotFail() throws InterruptedException {
+        AtomicBoolean failed = new AtomicBoolean(false);
```
jonatan-ivanov, line 107: Instead of checking the `AtomicBoolean` and `doesNotThrowAnyException`, can we just tell the registry to always throw an exception:
```java
private final PrometheusMeterRegistry registry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT,
            prometheusRegistry, clock).throwExceptionOnRegistrationFailure();
```
and in the tests we would just do normal meter registrations, e.g. this test would simplify to:
```java
registry.counter("my.counter");
registry.counter("my.counter", "test.k1", "v1").increment();
registry.counter("my.counter", "test.k2", "v2");
registry.counter("my.counter", "test.k3", "v3");
```
And it would also help with other tests (except `registerAndScrapeGaugeAndCounterWithSameNameSameTagKeyDifferentTagValueNoException` will need a new registry instance).

jonatan-ivanov:
@zeitlinger We are looking into this. I found a few differences in the behavior (because of the Prometheus change and because of this PR) and we need to discuss what to do them. Right now I think the Prometheus changes are probably ok but could you please give us some time till Monday (your Tuesday)?

zeitlinger:
> @zeitlinger We are looking into this. I found a few differences in the behavior (because of the Prometheus change and because of this PR) and we need to discuss what to do them. Right now I think the Prometheus changes are probably ok but could you please give us some time till Monday (your Tuesday)?

yes, sure :smile:

shakuzen:
> IIRC that option is off by default so people need to explicitly ask for it.

The code seems to indicate it is on by default and people need to explicitly disable it.
https://github.com/micrometer-metrics/micrometer/blob/a4e0866620ebd2f25a8b788f344a8d9b17074fd9/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/TestObservationRegistry.java#L147

shakuzen:
I updated the ObservationValidator message and disabled the capability by default. Let me know what you think @jonatan-ivanov .

Remaining TODO is to catch registration of metrics that will end up as duplicates in Prometheus and fail to register after the first. Currently cases like the following result in an exception when scraping, as per @jonatan-ivanov's investigation.

```java
registry.counter("test").increment();
registry.counter("test.total").increment(42);
```

Before: only the second one is registered (counter_total 42.0), no error
After: io.prometheus.metrics.model.snapshots.DuplicateLabelsException: Duplicate labels for metric "test": {}
Desired: only the first is registered, the second results in a call to `meterRegistrationFailed`.

We should document this somewhere. Another similar issue:

```java
Timer.builder("test").register(registry).record(Duration.ofMillis(10));
Gauge.builder("test_seconds_max", () -> 42).register(registry);
```

This doesn't fail to register but fails on scrape.
Before: java.lang.IllegalStateException: test_seconds_max: duplicate metric name.
After: io.prometheus.metrics.model.snapshots.DuplicateLabelsException: Duplicate labels for metric "test_seconds_max": {}
It would be better to fail to register the second meter like the desired behavior mentioned in the above scenario.

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
shakuzen: I updated the tests and simplified some things. Let me know what you think. I would rather not register the failure listener for tests that aren't testing that, since it's not the default for the registry. We also have more tests now that specifically are testing meter registration failure.

shakuzen:
For now I've disabled the test for the second scenario mentioned above:

```java
Timer.builder("test").register(registry);
Gauge.builder("test_seconds_max", () -> 42).register(registry);
```

It is harder to detect this in an efficient manner at registration time. We might revisit this later. For now, it will continue to fail at scrape time, which is the existing behavior. The exception thrown by the Prometheus client has changed, however. We will document this limitation and advise instrumentation that does this should be fixed.

We'll have to figure out what should go in a migration guide for upgrading to Micrometer 1.17.0 and what should go in the documentation for our Prometheus registry.

PR #7177, opened 2026-02-09
Downgrade Netty from 4.2.10 to 4.1.131

shakuzen:
Downgrade to match the version of netty that reactor-netty depends on. We can upgrade Netty when we upgrade the Reactor train. Netty is used as an optional dependency in micrometer-core for instrumentation, so it is not a transitive dependency for users. It is also used in the Statsd registry but it is shaded so again there should be no impact to users' dependencies.

Adds a separate entry in the version catalog for Netty 4.2 which is still used in the docs tests.

This was commented on in https://github.com/micrometer-metrics/micrometer/pull/6532 but it seems we forgot to follow up on it. It was raised again recently, which is what reminded me to finally do something about it.

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetricsTests.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/netty4/NettyEventExecutorMetricsTests.java
@@ -26,10 +26,8 @@ import org.junit.jupiter.api.Test;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


shakuzen:
I had to undo the parts of https://github.com/micrometer-metrics/micrometer/pull/6621 and https://github.com/micrometer-metrics/micrometer/pull/6612 that were using types only available in Netty 4.2 in tests. I left using Netty 4.2 in the docs tests. And I polished the tests a bit to use ~try-with-resources~ try-finally blocks. I tried using try-with-resources blocks, but those don't work with Java 11/17 because ExecutorService is not AutoCloseable in them (it is in later versions).

PR #7196, opened 2026-02-12
fix measure parameter name used in distribution average

kamilszymanski:


jonatan-ivanov:
As I mentioned in https://github.com/micrometer-metrics/micrometer-docs/pull/379#issuecomment-3881560726, the original version makes sense to me.

First, it talks about avg latency:
> This is a useful measure of average latency.

(`totalTime / count`)
And then it applies this in general (distribution summaries) which is not time based so `totalTime` should not be used there but you can say the average of a distribution summary is `totalAmount / count`:
>When the same idea is applied to the `totalAmount` and `count` emanating from distribution summaries, the measure is called a distribution average. 

Timer: `totalTime / count`
Distribution Summary: `totalAmount / count`

Does it makes sense? What should we change to make it more clear?

Issue #3795, opened 2023-05-01
OTLP exemplar support


shakuzen:
Similar to the exemplar support we added for the Prometheus registry, we can add exemplar support for the OTLP registry. See https://opentelemetry.io/docs/reference/specification/metrics/data-model/#exemplars

0x006EA1E5:
Hi,

Is there are movement on this?

Thanks

sonallux:
Hi @shakuzen and @jonatan-ivanov, is someone already working on this? We stumpled about this missing feature of the OTLP micrometer registry and I would like to work on this.

shakuzen:
I believe @jonatan-ivanov has started work on this, but I'll let him give any update or say if he would like help with it.

jonatan-ivanov:
@sonallux Thank you for the offer! If you want to work on it, please feel free to do so but there is a catch. :) There are multiple different ways to implement this and I'm right now still exploring them. If you have thoughts on how you would implement this or want to create a draft/prototype, please don't hesitate to share, we are appreciating inputs on this.
It is not guaranteed that we will chose the direction you did but feedback is welcomed. I'm only saying this because I don't want you to be disappointed if you work a lot on a PR and we go to a different direction but it can also happen that we will apply the code you come up with or do something influenced by you. Does this sound fair?
I will be in and out in the next two weeks due to conferences so might not be very responsive in the near future but please feel free to share what you have. Other than this issue, you can also create a room with Tommy (`@Tommy Ludwig`) and me (`@Jonatan Ivanov`) on https://slack.micrometer.io if you want to discuss/share anything.

sonallux:
> It is not guaranteed that we will chose the direction you did but feedback is welcomed. I'm only saying this because I don't want you to be disappointed if you work a lot on a PR and we go to a different direction but it can also happen that we will apply the code you come up with or do something influenced by you. Does this sound fair?

@jonatan-ivanov Yes, that is okay for me. I have already created a working prototype (see PR #5486). Please let me know, what you think about this draft.

Nikzz3:
I would also love to have this feature.

Is there already an estimated time for release?

Thanks.

ryleighmikami:
Hello team, upvoting this

illenko:
Waiting for this feature.

Thanks!

ThomasVitale:
Is there anything I can do to help with this?

vmeunier:
Hello
[...]

PR #7199, opened 2026-02-14
OTLP Exemplars support

jonatan-ivanov:
## Usage

If you want to try this out, please see `OtlpExemplarsSample` or the docs about how to enable it ():
```java
OtlpConfig config = OtlpConfig.DEFAULT;
ExemplarContextProvider contextProvider = new TracingExemplarContextProvider();
OtlpMeterRegistry meterRegistry = OtlpMeterRegistry.builder(config)
    .exemplarContextProvider(contextProvider)
    .build();
```

If you are using Spring Boot, this should help:
<details>
<summary>click to reveal</summary>

```java
@Bean
OtlpMeterRegistry otlpMeterRegistry(OtlpConfig otlpConfig, Clock clock, ObjectProvider<OtlpMetricsSender> metricsSender, ExemplarContextProvider exemplarContextProvider) {
	OtlpMeterRegistry.Builder builder = OtlpMeterRegistry.builder(otlpConfig)
			.clock(clock)
			.exemplarContextProvider(exemplarContextProvider);
	metricsSender.ifAvailable(builder::metricsSender);
	return builder.build();
}

@Bean
ExemplarContextProvider exemplarContextProvider(ObjectProvider<Tracer> tracerProvider) {
	return new DemoExemplarContextProvider(tracerProvider);
}

static class DemoExemplarContextProvider implements ExemplarContextProvider {

	private final SingletonSupplier<Tracer> tracer;

	DemoExemplarContextProvider(ObjectProvider<Tracer> tracerProvider) {
		this.tracer = SingletonSupplier.of(tracerProvider::getObject);
	}

	@Override
	public @Nullable OtlpExemplarContext getExemplarContext() {
		Span span = tracer.obtain().currentSpan();
		if (isSampled(span)) {
			TraceContext context = span.context();
			return new OtlpExemplarContext(context.traceId(), context.spanId());
		}
		return null;
	}

	private boolean isSampled(@Nullable Span span) {
		if (span == null) {
			return false;
		}
		Boolean sampled = span.context().sampled();
		return sampled != null && sampled;
	}
}
```
</details> 

Closes https://github.com/micrometer-metrics/micrometer/issues/3795

### Sizing

Right now the max size of the exemplars storage by use-case:
- counter: 16
- timer/distribution summary
  - w/o histogram: 1 
  - w/ bucket histogram: number of buckets
  - w/ exponential histogram: 16

OTel uses `Runtime.getRuntime().availableProcessors()` as the size for the exemplar store in case of counters and exponential histograms and the number of buckets for bucket histograms so the two are similar but also raises questions like:
- Are 16 buckets too much?
- Shouldn't timer/distribution summary w/o histogram match to counters?

### Rate limiting

Right now the sampler calls `ExemplarContextProvider` every single time a measurement is made. Maybe we could rate-limit this and only call `ExemplarContextProvider` for the same sampler after certain amount of time passed (10ms). This is somewhat tricky since spans are also sampled and this can reduce the possibility a span being sampled for exemplars.

### Sampling improvements

Sampling strategies right now do not take the age of the exemplars into account in a way that "freshly" sampled exemplars can be overwritten by the next measurements. Having some sort of a retention strategy could be an improvement. On the other hand it can significantly complicate sampling logic.

### Set fields on sampled Spans

A Span being sampled for an exemplar is an immutable event from the Span's perspective. This can be problematic for backends that do server-side sampling since Spans that are sampled for an exemplar can be dropped by the backend. If we could modify these spans during sampling (e.g. adding an exemplar=true flag), backends can make informed decisions about what spans to drop and what to keep.

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
@@ -16,11 +16,13 @@
 package io.micrometer.registry.otlp;
 
 import com.google.protobuf.ByteString;
+import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.step.StepValue;
 import io.opentelemetry.proto.common.v1.AnyValue;
 import io.opentelemetry.proto.common.v1.KeyValue;
 import io.opentelemetry.proto.metrics.v1.Exemplar;
+import org.jspecify.annotations.Nullable;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -32,6 +34,8 @@ import java.util.function.DoubleUnaryOperator;
 import java.util.function.Supplier;
 
 class OtlpExemplarSampler implements ExemplarSampler {
+
+    private static final WarnThenDebugLogger LOGGER = new WarnThenDebugLogger(OtlpExemplarSampler.class);
 
     private final ExemplarContextProvider exemplarContextProvider;
 
@@ -140,32 +144,37 @@ class OtlpExemplarSampler implements ExemplarSampler {
 
         private static Exemplar createExemplar(double measurement, DoubleUnaryOperator converter,
                 OtlpExemplarContext exemplarContext, Clock clock) {
-            String traceId = exemplarContext.getTraceId();
-            String spanId = exemplarContext.getSpanId();
+            ByteString traceId = fromHex(exemplarContext.getTraceId());
+            ByteString spanId = fromHex(exemplarContext.getSpanId());
 
             Exemplar.Builder builder = Exemplar.newBuilder()
                 .setAsDouble(converter.applyAsDouble(measurement))
                 .setTimeUnixNano(TimeUnit.MILLISECONDS.toNanos(clock.wallTime()));
 
             if (traceId != null) {
-                builder.setTraceId(ByteString.fromHex(traceId));
-                // .addFilteredAttributes(KeyValue.newBuilder()
-                // .setKey("originalTraceId")
-                // .setValue(AnyValue.newBuilder().setStringValue(traceId))
-                // .build());
+                builder.setTraceId(traceId);
             }
             if (spanId != null) {
-                builder.setSpanId(ByteString.fromHex(spanId));
-                // .addFilteredAttributes(KeyValue.newBuilder()
-                // .setKey("originalSpanId")
-                // .setValue(AnyValue.newBuilder().setStringValue(spanId))
-                // .build());
+                builder.setSpanId(spanId);
             }
             for (io.micrometer.common.KeyValue keyValue : exemplarContext.getKeyValues()) {
                 builder.addFilteredAttributes(toOtelKeyValue(keyValue));
             }
 
             return builder.build();
+        }
+
+        private static @Nullable ByteString fromHex(@Nullable String value) {
+            if (value == null) {
+                return null;
+            }
+            try {
+                return ByteString.fromHex(value);
+            }
+            catch (Exception e) {
+                LOGGER.log("Unable to parse value!", e);
+                return null;
+            }
         }
 
         private static KeyValue toOtelKeyValue(io.micrometer.common.KeyValue micrometerKeyValue) {
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
@@ -19,7 +19,6 @@ import io.micrometer.core.instrument.*;
 import io.restassured.response.Response;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.testcontainers.containers.GenericContainer;
@@ -33,7 +32,6 @@ import java.time.Duration;
 import static io.micrometer.registry.otlp.CompressionMode.GZIP;
 import static io.micrometer.registry.otlp.CompressionMode.NONE;
 import static io.restassured.RestAssured.given;
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 import static org.hamcrest.Matchers.*;
 import static org.testcontainers.containers.BindMode.READ_ONLY;
@@ -76,18 +74,6 @@ class OTelCollectorIntegrationTest {
                     "System property 'otel-collector-image.version' is not set. This should be set in the build configuration for running from the command line. If you are running OTelCollectorIntegrationTest from an IDE, set the system property to the desired collector image version.");
         }
         return version;
-    }
-
-    @BeforeEach
-    void preCheck() {
-        assertThat(container.isRunning()).isTrue();
-        // @formatter:off
-        await()
-            .atMost(Duration.ofSeconds(10))
-            .pollDelay(Duration.ofMillis(100))
-            .pollInterval(Duration.ofMillis(100))
-            .untilAsserted(() -> whenPrometheusScraped().then().statusCode(200));
-        // @formatter:on
     }
 
     @Test
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpExemplarSamplerTests.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpExemplarSamplerTests.java
@@ -153,6 +153,40 @@ class OtlpExemplarSamplerTests {
             assertThat(sampler.collectExemplars()).isEmpty();
         }
 
+        @Test
+        void nonHexEncodedTraceIdShouldNotThrowException() {
+            assertThat(sampler.collectExemplars()).isEmpty();
+            contextProvider.setExemplar("abcxyz", "00f067aa0ba902b7", KeyValues.empty());
+            sampler.sampleMeasurement(42.0);
+
+            assertThat(sampler.collectExemplars()).isEmpty();
+            clock.add(STEP);
+
+            assertThat(sampler.collectExemplars()).singleElement().satisfies(exemplar -> {
+                assertThat(encodeHexString(exemplar.getTraceId())).isEqualTo("");
+                assertThat(encodeHexString(exemplar.getSpanId())).isEqualTo("00f067aa0ba902b7");
+                assertThat(exemplar.getAsDouble()).isEqualTo(42.0);
+                assertThat(exemplar.getFilteredAttributesList()).isEmpty();
+            });
+        }
+
+        @Test
+        void nonHexEncodedSpanIdShouldNotThrowException() {
+            assertThat(sampler.collectExemplars()).isEmpty();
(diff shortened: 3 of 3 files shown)


implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
```diff
@@ -0,0 +1,248 @@
...
+
+            if (traceId != null) {
+                builder.setTraceId(ByteString.fromHex(traceId));
+                // .addFilteredAttributes(KeyValue.newBuilder()
+                // .setKey("originalTraceId")
+                // .setValue(AnyValue.newBuilder().setStringValue(traceId))
+                // .build());
+            }
+            if (spanId != null) {
+                builder.setSpanId(ByteString.fromHex(spanId));
```
shakuzen: An exception will be thrown if `spanId` and `traceId` are not in hex format. We should either relax that requirement or at least document it in `OtlpExemplarContext`.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarContext.java
```diff
@@ -0,0 +1,67 @@
...
+ * @author Jonatan Ivanov
+ * @since 1.17.0
+ */
+public class OtlpExemplarContext {
+
+    private final @Nullable String traceId;
+
+    private final @Nullable String spanId;
+
+    private final @Nullable Iterable<KeyValue> keyValues;
```
shakuzen: It will probably be nicer to make this non-nullable and of type `KeyValues`. The constructor that doesn't take keyvalues can pass `KeyValues.empty()` to the other constructor.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
jonatan-ivanov: Do you have any suggestion for relaxing it?

According to the OTLP specs, what we pass here must be hex encoded:
>The `traceId` and `spanId` byte arrays are represented as [case-insensitive hex-encoded strings](https://tools.ietf.org/html/rfc4648#section-8);

We can encode any arbitrary string to hex but we don't really know the semantics/intent of the user input: is it a hexadecimal number or a decimal number or just a string?

Since only the implementor of the `ExemplarContextProvider` knows this and this is OTLP-specific, I think we should document it: https://github.com/micrometer-metrics/micrometer/pull/7199/commits/7512dd013e95c8fb4812fc5dc2d649982524fd6f and let the implementor do the necessary encodings (if any).

shakuzen: > According to the OTLP specs, what we pass here must be hex encoded

Could you give a link to that? I'm struggling to find it myself.

docs/modules/ROOT/pages/implementations/otlp.adoc
```diff
@@ -183,3 +183,20 @@ Alternatively, if you want to generate Summary data point for a timer with name
 ----
 management.metrics.distribution.percentiles.test.timer=0.9,0.99
 ----
+
+== Exemplars
+
+Exemplars are metadata that you can attach to the value of your time series. They can reference data outside of your metrics. A common use case is storing tracing information (`traceId`, `spanId`) but any arbitrary key-value pairs can be attached. Exemplars are not tags/dimensions (attributes in OTLP terminology); they will not increase cardinality since they belong to the values of the time series. Using them allows you to correlate metrics with traces/spans (or other concepts).
```
shakuzen: I don't think the current docs are consistent or have anything setup to enforce, but I find for things like markdown or asciidoc, having one sentence per line helps make diffs and review easier.

shakuzen: A polish suggestion. Feel free to edit differently.

```suggestion
Exemplars are metadata that you can attach to the value of your metrics time series. They can reference data outside of your metrics. A common use case is cross-referencing tracing information (`traceId`, `spanId`) but any arbitrary key-value pairs can be attached. Exemplars are not tags/dimensions (attributes in OTLP terminology); they will not increase cardinality since they belong to the values of the time series.
```

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
```diff
@@ -69,33 +78,45 @@ private static String getCollectorImageVersion() {
...
 
+    @BeforeEach
+    void preCheck() {
+        assertThat(container.isRunning()).isTrue();
+        // @formatter:off
+        await()
+            .atMost(Duration.ofSeconds(10))
+            .pollDelay(Duration.ofMillis(100))
+            .pollInterval(Duration.ofMillis(100))
+            .untilAsserted(() -> whenPrometheusScraped().then().statusCode(200));
```
shakuzen: Is this `preCheck` needed? Shouldn't that all be covered by the `waitingFor` set on the container?

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
```diff
@@ -0,0 +1,245 @@
...
+            DoubleUnaryOperator converter) {
+        this.exemplarContextProvider = exemplarContextProvider;
+        this.clock = clock;
+        this.exemplars = exemplars;
+        this.converter = converter;
+    }
+
+    @Override
+    public void sampleMeasurement(double measurement) {
+        OtlpExemplarContext exemplarContext = exemplarContextProvider.getExemplarContext();
```
shakuzen, line 68: Since the cost of `getExemplarContext()` is non-zero (when backed by a real Tracer implementation) and this will be called on each measurement, but the selector might not select the measurement, I guess we're trading off avoiding the cost of the CellSelector in case this is null. I wonder what typically wastes less. We should check some scenarios, but that doesn't need to be done before merging this PR.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
```diff
@@ -0,0 +1,245 @@
...
+            Exemplar.Builder builder = Exemplar.newBuilder()
+                .setAsDouble(converter.applyAsDouble(measurement))
+                .setTimeUnixNano(TimeUnit.MILLISECONDS.toNanos(clock.wallTime()));
+
+            if (traceId != null) {
+                builder.setTraceId(ByteString.fromHex(traceId));
+                // .addFilteredAttributes(KeyValue.newBuilder()
+                // .setKey("originalTraceId")
+                // .setValue(AnyValue.newBuilder().setStringValue(traceId))
+                // .build());
```
shakuzen: I assume we want to clean up this commented code before merging.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
jonatan-ivanov: Oops, sorry, I meant to link that part of the [OTLP specs](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding).

Oh wait, that one is for "JSON Protobuf Encoding" and there is also "Binary Protobuf Encoding". Let me check what the collector does.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
jonatan-ivanov: It's not needed in the sense that this check could go to the `@Test`s or can be omitted since we check the body so tests would still fail/pass without this.

But there are essentially two states here: 1. the container is up and produces an empty http 200 response (this check) and 2. the container responses with a non-empty 200 (body check). There can be significant amount of time between the two (up to 5-10 seconds).
So even if this is not strictly necessary, I found it helpful if there is a problem with the container:
- The `@Test`s can forget to do this (and the check on the body might fail in a non-trivial way)
- It is separated from the checks in the `@Test`s so it can fail faster
- Because of this separation, it can be also configured separately

jonatan-ivanov:
>On exemplar sizing, (up to) 16 exemplars on counters sounds like too much to me. Only (up to) 16 may be low for some exponential histograms. This is something on which it would be good to get feedback from users before the 1.17.0 release candidate.

The more I think about the exemplars sizing the more I think we should improve it:
- Reducing exemplars for counters
- Set size of timer/ds to the same value as counters (right now it's 1)
- Look into a smarter way of setting the size for exponential histograms (maybe dynamically?)

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
jonatan-ivanov: I think this is a great idea, there is a trick though: the CellSelector can be stateful and calling it in cases when there is no span or it is not sampled changes its "dynamics" (probability that it can select a span when it will be sampled in case of "RandomDecayingProbabilityCellSelector"). Though I think this can be worked around by separating the logic of getting an index and mutating the state, I'll look into this.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
jonatan-ivanov: That's a good question, I found it useful many times when I was looking into the payload since the raw traceId/spanId are not very readable so this can be helpful for people who play with the code but we can remove it.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
jonatan-ivanov: I played with this and "what we pass here must be hex encoded" above I think should have been worded as: what we pass here is really a fixed-sized byte array (so the hex-encoding is ~implicit), otherwise we get HTTP 400 Bad Request from the collector during publishing.

This does not really change the outcome though, since we need to convert a String to a byte array, we should know what the input is, is it a hex number as String, is it a decimal number as String or is it a String that (probably consists of invalid characters) and produces the bytes of the number we need if we choose the right charset during the conversion.

jonatan-ivanov: Note: we discussed this internally, and because of the issues above we will let the user do the encoding but we added error handling so a warn message will be logged if the input is not parseable (not hex-encoded).

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpExemplarSampler.java
jonatan-ivanov: Note: we discussed this internally, when this can be useful, test can add it as KeyValues to the context.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OTelCollectorIntegrationTest.java
jonatan-ivanov: Note: we discussed this internally, I misunderstood the comment, it's about the testcontainer status check vs. this. With it this seems unnecessary (though I remember having some issues at some point with the status check but don't remember exactly what), I removed the pre-check (and will look into the issues I had previously).

PR #7234, opened 2026-02-25
Add Kafka latest version integration tests

### LINKED ISSUE #7233: Integration test Kafka instrumentation against latest minor version
Related and more important than https://github.com/micrometer-metrics/micrometer/issues/7232, we should be integration testing our instrumentation against the latest version of Kafka to ensure that our instrumentation works with the latest version despite being compiled against an older version. API could be removed or changed in breaking way in new major versions, so it is not guaranteed to work.

utafrali:
Adds a test task for Kafka instrumentation against the latest minor version (3.9.0) in Docker. This runs alongside the existing stable tests to catch compatibility issues early. Closes #7233.


micrometer-core/build.gradle
```diff
@@ -303,3 +303,26 @@ tasks.register("openj9ConcurrentScavengeTest", Test) {
 
     jvmArgs '-Xgc:concurrentScavenge'
 }
+
+configurations {
+    kafkaLatestTestRuntime {
+        extendsFrom configurations.testRuntimeClasspath
+        resolutionStrategy.force "org.apache.kafka:kafka-clients:${libs.versions.kafkaLatest.get()}"
```
shakuzen, line 310: There is also the kafka-streams dependency.

utafrali:
Thanks for the review! Good catch on the kafka-streams dependency. I'll update both Kafka and kafka-streams to 4.2.0 and ensure the tests pass against these versions. Happy to wait for your discussion with @jonatan-ivanov on the overall approach and adjust accordingly if needed.

utafrali:
Closing — can't sign the DCO. Thanks\!

PR #7241, opened 2026-03-01
docs: add UDS configuration guide for StatsD (#5730)

### LINKED ISSUE #5730: Document statsd UDS config
Is there any official docs on how to set this up?

_Originally posted by @sqlboy in https://github.com/micrometer-metrics/micrometer/issues/792#issuecomment-2513125549_

kangdaeun1022:
### Description

This PR documents how to configure the StatsD registry to use Unix Domain Socket (UDS) datagrams.

It adds:
- A brief explanation of UDS support.
- A note clarifying that the `host` must be an absolute path (without a `unix://` prefix).
- Configuration examples for both Java and Spring Boot (`application.yml`).

Closes #5730

### DIFF
--- a/docs/modules/ROOT/pages/implementations/statsD.adoc
+++ b/docs/modules/ROOT/pages/implementations/statsD.adoc
@@ -99,17 +99,7 @@ StatsdConfig config = new StatsdConfig() {
 MeterRegistry registry = new StatsdMeterRegistry(config, Clock.SYSTEM);
 ----
 
-*Spring Boot configuration example (`application.yml`):*
 
-[source,yaml]
-----
-management:
-  metrics:
-    export:
-      statsd:
-        protocol: uds_datagram
-        host: /var/run/datadog/dsd.socket
-----
 
 === Using Apache Kafka for Line Sink
 


docs/modules/ROOT/pages/implementations/statsD.adoc
```diff
@@ -64,6 +64,53 @@ MeterRegistry registry = StatsdMeterRegistry.builder(StatsdConfig.DEFAULT) <2>
...
+*Spring Boot configuration example (`application.yml`):*
+
+[source,yaml]
+----
+management:
+  metrics:
+    export:
+      statsd:
+        protocol: uds_datagram
+        host: /var/run/datadog/dsd.socket
```
shakuzen: Please remove this Spring Boot configuration example. We have had these in the docs historically, but it is better going forward that we are not duplicating documentation that should be in the Spring Boot docs. It gets out of sync with the properties for the latest version of Boot and Micrometer does not depend on Spring Boot.

kangdaeun1022: Thanks for the clarification!

I've removed the Spring Boot configuration example to avoid duplicating documentation that belongs in the Spring Boot docs.

Please let me know if anything else should be adjusted.

kangdaeun1022:
Thanks for the clarification!

I've removed the Spring Boot configuration example to avoid duplicating documentation that belongs in the Spring Boot docs.

Please let me know if anything else should be adjusted.

Issue #6404, opened 2025-06-14
Remove @NullUnmarked workaround for NullAway bug on generic types


jonatan-ivanov:
It seems `<T extends @Nullable Object>` does not work well in some situations, see https://github.com/uber/NullAway/issues/1075.

Because of this bug in NullAway, we need to use `@NullUnmarked` and "invert" the contract unnecessarily, see [here](https://github.com/micrometer-metrics/micrometer/pull/6385/files#diff-ded137c872c542917642b07bf544b3c5d4319a17a2f10fe771bf158f54390c11R59) and [here](https://github.com/micrometer-metrics/micrometer/pull/6385/files#diff-ded137c872c542917642b07bf544b3c5d4319a17a2f10fe771bf158f54390c11R72-R73).

seonghyeoklee:
I'd like to work on this. The upstream NullAway bug (uber/NullAway#1075) has been fixed in v0.12.10, and Micrometer already uses v0.13.1, so the `@NullUnmarked` workarounds can be safely removed.

seonghyeoklee:
I attempted to remove the `@NullUnmarked` workarounds (PR #7242), but found that the NullAway fix in uber/NullAway#1075 does not fully cover this use case.

Specifically, removing `@NullUnmarked` from `check()` and `checkRequired()` and using `T extends @Nullable Object` causes downstream compilation errors in registry modules (e.g., `micrometer-registry-azure-monitor`). When a `@Nullable`-returning method reference like `AzureMonitorConfig::connectionString` is passed as `Function<M, T>`, NullAway fails to infer `T` as nullable and reports:

```
referenced method returns @Nullable, but functional interface method
java.util.function.Function.apply(T) returns @NonNull
```

This is a separate inference limitation in NullAway beyond what #1075 addressed. This workaround removal may need to wait for a further upstream fix.

shakuzen:
Thank you for the investigation and trying to fix it. It's appreciated, and your analysis will help check in the future if this has been fixed in NullAway or not.

seonghyeoklee:
Thank you for the kind words\! I'll keep an eye on NullAway's progress regarding this pattern and will be happy to revisit when the upstream support is ready.

PR #7242, opened 2026-03-02
Remove @NullUnmarked workaround for NullAway bug on generic types

seonghyeoklee:
The upstream NullAway bug (uber/NullAway#1075) affecting inference for calls to generic methods has been fixed in v0.12.10, and Micrometer already uses v0.13.1. Remove the `@NullUnmarked` workarounds in `MeterRegistryConfigValidator` and use proper JSpecify annotations: replace `@NullUnmarked` with `T extends @Nullable Object` on `check()` and `checkRequired()`.

Closes gh-6404

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterRegistryConfigValidator.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/config/MeterRegistryConfigValidator.java
@@ -17,8 +17,7 @@ package io.micrometer.core.instrument.config;
 
 import io.micrometer.core.instrument.config.validate.Validated;
 import io.micrometer.core.instrument.config.validate.ValidationException;
-import org.jspecify.annotations.NonNull;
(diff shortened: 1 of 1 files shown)


seonghyeoklee:
Closing this PR. While the upstream NullAway bug (uber/NullAway#1075) for generic type inference has been fixed in v0.12.10, removing `@NullUnmarked` from `check()` / `checkRequired()` causes downstream compilation failures in registry modules (e.g. `micrometer-registry-azure-monitor`).

The issue is that NullAway does not properly infer nullable type arguments when a `@Nullable`-returning method reference is passed as a `Function<M, T>` parameter — even with `T extends @Nullable Object`. This results in errors like:

```
referenced method returns @Nullable, but functional interface method
java.util.function.Function.apply(T) returns @NonNull
```

This appears to require a further NullAway improvement beyond what uber/NullAway#1075 addressed. I'll file an issue upstream.

PR #7251, opened 2026-03-05
Revert behavior to return no-op observations that scope

### LINKED ISSUE #7200: Using context-propagation with a no-op Observation corrupts the current Observation
**Describe the bug**
After upgrade to Spring Boot 4, application log is full of these warnings (abridged):

```
2026-02-15T15:23:03.414+01:00  WARN 4561 --- [           main] i.m.o.c.ObservationThreadLocalAccessor   : There is no current scope in thread local. This situation should not happen
2026-02-15T15:23:03.414+01:00  WARN 4561 --- [           main] i.m.o.c.ObservationThreadLocalAccessor   : Observation <...> to which we're restoring is not the same as the one set as this scope's parent observation <null>. Most likely a manually created Observation has a scope opened that was never closed. This may lead to thread polluting and memory leaks.
```

**Environment**
I have created a minimal Spring Boot application that demonstrates the issue:
https://github.com/podlesh/micrometer-log-spam

In particular, the following is necessary for issue to appear:
- micrometer version 1.16.x (1.15.x is OK)
- `context-propagation` must be used (version does not seem to have any effect)
- the `Hooks.enableAutomaticContextPropagation();` line must be present

**To Reproduce**
Clone the demo repository and run the app
```
mvn spring-boot:run
```

**Expected behavior**
No such warning should happen.

**Additional context**
After some digging and code analysis, I have concluded that the replacement of `null` values by "noop" ones breaks compatibility with https://github.com/micrometer-metrics/context-propagation/blob/main/context-propagation/src/main/java/io/micrometer/context/DefaultContextSnapshotFactory.java

shakuzen:
In #6700 the behavior was changed to return true no-op observations that did not scope in certain cases. That causes problems that weren't caught by existing test cases. This adds a test case that fails without the fix to revert behavior to return no-op observations that still open scopes.

Fixes #7200

### DIFF
--- a/micrometer-observation/src/test/java/io/micrometer/observation/ObservationRegistryTest.java
+++ b/micrometer-observation/src/test/java/io/micrometer/observation/ObservationRegistryTest.java
@@ -20,7 +20,6 @@ import io.micrometer.common.KeyValues;
 import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.api.Test;
 
-import static io.micrometer.observation.Observation.NOOP;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.BDDAssertions.then;
 import static org.mockito.Mockito.mock;


micrometer-observation/src/test/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessorTests.java
```diff
@@ -229,6 +227,27 @@ void assertFalse(String message) {
...
+
+        Observation outer = Observation.createNotStarted("outer", observationRegistry);
+        outer.observe(() -> {
+            assertThat(observationRegistry.getCurrentObservation()).isSameAs(outer);
+
+            Observation inner = Observation.createNotStarted("inner", ObservationRegistry.NOOP);
+            Map<String, Object> context = Map.of(ObservationThreadLocalAccessor.KEY, inner);
+
+            try (var scope = contextSnapshotFactory.captureFrom(context).setThreadLocals()) {
+                assertThat(observationRegistry.getCurrentObservation()).isSameAs(inner);
```
shakuzen, line 244: Without the changes in Observation, this assert fails. The assert doesn't make sense with the previous behavior, but the test would fail even without the assert when the context propagation Scope in this try block is closed. The assert in ObservationThreadLocalAccessor fails in that case because it would end up closing the `outer` Observation Scope, which is wrong and even if allowed would cause the next assert on line 247 to fail.

micrometer-observation/src/test/java/io/micrometer/observation/contextpropagation/ObservationThreadLocalAccessorTests.java
```diff
@@ -362,7 +381,7 @@ public Class<? extends Map> readableType() {
 
         @Override
         public void readValues(Map sourceContext, Predicate<Object> keyPredicate, Map<Object, Object> readValues) {
-
+            readValues.putAll(sourceContext);
```
shakuzen, line 384: This implementation was missing and is needed to recreate the failure reported.

shakuzen:
I've merged in the interest of getting snapshots for this published, but feel free to leave review post-merge and we can follow-up with any changes.

PR #7252, opened 2026-03-05
Stabilize CircleCI builds

shakuzen:
We are getting consistent build failures on CircleCI on pull request builds now, and limited build failures on branches. This is strange because nothing has changed in the source that should affect build behavior. It appears the builds are failing with an OOM. This adds some diagnostics, tries to curb OOMs by limiting Gradle workers, and increases resources for the concurrency-test job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,28 +4,28 @@ executors:
   jdk24-executor:
     working_directory: ~/micrometer
     environment:
-      GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError" -Dscan.uploadInBackground=false -Dorg.gradle.workers.max=2'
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -44,7 +52,20 @@ commands:
...
       - checkout
       - restore_cache:
-          key: gradle-dependencies-{{ .Branch }}-{{ checksum "build.gradle" }}-{{ checksum ".circleci/config.yml" }}
+          keys:
+            # Branch-specific cache (best match)
+            - gradle-dependencies-{{ .Branch }}-{{ checksum "build.gradle" }}-{{ checksum ".circleci/config.yml" }}
+            # Fallback: same build config, any branch (e.g. main cache for PRs)
+            - gradle-dependencies-{{ checksum "build.gradle" }}-{{ checksum ".circleci/config.yml" }}
+            # Last resort: any gradle-dependencies cache
+            - gradle-dependencies-
```
shakuzen, line 61: I suspect these cache keys could be further improved, but I don't know what would be best. For now, I just want to get the CI back to green. We can revisit more later.

PR #3731, opened 2023-03-31
Improves the ThreadLocalAccessor story of continuing scopes

marcingrzejszczak:
@chemicL

PR #7255, opened 2026-03-06
Deprecate unused `Observation.Scope`-related methods

shakuzen:
These methods were introduced in https://github.com/micrometer-metrics/micrometer/pull/3731 / https://github.com/micrometer-metrics/micrometer/pull/3613 and subsequently they were no longer called (outside of tests) with the changes in https://github.com/micrometer-metrics/micrometer/pull/3831. They are public methods, however, so there is a possibility some code outside of Micrometer calls them. Hence, I've marked them deprecated but kept the implementation in place. Assuming we don't get any feedback, we should remove the implementation in 1.18.0.

### DIFF
--- a/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java
@@ -920,6 +920,7 @@ public interface Observation extends ObservationView {
             }
 
             @Override
+            @Deprecated
             public void reset() {
             }
 
@@ -966,7 +967,10 @@ public interface Observation extends ObservationView {
          * method if you know what you are doing and your use-case demands the usage of
          * it.
          * @since 1.10.4
+         * @deprecated No longer used by Micrometer code. Use {@link #close()} instead.
+         * This method will be removed in a future release.
          */
+        @Deprecated
         void reset();
 
         /**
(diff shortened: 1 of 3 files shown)


shakuzen:
@chemicL I used GitHub's search to check and as far as I could tell, these methods are not used (anymore?) by Reactor, but just a heads up about deprecating them (also since you were a reviewer on the PR that added some of them). Let us know if there are any concerns.
/cc @marcingrzejszczak

marcingrzejszczak:
Yeah, that was one way of trying to figure out the reactor solution that we later dumped AFAIR. Can you double check that in tracing we're not using these methods?

shakuzen:
> Can you double check that in tracing we're not using these methods?

The only thing is the `TracingObservationHandler` implementing `onScopeReset`. Tracing has its own `makeCurrent` method on `Baggage` but that's separate API from this.

shakuzen:
It is already there everywhere the methods are declared. The `@Override` methods inherit the JavaDoc.

shakuzen:
Also, heads up to @ttddyy about this change. I did use GitHub search to check that datasource-micrometer should be unaffected, but let us know if that's not the case.

ttddyy:
Thanks for the nudge.
Yeah, datasource-micrometer doesn’t use those APIs, so there should be no impact.

chemicL:
@shakuzen I don't see any use of these methods in reactor's codebase.

PR #7262, opened 2026-03-08
Upgrade to ubuntu-2404:2026.02.2 in CircleCI

izeye:
This PR upgrades to `ubuntu-2404:2026.02.2` in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ executors:
     working_directory: ~/micrometer
     machine:
       image: ubuntu-2404:2026.02.2
-    resource_class: large
+    resource_class: arm.medium
 
 commands:
   gradlew-build:


izeye:
I'm not sure why, but this is failing with the following:

> Job was rejected because resource class medium, image ubuntu-2404:2026.02.2 is not a valid resource class

See https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/11716/workflows/b2117c61-d6e4-4325-86b4-c6339bd563c7/jobs/65139

.circleci/config.yml
```diff
@@ -40,7 +40,7 @@ executors:
   machine-executor:
     working_directory: ~/micrometer
     machine:
-      image: ubuntu-2404:2025.09.1
+      image: ubuntu-2404:2026.02.2
```
shakuzen, line 43: I don't know why the error is happening, but I wonder if it helps to explicitly specify `resource_class`? It seems we're relying on the default now.

izeye: @shakuzen Thanks for the feedback!

Actually, I gave it a couple of tries ([`medium`](https://github.com/micrometer-metrics/micrometer/commit/325ac6b30c13838a51028687eb347a6ea35e0ea3) and [`medium+`](https://github.com/micrometer-metrics/micrometer/commit/7576306687b871da3ef2354d6704cec18efa1d31)), but they didn't work.

izeye: I also tried `large` and `arm.medium`, but they didn't work, either.

It might be a CircleCI problem.

shakuzen: I've posted [this](https://discuss.circleci.com/t/job-rejected-due-to-not-a-valid-resource-class-with-machine-image-ubuntu-2404-2026-02-2/54445) to their discussion forum. We'll see what response we get there.

shakuzen:
Based on the discussion in the CircleCI discussion forum, I think we should close this and wait to upgrade until there is an (intended) new publicly available version of the machine image.

PR #7273, opened 2026-03-11
Remove lastScope field and deprecated implementations from SimpleObservation

shakuzen:
This is a follow-up to #7255. With removing the `lastScopes` field on `SimpleObservation`, we avoid CPU cycles tracking info that is never used and we save memory per `SimpleObservation` instance.

## JOL
We go from 40 bytes per instance to 32 bytes object overhead. See JOL output below. (with [CompactObjectHeaders](https://openjdk.org/jeps/519), it gets down to 24 bytes)

```
io.micrometer.observation.SimpleObservation object internals:
OFF  SZ                            TYPE DESCRIPTION                    VALUE
  0   8                                 (object header: mark)          N/A
  8   4                                 (object header: class)         N/A
 12   4             ObservationRegistry SimpleObservation.registry     N/A
 16   4                         Context SimpleObservation.context      N/A
 20   4           ObservationConvention SimpleObservation.convention   N/A
 24   4       Deque<ObservationHandler> SimpleObservation.handlers     N/A
 28   4   Collection<ObservationFilter> SimpleObservation.filters      N/A
Instance size: 32 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

<details><summary>Previous JOL output</summary>
<p>

```
io.micrometer.observation.SimpleObservation object internals:
OFF  SZ                            TYPE DESCRIPTION                    VALUE
  0   8                                 (object header: mark)          N/A
  8   4                                 (object header: class)         N/A
 12   4             ObservationRegistry SimpleObservation.registry     N/A
 16   4                         Context SimpleObservation.context      N/A
 20   4           ObservationConvention SimpleObservation.convention   N/A
 24   4       Deque<ObservationHandler> SimpleObservation.handlers     N/A
 28   4   Collection<ObservationFilter> SimpleObservation.filters      N/A
 32   4              Map<Thread, Scope> SimpleObservation.lastScope    N/A
 36   4                                 (object alignment gap)         
Instance size: 40 bytes
Space losses: 0 bytes internal + 4 bytes external = 4 bytes total
```

</p>
</details> 

## Benchmarks

We see allocation savings of 72 bytes per operation (explained by the object layout difference above and one less ConcurrentHashMap being instantiated). Note that is irrespective of the number of key-values used, so it will be a higher percent savings for Observations with less key-values.

### This PR

```
ObservationBenchmark.observation                          avgt    2     96.621           ns/op
ObservationBenchmark.observation:gc.alloc.rate.norm       avgt    2    672.000            B/op
```

### `1.17.0-M3`

```
ObservationBenchmark.observation                          avgt    2     96.910           ns/op
ObservationBenchmark.observation:gc.alloc.rate.norm       avgt    2    744.000            B/op
```

### DIFF
--- a/micrometer-observation-test/src/test/java/io/micrometer/observation/tck/ObservationValidatorTests.java
+++ b/micrometer-observation-test/src/test/java/io/micrometer/observation/tck/ObservationValidatorTests.java
@@ -20,6 +20,7 @@ import io.micrometer.observation.Observation;
 import io.micrometer.observation.Observation.Event;
 import io.micrometer.observation.Observation.Scope;
 import io.micrometer.observation.ObservationRegistry;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
 import java.util.concurrent.CountDownLatch;
@@ -37,6 +38,13 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 class ObservationValidatorTests {
 
     private final ObservationRegistry registry = TestObservationRegistry.create();
+
+    @AfterEach
+    void tearDown() {
+        // Clean up the thread-local current observation scope after each test to prevent
+        // thread pollution and ensure test isolation.
+        registry.setCurrentObservationScope(null);
+    }
 
     @Test
     void doubleStartShouldBeInvalid() {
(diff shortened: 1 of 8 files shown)


micrometer-observation-test/src/test/java/io/micrometer/observation/tck/ObservationValidatorTests.java
```diff
@@ -168,15 +176,6 @@ void scopeOpenAfterStopShouldBeValid() {
         scope.close();
     }
 
-    @Test
-    @SuppressWarnings("resource")
-    void scopeResetAfterStopShouldBeValid() {
-        Observation observation = Observation.start("test", registry);
-        Scope scope = observation.openScope();
-        observation.stop();
-        scope.reset();
```
shakuzen, line 177: `scope.close()` is the recommended replacement and that is tested in the test below this removed one.

jonatan-ivanov (approved):

Issue #5731, opened 2024-12-04
Does Micrometer support GRPC Opentelemetry metrics?


alexFilichkin:
Hi,

I'm trying to configure Micrometer for metrics to send custom metrics to OTEL collector in GRPC format. Everything works with HTTP protocol, but doesn't work with GRPC.
I even cannot see it documentation https://docs.micrometer.io/micrometer/reference/implementations/otlp.html

jonatan-ivanov:
OTel supports 3 transports:
- HTTP/json
- HTTP/protobuf
- gRPC(/protobuf)

According to the specification, clients need to support one of these but backends/collectors must support all. So from the backend perspective, it should not matter which one you use. Micrometer supports HTTP/protobuf.

gRPC is not the "format" but the "protocol". gRPC uses protobuf to encode/serialize the data and it uses HTTP/2 to transfer the data. So HTTP/protobuf (what Micrometer uses) is very similar to gRPC (same encoding over HTTP).

With this in mind, could you please tell us why you need gRPC and why HTTP/protobuf does not work for you?

It also worth to mention that if you already have a gRPC client and Micrometer would bring you another one, that could case issues (as it did in the OTel SDK iirc).

alexFilichkin:
Hi @jonatan-ivanov thank you for the comment. 
Our OTEL collector is configured only for grpc and with this setup we cannot use micrometer, and have to use OTEL SDK 😞

jonatan-ivanov:
Can you add http to the collector config?
Something like this:
```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318
```

What do you use the collector for? Are you pushing metrics to a backend that Micrometer supports?

alexFilichkin:
I see your point. It's clear that we can expose the HTTP protocol in the OTEL collector or send the metric directly to the backend, bypassing the OTEl collector.
But my initial request was about the Micrometer. Why the most popular Observability library doesn't support GRPC?
gRPC offers better performance(compared to HTTP with Protobuf), especially for high-throughput systems, thanks to persistent connections and multiplexing.

shakuzen:
> especially for high-throughput systems

[...]

Issue #5040, opened 2024-05-05
Add gRPC support for OTLP in addition to HTTP


gayanper:
**Please describe the feature request.**
Add support for grpc transport protocol for OTLP implementation for metrics.

**Rationale**
Some of the tools such as Microsoft .NET Aspire Dashboard which can be used as a developer testing tool to run locally for verification of metrics and traces can only be used if all three traces, metrics and logs are transported using grpc protocol. So it would be nice to implement the support for transporting metrics using grpc in addition to http as well.

**Additional context**
https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard/overview

gayanper:
If this is something that it acceptable to be added, I would very much like to contribute with the implementation.

jonatan-ivanov:
As far as I remember (according to the specs and the OTel maintainers), collector implementation must support all three transports (http/json, http/protobuf, grpc) so that they can talk to any clients. It's not the client (Micrometer) who should support all of them but the server. For example the OTel SDK only supports two out of three. I would try to open an issue to the Aspire Dashboard to support all the transport protocols.

hackery:
@jonatan-ivanov any chance you could unearth a reference for this "must support all three transports"? Would be useful to give weight to requests against other projects to round out their implementations.

jonatan-ivanov:
I think so, but now that I just did, I think the current situation might be slightly different but it should not affect this issue.

First see the [OTel SDK specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol):
>SDKs SHOULD support both `grpc` and `http/protobuf` transports and MUST support at least one of them. If they support only one, it SHOULD be `http/protobuf`. They also MAY support `http/json`.

I also [asked OTel maintainers](https://cloud-native.slack.com/archives/C014L2KCTE3/p1677784529226719?thread_ts=1677780415.787389&cid=C014L2KCTE3):
>Jonatan Ivanov Mar 2nd, 2023 at 11:15 AM
>Thank you!
>This means that every collector implementation should support both HTTP/JSON and HTTP/protobuf since there can be clients that are only supporting one of them I guess.
>
>Jack Berg Mar 2nd, 2023 at 11:53 AM
>Correct.

I did not find if this is specified for the server but based on the SDK specs servers must support both `grpc` and `http/protobuf` since SDKs only need to support one of those (it seems servers can get away by not supporting `http/json`).
[...]

PR #7275, opened 2026-03-11
feat(otlp): Add OtlpGrpcMetricsSender and protocol detection via OTel environment variables

mmorel-35:
## Description

This PR adds OTLP/gRPC transport support to the `micrometer-registry-otlp` module by introducing a new `OtlpGrpcMetricsSender` and a companion `OtlpTransportProtocol` enum. It also extends `OtlpConfig` to detect the configured transport protocol from standard OpenTelemetry environment variables, keeping Micrometer aligned with the OTel specification for exporter configuration.

### Motivation

The OTLP registry previously only supported `http/protobuf` as its transport. When running alongside an OpenTelemetry SDK or Collector, the protocol is already expressed via well-known environment variables (`OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL`). This PR makes Micrometer respect those variables so that users don't need to duplicate configuration.

#### New: `OtlpGrpcMetricsSender`
A new implementation of `OtlpMetricsSender` that exports metrics over gRPC using the `MetricsServiceGrpc` blocking stub. The caller supplies a `ManagedChannel`, giving full control over TLS, keepalive settings, and the transport implementation.

Key behaviors:
- Forwards all configured request headers as gRPC ASCII metadata (lowercase-normalized), and rejects binary headers (keys ending in `-bin`) with a clear `IllegalArgumentException`.
- Supports GZIP compression via the gRPC `withCompression("gzip")` mechanism when `CompressionMode.GZIP` is configured.
- Parses the serialized Protobuf payload back into an `ExportMetricsServiceRequest` and delegates to the stub's `export` call.

Default OTLP/gRPC endpoint: `localhost:4317`.

#### New: `OtlpTransportProtocol`
A new enum with two values: `GRPC` and `HTTP_PROTOBUF`. A `fromString(String)` factory method parses the standard OTel string values (`"grpc"`, `"http/protobuf"`) as well as the underscore alias `"http_protobuf"` accepted in property files.

#### Updated: `OtlpConfig#protocol()`
A new default method on `OtlpConfig` that resolves the transport protocol using the following priority order:
1. Explicit `otlp.protocol` property (via `get(String)`).
2. `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` environment variable (signal-specific).
3. `OTEL_EXPORTER_OTLP_PROTOCOL` environment variable (global fallback).
4. Defaults to `HTTP_PROTOBUF`.

> **Note:** `protocol()` influences the default URL returned by `url()` for gRPC (defaulting to `localhost:4317`). It does **not** automatically switch the sender implementation — users must explicitly configure `OtlpGrpcMetricsSender` on the registry builder to use OTLP/gRPC.

Invalid protocol values are surfaced as a `ValidationException` and are caught by `OtlpConfig#validate()`.

#### Build: gRPC dependencies
`grpc-protobuf` and `grpc-stub` are added to `gradle/libs.versions.toml`. They are declared as `compileOnly` in the module build (so they don't become mandatory transitive dependencies) and as `testImplementation` alongside `grpc-inprocess` for the in-process server used in tests.

### Testing

- **`OtlpGrpcMetricsSenderTests`** — a new test class using an in-process gRPC server (`grpc-inprocess`) with a `HeaderCapturingInterceptor` to validate:
  - Metrics are exported correctly.
  - Headers are transmitted and normalised.
  - GZIP compression is applied.
  - gRPC errors (`UNAVAILABLE`) propagate as `StatusRuntimeException`.
  - Binary-suffix header keys are rejected.

- **`OtlpConfigTest`** — new test cases covering:
  - Default protocol is `HTTP_PROTOBUF`.
  - Explicit property takes precedence over environment variables.
  - `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` takes precedence over `OTEL_EXPORTER_OTLP_PROTOCOL`.
  - Fallback to the generic env var when the signal-specific one is absent.
  - Invalid protocol values are captured by validation.

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpGrpcMetricsSender.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpGrpcMetricsSender.java
@@ -24,8 +24,7 @@ import io.opentelemetry.proto.collector.metrics.v1.MetricsServiceGrpc;
 import java.util.Locale;
 
 /**
- * An implementation of {@link OtlpMetricsSender} that uses gRPC (OTLP/gRPC). Preferred
- * over HTTP for in-cluster communication (lower overhead, HTTP/2 multiplexing).
+ * An implementation of {@link OtlpMetricsSender} that uses gRPC (OTLP/gRPC).
  *
  * <p>
  * The {@link ManagedChannel} is provided by the caller, allowing full control over TLS,


jonatan-ivanov:
Taking a step back, could you please check these issues?
- https://github.com/micrometer-metrics/micrometer/issues/5040
- https://github.com/micrometer-metrics/micrometer/issues/5731

They discuss some of the points you mentioned above, especially:
>keeping Micrometer aligned with the OTel specification for exporter configuration

What specification are you referring to? Collector implementations must support all three transports (http/json, http/protobuf, grpc) so that they can talk to any clients. It's not the client (Micrometer) who should support all three of them but the server. For example the OTel SDK only supports two out of three.

>In practice, gRPC (OTLP/gRPC) is often preferred for in-cluster telemetry pipelines due to lower overhead and HTTP/2 multiplexing.

Since gRPC uses HTTP/2, I'm not sure why would it have lower overhead compared to http/protobuf in our use-case. I think it worth mentioning that HTTP/1.1 uses persistent connections just like gRPC/HTTP2 (by default). Although HTTP/2 allows multiple concurrent requests to be multiplexed over a single connection, this might not be happening in case of Micrometer since we don't send requests concurrently to an OTLP backend.

Please see this https://github.com/micrometer-metrics/micrometer/issues/5731#issuecomment-2521627559 especially this quote from one of the OTel Java maintainers:
>I have struggled to produce or find any compelling evidence suggesting an advantage of `grpc` over `http/protobuf`. On the contrary, as an employee of a company that accepts OTLP natively, we have routinely run into issues with `grpc`. Mostly proxys and load balancers not being `grpc` aware and returning error scenario responses which clients don’t correctly interpret as retriable.

Based on these, is this something that you want to use?
If so, what is the reason behind it, and why `http/protobuf` does not work for you?

mmorel-35:
Thank you for the feedback and the linked issues.

To answer your questions directly:

Why http/protobuf does not work for my use case ?

In our cluster, all OpenTelemetry instrumentation uses gRPC as the standard transport. The OTel Operator injects a Java agent into each instrumented pod via annotations, and this agent reads OTEL_EXPORTER_OTLP_ENDPOINT and OTEL_EXPORTER_OTLP_PROTOCOL=grpc from the pod environment to export to the in-cluster collector (ports 4317/4318 standard).

Micrometer runs in the same JVM as the injected OTel Java agent, so it shares the exact same environment variables. The issue is not about performance differences between gRPC and http/protobuf. I fully agree those are negligible. The issue is operational consistency and zero-touch configuration.

If Micrometer does not support reading OTEL_EXPORTER_OTLP_PROTOCOL=grpc from the environment, then every Java application using Micrometer alongside the OTel Java agent requires an explicit per-application override to force http/protobuf with a separate endpoint. This breaks the uniformity of the cluster setup where gRPC is the enforced standard for all in-cluster telemetry.

In other words: it is not that http/protobuf is technically impossible, it is that making it work requires patching the default behavior for every single application, while every other OTel component in the cluster works out of the box with the existing environment variables.

jonatan-ivanov:
The default and recommended transport by OpenTelemetry is also `http/protobuf`.

I'm a bit puzzled since if I remember correctly, in some cases, OTel disables Micrometer and tracing also has issues since the agent does not let you inject the Tracer.
So you are using the OTel Java agent and Micrometer and they both publish metrics? We usually don't recommend doing it and recommend using one of them instead.

>If Micrometer does not support reading OTEL_EXPORTER_OTLP_PROTOCOL=grpc from the environment, then every Java application using Micrometer alongside the OTel Java agent requires an explicit per-application override to force http/protobuf with a separate endpoint. This breaks the uniformity of the cluster setup where gRPC is the enforced standard for all in-cluster telemetry.

I assume you don't have a "common library" for every application where you could include such a change (mostly what you did in this PR)? Also, why does the agent and Micrometer have to use the same transport, you don;t need to reconfigure the agent to use `http/protobuf` because of Micrometer and Micrometer will not mind the agent using any protocol it support.

mmorel-35:
On using Micrometer alongside the OTel Java agent:

This is actually a well-established and documented pattern. The OTel Java agent explicitly bridges Micrometer through its micrometer instrumentation module — it registers a PrometheusMeterRegistry (or an OTLP-backed one depending on configuration) so that Micrometer metrics are exported through the agent's pipeline. See the [OpenTelemetry Java agent documentation](https://opentelemetry.io/docs/zero-code/java/agent/configuration/) and the [opentelemetry-java-instrumentation Micrometer bridge](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/micrometer). So the two coexist by design in this model Micrometer is not "competing" with the agent, it is instrumented by the agent.

On "why must they share the same transport":

 They don't have to but the point of this PR is zero-touch configuration. When the OTel Operator manages pods, it injects `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` into the pod environment. Every OTel-aware component, the Java agent, the Collector sidecar, reads this variable and configures itself automatically. If Micrometer ignores it, it becomes the only component that requires an explicit per-application override. That is the friction this PR eliminates.

On the "common library" workaround:

In a large organisation with dozens or hundreds of independent application teams, requiring every team to adopt an internal library to paper over a missing feature is not a sustainable answer. It also means that any application that does not use the internal library silently falls back to a broken configuration.

The ask is simple, Micrometer should respect the same standard OTel environment variables that every other component in the ecosystem already respects. This is not about gRPC vs http/protobuf performance it is about configuration consistency in OTel-managed environments.

jonatan-ivanov:
I feel we have a disconnect here, I'm aware of the Micrometer bridge but I'm not sure that:
- it is "well-established" (it's unstable, check `-alpha` in the version)
- "it registers a PrometheusMeterRegistry" (I don't think it does, it registers the bridge which is a registry)
- "so that Micrometer metrics are exported through the agent's pipeline" (without the "so that" part this is true, the bridge connects the Micrometer API to the OTel SDK)
- "[OpenTelemetry Java agent documentation](https://opentelemetry.io/docs/zero-code/java/agent/configuration/)" has anything Micrometer-specific
- "it is instrumented by the agent" (it is not, but it is "bridged")

Because of "Micrometer metrics are exported through the agent's pipeline", I'm not sure what you are trying to do, if the data recorded by Micrometer will be tunneled to the OTel SDK then the OTel SDK will export the metrics not Micrometer's `OtlpRegistry` so you don't need any of the changes in this PR.

At this point I need to ask: are you using some sort of AI/LLM that is telling you these things?

>In a large organisation with dozens or hundreds of independent application teams, requiring every team to adopt an internal library to paper over a missing feature is not a sustainable answer. It also means that any application that does not use the internal library silently falls back to a broken configuration.

Organizations with dozens of thousands of apps (x1000 the size you mentioned) do this.
Also, I'm not sure I understand why apps are unable to use am internal library but they can use a specific version of a non-internal one (if this would be added today, every app needs to upgrade Micrometer anyways).

>The ask is simple, Micrometer should respect the same standard OTel environment variables that every other component in the ecosystem already respects. This is not about gRPC vs http/protobuf performance it is about configuration consistency in OTel-managed environments.

I understand that the ask is simple but I think this is not about the simplicity of it. It is also not about OTel environment variables but it is about adding an extra transport that is not needed by most of the users and can be implemented using the `MetricsSender` component.

mmorel-35:
Thank you for the feedback. Given the concerns raised, I'll close this PR. The use case remains valid for my context and I'll handle it at the application level.

shakuzen:
> The use case remains valid for my context and I'll handle it at the application level.

Are you sure about that? Why would you want to export the same metrics from both Micrometer's OtlpMeterRegistry and the OTel SDK configured by the OTel Java Agent?

jonatan-ivanov:
I would definitely further investigate the question Tommy raised above: if you use the agent, data will flow through it. You need Micrometer's OtlpMeterRegistry if you don't use the agent.

If you still end-up implementing a gRPC sender, other than the things I wrote above, gRPC libraries (you use them in this PR) can conflict with each other (i.e.: your app is also using one to call other apps), see: https://github.com/open-telemetry/opentelemetry-java/issues/3123 You can look at `OkHttpGrpcSender` in OTel which uses okhttp as a gRPC client.
(But overall I would still recommend reconsidering this, it's not a coincidence that OTel also moved away from gRPC being the default, gRPC was the default in the past but users had bad experience.)

Issue #7288, opened 2026-03-13
ObservationValidator scope LIFO validation produces false positives for parallel observations


seonghyeoklee:
## Description

After merging #7253, @jonatan-ivanov identified that the LIFO scope ordering validation produces false positives when observations run in parallel on different threads.

The current implementation tracks scopes in a single global `Deque<Context>`, so sibling scopes on separate threads are incorrectly flagged as out-of-order.

## Example

```java
// Thread-1: scope1.open  →  (wait)      →  scope1.close
// Thread-2:    (wait)    →  scope2.open  →     (wait)     →  scope2.close
```

Global order: `scope1.open → scope2.open → scope1.close → scope2.close`

This is valid because scope1 and scope2 are independent siblings on different threads (e.g., two separate HTTP requests), but the validator reports an invalid closing order.

See: https://github.com/micrometer-metrics/micrometer/pull/7253#issuecomment-3775921131

## Proposed Fix

Replace the global `Deque<Context>` with a per-thread stack using `ThreadLocal<Deque<Context>>` so that each thread tracks its own scope ordering independently.

## Reproducer

<details>
<summary>Click to reveal test</summary>

```java
@Test
void siblings() throws InterruptedException {
    TestObservationRegistry registry = TestObservationRegistry.builder()
        .validateScopesClosedInReverseOrderOfOpening(true)
        .build();
    registry.observationConfig().observationHandler(new ObservationTextPublisher());

    CountDownLatch scopeOneOpenLatch = new CountDownLatch(1);
    CountDownLatch scopeTwoOpenLatch = new CountDownLatch(1);
    CountDownLatch scopeOneCloseLatch = new CountDownLatch(1);

    AtomicReference<Exception> error = new AtomicReference<>();
    Thread thread1 = new Thread(() -> {
        Observation observation1 = Observation.start("one", registry);
        try {
            Scope scope1 = observation1.openScope();
            scopeOneOpenLatch.countDown();
            scopeTwoOpenLatch.await();
            scope1.close();
            scopeOneCloseLatch.countDown();
        }
        catch (Exception e) {
            error.set(e);
        }
        finally {
            scopeOneCloseLatch.countDown();
        }
    });
    Thread thread2 = new Thread(() -> {
        try {
            scopeOneOpenLatch.await();
            Observation observation2 = Observation.start("two", registry);
            Scope scope2 = observation2.openScope();
            scopeTwoOpenLatch.countDown();
[...]

PR #7295, opened 2026-03-15
Use per-thread scope stack in ObservationValidator LIFO validation

seonghyeoklee:
## Summary

The LIFO scope ordering validation in `ObservationValidator` used a single global `Deque<Context>`, which produced false positives when observations ran in parallel on different threads (sibling scopes).

This replaces the global stack with a `ThreadLocal<Deque<Context>>` so that each thread tracks its own scope ordering independently.

## Changes

- `ObservationValidator`: `Deque<Context>` → `ThreadLocal<Deque<Context>>` with `ThreadLocal.withInitial(ArrayDeque::new)`
- `ObservationValidatorTests`: Add `siblingScopesOnDifferentThreadsShouldBeValid` test that verifies parallel observations on separate threads no longer trigger false positives

Closes gh-7288

### DIFF
--- a/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
+++ b/micrometer-observation-test/src/main/java/io/micrometer/observation/tck/ObservationValidator.java
@@ -50,7 +50,7 @@ class ObservationValidator implements ObservationHandler<Context> {
 
     private final Map<String, Set<String>> lowCardinalityKeysByObservationName;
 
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
Thank you!

jonatan-ivanov:
Fyi: I created https://github.com/micrometer-metrics/micrometer/pull/7295 to be able to clear the state of the validator.

PR #7297, opened 2026-03-16
Add FunctionCounter for MemoryMXBean.getTotalGcCpuTime

### LINKED ISSUE #7245: Add support for JDK 26's MemoryPoolMXBean.getTotalGcCpuTime()
This returns the total time spent doing GC, which would fit the GC metrics in Micrometer. See https://bugs.openjdk.org/browse/JDK-8368527 and https://jdk.java.net/26/release-notes#JDK-8368527.

shakuzen:
Closes #7245

### DIFF
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetricsTest.java
@@ -25,6 +25,7 @@ import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics.GcMetricsNotificationListener;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.condition.EnabledIf;
@@ -61,6 +62,11 @@ class JvmGcMetricsTest {
     SimpleMeterRegistry registry = new SimpleMeterRegistry();
 
     JvmGcMetrics binder = new JvmGcMetrics(DEFAULT_TAGS);
+
+    @AfterEach
+    void cleanUp() {
+        binder.close();
+    }
 
     @Test
     void noJvmImplementationSpecificApiSignatures() {


.circleci/config.yml
```diff
@@ -84,6 +84,12 @@ commands:
           path: ~/micrometer/test-results/
 
 jobs:
+  build-jdk26:
+    executor: jdk25-executor
+    steps:
+      - gradlew-build:
+          toolchainVersion: '26'
```
shakuzen, line 91: This fails until there is a Java 26 version returned by the disco API. Java 26 GA date is tomorrow.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -161,6 +164,28 @@ public void bindTo(MeterRegistry registry) {
...
+                        long nanos = (Long) m.invoke(bean);
+                        return nanos >= 0 ? TimeUtils.nanosToUnit((double) nanos, TimeUnit.SECONDS) : 0;
+                    }
+                    catch (Exception e) {
+                        return 0;
+                    }
+                })
+                    .tags(tags)
+                    .description("Approximate accumulated time spent in garbage collection")
+                    .baseUnit("seconds")
```
shakuzen: It would be more flexible to use a time-based meter instead of picking a specific time unit, but we don't have a time-based function counter. We do have a FunctionTimer, and I did consider implementing that using the number of GC events as the count, but I don't think that's semantically right for this. The [JavaDoc](https://download.java.net/java/early_access/jdk26/docs/api/java.management/java/lang/management/MemoryMXBean.html#getTotalGcCpuTime()) explicitly mentions that time may be included here that does not correspond to a collection event, so tying them together in a FunctionTimer is not correct. Alternatively we have a TimeGauge, which solves the time part of things, but we lose the semantics that the value returned is monotonically increasing (except for when it may return `-1` such as during shutdown).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -161,6 +164,28 @@ public void bindTo(MeterRegistry registry) {
 
         allocationPoolSizeAfter = new AtomicLong(0L);
 
+        MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
+        try {
+            Method m = MemoryMXBean.class.getMethod("getTotalGcCpuTime");
+            if ((Long) m.invoke(memoryMXBean) >= 0) {
+                FunctionCounter.builder("jvm.gc.cpu.time", memoryMXBean, bean -> {
```
shakuzen, line 170: I did not see a semantic convention for this in [OTel's JVM metrics](https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/). I'll check if there is an open discussion about this or figure out the process to start one.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -161,6 +164,28 @@ public void bindTo(MeterRegistry registry) {
 
         allocationPoolSizeAfter = new AtomicLong(0L);
 
+        MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
+        try {
+            Method m = MemoryMXBean.class.getMethod("getTotalGcCpuTime");
+            if ((Long) m.invoke(memoryMXBean) >= 0) {
```
jonatan-ivanov, line 169: For others/future me who are wondering if `m.invoke(memoryMXBean)` can return `null` (we catch the NPE and do nothing): it cannot, the [method signature](https://download.java.net/java/early_access/jdk26/docs/api/java.management/java/lang/management/MemoryMXBean.html#getTotalGcCpuTime()) is:
```
long getTotalGcCpuTime()
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
jonatan-ivanov, line 170: I'm thinking if it would worth it to move the lambda out to a method but I'm not sure since we need both the `bean` and the `method` too.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
```diff
@@ -161,6 +164,28 @@ public void bindTo(MeterRegistry registry) {
...
+                FunctionCounter.builder("jvm.gc.cpu.time", memoryMXBean, bean -> {
+                    try {
+                        long nanos = (Long) m.invoke(bean);
+                        return nanos >= 0 ? TimeUtils.nanosToUnit((double) nanos, TimeUnit.SECONDS) : 0;
```
jonatan-ivanov: I think converting to seconds is ok but in similar cases we use the unit we get back from the MxBean:
- `jvm.compilation.time` is `ms` (`JvmCompilationMetrics`)
- `process.cpu.time` is `ns` (`ProcessorMetrics`)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
jonatan-ivanov: This makes sense to me, in similar cases we do the same:
- `jvm.compilation.time` is a `FunctionCounter` (`JvmCompilationMetrics`)
- `process.cpu.time` is a `FunctionCounter` (`ProcessorMetrics`)

except `process.uptime` is a `TimeGauge` (`UptimeMetrics`).
Should we create an issue for `process.uptime`?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
shakuzen: I did not find an existing issue so I opened https://github.com/open-telemetry/semantic-conventions/issues/3551

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
shakuzen: Given we expect this new metric to be used with `process.cpu.time`, it would be ideal if they have the same unit, so users don't need to do extra math when writing the query. I wonder if we shouldn't explore an option for FunctionCounter with a TimeUnit.

shakuzen: See #7329 for the proposal to have a time-based FunctionCounter, which doesn't exactly help here since it would be a breaking change to use it with existing meters (units will change, depending on the registry used).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
shakuzen: I've opened https://github.com/micrometer-metrics/micrometer/issues/7346

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmGcMetrics.java
shakuzen: I've switched the units to nanoseconds to match `process.cpu.time`

PR #7301, opened 2026-03-17
Create separate Benchmark for DefaultMeterObservationHandler

shakuzen:
Focus the ObservationBenchmark on only benchmarking the Observation infrastructure. Separately benchmark with the DefaultMeterObservationHandler, an implementation of an ObservationHandler.


shakuzen:
This came up in the context of benchmarking #7273 where allocation of ObservationBenchmark was not consistent because the default implementation of LongTaskTimer that gets used when DefaultMeterObservationHandler is configured has variable allocation. It will be useful to separately benchmark the Observation infrastructure alone and with a substantial implementation of an ObservationHandler (such as DefaultMeterObservationHandler).

Issue #6908, opened 2025-11-20
Log warning when a registry is added to composite after meter registration


shakuzen:
I also think we should pursue adding a warning log so that users who can do all setup before registering metrics can avoid surprises like this and others (such as the values being recorded before a non-composite registry is added being dropped).

_Originally posted by @shakuzen in https://github.com/micrometer-metrics/micrometer/pull/6846#pullrequestreview-3480848960_

See also https://github.com/micrometer-metrics/micrometer/issues/6811#issuecomment-3451518473

shakuzen:
It could be done in a separate issue, but probably there is some documentation update that should accompany this. Perhaps the warning logged could link to the documentation (although that's potentially fragile as the link needs to always work going forward).

jonatan-ivanov:
Fixed by #7322

shakuzen:
We're still missing the documentation part of this issue. I'll open a new issue for that.

PR #7322, opened 2026-03-20
Add warning when registry is added after meters are already registered

kalakotima:
Problem:
When a MeterRegistry is added to a CompositeMeterRegistry after meters have already been registered, those existing meters are not propagated to the newly added registry. This can lead to missing metrics and unexpected behavior.

Solution:
Added a warning log in CompositeMeterRegistry#add(MeterRegistry registry) to notify users when a registry is added after meters already exist.

This helps improve visibility and prevents silent misconfiguration.

Closes https://github.com/micrometer-metrics/micrometer/issues/6908

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/composite/CompositeMeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/composite/CompositeMeterRegistry.java
@@ -15,13 +15,12 @@
  */
 package io.micrometer.core.instrument.composite;
 
+import io.micrometer.common.util.internal.logging.WarnThenDebugLogger;
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
 import io.micrometer.core.instrument.distribution.pause.PauseDetector;
 import org.jspecify.annotations.Nullable;
-import io.micrometer.common.util.internal.logging.InternalLogger;
-import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
 
 import java.util.Collections;
 import java.util.IdentityHashMap;
@@ -41,7 +40,8 @@ import java.util.function.ToLongFunction;
  */
 public class CompositeMeterRegistry extends MeterRegistry {
 
-    private static final InternalLogger logger = InternalLoggerFactory.getInstance(CompositeMeterRegistry.class);
+    private static final WarnThenDebugLogger addRegistryAfterMetersLogger = new WarnThenDebugLogger(
+            CompositeMeterRegistry.class);
 
     private final AtomicBoolean registriesLock = new AtomicBoolean();
 
@@ -140,9 +140,9 @@ public class CompositeMeterRegistry extends MeterRegistry {
 
     public CompositeMeterRegistry add(MeterRegistry registry) {
 
-        if (!this.getMeters().isEmpty()) {
-            logger.warn(
-                    "Adding a MeterRegistry after meters are already registered may cause missing metrics in the new registry.");
+        if (addRegistryAfterMetersLogger.isEnabled() && !this.getMeters().isEmpty()) {
+            addRegistryAfterMetersLogger.log(
+                    "A MeterRegistry is being added to this CompositeMeterRegistry that already has registered Meters. Meter recordings (e.g. increments, recordings) that occurred before this registry was added will not be reflected in it. If adding the registry after registering meters was not intentional, update your configuration to add all MeterRegistries to the composite before registering any Meters.");
         }
 
         lock(registriesLock, () -> {


shakuzen:
> When a MeterRegistry is added to a CompositeMeterRegistry after meters have already been registered, those existing meters are not propagated to the newly added registry.

That's not true. See the following test:

https://github.com/micrometer-metrics/micrometer/blob/87cac5c340f421aa68d2240babd8b999cef8a1f2/micrometer-core/src/test/java/io/micrometer/core/instrument/composite/CompositeMeterRegistryTest.java#L110-L127

So I'm confused about the premise of this change. Could you elaborate on the problem you noticed?

kalakotima:
@shakuzen  
Thanks for pointing this out and sharing the test case : that helps clarify the behavior.

You are right that meter instances themselves are propagated when a registry is added. However, what I observed is slightly different: metric updates that occur before a registry is attached are effectively NOOP’d and are not reflected in the newly added registry.

As demonstrated in the test, increments performed before `composite.add(...)` are not captured, which can lead to silent loss of metric data. In real-world scenarios, especially in dynamic or late-binding configurations, this can result in missing initial telemetry without any indication to the user.

My intention with this change was to improve observability of this behavior by adding a warning when a registry is added after meters already exist, so users are aware that prior metric activity was not recorded.

That said, I understand this may be considered expected behavior rather than a misconfiguration. I’m happy to adjust the approach : for example:
- Logging at debug level instead of warn
- Improving documentation instead of logging
- Or dropping the change if this aligns with the intended design

Would appreciate your thoughts on what would be the most appropriate way to surface this behavior, if at all.

jonatan-ivanov:
@kalakotima This is an expected behavior, if you add a registry to a composite, recordings through the composite will only happen in the registry after it was added to the composite.

I think this is a misconfiguration issue similar to https://github.com/micrometer-metrics/micrometer/pull/4917. Registries should be added to the composite before the first recording happens. I think we can do something similar here as https://github.com/micrometer-metrics/micrometer/pull/4917.

kalakotima:
@jonatan-ivanov 

Thanks for the shoot.

Correction, The behavior is not that meters are missing or not propagated, but rather that metric interactions that occur before a registry is attached are effectively NOOP’d and are not replayed.

From a design standpoint, this makes sense and is consistent with Micrometer’s approach of keeping instrumentation lightweight and non-buffering. However, what I observed is that this behavior may not be immediately obvious to users, especially in dynamic setups where registries are attached after some meters have already been used.

My intention with the change was to improve visibility around this behavior, not to suggest it is incorrect. I agree that this is expected behavior, so a documentation clarification (rather than logging) seems more appropriate.

Happy to adjust the PR accordingly or move this into reference documentation if that’s preferred. Thx

shakuzen:
Thanks for the clarification @kalakotima. We'll discuss this and get back to you.

shakuzen:
@kalakotima the DCO check is failing. Could you please check it? See [this section](https://github.com/micrometer-metrics/micrometer?tab=contributing-ov-file#include-a-signed-off-by-trailer) of the contributing guide.

kalakotima:
@shakuzen Thanks, done DCO check looks good now.

micrometer-core/src/main/java/io/micrometer/core/instrument/composite/CompositeMeterRegistry.java
```diff
@@ -135,6 +139,12 @@ protected Meter newMeter(Meter.Id id, Meter.Type type, Iterable<Measurement> mea
     }
 
     public CompositeMeterRegistry add(MeterRegistry registry) {
+
+        if (!this.getMeters().isEmpty()) {
+            logger.warn(
+                    "Adding a MeterRegistry after meters are already registered may cause missing metrics in the new registry.");
```
jonatan-ivanov: I'm thinking using wording something like this instead:
>A MeterRegistry is being added after a Meter has been registered to this CompositeMeterRegistry. This can indicate a configuration issue since MeterRegistries should be added to the composite registry before any Meters are registered. If this is not possible or you have a valid use case for doing this and you know what you are doing, please ignore this warning. Otherwise, please fix your configuration and add all MeterRegistries to the composite before registering any Meters.

@shakuzen What do you think?

micrometer-core/src/main/java/io/micrometer/core/instrument/composite/CompositeMeterRegistry.java
```diff
@@ -39,6 +41,8 @@
  */
 public class CompositeMeterRegistry extends MeterRegistry {
 
+    private static final InternalLogger logger = InternalLoggerFactory.getInstance(CompositeMeterRegistry.class);
```
jonatan-ivanov: Could you please use `WarnThenDebugLogger` instead?

jonatan-ivanov (changes requested):

micrometer-core/src/main/java/io/micrometer/core/instrument/composite/CompositeMeterRegistry.java
shakuzen: I think that generally sounds good. I may polish it a bit when merging.

shakuzen: I've pushed a commit with a log message that's a little different. Let me know what you think.

jonatan-ivanov (approved):

PR #7324, opened 2026-03-20
Improve exemplars sizing and add exemplarsSize config

jonatan-ivanov:
Default exemplar sizes:
- no histogram (counter, timer, distribution summary): 1
- explicit bucket histograms: number of buckets
- exponential histograms: 40 (maxBuckets/4 = 160/4)

### DIFF
--- a/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
+++ b/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
@@ -17,6 +17,7 @@ package io.micrometer.registry.otlp;
 
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.config.InvalidConfigurationException;
+import io.micrometer.core.instrument.config.validate.InvalidReason;
 import io.micrometer.core.instrument.config.validate.Validated;
 import io.micrometer.core.instrument.push.PushRegistryConfig;
 
@@ -34,6 +35,7 @@ import static io.micrometer.core.instrument.config.validate.PropertyValidator.*;
  *
  * @author Tommy Ludwig
  * @author Lenin Jaganathan
+ * @author Jonatan Ivanov
  * @since 1.9.0
  */
 public interface OtlpConfig extends PushRegistryConfig {
@@ -234,8 +236,8 @@ public interface OtlpConfig extends PushRegistryConfig {
      * {@link HistogramFlavor#EXPLICIT_BUCKET_HISTOGRAM} is used for the supported meters.
      * When this is set to {@link HistogramFlavor#BASE2_EXPONENTIAL_BUCKET_HISTOGRAM} and
      * {@code publishPercentileHistogram} is enabled
-     * {@link io.micrometer.registry.otlp.internal.Base2ExponentialHistogram} is used for
-     * recording distributions.
+     * {@link io.micrometer.registry.otlp.Base2ExponentialHistogram} is used for recording
+     * distributions.
      * <p>
      * Note: If specific SLO's are configured, this property is not honored and
      * {@link HistogramFlavor#EXPLICIT_BUCKET_HISTOGRAM} is used for those meters.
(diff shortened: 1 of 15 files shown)


shakuzen:
Could people who were interested in the OTLP exemplars feature give feedback on this, please? Mentioning people from https://github.com/micrometer-metrics/micrometer/issues/3795 @0x006EA1E @sonallux @Nikzz3 @ryleighmikami @illenko @ThomasVitale @vmeunier

vmeunier:
@shakuzen I have replied in the mentioned issue. 

Thanks for the feature, I think it works fine, I'm just having some gaps in the exemplars that I don't understand, but the problem doesn't come from micrometer, it's something probably happening between otelcol and mimir. I haven't had time to investigate and send them an issue though.

shakuzen:
@vmeunier thanks for the quick reply. We appreciate your feedback on the issue there. Sorry that it wasn't clear from my previous comment, but I was asking for feedback here specifically on the number of exemplars configured, which is what this PR is adjusting (see the PR description). Perhaps it's hard for you to give feedback on that aspect if you're having other trouble due to other issues preventing you from using the feature in your setup, though.

jonatan-ivanov:
@vmeunier Regarding having sparse exemplars in Grafana and this PR: sending more exemplars did not change anything, Grafana only shows one exemplar per time series per minute for counters and two exemplars per time series per minute for histograms (heatmap) for me even if I send 50-100 exemplars at each publish and even if I publish at every 10s (instead of every minute).

Also, if I query Prometheus inside of the lgtm stack: 
```
http://localhost:9090/api/v1/query_exemplars?query=my_counter_total
```
I see only ~5 exemplars per minute which I might be able to explain if the Prometheus export is happening ~5 times per minute since Prometheus can only have one exemplar per time series. But ~5 is still more than the single one that Grafana shows.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java
```diff
@@ -290,6 +290,17 @@ default boolean publishMaxGaugeForHistograms() {
...
 
+    /**
+     * Max number of exemplars per time series. Histograms use their own strategy for the
+     * number of exemplars and may ignore this value.
+     * @return exemplarsSize
+     *
+     * @since 1.17.0
+     */
+    default int exemplarsSize() {
+        return getInteger(this, "exemplarsSize").orElse(4);
```
shakuzen: At a minimum, we should have config validation somewhere that the size is nonnegative. I'm not sure whether that should be in the default implementation, the validate method, or both.

Similarly, we should consider how defensive we want to be in the code that will use this about receiving negative values. If we are going to explicitly handle it, we should also have tests for it. Otherwise, I guess we are relying on the config validation code to be added.

jonatan-ivanov: I added validation (to the validate method).

The places where this value is used are not public (`OtlpExemplarSamplerFactory` and `OtlpExemplarSampler`) so users cannot just pass an unvalidated config object easily. Also, these components are called through `OtlpMeterRegistry` which validates the config first so users need to override the validate method to disable validation and get trouble with this.

ThomasVitale:
Thanks for pinging me here. I'm happy to see this feature! Tomorrow I'll have time to try it out and will share feedback.

ThomasVitale:
The new default exemplar sizes seem good to me.

These two are also in alignment with the defaults used in the OpenTelemetry Metrics SDK ([docs](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#exemplar-defaults)):
- no histogram (counter, timer, distribution summary) => 1
- explicit bucket histograms => number of buckets

The default for exponential histograms in the OpenTelemetry Metrics SDK is `min(20, max_buckets)`. So, in case of maxBuckets = 160, the size would be 20 instead of 40. I don't have strong opinions either way, but I wanted to mention this difference. I wonder if it would be useful to have a way to customize the exemplar strategy used for the exponential histograms.

jonatan-ivanov:
Thank you for checking it and for the feedback!

> I wonder if it would be useful to have a way to customize the exemplar strategy used for the exponential histograms.

I think just like we have `maxBucketCount` and `maxBucketsPerMeter`, we could have `maxExponentialHistogramExemplarsCount` and `maxExponentialHistogramExemplarsCountPerMeter` but I would not complicate things if there is no user need for it.
Is this something you would use? Or do you have any alternative ideas? In both cases could you please create an issue?

Issue #7302, opened 2026-03-17
Negative counter values are invalid for the Prometheus Java Client


trip-pguedes:
**Describe the bug**
[CumulativeFunctionCounter](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/cumulative/CumulativeFunctionCounter.java) passes negative values through unchecked "count" function.
The count() method returns f.applyAsDouble(obj) — when the source (e.g., a JMX int attribute) overflows, the negative double throws IllegalArgumentException and breaks the entire prometheus scrape endpoint with a 500. 

Example of an error log:
`Forwarding to error page from request [/actuator/prometheus] due to exception [-1.[product_code]: counters cannot have a negative value]`

**Environment**
Production

- micrometer-commons : 1.15.4
- micrometer-core : 1.15.4
- micrometer-jakarta9 : 1.15.4
- micrometer-observation : 1.15.4
- micrometer-registry-prometheus : 1.15.4 
- openjdk version "21.0.10" 2026-01-20 LTS
- OpenJDK Runtime Environment Temurin-21.0.10+7 (build 21.0.10+7-LTS)
- OpenJDK 64-Bit Server VM Temurin-21.0.10+7 (build 21.0.10+7-LTS, mixed mode, sharing)

**To Reproduce**
How to reproduce the bug:
1. Create a pod running micrometer
2. Have prometheus scrape that pod
3. Increment a counter metric until it overflows

**Expected behavior**
1. Counter should never go to negative
2. Counter, if it reaches max value, should go to 0

**Additional context**
My suggestion would be either changing the counter to long or having a Math.max(0, value)

jonatan-ivanov:
This is a bit weird since iirc the Prometheus server is ok with overflow.

Here's a reproducer using only the Prometheus Java Client (no Micrometer):
```java
Counter counter = Counter.builder().name("test").register();
counter.inc(Long.MAX_VALUE);
counter.inc(Long.MAX_VALUE);
ExpositionFormats.init().findWriter(null).write(System.out, PrometheusRegistry.defaultRegistry.scrape());
```
Micrometer is ok with negative values, here is a reproducer with the Simple and the Prometheus registries:
```java
CompositeMeterRegistry registry = new CompositeMeterRegistry();
SimpleMeterRegistry smr = new SimpleMeterRegistry();
[...]

PR #7327, opened 2026-03-23
Improve FunctionCounter scrape error diagnostics

k-antonov:
## Summary

Improve diagnostics when a Prometheus `FunctionCounter` produces a negative value during scrape. (Issue https://github.com/prometheus/client_java/issues/1090)

Previously, the scrape failed with:

`-2.0: counters cannot have a negative value`

This made it hard to identify which meter was responsible.

This change preserves the existing fail-fast behavior, but enriches the exception message with the Micrometer meter name.

## What changed

- Enrich the `IllegalArgumentException` raised from the `FunctionCounter` Prometheus export path with the meter name
- Add test covering the negative `FunctionCounter` scrape scenario
- Now the exception message looks like `Failed to create Prometheus function counter snapshot for meter 'api.requests': -2.0: counters cannot have a negative value`

## Verification

Verified with a PrometheusMeterRegistryTest and small Spring Boot demo app with this Micrometer branch

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
@@ -395,8 +395,8 @@ public class PrometheusMeterRegistry extends MeterRegistry {
         }
         catch (IllegalArgumentException ex) {
             if (value < 0.0) {
-                throw new IllegalArgumentException("Failed to create Prometheus function counter snapshot for meter '"
-                        + id.getName() + "': " + ex.getMessage(), ex);
+                throw new IllegalArgumentException(
+                        "Failed to create Prometheus function counter snapshot for meter '" + id.getName() + "'", ex);
             }
             throw ex;
         }
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
@@ -380,13 +380,26 @@ public class PrometheusMeterRegistry extends MeterRegistry {
         applyToCollector(id, (collector) -> {
             List<String> tagValues = tagValues(id);
             List<String> tagKeys = tagKeys(id);
-            collector.add(id,
-                    (conventionName) -> Stream.of(new MicrometerCollector.Family<>(conventionName,
-                            family -> new CounterSnapshot(family.metadata, family.dataPointSnapshots),
-                            getMetadata(conventionName, id.getDescription()), new CounterDataPointSnapshot(fc.count(),
-                                    Labels.of(tagKeys, tagValues), null, createdTimestampMillis))));
+            collector.add(id, (conventionName) -> Stream.of(new MicrometerCollector.Family<>(conventionName,
+                    family -> new CounterSnapshot(family.metadata, family.dataPointSnapshots),
+                    getMetadata(conventionName, id.getDescription()),
+                    newFunctionCounterDataPointSnapshot(id, fc.count(), tagKeys, tagValues, createdTimestampMillis))));
         });
         return fc;
+    }
+
+    private CounterDataPointSnapshot newFunctionCounterDataPointSnapshot(Meter.Id id, double value,
+            List<String> tagKeys, List<String> tagValues, long createdTimestampMillis) {
+        try {
+            return new CounterDataPointSnapshot(value, Labels.of(tagKeys, tagValues), null, createdTimestampMillis);
+        }
+        catch (IllegalArgumentException ex) {
+            if (value < 0.0) {
+                throw new IllegalArgumentException("Failed to create Prometheus function counter snapshot for meter '"
+                        + id.getName() + "': " + ex.getMessage(), ex);
+            }
+            throw ex;
+        }
     }
 
     @Override
--- a/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
+++ b/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
@@ -450,6 +450,20 @@ class PrometheusMeterRegistryTest {
         assertThat(registry.scrape()).contains("api_requests_total 1.0");
     }
 
+    @Test
+    void functionCounterNegativeValueIncludesMeterNameInScrapeException() {
+        AtomicInteger value = new AtomicInteger(1);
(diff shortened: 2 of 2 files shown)


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -380,15 +380,28 @@ protected <T> FunctionCounter newFunctionCounter(Meter.Id id, T obj, ToDoubleFun
...
 
+    private CounterDataPointSnapshot newFunctionCounterDataPointSnapshot(Meter.Id id, double value,
+            List<String> tagKeys, List<String> tagValues, long createdTimestampMillis) {
+        try {
+            return new CounterDataPointSnapshot(value, Labels.of(tagKeys, tagValues), null, createdTimestampMillis);
+        }
+        catch (IllegalArgumentException ex) {
+            if (value < 0.0) {
+                throw new IllegalArgumentException("Failed to create Prometheus function counter snapshot for meter '"
+                        + id.getName() + "': " + ex.getMessage(), ex);
```
shakuzen: It's probably more idiomatic to not add the exception message to the message of the exception we will throw since we are passing that exception as the cause. If the stack trace is printed, it will show the message from each, with the cause's message after this exception's message. It ends up duplicating the caught exception's message in the output. What do you think?

shakuzen (commented):
Thanks for taking the time to put together a PR. In general, I think it's a nice improvement. However, from the linked issue in the Prometheus client, it seems they are amenable to changing the behavior in the client, which might be at odds or redundant with these changes. Probably that doesn't have to block doing something here but it's something to keep in mind. We should also consider whether to make this change in maintenance branches.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
k-antonov: Thank you, makes sense, updated the PR. I also got rid of strict exception message check, so the test is independent of client message.
I agree we should keep potential client changes in mind. But even if it updates behaviour in this area, in my opinion I suggest a low-risk improvement. In the worst case, the PR may become partially redundant or a candidate for simplification later. However, for now, it will already solve a real user problem.

jonatan-ivanov:
I think this issue is connected: https://github.com/micrometer-metrics/micrometer/issues/7302

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -380,15 +380,28 @@ protected <T> FunctionCounter newFunctionCounter(Meter.Id id, T obj, ToDoubleFun
...
+        }
+        catch (IllegalArgumentException ex) {
+            if (value < 0.0) {
+                throw new IllegalArgumentException(
+                        "Failed to create Prometheus function counter snapshot for meter '" + id.getName() + "'", ex);
+            }
+            throw ex;
+        }
+    }
+
```
jonatan-ivanov, line 404: I have two questions here:
- Should we instead of throwing an exception at scrape time, log a warning and return `NaN` as the value?
- Would not this be cleaner in a separate `PrometheusFunctionCounter` class?

See this as an example for both: https://github.com/jonatan-ivanov/micrometer/commit/8dec1bdde74f2e600604b6615032dc9e7313b373

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
```diff
@@ -450,6 +450,21 @@ void functionCounterNamingConvention() {
...
+        AtomicInteger value = new AtomicInteger(1);
+        FunctionCounter.builder("api.requests", value, AtomicInteger::doubleValue).register(registry);
+
+        assertThat(registry.scrape()).contains("api_requests_total 1.0");
+
+        value.set(-2);
+
+        Throwable thrown = catchThrowable(registry::scrape);
+
+        assertThat(thrown).isInstanceOf(IllegalArgumentException.class).hasMessageContaining("api.requests");
```
jonatan-ivanov, line 464: How about something like this? (Maybe with adding a bit more details for the error message?)
```java
assertThatThrownBy(registry::scrape)
    .isInstanceOf(IllegalArgumentException.class)
    .hasMessageContaining("api.requests")
    .hasCauseInstanceOf(IllegalArgumentException.class);
```

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
jonatan-ivanov: I think this might be a problem for `FunctionTimer` too, not sure if `SummaryDataPointSnapshot` validates `count` and `sum` but I feel it should similarly to `CounterDataPointSnapshot`.

jonatan-ivanov (commented):
Thank you for the PR!

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
k-antonov: Thank you for review!
1) Your https://github.com/jonatan-ivanov/micrometer/commit/8dec1bdde74f2e600604b6615032dc9e7313b373 solution is clean and it resolves the initial issue
2) Unlike `FunctionCounter`, `FunctionTimer` already clamps negative values to 0 in [CumulativeFunctionTimer](https://github.com/micrometer-metrics/micrometer/blob/04096266e526d6ff3817d245ef12477625daca20/micrometer-core/src/main/java/io/micrometer/core/instrument/cumulative/CumulativeFunctionTimer.java#L62-L79) before the `SummaryDataPointSnapshot` is created. I don't see same problem here. Should we add `WarnThenDebugLogger` log to [count and totalTime](https://github.com/micrometer-metrics/micrometer/blob/04096266e526d6ff3817d245ef12477625daca20/micrometer-core/src/main/java/io/micrometer/core/instrument/cumulative/CumulativeFunctionTimer.java#L62-L79)?

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
k-antonov: I agree, I would accept this if we still needed this test

trip-pguedes: Bumping this as we still have the problem reported in https://github.com/micrometer-metrics/micrometer/issues/7302

PR #7329, opened 2026-03-24
Add option for time-based FunctionCounter

shakuzen:
Without this change, FunctionCounters that track time do not get converted to the registry's base time unit. That's unexpected for some users. Prometheus, for example, expects time to be in seconds, but many (all?) current metrics are not in that base unit because the underlying function does not return seconds. Even if it did, another registry may expect a different time unit.

* Add a `TimeBasedBuilder` to FunctionCounter and a static `builder` overload that takes a `TimeUnit` along with the `ToDoubleFunction`
* Add package-private method `functionTimeCounter` to `MeterRegistry.More` to create a `FunctionCounter` that scales its value to the registry's base time unit
* This is analogous to how `TimeGauge` works, and provides a way to register monotonically increasing time values (like `jvm.gc.cpu.time` or `process.cpu.time`) in a way that automatically scales to the preferred time unit of the metrics backend.

See comments on https://github.com/micrometer-metrics/micrometer/pull/7297 for background motivating this change.

An alternative would be introducing a dedicated type such as `FunctionTimeCounter` which would perhaps make the feature more discoverable but obviously has more public API to support when the difference to a `FunctionCounter` is tiny.

### DIFF
--- a/docs/modules/ROOT/pages/concepts/counters.adoc
+++ b/docs/modules/ROOT/pages/concepts/counters.adoc
@@ -154,6 +154,20 @@ FunctionCounter counter = FunctionCounter
     .register(registry);
 ----
 
+If the function tracks a time value, you can use `timeUnit` on the builder to create a `FunctionCounter` that scales its value to the registry's base time unit. This is analogous to how `TimeGauge` works, and provides a way to register monotonically increasing time values (like `jvm.gc.cpu.time` or `process.cpu.time`) in a way that automatically scales to the preferred time unit of the metrics backend.
+
+[source, java]
+----
+MyCounterState state = ...;
+
+FunctionCounter counter = FunctionCounter
+    .builder("counter.time", state, state -> state.totalTime())
+    .timeUnit(TimeUnit.NANOSECONDS) // optional, scales the value to the registry's base time unit
+    .description("a description of what this counter does") // optional
+    .tags("region", "test") // optional
+    .register(registry);
+----
+
 WARNING: Attempting to construct a function-tracking counter with a primitive number or one of its `java.lang` object forms is always incorrect. These numbers are immutable. Thus, the function-tracking counter cannot ever be changed. Attempting to "re-register" the function-tracking counter with a different number does not work, as the registry maintains only one meter for each unique combination of name and tags. "Re-registering" a function-tracking counter can happen indirectly for example as the result of a `MeterFilter` modifying the name and/or the tags of two different function-tracking counters so that they will be the same after the filter is applied.
 
 Attempting to "re-register" a function-tracking counter will result in a warning like this:
--- a/docs/modules/ROOT/pages/concepts/counters.adoc
+++ b/docs/modules/ROOT/pages/concepts/counters.adoc
@@ -160,8 +160,7 @@ If the function tracks a time value, you can use `timeUnit` on the builder to cr
 ----
 MyCounterState state = ...;
 
-FunctionCounter counter = FunctionCounter
-    .builder("counter.time", state, state -> state.totalTime())
+FunctionCounter.builder("counter.time", state, state -> state.totalTime())
     .timeUnit(TimeUnit.NANOSECONDS) // optional, scales the value to the registry's base time unit
     .description("a description of what this counter does") // optional
     .tags("region", "test") // optional
--- a/docs/modules/ROOT/pages/concepts/counters.adoc
+++ b/docs/modules/ROOT/pages/concepts/counters.adoc
@@ -154,14 +154,13 @@ FunctionCounter counter = FunctionCounter
     .register(registry);
 ----
 
-If the function tracks a time value, you can use `timeUnit` on the builder to create a `FunctionCounter` that scales its value to the registry's base time unit. This is analogous to how `TimeGauge` works, and provides a way to register monotonically increasing time values (like `jvm.gc.cpu.time` or `process.cpu.time`) in a way that automatically scales to the preferred time unit of the metrics backend.
+If the function tracks a time value, you can specify the time unit of the value returned by the function when calling `FunctionCounter.builder(...)`. This creates a `FunctionCounter` that automatically scales its time-based count to the registry's base time unit. This is analogous to how `TimeGauge` works, and provides a way to register monotonically increasing time values (e.g. `RuntimeMXBean::getUptime`) in a way that automatically scales to the preferred time unit of the metrics backend.
 
 [source, java]
 ----
 MyCounterState state = ...;
 
-FunctionCounter.builder("counter.time", state, state -> state.totalTime())
-    .timeUnit(TimeUnit.NANOSECONDS) // optional, scales the value to the registry's base time unit
+FunctionCounter.builder("counter.time", state, state -> state.totalTime(), TimeUnit.NANOSECONDS)
     .description("a description of what this counter does") // optional
     .tags("region", "test") // optional
     .register(registry);
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
@@ -121,6 +121,7 @@ public interface FunctionCounter extends Meter {
          * the {@link MeterRegistry}.
          * @param timeUnit time unit of the provided function return value.
          * @return The counter builder with added time unit.
+         * @since 1.17.0
          */
         public Builder<T> timeUnit(@Nullable TimeUnit timeUnit) {
             this.functionTimeUnit = timeUnit;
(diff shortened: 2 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
```diff
@@ -112,6 +116,20 @@ public Builder<T> baseUnit(@Nullable String unit) {
...
+        /**
+         * Only set this if this function counter is time-based. This will be used to
+         * convert from the time unit provided by the function to the base time unit for
+         * the {@link MeterRegistry}. If this is set, any {@link #baseUnit(String)} will
+         * be ignored and the registry's base time unit will be used instead.
+         * @param timeUnit time unit of the provided function return value.
+         * @return The counter builder with added time unit.
+         * @since 1.17.0
+         */
+        public Builder<T> timeUnit(@Nullable TimeUnit timeUnit) {
```
shakuzen: Perhaps it's more clear to name the method `functionTimeUnit` to distinguish from the time unit that will end up on the Meter (`Id`), which is actually the registry's base time unit?

shakuzen: Another thought I had was do we want to make it possible to override the target time unit? In general, I would say surely no, but this feature as it is, only solves our problems for new functional time counters going forward. We can't convert existing FunctionCounter instances that are time based because it will result in a different eventual base unit compared to currently, depending on the registry.

docs/modules/ROOT/pages/concepts/counters.adoc
```diff
@@ -154,6 +154,20 @@ FunctionCounter counter = FunctionCounter
     .register(registry);
 ----
 
+If the function tracks a time value, you can use `timeUnit` on the builder to create a `FunctionCounter` that scales its value to the registry's base time unit. This is analogous to how `TimeGauge` works, and provides a way to register monotonically increasing time values (like `jvm.gc.cpu.time` or `process.cpu.time`) in a way that automatically scales to the preferred time unit of the metrics backend.
```
shakuzen: I'm out of time right now, but I think this documentation can be improved. Either we should link to our documentation on the mentioned metrics (if we have any) or instead reference the JMX method from which they're derived, as that is more universally understandable.

micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
jonatan-ivanov: I was thinking about this, I remember people getting confused with the `TimeUnit` parameter of `FunctionTimer`/`TimeGauge`.

docs/modules/ROOT/pages/concepts/counters.adoc
```diff
@@ -154,6 +154,20 @@ FunctionCounter counter = FunctionCounter
...
+
+[source, java]
+----
+MyCounterState state = ...;
+
+FunctionCounter counter = FunctionCounter
+    .builder("counter.time", state, state -> state.totalTime())
```
jonatan-ivanov: Is the assignment needed?
```suggestion
FunctionCounter.builder("counter.time", state, state -> state.totalTime())
```

micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
jonatan-ivanov: Where/how would that be used?

If you want to convert an existing (time-based) `FunctionCounter` to the unit of the registry:
```java
FunctionCounter.builder("counter.time", state, state -> state.totalTime())
    .timeUnit(TimeUnit.NANOSECONDS) // from .baseUnit("ns")
    .register(registry);
```
but if you don't (since its a breaking change), you don't need to do anything:
```java
FunctionCounter.builder("counter.time", state, state -> state.totalTime())
    .baseUnit("ns")
    .register(registry);
```
instead of converting from the unit of the function to the unit of the registry and (most probably) back to the unit of the function:
```java
FunctionCounter.builder("counter.time", state, state -> state.totalTime())
    .sourceTimeUnit(TimeUnit.NANOSECONDS)
    .targetTimeUnit(TimeUnit.NANOSECONDS)
    .register(registry);
```
I think it can help if users of this did a conversion in the function they passed to the builder; with this they can remove the convention and specify the right source/target units.

shakuzen: Do you mean to say it *would* be clear to name it `functionTimeUnit` or should we do more/something else to make it more clear?

shakuzen: Yeah, I guess it doesn't solve anything but doing the conversion for users to avoid them making an error in converting units themselves. Which has some value, but I think it's not worth the added API and hurdle it makes to understand what to use when. Let me know if you disagree.

shakuzen: Another idea I had was adding a variant of the `builder` method that takes a function with return type `Duration`, and then we don't need to take a `TimeUnit` or `baseUnit` at all (we can return a Builder that doesn't have those methods). I mostly like this because it removes the issue now where `timeUnit` and `baseUnit` are mutually exclusive options, which is something users need to understand about the API. That said, I don't know if it is worth adding now unless we have a clear idea where it would be used. I don't think many (any?) backing functions for typical FunctionCounters currently return a `Duration`. Converting number return types to Duration is straightforward, but it requires wrapping a method instead of using it directly, as is most common for backing functions.

jonatan-ivanov: Oh sorry, I meant to say that using `functionTimeUnit` would be clear.

jonatan-ivanov: I really like this idea. I'm also thinking if we should call the time-based property out in the method name:

```java
static <T> TimeBasedBuilder<T> timeBasedBuilder(String name, T obj, Function<T, Duration> f) { ... }

static <T> TimeBasedBuilder<T> timeBasedBuilder(String name, T obj, ToDoubleFunction<T> f, TimeUnit sourceUnit) { ... }
```

Other than providing an option with `Duration`, I think we should still provide an option with a `ToDoubleFunction` + `TimeUnit` since one tricky part with Duration is that it only accepts `long` values so if for some reason someone only have `double`, that conversion must be made (right now we can avoid this conversion by using our own `TimeUtils.convert` that works with `double`).

For the straightforward conversions to `Duration` (`long` -> `Duration`), we could provide a helper, like:
```java
static <T> Function<T, Duration> toDuration(Function<T, Long> f, TemporalUnit sourceUnit) { ... }
```
So users can do this:
```java
FunctionCounter.timeBasedBuilder("test", Demo.class, toDuration(Demo::getElapsedTimeInNanos, ChronoUnit.NANOS));
```
But now that I just wrote this down, that helper seems unnecessarry if we provide the extra option above (`ToDoubleFunction<T> f, TimeUnit sourceUnit`):
```java
FunctionCounter.timeBasedBuilder("test", Demo.class, Demo::getElapsedTimeInNanos, TimeUnit.NANOSECONDS);
```

shakuzen: As discussed offline, if we want to go this route, I think the changes need to wait for 1.18.0-M1 so we have more time to give this the thought it deserves and leave room for feedback. RC1 is too late to introduce that much new public API.

shakuzen: I've updated the PR with a TimeBasedBuilder. I opted to make the static builder method share the `builder` name for discoverability. Users are used to static builder methods being named that on built-in Meter types, so I thought a `timeBasedBuilder` would be less discoverable. I also chose to not add the Duration-based option for now. I don't think we know anywhere specifically this would be used, and it creates some friction. Both `Function<T, Duration>` and `ToDoubleFunction` are functional interfaces and so the compiler gets confused. We could introduce the duration function option with a different method name like the previously proposed `timeBasedBuilder` but that makes for an annoying inconsistency. I think we should defer until we have a good reason to put more thought into a Duration-based option.

docs/modules/ROOT/pages/concepts/counters.adoc
shakuzen: Updated to give `RuntimeMXBean::getUptime` as an example.

micrometer-core/src/main/java/io/micrometer/core/instrument/FunctionCounter.java
```diff
@@ -113,18 +145,95 @@ public Builder<T> baseUnit(@Nullable String unit) {
...
 
     }
 
+    /**
+     * Fluent builder for function counters that count monotonically increasing time.
+     *
+     * @param <T> The type of the state object from which the counter value is extracted.
+     * @since 1.18.0
+     */
+    class TimeBasedBuilder<T> {
```
shakuzen: The duplication between `Builder` and `TimeBasedBuilder` is a bit unfortunate, but I didn't have another solution that wasn't worse in some way.

jonatan-ivanov (approved):

PR #7344, opened 2026-03-26
Polish Shadow plugin configuration

Goooler:
`exclude("META-INF/stuff")` should be placed under the root `shadowJar` block.

```
OLD: old.jar
NEW: new.jar

 JAR   │ old       │ new       │ diff
───────┼───────────┼───────────┼──────
 class │  12.9 MiB │  12.9 MiB │  0 B
 other │ 375.2 KiB │ 375.2 KiB │ -8 B
───────┼───────────┼───────────┼──────
 total │  13.2 MiB │  13.2 MiB │ -8 B

 CLASSES │ old   │ new   │ diff
─────────┼───────┼───────┼───────────
 classes │  3308 │  3308 │ 0 (+0 -0)
 methods │ 37498 │ 37498 │ 0 (+0 -0)
  fields │ 15282 │ 15282 │ 0 (+0 -0)

=================
====   JAR   ====
=================

 size    │ diff │ path
─────────┼──────┼──────────────────────────────────────────────────
   960 B │ -4 B │ ∆ META-INF/MANIFEST.MF
   830 B │ -4 B │ ∆ META-INF/micrometer-registry-statsd.properties
─────────┼──────┼──────────────────────────────────────────────────
 1.7 KiB │ -8 B │ (total)
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,7 @@ buildscript {
 
 plugins {
     alias(libs.plugins.kotlin) apply false
+    alias(libs.plugins.shadow) apply false
 }
 
 // Hacks because of Antora's clone/checkout/worktrees behavior
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -235,3 +235,4 @@ plugin-bnd = "biz.aQute.bnd:biz.aQute.bnd.gradle:7.2.3"
 kotlin = { id = "org.jetbrains.kotlin.jvm", version = "2.2.21" }
 jcstress = { id = "io.github.reyerizo.gradle.jcstress", version = "0.9.0" }
 aspectj = { id = 'io.freefair.aspectj.post-compile-weaving', version = '8.14.4' }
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -35,6 +35,7 @@ buildscript {
 
 plugins {
     alias(libs.plugins.kotlin) apply false
+    alias(libs.plugins.shadow) apply false
```
shakuzen, line 38: Do we need this here, since the shadow plugin is only used in one place now (the statsd module)?

Goooler: It's recommanded to add the plugins in the root project and apply them in subprojects.

Goooler:
I'm the maintainer of Shadow, I noticed there is a misusage in the configurations, so I correct it by the way.

The real changed things are date fields:

image

shakuzen (approved):
Thanks for the explanation. It doesn't seem like this makes any real difference in the produced JARs, but it's better to configure things in the correct way.

PR #7357, opened 2026-03-28
Add logger exclusion support to Log4j2Metrics and LogbackMetrics

### LINKED ISSUE #6233: provide a way to exclude some logger from being collected by Log4j2metrics

Now the Log4j2Metrics will count all log events with all levels, that can be noisy if we don't care about some loggers, we don't want the loggers be collected into the log4j.events counters

We are hoping that we can exclude some loggers from being collected by the Log4j2Metrics.

For example:

I don't want the logger of "com.a.b.c" to be collected by the Log4j2Metrics

yhyh4420:
closes gh-6233 
This pr adds the ability to exclude specific loggers from being counted by `Log4j2Metrics` and `LogbackMetrics`. Excluded loggers are matched by exact name using a `HashSet` for O(1) lookup performance.

## Changes
- Add `Set<String> excludeLoggerNames` parameter to `Log4j2Metrics` and `LogbackMetrics` through new constructor overloads.
- Existing constructors delegate with `emptySet()` to maintain backward compatibility.

## Design Decisions
- Exact match over hierarchical matching: as suggested in [issue-comment](https://github.com/micrometer-metrics/micrometer/issues/6233#issuecomment-2877494673), exact match keeps the implementation simple and avoids performance concerns on the hot path
- `Set<String>` over `Predicate<String>`: guarantees high performance(O(1) lookup and zero overhead) when the set is empty.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
@@ -33,12 +33,16 @@ import org.apache.logging.log4j.core.filter.CompositeFilter;
 
 import java.beans.PropertyChangeListener;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import static java.util.Collections.emptyList;
+import static java.util.Collections.emptySet;
 
 /**
  * {@link MeterBinder} for Apache Log4j 2. Please use at least 2.21.0 since there was a
@@ -64,6 +68,8 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
     private final LoggerContext loggerContext;
 
+    private final Set<String> excludedLoggerNames;
+
     private final ConcurrentMap<MeterRegistry, MetricsFilter> metricsFilters = new ConcurrentHashMap<>();
 
     private final List<PropertyChangeListener> changeListeners = new CopyOnWriteArrayList<>();
@@ -77,8 +83,13 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
     }
 
     public Log4j2Metrics(Iterable<Tag> tags, LoggerContext loggerContext) {
+        this(tags, loggerContext, emptySet());
+    }
+
+    public Log4j2Metrics(Iterable<Tag> tags, LoggerContext loggerContext, Set<String> excludedLoggerNames) {
         this.tags = tags;
         this.loggerContext = loggerContext;
+        this.excludedLoggerNames = Collections.unmodifiableSet(new HashSet<>(excludedLoggerNames));
     }
 
     @Override
@@ -135,7 +146,7 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
     private MetricsFilter getOrCreateMetricsFilterAndStart(MeterRegistry registry) {
         return metricsFilters.computeIfAbsent(registry, r -> {
-            MetricsFilter metricsFilter = new MetricsFilter(r, tags);
+            MetricsFilter metricsFilter = new MetricsFilter(r, tags, excludedLoggerNames);
             metricsFilter.start();
             return metricsFilter;
         });
@@ -173,6 +184,8 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
     static class MetricsFilter extends AbstractFilter {
 
+        private final Set<String> excludedLoggerNames;
+
         private final Counter fatalCounter;
 
         private final Counter errorCounter;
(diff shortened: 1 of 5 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
```diff
@@ -185,7 +198,8 @@ static class MetricsFilter extends AbstractFilter {
 
         private final Counter traceCounter;
 
-        MetricsFilter(MeterRegistry registry, Iterable<Tag> tags) {
+        MetricsFilter(MeterRegistry registry, Iterable<Tag> tags, Set<String> excludedLoggerNames) {
+            this.excludedLoggerNames = excludedLoggerNames;
             fatalCounter = Counter.builder(METER_NAME)
```
reachvamsikrishnad-lgtm, line 203: Users should use Log4j 2 version 2.21.0 or newer due to a bug in earlier versions that caused incorrect counter increments. This information is vital for correct usage.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
reachvamsikrishnad-lgtm, line 170: The internalLogger is used for potential issues during filter management (e.g., in the close () method). Ensuring comprehensive logging around filter removal would be beneficial for troubleshooting configuration issues.

reachvamsikrishnad-lgtm (changes requested):
The internalLogger is used for potential issues during filter management (e.g., in the close () method). Ensuring comprehensive logging around filter removal would be beneficial for troubleshooting configuration issues.


Users should use Log4j 2 version 2.21.0 or newer due to a bug in earlier versions that caused incorrect counter increments. This information is vital for correct usage.

jonatan-ivanov:
@reachvamsikrishnad-lgtm Please stop spamming issues and PRs with with unrelated, AI-generated comments.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
```diff
@@ -77,8 +83,13 @@ public Log4j2Metrics(Iterable<Tag> tags) {
     }
 
     public Log4j2Metrics(Iterable<Tag> tags, LoggerContext loggerContext) {
+        this(tags, loggerContext, emptySet());
+    }
+
+    public Log4j2Metrics(Iterable<Tag> tags, LoggerContext loggerContext, Set<String> excludedLoggerNames) {
         this.tags = tags;
         this.loggerContext = loggerContext;
+        this.excludedLoggerNames = Collections.unmodifiableSet(new HashSet<>(excludedLoggerNames));
```
jonatan-ivanov, line 92: Is `new HashSet` necessary here? Is it there to ensure the `Set` is always a `HashSet` not a `TreeSet` or something else?
(Same for `LogbackMetrics`)

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
```diff
@@ -231,7 +245,9 @@ static class MetricsFilter extends AbstractFilter {
 
         @Override
         public Result filter(LogEvent event) {
-            incrementCounter(event);
+            if (!excludedLoggerNames.contains(event.getLoggerName())) {
```
jonatan-ivanov, line 248: Since this is on the hot path (every time a logger is called, this check will happen too), should not we prove it with a benchmark if JIT can eliminate this call if the `Set` is empty?
If not, I can help coming up with a solution that can eliminate the contains call if nothing is excluded.
(Same for `LogbackMetrics`)

jonatan-ivanov (commented):
Thank you for the PR!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
yhyh4420: Primary purpose is a defensive copy. unmodifiableSet alone only wraps the original, so callers could still mutate it through their reference. `new HashSet` ensures the immutability. Guaranteeing Type(HashSet) is side-benefit.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
yhyh4420: I just made a benchmark to verify whether JIT can eliminate the contains call when the Set is empty.(a37768acb7901610538c3296af68fbe87003ac99)
Benchmark results show that JIT successfully eliminates it, and I think there is no measurable overhead on that hot path. [results.txt](https://github.com/user-attachments/files/26381706/results.txt)

There are three exclusionMode options in my benchmark: 
 - none: no exclude option
 - empty: empty Set
 - small: with exclude option

**none vs empty: 0.37 ns/op (1.3%)**
**none vs small: 2 ns/op**

Excluding 2 loggers has no measurable impact on hot path. Maybe this result shows that the current HashSet-based implementation is well-optimized by JIT.

yhyh4420:
@jonatan-ivanov Hello. I just wanted to check in on the progress of this PR. Please let me know if anything else is needed. Thank you!

jonatan-ivanov:
Sorry for the delay, some other work took precedence (see https://github.com/micrometer-metrics/micrometer/issues/6233#issuecomment-4120497483), I'm going to be off next week but this is still on my todo list and planning to take a deeper look after.

yhyh4420:
@jonatan-ivanov Hello. gentle follow-up on this one — the earlier feedback(defensive-copy rationale and benchmark of hot-path) has been addressed, so it should be ready for another review whenever you have time. No rush, just keeping it on your radar. Thanks.

PR #7369, opened 2026-04-01
Introduce per-convention meter convention types and builder pattern

shakuzen:
Replace group interfaces (e.g., JvmMemoryMeterConventions) with specific per-convention interfaces (e.g., JvmMemoryUsedMeterConvention). This resolves issues where the OTel variant extending the Micrometer variant made Spring Boot auto-configuration awkward (the OTel variant is an `instanceof` the Micrometer variant), and where customizing a single meter required implementing the entire group interface.

Each binder now has a builder with per-convention setters and an `openTelemetryConventions()` convenience method. The binder merges extra tags itself so conventions stay purely semantic. Each specific interface provides a static `of()` factory for inline custom implementations.

Group interfaces and their implementations are deprecated in favor of the new per-convention types and builders.

---

The second commit in this PR https://github.com/micrometer-metrics/micrometer/pull/7369/commits/f8fc24528f4236e7d86150f5c8927af2bed4a288 removes `MeterConvention<C>` from the type hierarchy and deprecates it and SimpleMeterConvention. One problem with `MeterConvention` is its type parameter is problematic for conventions that don't derive their tags from any context object that needs to be passed. They have to set the type parameter to `Void` and users of the convention need to pass `null` to the `getTags` method, which is awkward API design that smells. Removing it from the hierarchy solves this problem because each specific convention interface defines the method signature that makes sense for it. There is no current use of a common type. There is no consumer of conventions that acts on a generic `MeterConvention` type. There is no registry of conventions. Each instrumentation needs specific conventions and uses the specific types for those conventions.

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -14,7 +14,7 @@
 ** xref:concepts/long-task-timers.adoc[Long Task Timers]
 ** xref:concepts/histogram-quantiles.adoc[Histograms and Percentiles]
 ** xref:concepts/meter-provider.adoc[Meter Provider]
-** xref:concepts/meter-convention.adoc[Meter Convention]
+** xref:concepts/meter-convention.adoc[Meter Conventions]
 ** xref:concepts/high-cardinality-tags-detector.adoc[High Cardinality Tags Detector]
 * xref:implementations.adoc[Implementations]
 ** xref:implementations/appOptics.adoc[AppOptics]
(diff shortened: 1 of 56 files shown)


shakuzen:
I've added a note to the `openTelemetryConventions()` convenience method on binder builders that explicitly mentions future minor versions may configure additional stable conventions, which is a change in behavior. Those who want strict stability should configure individual conventions rather than use the convenience method.

I'm also not sure if we should ship the convenience factor `of` methods on the conventions interfaces. I suspect the vast majority of users will use either the provided Micrometer or OTel convention. Implementing the interface without the `of` static factory is not so much to ask for custom conventions. I guess we'll keep it for now and reconsider for RC1 what we finally do.

I'll go ahead and merge to unblock these changes in the related Boot PR. Let me know if there's feedback on these changes.

PR #7372, opened 2026-04-02
Add opt-in consistent naming for ForkJoinPool metrics

### LINKED ISSUE #7089: Differences between monitoring ThreadPoolExecutor and ForkJoinPool
While I was reviewing https://github.com/micrometer-metrics/micrometer/pull/7083, I saw that if I use `ExecutorServiceMetrics.monitor` for a `ThreadPoolExecutor` (`Executors.newFixedThreadPool`) and a `ForkJoinPool` I get a few unexpected differences in the outputs.

`executor.active` and `executor.queued` have units (`threads` and `tasks`) for `ThreadPoolExecutor` while for `ForkJoinPool` they do not, this means that changing this:
```diff
- ExecutorServiceMetrics.monitor(registry, Executors.newFixedThreadPool(2), name, tags);
+ ExecutorServiceMetrics.monitor(registry, new ForkJoinPool(2), name, tags);
```
will result in different time series (`executor_active_threads` vs. `executor_active`) in the Prometheus output:
```diff
- executor_active_threads{name="test"} 0.0
+ executor_active{name="test"} 0.0
- executor_queued_tasks{name="test"} 0.0
+ executor_queued{name="test"} 0.0
```
(Fixing this is a breaking change.)

<details>
<summary>Click to reveal demo code</summary>

```java
public class Demo {
        generateAndPrintMetricsFor(Executors.newFixedThreadPool(2));
        generateAndPrintMetricsFor(new ForkJoinPool(2));
    }

    private static void generateAndPrintMetricsFor(ExecutorService executorServiceToMonitor) {
        PrometheusMeterRegistry registry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
        ExecutorService executorService = ExecutorServiceMetrics.monitor(registry, executorServiceToMonitor, "test", Tags.empty());
        List<Future<?>> futures = new ArrayList<>();
        for (int i = 0; i < 5; i++) {
            futures.add(executorService.submit(Demo::sleep));
        }
        for (Future<?> future : futures) {
            try {
                future.get(1_000, TimeUnit.MILLISECONDS);
            }
            catch (Exception e) {
                System.out.println(e.getMessage());
            }
            finally {
                future.cancel(true);
            }
        }

        System.out.println("===== Metrics for " + executorService.getClass().getSimpleName());
        System.out.println(registry.scrape());
        executorService.shutdown();
    }

    private static void sleep() {
        try {
            Thread.sleep(100);
        }
        catch (InterruptedException e) {
            // noop
        }
    }
}
```
</details>

kwondh5217:
Closes #7089

* Adds `enableConsistentForkJoinPoolBaseUnits()` static method to opt-in to consistent base units for `ForkJoinPool` metrics
* When enabled, aligns `ForkJoinPool` metrics with `ThreadPoolExecutor` counterparts by adding base units:
  * `executor.active` - `threads`
  * `executor.queued` - `tasks`
  * `executor.running` - `threads`
  * `executor.steals` - `tasks`
* Disabled by default to avoid breaking existing dashboards and alerts in monitoring systems such as Prometheus

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
@@ -426,34 +426,32 @@ public class ExecutorServiceMetrics implements MeterBinder {
     }
 
     private void monitor(MeterRegistry registry, ForkJoinPool fj) {
-        FunctionCounter.Builder<ForkJoinPool> stealsBuilder = FunctionCounter
-            .builder(metricPrefix + "executor.steals", fj, ForkJoinPool::getStealCount)
-            .tags(tags)
-            .description("Estimate of the total number of tasks stolen from "
-                    + "one thread's work queue by another. The reported value "
-                    + "underestimates the actual total number of steals when the pool " + "is not quiescent");
-        Gauge.Builder<ForkJoinPool> queuedBuilder = Gauge
-            .builder(metricPrefix + "executor.queued", fj,
-                    pool -> pool.getQueuedTaskCount() + pool.getQueuedSubmissionCount())
-            .tags(tags)
-            .description("The approximate number of tasks that are queued for execution");
-        Gauge.Builder<ForkJoinPool> activeBuilder = Gauge
-            .builder(metricPrefix + "executor.active", fj, ForkJoinPool::getActiveThreadCount)
-            .tags(tags)
-            .description("An estimate of the number of threads that are currently stealing or executing tasks");
-        Gauge.Builder<ForkJoinPool> runningBuilder = Gauge
-            .builder(metricPrefix + "executor.running", fj, ForkJoinPool::getRunningThreadCount)
-            .tags(tags)
-            .description(
-                    "An estimate of the number of worker threads that are not blocked waiting to join tasks or for other managed synchronization threads");
-        if (consistentForkJoinPoolBaseUnits) {
-            stealsBuilder.baseUnit(BaseUnits.TASKS);
-            queuedBuilder.baseUnit(BaseUnits.TASKS);
-            activeBuilder.baseUnit(BaseUnits.THREADS);
-            runningBuilder.baseUnit(BaseUnits.THREADS);
-        }
-        List<Meter> meters = asList(stealsBuilder.register(registry), queuedBuilder.register(registry),
-                activeBuilder.register(registry), runningBuilder.register(registry),
+        List<Meter> meters = asList(
+                FunctionCounter.builder(metricPrefix + "executor.steals", fj, ForkJoinPool::getStealCount)
+                    .tags(tags)
+                    .description("Estimate of the total number of tasks stolen from "
+                            + "one thread's work queue by another. The reported value "
+                            + "underestimates the actual total number of steals when the pool " + "is not quiescent")
+                    .baseUnit(consistentForkJoinPoolBaseUnits ? BaseUnits.TASKS : null)
+                    .register(registry),
+                Gauge
+                    .builder(metricPrefix + "executor.queued", fj,
+                            pool -> pool.getQueuedTaskCount() + pool.getQueuedSubmissionCount())
+                    .tags(tags)
+                    .description("The approximate number of tasks that are queued for execution")
+                    .baseUnit(consistentForkJoinPoolBaseUnits ? BaseUnits.TASKS : null)
+                    .register(registry),
+                Gauge.builder(metricPrefix + "executor.active", fj, ForkJoinPool::getActiveThreadCount)
+                    .tags(tags)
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -424,31 +426,34 @@ private void monitor(MeterRegistry registry, @Nullable ThreadPoolExecutor tp) {
...
+        Gauge.Builder<ForkJoinPool> activeBuilder = Gauge
+            .builder(metricPrefix + "executor.active", fj, ForkJoinPool::getActiveThreadCount)
+            .tags(tags)
+            .description("An estimate of the number of threads that are currently stealing or executing tasks");
+        Gauge.Builder<ForkJoinPool> runningBuilder = Gauge
+            .builder(metricPrefix + "executor.running", fj, ForkJoinPool::getRunningThreadCount)
+            .tags(tags)
+            .description(
+                    "An estimate of the number of worker threads that are not blocked waiting to join tasks or for other managed synchronization threads");
+        if (consistentForkJoinPoolBaseUnits) {
```
jonatan-ivanov: `baseUnit` is `@Nullable`, would not it be simpler to leave things as they were but add
```java
.baseUnit(consistentForkJoinPoolBaseUnits ? BaseUnits.TASKS : null)
```
and
```java
.baseUnit(consistentForkJoinPoolBaseUnits ? BaseUnits.THREADS : null)
```
inline?

so for example:
```diff
FunctionCounter.builder(metricPrefix + "executor.steals", fj, ForkJoinPool::getStealCount)
    .tags(tags)
    .description("Estimate of the total number of tasks stolen from "
            + "one thread's work queue by another. The reported value "
            + "underestimates the actual total number of steals when the pool " + "is not quiescent")
+   .baseUnit(consistentForkJoinPoolBaseUnits ? BaseUnits.TASKS : null)
    .register(registry),
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -76,6 +76,8 @@ public class ExecutorServiceMetrics implements MeterBinder {
 
     private static boolean allowIllegalReflectiveAccess = true;
 
+    private static boolean consistentForkJoinPoolBaseUnits = false;
```
jonatan-ivanov, line 79: I'm not a big fan of doing this statically but let me thing about this.
@shakuzen WDYT?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
```diff
@@ -545,4 +550,36 @@ public static void disableIllegalReflectiveAccess() {
...
+     * <p>
+     * <strong>Note:</strong> This is an opt-in breaking change. Enabling this will rename
+     * the corresponding time series in Prometheus (e.g. {@code executor_active} becomes
+     * {@code executor_active_threads}), which may affect existing dashboards and alerts.
+     */
+    public static void enableConsistentForkJoinPoolBaseUnits() {
+        consistentForkJoinPoolBaseUnits = true;
+    }
+
+    static void resetConsistentForkJoinPoolBaseUnits() {
```
jonatan-ivanov: I think it should say disable:
```suggestion
    static void disableConsistentForkJoinPoolBaseUnits() {
```

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
kwondh5217: Fixed as suggested. Thanks!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
kwondh5217: Fixed as suggested. Thanks!

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/ExecutorServiceMetrics.java
shakuzen: I think a non-static field is a better option. Users may not want to apply this change to all executor metrics. A non-static option allows to control this at a more fine-grain level. The complication, of course, is that means it needs to be added as a parameter to places. We have a lot of static methods already. So I'm not sure what, if anything, we want to try to do to address that. We could introduce a builder, but that won't help with all the existing API for creating/binding these metrics.

PR #7374, opened 2026-04-02
fix: capture actual response status in TimedHandler.onComplete()

### LINKED ISSUE #7276: Jetty 12's TimedHandler marks some requests with outcome UNKNOWN
**Describe the bug**
After updating to Jetty12 a lot of requests that were previously marked as Outcome SUCCESS are now Outcome UNKNOWN. One example of this happening is when we are showing index.jsp files through the "welcome" system of a  `WebAppContext`.

The `TimedHandler` stores the request status in `onResponseBegin`. At that moment, the status is still 0, so is marked as UNKNOWN. In `onComplete` we do have the correct status available, so should the capturing of the status move to onComplete instead?

**Workaround**
Override the status stored in `onResponseBegin` in the `onComplete`.

```
TimedHandler timedHandler = new TimedHandler(globalRegistry, Tags.empty()) {
	// The Response status code is originally stored onResponseBegin, which is too early 
	// This override copies it to onComplete, which is the last step of the request processing, ensuring that the correct status code is recorded in the metrics.
	@Override
	protected void onComplete(Request request, int status, HttpFields headers, Throwable failure) {
		request.setAttribute(RESPONSE_STATUS_ATTRIBUTE, status);
		super.onComplete(request, status, headers, failure);
	}
};
```

**Environment**
 - Micrometer version: 1.16.3
 - OS: Windows and Linux
 - Java version: openjdk 21.0.9 2025-10-21 LTS

**Expected behavior**
The request outcome is properly marked as SUCCESS instead of UNKNOWN.

blaspat:
## Summary

Jetty 12 fires `onResponseBegin` before the handler sets a response status, so the status captured in `onResponseBegin` is 0 (unset). This causes all responses to be tagged as `Outcome.UNKNOWN` instead of the correct outcome.

The fix stores the `Response` as a request attribute in `handle()` and retrieves it in `onComplete()` to update `RESPONSE_STATUS_ATTRIBUTE` with the actual status. This ensures correct status and outcome tags are recorded.

## Changes

- `TimedHandler.handle()`: stores the `Response` object as a request attribute
- `TimedHandler.onComplete()`: retrieves the `Response` and updates `RESPONSE_STATUS_ATTRIBUTE` with the actual status
- Added 4 new tests covering status codes 200, 404, 500, and the specific bug scenario

## Testing

All new tests pass:
- `statusIsCorrectlyRecordedForSuccessfulRequest` — verifies 200 → SUCCESS
- `statusIsCorrectlyRecordedForNotFoundRequest` — verifies 404 → CLIENT_ERROR
- `statusIsCorrectlyRecordedForServerError` — verifies 500 → SERVER_ERROR
- `statusIsCorrectlyRecordedWhenSetAfterOnResponseBegin` — the bug scenario from issue #7276

Fixes #7276

---

_DCO sign-off: Blasius Patrick_

### DIFF
--- a/micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
+++ b/micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
@@ -99,6 +99,16 @@ public class TimedHandler extends EventsHandler implements Graceful {
 
     @Override
     protected void onAfterHandling(Request request, boolean handled, Throwable failure) {
+        // Same issue as onComplete: onResponseBegin fires before the handler sets
+        // the status, so the attribute may still be 0. Update it here too so
+        // stopHandlerTiming sees the correct status for the handler timer.
+        Response response = (Response) request.getAttribute(RESPONSE_ATTRIBUTE);
+        if (response != null) {
+            int status = response.getStatus();
+            if (status > 0) {
+                request.setAttribute(RESPONSE_STATUS_ATTRIBUTE, status);
+            }
+        }
         stopHandlerTiming(request);
         super.onAfterHandling(request, handled, failure);
     }


micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
```diff
@@ -43,7 +43,9 @@ public class TimedHandler extends EventsHandler implements Graceful {
 
     private static final String SAMPLE_HANDLER_LONG_TASK_TIMER_ATTRIBUTE = "__micrometer_handler_ltt_sample";
 
-    protected static final String RESPONSE_STATUS_ATTRIBUTE = "__micrometer_jetty_core_response_status";
+    public static final String RESPONSE_STATUS_ATTRIBUTE = "__micrometer_jetty_core_response_status";
```
jonatan-ivanov: Why does this need to be `public`?

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
```diff
@@ -43,7 +43,9 @@ public class TimedHandler extends EventsHandler implements Graceful {
...
+
+    private static final String RESPONSE_ATTRIBUTE = "__micrometer_jetty_response";
```
jonatan-ivanov: Should not we make this aligned with `__micrometer_jetty_core_response_status`?
So it would be:
```suggestion
    private static final String RESPONSE_ATTRIBUTE = "__micrometer_jetty_core_response";
```

micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
```diff
@@ -250,6 +250,150 @@ else if ("/b".equals(path)) {
...
+        // With our fix, onComplete updates the attribute to the real status.
+        CountDownLatch latch = new CountDownLatch(1);
+        timedHandler.setHandler(new Handler.Abstract() {
+            @Override
+            public boolean handle(Request request, Response response, Callback callback) {
+                // Status is set inside handle() AFTER Jetty has already fired
+                // onResponseBegin with status 0. This is the exact scenario
+                // described in issue #7276.
+                response.setStatus(200);
+                response.write(true, BufferUtil.EMPTY_BUFFER, callback);
```
shakuzen, line 430: Doesn't `response.write` need to be called *before* `setStatus` is called to reproduce the bug? The test as is passes without the changes in main code.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
```diff
@@ -110,6 +112,16 @@ protected void onResponseBegin(Request request, int status, HttpFields headers)
...
     @Override
     protected void onComplete(Request request, Throwable failure) {
+        // Jetty 12 fires onResponseBegin before the handler sets a response status,
+        // so the status captured in onResponseBegin is 0 (unset). We need to capture
+        // the actual status here in onComplete so that tags reflect the real status.
+        Response response = (Response) request.getAttribute(RESPONSE_ATTRIBUTE);
+        if (response != null) {
+            int status = response.getStatus();
+            if (status > 0) {
+                request.setAttribute(RESPONSE_STATUS_ATTRIBUTE, status);
```
shakuzen, line 132: Does the handler Timer (stopped in `onAfterHandling`) have the same issue? Do we need to do this there as well?

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
blaspat: You're right, protected is not needed here, my bad. Reverting to protected now

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
blaspat: Added the same fix there so the `jetty.server.handling` timer gets the correct status tag too

micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
blaspat: Added a new test statusIsCorrectlyRecordedWhenWriteBeforeSetStatus that correctly calls write() first. Should I remove the old one?

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/7378

Issue #7373, opened 2026-04-02
Wrong Nullability Information in OkHttpMetricsEventListener


jckoenen:
**Describe the bug**

#### Background
In the `OkHttpMetricsEventListener`, users have the option to configure their own context specific tags via the `tag` function on the builder, taking an argument of `BiFunction<Request, Response, Tag>` ([here](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java#L300)). Both `Request` and `Response` are sourced from local state, where they are annotated as `@Nullable`([here](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java#L243)). The corresponding `package-info.java` declares `@NullMarked` ([here](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/package-info.java#L20)).

#### Problem

Due to the package level annotation, the callback effectively becomes `@NotNull BiFunction<@NotNull Request, @NotNull Response, @NotNull Tag>`. This causes the kotlin compiler to allow calls without `?`, and inserting the intrinsic checks. However, since both `Request` and `Response` are actually nullable, this can blow up at runtime.

```kotlin
OkHttpMetricsEventListener.builder(registry, "client.requests")
   .tag { request, _ -> Tag.of("key", "value") }
```

causes failures like this:

```
java.lang.NullPointerException: Parameter specified as non-null is null: method com.company.OkhttpKt.withMetrics$lambda$0$1, parameter <unused var>
    at com.company.OkhttpKt.withMetrics$lambda$0$1(Okhttp.kt)
    at io.micrometer.core.instrument.binder.okhttp3.OkHttpMetricsEventListener.lambda$time$0(OkHttpMetricsEventListener.java:170)
    at java.util.stream.ReferencePipeline$3$1.accept
    at java.util.ArrayList$ArrayListSpliterator.forEachRemaining
    at java.util.stream.AbstractPipeline.copyInto
    at java.util.stream.AbstractPipeline.wrapAndCopyInto
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential
    at java.util.stream.AbstractPipeline.evaluate
    at java.util.stream.ReferencePipeline.collect
[...]

PR #7375, opened 2026-04-03
Harmonize @Nullable annotations in okhttp instrumentation

jonatan-ivanov:
This is a just firefighting and not a proper fix.
The proper fix would be non-null request and nullable response but fixing that seems to be a very deep rabbit hole and also involves breaking changes.

Closes gh-7373

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/DefaultOkHttpObservationConvention.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/DefaultOkHttpObservationConvention.java
@@ -76,9 +76,10 @@ public class DefaultOkHttpObservationConvention implements OkHttpObservationConv
         Request request = state != null ? state.request : null;
         Response response = state != null ? state.response : null;
         IOException exception = state != null ? state.exception : null;
-        Function<Request, String> urlMapper = context.getUrlMapper();
+        Function<@Nullable Request, String> urlMapper = context.getUrlMapper();
         Iterable<KeyValue> extraTags = context.getExtraTags();
-        Iterable<BiFunction<Request, Response, KeyValue>> contextSpecificTags = context.getContextSpecificTags();
+        Iterable<BiFunction<@Nullable Request, @Nullable Response, KeyValue>> contextSpecificTags = context
+            .getContextSpecificTags();
         Iterable<KeyValue> unknownRequestTags = context.getUnknownRequestTags();
         boolean includeHostTag = context.isIncludeHostTag();
         // TODO: Tags to key values and back - maybe we can improve this?
@@ -98,7 +99,7 @@ public class DefaultOkHttpObservationConvention implements OkHttpObservationConv
         return keyValues;
     }
 
-    private String getUriTag(Function<Request, String> urlMapper, @Nullable Request request) {
+    private String getUriTag(Function<@Nullable Request, String> urlMapper, @Nullable Request request) {
         if (request == null) {
             return TAG_VALUE_UNKNOWN;
         }
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
@@ -89,11 +89,11 @@ public class OkHttpMetricsEventListener extends EventListener {
 
     private final String requestsMetricName;
 
-    private final Function<Request, String> urlMapper;
+    private final Function<@Nullable Request, String> urlMapper;
 
     private final Iterable<Tag> extraTags;
 
-    private final Iterable<BiFunction<Request, Response, Tag>> contextSpecificTags;
+    private final Iterable<BiFunction<@Nullable Request, @Nullable Response, Tag>> contextSpecificTags;
 
     private final Iterable<Tag> unknownRequestTags;
 
@@ -103,13 +103,14 @@ public class OkHttpMetricsEventListener extends EventListener {
     final ConcurrentMap<Call, CallState> callState = new ConcurrentHashMap<>();
 
     protected OkHttpMetricsEventListener(MeterRegistry registry, String requestsMetricName,
-            Function<Request, String> urlMapper, Iterable<Tag> extraTags,
-            Iterable<BiFunction<Request, Response, Tag>> contextSpecificTags) {
+            Function<@Nullable Request, String> urlMapper, Iterable<Tag> extraTags,
+            Iterable<BiFunction<@Nullable Request, @Nullable Response, Tag>> contextSpecificTags) {
         this(registry, requestsMetricName, urlMapper, extraTags, contextSpecificTags, emptyList(), true);
     }
 
-    OkHttpMetricsEventListener(MeterRegistry registry, String requestsMetricName, Function<Request, String> urlMapper,
-            Iterable<Tag> extraTags, Iterable<BiFunction<Request, Response, Tag>> contextSpecificTags,
+    OkHttpMetricsEventListener(MeterRegistry registry, String requestsMetricName,
+            Function<@Nullable Request, String> urlMapper, Iterable<Tag> extraTags,
+            Iterable<BiFunction<@Nullable Request, @Nullable Response, Tag>> contextSpecificTags,
             Iterable<String> requestTagKeys, boolean includeHostTag) {
         this.registry = registry;
         this.requestsMetricName = requestsMetricName;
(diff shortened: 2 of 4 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/DefaultOkHttpObservationConvention.java
```diff
@@ -98,7 +99,7 @@ public KeyValues getLowCardinalityKeyValues(OkHttpContext context) {
         return keyValues;
     }
 
-    private String getUriTag(Function<Request, String> urlMapper, @Nullable Request request) {
+    private String getUriTag(Function<@Nullable Request, String> urlMapper, @Nullable Request request) {
```
jckoenen: Taking into context the `if` condition immediately below, the function is simply not called when there is no request. Since there are no other usages of `urlMapper` or `context.urlMapper` I think this is correct to stay implicitly `@NotNull` which then also doesn't alter the public API.

(Obviously I don't have experience in your codebase, so I'm probably missing something. Please disregard in that case)

jckoenen, line 102: Taking into account the `if` condition on the following line, I believe the `urlMapper` can remain untouched. AFAICT there are no unchecked usages of `OkhttpContext.getUrlMapper` in the micrometer codebase, so it should be safe to keep the public API in this case.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
```diff
@@ -89,11 +89,11 @@ public class OkHttpMetricsEventListener extends EventListener {
 
     private final String requestsMetricName;
 
-    private final Function<Request, String> urlMapper;
+    private final Function<@Nullable Request, String> urlMapper;
 
     private final Iterable<Tag> extraTags;
 
-    private final Iterable<BiFunction<Request, Response, Tag>> contextSpecificTags;
+    private final Iterable<BiFunction<@Nullable Request, @Nullable Response, Tag>> contextSpecificTags;
```
jckoenen, line 96: I went through the codebase this morning with a fresh mind, so if I understand correctly the whole rabbit hole of nullability boils down to the fact that the public API `OkHttpContext` does not override the nullable [`SenderContext.getCarrier` ](https://github.com/micrometer-metrics/micrometer/blob/5452aeae740ccf1b67229b779a0a753fc8312485/micrometer-observation/src/main/java/io/micrometer/observation/transport/SenderContext.java#L60)

I understand that changing the context is not really possible without _widely_ affecting public API, but I believe this doesn't apply to `OkHttpMetricsEventListener`, which is completely isolated from that context (and pretty isolated in terms of public API, too). In this class, I think we can fully determine nullability and only need `<_, @Nullable Response, _>`:

- The only entry-point is [`callStart`](https://github.com/micrometer-metrics/micrometer/blob/edd49a28c09b95dedf9059f5cc207146d158c601/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java#L134). Here, `Request` is guaranteed not null, `Response` is guaranteed null
- Other lifecycle callbacks do not add new entries to the shared `callState` map, and do not alter any entry's `Request`. So, its still guaranteed not null
- [`responseHeadersEnd`](https://github.com/micrometer-metrics/micrometer/blob/edd49a28c09b95dedf9059f5cc207146d158c601/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java#L152) potentially updates the `Response`

Since `CallState` is private API, we know these are all relevant places, so we should be fine to drop the `@Nullable` annotation from `CallState.Request`

jckoenen:
Existing tests actually already cover the failing scenario, they are just missing the assertions which kotlin would insert. This will cause a failure:

```diff
diff --git a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
index 3e6809089..5f9268805 100644
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListenerTest.java
@@ -100,6 +100,11 @@ class OkHttpMetricsEventListenerTest {
 
         OkHttpClient client = new OkHttpClient.Builder().connectTimeout(1, TimeUnit.MILLISECONDS)
             .eventListener(OkHttpMetricsEventListener.builder(registry, "okhttp.requests")
+                .tag((req, res) -> {
+                    assertThat(req).isNotNull();
+                    assertThat(res).isNotNull();
+                    return Tag.of("key", "asdf");
+                })
                 .tags(Tags.of("foo", "bar"))
                 .uriMapper(URI_MAPPER)
                 .build())
```

I can also confirm that updating the annotations will really produce the expected behaviour when using kotlin (using the exact piece of code that triggered it before):

<details>
<summary> Bytecode with <code>@Nullable Response</code> </summary>

```
  private final static withMetrics$lambda$0$1(Lokhttp3/Request;Lokhttp3/Response;)Lio/micrometer/core/instrument/Tag;
   L0
    ALOAD 0
    LDC "request"
    INVOKESTATIC kotlin/jvm/internal/Intrinsics.checkNotNullParameter (Ljava/lang/Object;Ljava/lang/String;)V
   L1
    LINENUMBER 50 L1
    ALOAD 0
    INVOKEVIRTUAL okhttp3/Request.url ()Lokhttp3/HttpUrl;
    
    ...
```

</details>

<details>
<summary> Original Bytecode </summary>

```
  private final static withMetrics$lambda$0$1(Lokhttp3/Request;Lokhttp3/Response;)Lio/micrometer/core/instrument/Tag;
   L0
    ALOAD 0
    LDC "request"
    INVOKESTATIC kotlin/jvm/internal/Intrinsics.checkNotNullParameter (Ljava/lang/Object;Ljava/lang/String;)V
    ALOAD 1
    LDC "<unused var>"
    INVOKESTATIC kotlin/jvm/internal/Intrinsics.checkNotNullParameter (Ljava/lang/Object;Ljava/lang/String;)V
   L1
    LINENUMBER 50 L1
    ALOAD 0
    INVOKEVIRTUAL okhttp3/Request.url ()Lokhttp3/HttpUrl;
    
    ...
```

</details>

jonatan-ivanov:
@shakuzen

>I don't think we can know the Request is not null unless we check because it is marked @Nullable on the CallState, and the person who reported the issue apparently ran into it being null in some case.

Not in this PR, but I would get rid of the `CallState` entirely (there are multiple problems with it: being package-private but "leaked" on a public API is the most pressing one). In OkHttp's Kotlin API the request is not null-marked so it cannot be null which I think we can trust, it becomes nullable because of our own code. :(

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/DefaultOkHttpObservationConvention.java
jonatan-ivanov: You are right (this is the only place when it is called). I synchronized it with `BiFunction` because leaving them different might make things worse for the users when they pass them to the ctor of `OkHttpContext`.
```java
OkHttpContext(
    Function<@Nullable Request, String> urlMapper,
    ...,
    Iterable<BiFunction<@Nullable Request, @Nullable Response, KeyValue>> contextSpecificTags,
    ...)
```
I found it confusing to leave Request non-null for the `Function` but nullable on `BiFunction` but not sure which one is better for the future (I want to properly fix this).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/okhttp3/OkHttpMetricsEventListener.java
jonatan-ivanov: >I understand that changing the context is not really possible without widely affecting public API,

I think I might be able to fix things by just breaking this:
```java
    public void setState(OkHttpObservationInterceptor.CallState state) {
        this.state = state;
    }

    public OkHttpObservationInterceptor.@Nullable CallState getState() {
        return state;
    }
```
These are public methods but `CallState` is package-private so users are not able to use it anyways. (Though I don't like the parameters of the `OkHttpContext` either, they should be on the convention I think and that's really a breaking change.)

>I believe this doesn't apply to OkHttpMetricsEventListener, which is completely isolated from that context (and pretty isolated in terms of public API, too). In this class, I think we can fully determine nullability and only need <_, @Nullable Response, _>

I applied the same thought-process here as in https://github.com/micrometer-metrics/micrometer/pull/7375#discussion_r3042144639, to keep things consistent(ly wrong).

shakuzen: > These are public methods but `CallState` is package-private so users are not able to use it anyways.

I wonder if there is some static analysis or ErrorProne capability we could use to prevent adding such broken public API to begin with (public API that requires non-public types). I'm fine breaking that in 1.17.0-RC1 since external users can't use the API anyway, right?

jonatan-ivanov: They can do `context.setState(null)` and/or `Object obj = context.getState()` but neither makes a lot of sense.
I haven't seen any errorprone check for this: https://www.jetbrains.com/help/inspectopedia/ClassEscapesDefinedScope.html#locating-this-inspection
Maybe we should open an issue and/or check if we can use spotbugs or pmd for this.

jonatan-ivanov:
Fyi, I needed to fix tests after merge: https://github.com/micrometer-metrics/micrometer/commit/55b8f066ce6dbb637ccb54e626dadab605885a3e

Tests were ok locally for me so I did not even followed them through, NullAway should have screamed that my changes broke tests. The PR build was also fine. But after merge, this change broke the build. The failure is valid, the tests did not follow the implementation so I fixed them separately.

Right now I have no idea why this did not fail locally and (more importantly) on the PR build.

PR #7378, opened 2026-04-04
Capture actual response status in TimedHandler.onComplete()

### LINKED ISSUE #7276: Jetty 12's TimedHandler marks some requests with outcome UNKNOWN

blaspat:
Jetty 12 fires onResponseBegin before the handler sets a response status, so the status captured in onResponseBegin is 0 (unset). This causes all responses to be tagged with Outcome.UNKNOWN instead of the correct outcome based on the actual HTTP status code.

The fix stores the Response as a request attribute in handle() and retrieves it in onComplete() to update RESPONSE_STATUS_ATTRIBUTE with the actual status. This ensures correct status and outcome tags are recorded regardless of when the handler sets the response status.

Fixes #7276

Signed-off-by: Blasius Patrick <blasius.patrick@gmail.com>

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -39,7 +39,7 @@ javax-inject = "1"
 jaxb = "2.3.1"
 jetty9 = "9.4.58.v20250814"
 jetty11 = "11.0.16"
-jetty12 = "12.0.6"
+jetty12 = "12.0.8"
 jersey3 = "3.1.11"
 jmh = "1.37"
 # latest version of jOOQ to run tests against
--- a/micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
+++ b/micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
@@ -97,17 +97,12 @@ public class TimedHandler extends EventsHandler implements Graceful {
 
     @Override
     protected void onAfterHandling(Request request, boolean handled, Throwable failure) {
-        // Same issue as onComplete: onResponseBegin fires before the handler sets
-        // the status, so the attribute may still be 0. Update it here too so
-        // stopHandlerTiming sees the correct status for the handler timer.
         stopHandlerTiming(request);
         super.onAfterHandling(request, handled, failure);
     }
 
     @Override
     protected void onResponseBegin(Request request, int status, HttpFields headers) {
-        // If we see a status of 0 here, that mean the Handler hasn't set a status code.
-        request.setAttribute(RESPONSE_STATUS_ATTRIBUTE, status);
         super.onResponseBegin(request, status, headers);
     }
 
--- a/micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
+++ b/micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
@@ -97,6 +97,9 @@ public class TimedHandler extends EventsHandler implements Graceful {
 
     @Override
     protected void onAfterHandling(Request request, boolean handled, Throwable failure) {
+        // Same issue as onComplete: onResponseBegin fires before the handler sets
+        // the status, so the attribute may still be 0. Update it here too so
+        // stopHandlerTiming sees the correct status for the handler timer.
         stopHandlerTiming(request);
         super.onAfterHandling(request, handled, failure);
     }
@@ -109,9 +112,14 @@ public class TimedHandler extends EventsHandler implements Graceful {
     }
 
     @Override
-    protected void onComplete(Request request, Throwable failure) {
+    protected void onComplete(Request request, int status, HttpFields headers, Throwable failure) {
+        // Jetty 12 fires onResponseBegin before the handler sets a response status,
+        // so the status captured in onResponseBegin is 0 (unset). Use the status
+        // parameter from this 4-param onComplete overload — Jetty 12.1+ provides
+        // the actual status here directly.
+        request.setAttribute(RESPONSE_STATUS_ATTRIBUTE, status);
         stopRequestTiming(request);
-        super.onComplete(request, failure);
+        super.onComplete(request, status, headers, failure);
     }
 
     private void beginRequestTiming(Request request) {
--- a/micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
+++ b/micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
@@ -29,7 +29,7 @@ import org.eclipse.jetty.util.component.Graceful;
 import java.util.concurrent.CompletableFuture;
 
 /**
- * Jetty 12 Metrics Handler.
+ * Jetty 12 Server Metrics Handler. The implementation requires at least Jetty 12.0.8.
  *
  * @author Jon Schneider
  * @author Joakim Erdfelt
@@ -108,10 +108,6 @@ public class TimedHandler extends EventsHandler implements Graceful {
 
     @Override
     protected void onComplete(Request request, int status, HttpFields headers, Throwable failure) {
-        // Jetty 12 fires onResponseBegin before the handler sets a response status,
-        // so the status captured in onResponseBegin is 0 (unset). Use the status
-        // parameter from this 4-param onComplete overload — Jetty 12.1+ provides
-        // the actual status here directly.
         request.setAttribute(RESPONSE_STATUS_ATTRIBUTE, status);
         stopRequestTiming(request);
         super.onComplete(request, status, headers, failure);
--- a/micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
+++ b/micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
@@ -250,6 +250,135 @@ class TimedHandlerTest {
         }
     }
 
+    @Test
+    void statusIsCorrectlyRecordedForNotFoundRequest() throws Exception {
+        CountDownLatch latch = new CountDownLatch(1);
+        timedHandler.setHandler(new Handler.Abstract() {
+            @Override
+            public boolean handle(Request request, Response response, Callback callback) {
+                response.setStatus(404);
+                response.write(true, BufferUtil.EMPTY_BUFFER, callback);
+                latch.countDown();
+                return true;
+            }
+        });
+        server.start();
+
+        try (LocalConnector.LocalEndPoint endpoint = connector.connect()) {
+            String request = "GET /notfound HTTP/1.1\r\n" + "Host: localhost\r\n" + "\r\n";
+            endpoint.addInputAndExecute(request);
+
+            assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+            assertThat(timedHandler.awaitOnComplete(5, TimeUnit.SECONDS)).isTrue();
+
+            HttpTester.Response response = HttpTester.parseResponse(endpoint.getResponse());
+            assertThat(response.getStatus()).isEqualTo(HttpStatus.NOT_FOUND_404);
+
+            // Verify correct status and outcome tags
+            assertThat(registry.get("jetty.server.requests")
+                .tag("outcome", Outcome.CLIENT_ERROR.name())
+                .tag("method", "GET")
+                .tag("status", "404")
+                .timer()
+                .count()).isEqualTo(1);
+        }
+    }
+
+    @Test
+    void statusIsCorrectlyRecordedForServerError() throws Exception {
+        CountDownLatch latch = new CountDownLatch(1);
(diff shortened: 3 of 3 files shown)


blaspat:
Sorry for the confusion — the previous PR #7374 was closed due to a technical issue on my end. This is the same fix with all the same commits, re-opened as a new PR. Thanks for your patience!

micrometer-jetty12/src/test/java/io/micrometer/jetty12/server/TimedHandlerTest.java
```diff
@@ -250,6 +250,211 @@ else if ("/b".equals(path)) {
...
+            assertThat(registry.get("jetty.server.requests")
+                .tag("outcome", Outcome.SUCCESS.name())
+                .tag("status", "200")
+                .timer()
+                .count()).isEqualTo(1);
+        }
+    }
+
+    @Test
+    void statusIsCorrectlyRecordedWhenSetAfterOnResponseBegin() throws Exception {
```
shakuzen: This test is redundant with `statusIsCorrectlyRecordedForSuccessfulRequest` I think. I would remove this and leave `statusIsCorrectlyRecordedForSuccessfulRequest`.

micrometer-test/src/main/java/io/micrometer/common/util/internal/logging/MockLogger.java
shakuzen, line 1: Changes to this file look unrelated to the other changes in the pull request. Why are these changes included?

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
```diff
@@ -97,6 +99,16 @@ protected void onBeforeHandling(Request request) {
...
+        // Same issue as onComplete: onResponseBegin fires before the handler sets
+        // the status, so the attribute may still be 0. Update it here too so
+        // stopHandlerTiming sees the correct status for the handler timer.
+        Response response = (Response) request.getAttribute(RESPONSE_ATTRIBUTE);
+        if (response != null) {
+            int status = response.getStatus();
+            if (status > 0) {
+                request.setAttribute(RESPONSE_STATUS_ATTRIBUTE, status);
+            }
+        }
```
shakuzen: It looks like this added code could be made into a private method and called from the two places it is used.

shakuzen (commented):
Overall looks good. Thank you for this. I left a few comments.

micrometer-test/src/main/java/io/micrometer/common/util/internal/logging/MockLogger.java
blaspat: reverted for this file

shakuzen (approved):
The changes look good. The DCO check is failing, though.

shakuzen:
Mentioning @joakime in case you have any concerns about this fix.

shakuzen:
We'll want to rebase this on `1.15.x` so we can merge forward the fix from there. I can handle that when merging, though.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
```diff
@@ -97,6 +99,11 @@ protected void onBeforeHandling(Request request) {
 
     @Override
     protected void onAfterHandling(Request request, boolean handled, Throwable failure) {
+        // Same issue as onComplete: onResponseBegin fires before the handler sets
+        // the status, so the attribute may still be 0. Update it here too so
+        // stopHandlerTiming sees the correct status for the handler timer.
```
joakime: The `onAfterHandling()` occurs when the `Handler.handle()` call returns (either successfully, or due to unhandled exception).

If the status code being set is deferred (in an AsyncContext for example), then this event will not show a status code.

If the status is null during `onResponseBegin()` then the status code wasn't set by the `Handler` (the webapp, the servlet, the filter, the asynccontext, etc), which will result in Jetty determining what status code to actually use (for both the success case, and the error case).

An alternative approach would be to use `Request.addCompletionListener(Request request, Consumer<Throwable> listener)` and capture the suceeded/failed event of the handled request.

If you want to know what the application (or the error handler) set the status code to, you can always wrap the `Response` instance and capture the `setStatus(int)` call for your own purposes.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
```diff
@@ -110,10 +117,24 @@ protected void onResponseBegin(Request request, int status, HttpFields headers)
 
     @Override
     protected void onComplete(Request request, Throwable failure) {
```
joakime: You are using a deprecated event here.

This was addressed in 

* https://github.com/jetty/jetty.project/pull/11487

Use ...

``` java
@Override
protected void onComplete(Request request, int status, HttpFields headers, Throwable failure)
{
    // the status code here is valid
}
```

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
joakime: Ultimately, I think you should use the new `onComplete(Request request, int status, HttpFields headers, Throwable failure)` method instead of this Request/Response/Attribute juggling.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
```diff
@@ -170,6 +191,7 @@ private Timer.Sample getTimerSample(Request request) {
     public boolean handle(Request request, Response response, Callback callback) throws Exception {
         Timer.Sample sample = Timer.start(registry);
         request.setAttribute(SAMPLE_TIMER_ATTRIBUTE, sample);
+        request.setAttribute(RESPONSE_ATTRIBUTE, response);
```
joakime: Not necessary anymore, use the new onComplete method.

gradle/libs.versions.toml
```diff
@@ -40,7 +40,7 @@ javax-inject = "1"
 jaxb = "2.3.1"
 jetty9 = "9.4.58.v20250814"
 jetty11 = "11.0.16"
-jetty12 = "12.0.6"
+jetty12 = "12.1.6"
```
joakime: Note, Jetty 12.1.8 is now available

blaspat: Done bump version
Thanks for the info

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
blaspat: Done, updated with newer method

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
blaspat: Done, switched to the 4-param `onComplete` as you recommended.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
blaspat: Done, switched to the 4-param onComplete as you recommended.

shakuzen:
@blaspat help me understand the status of this PR. It looks like the last force-pushed commit reverts back to the implementation before Joakim's review. Was that intended?

Since we want to merge this into `1.15.x`, we need to be careful about depending on API that effectively bumps the minimum version of Jetty required. That doesn't mean we absolutely can't, but we need to be conscious of it and call it out in the release notes.

If some API that makes this fix better is only available in Jetty 12.1.x, we can apply that fix separately in the next minor version of Micrometer (1.18.0) and call out the new minimum Jetty version. Separate from that, it would be good to have a fix for maintenance branches (1.15.x, 1.16.x, and soon 1.17.x) that does not increase the minimum version of Jetty, at least not by a new minor version.

joakime:
@shakuzen I strongly encourage use of the new `onComplete(Request request, int status, HttpFields headers, Throwable failure)` method, not the deprecated one + the hacks present here.

See:

* https://github.com/micrometer-metrics/micrometer/pull/7378#discussion_r3045035289

shakuzen:
@joakime based on the PR you linked in Jetty, it looks like that's been available since 12.0.8. Is that right? If so, we should note that this instrumentation requires at least 12.0.8.

joakime:
@shakuzen yes, that's correct.

Using versions before Jetty 12.0.8 is discouraged anyway, as they have active vulnerabilities.

* https://jetty.org/security.html

blaspat:
@shakuzen Yes the latest force push was a mistake. I’m sorry. As for now, I will update using latest Jetty12 version (12.1.9) if thats okay?

shakuzen:
@blaspat the plan is to rebase this change on the `1.15.x` branch so we can fix this in maintenance releases. For that reason, we do not want to bump the minor version of Jetty with this change (because we might accidentally use some API only available in 12.1.x and users on 12.0.x will be broken). Jetty dependencies are optional dependencies for the Jetty 12 module that we only compile against. So generally, we are compiling against the oldest version the instrumentation should work with. Given the discussion above, I believe that is 12.0.8. Users need to decide what version they want separately.

As an improvement that could be considered in a separate change, we could consider running tests with the latest version(s) in addition to compiling against the oldest supported version.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
shakuzen: It looks like this still needs to be addressed, right?

blaspat:
Now updated using Jetty 12.0.8. and using 4-param `onComplete`

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
shakuzen: Does this serve any purpose if we're setting it in `onComplete` now?

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
```diff
@@ -97,6 +97,9 @@ protected void onBeforeHandling(Request request) {
 
     @Override
     protected void onAfterHandling(Request request, boolean handled, Throwable failure) {
+        // Same issue as onComplete: onResponseBegin fires before the handler sets
+        // the status, so the attribute may still be 0. Update it here too so
+        // stopHandlerTiming sees the correct status for the handler timer.
```
shakuzen: This comment is outdated now and can be removed, unless I'm missing some reason to keep it.

shakuzen (commented):
I've rebased onto `1.15.x` and left a couple review comments.

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
blaspat: comment removed

micrometer-jetty12/src/main/java/io/micrometer/jetty12/server/TimedHandler.java
blaspat: no, i removed it now. thanks

shakuzen (approved):
Thanks for the pull request and follow-ups to the feedback. I've added a polish commit to clean up some of the comments and document in the class JavaDoc the minimum Jetty version required for the implementation.

shakuzen:
Also thank you @joakime for the feedback that resulted in a better implementation of this fix. We'll release this fix in our upcoming maintenance releases.

PR #7384, opened 2026-04-07
feat: add metrics for JVM safepoints

schiemon:


jonatan-ivanov:
@schiemon Did you close this by accident or is there any other reason for it?
If this is still valid and we decide JFR is the way for this, I think we should wait with this and do something like this first: https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:jfr-binder

schiemon:
@jonatan-ivanov 

> Did you close this by accident

Jep :)

> If this is still valid and we decide JFR is the way for this, I think we should wait with this and do something like this first

Agree, I submitted a PR: #7531

Issue #5658, opened 2024-11-12
Replace the OpenTelemetry protobuf library


GFriedrich:
**Please describe the feature request.**
Get rid of the OpenTelemetry protobuf library as it is not intended for public use.

**Rationale**
Recently https://github.com/micrometer-metrics/micrometer/issues/5493 was opened to express that some vulnerable protobuf version is used. The ticket was closed though, because the protobuf version is coming from the OpenTelemetry protobuf library. Now though, the OpenTelemetry maintainers decided that this library is not intended for public use (see https://github.com/open-telemetry/opentelemetry-proto-java/pull/20). Therefore I guess the only option here is to get rid of this library and generating the files for protobuf by yourselves.

jonatan-ivanov:
Thank you for the issue!

It seems [CVE-2024-7254](https://github.com/advisories/GHSA-735f-pc8j-v9w8) is about parsing (untrusted) data. I don't think Micrometer is affected since it does not parse any protobuf payload (deserializing), it does the opposite, it produces protobuf payload (serializing). From security scanners perspective, this seems like a false-positive to me. It also worth to mention that other than being affected or not, from Micrometer's perspective, this should not be a CVE since users should be able to upgrade the protobuf dependency (`protobuf-java:3.25.5` should be fixed).

We need to discuss internally how to handle the new `opentelemetry-proto` guidance. According to it:
>[...] please generate your own bindings, consulting [grpc codegen](https://grpc.io/docs/languages/java/generated-code/#codegen) and possibly [build.gradle.kts](https://github.com/open-telemetry/opentelemetry-proto-java/blob/main/build.gradle.kts)

and as you mentioned, we should be able to simply generate the code from the proto specs using grpc codegen. Unfortunately, this is not as simple as it sounds. I asked around the [OTel Slack](https://cloud-native.slack.com/archives/C014L2KCTE3/p1731457239819089), it seems that:
- The [types](https://github.com/open-telemetry/opentelemetry-java/tree/15669842624cd95b662c90575d5678837d362293/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data) that otel-java itself uses are not generated using codegen either but manually created and maintained.
- otel-java cannot rely on the default grpc codegen tooling because upgrading codegen can introduce binary incompatible changes in the generated code (this happened in OTel multiple occasions in the past).
- The spec has a few customizations, in some cases (traceId/spanID), it defines a different serialized representation than the default behavior of codegen.
[...]

PR #7430, opened 2026-04-24
fix: Ensure OTLP histogram count is exactly the sum of buckets count

### LINKED ISSUE #6006: OTLP histogram bucket count does not equal count intermittently
**Describe the bug**
We upgraded to Micrometer 1.14.0 and started using exponential histograms around 2 months ago.
Since day 1 of using exponential histograms we noticed this error on our metrics backend (Grafana GEM):
```
ts=2025-03-05T05:07:37.899310792Z caller=grpc_logging.go:76 level=warn method=/cortex.Ingester/Push duration=3.033186ms msg=gRPC err="user=default-tenant: err: 8 observations found in buckets, but the Count field is 10: histogram's observation count should equal the number of observations found in the buckets (in absence of NaN). timestamp=2025-03-05T05:07:36.402Z, series=[{__name__ proxy_convergence_duration}] (err-mimir-native-histogram-count-mismatch) (sampled 1/10)
```

It's saying total count of observations in buckets doen't match the Count field. This seems like a bug

 - Micrometer version: 1.14.0
 - Micrometer registry: OTLP
 - OS: linux
 - Java version: 11

bogdandrutu:
Fixes https://github.com/micrometer-metrics/micrometer/issues/6006, see issue comments for more explanations.

### DIFF
--- a/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMetricConverterTest.java
+++ b/implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMetricConverterTest.java
@@ -20,7 +20,6 @@ import io.micrometer.core.instrument.config.NamingConvention;
 import io.micrometer.core.instrument.distribution.CountAtBucket;
 import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
 import io.micrometer.core.instrument.distribution.HistogramSnapshot;
-import io.micrometer.core.instrument.distribution.HistogramSupport;
 import io.micrometer.core.instrument.distribution.ValueAtPercentile;
 import io.micrometer.registry.otlp.internal.ExponentialHistogramSnapShot;
 import io.opentelemetry.proto.metrics.v1.ExponentialHistogramDataPoint;
@@ -33,7 +32,6 @@ import org.junit.jupiter.api.Test;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


shakuzen:
Thanks for the PR, Bogdan. Sorry for the long delay on reviewing this. This has been obviated on `main` with the transition to using the shaded OTel Metrics SDK under the hood (see https://github.com/micrometer-metrics/micrometer/issues/5658). With that, we no longer explicitly set the count and instead it is derived from the bucket counts:

https://github.com/micrometer-metrics/micrometer/blob/02917d0870ea96e1e895f643bfd2b09be91e677c/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java#L225-L226

So I think there's no longer anything to fix on `main`. We should probably adopt this fix for maintenance branches, however. Would you rebase on `1.16.x`? Otherwise, I can take care of doing that and add a test for this case.

shakuzen:
I've rebased your changes on `1.16.x` and added some unit tests.

jonatan-ivanov (approved):

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMetricConverterTest.java
```diff
@@ -191,4 +202,104 @@ void addMeterWithDistributionSummary() {
...
+            return snapshot.max();
+        }
+
+    }
+
+    @Test
+    void histogramCountShouldBeSumOfBucketCounts() {
+        Meter.Id id = new Meter.Id("test.histogram", Tags.empty(), null, null, Meter.Type.DISTRIBUTION_SUMMARY);
+        HistogramSnapshot snapshot = new HistogramSnapshot(10, 100, 20, new ValueAtPercentile[0],
+                new CountAtBucket[] { new CountAtBucket(10.0, 3.0), new CountAtBucket(20.0, 5.0) }, null);
```
shakuzen, line 259: Validation on `main` causes this to fail. We always ensure the positive infinity bucket is included, so it should be here as well. Fixed in 77b2902cdfc0e0d1370247164e7e8140b63a7342

Issue #7409, opened 2026-04-17
MeterRegistry closes HighCardinalityTagsDetector outside closed conditional check


edwardsre:
I am using Micrometer 1.17.0-M3 via Spring Boot 4.1.0-M3, but this problem exists prior to this.

Spring Boot calls `close()` on the MeterRegistry as part of application shutdown and then calls it again because it is a bean with a `close()` method. See [this Spring Boot issue](https://github.com/spring-projects/spring-boot/issues/50094#issuecomment-4268162680). This causes multiple log output statements when a HighCardinalityTagsDetector is configured. This is exacerbated by having multiple meter registries configured with a detector.

This seems benign, but the multiple logs makes it appear as though there are more detector instances being shutdown than actually are which leads to confusion.

In `MeterRegistry.close()`, the detector close method is called outside the if closed conditional. Not sure if this is intentional or not.

```java
public void close() {
  if (closed.compareAndSet(false, true)) {
    synchronized (meterMapLock) {
      for (Meter meter : meterMap.values()) {
        meter.close();
      }
    }
  }

  if (highCardinalityTagsDetector != null) {
    highCardinalityTagsDetector.close();  // This method outputs the log statement
  }
}
```

**Log output**
```
2026-04-16T16:08:11.890-06:00  INFO 68464 --- [demo] [ionShutdownHook] i.m.c.i.HighCardinalityTagsDetector      : Stopping HighCardinalityTagsDetector
2026-04-16T16:08:11.891-06:00  INFO 68464 --- [demo] [ionShutdownHook] o.s.boot.tomcat.GracefulShutdown         : Commencing graceful shutdown. Waiting for active requests to complete
2026-04-16T16:08:13.907-06:00  INFO 68464 --- [demo] [tomcat-shutdown] o.s.boot.tomcat.GracefulShutdown         : Graceful shutdown complete
2026-04-16T16:08:13.917-06:00  INFO 68464 --- [demo] [ionShutdownHook] i.m.c.i.HighCardinalityTagsDetector      : Stopping HighCardinalityTagsDetector
```

**Log output with two meter registries configured in Spring Boot**
```
2026-04-16T16:41:46.250-06:00  INFO 77024 --- [demo] [ionShutdownHook] i.m.c.i.HighCardinalityTagsDetector      : Stopping HighCardinalityTagsDetector
[...]

PR #7436, opened 2026-04-27
Fix/meter registry close detector

blaspat:
## Summary
Move `highCardinalityTagsDetector.close()` inside the `if (closed.compareAndSet(false, true))` guard block in `MeterRegistry.close()`, preventing duplicate log output when frameworks like Spring Boot call close() multiple times.

## Fix
`MeterRegistry.close()` was calling `highCardinalityTagsDetector.close()` outside the closed-guard, causing `HighCardinalityTagsDetector.shutdown()` to log "Stopping HighCardinalityTagsDetector" every time close() was invoked — even on subsequent calls after the first.

## Test
Added `closeShouldOnlyCloseHighCardinalityTagsDetectorOnce()` test in `MeterRegistryTest.java` that registers a detector and calls `registry.close()` twice, verifying the detector's close is only invoked once.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -1271,10 +1271,9 @@ public abstract class MeterRegistry {
                     meter.close();
                 }
             }
-        }
-
-        if (highCardinalityTagsDetector != null) {
-            highCardinalityTagsDetector.close();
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
Could you please rebase this on top of 1.15.x? See: https://github.com/micrometer-metrics/micrometer/issues/7409#issuecomment-4282616638

blaspat:
Done, here's the new PR https://github.com/micrometer-metrics/micrometer/pull/7441

PR #7454, opened 2026-04-30
Prometheus: construct MetricMetadata and Labels on meter creation

### LINKED ISSUE #7450: Prometheus: unnecessary allocation and validation performed on every scrape
See this part of code (taken from commit at v1.16.5):

https://github.com/micrometer-metrics/micrometer/blob/cf727d0a9605c1aea58c26c5e72fb9f6fe84df2a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java#L209-L222

`getMetadata` and `Labels.of` are called within the inner lambda, the one that implements `Child` and is called on every scrape. These calls construct the same `Metadata` and `Labels` objects over and over, which in turn also performs repeated validation (`PrometheusNaming.validateMetricName` and `PrometheusNaming.validateLabelName`), verifying _on each scrape_ that the metric name and label names are valid UTF-8.

The `Metadata` and `Labels` can in fact be created and validated once, because the `conventionName` and `tagKeys` are stored in the `MicrometerCollector` and not modified during its lifetime. This will make scraping more efficient in terms of performance and garbage.

The same applies to all meter types, not only to counters.

If the maintainers agree with the idea of this change, I can open a PR to implement it.

m-burst:
Currently `MetricMetadata` and `Labels` are constructed on each scrape. The downsides of this are
* unnecessary allocation of these two objects themselves
* unnecessary validation of the metric name and label names via `PrometheusNaming.validateMetricName` and `PrometheusNaming.isValidLabelName`
	* which perform UTF-8 validation via `StandardCharsets.UTF_8.newEncoder().canEncode(name)`
	    * which in turn allocates the encoder and a `CharBuffer` to hold the encoded string data, and consumes CPU to perform the validation

In fact these objects can only be constructed once because the `conventionName` is actually fixed (`MicrometerCollector#conventionName` is final), and both `tagKeys` and `tagValues` are already computed once.

The first commit implements the pre-construction of `MetricMetadata` and labels.

The second commit removes the `conventionName` from `MicrometerCollector$Child` signature. Since we pre-create the `MetricMetadata` using the `conventionName`, we already assume that `conventionName` is immutable, so passing it on every scrape seems unnecessary.

Fixes #7450

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
@@ -221,7 +221,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
         applyToCollector(id, (collector) -> {
             List<String> tagKeys = tagKeys(id);
             Labels labels = Labels.of(tagKeys, tagValues(id));
-            String conventionName = collector.getConventionName();
+            String conventionName = collector.conventionName;
             MetricFamilyDescriptor familyDescriptor = familyDescriptor(MetricType.COUNTER, conventionName, tagKeys,
                     id.getDescription());
             collector.add(id, () -> Stream.of(new MicrometerCollector.Family<>(conventionName,
@@ -242,7 +242,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
             List<String> tagKeys = tagKeys(id);
             Labels labels = Labels.of(tagKeys, tagValues(id));
             MetricType primaryType = summary.histogramCounts().length == 0 ? MetricType.SUMMARY : MetricType.HISTOGRAM;
-            String conventionName = collector.getConventionName();
+            String conventionName = collector.conventionName;
             MetricFamilyDescriptor familyDescriptor = familyDescriptor(primaryType, conventionName, tagKeys,
                     id.getDescription());
             String conventionNameMax = conventionName + "_max";
@@ -267,7 +267,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
                     }
 
                     Exemplars exemplars = summary.exemplars();
-                    families.add(new MicrometerCollector.Family<>(collector.getConventionName(),
+                    families.add(new MicrometerCollector.Family<>(collector.conventionName,
                             family -> new SummarySnapshot(familyDescriptor.getMetadata(), family.dataPointSnapshots),
                             new SummaryDataPointSnapshot(count, sum, quantiles, labels, exemplars,
                                     createdTimestampMillis)));
@@ -296,7 +296,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
                     }
 
                     Exemplars exemplars = summary.exemplars();
-                    families.add(new MicrometerCollector.Family<>(collector.getConventionName(),
+                    families.add(new MicrometerCollector.Family<>(collector.conventionName,
                             family -> new io.prometheus.metrics.model.snapshots.HistogramSnapshot(false,
                                     familyDescriptor.getMetadata(), family.dataPointSnapshots),
                             new HistogramDataPointSnapshot(ClassicHistogramBuckets.of(buckets, counts), sum, labels,
@@ -340,7 +340,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
         applyToCollector(id, (collector) -> {
             List<String> tagKeys = tagKeys(id);
             Labels labels = Labels.of(tagKeys, tagValues(id));
-            String conventionName = collector.getConventionName();
+            String conventionName = collector.conventionName;
             MetricFamilyDescriptor familyDescriptor;
             if (id.getName().endsWith(".info")) {
                 familyDescriptor = familyDescriptor(MetricType.INFO, conventionName, tagKeys, id.getDescription());
@@ -379,7 +379,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
         applyToCollector(id, (collector) -> {
             List<String> tagKeys = tagKeys(id);
             Labels labels = Labels.of(tagKeys, tagValues(id));
-            String conventionName = collector.getConventionName();
+            String conventionName = collector.conventionName;
             MetricFamilyDescriptor familyDescriptor = familyDescriptor(MetricType.SUMMARY, conventionName, tagKeys,
                     id.getDescription());
             collector.add(id,
@@ -399,7 +399,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
         applyToCollector(id, (collector) -> {
             List<String> tagKeys = tagKeys(id);
             Labels labels = Labels.of(tagKeys, tagValues(id));
-            String conventionName = collector.getConventionName();
+            String conventionName = collector.conventionName;
             MetricFamilyDescriptor familyDescriptor = familyDescriptor(MetricType.COUNTER, conventionName, tagKeys,
                     id.getDescription());
             collector.add(id,
@@ -423,7 +423,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
             List<CustomMeterMeasurement> customMeasurements = new ArrayList<>();
 
             for (Measurement measurement : measurements) {
-                CustomMeterMetric metric = customMeterMetric(collector.getConventionName(), measurement.getStatistic());
+                CustomMeterMetric metric = customMeterMetric(collector.conventionName, measurement.getStatistic());
 
                 // De-duplicate names: metrics with the same Prometheus name will use the
                 // same descriptor and metadata
@@ -552,7 +552,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
         Labels labels = Labels.of(tagKeys, tagValues);
         MetricType primaryType = histogramSupport.takeSnapshot().histogramCounts().length == 0 ? MetricType.SUMMARY
                 : MetricType.HISTOGRAM;
-        String conventionName = collector.getConventionName();
+        String conventionName = collector.conventionName;
         MetricFamilyDescriptor familyDescriptor = familyDescriptor(primaryType, conventionName, tagKeys,
                 id.getDescription());
         String conventionNameMax = conventionName + "_max";
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
@@ -46,8 +46,9 @@ import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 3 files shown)


shakuzen:
It looks like this is going to heavily conflict with #7487. I think the best plan to reviewing and merging these will be to finish reviewing and merging #7487 first, and then rebase these changes on top of that after.

m-burst:
Hi @shakuzen, just checking in—is there any update or ETA on #7487? It looks like there hasn't been much activity on it recently, and I'm wondering when this PR might be unblocked. Thanks!

shakuzen:
Sorry for the long delay. The plan is to review and merge #7487 for the upcoming 1.18.0-M1 milestone. We have been busy with other work but should hopefully get back to that soon.

shakuzen:
@m-burst thank you for your patience. #7487 has been merged now. Would you resolve the conflicts? Then we can review this PR.

m-burst:
I will resolve the conflicts, but it may take me a while to reconcile my approach with the new changes in main.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -408,45 +421,38 @@ protected Meter newMeter(Meter.Id id, Meter.Type type, Iterable<Measurement> mea
...
             List<String> statKeys = new ArrayList<>(tagKeys);
             statKeys.add("statistic");
-            List<MetricFamilyDescriptor> registrationDescriptors = new ArrayList<>();
-            String name = getConventionName(id);
-            Set<String> descriptorNames = new HashSet<>();
+
+            Map<String, MetricFamilyDescriptor> descriptors = new LinkedHashMap<>();
+            List<CustomMeterMeasurement> customMeasurements = new ArrayList<>();
+
             for (Measurement measurement : measurements) {
```
m-burst, line 423: We now iterate over `measurements` once on creation time to cache required data in `customMeasurements`, and iterate over `customMeasurements` during scrape instead.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -470,21 +487,19 @@ private CustomMeterMetric customMeterMetric(String conventionName, Statistic sta
         }
     }
 
-    private MicrometerCollector.Family<CounterDataPointSnapshot> customCounterFamily(Meter.Id id,
-            CustomMeterMetric metric, Labels labels, double value) {
+    private MicrometerCollector.Family<CounterDataPointSnapshot> customCounterFamily(
+            CustomMeterMeasurement measurement) {
         long createdTimestampMillis = clock.wallTime();
```
m-burst, line 492: This is a separate bug: this timestamp should be captured on creation time. I did not fix it in this PR to avoid mixing unrelated changes, but after this PR the timestamp can be added to the `CustomMeterMeasurement` class

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -218,15 +219,16 @@ public Counter newCounter(Meter.Id id) {
...
         applyToCollector(id, (collector) -> {
-            List<String> tagValues = tagValues(id);
             List<String> tagKeys = tagKeys(id);
-            collector.add(id,
-                    (conventionName) -> Stream.of(new MicrometerCollector.Family<>(conventionName,
-                            family -> new CounterSnapshot(getMetadata(family.conventionName, id.getDescription()),
-                                    family.dataPointSnapshots),
-                            new CounterDataPointSnapshot(counter.count(), Labels.of(tagKeys, tagValues),
-                                    counter.exemplar(), createdTimestampMillis))),
-                    familyDescriptor(MetricType.COUNTER, getConventionName(id), tagKeys, id.getDescription()));
```
m-burst, line 229: I removed most calls to `getConventionName(id)` in favor of `collector.getConventionName()`. The only remaining call to `getConventionName(id)` is inside `applyToCollector` where the collector is actually created

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -585,19 +623,18 @@ private void addDistributionStatisticSamples(Meter.Id id, MicrometerCollector co
...
             return families.build();
-        }, familyDescriptor(primaryType, getConventionName(id), tagKeys, id.getDescription()),
-                familyDescriptor(MetricType.GAUGE, getConventionName(id) + "_max", tagKeys, id.getDescription()));
+        }, familyDescriptor, familyDescriptorMax);
     }
 
     private MetricFamilyDescriptor familyDescriptor(MetricType metricType, String name, Collection<String> labelNames,
             @Nullable String description) {
+        // Unit is intentionally not set, see:
+        // https://github.com/OpenObservability/OpenMetrics/blob/1386544931307dff279688f332890c31b6c5de36/specification/OpenMetrics.md#unit
```
m-burst, line 625: Since we are now using the `MetricMetadata` created by the `MetricFamilyDescriptor` instead of creating it ourselves, this comment moved here

m-burst (commented):
The PR should be ready for review now. I added a few inline comments to highlight some of the changes

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
m-burst: On second thought, this may have been incorrect.

The contract of `Meter#measure` requires the iterable to `return the same number of measurements and in the same order`, but these can be different instances with different value functions. If metric values are captured in the `Measurement` instances at iteration time, the scrapes will still obtain the metric values from the original measurements, and these values will be obsolete.

Alternative implementations may be:
* iterate over measurements at scrape time and get cached metadata by index (this would require some extra handling for misbehaving `Iterable<Measurement>` which return a changing number of measurements — even though such iterables violate the contract, we don't want to crash during scrape)
* revert the custom-meter part, so that metadata and labels for them are created at scrape time
* add a cache:
	* initially populate it from the initial measurements
	* use the `Statistic` value as the cache key, because the same `Statistic` would give the same metric name (hence the same `MetricFamilyDescriptor`) and the same label values
	* if at scrape time the cache does not have a value for a measurement, we need to decide whether to add it to the cache or not (adding will require synchronization)

I would appreciate the maintainers' feedback on this, cc @shakuzen

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -290,11 +296,11 @@ public DistributionSummary newDistributionSummary(Meter.Id id,
                     }
 
                     Exemplars exemplars = summary.exemplars();
-                    families.add(new MicrometerCollector.Family<>(conventionName,
+                    families.add(new MicrometerCollector.Family<>(collector.getConventionName(),
```
shakuzen: Nit: this could use `conventionName` instead of calling `getConventionName()`.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -260,11 +267,10 @@ public DistributionSummary newDistributionSummary(Meter.Id id,
                     }
 
                     Exemplars exemplars = summary.exemplars();
-                    families.add(new MicrometerCollector.Family<>(conventionName,
-                            family -> new SummarySnapshot(getMetadata(family.conventionName, id.getDescription()),
-                                    family.dataPointSnapshots),
-                            new SummaryDataPointSnapshot(count, sum, quantiles, Labels.of(tagKeys, tagValues),
-                                    exemplars, createdTimestampMillis)));
+                    families.add(new MicrometerCollector.Family<>(collector.getConventionName(),
```
shakuzen: Nit: this could use `conventionName` instead of calling `getConventionName()`.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
shakuzen: Thanks for all the thoughtful consideration on this.

> revert the custom-meter part, so that metadata and labels for them are created at scrape time

I think this will be the best option for now. We can follow-up on optimizing custom meters if we think it is worth it. I would probably wait until we get feedback from users relying on custom meters and affected by this behavior. I generally expect custom meters to be quite rare and, thus, not worth the attention built-in meter types deserve.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
shakuzen: Thanks for catching this, and good call to not fix it in the same PR. Would you open an issue/PR for this bug? I suspect it will need to be fixed in maintenance branches as well.

shakuzen (commented):
Thanks a lot for this. Looks good overall. I left some comments and responded to your question about the handling of custom meters.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
m-burst: This is now fixed everywhere: I changed the field from `private` to package-private, and removed the getter method.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
m-burst: I will open an issue today

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -585,22 +608,34 @@ private void addDistributionStatisticSamples(Meter.Id id, MicrometerCollector co
...
+        // Unit is intentionally not set, see:
+        // https://github.com/OpenObservability/OpenMetrics/blob/1386544931307dff279688f332890c31b6c5de36/specification/OpenMetrics.md#unit
         return MetricFamilyDescriptor.of(metricType, name).help(helpText(description)).labelNames(labelNames).build();
     }
 
+    /**
+     * Used only for custom meters; all other meters' metadata is obtained through
+     * {@link MetricFamilyDescriptor} returned from {@link familyDescriptor}.
+     */
+    private MetricMetadata getMetadata(String name, @Nullable String description) {
```
m-burst, line 633: I had to restore the `getMetadata` method because the reverted custom meter code still needs it.

However, care must be taken going forward that `getMetadata` is consistent with the `MetricFamilyDescriptor`.

We could reimplement `getMetadata` to call `familyDescriptor`. In terms of optimization this is a step in the opposite direction; however, if we consider custom meters rare, this could be an acceptable trade-off.

shakuzen:
It looks like the DCO check is failing. Would you please take a look?

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
m-burst: #7798

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
shakuzen: We'll take care of it in a follow-up refactor. Thanks for pointing it out.

shakuzen:
Thanks for the PR and quick updates. I've merged this, and I've been working on a larger refactor of the Prometheus code in #7802. This and #7487 added some allocation and work to meter creation in order to reduce overall allocation and CPU cycles. Meter creation is generally on the hot path for applications while scraping is done in a separate thread. With the mentioned refactor, I'm hoping to get back most of the performance on meter creation while improving the code structure and further reducing scrape time/allocation.

m-burst:
Thanks for reviewing and merging.

> This and https://github.com/micrometer-metrics/micrometer/pull/7487 added some allocation and work to meter creation in order to reduce overall allocation and CPU cycles. Meter creation is generally on the hot path for applications while scraping is done in a separate thread.

That is indeed true; however, as far as I understand, most added work only happens for the truly new meters. If an application has e.g. `Counter.builder(...).tags(...).register(meterRegistry).increment()` on a hot path, most of the time it will only need to lookup a meter by its id.

shakuzen:
> most added work only happens for the truly new meters

Yes, that's an important distinction I should have included in my statement. Still, the first call per meter is on the hot path for the application and we want to keep overhead minimal there so max latency does not degrade because of Micrometer.

Issue #7311, opened 2026-03-19
Add deprecation concept to meters


shakuzen:
Sometimes new metrics are added that make existing metrics obsolete. Unlike Java API, there is no way currently to mark those as deprecated. This is a challenge because removing a meter will, in general, not break an application, but it is still a breaking change if the output of that meter is relied upon by another system. For example, if the time-series data from the meter is used in a chart in a dashboard or used in an alert query. The breaking change in behavior may not be immediately apparent to the developers working on the application, which makes it harder to notice and correlate to the change that removed the meter.

### Proposal

Add a way to indicate that a meter is deprecated in the instrumentation API. Separately, there should be a way to configure the behavior with respect to deprecated meters.

PR #7458, opened 2026-05-02
Change process.uptime from TimeGauge to FunctionCounter

### LINKED ISSUE #7346: Change `process.uptime` to `FunctionCounter` in `UptimeMetrics`
As discussed in [this PR comment](https://github.com/micrometer-metrics/micrometer/pull/7297#discussion_r2942501228), `process.uptime` is currently implemented as a `TimeGauge` in `UptimeMetrics`. 

Semantically, it should be a `FunctionCounter` since it represents a monotonically increasing accumulated value.

However, for some registries, the change from a gauge to a counter is a breaking change. So we will need to consider how to do this in the least disruptive way, or wait for 2.0 to do this.

mgustimz:
## Summary

Changed `process.uptime` meter from `TimeGauge` to `FunctionCounter` in `UptimeMetrics`. Semantically, `process.uptime` is a monotonically increasing accumulated value, making `FunctionCounter` the more appropriate meter type — matching the discussion in PR #7297.

### UptimeMetrics.java
- Changed meter type from `TimeGauge` to `FunctionCounter`
- Added `.baseUnit("ms")` for consistency with the native unit returned by `RuntimeMXBean::getUptime`

### UptimeMetricsTest.java
- Updated test assertions from `.timeGauge()` to `.functionCounter()`
- Adjusted expected value from `1.337` (converted to seconds) to `1337.0` (raw milliseconds from mock)

## Breaking change note

For registries that treat gauges and counters differently, this is a breaking change for `process.uptime`. Users monitoring this metric may need to update their dashboards or alerting rules. The issue #7346 notes this was a consideration and it may be appropriate to wait for 2.0 for this change, but the semantic correctness makes it worth proposing now.

## Fixes

Fixes #7346

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/UptimeMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/UptimeMetrics.java
@@ -15,6 +15,7 @@
  */
 package io.micrometer.core.instrument.binder.system;
 
+import io.micrometer.core.instrument.FunctionCounter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.TimeGauge;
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
Thank you for the PR!

If this PR was (at least partly) created by some kind of AI tooling, could you please tell us what tool/tools you used so that we can keep the records of it?

I think this should wait for https://github.com/micrometer-metrics/micrometer/pull/7329 but also, as the original issue mentioned (https://github.com/micrometer-metrics/micrometer/issues/7346) as well as the description of this PR: this is a breaking change and I think we should not do this as-is. At least we should introduce some kind of an opt-in mechanism with probably deprecating the old one: https://github.com/micrometer-metrics/micrometer/issues/7311

mgustimz:
Thank you for the review!

Yeah, this PR was created with assistance from an AI tool (OpenClaw). I'll note that going forward.

Agreed on waiting for #7329 and the opt-in mechanism. The semantic correctness of FunctionCounter for a monotonically increasing value is clear, but you're right that a breaking change deserves a proper deprecation path rather than a silent swap. I'll leave this PR open as a reference and happy to implement the opt-in approach if #7311 moves forward.

PR #7479, opened 2026-05-12
Remove duplicate package-info classes from micrometer-test

### LINKED ISSUE #7067: Duplicate classes across modules causes Bazel build failure
I am experiencing some friction in getting my corporate build system to upgrade to latest micrometer because of duplicate classes issue.

Duplicates in micrometer-test-1.17.0-M1.jar

Conflicts with micrometer-core-1.17.0-M1.jar:

✗ io/micrometer/core/instrument/binder/cache/package-info.class
✗ io/micrometer/core/instrument/package-info.class
✗ io/micrometer/core/instrument/step/package-info.class
✗ io/micrometer/core/ipc/http/package-info.class

Conflicts with micrometer-commons-1.17.0-M1.jar:

 ✗ io/micrometer/common/util/internal/logging/package-info.class

Please ensure that the micrometer-test package does not contain overlapping classes.

officialasishkumar:
## Summary
Removes `package-info.java` sources from `micrometer-test` when the same package already belongs to `micrometer-core` or `micrometer-commons`, preventing duplicate `package-info.class` entries for Bazel consumers.

Keeps the `micrometer-test` package-info sources that are unique to that module.

Fixes #7067

## Testing
- `./gradlew :micrometer-test:jar --console=plain --max-workers=2`
- `jar tf micrometer-test/build/libs/micrometer-test-*.jar | grep 'package-info.class'`
- `git diff --cached --check`


shakuzen:
Thank you for the pull request, but I don't think this is the right solution to the problem, even if it might technically solve the Bazel issue. We need to fix the split package issue, but it will be a breaking change if users are depending on any public API in classes we need to move. So we'll need to take a multi-step approach to moving things. I'll close this pull request since this isn't the direction we want to go to fix this.

PR #7480, opened 2026-05-12
Clarify custom MeterTag value expression resolver docs

### LINKED ISSUE #6465: Missing class SpelValueExpressionResolver from documentation
When configuring Micrometer to use `@Metertag` annotation according to the [documentation](https://docs.micrometer.io/micrometer/reference/concepts/timers.html#_metertag_on_Method_parameters) micrometer dependencies missing class `SpelValueExpressionResolver ` even in spring-boot application.

1. I propose to add some information into documentation that will describe that `SpelValueExpressionResolver ` should be implemented independently atm or should be requested in spring-boot project for spring-boot applications.
2. In addition, I propose to create example of simple implementation in documentation that does not use `SPEL` for cases when the application does not need to calculate the values ​​of tags through `SPEL` at all.

**Rationale**
According to the [documentation](https://docs.micrometer.io/micrometer/reference/concepts/timers.html#_metertag_on_Method_parameters) code snippet does not work cause class `SpelValueExpressionResolver ` is missing.

```
ValueResolver valueResolver = parameter -> "Value from myCustomTagValueResolver [" + parameter + "]";
// Example of a ValueExpressionResolver that uses Spring Expression Language
ValueExpressionResolver valueExpressionResolver = new SpelValueExpressionResolver();
// Setting the handler on the aspect
timedAspect.setMeterTagAnnotationHandler(
        new MeterTagAnnotationHandler(aClass -> valueResolver, aClass -> valueExpressionResolver));
```

**Additional context**
Im facing this issue with:
- spring-boot 3.5.0
- micrometer 1.15.0

After some research I found that private class exists [SpelTagValueExpressionResolver](https://github.com/spring-projects/spring-boot/blob/19818fb844c66cd4fc0e2e45581335e22aba1ce2/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/MicrometerTracingAutoConfiguration.java#L136) for the spring-boot applications with a bit misspelling name:
- Spel**Tag**ValueExpressionResolver in the spring-boot
- SpelValueExpressionResolver in the micrometer documentation

Unfortunately, there is no bean for this class and single usage is [here](https://github.com/spring-projects/spring-boot/blob/19818fb844c66cd4fc0e2e45581335e22aba1ce2/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/MicrometerTracingAutoConfiguration.java#L117) as part of `SpanTagAnnotationHandler`. May be this class can be created as bean for Timed and Counted aspects as additional changes in spring-boot.

officialasishkumar:
## Summary
Renames the docs-only example resolver to `CustomValueExpressionResolver` and updates the `@MeterTag` docs to make clear that this resolver is application-provided, not a Micrometer class.

Updates the Timed, Counted, and Observation docs tests to use the renamed resolver.

Fixes #6465

## Testing
- `./gradlew :docs:test --tests io.micrometer.docs.metrics.TimedAspectTest --tests io.micrometer.docs.metrics.CountedAspectTest --tests io.micrometer.docs.observation.ObservationHandlerTests --console=plain --max-workers=2`
- `rg -n "SpelValueExpressionResolver|CustomValueExpressionResolver" docs/src/test/java docs/modules/ROOT/pages`
- `git diff --cached --check`

### DIFF
--- a/docs/modules/ROOT/pages/concepts/counters.adoc
+++ b/docs/modules/ROOT/pages/concepts/counters.adoc
@@ -97,6 +97,8 @@ include::{include-java}/metrics/CountedAspectTest.java[tags=resolvers,indent=0]
 include::{include-java}/metrics/CountedAspectTest.java[tags=meter_tag_annotation_handler,indent=0]
 -----
 
+NOTE: `CustomValueExpressionResolver` is an application-provided `ValueExpressionResolver` example. Micrometer does not provide this class. You need to provide an implementation that matches your application framework if you use `@MeterTag` expressions.
+
 Let's assume that we have the following interface.
 
 [source,java,subs=+attributes]
--- a/docs/modules/ROOT/pages/concepts/counters.adoc
+++ b/docs/modules/ROOT/pages/concepts/counters.adoc
@@ -97,7 +97,7 @@ include::{include-java}/metrics/CountedAspectTest.java[tags=resolvers,indent=0]
 include::{include-java}/metrics/CountedAspectTest.java[tags=meter_tag_annotation_handler,indent=0]
 -----
 
-NOTE: `CustomValueExpressionResolver` is an application-provided `ValueExpressionResolver` example. Micrometer does not provide this class. You need to provide an implementation that matches your application framework if you use `@MeterTag` expressions.
+NOTE: `CustomValueExpressionResolver` is a sample implementation of `ValueExpressionResolver` for documentation purposes. Micrometer does not provide this class. An implementation is required to use `@MeterTag` expressions.
 
 Let's assume that we have the following interface.
 
--- a/docs/modules/ROOT/pages/concepts/timers.adoc
+++ b/docs/modules/ROOT/pages/concepts/timers.adoc
@@ -125,6 +125,8 @@ include::{include-java}/metrics/TimedAspectTest.java[tags=resolvers,indent=0]
 include::{include-java}/metrics/TimedAspectTest.java[tags=meter_tag_annotation_handler,indent=0]
 -----
 
+NOTE: `CustomValueExpressionResolver` is an application-provided `ValueExpressionResolver` example. Micrometer does not provide this class. You need to provide an implementation that matches your application framework if you use `@MeterTag` expressions.
+
 Let's assume that we have the following interface.
 
 [source,java,subs=+attributes]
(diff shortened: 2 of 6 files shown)


docs/modules/ROOT/pages/concepts/counters.adoc
```diff
@@ -97,6 +97,8 @@ include::{include-java}/metrics/CountedAspectTest.java[tags=resolvers,indent=0]
 include::{include-java}/metrics/CountedAspectTest.java[tags=meter_tag_annotation_handler,indent=0]
 -----
 
+NOTE: `CustomValueExpressionResolver` is an application-provided `ValueExpressionResolver` example. Micrometer does not provide this class. You need to provide an implementation that matches your application framework if you use `@MeterTag` expressions.
```
shakuzen: I would reword this a bit. While Micrometer doesn't provide an implementation, the end-user doesn't _necessarily_ need to provide one. The framework they're using or some other third-party library may provide one.

```suggestion
NOTE: `CustomValueExpressionResolver` is a sample implementation of `ValueExpressionResolver` for documentation purposes. Micrometer does not provide this class. An implementation is required to use `@MeterTag` expressions.
```

shakuzen (commented):
Thank you for the pull request. I agree the documentation is better with it made more clear that the implementation referenced isn't available for use from Micrometer. I left some feedback about wording that applies to both places in the documentation where the note was added.

docs/modules/ROOT/pages/concepts/counters.adoc
officialasishkumar: Updated the note so it allows resolvers provided by frameworks or third-party libraries in a757bac.

shakuzen:
Thanks for updating things. The [DCO check](https://github.com/micrometer-metrics/micrometer/pull/7480/checks?check_run_id=76899900934) is failing. Please check it.

officialasishkumar:
Fixed the DCO signoff in 8393ab2; the DCO check is passing.

docs/modules/ROOT/pages/concepts/timers.adoc
```diff
@@ -125,6 +125,8 @@ include::{include-java}/metrics/TimedAspectTest.java[tags=resolvers,indent=0]
 include::{include-java}/metrics/TimedAspectTest.java[tags=meter_tag_annotation_handler,indent=0]
 -----
 
+NOTE: `CustomValueExpressionResolver` is an application-provided `ValueExpressionResolver` example. Micrometer does not provide this class. You need to provide an implementation that matches your application framework if you use `@MeterTag` expressions.
```
shakuzen, line 128: Please update this note to match the one in counters.adoc

shakuzen (changes requested):
Could you also rebase this on `1.15.x` so we can include this documentation update in the maintenance line documentation and merge forward from there?

PR #7481, opened 2026-05-12
Mark nullability on Meter.match

### LINKED ISSUE #6408: Revise @NullUnmarked usage in io.micrometer.core.instrument.Meter
See https://github.com/micrometer-metrics/micrometer/pull/6385#discussion_r2141192698

nikita-kibitkin:
Removes the `@NullUnmarked` workaround on `Meter.match` and expresses the nullability contract with JSpecify.

`match` returns whatever the selected visitor returns, so the result type has to allow `null` when a visitor returns `null`. Switched the type parameter to `<T extends @Nullable Object>`; the visitor functions themselves stay non-null.

Updated test call sites that passed `null` for unused visitors to pass `x -> null` instead, since `Function` params are no longer nullable. Added a test for the `match`-returns-`null` case.

Closes gh-6408


shakuzen (approved):
Thanks for the PR. The changes look good to me.

jonatan-ivanov (approved):

jonatan-ivanov:
@nikita-kibitkin Thank you for fixing this!

PR #7487, opened 2026-05-13
Fail collision validation at registration instead of at scrape with typed Prometheus family descriptors

zeitlinger:
## Summary

This is a cleaner follow-up to the Prometheus metadata compatibility work for Micrometer.

Instead of reconstructing `MetricMetadata` from Micrometer-side suffix logic, this change switches the Prometheus registry integration to use the new typed registration descriptor API from `client_java`.

Depends on:
- prometheus/client_java#2114

Related context:
- micrometer-metrics/micrometer#7460

## What changed

- replace Micrometer's stored registration metadata shape with `MetricFamilyDescriptor`
- remove Micrometer-side `MetricMetadata` reconstruction for registration metadata
- derive registration metadata directly from the typed descriptor API
- keep scrape-time snapshot creation on typed snapshot builders
- add/update registration collision coverage in Prometheus registry tests

## Why

Micrometer needs registration-time family metadata for collision validation, but the old shape forced Micrometer to duplicate Prometheus naming derivation details.

With `client_java` exposing typed family descriptors, Micrometer can now treat Prometheus metadata as an upstream concern instead of rebuilding it locally.

## Validation

Tested locally against a locally installed `prometheus/client_java` `1.6.2-SNAPSHOT` containing the descriptor API from prometheus/client_java#2114:

```bash
./gradlew --no-build-cache \
  -I /tmp/micrometer-prom-local.init.gradle \
  :micrometer-registry-prometheus:test \
  --tests io.micrometer.prometheusmetrics.MicrometerCollectorTest \
  --tests io.micrometer.prometheusmetrics.PrometheusMeterRegistryTest
```

## Notes

- This draft PR is expected to need `prometheus/client_java#2114` first.
- Until that API is released, this branch is primarily for review of the end-state design.


shakuzen:
@zeitlinger thank you for the PR. We'll look to review and merge this for the Micrometer 1.18.0 release. In the meantime, dependabot sent a PR to upgrade the Prometheus Java client version to 1.7.0, which was merged automatically since the build passed: https://github.com/micrometer-metrics/micrometer/pull/7572. That would be included in the upcoming Micrometer 1.17.0 GA release. Is there any concern about making that upgrade on its own without this PR?

zeitlinger:
Thanks for the heads up. Yes, safe to ship 1.7.0 in Micrometer 1.17.0 without #7487 — the two are independent. (And worth noting: Micrometer skipped the 1.6 line because of a bug there; 1.7.0 is the first clean upgrade target.) We've also added tests on the `client_java` side that pin the naming/suffix behavior Micrometer depends on, so future regressions of the kind that caused the 1.6 skip should be caught upstream — with or without #7487 merged.

**What #7487 adds on top (still worth merging for 1.18.0):**
- **Correctness:** Today Micrometer rebuilds `MetricMetadata` by replicating Prometheus's suffix/naming rules on its side. Every time those rules evolve upstream, the two implementations can silently drift. Consuming the typed descriptor directly makes Prometheus the single source of truth and eliminates that class of bug.
- **Cleaner code:** Removes the Micrometer-side reconstruction logic and its tests. Naming concerns stay where they belong — in `client_java`.
- **Perf:** Minor — avoids re-deriving family metadata the registry already knows. Not the motivation, but a small win.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -72,6 +87,42 @@ Meter.Id getOriginalId() {
         return originalMeterId;
     }
 
+    @Override
+    public List<String> getPrometheusNames() {
```
shakuzen: Aren't the newly overridden methods all deprecated? Should we override `getMetricFamilyDescriptors` instead? I have some changes locally that simplify the code quite a bit when using that instead. We can get rid of the `RegisteredFamily` type which is just a wrapper for `MetricFamilyDescriptor`.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -572,20 +621,86 @@ private void onMeterRemoved(Meter meter) {
...
+            .name(name)
+            .help(help)
+            .gaugeHistogram(gaugeHistogram);
+        family.dataPointSnapshots.forEach(builder::dataPoint);
+        return builder.build();
+    }
+
+    private String counterMetricName(String prometheusName, String originalName) {
+        return normalizeBaseName(prometheusName, originalName, ".bucket", "_bucket", ".created", "_created", ".info",
+                "_info", ".total", "_total");
```
shakuzen: These normalization methods seem redundant with the default `PrometheusNamingConvention` implementation which delegates to the `PrometheusNaming` methods from the client. Why are they needed?

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -572,20 +621,86 @@ private void onMeterRemoved(Meter meter) {
...
+        InfoSnapshot.Builder builder = InfoSnapshot.builder().name(name).help(help);
+        family.dataPointSnapshots.forEach(builder::dataPoint);
+        return builder.build();
+    }
+
+    private SummarySnapshot summarySnapshot(MicrometerCollector.Family<SummaryDataPointSnapshot> family, String name,
+            String help) {
+        SummarySnapshot.Builder builder = SummarySnapshot.builder().name(name).help(help);
+        family.dataPointSnapshots.forEach(builder::dataPoint);
+        return builder.build();
```
shakuzen: Before we were directly calling the constructor, e.g. `new SummarySnapshot(family.metadata, family.dataPointSnapshots)`. That seems better than this way via a builder. Is there some advantage to this way of doing things that I'm missing?

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
```diff
@@ -90,6 +90,19 @@ void differentMicrometerNameSamePrometheusNameFailsToRegister() {
...
+    void functionTimerCollisionUsesPrometheusRegistrationMetadata() {
+        registry.throwExceptionOnRegistrationFailure();
+
+        registry.more().timer("test", Tags.empty(), this, o -> 1, o -> 2, TimeUnit.SECONDS);
+
+        assertThatThrownBy(
+                () -> Gauge.builder("test.seconds", new AtomicInteger(42), AtomicInteger::get).register(registry))
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessageContaining("same Prometheus name (test_seconds)")
+            .hasMessageContaining("would fail with an exception on scrape");
```
shakuzen, line 123: Am I understanding correctly that this registration-time validation is the real benefit users are gaining from us adopting the typed family descriptors API? Whereas before registration would succeed and then it would fail at scrape time?

shakuzen (commented):
@zeitlinger thank you for the work on this. Sorry for the long delay in getting it reviewed. I have some changes locally that I think make sense, but I've asked for clarification on a few things before I change anything.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
zeitlinger: Yes, I agree. Overriding `getMetricFamilyDescriptors` is the right target now, and removing `RegisteredFamily` sounds like the cleaner implementation. Please feel free to push your local simplification to this branch (maintainer edits are enabled).

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
zeitlinger: Agreed. These were carried over from the earlier metadata-reconstruction approach, but with the default `PrometheusNamingConvention` delegating to the client they are redundant and should be removed.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
zeitlinger: No advantage here—the direct snapshot constructors are clearer. The builder conversion was incidental to the refactor and should be reverted.

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
zeitlinger: Yes, exactly. The immediate user-visible benefit is moving incompatible-family validation to registration time instead of failing later during scrape. The other benefit is consuming the client's typed descriptor as the source of truth, which avoids Micrometer duplicating metadata/naming logic.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
zeitlinger: Implemented in 00c7e1472: `MicrometerCollector` now overrides `getMetricFamilyDescriptors()` directly and the `RegisteredFamily` wrapper is gone.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
zeitlinger: Implemented in 00c7e1472: removed the redundant normalization helpers and now use the naming convention result directly.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
zeitlinger: Implemented in 00c7e1472: restored direct snapshot constructor calls.

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
zeitlinger: The registration-time validation test remains in place, and 00c7e1472 simplifies the implementation around the typed descriptor API.

zeitlinger:
Implemented the suggested simplification in `00c7e1472`:

- override `getMetricFamilyDescriptors()` directly and remove the `RegisteredFamily` wrapper
- remove the redundant metric-name normalization helpers and rely on `PrometheusNamingConvention`
- use the direct snapshot constructors instead of builders
- update the collector test to assert the typed descriptor directly

Validation:

- `./gradlew format --no-daemon`
- `./gradlew :micrometer-registry-prometheus:checkFormat :micrometer-registry-prometheus:test --no-daemon`

The review threads have been addressed and resolved. The current JDK 17 CircleCI failure is unrelated: downloading the Gradle 9.5.1 distribution timed out after 10 seconds; the other jobs are still running.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -394,66 +406,100 @@ protected Meter newMeter(Meter.Id id, Meter.Type type, Iterable<Measurement> mea
...
             List<String> tagKeys = tagKeys(id);
+            List<String> statKeys = new ArrayList<>(tagKeys);
+            statKeys.add("statistic");
+            List<MetricFamilyDescriptor> registrationDescriptors = new ArrayList<>();
+            String name = getConventionName(id);
+            Set<String> descriptorNames = new HashSet<>();
+            for (Measurement measurement : measurements) {
+                MetricFamilyDescriptor descriptor = customMeterDescriptor(id, name, measurement.getStatistic(),
+                        statKeys);
+                // de-duplicate names
```
shakuzen, line 417: This fixes a test failure caught by the docker tests. I added a unit test to catch it as well - see `customMeterWithAllStatistics` in `PrometheusMeterRegistryTest`

shakuzen (approved):
I polished the changes a bit and added some test. I think this is good to go now. Let me know if you think anything looks off after the polishing.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -538,12 +586,23 @@ private void addDistributionStatisticSamples(Meter.Id id, MicrometerCollector co
...
 
             return families.build();
-        });
+        }, familyDescriptor(primaryType, getConventionName(id), tagKeys, id.getDescription()),
+                familyDescriptor(MetricType.GAUGE, getConventionName(id) + "_max", tagKeys, id.getDescription()));
+    }
+
+    private MetricFamilyDescriptor familyDescriptor(MetricType metricType, String name, Collection<String> labelNames,
+            @Nullable String description) {
+        return MetricFamilyDescriptor.of(metricType, name).help(helpText(description)).labelNames(labelNames).build();
```
shakuzen, line 601: Does setting the help text matter for the purposes we use MetricFamilyDescriptor? It's not a big deal, but it seems like we could avoid the builder if we don't care about the help text. This is something we can follow-up with post merge, but I wanted to ask the question while I noticed it.

PR #7502, opened 2026-05-19
Add config option to compress datadog payload

### LINKED ISSUE #2297: Add config option to compress Datadog payload
Compression cannot be enabled with `DatadogConfig` right now, but we do have the `compress` method on our abstract `Request.Builder` for `HttpSender`. Enabling compression by default seems reasonable and would help avoid the issue of going over the max payload size as mentioned in #1500.

_Originally posted by @shakuzen in https://github.com/micrometer-metrics/micrometer/issue_comments/705314786_

antonio-pedro99:
This PR fixes https://github.com/micrometer-metrics/micrometer/issues/2297.

### DIFF
--- a/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
+++ b/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogConfig.java
@@ -70,6 +70,13 @@ public interface DatadogConfig extends StepRegistryConfig {
         return getBoolean(this, "descriptions").orElse(true);
     }
 
+    /**
+     * @return {@code true} if payload should be compressed.
+     */
+    default boolean compress() {
+        return getBoolean(this, "compress").orElse(true);
+    }
+
     @Override
     default Validated<?> validate() {
         return checkAll(this, c -> StepRegistryConfig.validate(c), checkRequired("apiKey", DatadogConfig::apiKey),
--- a/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
+++ b/implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
@@ -142,6 +142,7 @@ public class DatadogMeterRegistry extends StepMeterRegistry {
 
                 httpClient.post(datadogEndpoint)
                     .withJsonContent(body)
+                    .compressWhen(config::compress)
                     .send()
                     .onSuccess(response -> logger.debug("successfully sent {} metrics to datadog", batch.size()))
                     .onError(response -> logger.error("failed to send metrics to datadog: {}", response.body()));
--- a/implementations/micrometer-registry-datadog/src/test/java/io/micrometer/datadog/DatadogConfigTest.java
+++ b/implementations/micrometer-registry-datadog/src/test/java/io/micrometer/datadog/DatadogConfigTest.java
@@ -46,4 +46,16 @@ class DatadogConfigTest {
         assertThat(config.hostTag()).isEqualTo("instance");
     }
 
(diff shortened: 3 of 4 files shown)


shakuzen (commented):
Don't we need zlib compression? I also wonder how hard it would be to have some kind of integration test that would catch such issues, since no test is failing even though I believe it is using the wrong compression algorithm now.

antonio-pedro99:
> I also wonder how hard it would be to have some kind of integration test that would catch such issues,

I can add integration tests. Regarding the compression type, I do not think we need only zlib. I was playing with the Datadog metrics API v1 on Postman, and I saw that they might support other metrics like `deflate` and `identity`. I have tested this implementation with the default gzip compression, and it worked just fine. 

Right now, the compression is enabled by default, so all Datadog current integration tests are compressing the payload. But let me add more specific test maybe

antonio-pedro99:
> > I also wonder how hard it would be to have some kind of integration test that would catch such issues,
> 
> I can add integration tests. Regarding the compression type, I do not think we need only zlib. I was playing with the Datadog metrics API v1 on Postman, and I saw that they might support other metrics like `deflate` and `identity`. I have tested this implementation with the default gzip compression, and it worked just fine.
> 
> Right now, the compression is enabled by default, so all Datadog current integration tests are compressing the payload. But let me add more specific test maybe

I spent some time again playing with Datadog v1 API, and found out the following supported format:
- gzip is accepted, when produced by `new GZIPOutputStream(bos)`.
- deflate(zlib wrapped) is accept in the following format: `zlib header + raw DEFLATE stream + Adler-32 checksum`. This is produced by `new Deflater(Deflater.BEST_COMPRESSION, false)`.
- identity(is same as, do not use any compression so it would work if payload size does not exceed), this one really does not interest us.

Raw deflate is not supported by Datadog API. If set the Deflater to `new Deflater(Deflater.BEST_COMPRESSION, true)`, we will send raw deflate bytes, and the API will throw `Unable to decompress compressed payload`.

That being said, two paths we can take:
- Continuing using gzip only
- Add support for other accepted `Content-Enconding` values.

shakuzen:
Thank you for the investigation. That's good info. Is there a link to some official documentation from Datadog regarding supported compression that we could reference?

On integration tests (other than the type of tests we currently have), we can defer that to a separate change. I was thinking about tests that send data to a real Datadog endpoint, but we need to deal with API key management in order to do that.

shakuzen:
> * Continuing using gzip only

That sounds easiest to me. Unless there's a compelling reason to complicate things, why not go with that? I suppose it would make the configuration more difficult to change to support different algorithms later if we use a boolean now, though. Is there a reason users would care much about the compression algorithm used?

antonio-pedro99:
> Is there a link to some official documentation from Datadog regarding supported compression that we could reference?

No, I could not find any reference. That's why I did the investigation.

antonio-pedro99:
> but we need to deal with API key management in order to do that.

This is adding more complexity, plus I am not sure if these API keys have an expiration time(I bet they have), so if they expire after a certain period, it will break the tests.

shakuzen:
I've marked this PR as blocked for now, since we need to branch `1.17.x` after the 1.17.0 GA release (currently scheduled for next week) before we can start merging things for 1.18.0. We'll come back to this once that's done. Thanks for the PR and quick responses so far.

antonio-pedro99:
> I've marked this PR as blocked for now, since we need to branch `1.17.x` after the 1.17.0 GA release (currently scheduled for next week) before we can start merging things for 1.18.0. We'll come back to this once that's done. Thanks for the PR and quick responses so far.

Sure. I will wait for your reviews later after 1.18.0 is merged.

shakuzen (approved):
Thank you for the pull request and the patience on review. It looks good to me.

PR #7508, opened 2026-05-20
Resolve `DuplicateLabelsException` in `MicrometerCollector`

### LINKED ISSUE #6851: Registering/Updating a MultiGauge while a Prometheus scrape occurs can lead to DuplicateLabelsException
**Describe the bug**

Registering/Updating a MultiGauge while a Prometheus scrape requests occur can lead to occasional 
io.prometheus.metrics.model.snapshots.DuplicateLabelsException: Duplicate labels for metric exception

**Environment**

Dependencies:
• Spring Boot BOM: 3.5.7
• io.micrometer:micrometer-core:1.15.5
• io.micrometer:micrometer-registry-prometheus:1.15.5
• io.prometheus:prometheus-metrics-bom: 1.4.2

**To Reproduce**

A small Test with a PrometheusMeterRegistry is enough.

1. Setup 
io.micrometer.prometheusmetrics.PrometheusMeterRegistry(io.micrometer.prometheusmetrics.PrometheusConfig.DEFAULT)                                                                                                           
2. Register MultiGauges: 
   MultiGauge.builder("my_metric")                                                                            
               .register(prometheusMeterRegistry)                                                             
                                                                                                              
  With multiple rows:
```java
   MultiGauge.Row.of(                                                                                         
                   io.micrometer.core.instrument.Tags.of(                                                     
                       "Tag1", "abc",                                                                         
                       "poll_count", "a counter for example: 1", // <-- this increments for example           
                       "SomeOtherTags", "cde",                                                                
                   ),
   MultiGauge.Row.of(                                                                                         
                   io.micrometer.core.instrument.Tags.of(                                                     
                       "Tag1", "abc",                                                                         
                       "poll_count", "2",         
                       "SomeOtherTags", "cde",                                                                
                   ),
...
   MultiGauge.Row.of(                                                                                         
                   io.micrometer.core.instrument.Tags.of(                                                     
                       "Tag1", "abc",                                                                         
                       "poll_count", "50"           
                       "SomeOtherTags", "cde",                                                                
                   ),                                                                                               
                                                                                                              
```

3. Start a new thread that scrapes prometheusMeterRegistry.scrape() frequently to increase chance of exception
4. While the scrape happens, register/update the MultiGauge rows:

   multiGauge.register(rows, true) // <-- Override true                                                       
                                                                                                              

5. Now occasionally the scrape thread throws io.prometheus.metrics.model.snapshots.DuplicateLabelsException, 
even though there should be no duplicate labels because the "poll_count" tag for example changes or the metric override should handle this properly

**Expected behavior**

multiGauge.register(rows, true) should not cause a 
io.prometheus.metrics.model.snapshots.DuplicateLabelsException when prometheusMeterRegistry.scrape() occurs

schiemon:
Fixes #6851

`MultiGauge.register(rows, true)` removes and re-adds row gauges:

  https://github.com/micrometer-metrics/micrometer/blob/32a3fce5405085b70e0704a922d9a44b5422f548/micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java#L61-L91

While this is happening, a concurrent Prometheus scrape could iterate `MicrometerCollector.children` through a weakly consistent `ConcurrentHashMap` iterator and observe both an old child and its replacement in the same scrape:

  https://github.com/micrometer-metrics/micrometer/blob/32a3fce5405085b70e0704a922d9a44b5422f548/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java#L75-L91
  
That produced duplicate samples with the same Prometheus label set (`families`), causing the Prometheus client to throw `DuplicateLabelsException`.

The fix catches the DuplicateLabelsException and removes all duplicates but the first match for labels.


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -75,9 +75,17 @@ Meter.Id getOriginalId() {
     @Override
     public MetricSnapshots collect() {
         Map<String, Family> families = new HashMap<>();
-
-        for (Child child : children.values()) {
-            child.samples(conventionName)
+        Set<Meter.Id> seen = new HashSet<>(children.size());
+
+        for (Map.Entry<Meter.Id, Child> child : children.entrySet()) {
+            // `children` is a `ConcurrentHashMap` and its iterator is weakly-consistent.
```
schiemon: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/package-summary.html#Weakly

schiemon:
I thought about just adding an `Object`-lock protecting `children` but just filtering out duplicates seems to be least invasive from a code and performance perspective.

shakuzen:
Thanks for the analysis and PR. We'll want a JMH benchmark to quantify the performance impact of this, including measuring allocations with the GCProfiler.

schiemon:
We see a ~5% increase in allocation rate inside `collect()` and a similar increase in CPU usage. This is consistent with the differential async-profiler flamegraphs for the `children=100K` case.

```
Benchmark                                                Params    Variant            Score           Error  Units
------------------------------------------------------------------------------------------------------------------
MicrometerCollectorBenchmark.collect                     1000      baseline      764776.756       81479.363  ns/op
                                                                   fix           724178.292       21882.560  ns/op
                                                                   Δ abs         -40598.464      -59596.803  ns/op
                                                                   Δ %               -5.31%         -73.14%

MicrometerCollectorBenchmark.collect:gc.alloc.rate       1000      baseline        1055.006         104.735  MB/sec
                                                                   fix             1158.464          34.090  MB/sec
                                                                   Δ abs            103.458         -70.645  MB/sec
                                                                   Δ %               +9.81%         -67.45%

MicrometerCollectorBenchmark.collect:gc.alloc.rate.norm  1000      baseline      838872.532           0.056  B/op
                                                                   fix           879128.503           0.015  B/op
                                                                   Δ abs          40255.971          -0.041  B/op
                                                                   Δ %               +4.80%         -73.21%

MicrometerCollectorBenchmark.collect:gc.count            1000      baseline          32.000                  counts
                                                                   fix               36.000                  counts
                                                                   Δ abs              4.000                  counts
                                                                   Δ %              +12.50%                

MicrometerCollectorBenchmark.collect:gc.time             1000      baseline         183.000                  ms
                                                                   fix              184.000                  ms
                                                                   Δ abs              1.000                  ms
                                                                   Δ %               +0.55%                

MicrometerCollectorBenchmark.collect                     10000     baseline     7468692.463      310471.146  ns/op
                                                                   fix          7758603.854      309939.626  ns/op
                                                                   Δ abs         289911.391        -531.520  ns/op
                                                                   Δ %               +3.88%          -0.17%

MicrometerCollectorBenchmark.collect:gc.alloc.rate       10000     baseline         974.137          37.807  MB/sec
                                                                   fix              985.164          38.893  MB/sec
                                                                   Δ abs             11.027           1.086  MB/sec
                                                                   Δ %               +1.13%          +2.87%

MicrometerCollectorBenchmark.collect:gc.alloc.rate.norm  10000     baseline     7619653.187           0.224  B/op
                                                                   fix          8005253.403           0.214  B/op
                                                                   Δ abs         385600.216          -0.010  B/op
                                                                   Δ %               +5.06%          -4.46%

MicrometerCollectorBenchmark.collect:gc.count            10000     baseline          30.000                  counts
                                                                   fix               30.000                  counts
                                                                   Δ abs              0.000                  counts
                                                                   Δ %               +0.00%                

MicrometerCollectorBenchmark.collect:gc.time             10000     baseline         148.000                  ms
                                                                   fix              176.000                  ms
                                                                   Δ abs             28.000                  ms
                                                                   Δ %              +18.92%                

MicrometerCollectorBenchmark.collect                     100000    baseline    95719241.624     1648868.077  ns/op
                                                                   fix         97474594.446     2577931.047  ns/op
                                                                   Δ abs        1755352.822      929062.970  ns/op
                                                                   Δ %               +1.83%         +56.35%

MicrometerCollectorBenchmark.collect:gc.alloc.rate       100000    baseline         778.681          13.281  MB/sec
                                                                   fix              806.464          20.666  MB/sec
                                                                   Δ abs             27.783           7.385  MB/sec
                                                                   Δ %               +3.57%         +55.61%

MicrometerCollectorBenchmark.collect:gc.alloc.rate.norm  100000    baseline    78143186.213           1.251  B/op
                                                                   fix         82391827.301           1.810  B/op
                                                                   Δ abs        4248641.088           0.559  B/op
                                                                   Δ %               +5.44%         +44.68%

MicrometerCollectorBenchmark.collect:gc.count            100000    baseline          24.000                  counts
                                                                   fix               25.000                  counts
                                                                   Δ abs              1.000                  counts
                                                                   Δ %               +4.17%                

MicrometerCollectorBenchmark.collect:gc.time             100000    baseline         194.000                  ms
                                                                   fix              247.000                  ms
                                                                   Δ abs             53.000                  ms
                                                                   Δ %              +27.32%                
```

Allocations (`--total`, bytes):

alloc_diff

[run-all-test-async-alloc-heap-pressure-flamegraph-diff.html](https://github.com/user-attachments/files/28470498/run-all-test-async-alloc-heap-pressure-flamegraph-diff.html)

CPU:

cpu_diff

[run-all-test-async-cpu-flamegraph-diff.html](https://github.com/user-attachments/files/28470540/run-all-test-async-cpu-flamegraph-diff.html)

To summarize:

1. Even in the worst case with `100K` children, we observed roughly a 5% increase in both CPU time and allocation rate.
2. The additional allocations are short-lived objects that remain confined to young-generation collections.
3. The affected code path is scraping, which is [less performance-critical](https://github.com/micrometer-metrics/micrometer/wiki/Performance:-criticality-of-code-paths) than for example meter recording paths.

@shakuzen wdyt?

schiemon:
@shakuzen friendly bump

shakuzen:
Sorry for the delay in getting back to reviewing this. I just opened https://github.com/micrometer-metrics/micrometer/pull/7851 which I think provides a more fundamental fix for the MultiGauge re-register problem, but probably those changes should not be applied to maintenance branches. I'll take a closer look at this PR tomorrow as something to fix the reported issue in maintenance branches. The PR will need to be retargeted to `1.16.x` because `1.15.x` is out of OSS support now.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
shakuzen: I was chatting with an LLM exploring other implementations to solve this problem, and it came up with a solution around this code that does what you've done but it makes it conditional to when we have the duplicate problem. I like this solution because it limits the added cost to when the problem occurs, which should be rare.

```java
    MetricSnapshot toMetricSnapshot() {
        try {
            return metricSnapshotFactory.apply(this);
        }
        catch (DuplicateLabelsException ex) {
            Set<Labels> seenLabels = new HashSet<>();
            dataPointSnapshots.removeIf(dataPoint -> !seenLabels.add(dataPoint.getLabels()));
            return metricSnapshotFactory.apply(this);
        }
    }
```

What do you think?

shakuzen: I've updated the pull request with this implementation.

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/prometheus/PrometheusRegistryBenchmark.java
shakuzen, line 1: I've backported this benchmark from `main`.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -129,7 +131,14 @@ Family<T> addSamples(Collection<T> dataPointSnapshots) {
...
 
         MetricSnapshot toMetricSnapshot() {
-            return metricSnapshotFactory.apply(this);
+            try {
+                return metricSnapshotFactory.apply(this);
+            }
+            catch (DuplicateLabelsException ex) {
+                Set<Labels> seenLabels = new HashSet<>();
+                dataPointSnapshots.removeIf(dataPoint -> !seenLabels.add(dataPoint.getLabels()));
+                return metricSnapshotFactory.apply(this);
```
shakuzen, line 148: I wonder if we should log something here. If the cause is the MultiGauge issue we know about, I'd rather not log anything or log it at trace level, but if it is some other cause, it would be good to have a higher level log about it. I don't know if we can differentiate, though. It seems like it would be bad to have another issue producing duplicates and not know about it. Granted, I think the plan should be to remove this try/catch behavior on `main` because the MultiGauge problem is being fixed another way there, so this is really just for maintenance branches. Thoughts?

jonatan-ivanov: Since `MultiGauge` "just" registers `Gauge`s, I'm not sure there is a relatively easy way to detect this scenario (unless the `MultiGauge` somehow signals it on the `Gauge`, e.g. in the description).

I think since Micrometer should prevent this happening but if it happens, the behavior before this change was throwing out an exception during scrape, I think we should log out the exception here so we won't hide the fact that something bad happened. As for the `MultiGauge` scenario, We can mention that if this was a MultiGauge, this is probably ok and upgrading Micrometer to 1.18.0 should make this message disappear.

If we log out, the exception, it looks like this:
```
io.prometheus.metrics.model.snapshots.DuplicateLabelsException: Duplicate labels for metric "test": {k1="1"}
```
so users can find out if there is a `MultiGauge` that's name is "test" and has a tag: "k1=1".

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -129,7 +131,14 @@ Family<T> addSamples(Collection<T> dataPointSnapshots) {
         }
 
         MetricSnapshot toMetricSnapshot() {
-            return metricSnapshotFactory.apply(this);
+            try {
+                return metricSnapshotFactory.apply(this);
+            }
+            catch (DuplicateLabelsException ex) {
+                Set<Labels> seenLabels = new HashSet<>();
```
jonatan-ivanov, line 146: Since this is weird enough to do, can we leave a comment here that mentions the issue this fixes, e.g.:
```suggestion
                // Fix for https://github.com/micrometer-metrics/micrometer/issues/6851
                Set<Labels> seenLabels = new HashSet<>();
```

benchmarks/benchmarks-core/src/jmh/java/io/micrometer/benchmark/prometheus/PrometheusRegistryBenchmark.java
shakuzen: Perhaps worth noting that there is no MultiGauge use in the benchmark, but I think that's okay because we're mostly concerned with checking performance regression on the most typical usage.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
shakuzen: I'm not sure about the exact log message and the level to log at, though. It's kind of ugly to log something like "if this is a MultiGauge it is a known issue" but I'm not sure what else to do. Should we use WarnThenDebugLogger? I wouldn't want constant noise for users affected by the MultiGauge issue, but the problem is if the MultiGauge issue and another issue happen in the same app but the MultiGauge issue happens first, it masks the other issue (assuming debug level isn't enabled by default).

shakuzen: I've updated the PR with a debug log. I think that's probably fine unless we have some specific scenario where we think it'd be better to be louder about this. What do you think?

jonatan-ivanov (approved):

PR #7509, opened 2026-05-20
fix exception message with `MeterNotFoundException` in concurrent case

### LINKED ISSUE #2307: Race condition in registry searching and MeterNotFoundException creation may result in a failure that looks as if it should have succeeded
I think there's a race condition between meter registration and the evaluation of a search and creation of a  `MeterNotFoundException`. If a thread registers a meter that meets the search's criteria after the search has been performed but before the `MeterNotFoundException` is created, the exception will report a failure that looks as if it should have succeeded:

```
io.micrometer.core.instrument.search.MeterNotFoundException: Unable to find a meter that matches all the requirements at once. Here's what was found: |  
-- | --
  | OK: A meter with name 'http.server.requests' was found. |  
  | OK: Meters with type timer were found.
```

schiemon:
fixes #2307 

Use the same meter snapshot for failed RequiredSearch lookups and MeterNotFoundException diagnostics so concurrent registrations cannot make the failure message report meters that were not visible to the failed search.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/search/MeterNotFoundException.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/search/MeterNotFoundException.java
@@ -42,8 +42,9 @@ public class MeterNotFoundException extends RuntimeException {
                 + (tagDetail == null ? "" : "\n   " + tagDetail.stream().collect(joining("\n   "))));
     }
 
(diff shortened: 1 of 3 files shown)


shakuzen (commented):
Thanks for the pull request. I'll take a closer look once we've branched `1.17.x` from `main` but so far this looks good.

schiemon:
It seem like `1.17.x` was branched off, friend reminder of the existence of this PR :)

shakuzen (approved):
Looks good to me. Thanks!

PR #7531, opened 2026-05-26
Extract JFR recording stream configuration into `JfrMeterBinder`

schiemon:
This PR introduces `JfrMeterBinder` to unify `RecordingStream` configuration and JFR event handler registration for JFR-backed `MeterBinder`s.

`JfrMeterBinder` was put into an newly added `micrometer-java17` module as `RecodingStream` was introduced in Java 14 and Java 17 is the next-in-line LTS version.

This is also pre-work for #7384 in which we are going to introduce `JvmSafepointMetrics`, after `VirtualThreadMetrics` the second `JfrMeterBinder` in Micrometer.


micrometer-java17/src/main/java/io/micrometer/java17/instrument/binder/jdk/JfrMeterBinder.java
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * The same as {@link MeterBinder#bindTo(MeterRegistry)}, but with an additional
+     * {@code eventRegistrar} parameter that can be used to register handlers for JFR
+     * events.
+     * @param registry the registry to bind to
+     * @param eventHandlerRegistrar the registrar to use for registering handlers for JFR
+     * events
+     */
+    protected abstract void bindTo(MeterRegistry registry, EventHandlerRegistrar eventHandlerRegistrar);
```
schiemon, line 72: I am intentionally not simply handing over the `RecordingStream` or `EventStream` as I don't want to allow the implementation to `start` or `close` the stream.

jonatan-ivanov:
@shakuzen and I need to discuss if we want to generalize the JFR binders in the next minor release. I think we should but I'm in favor of a simpler approach, like the code you copied in this PR: https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:jfr-binder

schiemon:
> @shakuzen and I need to discuss if we want to generalize the JFR binders in the next minor release.

Alright, please notify me with your decision when you are ready 👍 

> but I'm in favor of a simpler approach, like the code you copied in this PR: https://github.com/micrometer-metrics/micrometer/compare/main...jonatan-ivanov:micrometer:jfr-binder

Personally, I also like that it is simple but dislike that it leaks the `RecordingStream` to the implementation because then, in principle, the implementation could start or stop and reconfigure the stream. IMHO the `JfrBinder` should completely own the stream. But again, please let me know how you guys decide so I can move on with this one and https://github.com/micrometer-metrics/micrometer/pull/7384

schiemon:
@jonatan-ivanov friendly bump :)

Issue #6507, opened 2025-07-03
Refinement/Clarification of LongTaskTimer when using Prometheus


jjank:
### Please describe the feature request 

Following a question in Slack with @shakuzen https://micrometer-metrics.slack.com/archives/C662HUJC9/p1751460825070189

The Migration to Prometheus 1.x client libs in micrometer 1.13. changed the behaviour for `LongTaskTimer`. 

Specifically in said Slack thread it was concluded that:
* The [migration guide](https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide) could be more precise in outlining two distinct changes
    * Histogram -> GaugeHistogram for LongTaskTimers with histogram buckets.
    * No more `active` or `duration` suffix. `_active_count` -> `_count` or `_gcount`, `_duration_sum` -> `_sum` or `_gsum`
* The [docs for LongTaskTimer](https://docs.micrometer.io/micrometer/reference/concepts/long-task-timers.html#page-title) or [Micrometer Prometheus docs](https://docs.micrometer.io/micrometer/reference/implementations/prometheus.html) don't make it clear that `.publishPercentileHistogram()` might be necessary for certain use cases
* Consider making all `LongTaskTimer`s produce HistogramGauges and ditch Summary due to the semantic mismatch

shakuzen:
Thanks for opening the issue. If anyone wants to help with a pull request to improve the Micrometer Prometheus docs, it'd be appreciated. I think a table or something showing the mapping from Micrometer types to Prometheus types in different situations would help make things more clear.

PR #7546, opened 2026-05-30
docs: clarify Prometheus LongTaskTimer output

jewoodev:
**Summary**

Addresses #6507.

- Adds a small Timer/LongTaskTimer mapping table to the Prometheus implementation page that documents the default `${name}_seconds_count`/`${name}_seconds_sum`/`${name}_seconds_max` suffixes and the gauge-histogram `${name}_seconds_gcount`/`${name}_seconds_gsum` series exposed when `publishPercentileHistogram()` is enabled on a `LongTaskTimer`.
- Notes that `${name}_seconds_max` is always exposed as a separate metric family in the Prometheus Java Client 1.x registry, and adds a short guidance line on when `publishPercentileHistogram()` is useful (`histogram_quantile()` queries, histogram-based alerts).
- Adds a short note and an xref from the registry-neutral `concepts/long-task-timers.adoc` to the Prometheus section, so readers reach the implementation-specific details from either entry point.

This PR only touches the docs and does not change the migration guide or the runtime behavior discussed in the issue.

**Testing**

- `./gradlew :docs:antora`

### DIFF
--- a/docs/modules/ROOT/pages/implementations/prometheus.adoc
+++ b/docs/modules/ROOT/pages/implementations/prometheus.adoc
@@ -168,22 +168,24 @@ In contrast, without rate normalization, the counter drops back to zero on servi
 
 === Timers
 
-The Prometheus `Timer` produces two counter time series with different names:
+The Prometheus `Timer` produces the following time series:
 
-* `$\{name}_count`: Total number of all calls.
-* `$\{name}_sum`: Total time of all calls.
+* `$\{name}_seconds_count`: Total number of all calls.
+* `$\{name}_seconds_sum`: Total time of all calls.
+* `$\{name}_seconds_max`: Maximum recorded value within a configurable time window.
+
+Calling `publishPercentileHistogram()` on the builder adds `$\{name}_seconds_bucket` time series; the count, sum, and max series are unchanged.
 
 Again, representing a counter without rate normalization over some time window is rarely useful, as the representation is a function of both the rapidity with which the counter is incremented and the longevity of the service.
 
 Using the following Prometheus queries, we can graph the most commonly used statistics about timers:
 
-* Average latency: `rate(timer_sum[10s])/rate(timer_count[10s])`
-* Throughput (requests per second): `rate(timer_count[10s])`
+* Average latency: `rate(timer_seconds_sum[10s])/rate(timer_seconds_count[10s])`
+* Throughput (requests per second): `rate(timer_seconds_count[10s])`
 
 .Timer over a simulated service.
 image::implementations/prometheus-timer.png[Grafana-rendered Prometheus timer]
 
-[[prometheus-long-task-timers]]
 === Long task timers
 
 Long task timers measure tasks that are still running when Prometheus scrapes the registry.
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/implementations/prometheus.adoc
```diff
@@ -183,9 +183,31 @@ Using the following Prometheus queries, we can graph the most commonly used stat
 .Timer over a simulated service.
 image::implementations/prometheus-timer.png[Grafana-rendered Prometheus timer]
 
+[[prometheus-long-task-timers]]
```
shakuzen: What's the reason to add an explicit anchor name here? I think it's better to not do this because it will break existing links to this anchor.

docs/modules/ROOT/pages/implementations/prometheus.adoc
```diff
@@ -183,9 +183,31 @@ Using the following Prometheus queries, we can graph the most commonly used stat
...
-The following example shows a Prometheus query to plot the duration of a long task timer for a serial task is `long_task_timer_sum`. In Grafana, we can set an alert threshold at some fixed point.
+Long task timers measure tasks that are still running when Prometheus scrapes the registry.
+This differs from a `Timer`, which records completed events.
+
+The exposed Prometheus series depend on whether `publishPercentileHistogram()` is enabled:
+
+|===
+| Micrometer meter | Default Prometheus output | With `publishPercentileHistogram()`
+
+| `Timer`
```
shakuzen: I think it would be better to have the information about Timer in the Timer section of the documentation rather than under the LTT section. That would also eliminate the need to have the "Micrometer meter".

docs/modules/ROOT/pages/implementations/prometheus.adoc
```diff
@@ -183,9 +183,31 @@ Using the following Prometheus queries, we can graph the most commonly used stat
...
+| `$\{name}_seconds_count`, `$\{name}_seconds_sum`, and `$\{name}_seconds_max`
+| Adds `$\{name}_seconds_bucket`; the count, sum, and max series are unchanged.
+
+| `LongTaskTimer`
+| `$\{name}_seconds_count`, `$\{name}_seconds_sum`, and `$\{name}_seconds_max`
+| Replaces `$\{name}_seconds_count` and `$\{name}_seconds_sum` with the gauge-histogram series `$\{name}_seconds_gcount` and `$\{name}_seconds_gsum`, adds `$\{name}_seconds_bucket`, and keeps `$\{name}_seconds_max`.
```
shakuzen: I wonder if it would be more clear to list the names without explanation about time series being replaced. A link to https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#gaugehistogram might also be useful.

shakuzen (changes requested):
Thank you for the PR. Overall it looks good. I've left some feedback. Also, could you rebase on `1.15.x` and target that branch so we can include this in the docs for maintenance branches as well?

docs/modules/ROOT/pages/implementations/prometheus.adoc
jewoodev: Removed the explicit `[[prometheus-long-task-timers]]` anchor. The cross-reference in `concepts/long-task-timers.adoc` now targets the auto-generated `_long_task_timers` anchor.

docs/modules/ROOT/pages/implementations/prometheus.adoc
jewoodev: Moved the Timer series-suffix details into the `=== Timers` section, removed the `Micrometer meter` column from the LongTaskTimer table, and adjusted the nearby query/prose examples to use the same `_seconds_*` names.

docs/modules/ROOT/pages/implementations/prometheus.adoc
jewoodev: Reduced the `With publishPercentileHistogram()` cell to just the series names (`${name}_seconds_gcount`, `${name}_seconds_gsum`, `${name}_seconds_bucket`, and `${name}_seconds_max`), and added a line below the table linking to `GaugeHistogram` in the OpenMetrics specification.

jewoodev:
Thanks for the feedback. Addressed the inline comments and rebased onto `1.15.x`; PR base updated accordingly.

shakuzen (approved):
Thank you

shakuzen:
I added a polish commit with https://github.com/micrometer-metrics/micrometer/commit/c5705a372aa1dfb46f101f96e5eb182c03cd145b for this. We'll get it merged forward for inclusion in the upcoming maintenance releases' documentation.

PR #7547, opened 2026-05-30
Match HTTP/2 GlobalRequestProcessor MBean in TomcatMetrics

### LINKED ISSUE #7535: Include HTTP/2 GlobalRequestProcessor MBean in TomcatMetrics
**Describe the bug**
With `server.http2.enabled=true` on embedded Tomcat, `tomcat.global.*` meters (`received`, `sent`, `request`, `error`, `request.max`) reflect HTTP/1.1 traffic only. HTTP/2 traffic is collected by Tomcat but not exported.

**Root cause**
Tomcat registers two MBeans per connector when HTTP/2 is enabled:
- `Catalina:type=GlobalRequestProcessor,name="http-nio-XXXX"` — HTTP/1.1
- `Catalina:type=GlobalRequestProcessor,name="http-nio-XXXX",Upgrade="h2c"` — HTTP/2
`TomcatMetrics#registerGlobalRequestMetrics` queries `:type=GlobalRequestProcessor,name=*`. Per `javax.management.ObjectName` matching rules, MBeans with an extra `Upgrade=` key are not matched, so the HTTP/2 aggregator is silently ignored.
`registerServletMetrics` already uses the trailing-`,*` form (`:j2eeType=Servlet,name=*,*`) for the same reason; the global one was not updated when HTTP/2 landed in Tomcat.
A bare pattern change is not enough — `nameTag(ObjectName)` only reads the `name` key, so both MBeans would produce identical tag sets and the second registration would be rejected as a duplicate. The fix needs:
1. Broaden the pattern to match upgrade-protocol MBeans.
2. Include `Upgrade` as a tag (when present) to disambiguate the series.
Structurally the same issue as #7214 (Servlet + `WebModule`).

**Environment**
- Micrometer: 1.15.x / 1.16.x / `main` (`TomcatMetrics.java` has not been functionally changed since 2020)
- Registry: prometheus
- Tomcat: 10.1.x embedded

**Expected behavior**
`tomcat.global.*` should cover both HTTP/1.1 and HTTP/2, either summed or split via a protocol/upgrade tag.

jewoodev:
## Summary

When `server.http2.enabled` is set, Tomcat registers a second `GlobalRequestProcessor` MBean per connector with an extra `Upgrade` key (e.g. `Upgrade="h2c"`) for HTTP/2 traffic:

- `Catalina:type=GlobalRequestProcessor,name="http-nio-XXXX"` (HTTP/1.1)
- `Catalina:type=GlobalRequestProcessor,name="http-nio-XXXX",Upgrade="h2c"` (HTTP/2)

The previous `:type=GlobalRequestProcessor,name=*` pattern only matches MBeans whose key set is exactly `{name}` (per `javax.management.ObjectName` matching rules), so the HTTP/2 aggregator was silently ignored and `tomcat.global.*` reflected HTTP/1.1 only.

This change:

1. Broadens the pattern with a trailing `,*` so MBeans with extra keys are matched (mirrors the Servlet pattern already used in `registerServletMetrics`).
2. Surfaces the `Upgrade` key as an `upgrade` tag on the registered meters so the two per-connector series register with distinct label values instead of being rejected as duplicates.
3. Filters out `UpgradeGroupInfo` MBeans (WebSocket and other servlet upgrades), which share the ObjectName shape but expose a different attribute set.

> **Compatibility:** this adds an `upgrade` tag to every `tomcat.global.*` meter. Existing HTTP/1.1 series get `upgrade=none`; the HTTP/2 series gets `upgrade=h2c` (or `h2`). The label set stays consistent across protocols so dimensional registries like Prometheus accept both as separate series, but downstream queries/dashboards that depend on the exact label set will see one additional key.

> **UpgradeGroupInfo exclusion:** the `Upgrade` key on `:type=GlobalRequestProcessor` is shared between `RequestGroupInfo` (HTTP/2 per-connector aggregator) and `UpgradeGroupInfo` (servlet upgrades such as WebSocket). `UpgradeGroupInfo` exposes `msgsSent` / `msgsReceived` instead of `requestCount` / `processingTime` / ..., so registering meters against it would produce silently broken NaN/0 series. The binder discriminates on the presence of the `requestCount` attribute. (Tomcat wraps `RequestGroupInfo` in a `BaseModelMBean`, which makes `MBeanServer.isInstanceOf` and className-based discrimination both return false.)

> **Defect-class sweep:** the other `registerMetricsEventually` call sites in `TomcatMetrics` (`ThreadPool`, `StringCache`) and the `GlobalRequestProcessor` / `UpgradeGroupInfo` namespace itself were verified — no other ObjectName patterns in this binder collide with multi-key MBeans that Tomcat registers under the same shape.

## Testing

Added to `TomcatMetricsTest`:

- `globalRequestMetrics_areRegisteredForBothHttp11AndHttp2UpgradeMBeans` — registers two fake `GlobalRequestProcessor` MBeans (with and without an `Upgrade` key) against the platform MBean server and asserts that both produce `tomcat.global.received` meters tagged with `upgrade=none` and `upgrade=h2c` respectively. Fails on `main`, passes with this change.
- `globalRequestMetrics_skipNonRequestGroupInfoMBeans` — registers an MBean that matches the ObjectName pattern but lacks `requestCount` (mirrors `UpgradeGroupInfo`) and asserts no `tomcat.global.*` meters are registered.
- `globalRequestMetrics_areRegisteredEventuallyForLateMBeans` — registers the MBeans *after* `bindTo`, exercising the `MBeanServer` notification-listener path.

Local verification on Tomcat 8.5.100 (test classpath):

```bash
./gradlew :micrometer-core:test --tests TomcatMetricsTest
./gradlew :micrometer-core:checkFormat :micrometer-core:checkstyleMain :micrometer-core:checkstyleTest
```

`TomcatMetricsTest` (existing and new cases) and the style checks above all pass.

Closes gh-7535

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java
@@ -53,6 +53,16 @@ public class TomcatMetrics implements MeterBinder, AutoCloseable {
 
     private static final String OBJECT_NAME_SERVER_STANDALONE = JMX_DOMAIN_STANDALONE + OBJECT_NAME_SERVER_SUFFIX;
 
+    // The :type=GlobalRequestProcessor,name=*,Upgrade=* ObjectName shape is shared by
+    // RequestGroupInfo (HTTP/1.1 + HTTP/2 per-connector aggregator) and UpgradeGroupInfo
+    // (servlet upgrades such as WebSocket). They expose different attribute sets, so we
+    // only register meters for the former. Tomcat wraps RequestGroupInfo in a
+    // BaseModelMBean, so MBeanServer.isInstanceOf and className comparison both fail;
+    // we discriminate on the presence of "requestCount" (defining attribute of
+    // RequestGroupInfo, absent from UpgradeGroupInfo which exposes
+    // msgsReceived/msgsSent).
+    private static final String REQUEST_GROUP_INFO_ATTRIBUTE = "requestCount";
+
     private final @Nullable Manager manager;
 
     private final MBeanServer mBeanServer;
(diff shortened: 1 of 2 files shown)


jewoodev:
Hi @shakuzen — could you clarify what's blocking this PR? The three plausible blockers I see:

1. The added `upgrade` tag widens the existing `tomcat.global.*` label set (incl. `upgrade=none` for HTTP/1.1) — a breaking cardinality change.
2. Split vs. sum (issue #7535 left both open) needs a maintainer call.
3. The `UpgradeGroupInfo` vs `RequestGroupInfo` discrimination should land in Tomcat upstream instead.

I can rework toward an opt-in builder option if (1) is the blocker — just want to target the right one.

shakuzen:
Sorry. I should have left a comment clarifying the reason for the `blocker` label. The reason is that we still haven't released the 1.17.0 GA version, so we have not switched `main` to development of the 1.18.0 release yet. Hence we are blocked on merging this until we do that. We may still leave some review before it is unblocked.

shakuzen (approved):
Thanks for the PR. It looks good to me. I updated the tests to run with a real embedded Tomcat instance.

PR #7548, opened 2026-06-02
docs: clarify MultiGauge row registration

### LINKED ISSUE #4403: Multigauge Documentation lacks overwrite=true
**Describe the bug**
It's a bug in the documentation only!

I tried the code from https://micrometer.io/docs/concepts#_multi_gauge
it suggests that you run "register" periodically when running the query, and for me it is implied, that the "count" would be updated every time.
For this to work, the second parameter "overwrite" in my understanding must be set to true, the default is false.

I suggest changing the doumentation to 

```
statuses.register(
    resultSet.stream()
        .map(result -> Row.of(Tags.of("status", result.getAsString("status")), result.getAsInt("count")))
        .collect(toList()),
/* overwrite */ true
);
```

**Environment**
Micrometer 1.12

**To Reproduce**
```
 @Test
    fun `min sample`() {
        val gauge = MultiGauge.builder("test").register(meterRegistry)
        gauge.register(
            listOf(MultiGauge.Row.of(Tags.of("status", "some status"), 1)),
        )

        printGauge()

        gauge.register(
            listOf(MultiGauge.Row.of(Tags.of("status", "some status"), 2)),
        )

        printGauge()

        gauge.register(
            listOf(MultiGauge.Row.of(Tags.of("status", "some status"), 3)),
            /* overwrite = */ true,
        )

        printGauge()
    }

    private fun printGauge() {
        println(
            meterRegistry
                .find("test")
                .gauges()
                .map { Pair(it.id.tags, it.value()) },
        )
    }
    
prints:

[([tag(status=some status)], 1.0)]
[([tag(status=some status)], 1.0)] // not updated
[([tag(status=some status)], 3.0)] // updated    
```

**Additional context**
A related thought: Looking at the implementation of Multigauge.register, I see that gauges are deleted and recreated in the MeterRegistry. I can't see how this is multithreading save: If the metrics are collected by Prometheus just as the gauge is deleted, the gauge will be missing from the metrics. Should I open a ticket for this?

codingkiddo:
## Summary

Clarifies `MultiGauge` row registration behavior in both the reference documentation and JavaDoc.

This adds more explanation around the `overwrite` argument used by `MultiGauge.register(...)`, including how rows with the same tags are handled across repeated registrations and what happens when rows are no longer present.

## Related issue

Fixes #4403

## Notes

This is a documentation-only change. No runtime behavior is changed.

## Validation

- `./gradlew :micrometer-core:compileJava`
- `./gradlew :micrometer-core:test --tests "*MultiGauge*"`
- `./gradlew :micrometer-core:javadoc`
- `./gradlew checkFormat`

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java
@@ -29,6 +29,14 @@ import static java.util.Collections.emptySet;
 import static java.util.stream.Collectors.toSet;
 
 /**
+ * A virtual meter that manages a dynamic set of gauges with a common name and common
+ * tags.
+ * <p>
+ * A {@code MultiGauge} is not itself a {@link Meter}; it creates and removes individual
+ * {@link Gauge} meters as rows are registered. Each row contributes unique tags and a
+ * value function. Re-registering rows reconciles the gauges currently managed by this
+ * instance with the supplied rows.
+ *
  * @author Jon Schneider
  * @since 1.1.0
  */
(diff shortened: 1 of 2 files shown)


micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java
```diff
@@ -29,6 +29,12 @@
 import static java.util.stream.Collectors.toSet;
 
 /**
+ * A gauge that manages a set of rows with a common name and common tags.
```
shakuzen: I'm not sure the best wording for this, but I think it's better to avoid calling a `MultiGauge` "a gauge" even though the name `MultiGauge` makes it sound like a reasonable thing to do.

I would instead call it a "collection of gauges" or a virtual meter representing multiple gauges. Even these descriptions have problems, I think... `MultiGauge` is not actually a `Meter` and it doesn't *really* get registered with a MeterRegistry, despite having a `register(MeterRegistry)` method.

I'm open to other ideas. Maybe someone can come up with a better description. If not, I think calling it a "virtual meter" probably is best along with a brief explanation that it's not an actual `Meter`.

shakuzen (commented):
Thank you for working on this. I left one piece of feedback. Could you also rebase onto `1.15.x` and target the PR at that branch so we can include this in maintenance branches and merge forward from there?

micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java
codingkiddo: Thank you for the feedback.

I updated the class-level JavaDoc to avoid describing `MultiGauge` as a gauge and clarified that it is not itself a `Meter`, but manages individual `Gauge` meters for registered rows.

I also rebased the branch onto `1.15.x` and changed the PR target branch to `1.15.x`.

shakuzen (approved):
Looks great. Thank you.

PR #7562, opened 2026-06-03
Skip registering JCache meters when statistics are not enabled

jewoodev:
**Summary**

Follow-up to gh-5402 and gh-5409 for `JCacheMetrics`. See gh-5066.

When a JCache cache is bound without statistics enabled, the registered meters silently report zero, which is hard to diagnose. This change mirrors the Caffeine pattern: detect the stats-disabled state via `CompleteConfiguration#isStatisticsEnabled()`, log a warning explaining how to enable statistics, and skip registering the stats-dependent meters (`cache.gets`, `cache.puts`, `cache.evictions`, `cache.removals`).

If the JCache provider does not expose `CompleteConfiguration` (the configuration is `null` or `getConfiguration(CompleteConfiguration.class)` throws `IllegalArgumentException`), this assumes statistics are enabled to avoid a false-positive warning.

**Status of other cache binders**

- **Caffeine** — already handled in gh-5402 and gh-5409.
- **Hazelcast** — `MapConfig#isStatisticsEnabled()` makes the same detection possible, but `HazelcastCacheMetrics` accesses the cache through reflection via `HazelcastIMapAdapter`, so adding this check requires extending the adapter surface. Hazelcast remains applicable but is left out to keep this PR focused.
- **EhCache 2** — EhCache 2.5+ has statistics always on, so the check is unnecessary.
- **Guava** — Guava does not expose a public API to detect whether `recordStats()` was called. The upstream request (google/guava#7381) was closed in 2024 with no plans to add cache features.

**Testing**

- New test `JCacheMetricsTest#doNotReportMetricsWhenStatisticsAreDisabled` asserts the warning is emitted and that `cache.gets`, `cache.puts`, `cache.evictions`, and `cache.removals` are not registered when statistics are disabled.
- Existing tests remain unchanged: their mock `Cache` does not stub `getConfiguration`, so the helper falls into the `null → assume enabled` branch and they continue to register meters as before.
- `./gradlew :micrometer-core:check` passes locally.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java
@@ -195,6 +195,9 @@ public class JCacheMetrics<K, V, C extends Cache<K, V>> extends CacheMeterBinder
         }
     }
 
+    // Captured once at construction; runtime toggles via
+    // CacheManager#enableStatistics(...) are not tracked.
+    // Rebind the metrics if statistics are enabled later.
     private static boolean isStatisticsEnabled(Cache<?, ?> cache) {
         try {
             CompleteConfiguration<?, ?> config = cache.getConfiguration(CompleteConfiguration.class);
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java
@@ -15,12 +15,15 @@
  */
 package io.micrometer.core.instrument.binder.cache;
 
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.config.InvalidConfigurationException;
 import org.jspecify.annotations.Nullable;
 
 import javax.cache.Cache;
 import javax.cache.CacheManager;
+import javax.cache.configuration.CompleteConfiguration;
 import javax.management.*;
 import java.util.List;
 
@@ -36,10 +39,14 @@ import java.util.List;
  */
 public class JCacheMetrics<K, V, C extends Cache<K, V>> extends CacheMeterBinder<C> {
 
+    private static final InternalLogger log = InternalLoggerFactory.getInstance(JCacheMetrics.class);
+
     // VisibleForTesting
     @Nullable ObjectName objectName;
 
     private final boolean registerCacheRemovalsAsFunctionCounter;
+
+    private final boolean statisticsEnabled;
 
     /**
      * Record metrics on a JCache cache.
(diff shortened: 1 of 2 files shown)


ben-manes:
just fyi and likely acceptable, JCache allows statistics to be enabled at runtime by adding or removing a JMX MBean. I don't think anyone would or should do that, but its worth stating explicitly the decision to disregard supporting.

```java
  /**
   * Enables or disables statistics gathering for a managed {@link Cache} at
   * runtime.
   * <p>
   * Each cache's statistics object must be registered with an ObjectName that
   * is unique and has the following type and attributes:
   * <p>
   * Type:
   * <code>javax.cache:type=CacheStatistics</code>
   * <p>
   * Required Attributes:
   * <ul>
   * <li>CacheManager the URI of the CacheManager
   * <li>Cache the name of the Cache
   * </ul>
   *
   * @param cacheName the name of the cache to register
   * @param enabled   true to enable statistics, false to disable.
   * @throws IllegalStateException if the {@link CacheManager} or
   *                               {@link Cache} {@link #isClosed()}
   * @throws NullPointerException  if cacheName is null
   * @throws SecurityException when the operation could not be performed
   *                           due to the current security settings
   */
  void enableStatistics(String cacheName, boolean enabled);
```

jewoodev:
Thanks for pointing this out. This PR intentionally treats JCache statistics as a bind-time precondition: `JCacheMetrics` checks the setting when it is constructed and does not dynamically add or remove meters when `CacheManager#enableStatistics(...)` changes later. If statistics are enabled after the cache was initially bound, metrics should be bound again after enabling them.

I've added a short comment above `isStatisticsEnabled` so that decision is visible in the code.

If Micrometer should support runtime statistics toggles for JCache, I think that should be tracked as a separate issue from this fix.

shakuzen:
Thanks for the PR. We'll follow-up once we start merging things for 1.18

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java
```diff
@@ -167,6 +195,22 @@ protected void bindImplementationSpecificMetrics(MeterRegistry registry) {
         }
     }
 
+    // Captured once at construction; runtime toggles via
+    // CacheManager#enableStatistics(...) are not tracked.
+    // Rebind the metrics if statistics are enabled later.
```
shakuzen, line 200: This is probably good information to include in the JavaDoc visible to users, but we can follow-up with that polish so as to not block getting this merged.

shakuzen (approved):
Thanks for the PR. The changes look good to me. I just left one comment we can follow-up on about moving a comment into the relevant JavaDocs.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/JCacheMetrics.java
jewoodev: Thanks for the suggestion. I agree that this bind-time behavior should be visible in the public API documentation. I followed up in #7837 by moving the note to the `JCacheMetrics` class Javadoc and clarifying that enabling statistics at runtime requires creating and binding a new `JCacheMetrics` instance.

PR #7568, opened 2026-06-04
Skip registering Hazelcast meters when statistics are not enabled

### LINKED ISSUE #5066: Log a warning when instrumenting a cache that is not recording stats

jewoodev:
**Summary**

Follow-up to gh-5402, gh-5409, and gh-7562 for `HazelcastCacheMetrics`. See gh-5066.

When a Hazelcast `IMap` is bound without statistics enabled, its meters silently report zero, which can be misleading. This change adds `Config`-aware constructor and `monitor` overloads that resolve the effective map configuration through the public `Config#findMapConfig(name)` API and inspect `MapConfig#isStatisticsEnabled()`.

When statistics are disabled, `HazelcastCacheMetrics` logs a warning and skips the statistics-dependent meters: `cache.size`, `cache.gets`, `cache.puts`, `cache.entries`, `cache.entry.memory`, `cache.partition.gets`, the near-cache gauges, and the `cache.gets.latency`, `cache.puts.latency`, and `cache.removals.latency` timers. `cache.size` is included because it reads `LocalMapStats#getOwnedEntryCount()`, which is not meaningful when statistics are disabled.

The existing constructor and `monitor` overloads remain unchanged and assume statistics are enabled because they do not receive a `Config`. This preserves existing behavior and the Hazelcast 3/4 compatibility surface in `HazelcastIMapAdapter` without relying on Hazelcast internal SPI. The statistics-enabled state is captured when the binder is created; if statistics are enabled later, a new `HazelcastCacheMetrics` instance must be created and bound.

**Status of other cache binders**

- **Caffeine** — handled in gh-5402 and gh-5409.
- **JCache** — handled in gh-7562.
- **Hazelcast** — covered by this change.
- **EhCache 2** — EhCache 2.5+ has statistics always enabled, so no check is necessary.
- **Guava** — Guava does not expose a public API to detect whether `recordStats()` was called. The upstream request (google/guava#7381) was closed with no plans to add further cache features.

**Testing**

- `HazelcastCacheMetricsTest#doNotReportMetricsWhenStatisticsAreDisabled` verifies that the `Config`-aware API logs a warning and does not register statistics-dependent meters when map statistics are disabled.
- `HazelcastCacheMetricsTest#assumeStatisticsAreEnabledWhenConfigIsNotProvided` verifies that the existing API continues to register meters when no `Config` is provided.
- The `micrometer-core` checks and the latest-Hazelcast compatibility test pass.

### DIFF
--- a/micrometer-core/build.gradle
+++ b/micrometer-core/build.gradle
@@ -52,6 +52,7 @@ jar {
         Import-Package: \
             org.aspectj.*;resolution:=dynamic,\
             com.github.benmanes.caffeine.*;resolution:=dynamic;version="${@}",\
+            com.hazelcast.*;resolution:=dynamic;version="${@}",\
             net.sf.ehcache.*;resolution:=dynamic;version="${@}",\
             javax.cache.*;resolution:=dynamic;version="${@}",\
             org.hibernate.*;resolution:=dynamic;version="${@}",\
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/HazelcastCacheMetrics.java
@@ -15,6 +15,8 @@
  */
 package io.micrometer.core.instrument.binder.cache;
 
+import io.micrometer.common.util.internal.logging.InternalLogger;
+import io.micrometer.common.util.internal.logging.InternalLoggerFactory;
 import io.micrometer.core.instrument.*;
 import io.micrometer.core.instrument.binder.BaseUnits;
 import io.micrometer.core.instrument.binder.cache.HazelcastIMapAdapter.LocalMapStats;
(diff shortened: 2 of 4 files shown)


shakuzen:
Thanks for the PR. We'll follow-up once we start merging things for 1.18

shakuzen (commented):
I wonder how safe it is to get the MapConfig this way. Given the API design, it seems they intentionally don't want to give access to it. From what I gather, the normal way to access this (in embedded mode anyway) is via the HazelcastInstance or Config, but we don't have access to that from HazelcastCacheMetrics. I suppose we can test that this reflection continues to work and decide what to do if it ever breaks, but that feels like putting a future trap in the code.

I'm not sure what is best for us to do in this case.

jewoodev:
Thanks, that’s a good point. I checked the Hazelcast API more closely, and `NodeEngine` is marked as `@PrivateApi`, so I agree that keeping this reflection chain would create a compatibility trap.

I have a local revision that removes the internal reflection. It keeps the existing constructor and `monitor` behavior unchanged, and adds overloads that accept Hazelcast `Config`. When one of those overloads is used, the binder reads the effective map configuration through the public `Config#findMapConfig(name)` API and skips statistics-dependent meters when statistics are disabled.

Would this API direction make sense to you? If so, I can push the revision.

shakuzen:
> Would this API direction make sense to you?

Yes, I think that makes sense. I can't remember exactly why we didn't take a `Config` argument to begin with other than I guess we didn't think we needed it.

shakuzen (approved):

PR #7569, opened 2026-06-04
Apply MeterFilters when building Kafka meter id for comparison

### LINKED ISSUE #7541: KafkaMetrics dedup/removal miss when a MeterFilter transforms tags (e.g. replaceTagValues)
**Describe the bug**

After upgrading to micrometer 1.14.13 we started seeing intermittent `This <X> has been already registered` WARN logs from `MeterRegistry`. They kept reappearing on the 60s `checkAndBindMetrics` cycle and never went away on their own, and a few `kafka_consumer_*` series in `/actuator/prometheus` were stuck at a frozen value — still being scraped after the underlying Kafka metric was gone.

`MeterRegistry` stores meters under the fully-mapped id (every configured `MeterFilter` applied). `KafkaMetrics` builds its comparison/removal ids via `meterTags(metricName, true)`, which only manually mimics `commonTags`. Any other `MeterFilter` (`replaceTagValues`, `renameTag`, custom `map()`, …) is invisible to that path, so the lookup never matches what's actually stored — both the dedup check and the removal call miss for the same reason.

Concrete trace with `replaceTagValues("topic", v -> v.replace('_', '.'))` (a common filter to undo Kafka's `.`→`_` sanitization on the `topic` tag):

- First bind — Kafka exposes `{topic: "my_topic"}`. `bindMeter` goes through the registry's normal path, the filter rewrites it to `topic="my.topic"`, stored under `id(topic="my.topic")`.
- Next refresh, dedup — `KafkaMetrics` builds the comparison id with raw `topic="my_topic"`. `containsAll` against the registered (mapped) tags returns false, falls through to `bindMeter`, triggers `This <X> has been already registered`.
- Later, removal — Kafka drops the metric and `KafkaMetrics` calls `registry.remove(id(topic="my_topic"))`. `meterMap` is keyed by `id(topic="my.topic")` → silent no-op → zombie series persists in `/actuator/prometheus` at its last value.

**Environment**

- Micrometer version: 1.14.13 (also reproduces on `main`)
- Kafka clients: 3.8.1

**To Reproduce**

```java
PrometheusMeterRegistry registry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
// Undo Kafka's "." → "_" sanitization on the topic tag (Fetcher.java) so
// dashboards show real topic names — a fairly common filter.
registry.config().meterFilter(MeterFilter.replaceTagValues("topic", v -> v.replace('_', '.')));

new KafkaClientMetrics(consumer).bindTo(registry);

// 1. Let the consumer fetch (topic-level meter registered with the mapped topic).
// 2. Any metric-set change observed by the next checkAndBindMetrics tick
//    (partition revocation, sensor expiry, metadata refresh, …)
// → zombie series persists in /actuator/prometheus at its last value
// → WARN fires on the refresh cycle
```

**Expected behavior**

`registry.remove(...)` for an id derived from a metric whose tags are transformed by a filter should succeed (no zombies), and the dedup check should recognize the already-registered meter (no `already registered` WARN).

yoon-youngjin:
Fixes gh-7541.

`KafkaMetrics` built the id used to compare against and remove already-registered meters from raw tags (plus a common-tags workaround), so configured `MeterFilter`s that transform tags (`replaceTagValues`, `renameTag`, ...) made the comparison id diverge from the id the registry actually stores. As a result, stale meters were never removed.

This change routes the comparison id through `MeterRegistry#getMappedId` so all configured `MeterFilter`s are applied, and drops the common-tags workaround. A regression test reproduces the issue with a `replaceTagValues` filter.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
@@ -659,7 +659,7 @@ public abstract class MeterRegistry {
         return meterClass.cast(m);
     }
 
-    Meter.Id getMappedId(Meter.Id id) {
+    public Meter.Id getMappedId(Meter.Id id) {
         Meter m = preFilterIdToMeterMap.get(id);
         if (m != null && !isStaleId(id)) {
             return m.getId();
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java
@@ -100,8 +100,6 @@ class KafkaMetrics implements MeterBinder, AutoCloseable {
 
     private final Duration refreshInterval;
 
-    private @Nullable Iterable<Tag> commonTags;
-
     /**
      * Keeps track of current set of metrics.
      */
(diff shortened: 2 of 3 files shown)


yoon-youngjin:
Hi @jonatan-ivanov, this is ready for review whenever you have a chance. It targets `1.16.x` as requested and includes a regression test for the issue. CI and DCO are green. Thanks!

Issue #7577, opened 2026-06-06
ValidatedKeyValue does not enforce @NonNull on key (allows null key via validator overload)


zbnerd:
## Summary

`KeyValue.of(String, T, Predicate<? super T>)` can create a `KeyValue` instance with a null key when the validator overload is used.

The resulting object behaves mostly normally (`getValue`, `equals`, `hashCode`, `toString`), but contains `getKey() == null` and later throws `NullPointerException` in `compareTo`.

After investigating the surrounding code and nullability contracts, this appears to be caused by a missing null check in `ValidatedKeyValue`.

## Context

Initially I suspected several null-handling inconsistencies, but after looking more closely they appear to be intentional:

* `KeyValue.of("k", null)` → `NullPointerException`
* `KeyValue.of("k", null, v -> true)` → value stored as `"null"`

This appears intentional because the validator overload allows callers to customize value acceptance, and `String.valueOf(null)` returns `"null"`.

Likewise:

* `KeyValue.of((KeyName) null, "v", v -> true)` → `NullPointerException`

also appears intentional because the package is `@NullMarked` and `KeyName` is dereferenced through `asString()` before reaching the validator path.

The behavior below is the one that appears problematic:

```java
KeyValue kv = KeyValue.of((String) null, "v", v -> true);
```

## Reproduction

Using the validator overload:

```java
KeyValue kv = KeyValue.of((String) null, "v", v -> true);

kv.getKey();      // null
kv.getValue();    // "v"
kv.hashCode();    // works
kv.equals(kv);    // works
kv.toString();    // works
```

However:

```java
kv.compareTo(KeyValue.of("k", "v"));
```

throws:

```text
NullPointerException
```

because the implementation eventually calls:

```java
getKey().compareTo(...)
```

with a null receiver.

I was able to reproduce the same behavior by directly constructing `ValidatedKeyValue` from within the package.

### 1. Violates the package's nullability contract

The package is annotated with `@NullMarked` and the `key` parameter is not annotated `@Nullable`.

As a result, callers, static analysis tools, and maintainers are led to assume that a `KeyValue` key is always non-null. The validator overload is currently the only path that allows a null key to bypass that contract.

### 2. Creates a partially invalid object

The object is successfully constructed despite violating the non-null key assumption.

| Method | Result |
|----------|----------|
| `getKey()` | returns `null` |
| `getValue()` | works |
| `hashCode()` | works |
| `equals()` | works |
| `toString()` | works |
[...]

PR #7578, opened 2026-06-06
Enforce non-null key in ValidatedKeyValue constructor

zbnerd:
Issue #7577.

## Summary

`ValidatedKeyValue` did not enforce the package-level `@NullMarked` contract on its `key` parameter.

As a result, `KeyValue.of(String, T, Predicate<? super T>)` could create a `KeyValue` instance with `getKey() == null`. The object remained usable for most operations (`equals`, `hashCode`, `toString`, etc.) but later failed with `NullPointerException` in `compareTo`.

This change adds the missing null check and aligns `ValidatedKeyValue` with `ImmutableKeyValue`, which already enforces non-null constructor arguments.

## Change

```java
ValidatedKeyValue(String key, T value, Predicate<? super T> validator) {
    this.key = Objects.requireNonNull(key, "key must not be null");
    this.value = String.valueOf(assertValue(value, validator));
}
```

## Tests

Regression tests for the null-key fix:

- null key is rejected by the `ValidatedKeyValue` constructor
- null key is rejected through the public `KeyValue.of(String, T, Predicate)` factory (the entry point reported in #7577)
- a happy-path `KeyValue.of(...)` confirms the validator/value path still works

Note: the null-value handling (`String.valueOf(null) -> "null"` when the validator accepts it) is intentionally left untouched here, as discussed in #7577.

### DIFF
--- a/micrometer-commons/src/main/java/io/micrometer/common/ValidatedKeyValue.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/ValidatedKeyValue.java
@@ -15,6 +15,7 @@
  */
 package io.micrometer.common;
 
+import java.util.Objects;
 import java.util.function.Predicate;
 
 /**
@@ -31,7 +32,7 @@ class ValidatedKeyValue<T> implements KeyValue {
     private final String value;
 
     ValidatedKeyValue(String key, T value, Predicate<? super T> validator) {
-        this.key = key;
+        this.key = Objects.requireNonNull(key, "key must not be null");
         this.value = String.valueOf(assertValue(value, validator));
     }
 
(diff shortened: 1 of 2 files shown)


zbnerd:
One thing that makes this look unintentional is the consistency of the other overloads.

If the validator overload were intended to allow null keys whenever the predicate accepts the value, I would expect all key-accepting overloads to behave similarly.

Instead:

- of(String, String) rejects null keys
- of(KeyName, String) rejects null keys
- of(KeyName, T, Predicate) rejects null keys
- only of(String, T, Predicate) silently stores a null key

So regardless of whether one views the predicate overload as a special case for value validation, the current behavior leaves a single entry point that handles null keys differently from every other overload.

That asymmetry is what made me think this is more likely a missing null check than an intentional API distinction.

zbnerd:
I reviewed both the implementation and the tests after your feedback, simplified the regression tests to focus only on the behavior addressed by this issue, and updated the PR description accordingly. The production fix itself remains unchanged.

PR #7596, opened 2026-06-09
GH-5461: Document lifecycle and thread-safety contract for MeterBinder

### LINKED ISSUE #5461: [Question] Does MeterBinder needs to be thread safe?
Should I make my MeterBinder implementations thread-safe? Docs do not specify that...

won-seoop:
## What

Closes #5461.

Adds Javadoc to the `MeterBinder` interface documenting:

1. **Lifecycle** — `bindTo()` is typically called once per registry at application startup (configuration time). The same `MeterBinder` instance may be bound to multiple registries (e.g. a composite registry plus an individual registry in Spring Boot), with each call registering fresh meters on the supplied registry.

2. **Thread safety** — implementations are **not** required to make `bindTo` thread-safe. Concurrent invocations of `bindTo` on the same instance are not a supported use case. In normal usage all `bindTo` calls for a given instance happen serially during the single-threaded application context initialization phase.

## Why

The `MeterBinder` contract had no documentation about when `bindTo` is expected to be called or what thread-safety guarantees implementors must provide. Users implementing custom binders were left guessing — some used `synchronized` or `ReentrantLock` unnecessarily, while others had questions (as raised in #5461) about whether that was required.

The maintainer comment on #5461 confirmed: _"We will update documentation on this."_

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/MeterBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/MeterBinder.java
@@ -23,6 +23,20 @@ import io.micrometer.core.instrument.MeterRegistry;
  * <p>
  * Binders are enabled by default if they source data for an alert that is recommended for
  * a production ready app.
+ *
+ * <h2>Lifecycle</h2>
+ * {@link #bindTo(MeterRegistry)} is typically called once per registry during application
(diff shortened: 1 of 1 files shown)


shakuzen:
Thank you for the pull request. The DCO check is failing. Please take a look.

PR #7606, opened 2026-06-10
GH-5461: Document thread-safety expectations for MeterBinder.bindTo()

won-seoop:
## Summary

Addresses #5461.

`MeterBinder.bindTo()` had no Javadoc at all, leaving implementors uncertain about whether their implementations need to be thread-safe. Based on [the maintainer's guidance in that issue](https://github.com/micrometer-metrics/micrometer/issues/5461#issuecomment-1829014870), this PR adds Javadoc to `bindTo()` that clarifies:

- It is typically called once per registry at application configuration time.
- The Micrometer framework does not provide any thread-safety guarantee for concurrent calls to `bindTo()` on the same instance.
- Implementations that need to support concurrent binding to multiple registries must ensure their own thread safety.

## Test plan

- [ ] No behaviour change (Javadoc only); existing tests pass

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/MeterBinder.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/MeterBinder.java
@@ -26,6 +26,16 @@ import io.micrometer.core.instrument.MeterRegistry;
  */
 public interface MeterBinder {
 
+    /**
+     * Bind meters to the given registry.
+     * <p>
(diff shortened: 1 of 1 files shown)


shakuzen:
Thank you for the pull request. The build is failing due to formatting issues. The DCO check is also failing here and other PRs you've opened. I've marked them all `waiting for feedback`. We can't merge changes if the DCO check is failing, so we will wait to review until that is resolved.

PR #7620, opened 2026-06-13
Make virtual thread pinned event threshold configurable

### LINKED ISSUE #7563: Make event threshold configurable in VirtualThreadMetrics
`VirtualThreadMetrics` utilizes JFR to capture `VirtualThreadPinned` events and consequently populate the `jvm.threads.virtual.pinned` metric. Currently only events with a duration >= 20ms will be captured. The threshold is defined in `VirtualThreadMetrics` and it is not possible to configure a different threshold outside of this class. I would like to be able to configure the threshold when creating a `VirtualThreadMetrics` instance.

**Rationale**
In the context of our applications 20ms is a lot of time. We assume that we miss a lot of information by filtering out `VirtualThreadPinned` events that lasted less than 20ms.

HwangRock:
I ran into the same thing — the 20ms pinned-event threshold in `VirtualThreadMetrics` is hardcoded, so any `jdk.VirtualThreadPinned` event shorter than 20ms is dropped and never shows up in the `jvm.threads.virtual.pinned` timer. In our apps a lot of pins are well under 20ms, so we were effectively blind to them. Opening this PR to make the threshold configurable.

The threshold was already plumbed through `RecordingConfig.pinnedThreshold` and applied via `withThreshold(...)`; it just wasn't reachable from the public API. This adds two constructors:

```java
new VirtualThreadMetrics(Duration pinnedThreshold)
new VirtualThreadMetrics(Duration pinnedThreshold, Iterable<Tag> tags)
```

The existing constructors and the 20ms default are unchanged, so this is backward compatible. A null threshold is rejected with `NullPointerException` and a negative one with `IllegalArgumentException`, mirroring the existing `maxSizeBytes` guard.

Tests pass on JDK 24:

```
VirtualThreadMetricsTests > tests=7, skipped=1, failures=0, errors=0
  customPinnedThresholdConstructorAcceptsValidDuration         PASSED
  customPinnedThresholdWithTagsConstructorAcceptsValidDuration PASSED
  customPinnedThresholdConstructorRejectsNullThreshold         PASSED
  customPinnedThresholdWithTagsConstructorRejectsNullThreshold PASSED
  customPinnedThresholdConstructorRejectsNegativeThreshold     PASSED
  defaultConstructorIsUnchanged                                PASSED
  pinnedEventsShouldBeRecorded                                 SKIPPED (@DisabledForJreRange min JAVA_24)
```

Resolves #7563

### DIFF
--- a/micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
+++ b/micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
@@ -50,6 +50,8 @@ public class VirtualThreadMetrics implements MeterBinder, Closeable {
 
     private static final long DEFAULT_MAX_SIZE_BYTES = 10L * 1024 * 1024;
 
+    private static final Duration DEFAULT_PINNED_THRESHOLD = Duration.ofMillis(20);
+
     private static final String LIVE_THREADS_DESCRIPTION = "Approximate current number of virtual threads that are unfinished";
 
     private static final String METER_NAME_PREFIX = "jvm.threads.virtual.";
@@ -182,7 +184,7 @@ public class VirtualThreadMetrics implements MeterBinder, Closeable {
 
     private record RecordingConfig(Duration maxAge, long maxSizeBytes, Duration pinnedThreshold) {
         private RecordingConfig() {
-            this(DEFAULT_MAX_AGE, DEFAULT_MAX_SIZE_BYTES, Duration.ofMillis(20));
+            this(DEFAULT_MAX_AGE, DEFAULT_MAX_SIZE_BYTES, DEFAULT_PINNED_THRESHOLD);
         }
 
         private RecordingConfig {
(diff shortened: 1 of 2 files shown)


micrometer-java21/src/main/java/io/micrometer/java21/instrument/binder/jdk/VirtualThreadMetrics.java
```diff
@@ -155,7 +182,7 @@ public void close() {
 
     private record RecordingConfig(Duration maxAge, long maxSizeBytes, Duration pinnedThreshold) {
         private RecordingConfig() {
-            this(Duration.ofSeconds(5), 10L * 1024 * 1024, Duration.ofMillis(20));
+            this(DEFAULT_MAX_AGE, DEFAULT_MAX_SIZE_BYTES, Duration.ofMillis(20));
```
brushmate: The threshold should also be defined as a static field, no?

HwangRock: Good catch — extracted it to DEFAULT_PINNED_THRESHOLD so it lines up with the other defaults. Thanks!

PR #7621, opened 2026-06-15
Move secondary dependency versions out of Version Catalog

shakuzen:
This PR moves secondary/older versions of dependencies (such as Jersey 2, jOOQ 3.14, and Jetty 11) out of the Gradle Version Catalog (`gradle/libs.versions.toml`) and declares them as variables in the root `build.gradle`'s `allprojects` block. 

This prevents Dependabot from blindly upgrading these secondary versions (such as Jersey 2, Logback 1.2, Netty BOM 4.2, jOOQ 3.14, Jetty 9, and Jetty 11) to the latest versions when it updates the primary versions in the catalog.

### 1. The Core Limitation of Dependabot
Dependabot's internal engine groups all dependencies by their Maven coordinate (`group:artifact`). It does not support managing multiple versions of the exact same coordinate in a single project. 

If a coordinate is declared multiple times in a Gradle Version Catalog (even under different aliases, and even if one uses `version = "..."` and the other uses `version.ref = "..."`), Dependabot treats them as a single dependency. When it finds an upgrade, it updates **every single occurrence** of that coordinate in the catalog to the same newest version, clobbering the older version (e.g., upgrading Jersey 2 to Jersey 4).

This is a known, open limitation in the official Dependabot repository: [dependabot/dependabot-core #14040](https://github.com/dependabot/dependabot-core/issues/14040).

---

### 2. The Workaround & Solution
To resolve this limitation while still ensuring both versions receive automatic patch updates on maintenance branches, we have implemented a multi-directory isolation strategy:

1. **Remove Secondary Versions from the Catalog**:
   * We keep the oldest supported/production versions (such as **Jersey 2** and **jOOQ 3.14**) in `gradle/libs.versions.toml`.
   * We remove the secondary/newer versions (such as **Jersey 3**, **jOOQ Latest**, and **Jetty 11**) from the Version Catalog and declare them as literal strings inside their respective subproject `build.gradle` files (e.g., `samples/micrometer-samples-jersey3/build.gradle`).
   * Because the secondary versions are no longer in the catalog, Dependabot's root scan (`directory: "/"`) only sees one version of the coordinate in the catalog, preventing any clobbering.

2. **Understand Subproject Discovery & Dynamic Inclusion**:
   * Dependabot's root scan (`directory: "/"`) parses `settings.gradle` to discover subprojects. However, because the sample subprojects are dynamically included using a Groovy loop (e.g., `['core', 'hazelcast', ...].each { include "micrometer-samples-$it" }`), Dependabot's static parser **fails to discover or scan these subprojects**.
   * Consequently, the hardcoded Jersey 3 and jOOQ Latest dependencies in these subprojects are completely invisible to the root scan.

3. **Configure Multi-Directory Dependabot Updates (Configured on `main`)**:
   See https://github.com/micrometer-metrics/micrometer/pull/7632


samples/micrometer-samples-jersey3/build.gradle
```diff
@@ -8,10 +8,10 @@ dependencies {
     // Test sample project with SLFJ4 2.x / Logback; see gh-3415
     runtimeOnly(libs.logbackLatest)
 
-    implementation libs.jersey3ContainerJdkHttp
-    runtimeOnly libs.jersey3Hk2
+    implementation "org.glassfish.jersey.containers:jersey-container-jdk-http:3.1.11"
+    runtimeOnly "org.glassfish.jersey.inject:jersey-hk2:3.1.11"
 
-    testImplementation libs.jersey3TestFrameworkJdkHttp
+    testImplementation "org.glassfish.jersey.test-framework.providers:jersey-test-framework-provider-jdk-http:3.1.11"
```
shakuzen, line 14: These should be updated independently of the 2.x versions by dependabot due to the explicit configuration for this sample build.gradle, which doesn't get picked up by the root configuration because it is not statically imported. See https://github.com/micrometer-metrics/micrometer/pull/7632/changes

shakuzen: Same for the jooq sample below.

build.gradle
```diff
@@ -61,6 +61,17 @@ apply plugin: 'io.github.gradle-nexus.publish-plugin'
 allprojects {
     group = 'io.micrometer'
     ext.'release.stage' = releaseStage ?: 'SNAPSHOT'
+    // Legacy/older versions of dependencies are defined as ext properties here instead of
```
jonatan-ivanov, line 64: We can also have these in the `gradle.properties` file, e.g.: https://github.com/spring-projects/spring-boot/blob/2edca39504f8e9c2943262d3e60ba34b0e7751ac/gradle.properties

jonatan-ivanov (approved):

PR #7635, opened 2026-06-17
Filter null Humio meter events

### LINKED ISSUE #7579: HumioMeterRegistry uses Meter.match with mixed nullability but doesn't handle null after
This looks like a potentially long-standing bug uncovered by NullAway in #7575. Some of the write* methods can return null and that will be included in the produced JSON as is. I have a fix to filter out nulls before producing the JSON, but I want to look into why this hasn't been caught before and how it is handled by Humio.

xfocus3:
Fixes #7579.

## Summary
- filter out null Humio event payloads before joining the publish batch
- add a regression test covering a mixed batch where one meter produces a valid event and another produces `null`

## Verification
- `./gradlew :micrometer-registry-humio:test --tests io.micrometer.humio.HumioMeterRegistryTest.publishDropsNullMeterEvents --no-daemon`
- `./gradlew :micrometer-registry-humio:test --no-daemon`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/implementations/micrometer-registry-humio/src/test/java/io/micrometer/humio/HumioMeterRegistryTest.java
+++ b/implementations/micrometer-registry-humio/src/test/java/io/micrometer/humio/HumioMeterRegistryTest.java
@@ -60,6 +60,18 @@ class HumioMeterRegistryTest {
     }
 
     @Test
+    void publishDropsNullMeterEvents(@WiremockResolver.Wiremock WireMockServer server) {
+        HumioMeterRegistry registry = humioMeterRegistry(server);
+        registry.timer("my.timer", "status", "success");
+        registry.gauge("my.gauge", Double.NaN);
+
+        server.stubFor(any(anyUrl()));
+        registry.publish();
+        server.verify(postRequestedFor(urlMatching("/api/v1/ingest/humio-structured")).withRequestBody(equalTo(
+                "[{\"events\": [{\"timestamp\":\"1970-01-01T00:00:00.001Z\",\"attributes\":{\"name\":\"my_timer\",\"count\":0,\"sum\":0,\"avg\":0,\"max\":0,\"status\":\"success\"}}]}]")));
+    }
+
+    @Test
     void datasourceTags(@WiremockResolver.Wiremock WireMockServer server) {
         HumioMeterRegistry registry = humioMeterRegistry(server, "name", "micrometer");
         registry.counter("my.counter").increment();
--- a/implementations/micrometer-registry-humio/src/test/java/io/micrometer/humio/HumioMeterRegistryTest.java
+++ b/implementations/micrometer-registry-humio/src/test/java/io/micrometer/humio/HumioMeterRegistryTest.java
@@ -62,13 +62,15 @@ class HumioMeterRegistryTest {
     @Test
     void publishDropsNullMeterEvents(@WiremockResolver.Wiremock WireMockServer server) {
         HumioMeterRegistry registry = humioMeterRegistry(server);
-        registry.timer("my.timer", "status", "success");
-        registry.gauge("my.gauge", Double.NaN);
+        registry.counter("my.counter").increment();
+        clock.add(config.step());
+        Measurement measurement = new Measurement(() -> Double.NaN, Statistic.VALUE);
+        Meter.builder("my.meter", Meter.Type.GAUGE, List.of(measurement)).register(registry);
 
         server.stubFor(any(anyUrl()));
         registry.publish();
         server.verify(postRequestedFor(urlMatching("/api/v1/ingest/humio-structured")).withRequestBody(equalTo(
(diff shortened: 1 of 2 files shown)


implementations/micrometer-registry-humio/src/test/java/io/micrometer/humio/HumioMeterRegistryTest.java
```diff
@@ -59,6 +59,18 @@ void writeTimer(@WiremockResolver.Wiremock WireMockServer server) {
                 "[{\"events\": [{\"timestamp\":\"1970-01-01T00:00:00.001Z\",\"attributes\":{\"name\":\"my_timer\",\"count\":0,\"sum\":0,\"avg\":0,\"max\":0,\"status\":\"success\"}}]}]")));
     }
 
+    @Test
+    void publishDropsNullMeterEvents(@WiremockResolver.Wiremock WireMockServer server) {
+        HumioMeterRegistry registry = humioMeterRegistry(server);
+        registry.timer("my.timer", "status", "success");
```
jonatan-ivanov: I think we should record something with the timer but maybe it would be simpler just:
```java
registry.counter("my.counter").increment();
```

implementations/micrometer-registry-humio/src/test/java/io/micrometer/humio/HumioMeterRegistryTest.java
```diff
@@ -59,6 +59,18 @@ void writeTimer(@WiremockResolver.Wiremock WireMockServer server) {
...
+        registry.gauge("my.gauge", Double.NaN);
```
jonatan-ivanov: I think this should produce a time series for `my.gauge` where the value is `NaN`, does humio support it?

xfocus3:
Thanks for the review. I updated the regression test to use a stepped counter for the retained event and a custom meter with only non-finite measurements for the filtered event, so the test no longer depends on publishing a NaN gauge.

Verification:
- `./gradlew :micrometer-registry-humio:test --tests io.micrometer.humio.HumioMeterRegistryTest.publishDropsNullMeterEvents --no-daemon`
- `./gradlew :micrometer-registry-humio:test --no-daemon`
- `git diff --check`

implementations/micrometer-registry-humio/src/test/java/io/micrometer/humio/HumioMeterRegistryTest.java
```diff
@@ -59,6 +59,20 @@ void writeTimer(@WiremockResolver.Wiremock WireMockServer server) {
                 "[{\"events\": [{\"timestamp\":\"1970-01-01T00:00:00.001Z\",\"attributes\":{\"name\":\"my_timer\",\"count\":0,\"sum\":0,\"avg\":0,\"max\":0,\"status\":\"success\"}}]}]")));
     }
 
+    @Test
+    void publishDropsNullMeterEvents(@WiremockResolver.Wiremock WireMockServer server) {
+        HumioMeterRegistry registry = humioMeterRegistry(server);
+        registry.counter("my.counter").increment();
+        clock.add(config.step());
+        Measurement measurement = new Measurement(() -> Double.NaN, Statistic.VALUE);
+        Meter.builder("my.meter", Meter.Type.GAUGE, List.of(measurement)).register(registry);
```
jonatan-ivanov: Same comment as before: I think this should produce a time series for `my.meter` where the value is `NaN`, does humio support it?

The problem is not the type of the Meter: `Gauge` vs. `Meter` vs. something else, but having a time series with a value (`NaN`) is semantically different than not having a time series at all (like `null` vs. empty string/list/array/etc). Not having a value can indicate that the app is down or it's not instrumented, `NaN` can indicate that the app is up and reporting data but the value is `NaN`.

xfocus3:
Thanks, I updated the regression test again so it no longer relies on a non-finite time-series value for the filtered event. The retained event is still a stepped counter, and the filtered event now uses an empty custom meter that makes the Humio writer return no event.

Verification:
- `./gradlew :micrometer-registry-humio:test --tests io.micrometer.humio.HumioMeterRegistryTest.publishDropsNullMeterEvents --no-daemon`
- `./gradlew :micrometer-registry-humio:test --no-daemon`
- `git diff --check`

jonatan-ivanov:
>Thanks, I updated the regression test again so it no longer relies on a non-finite time-series value for the filtered event. The retained event is still a stepped counter, and the filtered event now uses an empty custom meter that makes the Humio writer return no event.

We might have some miscommunication here and it is also possible that I misunderstand something. The problem is not the test but the implementation (not only the one filtering line that you added I think), I think if I do this:
```java
registry.gauge("my.gauge", Double.NaN);
```
I should see a time series. But right now I don't since `writeGauge` returns `null`. I agree that `null` should be filtered out if there is any but I think that there should not be any `null` values if Humio supports `NaN` since we should publish that instead of nothing. Does this make sense?

On the other hand fixing that might be too disruptive for existing users so maybe we should do just the filtering as-is in your PR, let me discuss this with @shakuzen and get back to you.

PR #7636, opened 2026-06-17
Add MdcToKeyValuesObservationFilter to copy MDC entries into KeyValues

philsttr:
## Summary

Adds a reusable [`ObservationFilter`](https://github.com/micrometer-metrics/micrometer/blob/64e9a2697e16d0bb77cdc06a6859c0fceddabfab/micrometer-observation/src/main/java/io/micrometer/observation/ObservationFilter.java) that copies entries from the [slf4j **MDC** (Mapped Diagnostic Context)](https://www.slf4j.org/api/org/slf4j/MDC.html) into the [`Observation.Context`](https://github.com/micrometer-metrics/micrometer/blob/64e9a2697e16d0bb77cdc06a6859c0fceddabfab/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java#L1004-L1345) as low or high cardinality [`KeyValue`](https://github.com/micrometer-metrics/micrometer/blob/64e9a2697e16d0bb77cdc06a6859c0fceddabfab/micrometer-commons/src/main/java/io/micrometer/common/KeyValue.java)s. This lets contextual logging data (request id, tenant, account, …) flow into metrics and traces without custom Observation instrumentation.

I find this helpful when adding observations to existing applications where the MDC is already well-instrumented for logging. For example, say that an application does not generate log messages for certain requests, but does generate metrics and spans. In this case, it is helpful to include certain MDC values in Spans, since no log messages containing the MDC values exist. Additionally, even if there were log messages, it is sometimes helpful to include the additional context in Spans so that they can be queried directly in the tracing backend.

Micrometer ships no production `ObservationFilter` implementation today, so users have to roll their own. This provides a configurable, well-tested one for a common use case.

## What it does

Which MDC keys are copied, the cardinality at which they are added, and whether they are renamed is configured through a fluent `Builder` using an **ordered, first-match-wins** set of rules. Rules are declared in three phases, enforced at runtime:

1. **exact key** rules: `includeKey` / `excludeKey`
2. **matching** rules: `includeKeysMatching` / `excludeKeysMatching`, by regular expression or `Predicate<String>`
3. a single optional **catch-all default**: `includeByDefault` / `excludeByDefault`

`build()` throws `IllegalStateException` unless at least one include rule is declared (otherwise the filter would copy nothing).

Per-rule you can control:

- **cardinality**: high by default, `lowCardinality()` to opt in to low
- **renaming**:
  - `renameTo(...)`: a literal name for exact keys, a `$`-group back-reference replacement template for regular expressions
  - `renameUsing(UnaryOperator<String>)`: custom logic applied to the original key (the only rename option for predicate and default rules, which can match many keys)
  - `renameTo` and `renameUsing` are mutually exclusive, last call wins

### Example

```java
MdcToKeyValuesObservationFilter filter = MdcToKeyValuesObservationFilter.builder()
    // include a specific high-cardinality key, without renaming it
    .includeKey("order.id")
    // include a specific high-cardinality key and rename it
    .includeKey("accountId", spec -> spec.renameTo("account.id"))
    // include a specific low-cardinality key
    .includeKey("tenant", spec -> spec.lowCardinality())
    // exclude keys that would otherwise be included by later rules
    .excludeKey("foobar")
    .excludeKeysMatching("foobarbaz.*")
    // include by regex, renaming with back-references
    .includeKeysMatching("foo(.*)", spec -> spec.renameTo("Foo$1"))
    // include by predicate
    .includeKeysMatching(key -> key.startsWith("order"))
    // include everything else, renamed
    .includeByDefault(spec -> spec.renameUsing(key -> "mdc." + key))
    .build();

registry.observationConfig().observationFilter(filter);
```

## Design decisions

- **Ordered, first-match-wins rules**: precedence between overlapping rules (e.g. an exclude vs a broad include) needs to be obvious. An explicit order with first-match-wins makes the outcome for any key easy to reason about, rather than implicit priorities between rule kinds.
- **Three enforced phases (exact → matching → default), checked at runtime**: a fixed phase order keeps configurations readable and, crucially, lets an exact-key match short-circuit the regex/predicate scan (see the fast path below). The order is enforced with runtime `IllegalStateException`s from a single `Builder` type rather than a compile-time staged-builder interface chain, which would multiply the API surface for little gain; a unit test covers the enforcement.
- **High cardinality is the default**: MDC values are frequently unbounded (request id, account id), so high cardinality is the safe default; `lowCardinality()` is an explicit opt-in. Internally this is a `lowCardinality` boolean so the safe default coincides with the Java default (`false`).
- **`Consumer<…Spec>` arguments instead of parameter-heavy overloads**: each rule has several optional facets (cardinality, rename-to-literal, rename-via-template, rename-via-function). Expressing those as method parameters would mean either a combinatorial set of overloads or long positional argument lists (e.g. `includeKey(key, lowCardinality, renameTemplate, renamer)`) that are easy to misread and brittle as options grow. A single `Consumer<…Spec>` lets callers set only what they need fluently, keeps the method signatures stable when new options are added, and avoids ambiguous positional booleans. It also lets each rule kind expose its own spec type, so the compiler only offers the options that are valid for that kind (e.g. a predicate rule's spec has no `renameTo`).
- **Per-rule-kind rename specs**: exact keys accept a literal `renameTo`, regex keys accept a `$`-back-reference `renameTo`, while predicate and default rules accept only `renameUsing`. A predicate/default rule can match many keys, so a literal `renameTo` would collapse them all onto one name (last-write-wins), almost never the intent, whereas `renameUsing` renames each key individually. `renameTo`/`renameUsing` are mutually exclusive (last call wins) to avoid an ambiguous combined state.
- **At least one include rule is required**: a filter with only excludes (or no rules) copies nothing, which is almost certainly a misconfiguration, so `build()` fails fast instead of silently producing a no-op filter.
- **Read MDC directly via `org.slf4j.MDC`, slf4j optional**: rather than a pluggable value supplier. This keeps the common case simple and matches how the module already declares optional integrations; the dependency stays optional so non-MDC users pay nothing.
- **Caching assumes pure rule functions**: the key-resolution cache and the renamed-key memoization rely on predicates and rename functions being pure functions of the key. This is documented on the relevant methods; it is what makes per-key memoization correct.
- **Cache is bounded and on by default**: MDC-key cardinality is assumed low, so caching helps and a small bound protects memory if that assumption is wrong; it can be resized or disabled for genuinely high-cardinality keys.

## Performance

- **O(1) exact-key fast path**: when no rule could include any key other than the configured exact keys, the filter reads only those keys via `MDC.get(...)` instead of copying the whole MDC map.
- **Bounded key-resolution cache**: resolution of regex/predicate/default rules for a given MDC key is a pure function of the key, so it is memoized (on by default, size-bounded). Tunable or disablable via `keyResolutionCache(spec -> spec.enabled(...).maxSize(...))` for high MDC-key cardinality.

## Caveats

- **Only the MDC visible at stop time is copied**: the filter runs when the observation is stopped, so it sees only the MDC entries present at that moment. If the observed code populates the MDC and then removes those entries before the observation stops (e.g. a key set and cleared inside the observed scope), those values are no longer visible and cannot be copied.
- **Runs synchronously on the stopping thread**: rules, predicates, and rename functions execute on the thread that stops the observation, so they must be cheap and must not throw.
- **Rule functions should be pure**: predicates and rename functions may be memoized by the key-resolution cache, so a non-pure function (one that depends on changing state) can yield stale or surprising results for a recurring key. Disable the cache if you truly need per-observation evaluation.
- **Rename collisions are order-dependent**: if two MDC entries rename to the same key, the later one overwrites the earlier, but MDC iteration order is non-deterministic, so which one wins is unspecified.

## Build / packaging

- `slf4j-api` is added as an **optional** dependency (and an optional OSGi `Import-Package`), matching how `contextPropagation`, `javax.servletApi`, and `aspectjrt` are declared. The filter reads MDC directly via `org.slf4j.MDC`.
- A checkstyle suppression is added for the new file, since the project bans `org.slf4j.*` imports by default (same mechanism already used for `MeterRegistryLoggingTest`).

## Testing

`MdcToKeyValuesObservationFilterTests` drives the filter through a real `ObservationRegistry` / `Observation.start(...).stop()` flow and covers: rule ordering and first-match-wins, exact/regex/predicate/default include & exclude, cardinality routing, `renameTo` (literal + back-reference) and `renameUsing` (including last-call-wins and null handling), the exact-key fast path, the key-resolution cache (hit, disabled, and bounded/full behavior), builder phase enforcement, and the "no include rule" build failure.

`./gradlew check` passes (tests, NullAway, checkstyle, license, formatting, and a clean japicmp, since the change is purely additive).

### DIFF
--- a/config/checkstyle/checkstyle-suppressions.xml
+++ b/config/checkstyle/checkstyle-suppressions.xml
@@ -21,6 +21,7 @@
 	<suppress id="SLF4JIllegalImportCheck" files="benchmarks[\\/].+" />
 	<suppress id="SLF4JIllegalImportCheck" files="micrometer-osgi-test" />
 	<suppress id="SLF4JIllegalImportCheck" files="MeterRegistryLoggingTest" />
+	<suppress id="SLF4JIllegalImportCheck" files="MdcToKeyValuesObservationFilter" />
 
 	<suppress id="sysout" files="ClearCustomMetricDescriptors.java"/>
 	<suppress id="sysout" files="HttpSender.java"/>
--- a/micrometer-observation/build.gradle
+++ b/micrometer-observation/build.gradle
@@ -31,6 +31,7 @@ jar {
             javax.servlet.*;resolution:=optional,\
             io.micrometer.context.*;resolution:=optional,\
             org.aspectj.*;resolution:=optional,\
+            org.slf4j.*;resolution:=optional,\
             *
         '''.stripIndent()
     }
@@ -47,6 +48,10 @@ dependencies {
     // Aspects
     optionalApi libs.aspectjrt
 
+    // MDC source for MdcToKeyValuesObservationFilter
+    optionalApi libs.slf4jApi
+
+    testImplementation libs.slf4jApi
     testRuntimeOnly libs.logback12
 
     testImplementation(libs.archunitJunit5) {
--- a/micrometer-observation/src/main/java/io/micrometer/observation/MdcToKeyValuesObservationFilter.java
+++ b/micrometer-observation/src/main/java/io/micrometer/observation/MdcToKeyValuesObservationFilter.java
@@ -1,0 +1,1075 @@
+/*
+ * Copyright 2026 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micrometer.observation;
+
+import io.micrometer.common.KeyValue;
+import org.jspecify.annotations.Nullable;
+import org.slf4j.MDC;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.function.UnaryOperator;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * An {@link ObservationFilter} that copies entries from the SLF4J {@link MDC Mapped
+ * Diagnostic Context} into the {@link Observation.Context} as low or high cardinality
+ * {@link KeyValue key values}.
+ * <p>
+ * Which MDC keys are copied, the cardinality level at which they are added, and whether
+ * they are renamed is controlled by an ordered set of rules configured through the
+ * {@link Builder}. For each MDC entry, the rules are consulted and the <em>first</em>
+ * matching rule decides the outcome; an entry that matches no rule is dropped. Rules are
+ * added in three phases that must be declared in order:
+ * <ol>
+ * <li>exact key rules ({@code includeKey}/{@code excludeKey})</li>
+ * <li>matching rules, by regular expression or predicate
+ * ({@code includeKeysMatching}/{@code excludeKeysMatching})</li>
+ * <li>a single optional catch-all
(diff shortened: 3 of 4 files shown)


philsttr:
Closing due to feedback received over slack from @jonatan-ivanov 

> Should not this be implemented via Context Propagation instead of an ObservationFilter?
Also, this is a bigger feature and you are the first one who has asked for it so far. I also think that the implementation is a bit over-complicated.

philsttr:
Opened issue #7650 to revisit this.  If you see this as beneficial, provide feedback on that issue instead of here.

PR #7657, opened 2026-06-23
Fix ineffective ExemplarSampler cache in DefaultExemplarSamplerFactory

shakuzen:
Fix ineffective `ExemplarSampler` cache in `DefaultExemplarSamplerFactory`.

Wrap `double[]` in a value-based `DoubleArrayKey` to ensure that `equals()` and `hashCode()` compare array elements rather than array identities.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/DefaultExemplarSamplerFactory.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/DefaultExemplarSamplerFactory.java
@@ -20,6 +20,7 @@ import io.prometheus.metrics.core.exemplars.ExemplarSampler;
 import io.prometheus.metrics.core.exemplars.ExemplarSamplerConfig;
 import io.prometheus.metrics.tracer.common.SpanContext;
 
+import java.util.Arrays;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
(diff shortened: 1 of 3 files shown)


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/DefaultExemplarSamplerFactory.java
```diff
@@ -34,8 +35,7 @@ class DefaultExemplarSamplerFactory implements ExemplarSamplerFactory {
 
     private final ConcurrentMap<Integer, ExemplarSamplerConfig> exemplarSamplerConfigsByNumberOfExemplars = new ConcurrentHashMap<>();
 
-    @SuppressWarnings("ArrayAsKeyOfSetOrMap")
-    private final ConcurrentMap<double[], ExemplarSamplerConfig> exemplarSamplerConfigsByHistogramUpperBounds = new ConcurrentHashMap<>();
+    private final ConcurrentMap<DoubleArrayKey, ExemplarSamplerConfig> exemplarSamplerConfigsByHistogramUpperBounds = new ConcurrentHashMap<>();
```
shakuzen: Presumably the intention was to cache `ExemplarSamplerConfig` per unique sets of bucket bounds, but this before effectively had one entry per meter with a histogram because each meter's bucket array is a unique instance even if they have the same bounds.

shakuzen:
I guess we missed this when reviewing https://github.com/micrometer-metrics/micrometer/pull/4867 and https://github.com/micrometer-metrics/micrometer/pull/6494

jonatan-ivanov (approved):

PR #7664, opened 2026-06-25
  Fix Kotlin nullable value support in CaffeineCacheMetrics

### LINKED ISSUE #7582: Kotlin interoperability for CaffeineCacheMetrics broken/requires unchecked cast to work
**Describe the bug**
I access CaffeineCacheMetrics from a Kotlin class. Since the implementation of the fix for https://github.com/micrometer-metrics/micrometer/issues/6973 I can no longer call `CaffeineCacheMetrics.monitor(meterRegistry, store, "cacheName")`, which gives me the following compile time error 

Image

What happens here from my understanding is that my store field cannot be resolved dynamically, which used to work before the aforementioned change. All that changed was the nullability of the variable, which I assume clashes with Kotlin's own nullability handling, which I use by setting my store parameter to `private final val store: Cache<String, FSCacheValue?>` in order to set the FSCacheValue to be nullable.

My workaround is to pretend that FSCacheValue is not nullable by casting to Cache<String, FSCacheValue>, which solves the issue, however I lose the ability to have compile time nullability checks on the parametrized value within Kotlin.

**Environment**
 - Micrometer version: 1.16.5
 - OS: Linux
 - Java version: openjdk 21.0.11
 - Kotlin version: 2.3
**To Reproduce**
How to reproduce the bug:
```
class Main (
    @Autowired private val meterRegistry: MeterRegistry,
    ) {
    private final val store: Cache<String, FSCacheValue?> =
        Caffeine.newBuilder()
            .maximumSize(1000)
            .expireAfterAccess(Duration.ofHours(1))
            .recordStats()
            .build()

    fun main(args: Array<String>) {
        CaffeineCacheMetrics.monitor(meterRegistry, store, "cacheName") // This line does not compile
    }   
}
}
```
Note: This is within a Spring Boot application, hence the @Autowired for the MeterRegistry.

**Expected behavior**
The value for Cache<String, FSCacheValue?> is accepted as a parameter for CaffeineCacheMetrics::monitor, as was previously possible before the API change, and does not need to be casted to a non-nullable parametrized value.

**Additional context**
Micrometer is provided in my application through Spring Boot 4.

cfcromn:
Fixes #7582.

The CaffeineCacheMetrics class allows nullable cache value types through its
`V extends @Nullable Object` bound, but its four static monitor overloads
still declared an unbounded `V`.

Because the cache binder package is `@NullMarked`, Kotlin treats those
method-level value types as non-null. As a result, `Cache<K, V?>` and
`AsyncCache<K, V?>` could not be passed to CaffeineCacheMetrics.monitor(...)
without an unchecked cast.

This change aligns the generic bounds of all `Cache` and `AsyncCache`
monitor overloads with the class-level declaration. It does not change the
method bodies, return types, or meter registration behavior.

A Kotlin regression test verifies compile compatibility for nullable value
types in both synchronous and asynchronous Caffeine caches.

Tests:

* ./gradlew :micrometer-core:compileTestKotlin
* ./gradlew :micrometer-core:test --tests '*CaffeineCacheMetrics*'
* ./gradlew :micrometer-core:test
* ./gradlew :micrometer-core:spotlessCheck

### DIFF
--- a/micrometer-core/src/test/kotlin/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetricsKtTests.kt
+++ b/micrometer-core/src/test/kotlin/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetricsKtTests.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2026 VMware, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.micrometer.core.instrument.binder.cache
+
+import com.github.benmanes.caffeine.cache.AsyncCache
+import com.github.benmanes.caffeine.cache.Cache
+import com.github.benmanes.caffeine.cache.Caffeine
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry
+import org.junit.jupiter.api.Test
+
+class CaffeineCacheMetricsKtTests {
+
+    @Test
+    fun shouldAcceptKotlinNullableCacheValueType() {
+        val cache: Cache<String, FSCacheValue?> = Caffeine.newBuilder().recordStats().build()
+
+        CaffeineCacheMetrics.monitor(SimpleMeterRegistry(), cache, "cache")
+    }
+
+    @Test
+    fun shouldAcceptKotlinNullableAsyncCacheValueType() {
+        val cache: AsyncCache<String, FSCacheValue?> = Caffeine.newBuilder().recordStats().buildAsync()
+
+        CaffeineCacheMetrics.monitor(SimpleMeterRegistry(), cache, "cache")
+    }
+
+    private class FSCacheValue
+}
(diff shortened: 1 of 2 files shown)


cfcromn:
Hi, just following up on this PR.

This PR addresses the Kotlin nullability issue reported in #7582 by aligning the method-level generic bounds with the class-level nullable value bound, without changing the method bodies or meter registration behavior.

I also added a Kotlin regression test for both Cache<K, V?> and AsyncCache<K, V?>.

Could you let me know if this direction looks acceptable, or if there is anything I should adjust before review?

cfcromn:
Hi @shakuzen, sorry for the ping.
Since you previously looked into #7582, would you mind taking a quick look at this PR when you have a chance?
The CI is green and I’ve included Kotlin regression tests for both Cache<K, V?> and AsyncCache<K, V?>. Thanks!

micrometer-core/src/test/kotlin/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetricsKtTests.kt
```diff
@@ -0,0 +1,42 @@
...
+        val cache: Cache<String, FSCacheValue?> = Caffeine.newBuilder().recordStats().build()
+
+        CaffeineCacheMetrics.monitor(SimpleMeterRegistry(), cache, "cache")
+    }
+
+    @Test
+    fun shouldAcceptKotlinNullableAsyncCacheValueType() {
+        val cache: AsyncCache<String, FSCacheValue?> = Caffeine.newBuilder().recordStats().buildAsync()
+
+        CaffeineCacheMetrics.monitor(SimpleMeterRegistry(), cache, "cache")
```
shakuzen, line 46: Could we add two more tests so we fully cover the API that was updated?

shakuzen (approved):
Thank you for the PR and sorry for the long wait for review.

The fix looks good. Would you please rebase on and retarget the PR against `1.16.x` so we can fix this for upcoming maintenance releases? We'll merge forward from there.

micrometer-core/src/test/kotlin/io/micrometer/core/instrument/binder/cache/CaffeineCacheMetricsKtTests.kt
cfcromn: Thanks for the review! I’ve rebased and retargeted the PR to 1.16.x and added the two additional tests to cover the remaining Iterable<Tag> overloads.

shakuzen (approved):
Looks good. Thank you.

PR #7728, opened 2026-07-09
Cache baseline JARs in CI

### LINKED ISSUE #7102: Cache baseline JARs on CI
https://app.circleci.com/pipelines/github/micrometer-metrics/micrometer/11311/workflows/78f87759-0807-45d7-8045-9a66d945481e/jobs/62275

We had a CI build failure:

```
Execution failed for task ':micrometer-registry-statsd:downloadBaseline'.
> A failure occurred while executing de.undercouch.gradle.tasks.download.internal.DefaultWorkerExecutorHelper$DefaultWorkAction
   > de.undercouch.gradle.tasks.download.org.apache.hc.core5.http.NoHttpResponseException: The target server failed to respond
```

I've never seen this before so it's not been a significant issue, but each branch will always download the same baseline versions, which makes it a good candidate to include in the CI cache, so it doesn't repeatedly download the same JARs every CI build.

Develop-KIM:
Fixes #7102

This moves downloaded baseline JARs into the root project's `build/baselineLibs` directory and adds that directory to the CircleCI Gradle cache. That lets CI reuse the same baseline artifacts across jobs and reruns instead of downloading them into each module's build directory every time.

Verification:
- `./gradlew :micrometer-core:downloadBaseline --dry-run`
- `./gradlew :micrometer-core:downloadBaseline`
- `./gradlew :micrometer-core:japicmp`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,7 @@ commands:
           paths:
             - ~/.gradle/caches/modules-2
             - ~/.gradle/wrapper
+            - ~/micrometer/build/baselineLibs
           when: always
       - run:
           name: collect test reports
--- a/build.gradle
+++ b/build.gradle
@@ -456,11 +456,11 @@ subprojects {
                     }
 
                     src "${rootUrl}io/micrometer/${project.name}/${compatibleVersion}/${project.name}-${compatibleVersion}.jar"
-                    dest layout.buildDirectory.file("baselineLibs/${project.name}-${compatibleVersion}.jar")
+                    dest rootProject.layout.buildDirectory.file("baselineLibs/${project.name}-${compatibleVersion}.jar")
                 }
 
                 tasks.register("japicmp", me.champeau.gradle.japicmp.JapicmpTask) {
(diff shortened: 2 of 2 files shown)


jonatan-ivanov:
Thank you for the PR!
I think this is a step forward but I feel its not very fortunate to move everything under the `build` folder of the root project. Also, we should do this starting from `1.16.x`.

I also think that it would be better to remove `the de.undercouch.download` plugin and use
Gradle's dependency resolution mechanism to download and cache the jars, see: https://github.com/micrometer-metrics/micrometer/issues/7102#issuecomment-3781386601

I created a PR for this, please feel free to give us feedback: https://github.com/micrometer-metrics/micrometer/pull/7741

Develop-KIM:
Makes sense, going through Gradle's resolution instead of the download plugin is cleaner and keeps it out of the root build dir. Fine to close this in favor of #7741, left a note there about the SKIP case.

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/7741

PR #7736, opened 2026-07-10
Guard Step function meters against negative counts on count function reset

### LINKED ISSUE #2489: FunctionCounter / FunctionTimer count return negative rates on count function resets
Originally reported in https://github.com/spring-projects/spring-boot/issues/25444, it is a general problem with the count function handling in *Step variants, including the DropwizardFunctionTimer and DropwizardFunctionCounter used by the GraphiteMeterRegistry. On the one hand, we expect the count function provided to be monotonically increasing and a reset breaks that contract. On the other, we are guarding against exceptions in the count function by defaulting to a 0 value, which is a reset if the value wasn't 0 before.

seonwooj0810:
Fixes #2489

## Root cause
`StepFunctionCounter` and `StepFunctionTimer` record the amount for a step interval as the delta between the current and previous reading of the count/totalTime functions. Those functions are expected to be monotonically increasing, but they can decrease — e.g. when the monitored object is reset/replaced, or a guarded function such as [`safeLong` in `TomcatMetrics`](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.java) returns `0` after an exception. The negative delta was added as-is, so `count()`/`totalTime()` reported a negative value for that interval (as noted in the issue, originally surfaced via spring-projects/spring-boot#25444).

## Change
Clamp the per-interval delta at zero in both `StepFunctionCounter` and `StepFunctionTimer` (count and totalTime), so a reset is treated as no activity for that step rather than producing a negative count/rate. `StepFunctionTimer` already clamped the absolute readings with `Math.max(..., 0)`, but not the delta; this extends the same guard to the delta.

## Test evidence
Added regression tests that reproduce the issue from the report:
- `StepFunctionCounterTest.countShouldNotGoNegativeWhenCountFunctionResets` (mirrors the failing test in the issue)
- `StepFunctionTimerTest.countAndTotalTimeShouldNotGoNegativeWhenFunctionsReset`

Both fail before the change (`expected: 0.0 but was: -100.0`) and pass after.

## Scope note
This addresses the Step variants, where `count()`/`totalTime()` return the per-step delta and therefore go negative on a reset. The Dropwizard variants (`DropwizardFunctionCounter`/`DropwizardFunctionTimer`) mentioned in the issue return the absolute cumulative value from `count()`, so they don't exhibit a negative `count()`; they do feed a negative increment into their `DropwizardRate` EWMA on reset, which is a related but distinct symptom. Happy to extend this PR to cover that too if you'd prefer it in one change.

Verification done: (1) searched open PRs by symbol (`StepFunctionCounter`, `negative count function reset`) — none in flight; (4) grepped main and confirmed the unclamped delta add still exists in both classes; ran targeted `:micrometer-core:test` for both test classes (pass with fix, fail without); `:micrometer-core:spotlessCheck` passes.


seonwooj0810:
Superseded by #7771 — recreated after this PR was inadvertently closed by a force-push on my fork.

PR #7741, opened 2026-07-10
Cache baseline JARs

### LINKED ISSUE #7102: Cache baseline JARs on CI

jonatan-ivanov:
Remove the `de.undercouch.download` plugin and use
Gradle's dependency resolution mechanism to download and cache the jars

Closes gh-7102

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -466,8 +466,6 @@ subprojects {
 
                     fieldExcludes = []
                     methodExcludes = []
-
-                    onlyIf { compatibleVersion != 'SKIP' }
                 }
 
                 tasks.check.dependsOn(japicmp)
--- a/build.gradle
+++ b/build.gradle
@@ -429,6 +429,7 @@ subprojects {
                 apply plugin: 'me.champeau.gradle.japicmp'
 
                 tasks.register('downloadBaseline') {
+                    onlyIf { !project.gradle.startParameter.isOffline() }
                     doLast {
                         String originalGroup = project.group
                         project.group = 'downloadBaseline' // hack, see: https://github.com/melix/japicmp-gradle-plugin/issues/16#issuecomment-455132296
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,6 @@ buildscript {
         classpath libs.plugin.nexusPublish
         classpath libs.plugin.javaformat
         classpath libs.plugin.japicmp
-        classpath libs.plugin.downloadTask
         classpath libs.plugin.spotless
         classpath libs.plugin.bnd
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -417,35 +417,24 @@ subprojects {
...
-                    } else if (compatibleVersion.contains('-SNAPSHOT') ) {
-                        rootUrl = 'https://repo.spring.io/snapshot/'
-                    } else {
-                        rootUrl = repositories.mavenCentral().url
+                tasks.register('downloadBaseline') {
+                    doLast {
+                        String originalGroup = project.group
+                        project.group = 'downloadBaseline' // hack, see: https://github.com/melix/japicmp-gradle-plugin/issues/16#issuecomment-455132296
+                        try {
+                            Set<File> baselineFiles = configurations.detachedConfiguration(dependencies.create("${originalGroup}:${project.name}:${compatibleVersion}@jar")).files
```
jonatan-ivanov: This is where the interesting things happen: this creates a separate/detached `configuration` (so that it does not conflict with anything else) and defines the `compatibleVersion` of the same artifact as a dependency that gradle will download into the dependency cache (under `.gradle/caches`).

Develop-KIM:
One thing on the SKIP/offline case: `downloadBaseline` doesn't have its `onlyIf` guard anymore, and the `onlyIf { compatibleVersion != 'SKIP' }` on `japicmp` won't cover it since `onlyIf` doesn't skip a task's dependencies. So when `compatibleVersion` is `SKIP` (or offline), `downloadBaseline` still runs and tries to resolve `io.micrometer:<module>:SKIP@jar`, which fails. Maybe move that guard back onto `downloadBaseline`?

jonatan-ivanov:
@Develop-KIM Thanks for catching this, I added a check for `offline` builds. For the `compatibleVersion == 'SKIP'` use-case, I think we never used it and it can be easily excluded via `-x japicmp`.

build.gradle
```diff
@@ -417,38 +416,30 @@ subprojects {
...
-                    compress true
-                    String rootUrl
-                    if (compatibleVersion.contains('-M') || compatibleVersion.contains('-RC')) {
-                        rootUrl = 'https://repo.spring.io/milestone/'
-                    } else if (compatibleVersion.contains('-SNAPSHOT') ) {
-                        rootUrl = 'https://repo.spring.io/snapshot/'
-                    } else {
-                        rootUrl = repositories.mavenCentral().url
+                tasks.register('downloadBaseline') {
+                    onlyIf { !project.gradle.startParameter.isOffline() }
```
shakuzen: I know this is just preserving existing behavior, but since we're using Gradle for resolution instead of the download plugin, won't the JARs be taken from the local maven repository if available? So do we need to do anything special for offline mode? Maybe we just need to add a catch block to log that the JARs weren't available offline if resolution fails?

Develop-KIM: Agree this reads cleaner. Since resolution now goes through Gradle's cache, an offline build that already has the baseline cached could still run the comparison, so try/catch is less aggressive than skipping outright — and it'd also cover the `SKIP` case I mentioned earlier, since that just becomes a failed resolution we log rather than needing its own guard.

One thing to watch: `japicmp` only guards on `compatibleVersion != 'SKIP'`, so if `downloadBaseline` swallows the resolution failure, `japicmp` would still run against a missing `baselineLibs` jar. Probably want to skip `japicmp` too when the baseline file isn't there.

jonatan-ivanov: If they are available: yes. But if they are not and one runs the `japicmp` task while offline, `downloadBaseline` fails:
```
10: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':micrometer-java21:downloadBaseline'.
> Could not resolve all files for configuration ':detachedConfiguration3'.
   > Could not download micrometer-java21-1.15.0.jar (io.micrometer:micrometer-java21:1.15.0)
      > Could not get resource 'https://repo.maven.apache.org/maven2/io/micrometer/micrometer-java21/1.15.0/micrometer-java21-1.15.0.jar'.
         > Could not GET 'https://repo.maven.apache.org/maven2/io/micrometer/micrometer-java21/1.15.0/micrometer-java21-1.15.0.jar'.
            > repo.maven.apache.org
```
Right now with `--offline`, the build will succeed since neither `downloadBaseline` nor `japicmp` are happening. (Assuming everything else is available locally.)

Though this is somewhat obscure, as I'm thinking more about this, I think we should let this task fail and let the user exclude it if they are offline (`-x japicmp`) so that it is explicit what is not happening during build.

jonatan-ivanov:
Fyi, I made a few changes:
1. Reverted the offline check, users can exclude the task if they want to, see: https://github.com/micrometer-metrics/micrometer/pull/7741#discussion_r3591314222
2. Removed the "SKIP" version check from `japicmp`, users can exclude the task if they want to
3. I think I found a better hack for the Gradle dependency configuration limitation:
instead of creating a detached configuration for the current subproject, we can create one for the root project. Since we don't use `downloadBaseline` for the root project and it has different coordinates as any of the subprojects, dependency coordinates won't collide.
Also, this is only needed for Gradle 8. Gradle 9 does not have this limitation so 1.17.x and forward does not need to do this, they can use the current project's detached config instead of the root's (courtesy of @erichaagdev).

shakuzen (approved):
Seems fine as far as I can tell. We could also consider upgrading Gradle in `1.16.x` to get rid of the Gradle 8 workaround.

jonatan-ivanov:
1.16.x was upgraded to Gradle 9 so I rebased this PR which enabled me to remove the rest of the hacks.

shakuzen (approved):
Still looks good.

PR #7754, opened 2026-07-20
Fast-forward time-window rotation after a long idle gap

TimurRakhmatullin86:
## Problem

`AbstractTimeWindowHistogram.rotate()` (used by `TimeWindowPercentileHistogram` and
`TimeWindowFixedBoundaryHistogram`) and `TimeWindowSum.rotate()` cap the catch-up loop at
`ringBuffer.length` iterations but never fast-forward `lastRotateTimestampMillis` when the
cap is reached:

```java
do {
    resetBucket(ringBuffer[currentBucket]);
    ...
    timeSinceLastRotateMillis -= durationBetweenRotatesMillis;
    lastRotateTimestampMillis += durationBetweenRotatesMillis;
}
while (timeSinceLastRotateMillis >= durationBetweenRotatesMillis && ++iterations < ringBuffer.length);
```

After the meter is idle (no records, no snapshot) for longer than the whole ring buffer
spans (`bufferLength × durationBetweenRotates`, e.g. 3 × 40s = 2 min at defaults), each
`rotate()` advances the timestamp by at most one buffer-length worth of time and then exits,
leaving `lastRotateTimestampMillis` still in the past. Since `rotate()` runs at the top of
**every** `recordLong`/`recordDouble`/`takeSnapshot`, each following call wipes the **entire**
ring buffer again — destroying the samples recorded since the previous call — until the
timestamp finally catches up.

**Effect:** a burst of traffic that arrives after a quiet period (the classic "cold endpoint
hit after an idle night", an infrequently-scraped meter, or a publish interval longer than
the expiry) silently loses most of its samples. For a `Timer` with `publishPercentiles`, p50
of `{100,200,300,400,500}` recorded 1 ms apart after a 10-minute gap comes back as the last
value instead of the median; for `serviceLevelObjectives`, the SLO bucket counts are wiped.
`TimeWindowSum` is likewise reachable via `JvmHeapPressureMetrics`, under-reporting
`jvm.gc.overhead` after a scrape gap.

The sibling `TimeWindowMax.rotate()` already handles exactly this (issue #2647) with a
dedicated fast-forward branch; that fix was never applied to the other two implementations.

## Fix

Mirror `TimeWindowMax.rotate()`: when `timeSinceLastRotateMillis` covers the whole buffer,
clear every bucket once, reset `currentBucket`, and fast-forward
`lastRotateTimestampMillis = now - timeSinceLastRotateMillis % durationBetweenRotatesMillis`.

## Tests

`TimeWindowRotationTest.samplesRecordedAfterLongIdleGapAreNotDiscarded`, parameterized over
both histogram implementations: record once, advance a `MockClock` 10 minutes, record a
5-value burst, and assert the SLO bucket count is 5. It fails against the current code (the
burst is wiped, count 0) and passes with the fix. Existing `TimeWindow*Test` suites,
`spotlessCheck`, and `checkstyleMain` all pass.

`TimeWindowSum` gets the identical fix; it reads `System.currentTimeMillis()` directly (no
injectable clock and no existing unit test), so it can't be exercised deterministically
without an API change — happy to follow up if you'd like it made testable.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/AbstractTimeWindowHistogram.java
@@ -201,7 +201,8 @@ abstract class AbstractTimeWindowHistogram<T, U> implements Histogram {
     }
 
     private void rotate() {
-        long timeSinceLastRotateMillis = clock.wallTime() - lastRotateTimestampMillis;
+        long wallTime = clock.wallTime();
+        long timeSinceLastRotateMillis = wallTime - lastRotateTimestampMillis;
         if (timeSinceLastRotateMillis < durationBetweenRotatesMillis) {
             // Need to wait more for next rotation.
             return;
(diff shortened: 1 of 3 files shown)


TimurRakhmatullin86:
Friendly ping — happy to address any feedback or rebase if needed.

PR #7758, opened 2026-07-21
Swap StepFunctionCounter's last value atomically to avoid double-counting

TimurRakhmatullin86:
## Problem

`StepFunctionCounter.count()` did a non-atomic read-modify-write on the `last` value:

```java
double prevLast = last;                 // volatile read
last = f.applyAsDouble(obj2);           // volatile write
count.getCurrent().add(last - prevLast); // add the delta
```

`count()` is called from several threads: the step-rollover poller
(`StepMeterRegistry.pollMetersToRollover`), the publish thread of a push registry, `close()`
(`StepMeterRegistry.stop()` shuts the poller down without `awaitTermination`, then closing
rollover calls `count()`), and any actuator/`/metrics` scrape. When two of them run
concurrently they can read the same `prevLast`, each compute the same new value, and each add
the full delta to the current step's `DoubleAdder` — so the same increment of the underlying
function is counted twice and the published rate for that step is inflated (up to ~2x).

## Fix

Store `last` as an `AtomicLong` of the double bits and swap it with `getAndSet`, so concurrent
callers telescope their deltas (caller *k*'s `prevLast` is exactly caller *k-1*'s new value)
instead of each adding the full delta from a shared previous value:

```java
double newLast = f.applyAsDouble(obj2);
double prevLast = Double.longBitsToDouble(last.getAndSet(Double.doubleToLongBits(newLast)));
count.getCurrent().add(newLast - prevLast);
```

Single-threaded behaviour is unchanged. `_closingRollover()` still flushes the residual delta
correctly.

## Tests

`StepFunctionCounterTest.concurrentCountDoesNotOverCount` — 4 threads increment a source and
call `count()` concurrently; after a step rollover the polled total must equal the number of
increments. With the fix it telescopes exactly; against the current code it over-counts
(measured 18–68% over across runs). All existing `StepFunctionCounterTest` tests, `spotlessCheck`,
and `checkstyle` pass.

`StepFunctionTimer.accumulateCountAndTotal()` has the same class of race on its
`lastCount`/`lastTime` fields (plus a `lastUpdateTime` check-then-act); that needs a larger
change and is left as a follow-up.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepFunctionCounter.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepFunctionCounter.java
@@ -20,6 +20,7 @@ import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.FunctionCounter;
 
 import java.lang.ref.WeakReference;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.ToDoubleFunction;
 
 public class StepFunctionCounter<T> extends AbstractMeter implements FunctionCounter, StepMeter {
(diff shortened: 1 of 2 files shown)


TimurRakhmatullin86:
Friendly ping — happy to rebase or adjust if needed.

PR #7774, opened 2026-07-26
Allow filtering specific loggers in Log4j2Metrics (#6233)

### LINKED ISSUE #6233: provide a way to exclude some logger from being collected by Log4j2metrics

tanujkapoor-cmd:
Resolves #6233.

Provides the ability to selectively exclude noisy or specific loggers (e.g. `com.a.b.c`) from being collected by `Log4j2Metrics`.

 Changes
* Added an overloaded constructor to `Log4j2Metrics` accepting an optional `Predicate<LogEvent> ignoreEventPredicate`. 
* Added a convenience constructor `Log4j2Metrics(Iterable<Tag>, Predicate<LogEvent>)` so users aren't forced to pass a `LoggerContext` manually just to use the predicate.
* The `MetricsFilter.filter(LogEvent event)` method now returns `Result.NEUTRAL` early if the predicate evaluates to `true`, bypassing the counter increment.
* Wrapped the predicate evaluation in a `try-catch` block to ensure that a faulty user-supplied predicate will never break the Log4j2 logging filter chain.
* Fully annotated the new fields and parameters with `org.jspecify.annotations.Nullable` to align with Micrometer's existing conventions.

 Testing
* Added comprehensive unit test coverage (`Log4j2MetricsTest.java`), including tests for null predicates, throwing predicates, and level-specific ignores.
* All tests pass cleanly (`./gradlew :micrometer-core:test`).

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/logging/Log4j2Metrics.java
@@ -37,6 +37,8 @@ import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.Predicate;
+import org.jspecify.annotations.Nullable;
 
 import static java.util.Collections.emptyList;
 
@@ -72,6 +74,8 @@ public class Log4j2Metrics implements MeterBinder, AutoCloseable {
 
     private final List<PropertyChangeListener> changeListeners = new CopyOnWriteArrayList<>();
 
+    private final @Nullable Predicate<LogEvent> ignoreEventPredicate;
+
     public Log4j2Metrics() {
         this(emptyList());
     }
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
Thanks for the PR!
What is your use-case? Are you doing something like this in prod right now?
Would this work for you: https://github.com/micrometer-metrics/micrometer/pull/7357?
(The build is also failing)

tanujkapoor-cmd:
"Ah, I completely missed that #7357 was already open for this! The Set<String> approach makes perfect sense for performance reasons, and simple logger name exclusion covers the standard use-case just fine. I will close this PR in favor of #7357. Thanks for reviewing!"

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/pull/7357

PR #7785, opened 2026-08-01
Swap StepFunctionTimer's last values atomically to avoid double-counting

TimurRakhmatullin86:
`StepFunctionTimer.accumulateCountAndTotal()` reads the previous `lastCount`/`lastTime`, recomputes the function, and adds the delta to the step — a non-atomic read-modify-write on plain `volatile` fields.

A `StepMeterRegistry` touches the same meter from two threads: the rollover poller (`pollMetersToRollover` → `FunctionTimer::count`) and the publisher (`publish` → `count()`/`totalTime()`). Both call `accumulateCountAndTotal()`, and if both pass the 1 ms throttle in the same window they can read the same previous value and each add the full increase, so the step reports up to 2× the true count/total. The 1 ms guard only throttles how often the user function is called; it does not make the update atomic.

Fix: store the last observed values in `AtomicLong` and swap them with `getAndSet`, so concurrent callers telescope the deltas — the same fix as #7758 for `StepFunctionCounter`, applied to the sibling timer.

Added `concurrentAccumulateDoesNotOverCount` (4 threads × 10k increments); it fails on the current code (observed ~42k for a true 40k) and passes with the swap. `spotless`/`checkFormat` clean.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepFunctionTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepFunctionTimer.java
@@ -21,6 +21,7 @@ import io.micrometer.core.instrument.util.TimeUtils;
 
 import java.lang.ref.WeakReference;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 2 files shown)


TimurRakhmatullin86:
Friendly ping — let me know if there's anything I should update here.

PR #7802, opened 2026-08-05
Optimize PrometheusMeterRegistry and MicrometerCollector snapshot collection

shakuzen:
> [!NOTE]
> **Stacked on #7454.** That PR's head branch lives on a fork, so this PR has to target `main` and its diff includes #7454's commits until it merges.

Further simplification of `MicrometerCollector`/`PrometheusMeterRegistry` on top of #7454, continuing the move away from code structured around the 0.x Prometheus client. Public API and exposition output are unchanged (except where noted below, in the spirit of #7487's fail-at-registration behavior).

### Changes

- **One `MetricSnapshot` per metric family again.** `MetricFamilyDescriptor` has identity equality, so keying collected samples by descriptor produced one single-data-point snapshot per meter (e.g. 20k snapshots for 20k same-name meters) and relied on the exposition writers merging duplicate families on every scrape. `MicrometerCollector` now groups data points by Prometheus family name, restoring the pre-1.16 snapshot granularity observable via `PrometheusRegistry#scrape`. `Child` is now `void collect(BiConsumer<MetricFamilyDescriptor, DataPointSnapshot>)`, which also removes the `computeIfAbsent` boilerplate from every meter factory.
- **Gauge histogram flag derived from the meter type.** `LongTaskTimer` is the only meter producing a gauge histogram, and all meters in a collector share one type, so the `SnapshotFactory`/`registerCustomSnapshotFactory` mechanism collapses to one boolean in `createSnapshot`.
- **Fail registration instead of scrape on conflicting family types.** `PrometheusRegistry` validates family descriptors only when a collector is first registered; meters joining an existing collector are never re-validated by the client. `MicrometerCollector#add` now rejects type conflicts (e.g. same-name timers where only one publishes histogram buckets) through the standard `meterRegistrationFailed` path. Previously this registered "successfully" and then every scrape threw from inside the client. Also enabled the previously `@Disabled` `differentTypesThatProduceSamePrometheusMetricFamilyFailsToRegister`, which passes since #7487.
- **`newDistributionSummary` unified onto `addDistributionStatisticSamples`** (~70 duplicated lines removed). A `@Nullable TimeUnit` distinguishes raw summary values from time-converted timer/LTT values; exemplar scaling moved into the per-type supplier; the summary path takes a single `takeSnapshot()` per scrape instead of 4–5 independent reads.
- **Custom meters precompute per-statistic descriptor + `Labels` at registration**; scrapes no longer rebuild label lists or re-run the statistic-name switch. The created timestamp is captured at creation time, consistent with all other meter types (previously read at scrape time).
- Collector internals merged into a single registration map; `Format` enum replaced with a constant; `scrape(String)` delegates to the two-arg overload; unused imports removed.

### Behavior notes

- `getPrometheusRegistry().scrape()` returns one snapshot per family instead of one per meter; text/OpenMetrics output is byte-identical.
- Conflicting histogram/summary config under one name now fails at registration (like #7487's collisions) instead of throwing on every scrape.
- Custom meters' `_created` timestamp is the meter creation time instead of the scrape time.

### Testing

`MicrometerCollectorTest` updated to the new `Child` signature plus new tests pinning: one snapshot per family across separate descriptor instances, gauge-histogram on/off by meter type, and add-time type-conflict rejection. New registry tests cover same-name timers with mismatched histogram config in both registration orders. All existing `PrometheusMeterRegistryTest` assertions pass unchanged (61 tests) along with the compatibility TCK.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
@@ -60,6 +60,10 @@ class MicrometerCollector implements MultiCollector {
         this.originalMeterId = id;
     }
 
+    public String getConventionName() {
+        return conventionName;
+    }
+
     public void add(Meter.Id id, Child child, MetricFamilyDescriptor... registeredFamilies) {
         children.put(id, child);
         this.registeredFamilies.put(id, Arrays.asList(registeredFamilies));
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
@@ -78,7 +78,7 @@ class MicrometerCollector implements MultiCollector {
 
     private final Map<Meter.Id, Child> children = new ConcurrentHashMap<>();
 
-    private final Map<String, MetricFamilyDescriptor> descriptorsByFamilyName = new ConcurrentHashMap<>();
+    private final Map<String, MetricFamilyDescriptor> descriptorsByFamilyName = new HashMap<>(2);
 
     final String conventionName;
 
@@ -278,10 +278,13 @@ class MicrometerCollector implements MultiCollector {
      */
     MetricFamilyDescriptor getOrCreateDescriptor(MetricType metricType, String familyName,
             Collection<String> labelNames, String help) {
-        MetricFamilyDescriptor descriptor = descriptorsByFamilyName.computeIfAbsent(familyName,
-                // Unit is intentionally not set, see:
-                // https://github.com/OpenObservability/OpenMetrics/blob/1386544931307dff279688f332890c31b6c5de36/specification/OpenMetrics.md#unit
-                name -> MetricFamilyDescriptor.of(metricType, name).help(help).labelNames(labelNames).build());
+        MetricFamilyDescriptor descriptor = descriptorsByFamilyName.get(familyName);
+        if (descriptor == null) {
+            // Unit is intentionally not set, see:
+            // https://github.com/OpenObservability/OpenMetrics/blob/1386544931307dff279688f332890c31b6c5de36/specification/OpenMetrics.md#unit
+            descriptor = MetricFamilyDescriptor.of(metricType, familyName).help(help).labelNames(labelNames).build();
+            descriptorsByFamilyName.put(familyName, descriptor);
+        }
         if (descriptor.getType() != metricType) {
             throw new IllegalArgumentException(
                     "Meters with the same name must produce the same Prometheus metric family types. The family ("
@@ -299,9 +302,9 @@ class MicrometerCollector implements MultiCollector {
 
     @Override
     public MetricSnapshots collect() {
+        int familyCount = descriptorsByFamilyName.size();
         // descriptors are canonical per family
-        Map<MetricFamilyDescriptor, List<DataPointSnapshot>> dataPointsByFamily = new IdentityHashMap<>(
-                descriptorsByFamilyName.size());
+        Map<MetricFamilyDescriptor, List<DataPointSnapshot>> dataPointsByFamily = new IdentityHashMap<>(familyCount);
 
         for (Child child : children.values()) {
             child.collect((family, dataPoint) -> dataPointsByFamily
@@ -309,7 +312,7 @@ class MicrometerCollector implements MultiCollector {
                 .add(dataPoint));
         }
 
-        List<MetricSnapshot> snapshots = new ArrayList<>(dataPointsByFamily.size());
+        List<MetricSnapshot> snapshots = new ArrayList<>(familyCount);
         dataPointsByFamily.forEach((family, dataPoints) -> snapshots.add(createSnapshot(family, dataPoints)));
         return new MetricSnapshots(snapshots);
     }
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
@@ -96,7 +96,7 @@ class MicrometerCollector implements MultiCollector {
         Map<String, Family> families = new HashMap<>();
 
         for (Child child : children.values()) {
-            child.samples(conventionName)
+            child.samples()
                 .forEach(family -> families.compute(family.getConventionName(),
                         (name, matchingFamily) -> matchingFamily != null
                                 ? matchingFamily.addSamples(family.dataPointSnapshots) : family));
@@ -112,7 +112,7 @@ class MicrometerCollector implements MultiCollector {
 
     interface Child {
 
-        Stream<Family<?>> samples(String conventionName);
+        Stream<Family<?>> samples();
 
     }
 
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
@@ -78,6 +78,9 @@ import static java.util.concurrent.TimeUnit.SECONDS;
  */
 class MicrometerCollector implements MultiCollector {
 
+    // avoid cloning each time we want to iterate the enum values
+    private static final Statistic[] STATISTICS = Statistic.values();
+
     private final Map<Meter.Id, Child> children = new ConcurrentHashMap<>();
 
     private final Map<String, MetricFamilyDescriptor> descriptorsByFamilyName = new ConcurrentHashMap<>();
@@ -199,16 +202,14 @@ class MicrometerCollector implements MultiCollector {
         List<String> statKeys = new ArrayList<>(context.tagKeys);
         statKeys.add("statistic");
 
-        // precompute the family and labels per statistic; statistics that map to the
-        // same Prometheus name (e.g. TOTAL and TOTAL_TIME) get the same family
+        // Pre-populate all Statistic enum values at registration time so the EnumMap
+        // is read-only while supporting dynamic measurements.
         Map<Statistic, MeasurementFamily> familiesByStatistic = new EnumMap<>(Statistic.class);
-        for (Measurement measurement : measurements) {
-            familiesByStatistic.computeIfAbsent(measurement.getStatistic(), statistic -> {
-                List<String> statValues = new ArrayList<>(context.tagValues);
-                statValues.add(statistic.toString());
-                return new MeasurementFamily(customMeterFamily(statistic, statKeys, context.help),
-                        Labels.of(statKeys, statValues));
-            });
+        for (Statistic statistic : STATISTICS) {
+            List<String> statValues = new ArrayList<>(context.tagValues);
+            statValues.add(statistic.toString());
+            familiesByStatistic.put(statistic, new MeasurementFamily(
+                    customMeterFamily(statistic, statKeys, context.help), Labels.of(statKeys, statValues)));
         }
 
         add(context.id, samples -> {
--- a/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
+++ b/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
@@ -16,6 +16,7 @@
 package io.micrometer.prometheusmetrics;
 
 import io.micrometer.core.instrument.Meter;
+import io.prometheus.metrics.model.registry.MetricType;
 import io.prometheus.metrics.model.registry.MultiCollector;
 import io.prometheus.metrics.model.snapshots.CounterSnapshot;
 import io.prometheus.metrics.model.snapshots.DataPointSnapshot;
@@ -29,15 +30,12 @@ import io.prometheus.metrics.model.snapshots.MetricSnapshots;
 import io.prometheus.metrics.model.snapshots.SummarySnapshot;
 
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiConsumer;
-
-import static java.util.stream.Collectors.toList;
+import java.util.function.Supplier;
 
 /**
  * {@link MultiCollector} for Micrometer.
@@ -48,7 +46,9 @@ import static java.util.stream.Collectors.toList;
  */
 class MicrometerCollector implements MultiCollector {
 
-    private final Map<Meter.Id, Registration> registrations = new ConcurrentHashMap<>();
+    private final Map<Meter.Id, Child> children = new ConcurrentHashMap<>();
+
+    private final Map<String, MetricFamilyDescriptor> descriptorsByFamilyName = new ConcurrentHashMap<>();
 
     final String conventionName;
 
@@ -62,50 +62,58 @@ class MicrometerCollector implements MultiCollector {
         this.originalMeterId = id;
     }
 
-    public void add(Meter.Id id, Child child, MetricFamilyDescriptor... familyDescriptors) {
-        validateTypesMatchExistingFamilies(familyDescriptors);
-        registrations.put(id, new Registration(child, Arrays.asList(familyDescriptors)));
+    public MetricFamilyDescriptor getOrCreateDescriptor(MetricType metricType, String familyName,
+            Supplier<MetricFamilyDescriptor> factory) {
+        MetricFamilyDescriptor existing = descriptorsByFamilyName.get(familyName);
+        if (existing != null) {
+            if (existing.getType() != metricType) {
+                throw new IllegalArgumentException(
+                        "Meters with the same name must produce the same Prometheus metric family types. The family ("
+                                + familyName + ") was already registered with type " + existing.getType()
+                                + ", but this meter produces type " + metricType
+                                + ". This can happen when only some meters with"
+                                + " the same name publish histogram buckets.");
+            }
+            return existing;
+        }
+        MetricFamilyDescriptor created = factory.get();
+        existing = descriptorsByFamilyName.putIfAbsent(familyName, created);
+        if (existing != null) {
+            if (existing.getType() != metricType) {
+                throw new IllegalArgumentException(
+                        "Meters with the same name must produce the same Prometheus metric family types. The family ("
+                                + familyName + ") was already registered with type " + existing.getType()
+                                + ", but this meter produces type " + metricType
+                                + ". This can happen when only some meters with"
+                                + " the same name publish histogram buckets.");
+            }
+            return existing;
+        }
+        return created;
     }
 
-    /**
-     * Fail registration instead of failing later on scrape if the given family
-     * descriptors conflict in type with the metric families of the already registered
-     * meters, e.g. only some meters with the same name publish histogram buckets.
-     * {@link io.prometheus.metrics.model.registry.PrometheusRegistry} only validates this
-     * for the families present when the collector is registered, not for meters added to
-     * this collector afterwards.
-     */
-    private void validateTypesMatchExistingFamilies(MetricFamilyDescriptor[] familyDescriptors) {
-        // Existing registrations are mutually consistent since each was validated when
-        // added, so validating against a single one of them is sufficient; only the
-        // primary family of distribution meters can vary in type (SUMMARY vs HISTOGRAM)
-        // and every distribution meter registers it.
-        Iterator<Registration> registrationIterator = registrations.values().iterator();
-        if (!registrationIterator.hasNext()) {
-            return;
-        }
-        List<MetricFamilyDescriptor> existingDescriptors = registrationIterator.next().familyDescriptors;
+    public void add(Meter.Id id, Child child, MetricFamilyDescriptor... familyDescriptors) {
         for (MetricFamilyDescriptor descriptor : familyDescriptors) {
-            for (MetricFamilyDescriptor existingDescriptor : existingDescriptors) {
-                if (descriptor.getPrometheusName().equals(existingDescriptor.getPrometheusName())
-                        && descriptor.getType() != existingDescriptor.getType()) {
-                    throw new IllegalArgumentException(
-                            "Meters with the same name must produce the same Prometheus metric family types. The family ("
-                                    + descriptor.getPrometheusName() + ") was already registered with type "
-                                    + existingDescriptor.getType() + ", but this meter produces type "
-                                    + descriptor.getType() + ". This can happen when only some meters with"
-                                    + " the same name publish histogram buckets.");
-                }
+            MetricFamilyDescriptor existing = descriptorsByFamilyName.putIfAbsent(descriptor.getPrometheusName(),
+                    descriptor);
+            if (existing != null && existing.getType() != descriptor.getType()) {
+                throw new IllegalArgumentException(
+                        "Meters with the same name must produce the same Prometheus metric family types. The family ("
+                                + descriptor.getPrometheusName() + ") was already registered with type "
+                                + existing.getType() + ", but this meter produces type " + descriptor.getType()
+                                + ". This can happen when only some meters with"
+                                + " the same name publish histogram buckets.");
             }
         }
+        children.put(id, child);
     }
 
     public void remove(Meter.Id id) {
-        registrations.remove(id);
+        children.remove(id);
     }
 
     public boolean isEmpty() {
-        return registrations.isEmpty();
+        return children.isEmpty();
     }
 
     Meter.Id getOriginalId() {
@@ -114,10 +122,7 @@ class MicrometerCollector implements MultiCollector {
 
     @Override
     public List<MetricFamilyDescriptor> getMetricFamilyDescriptors() {
-        return registrations.values()
-            .stream()
-            .flatMap(registration -> registration.familyDescriptors.stream())
-            .collect(toList());
+        return new ArrayList<>(descriptorsByFamilyName.values());
     }
 
     @Override
@@ -129,8 +134,8 @@ class MicrometerCollector implements MultiCollector {
             .computeIfAbsent(descriptor.getPrometheusName(), name -> new FamilySamples(descriptor)).dataPoints
             .add(dataPoint);
 
-        for (Registration registration : registrations.values()) {
-            registration.child.collect(sink);
+        for (Child child : children.values()) {
+            child.collect(sink);
         }
 
         List<MetricSnapshot> snapshots = new ArrayList<>(samplesByFamilyName.size());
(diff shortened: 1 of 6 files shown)


implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -46,9 +48,9 @@ class MicrometerCollector implements MultiCollector {
 
     private final Map<Meter.Id, Child> children = new ConcurrentHashMap<>();
 
-    private final Map<Meter.Id, List<MetricFamilyDescriptor>> registeredFamilies = new ConcurrentHashMap<>();
+    private final Map<String, MetricFamilyDescriptor> descriptorsByFamilyName = new ConcurrentHashMap<>();
 
-    private final String conventionName;
+    final String conventionName;
```
jonatan-ivanov, line 83: Can we leave this private and add a package-private getter?

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +62,133 @@ class MicrometerCollector implements MultiCollector {
         this.originalMeterId = id;
     }
 
-    public void add(Meter.Id id, Child child, MetricFamilyDescriptor... registeredFamilies) {
+    public MetricFamilyDescriptor getOrCreateDescriptor(MetricType metricType, String familyName,
```
jonatan-ivanov: Can this be package-private?
Since the class itself is package-private, can we modify the visibility of the other public methods too?

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -585,19 +464,63 @@ private void addDistributionStatisticSamples(Meter.Id id, MicrometerCollector co
...
+        return timeUnit == null ? countAtBucket.bucket() : countAtBucket.bucket(timeUnit);
+    }
 
-            return families.build();
-        }, familyDescriptor(primaryType, getConventionName(id), tagKeys, id.getDescription()),
-                familyDescriptor(MetricType.GAUGE, getConventionName(id) + "_max", tagKeys, id.getDescription()));
+    private MetricFamilyDescriptor familyDescriptor(MicrometerCollector collector, MetricType metricType, String name,
+            Collection<String> labelNames, @Nullable String description) {
+        return collector.getOrCreateDescriptor(metricType, name,
+                () -> familyDescriptor(metricType, name, labelNames, description));
```
jonatan-ivanov: Since this method (`familyDescriptor`) is only called from here, the method can be moved to `MicrometerCollector`, so that `getOrCreateDescriptor` can call it directly (no need for the supplier and this lambda). For that `getOrCreateDescriptor`'s signature needs to be changed a bit:
```java
MetricFamilyDescriptor getOrCreateDescriptor(MetricType metricType, String familyName, Collection<String> labelNames, String description)
```
One drawback is, since `helpText` needs `prometheusConfig`, we need to call that here (or we pass `prometheusConfig`):
```java
return collector.getOrCreateDescriptor(metricType, name, labelNames, helpText(description));
```

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +62,133 @@ class MicrometerCollector implements MultiCollector {
...
+            Supplier<MetricFamilyDescriptor> factory) {
+        MetricFamilyDescriptor existing = descriptorsByFamilyName.get(familyName);
```
jonatan-ivanov: We are duplicating this block I guess mainly because of concurrency reasons, could this be replaced by `compute`?
```java
    MetricFamilyDescriptor getOrCreateDescriptor(MetricType metricType, String familyName, Collection<String> labelNames, String description) {
        return descriptorsByFamilyName.compute(familyName, (name, existing) -> {
            if (existing != null) {
                if (existing.getType() != metricType) {
                    throw new IllegalArgumentException(
                        "Meters with the same name must produce the same Prometheus metric family types. The family ("
                            + familyName + ") was already registered with type " + existing.getType()
                            + ", but this meter produces type " + metricType
                            + ". This can happen when only some meters with"
                            + " the same name publish histogram buckets.");
                }
                return existing;
            }

            return familyDescriptor(metricType, familyName, labelNames, description);
        });
    }
```
I did the supplier -> direct method call (`familyDescriptor`) refactor that I left a comment about in the registry but I hope the intent is visible even if that change is also here.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +62,133 @@ class MicrometerCollector implements MultiCollector {
...
     }
 
-    static class Family<T extends DataPointSnapshot> {
-
-        final String conventionName;
+    @FunctionalInterface
+    interface Child {
 
-        final List<T> dataPointSnapshots = new ArrayList<>();
+        void collect(BiConsumer<MetricFamilyDescriptor, DataPointSnapshot> samples);
```
jonatan-ivanov, line 397: Is there a way to simplify this?
If I understand this, in the registry we define a Supplier that accepts a BiConsumer (this `collect` method) and calls that BiConsumer with the descriptor and snapshot. And above in the `collect()` method we define that BiConsumer which will be called and it adds the passed parameters to a map.

Can we simplify this by redefining `Child#collect` so that it would return the descriptor and the snapshot without passing them to a BiConsumer and in `MicrometerCollector#collect` we just go through all the children, ask them for the descriptor and the snapshot and add them to the map.
Does this make sense?

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -109,28 +97,34 @@ public PrometheusMeterRegistry(PrometheusConfig config, PrometheusRegistry regis
...
+        NamingConvention convention = config().namingConvention();
+        List<Tag> tags = id.getTags();
+        int size = tags.size();
+        List<String> tagKeys = new ArrayList<>(size);
+        List<String> tagValues = new ArrayList<>(size);
+        for (int i = 0; i < size; i++) {
+            Tag tag = tags.get(i);
+            tagKeys.add(convention.tagKey(tag.getKey()));
+            tagValues.add(convention.tagValue(tag.getValue()));
+        }
```
jonatan-ivanov, line 118: Would this make sense?
```suggestion
        for (Tag tag : tags) {
            tagKeys.add(convention.tagKey(tag.getKey()));
            tagValues.add(convention.tagValue(tag.getValue()));
        }
```

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
```diff
@@ -512,136 +273,34 @@ public PrometheusRegistry getPrometheusRegistry() {
...
+    /**
+     * Adds a meter to the collector of its Prometheus name, creating and registering the
+     * collector if this is the first meter with that name. Everything the collector needs
+     * from the naming convention and the registry configuration is resolved once here, so
+     * that it does not have to be resolved again on every scrape.
+     */
+    private void applyToCollector(Meter.Id id, BiConsumer<MicrometerCollector, MeterContext> consumer) {
+        MeterContext context = createMeterContext(id);
         collectorMap.compute(getConventionName(id), (name, existingCollector) -> {
             if (existingCollector == null) {
```
jonatan-ivanov: Can we move `MeterContext` creation inside of `compute`'s re-mapping lambda?
```suggestion
        collectorMap.compute(getConventionName(id), (name, existingCollector) -> {
            MeterContext context = createMeterContext(id);
            if (existingCollector == null) {
```

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
jonatan-ivanov: It seems it is only used in tests, I would make these fields similar: either non-private or private + getter. I think from readability perspective the latter might be better.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +94,326 @@ class MicrometerCollector implements MultiCollector {
...
+     * buckets. {@link io.prometheus.metrics.model.registry.PrometheusRegistry} only
+     * validates this for the families present when the collector is registered, not for
+     * meters added to this collector afterwards.
+     */
+    MetricFamilyDescriptor getOrCreateDescriptor(MetricType metricType, String familyName,
+            Collection<String> labelNames, String help) {
+        MetricFamilyDescriptor descriptor = descriptorsByFamilyName.computeIfAbsent(familyName,
+                // Unit is intentionally not set, see:
+                // https://github.com/OpenObservability/OpenMetrics/blob/1386544931307dff279688f332890c31b6c5de36/specification/OpenMetrics.md#unit
+                name -> MetricFamilyDescriptor.of(metricType, name).help(help).labelNames(labelNames).build());
```
jonatan-ivanov: I think it might help readability a bit moving this into a separate "create" method:
```java
MetricFamilyDescriptor getOrCreateDescriptor(MetricType metricType, String familyName, Collection<String> labelNames, String help) {
    MetricFamilyDescriptor descriptor = descriptorsByFamilyName.computeIfAbsent(familyName, name -> createDescriptor(metricType, name, labelNames, help));
```

And
```java
MetricFamilyDescriptor createDescriptor(MetricType metricType, String familyName, Collection<String> labelNames, String help) {
    // Unit is intentionally not set, see:
    // https://github.com/OpenObservability/OpenMetrics/blob/1386544931307dff279688f332890c31b6c5de36/specification/OpenMetrics.md#unit
    return MetricFamilyDescriptor.of(metricType, familyName).help(help).labelNames(labelNames).build();
}
```

jonatan-ivanov: I also think adding an overload for `getOrCreateDescriptor` can also help:
```java
MetricFamilyDescriptor getOrCreateDescriptor(MetricType metricType, String familyName, MeterContext context) {
    return getOrCreateDescriptor(metricType, familyName, context.tagKeys, context.help);
}
```
So that we can do this everywhere but in `customMeterFamily`:
```java
MetricFamilyDescriptor family = getOrCreateDescriptor(MetricType.COUNTER, conventionName, context);
```

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +94,326 @@ class MicrometerCollector implements MultiCollector {
         this.originalMeterId = id;
     }
 
-    public void add(Meter.Id id, Child child, MetricFamilyDescriptor... registeredFamilies) {
+    void addCounter(MeterContext context, PrometheusCounter counter) {
+        MetricFamilyDescriptor family = getOrCreateDescriptor(MetricType.COUNTER, conventionName, context.tagKeys,
+                context.help);
+        addSingleDataPoint(context, family, () -> new CounterDataPointSnapshot(counter.count(), context.labels,
```
jonatan-ivanov: I'm thinking moving the body of the lambda to a separate method:
```java
CounterDataPointSnapshot createCounterDataPoint(MeterContext context, PrometheusCounter counter) {
    return new CounterDataPointSnapshot(counter.count(), context.labels, counter.exemplar(), context.createdTimestampMillis);
}
```
So that (with other refactors I suggested):
```java
void addCounter(MeterContext context, PrometheusCounter counter) {
    MetricFamilyDescriptor family = getOrCreateDescriptor(MetricType.COUNTER, conventionName, context);
    addSingleDataPoint(context, family, () -> createCounterDataPoint(context, counter));
}
```

Might not be useful everywhere, e.g.:
```java
() -> new InfoDataPointSnapshot(context.labels)
```
but there are a few cases when it seems slightly better.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +94,326 @@ class MicrometerCollector implements MultiCollector {
...
-                                ? matchingFamily.addSamples(family.dataPointSnapshots) : family));
+            child.collect(sink);
         }
 
-        Collection<MetricSnapshot> metricSnapshots = families.values()
-            .stream()
-            .map(Family::toMetricSnapshot)
-            .collect(toList());
+        List<MetricSnapshot> snapshots = new ArrayList<>(dataPointsByFamily.size());
+        dataPointsByFamily.forEach((family, dataPoints) -> snapshots.add(createSnapshot(family, dataPoints)));
```
jonatan-ivanov: I guess this has no/minimal performance difference compared to this (no array resize):
```java
List<MetricSnapshot> snapshots = new ArrayList<>(dataPointsByFamily.size());
for (Map.Entry<MetricFamilyDescriptor, List<DataPointSnapshot>> entry : dataPointsByFamily.entrySet()) {
    snapshots.add(createSnapshot(entry.getKey(), entry.getValue()));
}
```
but probably it has a performance advantage compared to this:
```java
List<MetricSnapshot> snapshots = dataPointsByFamily.entrySet().stream()
    .map(this::createSnapshot)
```

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +94,326 @@ class MicrometerCollector implements MultiCollector {
...
+                            + ", but this meter produces type " + metricType + ". This can happen when only some"
+                            + " meters with the same name publish histogram buckets.");
+        }
+        return descriptor;
+    }
+
     @Override
     public List<MetricFamilyDescriptor> getMetricFamilyDescriptors() {
-        return registeredFamilies.values().stream().flatMap(Collection::stream).collect(toList());
+        return new ArrayList<>(descriptorsByFamilyName.values());
```
shakuzen, line 300: Note: from the usage of this method in the Prometheus client, it seems like the API could be Collection or even Iterable instead of List. If it were, we could do, e.g. `Collections.unmodifiableCollection(descriptorsByFamilyName.values())` which would avoid copying the array of values. It's not a big deal because this is only called when a collector is registered to the Prometheus registry, but it's wasteful. I don't think the Prometheus client can fix this without breaking compatibility or introducing a different method.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +94,326 @@ class MicrometerCollector implements MultiCollector {
...
+                List<String> statValues = new ArrayList<>(context.tagValues);
+                statValues.add(statistic.toString());
+                return new MeasurementFamily(customMeterFamily(statistic, statKeys, context.help),
+                        Labels.of(statKeys, statValues));
+            });
+        }
+
+        add(context.id, samples -> {
+            for (Measurement measurement : measurements) {
+                MeasurementFamily family = requireNonNull(familiesByStatistic.get(measurement.getStatistic()));
```
m-burst, line 215: As we discussed in #7454, this is technically a breaking change for users whose `Iterable<Measurement>` misbehaves and does not honor the contract of `Meter#measure`. If an iterable returns different sets of statistics during initial iteration (at registration time) and during subsequent iterations, this will now throw, while previously it worked. Not sure if this deserves a notice in comments or even in the release notes.

shakuzen: Thanks for pointing this out. It's definitely worth bringing extra attention to it because it's bound to get lost in all the other changes made here. Everything in the PR is still in flux, which is why I have left the PR in draft state. These finer points will be reevaluated after we have more confidence that the general structure of the code is the one we're likely to merge.

shakuzen: For now I've done this: https://github.com/micrometer-metrics/micrometer/pull/7802/changes/dfc0290c936861f79aac8ac754abe0ea016b773f. We now have a test for this situation.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
shakuzen: I agree this is an annoying API shape that continues to make things hard to follow. Unfortunately, we can't do what you suggested because of meters that produce multiple DataPointSnapshot for different families (e.g. Timer). We could have different implementations of Collector for different Micrometer Meters and simplify this in at least some of them. But maybe it's best to leave that as a part 2 refactor? At least if we agree the changes so far (or somewhere short of doing that) are an improvement and worth keeping.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
shakuzen: Yes, I would have thought it's equivalent (see later). I would argue the code as is is better than the first because we can name the key and value variables instead of using `entry.getKey()` and `entry.getValue()` which requires you to check the type parameters to know what it is you're passing.

I benchmarked both ways, and as is actually allocates *less* because `IdentityHashMap` has an optimized `forEach` implementation, whereas the enhanced for loop with `entrySet()` results in various allocation that does not get eliminated.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
shakuzen: Yes, it's functionally equivalent and reads much better. I went back and forth between the two. Enhanced for loops create iterators, though, and since we have index access we don't need to. It was a micro-optimization choice to squeeze out more allocations.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -15,38 +15,70 @@
...
  */
 class MicrometerCollector implements MultiCollector {
 
+    // avoid cloning each time we want to iterate the enum values
+    private static final Statistic[] STATISTICS = Statistic.values();
+
     private final Map<Meter.Id, Child> children = new ConcurrentHashMap<>();
 
-    private final Map<Meter.Id, List<MetricFamilyDescriptor>> registeredFamilies = new ConcurrentHashMap<>();
+    private final Map<String, MetricFamilyDescriptor> descriptorsByFamilyName = new ConcurrentHashMap<>();
```
shakuzen: I think this can be populated on construction and be read-only. Only meters with the same name will be added to a collector, and all of them will contribute DataPointSnapshots to the same families. We would have to change some things so we have the information necessary in the constructor. The current `getOrCreateDescriptor` method also is checking that types of added meters match, which is not an obvious behavior from the name of the method. I tried to see how this might look, but it got messy quickly. An easier improvement would be switching to HashMap with an initial size of 2. I don't think we need CHM.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
shakuzen: I think this is resolved now.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistry.java
shakuzen: I think this is resolved now.

shakuzen:
It's not quite as nice as I hoped it would turn out from an API/code structure perspective, but hopefully it's an improvement over the previous code. I'm out of time and good ideas. Efficiency is improved, tests were added and all are passing, benchmarks were added. I'll come back to this once @jonatan-ivanov has had a chance to review.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
```diff
@@ -60,84 +92,325 @@ class MicrometerCollector implements MultiCollector {
...
 
-        Stream<Family<?>> samples();
+    // TODO: remove this cumulative -> non cumulative conversion
+    // ClassicHistogramBuckets is not cumulative but the histograms we use are cumulative
+    // so we convert it to non-cumulative just for the Prometheus client library
+    // can convert it back to cumulative.
+    private static ClassicHistogramBuckets nonCumulativeBuckets(CountAtBucket[] histogramCounts, long count,
+            @Nullable TimeUnit timeUnit) {
+        List<Double> buckets = new ArrayList<>();
+        List<Number> counts = new ArrayList<>();
```
m-burst, line 362: Another possible optimization: these can be `double[]` and `long[]` instead of lists. The size is known in advance, and `ClassicHistogramBuckets#of` will convert the lists into primitive arrays anyway.

shakuzen: Good spot. It's worth looking into, but I'd like to leave it as a follow-up after this gets merged to help make review of this easier. We're running out of time to get this into 1.18.0-M1 and I'll be busy with other work, but we can come back to this after M1.

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
jonatan-ivanov: Changes after I made this comment improved a lot around this I think. Even though the method signature is the same but for me it's easier to follow what is going on. I think it is possible to define types that reflect what is happening (a meter can create multiple families and each family can have multiple datapoints) but I'm not sure it worth it now in reflection of the latest changes (it makes it simpler to follow the execution but constructing those objects will be a bit more complicated than just samples.apply(...)).

shakuzen:
I think we're good to merge this now. We can follow-up with more improvements, but I want to get this into M1 and move on to other things. The benchmarks show substantial improvement in scraping with all of the accumulated changes in `main`. There is a slight increase in the cost of creating new meters, but it is small enough in absolute terms I think it's probably worth it, given new meter creation is bounded and metrics should be low cardinality.

<details><summary>Benchmark results against 1.17.0</summary>
<p>

```
Benchmark                                                     (scrapeMeterCount)  Mode  Cnt       Score     Error   Units
PrometheusRegistryBenchmark.collect                                          500  avgt    5      48.825 ±   3.058   us/op
PrometheusRegistryBenchmark.collect:gc.alloc.rate                            500  avgt    5    3621.611 ± 225.243  MB/sec
PrometheusRegistryBenchmark.collect:gc.alloc.rate.norm                       500  avgt    5  185392.171 ±   0.011    B/op
PrometheusRegistryBenchmark.collect:gc.count                                 500  avgt    5      91.000            counts
PrometheusRegistryBenchmark.collect:gc.time                                  500  avgt    5      55.000                ms
PrometheusRegistryBenchmark.meterCreation                                    500  avgt    5     136.767 ±   2.917   ns/op
PrometheusRegistryBenchmark.meterCreation:gc.alloc.rate                      500  avgt    5    2750.551 ± 144.164  MB/sec
PrometheusRegistryBenchmark.meterCreation:gc.alloc.rate.norm                 500  avgt    5     470.564 ±   3.822    B/op
PrometheusRegistryBenchmark.meterCreation:gc.count                           500  avgt    5     107.000            counts
PrometheusRegistryBenchmark.meterCreation:gc.time                            500  avgt    5      58.000                ms
PrometheusRegistryBenchmark.scrape                                           500  avgt    5      96.422 ±   2.293   us/op
PrometheusRegistryBenchmark.scrape:gc.alloc.rate                             500  avgt    5    2595.714 ±  61.147  MB/sec
PrometheusRegistryBenchmark.scrape:gc.alloc.rate.norm                        500  avgt    5  262456.575 ±   2.066    B/op
PrometheusRegistryBenchmark.scrape:gc.count                                  500  avgt    5      99.000            counts
PrometheusRegistryBenchmark.scrape:gc.time                                   500  avgt    5      56.000                ms
```

</p>
</details> 

<details><summary>Benchmark results against this PR</summary>
<p>

```
Benchmark                                                     (scrapeMeterCount)  Mode  Cnt      Score     Error   Units
PrometheusRegistryBenchmark.collect                                          500  avgt    5     13.505 ±   8.032   us/op
PrometheusRegistryBenchmark.collect:gc.alloc.rate                            500  avgt    5   1379.244 ± 679.629  MB/sec
PrometheusRegistryBenchmark.collect:gc.alloc.rate.norm                       500  avgt    5  19224.047 ±   0.028    B/op
PrometheusRegistryBenchmark.collect:gc.count                                 500  avgt    5     51.000            counts
PrometheusRegistryBenchmark.collect:gc.time                                  500  avgt    5     54.000                ms
PrometheusRegistryBenchmark.meterCreation                                    500  avgt    5    167.206 ±   4.825   ns/op
PrometheusRegistryBenchmark.meterCreation:gc.alloc.rate                      500  avgt    5   2718.340 ± 100.738  MB/sec
PrometheusRegistryBenchmark.meterCreation:gc.alloc.rate.norm                 500  avgt    5    553.953 ±   3.444    B/op
PrometheusRegistryBenchmark.meterCreation:gc.count                           500  avgt    5    103.000            counts
PrometheusRegistryBenchmark.meterCreation:gc.time                            500  avgt    5     61.000                ms
PrometheusRegistryBenchmark.scrape                                           500  avgt    5     58.722 ±   3.776   us/op
PrometheusRegistryBenchmark.scrape:gc.alloc.rate                             500  avgt    5   1545.362 ±  98.487  MB/sec
PrometheusRegistryBenchmark.scrape:gc.alloc.rate.norm                        500  avgt    5  95144.206 ±   0.012    B/op
PrometheusRegistryBenchmark.scrape:gc.count                                  500  avgt    5     60.000            counts
PrometheusRegistryBenchmark.scrape:gc.time                                   500  avgt    5     33.000                ms
```

</p>
</details>

implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/MicrometerCollector.java
shakuzen: @m-burst were you interested in making a pull request for this? Otherwise we can open an issue so we don't lose track of this.

m-burst: @shakuzen I have opened #7838 with this improvement, please take a look

PR #7814, opened 2026-08-09
Use OTel Metrics SDK and OTLP Exporter with shading

### LINKED ISSUE #5658: Replace the OpenTelemetry protobuf library

shakuzen:
Switch from the opentelemetry-proto dependency to shading the OTel OTLP exporter and Metrics SDK parts we need for publishing metrics in OTLP protobuf format.

For follow-up: we are still using some internal API where a non-internal factory method was not available for MetricData or HistogramDataPoint that we needed.

We use `OtlpHttpMetricExporter` internally with a custom `HttpSender` that delegates to our `OtlpMetricsSender`. This is more abstraction than we had before but the serialization API is marked internal and this was the only supported way around that.

As noted in PR comments, the toString of Request is now of limited usefulness without a serialization API available.

Because we shade these dependencies, we will need to make an exception to our general dependency upgrade policy to regularly upgrade to new minor versions of these dependencies even in our maintenance branches (where we would normally only upgrade to new patch versions). As far as I understand, OTel does not make patch releases for anything except the latest minor version and they have regular monthly new minor releases. If there were a CVE or any bug in code we are relying on here, because it is shaded, end-users cannot upgrade the version of the dependencies themselves. So it is important we keep them up-to-date, and due to the OTel support policy, the only way to do that is to upgrade to new minor versions regularly. That perhaps puts more risk around using things marked internal.

Closes #5658


implementations/micrometer-registry-otlp/build.gradle
```diff
@@ -1,9 +1,19 @@
...
+
 description = 'Publishes Micrometer meters in OTLP format'
 
 dependencies {
     api project(':micrometer-core')
 
-    implementation libs.openTelemetry.proto
+    implementation platform(libs.openTelemetry.bom)
+    implementation('io.opentelemetry:opentelemetry-exporter-otlp-common') {
+        exclude group: 'io.opentelemetry', module: 'opentelemetry-sdk-extension-autoconfigure-spi'
```
shakuzen, line 19: `opentelemetry-sdk-extension-autoconfigure-spi` pulls in `opentelemetry-sdk` which pulls in logs and tracing, which we don't need. This exclusion keeps the dependencies minimal.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricsSender.java
```diff
@@ -93,12 +95,10 @@ public String toString() {
         }
 
         private String readableMetricsData() {
-            try {
-                return ExportMetricsServiceRequest.parseFrom(metricsData).toString();
-            }
-            catch (InvalidProtocolBufferException e) {
-                return new String(metricsData, StandardCharsets.UTF_8);
+            if (this.readableMetricsDataSupplier != null) {
+                return this.readableMetricsDataSupplier.get();
```
shakuzen: This is only needed because `Request.toString` tries to include readable output of the metrics data. I don't know if that was a good idea for various reasons. We don't use the toString in any production code, but since `Request` is public, any third-party code could be using it. If we didn't include it, we could avoid the headache of needing a way to print the request payload. I don't know whether we want to solve that in this PR or separately, but all this added code around this is distracting from the main purpose of the PR.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
```diff
@@ -172,157 +179,139 @@ private static Optional<ExponentialHistogramSnapShot> getExponentialHistogramSna
...
+            }
+        }
+        else {
+            if (isDelta()) {
+                point = ImmutableHistogramPointData.create(startTimeNanos, getTimeUnixNano(), attributes, total, false,
+                        0.0, true, max, explicitBounds, bucketCounts, exemplars);
+            }
+            else {
+                point = ImmutableHistogramPointData.create(startTimeNanos, getTimeUnixNano(), attributes, total, false,
+                        0.0, false, 0.0, explicitBounds, bucketCounts, exemplars);
```
shakuzen: `ImmutableHistogramPointData` is internal. There is non-internal `HistogramPointData.create` API but it does not take exemplars, which is why we're using `ImmutableHistogramPointData` here. We can either take the risk of using internal API or we can make our own implementation of `HistogramPointData`.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
```diff
@@ -337,82 +326,97 @@ private boolean isDelta() {
...
+                io.opentelemetry.sdk.metrics.data.AggregationTemporality temporality) {
+            String unit = meta.baseUnit != null ? meta.baseUnit : "";
+            String description = meta.description != null ? meta.description : "";
+
+            switch (meta.metricType) {
+                case DOUBLE_GAUGE:
+                    if (doublePoints.isEmpty())
+                        return null;
+                    return ImmutableMetricData.createDoubleGauge(resource, scope, meta.name, description, unit,
+                            GaugeData.createDoubleGaugeData(doublePoints));
```
shakuzen, line 377: `ImmutableMetricData` is internal. We could make our own implementation if we don't want the drift risk of using something in an internal package.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -173,32 +178,39 @@ public void stop() {
...
-                    .compressionMode(config.compressionMode())
-                    .build());
-            }
-            catch (Exception e) {
-                logger.warn(String.format("Failed to publish metrics to OTLP receiver (context: %s)",
-                        getConfigurationContext()), e);
+            Collection<MetricData> metrics = otlpMetricConverter.getAllMetrics();
+            if (!metrics.isEmpty()) {
+                try {
+                    MetricsRequestMarshaler marshaler = MetricsRequestMarshaler.create(metrics);
```
shakuzen: `MetricsRequestMarshaler` is internal. Serialization by itself is not part of public API. Either we take the risk of using internal code, or we go the route of adapting OTel's HttpSender interface, which provides a non-internal API route at the cost of more indirection. We would probably need to think about how the use case of configuring a gRPC sender works in that case.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpHttpMetricsSenderTests.java
```diff
@@ -78,9 +78,8 @@ void toStringOfRequestShouldBeHumanReadable() throws Exception {
 
         verify(metricsSender).send(assertArg(request -> assertThat(request.toString()).startsWith(
                 "OtlpMetricsSender.Request for address: http://localhost:4318/v1/metrics, headers: {test-key=test-value}, compressionMode: NONE, metricsData:")
-            .contains("name: \"test.counter\"")
-            .contains("as_double: 1.0")
-            .contains("aggregation_temporality: AGGREGATION_TEMPORALITY_CUMULATIVE")));
+            .contains("test.counter")
+            .contains("1.0")));
```
shakuzen: This test depends on the `toString` of `Request`, and that depends on the readable printing mentioned in another comment. It doesn't represent the binary protobuf format that's actually sent, and with this PR it's now JSON instead of protobuf text format.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMetricConverter.java
```diff
@@ -337,82 +329,97 @@ private boolean isDelta() {
...
+
+    private static class MetricPointCollector {
+
+        final List<DoublePointData> doublePoints = new ArrayList<>();
+
+        final List<HistogramPointData> histogramPoints = new ArrayList<>();
+
+        final List<ExponentialHistogramPointData> exponentialHistogramPoints = new ArrayList<>();
+
+        final List<SummaryPointData> summaryPoints = new ArrayList<>();
```
shakuzen, line 365: It's not efficient to instantiate all these when we know they won't all be used, but I think we should streamline the review of this PR and follow-up separately with improving efficiency by focusing on ensuring the functionality is equivalent.

implementations/micrometer-registry-otlp/src/test/java/io/micrometer/registry/otlp/OtlpMeterRegistryTest.java
```diff
@@ -977,91 +1025,31 @@ void testZeroCountForExponentialHistogram() {
...
-        clock.add(exponentialHistogramOtlpConfig().step());
-
-        assertThat(writeToMetrics(timerWithDefaultZeroThreshold)).filteredOn(Metric::hasExponentialHistogram)
-            .singleElement()
-            .satisfies(exponentialHistogram -> {
-                ExponentialHistogramDataPoint dataPoint = exponentialHistogram.getExponentialHistogram()
-                    .getDataPoints(0);
-                // Default minimum expected value for Timer is 1ms. Base unit is
-                // milliseconds.
-                assertThat(dataPoint.getZeroThreshold()).isEqualTo(Math.nextDown(1.0));
```
shakuzen, line 1036: We don't have the API to check the zero threshold after the change in dependencies. We still have a test above this for the zero count, though.

implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpMeterRegistry.java
```diff
@@ -173,28 +179,35 @@ public void stop() {
...
-                        .setResource(this.resource)
-                        .addScopeMetrics(ScopeMetrics.newBuilder()
-                            // we don't have instrumentation library/version
-                            // attached to meters; leave unknown for now
-                            // .setScope(InstrumentationScope.newBuilder().setName("").setVersion("").build())
-                            .addAllMetrics(otlpMetricConverter.getAllMetrics())
-                            .build())
-                        .build())
-                    .build();
-                logger.trace("Request: {}", request);
```
shakuzen, line 192: I forgot to include this trace log, but the output would be different now anyway (JSON vs protobuf text format). And it's a bit of a pain to pass around how to serialize the data to a text format. I don't know what we want to do about this.

PR #7816, opened 2026-08-10
Store AbstractTimer pause detector as Object for optional LatencyUtils

### LINKED ISSUE #7815: ClassNotFoundException: org.LatencyUtils.PauseDetector is thrown when introspecting AbstractTimer
See https://github.com/spring-projects/spring-boot/issues/51322

#6883 made `org.LatencyUtils` optional but the `pauseDetector` field is of type `org.LatencyUtils.PauseDetector` which means that reflection fails. I suspect creating an instance of the class will also fail if LatencyUtils is not present.

I suspect you'll need to make `pauseDetector` an `Object` and cast it when used.

kalayciburak:
## Statement of purpose

Fixes #7815.

When optional LatencyUtils is not on the classpath, Spring reflective introspection of `AbstractTimer` fails because the `pauseDetector` field is typed as `org.LatencyUtils.PauseDetector`. That field type forces class resolution even when pause detection is disabled (the default).

## Changes

- Store the pause detector value as `Object` and cast only on the LatencyUtils-enabled path
- Add a regression test under `@ClassPathExclusions("LatencyUtils-*.jar")` that resolves all `AbstractTimer` field types

## How to verify

```bash
./gradlew :micrometer-core:test --tests io.micrometer.core.instrument.MissingLatencyUtilsTest --tests io.micrometer.core.instrument.AbstractTimerTests
```

Executed locally: both test classes GREEN.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
@@ -52,7 +52,8 @@ public abstract class AbstractTimer extends AbstractMeter implements Timer {
     // Only used when pause detection is enabled
     private @Nullable Object intervalEstimator;
 
-    private org.LatencyUtils.@Nullable PauseDetector pauseDetector;
+    // Object so field resolution does not require optional LatencyUtils on the classpath
+    private @Nullable Object pauseDetector;
 
     /**
      * Creates a new timer.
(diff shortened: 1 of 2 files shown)


shakuzen (approved):
Thanks for the PR. The changes look good. Would you rebase and target the PR to `1.17.x`? We'll merge forward from there.

kalayciburak:
rebased onto 1.17.x and retargeted the base — thanks

PR #7812, opened 2026-08-08
fix(prometheus): capture custom meter creation timestamps at registration

arimu1:
## Description

Custom Prometheus meters (registered via `Meter.builder` / `newMeter`) were computing `_created` timestamps with `clock.wallTime()` inside `customCounterFamily`, which runs on every scrape. That made creation timestamps advance with scrape time instead of reflecting actual meter registration time.

Other meter types already capture `createdTimestampMillis` once at registration (outside the scrape lambda) and reuse it. This change does the same for custom meters: capture once in `newMeter` and pass the value through `customMeterFamily` → `customCounterFamily`.

Fixes #7798

## Related work

#7802 also notes fixing custom-meter `_created` as part of a broader `MicrometerCollector` / `PrometheusMeterRegistry` refactor. This PR is a focused fix for the bug only, so it can land independently (and is easier to backport toward 1.16.7 if desired).

## Testing

- Strengthened `PrometheusMeterRegistryTest#createdTimestampEnabled` (`@Issue("gh-7798")`): advances the mock clock by 10s after registration, asserts COUNT and TOTAL custom-meter `_created` stay at registration time (5.001), and asserts they do not report scrape-time 15.001.
- `./gradlew :micrometer-registry-prometheus:test --tests 'io.micrometer.prometheusmetrics.PrometheusMeterRegistryTest'`
- `./gradlew :micrometer-registry-prometheus:checkFormat`

JDK 21 (Temurin 21.0.12).

PR #7817, opened 2026-08-10
Fix/issue 7798 prometheus custom meter timestamp

yashsiwacha:
This PR addresses an issue in `PrometheusMeterRegistry` where the creation timestamp for custom counters was being incorrectly recomputed on every scrape event. This caused Prometheus to receive dynamic, changing `_created` timestamps for the same metric, which breaks the OpenMetrics standard and can cause data to be dropped.


jonatan-ivanov:
Duplicate of https://github.com/micrometer-metrics/micrometer/pull/7812

PR #7837, opened 2026-08-17
Document runtime statistics behavior in `JCacheMetrics`

jewoodev:
**Summary**

Follow-up to gh-7562.

Move the JCache runtime statistics limitation from an implementation comment to the `JCacheMetrics` class Javadoc. Clarify that changes made through `CacheManager#enableStatistics(...)` are not tracked and that enabling statistics later requires creating and binding a new `JCacheMetrics` instance.

This is a documentation-only follow-up; runtime behavior is unchanged.

**Testing**

- `./gradlew :micrometer-core:javadoc`


shakuzen (approved):
Looks good to me. Thank you. We will merge after the current round of releases are out (scheduled for this week).

PR #7507, opened 2026-05-20
Make `DefaultLongTaskTimer.takeSnapshot` handle multiple percentiles per active task duration

schiemon:
Fixes #3877.

The percentile interpolation loop in `DefaultLongTaskTimer.takeSnapshot()` advances at most one percentile per active-task duration:

https://github.com/micrometer-metrics/micrometer/blob/3c6fe3d9dd994b5ebe3f136e858078393927b37e/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/DefaultLongTaskTimer.java#L193-L219

However, multiple configured percentiles can map to the same active-task duration. For example, with one active task and percentiles `[0.1, 0.25, 0.5]`, all three percentile ranks (`0.2`, `0.5`, `1.0`) are satisfied once `count = 1`.

The existing loop recorded only `0.1` and omitted the remaining percentiles because there was no next duration to advance to. As a result, the snapshot contains fewer percentile values than the percentiles configured for the `DefaultLongTaskTimer`. This can trip an `ArrayIndexOutOfBoundsException` in `HistogramGauges`, which expects the snapshot percentiles to match the configured percentiles:

  https://github.com/micrometer-metrics/micrometer/blob/3c6fe3d9dd994b5ebe3f136e858078393927b37e/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/HistogramGauges.java#L113-L120

**Fix:** change the `if` to a `while` so every percentile satisfied by the current `count` is recorded before advancing to the next duration.

Issue #3993, opened 2023-07-24
Poor performance of MicrometerCollector.collect()


Hartigan:
Poor performance of `MicrometerCollector.collect()` which uses `DefaultLongTaskTimer`. `DefaultLongTaskTimer` contains a lot of calls of `ConcurrentLinkedDeque.size()` with `O(n)` complexity during histogram creation.

Micrometer version: 1.10.3

Might be worth saving the size of deque or full snapshot at the beginning of the method `DefaultLongTaskTimer.takeSnapshot()` execution. Also method may contain data races between histogram creation and updates of `activeTasks`.

jonatan-ivanov:
Thank you for the issue!

Do you have benchmarks that shows the bottleneck or is this a theoretical issue?
I would assume that you will have issues because of the huge amount of parallel tasks before you would see performance degradations because of `size()`. This assumption might be incorrect, some benchmarks can demonstrate the correctness of it.

>Also method may contain data races between histogram creation and updates of activeTasks.

Could you please elaborate this? Which method and where can race condition arise?

Hartigan:
True, I have service with huge amount of parallel requests and it's expected load (not issue). And `ConcurrentLinkedDeque.size()` consumes 8% of flamegraph -- actual issue. 
<img width="713" alt="Screenshot 2023-07-25 at 10 19 15" src="https://github.com/micrometer-metrics/micrometer/assets/3034112/ede82a58-0a06-4e0f-90e3-a563f66aff12">

I calculate the algorithmic complexity of this class:
`DefaultLongTaskTimer` store task in `ConcurrentLinkedDeque` and provide methods `start()` with `O(1) for append task into deque`, `SampleImpl.stop()` -- `O(n) for removing random element from deque` and `takeSnapshot()` -- `O(n^2) for iterating over deque with size() calls on each iteration`.

> Could you please elaborate this? Which method and where can race condition arise?

It's theoretical, but you use `ConcurrentLinkedDeque`. This implies that timers can be added and removed while the histogram is being calculated(calls of `start()/SampleImpl.stop()` during call of `takeSnapshot()`).

Hartigan:
I created draft pull request https://github.com/micrometer-metrics/micrometer/pull/4001

Hartigan:
[...]

PR #7848, opened 2026-08-18
Fix histogram gauges for changing snapshots

PongPong:
## Summary

Make synthetic histogram gauges resilient when a `HistogramSupport` implementation returns snapshots whose percentile or bucket arrays change shape between publish cycles.

## Context

`HistogramGauges` registers callbacks from an initial `HistogramSnapshot`. Each callback currently captures its array index and uses that same index in every later snapshot.

For example, if the initial percentile array is `[0.50, 0.75, 0.99]` and a later snapshot is `[0.50, 0.99]`:

- the `0.75` gauge reads the `0.99` value under the wrong label;
- the `0.99` gauge reads index `2` and throws `ArrayIndexOutOfBoundsException`.

This is a defensive follow-up to #3877 and #7507. That change fixed a deterministic omission in `DefaultLongTaskTimer.takeSnapshot()`, but a long-task timer can still change while a snapshot is being built. The snapshot consistency concerns are also related to #3993.

## Changes

- Capture the registered percentile or bucket boundary instead of its initial array index.
- Resolve each gauge value by that stable key in the current snapshot.
- Use the current snapshot maximum when a registered percentile is temporarily absent, consistent with long-task timer percentiles above the interpolatable range.
- Return `NaN` when a registered bucket is absent instead of reading a different bucket.
- Treat all positive-infinity bucket representations as the same boundary.
- Synchronize only snapshot rollover and poll accounting so concurrent callbacks use a coherent snapshot cycle; value conversion remains outside the synchronized section.
- Preserve the existing behavior of taking one snapshot for a complete gauge polling cycle.

## Tests

Added deterministic regression coverage for:

- a shortened percentile snapshot, including stable label/value association and maximum fallback;
- a shortened bucket snapshot, including a missing bucket and positive infinity;
- one shared snapshot per complete gauge polling cycle.

Validated with the repository wrapper on JDK 17, as supported by `CONTRIBUTING.md`:

```text
./gradlew :micrometer-core:test --tests io.micrometer.core.instrument.distribution.HistogramGaugesTest
./gradlew :micrometer-core:check
```

Both pass. `:micrometer-core:check` covers formatting, license headers, Checkstyle, binary compatibility, module checks, and the complete core test suite.

The commit includes the required DCO `Signed-off-by` trailer.

Draft for maintainer and human review.

### DIFF
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/HistogramGauges.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/HistogramGauges.java
@@ -111,12 +111,17 @@ public class HistogramGauges {
         this.polledGaugesLatch = new CountDownLatch(0);
 
         for (int i = 0; i < valueAtPercentiles.length; i++) {
-            final int index = i;
+            ValueAtPercentile registeredPercentile = valueAtPercentiles[i];
 
             ToDoubleFunction<HistogramSupport> percentileValueFunction = m -> {
-                snapshotIfNecessary();
-                polledGaugesLatch.countDown();
-                return percentileValue.apply(snapshot.percentileValues()[index]);
+                HistogramSnapshot currentSnapshot = pollSnapshot();
+                for (ValueAtPercentile currentPercentile : currentSnapshot.percentileValues()) {
+                    if (Double.compare(currentPercentile.percentile(), registeredPercentile.percentile()) == 0) {
+                        return percentileValue.apply(currentPercentile);
+                    }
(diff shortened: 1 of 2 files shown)


shakuzen:
Thanks for the pull request. I haven't looked at the details of the pull request yet, but we will need the DCO check passing before we could even consider merging anything. Could you take a look? All commits in the PR need to pass the DCO check.

PongPong (commented):
hi @shakuzen , May I ask you to review this?

PR #7851, opened 2026-08-19
Update MultiGauge Gauges instead of remove/add

### LINKED ISSUE #6851: Registering/Updating a MultiGauge while a Prometheus scrape occurs can lead to DuplicateLabelsException

shakuzen:
Previously, the corresponding Gauge for a MultiGauge Row was removed and re-registered if `overwrite` was `true`. This introduces a timing for the Gauge to be missed in publishing (or in the case of the Prometheus registry, to be read multiple times). Instead, update the backing object for the Gauge by introducing a RowState and tracking tracking it in the MultiGauge.

This fixes #6851, but due to the nature of the changes, I'm not sure we should apply this to maintenance branches. #7508 has a fix for the Prometheus registry specifically that looks safer to apply to maintenance branches.

### DIFF
--- a/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
+++ b/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
@@ -977,7 +977,7 @@ class PrometheusMeterRegistryTest {
         MultiGauge multiGauge = MultiGauge.builder("my.metric").register(registry);
         List<MultiGauge.Row<?>> rows = IntStream.range(0, 100)
             .mapToObj(i -> MultiGauge.Row
-                .of(Tags.of("tag1", "abc", "poll_count", Integer.toString(i), "some_other_tag", "cde"), i))
+                .of(Tags.of("tag1", "abc", "poll.count", Integer.toString(i), "some.other.tag", "cde"), i))
             .collect(toList());
 
         multiGauge.register(rows, true);
--- a/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
+++ b/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
@@ -40,10 +40,12 @@ import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import static io.micrometer.core.instrument.MockClock.clock;
 import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
 import static org.assertj.core.api.Assertions.*;
 
 /**
@@ -967,6 +969,39 @@ class PrometheusMeterRegistryTest {
         assertThat(registry.get("another.gauge").gauge().value()).isEqualTo(2d);
 
         executorService.shutdownNow();
+    }
+
+    @Issue("#6851")
+    @Test
+    void scrapeWhileOverwritingMultiGaugeRowsDoesNotCreateDuplicateLabels() throws Exception {
+        MultiGauge multiGauge = MultiGauge.builder("my.metric").register(registry);
+        List<MultiGauge.Row<?>> rows = IntStream.range(0, 100)
+            .mapToObj(i -> MultiGauge.Row
+                .of(Tags.of("tag1", "abc", "poll_count", Integer.toString(i), "some_other_tag", "cde"), i))
+            .collect(toList());
+
+        multiGauge.register(rows, true);
+
+        AtomicBoolean stop = new AtomicBoolean();
+        ExecutorService executor = Executors.newSingleThreadExecutor();
+        Future<?> scraperTask = executor.submit(() -> {
+            while (!stop.get()) {
+                registry.scrape();
+            }
+        });
+
+        try {
+            for (int i = 1; i < 1_000; i++) {
+                multiGauge.register(rows, true);
+            }
+        }
+        finally {
+            stop.set(true);
+            executor.shutdown();
+            assertThat(executor.awaitTermination(5, TimeUnit.SECONDS)).isTrue();
+        }
+
+        scraperTask.get();
     }
 
     @Test
--- a/micrometer-core/src/test/java/io/micrometer/core/instrument/MultiGaugeTest.java
+++ b/micrometer-core/src/test/java/io/micrometer/core/instrument/MultiGaugeTest.java
@@ -15,15 +15,19 @@
  */
 package io.micrometer.core.instrument;
 
+import io.micrometer.core.Issue;
 import io.micrometer.core.instrument.MultiGauge.Row;
 import io.micrometer.core.instrument.config.MeterFilter;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import org.junit.jupiter.api.Test;
 
 import java.util.*;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import static java.util.stream.Collectors.toList;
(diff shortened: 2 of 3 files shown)


implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
```diff
@@ -969,6 +971,39 @@ void meterTriggeringAnotherMeterWhenCollectingValue() {
         executorService.shutdownNow();
     }
 
+    @Issue("#6851")
+    @Test
+    void scrapeWhileOverwritingMultiGaugeRowsDoesNotCreateDuplicateLabels() throws Exception {
+        MultiGauge multiGauge = MultiGauge.builder("my.metric").register(registry);
+        List<MultiGauge.Row<?>> rows = IntStream.range(0, 100)
+            .mapToObj(i -> MultiGauge.Row
+                .of(Tags.of("tag1", "abc", "poll_count", Integer.toString(i), "some_other_tag", "cde"), i))
```
jonatan-ivanov:
```suggestion
                .of(Tags.of("tag1", "abc", "poll.count", Integer.toString(i), "some.other.tag", "cde"), i))
```

micrometer-core/src/test/java/io/micrometer/core/instrument/MultiGaugeTest.java
```diff
@@ -184,6 +188,84 @@ void withMeterFilterIgnoreTags() {
...
+            List.of(
+                Row.of(Tags.of("key", "2", "ignored", "4"), 4d),
+                Row.of(Tags.of("key", "2", "ignored", "5"), 5d),
+                Row.of(Tags.of("key", "2", "ignored", "6"), 6d)
+            ),
+            true
+        );
+        // @formatter:on
+        assertThat(registry.get("mg").tag("key", "2").gauges()).hasSize(1);
+        assertThat(registry.get("mg").tag("key", "2").gauge().value()).isEqualTo(6d);
```
shakuzen, line 209: I don't think this behavior is specified (we have no JavaDoc so practically nothing is specified). I think this should be a separate test. The other assertions in this test are about single row register calls. This seems to be about multiple rows that will be equivalent after the filter. We should improve the specification for MultiGauge (and remove Incubating since I think the ship has sailed on not supporting it), but I think that's a separate issue to this PR. I guess it is nice the behavior happens to be the same before and after the main code changes.

shakuzen: We should also backport any relevant tests to maintenance branches if/when we document the behavior in a JavaDoc

jonatan-ivanov: We do have this: https://github.com/micrometer-metrics/micrometer/blob/225b0b8fb05680ae9b441ecc433061491d4b9a6c/micrometer-core/src/main/java/io/micrometer/core/instrument/MultiGauge.java#L77-L81

I think we should call the effect of a `MeterFilter` out on this, as we did for other function-tracking meters:
>"Re-registering" a gauge can happen indirectly for example as the result of a MeterFilter modifying the name and/or the tags of two different gauges so that they will be the same after the filter is applied.

I also think that even if nothing would be specified, the scenario is definitely possible so I think the test is still valid since we probably want to know if the behavior changes.

But I agree, this test is definitely more connected to https://github.com/micrometer-metrics/micrometer/pull/6146 and https://github.com/micrometer-metrics/micrometer/pull/6070 than to this issue, I just discovered this scenario during reviewing this PR and and saw we did not have this test case.

At this point, I wanted to create a PR to add this test to `1.16.x` and here's the twist: it fails on `1.16.x`, `1.17.x`, and `main`. I think the current behavior on `main` is "wrong" and this PR "fixes" it. Let's discuss what to do (should we fix 1.16 and 1.17) on our next meeting. This also makes me wanting to add this test on top of these changes (in this PR or in a fast-follow one).

skrcode:
I ran JAIPilot Cloud against this exact PR head. It found one follow-up in `MultiGauge.register()`: avoid rebuilding a temporary `HashSet` for each registration while preserving row replacement and stale-gauge removal.

On a 200-row JMH workload using the same host and JDK 25, median time moved from 14.9035 to 8.4815 µs/op and allocation from 18,616 to 8,000 B/op. The focused 75-test suite and affected module builds passed.

Cloud-generated draft and full evidence: https://github.com/skrcode/micrometer/pull/2  
PR directly onto this source branch: https://github.com/shakuzen/micrometer/pull/1

This is optional and workload-qualified; feel free to merge or cherry-pick if it fits the direction of this PR.

shakuzen:
@skrcode thanks for taking a look. I think it will be best to consider that optimization after this PR as a follow-up. MultiGauge.register is not expected to be on the hot path, and the proposed changes I think make the code harder to follow, even if they're still correct and have better performance. So we'll have to consider the tradeoff.

micrometer-core/src/test/java/io/micrometer/core/instrument/MultiGaugeTest.java
shakuzen: I don't know how I missed the JavaDoc when I wrote my original comment. After reading the JavaDoc, I would still say the behavior for duplicates within a register call is unclear if not completely unspecified.

I went down the rabbit hole on this a bit. I think we should decide the right behavior going forward and since this PR actually changes the behavior, perhaps it should be specified along with this PR. I didn't want to block getting this merged on that since it wasn't the point of this PR originally, but I think changing the behavior without making it clear what the expected behavior is seems wrong too.

I think the problem with the behavior in maintenance branches is that it depends on the previous state. You get a different outcome depending on whether the old rows contain the same ID as the duplicates being registered. If the old rows don't contain it, you get first row wins and subsequent rows result in a failed gauge registration. If the old row does contain the ID, you get last row wins and a remove/re-register for each row.

With the changes in this PR, you get consistent behavior that only depends on what you pass for `overwrite`. It's unclear from the JavaDoc that `overwrite` *should* apply to **intra**-batch duplicates, as it was originally added for controlling the behavior of **inter**-batch duplicates, but I think it's a reasonable thing to do until we understand a reason for allowing different behavior.

I'm not sure how much effort we should put into changing anything related to this on maintenance branches until we at least have a reported problem about it. I would focus our efforts on figuring out what is right going forward, and then specifying and testing that.

shakuzen:
@jonatan-ivanov I have attempted to specify the behavior in the updated `register` JavaDoc. Would you take a look, please?

shakuzen:
I think a follow-up PR should look at removing the `@Incubating` annotation and fully checking test coverage and specification of behavior. I left that out-of-scope for this PR. I think we're missing some test cases and there might even be more decisions to make about what is the right behavior with edge cases.

jonatan-ivanov (approved):

Issue #7367, opened 2026-04-01
Document best practice about configuring registries on a composite


shakuzen:
This is the follow-up documentation part of https://github.com/micrometer-metrics/micrometer/issues/6908.

nick-montana:
@shakuzen

Hi. I am very much interested in actively contributing to Micrometer and trying to find an issue that could be a good starting point. I've been using this tool for a while now and can write decent enough documentation. I'll take up this issue if nobody is working on it already


**My initial Questions:**

- Does the best practices documentation specifically need attention to manual composite configuration (with code), property-based configuration, or both? 

- I've previously only configured two registries on the composite, do you ever see scenarios where devs add more? Seems like there would be memory or CPU considerations for exporting micrometer payloads to over two destinations

- Can you share any documentation-related acceptance criteria off the top of your head? 

Also want to tag @jonatan-ivanov to this after our brief conversation in slack.  

Thanks for taking the time to answer my questions... I know that the standard protocol for contributing is to just proactively find an issue and open a PR. I will try my best to get into that flow after finishing my first task

jonatan-ivanov:
Thank you for the offer I assigned the issue to you!

>Does the best practices documentation specifically need attention to manual composite configuration (with code), property-based configuration, or both?

Not sure if any code is needed here really, this should be a very simple doc update, it is only about writing down that meter registration should happen after the composite is configured (after registries are added), wether the user does this (code) or the framework (properties?) is not relevant I think.

>I've previously only configured two registries on the composite, do you ever see scenarios where devs add more? Seems like there would be memory or CPU considerations for exporting micrometer payloads to over two destinations

It does not matter, this can be an issue with a single registry:
1. create a composite
2. register a meter
3. add a single registry to the composite

>Can you share any documentation-related acceptance criteria off the top of your head?

- It should clearly align with the wording of https://github.com/micrometer-metrics/micrometer/pull/7322
- I think it should be in a `NOTE:` section
- `./gradlew antora` should generate the site and it should look ok
- Should be done on `1.15.x`

PR #7855, opened 2026-08-20
Improve CompositeMeterRegistry documentation with best practices

nick-montana:
add note section to composite meter registry documentation that adds additional context for appropriate use during configuration

Resolves: Git Issue #7367


nick-montana:
hey @jonatan-ivanov I worked on this way later than I originally planned for but better late than never. The doc update is simple but I think that is anticipated based off our previous conversation in the issue. Let me know if this works or if there are areas for improvement

nick-montana:
just realized i need to rebase to 1.16.X. opening a new PR

Issue #497, opened 2018-03-16
ElasticMeterRegistry reports NaN values to ES causing errors


tkp1n:
The `ElasticMeterRegistry` may report NaN for doubles which causes an error on Elasticsearch.

The culprit (most likely at least):
https://github.com/micrometer-metrics/micrometer/blob/3ec44d958736b0e4c6979479cbc2982fd491384b/implementations/micrometer-registry-elastic/src/main/java/io/micrometer/elastic/ElasticMeterRegistry.java#L310

The erorr as reported by Elasticsearch:
```json
"index": {
    "_index": "metrics-2018.03.15",
    "_type": "doc",
    "_id": "AWIqq3txcLB5uE05seul",
    "status": 400,
    "error": {
        "type": "mapper_parsing_exception",
        "reason": "failed to parse",
        "caused_by": {
            "type": "i_o_exception",
            "reason": "Non-standard token 'NaN': enable JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS to allow\n at [Source: org.elasticsearch.common.bytes.BytesReference$MarkSupportingStreamInputWrapper@32e12734; line: 1, column: 98]"
        }
    }
}
```

Possible solutions:
- Skip the meter if numeric value is not a number
- Report 0.0 instaed of NaN

/cc @jkschneider

jkschneider:
Gauge/TimeGauge are the only things that can report NaN. I guess we'll skip reporting them if they get to that value.

tkp1n:
I just retested your fixes to #497, #498, and #499.
So far everything is looking good! Thanks a lot

PR #7856, opened 2026-08-20
Make state objects non-nullable for function-based meter registration

### LINKED ISSUE #6416: State objects should not be nullable for function-based meter registration

ragulshanmugam:
Implements option 2 from #6416, per team's direction.

Annotation-only. no runtime requireNonNull, so callers passing null today behave exactly as before. 

the `@Contract on MeterRegistry.gauge(...)` came from #6546, so removing it partially reverts that.

The three tests that pass null are marked @NullUnmarked rather than deleted, since the implementations still tolerate null at runtime. The TCK one is renamed to nullSourceObjectReportsNaN — it never garbage-collected anything, it just passed a literal null, and a real GC test would be non-deterministic.

`./gradlew check` also fails on unmodified main here, in an unrelated class, so the local suite looks flaky under parallel execution.

fixes #6416

### DIFF
--- a/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticMeterRegistryTest.java
+++ b/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticMeterRegistryTest.java
@@ -18,7 +18,6 @@ package io.micrometer.elastic;
 import io.micrometer.core.Issue;
 import io.micrometer.core.instrument.*;
 import org.jspecify.annotations.Nullable;
-import org.jspecify.annotations.NullUnmarked;
 import org.junit.jupiter.api.Test;
 
 import java.time.Instant;
@@ -165,7 +164,7 @@ class ElasticMeterRegistryTest {
 
     @Issue("#497")
     @Test
-    @NullUnmarked
+    @SuppressWarnings("NullAway")
     void nullGauge() {
         Gauge g = Gauge.builder("gauge", null, o -> 1).register(registry);
         assertThat(registry.writeGauge(g)).isNotPresent();
--- a/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticMeterRegistryTest.java
+++ b/implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticMeterRegistryTest.java
@@ -18,6 +18,7 @@ package io.micrometer.elastic;
 import io.micrometer.core.Issue;
 import io.micrometer.core.instrument.*;
 import org.jspecify.annotations.Nullable;
+import org.jspecify.annotations.NullUnmarked;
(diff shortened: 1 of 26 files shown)


jonatan-ivanov (approved):

jonatan-ivanov:
Thank you very much for the PR!
It looks good, I polished it a little bit and documented what to do in case of nullable state objects.

implementations/micrometer-registry-elastic/src/test/java/io/micrometer/elastic/ElasticMeterRegistryTest.java
```diff
@@ -164,6 +164,7 @@ void writeTags() {
 
     @Issue("#497")
     @Test
+    @SuppressWarnings("NullAway")
```
shakuzen, line 167: This test was added in resolution of https://github.com/micrometer-metrics/micrometer/issues/497. I think the test can be updated so the ToDoubleFunction returns `NaN`.

micrometer-test/src/main/java/io/micrometer/core/tck/MeterRegistryCompatibilityKit.java
```diff
@@ -429,8 +428,9 @@ void mapSizeGauge() {
 
         @Test
         @DisplayName("gauges that reference an object that is garbage collected report NaN")
+        @SuppressWarnings("NullAway")
         void garbageCollectedSourceObject() {
-            registry.gauge("my.gauge", emptyList(), (Map) null, Map::size);
+            registry.gauge("my.gauge", emptyList(), (Map<?, ?>) null, Map::size);
```
shakuzen, line 433: This test implementation is a bit dubious, but that's an existing issue.

micrometer-core/src/test/java/io/micrometer/core/instrument/dropwizard/DropwizardMeterRegistryTest.java
```diff
@@ -60,6 +60,7 @@ protected Double nullGaugeValue() {
     };
 
     @Test
+    @SuppressWarnings("NullAway")
     void gaugeOnNullValue() {
         registry.gauge("gauge", emptyList(), null, obj -> 1.0);
         assertThat(registry.get("gauge").gauge().value()).isNaN();
```
shakuzen, line 66: Likewise, this test is not the best implementation of a state object being garbage collected, which is the scenario it is trying to simulate. But this is also an existing issue.

shakuzen (commented):
I'll follow-up with a PR based on the things I pointed out. Good work on this everyone. Thanks.

PR #7866, opened 2026-08-23
fix: sanitize carriage returns in Influx tag values

### LINKED ISSUE #7865: Sanitize carriage return in InfluxNamingConvention
Influx metrics are serialized using line protocol, where line terminators delimit metric records. If an application supplies a metric name or tag key containing \r or \n, the value is currently carried into the serialized line protocol and can alter record boundaries.

This is primarily a defensive-correctness issue: standard Micrometer binders use fixed metric names and tag keys, so remote exploitability depends on an application or custom extension deriving those values from untrusted input.

_Proposed fix_: replace both carriage return and line feed with spaces in all three Influx naming-convention methods before applying normal line-protocol escaping.

ThanosTsiamis:
This change extends the existing Influx tag-value sanitization to replace carriage returns (\r) with spaces, matching the current handling of line feeds (\n).
Meter names and tag keys are unchanged.
Adds a regression test covering carriage returns in tag values.
Closes #7865

### DIFF
--- a/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxNamingConvention.java
+++ b/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxNamingConvention.java
@@ -62,7 +62,7 @@ public class InfluxNamingConvention implements NamingConvention {
 
     @Override
     public String tagValue(String value) {
-        return escape(this.delegate.tagValue(value).replace('\n', ' '));
+        return escape(this.delegate.tagValue(value).replace('\r', ' ').replace('\n', ' '));
     }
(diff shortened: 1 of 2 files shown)


ThanosTsiamis:
@jonatan-ivanov Any thoughts on the updated PR (based on your comments in the issue)?

jonatan-ivanov:
Thank you for your PR, we decided to document this instead, see: https://github.com/micrometer-metrics/micrometer/issues/7865#issuecomment-5498237643

jonatan-ivanov:
Superseded by https://github.com/micrometer-metrics/micrometer/issues/7886

PR #7880, opened 2026-08-25
Enable NullAway JSpecifyExperimental and fix nullability issues

shakuzen:
Enable the `NullAway:JSpecifyMode` and `NullAway:JSpecifyExperimental` flags (on JDK 22+) and resolve all resulting NullAway errors across the codebase.

With NullAway 0.14.0 and `JSpecifyExperimental` enabled, NullAway activates stricter JSpecify enforcement, better wildcard support, and standard library nullability models from `jspecify/jdk`.

The `NullAway:JSpecifyMode` and `NullAway:JSpecifyExperimental` flags are restricted to JDK 22 and higher (`canCompileOrRun(22)`) because on JDK 21, `javac` has bytecode type-use annotation limitations ([JDK-8346471](https://bugs.openjdk.org/browse/JDK-8346471)). Restricting JSpecify checking to JDK 22+ ensures all nullability checks run cleanly without needing source code workarounds or JDK 21 compiler flags.

Key areas of note for review:
1. **Public API and Generics adjustments**:
   - `TimedExecutorService`, `TimedScheduledExecutorService`, `TimedCallable`: Added `<T extends @Nullable Object>` bounds on generic task and future methods, allowing tasks that return `null` (matching standard `ExecutorService` under JSpecify).
   - `ObservationPredicate`: First parameter is now `@Nullable String` (`BiPredicate<@Nullable String, Context>`), as convention names can be null before resolution.
   - `ResponseContext`: Updated `void setResponse(@Nullable RES response)` to accept `@Nullable RES` since client/server transport errors leave the response object null.
   - `LongTaskTimer.Builder`: Fixed `serviceLevelObjectives(Duration @Nullable ... slos)` (nullable array of non-null durations) to align with `AbstractTimerBuilder` and `Timer.Builder`.
   - `StatsdMeterRegistry.Builder`: Parameter type for `lineBuilder` changed to `BiFunction<Meter.Id, @Nullable DistributionStatisticConfig, StatsdLineBuilder>` since non-distribution meters do not supply a configuration.
2. **Null handling / logic improvements**:
   - `AbstractTimer`: Refactored pause detector initialization to avoid returning `null` in `computeIfAbsent`.
   - `CloudWatchMeterRegistry`: Used a null-safe builder helper for `Stream<MetricDatum>` to safely drop `NaN` metric values.
   - `TimedAspect` and `CountedAspect`: Added null check when looking up annotations before invoking join points.
   - `InfluxApiVersion`: Safely handled optional `config.org()` when building Influx V2 write endpoints.
3. **Suppressions**:
   - Added `@SuppressWarnings("NullAway")` for `MBeanServer.addNotificationListener` calls passing `null` handback objects (`CommonsObjectPool2Metrics`, `TomcatMetrics`, `KafkaConsumerMetrics`), and `Arrays.copyOf` array expansions (`MetricsDSLContext`, `NewRelicInsightsApiClientProvider`).

### DIFF
--- a/implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java
@@ -474,7 +474,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
                 }
             }
 
-            @Nullable Exemplar lastExemplar = lastExemplarSupplier.get();
+            Exemplar lastExemplar = lastExemplarSupplier.get();
             Collector.Type type = distributionStatisticConfig.isPublishingHistogram() ? Collector.Type.HISTOGRAM
                     : Collector.Type.SUMMARY;
             if (histogramCounts.length > 0) {
@@ -489,7 +489,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
                     case Prometheus:
                         histogramKeys.add("le");
 
-                        Exemplar @Nullable [] exemplars = histogramExemplarsSupplier.get();
+                        Exemplar[] exemplars = histogramExemplarsSupplier.get();
 
                         // satisfies
                         // https://prometheus.io/docs/concepts/metric_types/#histogram
--- a/implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java
+++ b/implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java
@@ -449,8 +449,10 @@ public class PrometheusMeterRegistry extends MeterRegistry {
     }
 
     private void addDistributionStatisticSamples(DistributionStatisticConfig distributionStatisticConfig,
-            MicrometerCollector collector, HistogramSupport histogramSupport, Supplier<Exemplar> lastExemplarSupplier,
-            Supplier<Exemplar[]> histogramExemplarsSupplier, List<String> tagValues, boolean forLongTaskTimer) {
+            MicrometerCollector collector, HistogramSupport histogramSupport,
+            Supplier<@Nullable Exemplar> lastExemplarSupplier,
+            Supplier<Exemplar @Nullable []> histogramExemplarsSupplier, List<String> tagValues,
+            boolean forLongTaskTimer) {
         collector.add(tagValues, (conventionName, tagKeys) -> {
             Stream.Builder<Collector.MetricFamilySamples.Sample> samples = Stream.builder();
 
@@ -472,7 +474,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
                 }
             }
 
-            Exemplar lastExemplar = lastExemplarSupplier.get();
+            @Nullable Exemplar lastExemplar = lastExemplarSupplier.get();
             Collector.Type type = distributionStatisticConfig.isPublishingHistogram() ? Collector.Type.HISTOGRAM
                     : Collector.Type.SUMMARY;
             if (histogramCounts.length > 0) {
@@ -487,7 +489,7 @@ public class PrometheusMeterRegistry extends MeterRegistry {
                     case Prometheus:
                         histogramKeys.add("le");
 
-                        Exemplar[] exemplars = histogramExemplarsSupplier.get();
+                        Exemplar @Nullable [] exemplars = histogramExemplarsSupplier.get();
 
                         // satisfies
                         // https://prometheus.io/docs/concepts/metric_types/#histogram
--- a/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
+++ b/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
@@ -511,13 +511,12 @@ class PrometheusMeterRegistryTest {
     }
 
     private Condition<Iterable<? extends MetricSnapshot>> withNameAndQuantile(String name) {
-        return new Condition<>(
-                metricSnapshots -> ((MetricSnapshots) metricSnapshots).stream()
-                    .filter(snapshot -> snapshot.getMetadata().getPrometheusName().equals(name))
-                    .flatMap(snapshot -> snapshot.getDataPoints().stream())
-                    .filter(SummarySnapshot.SummaryDataPointSnapshot.class::isInstance)
-                    .map(SummarySnapshot.SummaryDataPointSnapshot.class::cast)
-                    .anyMatch(summaryDataPoint -> summaryDataPoint.getQuantiles().size() > 0),
+        return new Condition<>(metricSnapshots -> ((MetricSnapshots) metricSnapshots).stream()
+            .filter(snapshot -> snapshot.getMetadata().getPrometheusName().equals(name))
+            .flatMap(snapshot -> snapshot.getDataPoints().stream())
+            .filter(SummarySnapshot.SummaryDataPointSnapshot.class::isInstance)
+            .map(SummarySnapshot.SummaryDataPointSnapshot.class::cast)
+            .anyMatch(summaryDataPoint -> summaryDataPoint != null && summaryDataPoint.getQuantiles().size() > 0),
                 "a summary with name `%s` and at least one quantile", name);
     }
 
@@ -1200,7 +1199,10 @@ class PrometheusMeterRegistryTest {
             @Override
             public Properties prometheusProperties() {
                 Properties properties = new Properties();
-                properties.putAll(PrometheusConfig.super.prometheusProperties());
+                Properties defaultProperties = PrometheusConfig.super.prometheusProperties();
+                if (defaultProperties != null) {
+                    properties.putAll(defaultProperties);
+                }
                 properties.setProperty("io.prometheus.exporter.includeCreatedTimestamps", "true");
                 return properties;
             }
@@ -1247,7 +1249,10 @@ class PrometheusMeterRegistryTest {
             @Override
             public Properties prometheusProperties() {
                 Properties properties = new Properties();
-                properties.putAll(PrometheusConfig.super.prometheusProperties());
+                Properties defaultProperties = PrometheusConfig.super.prometheusProperties();
+                if (defaultProperties != null) {
+                    properties.putAll(defaultProperties);
+                }
                 properties.setProperty("io.prometheus.exporter.includeCreatedTimestamps", "false");
                 return properties;
             }
@@ -1346,7 +1351,10 @@ class PrometheusMeterRegistryTest {
             @Override
             public Properties prometheusProperties() {
                 Properties mergedProperties = new Properties();
-                mergedProperties.putAll(PrometheusConfig.super.prometheusProperties());
+                Properties defaultProperties = PrometheusConfig.super.prometheusProperties();
+                if (defaultProperties != null) {
+                    mergedProperties.putAll(defaultProperties);
+                }
                 properties.forEach((key, value) -> mergedProperties.setProperty(key.toString(), value.toString()));
                 return mergedProperties;
             }
--- a/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
+++ b/implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
@@ -514,6 +514,7 @@ class PrometheusMeterRegistryTest {
         return new Condition<>(metricSnapshots -> ((MetricSnapshots) metricSnapshots).stream()
             .filter(snapshot -> snapshot.getMetadata().getPrometheusName().equals(name))
             .flatMap(snapshot -> snapshot.getDataPoints().stream())
+            .filter(Objects::nonNull)
             .filter(SummarySnapshot.SummaryDataPointSnapshot.class::isInstance)
             .map(SummarySnapshot.SummaryDataPointSnapshot.class::cast)
             .anyMatch(summaryDataPoint -> summaryDataPoint != null && summaryDataPoint.getQuantiles().size() > 0),
--- a/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/UserAgentHeaderProvider.java
+++ b/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/UserAgentHeaderProvider.java
@@ -54,7 +54,8 @@ class UserAgentHeaderProvider implements HeaderProvider {
 
     private String computeUserAgent(String component) {
         String library = "micrometer-registry-" + component;
-        String version = getClass().getPackage().getImplementationVersion();
+        Package pkg = getClass().getPackage();
+        String version = pkg != null ? pkg.getImplementationVersion() : null;
 
         return "Micrometer/" + version + " " + library + "/" + version;
     }
--- a/implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
+++ b/implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
@@ -44,7 +44,8 @@ class StackdriverIntegrationTest {
     static final StackdriverConfig CONFIG = new StackdriverConfig() {
         @Override
         public String projectId() {
-            return System.getenv("GOOGLE_CLOUD_PROJECT_ID");
+            String projectId = System.getenv("GOOGLE_CLOUD_PROJECT_ID");
+            return projectId != null ? projectId : "projectId";
         }
 
         @Override
--- a/implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
+++ b/implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
@@ -26,6 +26,7 @@ import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import java.time.Duration;
+import java.util.Objects;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -44,7 +45,8 @@ class StackdriverIntegrationTest {
     static final StackdriverConfig CONFIG = new StackdriverConfig() {
         @Override
         public String projectId() {
-            return Objects.requireNonNull(System.getenv("GOOGLE_CLOUD_PROJECT_ID"), "GOOGLE_CLOUD_PROJECT_ID env var needs to be set!");
+            return Objects.requireNonNull(System.getenv("GOOGLE_CLOUD_PROJECT_ID"),
+                    "GOOGLE_CLOUD_PROJECT_ID env var needs to be set!");
         }
 
         @Override
--- a/implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
+++ b/implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
@@ -44,8 +44,7 @@ class StackdriverIntegrationTest {
     static final StackdriverConfig CONFIG = new StackdriverConfig() {
         @Override
         public String projectId() {
-            String projectId = System.getenv("GOOGLE_CLOUD_PROJECT_ID");
-            return projectId != null ? projectId : "projectId";
+            return Objects.requireNonNull(System.getenv("GOOGLE_CLOUD_PROJECT_ID"), "GOOGLE_CLOUD_PROJECT_ID env var needs to be set!");
         }
 
         @Override
--- a/micrometer-commons/src/main/java/io/micrometer/common/util/internal/logging/JdkLogger.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/util/internal/logging/JdkLogger.java
@@ -74,7 +74,7 @@ class JdkLogger extends AbstractInternalLogger {
     final transient Logger logger;
 
     JdkLogger(Logger logger) {
-        super(logger.getName() != null ? logger.getName() : "");
+        super(logger.getName() != null ? logger.getName() : "unnamed");
         this.logger = logger;
     }
 
--- a/micrometer-commons/src/main/java/io/micrometer/common/util/internal/logging/JdkLogger.java
+++ b/micrometer-commons/src/main/java/io/micrometer/common/util/internal/logging/JdkLogger.java
@@ -74,7 +74,7 @@ class JdkLogger extends AbstractInternalLogger {
     final transient Logger logger;
 
     JdkLogger(Logger logger) {
-        super(logger.getName());
+        super(logger.getName() != null ? logger.getName() : "");
         this.logger = logger;
     }
 
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -194,6 +194,9 @@ public class TimedAspect {
             declaringClass = pjp.getTarget().getClass();
         }
         Timed timed = declaringClass.getAnnotation(Timed.class);
+        if (timed == null) {
+            return pjp.proceed();
+        }
 
         return perform(pjp, timed, method);
     }
@@ -209,6 +212,9 @@ public class TimedAspect {
         if (timed == null) {
             method = pjp.getTarget().getClass().getMethod(method.getName(), method.getParameterTypes());
             timed = method.getAnnotation(Timed.class);
+        }
+        if (timed == null) {
+            return pjp.proceed();
         }
 
         return perform(pjp, timed, method);
--- a/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
@@ -212,9 +212,9 @@ public class TimedAspect {
         if (timed == null) {
             method = pjp.getTarget().getClass().getMethod(method.getName(), method.getParameterTypes());
             timed = method.getAnnotation(Timed.class);
-        }
-        if (timed == null) {
-            return pjp.proceed();
+            if (timed == null) {
+                return pjp.proceed();
+            }
         }
 
         return perform(pjp, timed, method);
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
@@ -147,15 +147,12 @@ public abstract class AbstractTimer extends AbstractMeter implements Timer {
                         e);
                 return;
             }
-        }
-        pauseDetector = pauseDetectorCache.computeIfAbsent(pauseDetectorType, detector -> {
-            if (detector instanceof ClockDriftPauseDetector) {
+            pauseDetector = pauseDetectorCache.computeIfAbsent(pauseDetectorType, detector -> {
                 ClockDriftPauseDetector clockDriftPauseDetector = (ClockDriftPauseDetector) detector;
                 return new SimplePauseDetector(clockDriftPauseDetector.getSleepInterval().toNanos(),
                         clockDriftPauseDetector.getPauseThreshold().toNanos(), 1, false);
-            }
-            return null;
-        });
+            });
+        }
 
         if (pauseDetector instanceof SimplePauseDetector) {
             org.LatencyUtils.PauseDetector pauseDetector = (org.LatencyUtils.PauseDetector) this.pauseDetector;
--- a/micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
+++ b/micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
@@ -22,6 +22,7 @@ import org.jspecify.annotations.Nullable;
 
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.Objects;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.function.*;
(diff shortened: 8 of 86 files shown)


implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java
```diff
@@ -472,7 +474,7 @@ private void addDistributionStatisticSamples(DistributionStatisticConfig distrib
                 }
             }
 
-            Exemplar lastExemplar = lastExemplarSupplier.get();
+            @Nullable Exemplar lastExemplar = lastExemplarSupplier.get();
```
jonatan-ivanov: Is this needed for local variables? (There is another one below this.)
I think `@Nullable` annotation on local variables is unnecessary (/redundant) since IDEs (and errorprone) can figure this out and warn us about this automatically (nullability is implicit for local variables).

implementations/micrometer-registry-prometheus/src/test/java/io/micrometer/prometheusmetrics/PrometheusMeterRegistryTest.java
```diff
@@ -509,13 +509,12 @@ void functionCounterNamingConvention() {
...
-                    .flatMap(snapshot -> snapshot.getDataPoints().stream())
-                    .filter(SummarySnapshot.SummaryDataPointSnapshot.class::isInstance)
-                    .map(SummarySnapshot.SummaryDataPointSnapshot.class::cast)
-                    .anyMatch(summaryDataPoint -> summaryDataPoint.getQuantiles().size() > 0),
+        return new Condition<>(metricSnapshots -> ((MetricSnapshots) metricSnapshots).stream()
+            .filter(snapshot -> snapshot.getMetadata().getPrometheusName().equals(name))
+            .flatMap(snapshot -> snapshot.getDataPoints().stream())
+            .filter(SummarySnapshot.SummaryDataPointSnapshot.class::isInstance)
+            .map(SummarySnapshot.SummaryDataPointSnapshot.class::cast)
+            .anyMatch(summaryDataPoint -> summaryDataPoint != null && summaryDataPoint.getQuantiles().size() > 0),
```
jonatan-ivanov, line 520: I think we can do the null check earlier which might be sparing the `instanceof` check and the cast but also, it makes the predicate in `.anyMatch` a bit shorter:
```suggestion
        return new Condition<>(metricSnapshots -> ((MetricSnapshots) metricSnapshots).stream()
            .filter(snapshot -> snapshot.getMetadata().getPrometheusName().equals(name))
            .flatMap(snapshot -> snapshot.getDataPoints().stream())
            .filter(Objects::nonNull)
            .filter(SummarySnapshot.SummaryDataPointSnapshot.class::isInstance)
            .map(SummarySnapshot.SummaryDataPointSnapshot.class::cast)
            .anyMatch(summaryDataPoint.getQuantiles().size() > 0),
```

implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/UserAgentHeaderProvider.java
```diff
@@ -54,7 +54,8 @@ public Map<String, String> getHeaders() {
 
     private String computeUserAgent(String component) {
         String library = "micrometer-registry-" + component;
-        String version = getClass().getPackage().getImplementationVersion();
+        Package pkg = getClass().getPackage();
+        String version = pkg != null ? pkg.getImplementationVersion() : null;
```
jonatan-ivanov, line 58: This is not the concern of this PR but I'm thinking if we should create an issue to add a utility class that exposes the Micrometer version since we are doing this throughout the codebase (Boot and Framework are doing this).

implementations/micrometer-registry-stackdriver/src/test/java/io/micrometer/stackdriver/StackdriverIntegrationTest.java
```diff
@@ -44,7 +44,8 @@ class StackdriverIntegrationTest {
     static final StackdriverConfig CONFIG = new StackdriverConfig() {
         @Override
         public String projectId() {
-            return System.getenv("GOOGLE_CLOUD_PROJECT_ID");
+            String projectId = System.getenv("GOOGLE_CLOUD_PROJECT_ID");
+            return projectId != null ? projectId : "projectId";
```
jonatan-ivanov: If the env var is not set, but we use `"projectId"` instead, this will fail later on except if there is a project called `projectId`. I think in both cases we should fail fast (without calling the GCP APIs):
```suggestion
            return Objects.requireNonNull(System.getenv("GOOGLE_CLOUD_PROJECT_ID"), "GOOGLE_CLOUD_PROJECT_ID env var needs to be set!");
```

micrometer-commons/src/main/java/io/micrometer/common/util/internal/logging/JdkLogger.java
```diff
@@ -74,7 +74,7 @@ class JdkLogger extends AbstractInternalLogger {
     final transient Logger logger;
 
     JdkLogger(Logger logger) {
-        super(logger.getName());
+        super(logger.getName() != null ? logger.getName() : "");
```
jonatan-ivanov: I'm thinking if we should throw an exception here but it seems a `null` name is valid but in that case should not we give it a name like `unnamed`/`unknown`/`null` instead since empty logger names might not be easy to configure:
```suggestion
        super(logger.getName() != null ? logger.getName() : "unnamed");
```

micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java
```diff
@@ -210,6 +213,9 @@ private Function<ProceedingJoinPoint, Iterable<Tag>> makeSafe(
             method = pjp.getTarget().getClass().getMethod(method.getName(), method.getParameterTypes());
             timed = method.getAnnotation(Timed.class);
         }
+        if (timed == null) {
```
jonatan-ivanov: How about moving this into the previous if block?

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyClientMetrics.java
```diff
@@ -70,7 +70,7 @@ public class JettyClientMetrics implements Request.Listener {
     protected JettyClientMetrics(MeterRegistry registry, JettyClientTagsProvider tagsProvider, String timingMetricName,
             String contentSizeMetricName, int maxUriTags) {
         this(registry, ObservationRegistry.NOOP, null, tagsProvider, timingMetricName, contentSizeMetricName,
-                maxUriTags, (request, result) -> tagsProvider.uriPattern(result));
+                maxUriTags, (request, result) -> result != null ? tagsProvider.uriPattern(result) : "UNKNOWN");
```
jonatan-ivanov, line 73: Not necessarily the concern of this PR but I'm thinking if we should move this responsibility to `JettyClientTagsProvider` so implementors can map null responses to their own values (though they can do this with a `MeterFilter` too).

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
```diff
@@ -120,7 +120,7 @@ public void bindTo(MeterRegistry registry) {
     @Override
     public void close() throws Exception {
         if (registry != null) {
-            registeredMeterIds.forEach(registry::remove);
+            registeredMeterIds.forEach(id -> registry.remove(id));
```
jonatan-ivanov, line 123: I'm not getting this: other than this seems to be the same, `registeredMeterIds` should not contain `null` `Id`s.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm/JvmInfoMetrics.java
```diff
@@ -31,11 +31,16 @@ public class JvmInfoMetrics implements MeterBinder {
...
+                    getSystemProperty("java.vm.vendor", "unknown"), "runtime",
+                    getSystemProperty("java.runtime.name", "unknown"))
             .strongReference(true)
             .register(registry);
     }
 
+    private static String getSystemProperty(String key, String defaultValue) {
+        String value = System.getProperty(key, defaultValue);
+        return value != null ? value : defaultValue;
+    }
```
jonatan-ivanov, line 44: I'm not sure I get this, `System.getProperty(key, defaultValue)` does the null check, so if `defaultValue` is not null it will not return `null`.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
```diff
@@ -369,7 +369,7 @@ public void handleNotification(Notification notification2, Object handback2) {
                 MBeanServerNotification mbs2 = (MBeanServerNotification) notification2;
                 ObjectName o2 = mbs2.getMBeanName();
                 if (o2.equals(o)) {
-                    meters.stream().forEach(registry::remove);
+                    meters.forEach(m -> registry.remove(m));
```
jonatan-ivanov, line 372: This seems unnecessary too?

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryVictoriaMetricsHistogram.java
```diff
@@ -141,8 +144,9 @@ else if (inxs.offset == 1)
         AtomicLongArray hb = values.get(inxs.bucketIdx);
         if (hb == null) {
             hb = new AtomicLongArray(BUCKET_SIZE);
-            if (!values.compareAndSet(inxs.bucketIdx, null, hb))
-                hb = values.get(inxs.bucketIdx);
+            if (!values.compareAndSet(inxs.bucketIdx, null, hb)) {
+                hb = Objects.requireNonNull(values.get(inxs.bucketIdx));
```
jonatan-ivanov, line 148: But if we require non-null, should not AtomicReferenceArray require that too (`AtomicReferenceArray<AtomicLongArray>`)?

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/PercentileHistogramBuckets.java
```diff
@@ -78,8 +78,10 @@ public class PercentileHistogramBuckets {
      * @return The set of histogram buckets for use in computing aggregable percentiles.
      */
     public static NavigableSet<Double> buckets(DistributionStatisticConfig distributionStatisticConfig) {
-        return PERCENTILE_BUCKETS.subSet(distributionStatisticConfig.getMinimumExpectedValueAsDouble(), true,
-                distributionStatisticConfig.getMaximumExpectedValueAsDouble(), true);
+        Double min = distributionStatisticConfig.getMinimumExpectedValueAsDouble();
+        Double max = distributionStatisticConfig.getMaximumExpectedValueAsDouble();
+        return PERCENTILE_BUCKETS.subSet(min == null ? 1.0 : min, true, max == null ? Double.POSITIVE_INFINITY : max,
```
jonatan-ivanov, line 83: If min is null, should not we use 0th percentile instead of 1? Or if that could get us into issues with math (i.e.: divide by 0), can we use something smaller? (10E-9?)

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
```diff
@@ -175,10 +175,11 @@ private boolean shouldPublishDataForLastStep() {
      * Performs closing rollover on StepMeters.
      */
     private void closingRolloverStepMeters() {
-        getMeters().stream()
-            .filter(StepMeter.class::isInstance)
-            .map(StepMeter.class::cast)
-            .forEach(StepMeter::_closingRollover);
+        for (Meter meter : getMeters()) {
```
jonatan-ivanov, line 178: Better preformance?

micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
```diff
@@ -147,15 +147,12 @@ private void initPauseDetector(PauseDetector pauseDetectorType) {
                         e);
                 return;
             }
-        }
-        pauseDetector = pauseDetectorCache.computeIfAbsent(pauseDetectorType, detector -> {
-            if (detector instanceof ClockDriftPauseDetector) {
+            pauseDetector = pauseDetectorCache.computeIfAbsent(pauseDetectorType, detector -> {
```
jonatan-ivanov, line 150: I think this would cause a `ClassCastException`:

```java
class MyTimer extends AbstractTimer {
        MyTimer() {
            super(new Meter.Id("name", Tags.empty(), null, null, Meter.Type.TIMER), Clock.SYSTEM, DistributionStatisticConfig.DEFAULT, new MyPauseDetector(), TimeUnit.SECONDS, false);
        }
[...]
}

class MyPauseDetector implements PauseDetector {}
```

This seems like a valid test-case, I can add a test to `AbstractTimerTests` if you agree.

micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
```diff
@@ -357,10 +358,10 @@ public Builder description(@Nullable String description) {
          * @return This builder.
          * @since 1.5.0
          */
-        public Builder serviceLevelObjectives(@Nullable Duration... slos) {
+        public Builder serviceLevelObjectives(Duration @Nullable ... slos) {
```
jonatan-ivanov, line 360: Should we accept null here?

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -866,7 +866,7 @@ else if (reply == MeterFilterReply.ACCEPT) {
      */
     @Incubating(since = "1.2.0")
     public void clear() {
-        meterMap.keySet().forEach(this::remove);
+        meterMap.keySet().forEach(id -> remove(id));
```
jonatan-ivanov, line 869: Should we revert?

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
```diff
@@ -961,7 +961,8 @@ public Config onMeterRemoved(Consumer<Meter> meterRemovedListener) {
          * @since 1.6.0
          */
         @Incubating(since = "1.6.0")
-        public Config onMeterRegistrationFailed(BiConsumer<Meter.Id, String> meterRegistrationFailedListener) {
+        public Config onMeterRegistrationFailed(
+                BiConsumer<Meter.Id, @Nullable String> meterRegistrationFailedListener) {
```
jonatan-ivanov, line 965: I'm thinking if we should let the reason be null. Getting a notification saying registration failed for this id but not giving any reason is less helpful.
But if we say reason cannot be null we should change this:
```java
protected void meterRegistrationFailed(Meter.Id id, @Nullable String reason)
```

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/jetty/JettyClientMetricsTest.java
```diff
@@ -73,7 +73,10 @@ public void handle(String target, org.eclipse.jetty.server.Request baseRequest,
         httpClient.setFollowRedirects(false);
         // noinspection deprecation
         httpClient.getRequestListeners()
-            .add(JettyClientMetrics.builder(registry, result -> result.getRequest().getURI().getPath()).build());
+            .add(JettyClientMetrics.builder(registry,
```
jonatan-ivanov: Can we move this out to a method so it can be more readable?

jonatan-ivanov: As well as `JettyClientMetricsWithObservationTest`

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaMetricsTest.java
```diff
@@ -542,7 +542,7 @@ void removeShouldWorkForNonExistingMeters() {
         assertThat(registry.getMeters()).hasSize(1);
 
         kafkaMetricMap.clear();
-        registry.forEachMeter(registry::remove);
+        registry.forEachMeter(m -> registry.remove(m));
```
jonatan-ivanov, line 545: Revert?

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/DiskSpaceMetricsTest.java
```diff
@@ -38,7 +38,8 @@ class DiskSpaceMetricsTest {
     @Test
     void diskSpaceMetrics() {
         // tag::example[]
-        new DiskSpaceMetrics(new File(System.getProperty("user.dir"))).bindTo(registry);
+        String userDir = System.getProperty("user.dir");
+        new DiskSpaceMetrics(new File(userDir != null ? userDir : ".")).bindTo(registry);
```
jonatan-ivanov, line 42:
```suggestion
        new DiskSpaceMetrics(new File(System.getProperty("user.dir", "."))).bindTo(registry);
```

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/DiskSpaceMetricsTest.java
```diff
@@ -47,15 +48,17 @@ void diskSpaceMetrics() {
 
     @Test
     void diskSpaceMetricsWithTags() {
-        new DiskSpaceMetrics(new File(System.getProperty("user.dir")), Tags.of("key1", "value1")).bindTo(registry);
+        String userDir = System.getProperty("user.dir");
+        new DiskSpaceMetrics(new File(userDir != null ? userDir : "."), Tags.of("key1", "value1")).bindTo(registry);
```
jonatan-ivanov, line 52:
```suggestion
        new DiskSpaceMetrics(new File(System.getProperty("user.dir", ".")), Tags.of("key1", "value1")).bindTo(registry);
```

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/DiskSpaceMetricsTest.java
```diff
@@ -47,15 +48,17 @@ void diskSpaceMetrics() {
...
 
         assertThat(registry.get("disk.free").tags("key1", "value1").gauge().value()).isNotNaN().isGreaterThan(0);
         assertThat(registry.get("disk.total").tags("key1", "value1").gauge().value()).isNotNaN().isGreaterThan(0);
     }
 
     @Test
     void garbageCollectionDoesNotLoseGaugeValue() {
-        new DiskSpaceMetrics(new File(System.getProperty("user.dir"))).bindTo(registry);
+        String userDir = System.getProperty("user.dir");
+        new DiskSpaceMetrics(new File(userDir != null ? userDir : ".")).bindTo(registry);
```
jonatan-ivanov, line 61:
```suggestion
        new DiskSpaceMetrics(new File(System.getProperty("user.dir"), ".")).bindTo(registry);
```

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetricsTest.java
```diff
@@ -48,7 +48,8 @@ void fileDescriptorMetricsUnsupportedOsBeanMock() {
 
     @Test
     void unixFileDescriptorMetrics() {
-        assumeFalse(System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("win"));
+        String osName = System.getProperty("os.name");
+        assumeFalse(osName != null && osName.toLowerCase(Locale.ROOT).contains("win"));
```
jonatan-ivanov: This and the `windowsFileDescriptorMetrics` case are slightly different from nullness perspective: for this, `osName` being null is ok, for `windowsFileDescriptorMetrics` it is not.

Should not we do something like this instead?
```java
assumeNotNull(osName);
assumeFalse(osName.toLowerCase(Locale.ROOT).contains("win"));
```
and the same with `assumeTrue` for `windowsFileDescriptorMetrics`?

jonatan-ivanov: Or maybe we should migrate to JUnit5's `@EnabledOnOs`/`@DisabledOnOs`?
Same for `ProcessorMetricsTest`.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
```diff
@@ -316,11 +316,11 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
         });
     }
 
-    void runTomcat(HttpServlet servlet, Callable<Void> doWithTomcat) throws Exception {
+    void runTomcat(HttpServlet servlet, Callable<?> doWithTomcat) throws Exception {
```
jonatan-ivanov: What's wrong with `Void`?

micrometer-core/src/test/java/io/micrometer/core/instrument/step/StepMeterRegistryTest.java
```diff
@@ -635,7 +636,8 @@ private class MyStepMeterRegistry extends StepMeterRegistry {
 
         AtomicBoolean isPublishing = new AtomicBoolean(false);
 
-        CompletableFuture<Void> scheduledPublishingFuture = CompletableFuture.completedFuture(null);
+        Future<?> scheduledPublishingFuture = CompletableFuture.runAsync(() -> {
```
jonatan-ivanov: `Future<Void>`?
```suggestion
        Future<Void> scheduledPublishingFuture = CompletableFuture.runAsync(() -> {
```

micrometer-java11/src/main/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClient.java
```diff
@@ -269,7 +270,7 @@ public <T> CompletableFuture<HttpResponse<T>> sendAsync(HttpRequest httpRequest,
             }
             else {
                 stopObservationOrTimer(instrumentation, request, response);
-                return response;
+                return Objects.requireNonNull(response);
```
jonatan-ivanov, line 273: Is this only a workaround to make NullAway happy or can the response actually be `null` if throwable is also `null`?

micrometer-observation-test/src/main/java/io/micrometer/observation/tck/InvalidObservationException.java
```diff
@@ -71,15 +72,15 @@ public static class HistoryElement {
...
+        private final @Nullable StackTraceElement[] stackTrace;
 
         HistoryElement(EventName eventName) {
             this.eventName = eventName;
             StackTraceElement[] currentStackTrace = Thread.currentThread().getStackTrace();
             this.stackTrace = findRelevantStackTraceElements(currentStackTrace);
         }
 
-        private StackTraceElement[] findRelevantStackTraceElements(StackTraceElement[] stackTrace) {
+        private @Nullable StackTraceElement[] findRelevantStackTraceElements(@Nullable StackTraceElement[] stackTrace) {
```
jonatan-ivanov: I'm not sure this can return null.
Nor that `stackTrace` should be nullable.

micrometer-observation/src/main/java/io/micrometer/observation/ObservationPredicate.java
```diff
@@ -26,6 +28,6 @@
  * @author Marcin Grzejszczak
  * @since 1.10.0
  */
-public interface ObservationPredicate extends BiPredicate<String, Observation.Context> {
+public interface ObservationPredicate extends BiPredicate<@Nullable String, Observation.Context> {
```
jonatan-ivanov, line 31: Should we allow null name?

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/DiskMetricsSample.java
```diff
@@ -29,7 +29,8 @@ public class DiskMetricsSample {
 
     public static void main(String[] args) {
         MeterRegistry registry = SampleConfig.myMonitoringSystem();
-        new DiskSpaceMetrics(new File(System.getProperty("user.dir"))).bindTo(registry);
+        String userDir = System.getProperty("user.dir");
+        new DiskSpaceMetrics(new File(userDir != null ? userDir : ".")).bindTo(registry);
```
jonatan-ivanov, line 33:
```suggestion
        new DiskSpaceMetrics(new File(System.getProperty("user.dir", "."))).bindTo(registry);
```

samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/ObservationHandlerSample.java
```diff
@@ -127,7 +128,7 @@ public String toString() {
     static class IgnoringObservationPredicate implements ObservationPredicate {
 
         @Override
-        public boolean test(String name, Observation.Context context) {
+        public boolean test(@Nullable String name, Observation.Context context) {
```
jonatan-ivanov, line 131: Name should not be null.

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyServerThreadPoolMetrics.java
shakuzen: I think it's this, which should hopefully be fixed soon: https://github.com/uber/NullAway/issues/1719

micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaConsumerMetrics.java
shakuzen: Again https://github.com/uber/NullAway/issues/1719 I think

micrometer-core/src/main/java/io/micrometer/core/instrument/distribution/FixedBoundaryVictoriaMetricsHistogram.java
shakuzen: That would be nice but we can't since it is initialized with `null` values. I don't think there's a way in which `null` is set as a value after construction, but we can't express that (only initially null) as far as I know. Since we know `null` is never set we also know `values.get(inxs.bucketIdx)` is not null here, so it's a bit wasteful to add a null check. I'm not sure whether it is better to leave it or add a suppress warning to avoid it. It should get optimized away by the compiler but until/if that happens it is wasteful.

micrometer-core/src/main/java/io/micrometer/core/instrument/step/StepMeterRegistry.java
shakuzen: Maybe, but this was another case of https://github.com/uber/NullAway/issues/1719 I think. We should probably hold off merging this PR until the next NullAway release so we can clean up these things.

micrometer-core/src/main/java/io/micrometer/core/instrument/AbstractTimer.java
shakuzen: I don't think so. This is inside the `if (pauseDetectorType instanceof ClockDriftPauseDetector) {` block, so the call to `computeIfAbsent` passing `pauseDetectorType` is guaranteed to be an instance of `ClockDriftPauseDetector` in the mapping function. I don't think it's possible for it to throw a CCE.

Before these changes, this code was outside the instanceof check block for `ClockDriftPauseDetector`, so it was possible it wasn't an instance of ClockDriftPauseDetector. But in that case it was doing nothing. So the code changes look equivalent while being simpler and still safe. 

It also doesn't make sense to make a custom PauseDetector type because nothing will be done with it. It will just waste cycles doing the `instanceof` checks and match none. If we weren't on ancient Java, it would make sense to make it a sealed type.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: https://github.com/uber/NullAway/issues/1719 again

micrometer-core/src/main/java/io/micrometer/core/instrument/LongTaskTimer.java
shakuzen: Generally no, but I guess that's a breaking change in some sense to tighten the contract. It looks like the `@Nullable` annotation was not moved when switching to JSpecify, perhaps? It definitely doesn't make sense to accept null elements, so it was in the wrong place before. Whether we want to accept a null array or not is another question. I think other similar places accept null arrays.

shakuzen:
Marking blocked until a fix with at least https://github.com/uber/NullAway/issues/1719 is available.

micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java
shakuzen: I think that method signature is exactly why this was changed. We'd have to decide how big the risk of breaking users is by changing it to non-null.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/kafka/KafkaMetricsTest.java
shakuzen: https://github.com/uber/NullAway/issues/1719

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/FileDescriptorMetricsTest.java
shakuzen: I think the annotations are cleaner. Let's go with that. I'll push a change.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/system/DiskSpaceMetricsTest.java
shakuzen: Unfortunately NullAway is not that smart:
```
error: [NullAway] passing @Nullable parameter 'System.getProperty("user.dir", ".")' where @NonNull is required
```
The default value is considered nullable, and so the return value is considered still nullable, even though we're obviously passing a non-null value. I don't know if this is something that will be improved in a future version of NullAway. I think it needs [`@PolyNull` support](https://github.com/jspecify/jspecify/wiki/polynull) and there's no ETA on that.

micrometer-java11/src/main/java/io/micrometer/java11/instrument/binder/jdk/MicrometerHttpClient.java
shakuzen: It seems it's just to make NullAway happy, at least in the current JDK implementation. We could suppress the warning if we want.

micrometer-core/src/test/java/io/micrometer/core/instrument/binder/tomcat/TomcatMetricsTest.java
shakuzen: It needs to be marked `@Nullable` it seems because `Void` is always null. But things get a bit wonky between builds on different JDKs where the experimental flag is or isn't. Using `?` instead avoids issues in all builds, but it's an ugly hack. I think we can use `@Nullable Void` instead. I'm testing. Maybe we can only run jspecify mode on JDK 22+ and it will then match when this experimental mode is enabled and simplify things.

implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java
shakuzen: Updated. I don't know why it was added to begin with.

PR #7882, opened 2026-08-26
Make japicmp baseline dependency declarative

shakuzen:
## Summary
- Replaces the imperative `downloadBaseline` copy task with a declarative detached configuration wired directly to `japicmp.oldClasspath`.
- Bypasses Gradle's automatic project dependency substitution to ensure published baseline artifacts are resolved without self-referencing.
- Adds an execution-time assertion in `japicmp.doFirst` to verify that the resolved baseline JAR matches the expected baseline version.
- Wires `jar.archiveFile` directly into `japicmp.newClasspath` using Gradle Provider APIs.
- Adds `docs` to the non-published modules check in `build.gradle`, eliminating obsolete publishing and `japicmp` disable blocks from `docs/build.gradle`.
- Enables native Gradle `UP-TO-DATE` checks and build caching (`FROM-CACHE`) for `japicmp` tasks while unblocking Gradle Configuration Cache support.

## Test plan
- Verified `./gradlew :micrometer-core:japicmp` correctly compares against `micrometer-core-1.15.0.jar` (not self).
- Verified second execution of `./gradlew :micrometer-core:japicmp` is `UP-TO-DATE`.
- Verified `./gradlew japicmp --configuration-cache` stores and reuses configuration cache cleanly across all subprojects.
- Verified `./gradlew check -x test` builds successfully.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -424,20 +424,10 @@ subprojects {
             if (!(project.name in [])) { // add projects here that do not exist in the previous minor so should be excluded from japicmp
                 apply plugin: 'me.champeau.gradle.japicmp'
 
-                def baselineConfig = configurations.create('japicmpBaseline') {
-                    visible = false
-                    canBeConsumed = false
-                    canBeResolved = true
-                    description = "Baseline dependency for Japicmp binary compatibility check"
-                }
-                if (compatibleVersion != 'SKIP') {
-                    dependencies.add(baselineConfig.name, "${project.group}:${project.name}:${compatibleVersion}@jar")
-                }
-
                 tasks.register("japicmp", me.champeau.gradle.japicmp.JapicmpTask) {
                     dependsOn jar
-                    oldClasspath.from(baselineConfig)
-                    newClasspath.from(files(jar.archiveFile, project(":${project.name}").jar))
+                    oldClasspath.from(configurations.detachedConfiguration(dependencies.create("${project.group}:${project.name}:${compatibleVersion}@jar")))
+                    newClasspath.from(jar.archiveFile)
                     onlyBinaryIncompatibleModified = true
                     failOnModification = true
                     failOnSourceIncompatibility = true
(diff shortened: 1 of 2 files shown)


jonatan-ivanov (approved):

docs/build.gradle
```diff
@@ -113,15 +113,3 @@ jar {
...
-tasks.withType(AbstractPublishToMaven).configureEach {
-    enabled = false
-}
-
-downloadBaseline {
-    enabled = false
-}
-
-japicmp {
-    enabled = false
```
shakuzen, line 126: There were issues with the configuration cache and doing this because it still tried to resolve the inputs for the task. Instead, I added the docs module to the list of modules that don't publish. I think that's fine, but hopefully I'm not missing a reason we didn't do that originally.

build.gradle
```diff
@@ -425,23 +425,14 @@ subprojects {
...
+                String expectedJarName = "${project.name}-${compatibleVersion}.jar"
                 tasks.register("japicmp", me.champeau.gradle.japicmp.JapicmpTask) {
-                    dependsOn downloadBaseline, jar
-                    oldClasspath.from(layout.buildDirectory.file("baselineLibs/${project.name}-${compatibleVersion}.jar"))
-                    newClasspath.from(files(jar.archiveFile, project(":${project.name}").jar))
+                    dependsOn jar
+                    oldClasspath.from(configurations.detachedConfiguration(dependencies.create("${project.group}:${project.name}:${compatibleVersion}@jar")))
+                    newClasspath.from(jar.archiveFile)
+                    doFirst {
+                        assert oldClasspath.files.any { it.name.endsWith(expectedJarName) }
```
shakuzen, line 433: This assert is configuration cache-friendly according to Cursor because the expected name is resolved outside the task body. `doFirst` will only run when the task is not cached.

shakuzen:
As discussed offline, this change doesn't save a lot of time because japicmp was already cacheable and the download task is very fast when the baseline is already in the gradle cache, but it eliminates the download task entirely and makes this configuration cache-friendly.

jonatan-ivanov (approved):

Issue #7859, opened 2026-08-21
POM properties can be misleading


jonatan-ivanov:
`url`, `nebula_Build_Url`, and `nebula_Module_Origin` can be misleading when we release from a private repository, see: https://repo1.maven.org/maven2/io/micrometer/micrometer-core/1.18.0-M1/micrometer-core-1.18.0-M1.pom

Found by @m-burst in https://github.com/micrometer-metrics/micrometer/issues/7854

nick-montana:
should I just open a pull request for completing tasks like this? Rather than waiting on an authorized contributor to assign the issue

jonatan-ivanov:
This particular case is tricky since it involves our release train setup which is not public and we need to discuss and decide what to do. If you have an idea for fixing this and you don't mind that we might not go to that direction, feel free to share it in the comments or (if it's easier that way) in the form of a PR. Issues with the label `help wanted` are a bit more suitable for contributions but if you want to dig deeper on this feel free to go ahead.

nick-montana:
I'll leave it alone, thanks

PR #7887, opened 2026-08-26
Change POM properties to reflect the right urls

jonatan-ivanov:
Closes gh-7859

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -471,6 +471,22 @@ subprojects {
                 github 'jonatan-ivanov'
             }
         }
+
+        infoBroker.excludedManifestProperties.addAll('Module-Origin', 'Module-Source')
+
+        // modify publications if necessary after Nebula plugins applied
+        plugins.withId('maven-publish') {
+            publishing {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -471,6 +471,22 @@ subprojects {
...
         }
+
+        infoBroker.excludedManifestProperties.addAll('Module-Origin', 'Module-Source')
+
+        // modify publications if necessary after Nebula plugins applied
+        plugins.withId('maven-publish') {
+            publishing {
+                publications {
+                    nebula(MavenPublication) {
+                        if (pom.url.get().startsWith('https://github.com/micrometer-metrics/')) {
```
jonatan-ivanov, line 482: This makes sure there is no effect on forks though I'm not sure it matters a lot since the contacts are hardcoded without any checks.

jonatan-ivanov: Also, this does not change `nebula_Build_Url`.

shakuzen (approved):

PR #7890, opened 2026-08-28
Make HdrHistogram an optional dependency

raphw:
I suggest turning HdrHistogram into an optional dependency. This pattern of optional dependencies is already present within micrometer, and the majority of users of micrometer likely do not rely on it. If the feature is used without adding the dependency, this can be indicated by an explicit error message. The current behaviour can easily be reinstated by simply adding HdrHistogram explicitly, which also states this need in a clearer way.

The background for this suggestion is also that the project seems abandoned. There has not been any activity in almost two years, tickets are not answered and the project seems abandoned (https://github.com/HdrHistogram/HdrHistogram).

To make things worse, the latest release removed its previous module name as the build file was misconfigured.

For a possible alternative, DataDog offers a library that could fit, but this would maybe be an awkward dependency to a Vendor what does not sound well for a vendor-independent API.

Alternatively, one could keep HdrHistorgram as a regular dependency, but still make the code changes, so it can be removed cleanly with a meaningful error message if the library was excluded.

### DIFF
--- a/docs/modules/ROOT/pages/concepts.adoc
+++ b/docs/modules/ROOT/pages/concepts.adoc
@@ -19,4 +19,7 @@ It does not require any third-party (non-Micrometer) dependencies on the classpa
 Use of the xref:./concepts/timers.adoc#_pause_detection[pause detection] feature requires the https://github.com/LatencyUtils/LatencyUtils[LatencyUtils] dependency (`org.latencyutils:LatencyUtils`) to be available on the classpath at runtime.
 Micrometer does not include LatencyUtils on the runtime classpath by default because pause detection is turned off by default.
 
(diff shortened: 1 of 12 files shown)


shakuzen:
While I think a reasonable argument could be made for making it optional, we've always kept HdrHistogram included by default because it felt like a better user experience that client-side percentiles work out-of-the-box if configured. After almost a decade, it may be worth considering how much client-side percentiles are used nowadays. The popularity of backends that support server-side percentile estimation is certainly more now than when Micrometer was created. And we do encourage using server-side percentiles when possible. Still, it can be somewhat subtle that client-side percentiles are configured. We do [document](https://docs.micrometer.io/micrometer/reference/concepts.html#concepts-dependencies) that the dependency can be excluded if not using client-side percentiles.

> Alternatively, one could keep HdrHistorgram as a regular dependency, but still make the code changes, so it can be removed cleanly with a meaningful error message if the library was excluded.

This would be an acceptable first step for now, while we take more time to consider additional changes.

PR #7894, opened 2026-08-28
Send the Datadog API keys in the headers

jonatan-ivanov:


implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
```diff
@@ -265,15 +266,16 @@ String writeMetric(Meter.Id id, @Nullable String suffix, long wallTime, double v
      */
     // VisibleForTesting
     void postMetricMetadata(String metricName, DatadogMetricMetadata metadata) {
-        // already posted the metadata for this metric, or no data to post
-        if (metadata.editMetadataBody() == null || verifiedMetadata.contains(metricName)) {
+        // no application key, or no data to post, or already posted the metadata
+        if (config.applicationKey() == null || metadata.editMetadataBody() == null
```
shakuzen, line 270: Adding `config.applicationKey() == null` here seems like new behavior. We should add a test for it. I guess before it would send the request with a `null` application key?

shakuzen (approved):

implementations/micrometer-registry-datadog/src/main/java/io/micrometer/datadog/DatadogMeterRegistry.java
jonatan-ivanov: Yes, previously, it did this:
```
?api_key=s3cr3t&application_key=null
```
Which is invalid since DataDog needs an application key for this API call.
I added a test to verify the new behavior

PR #7895, opened 2026-08-28
Do not add web fonts to javadoc

jonatan-ivanov:
Prior to this change, every module that generated a javadoc.jar included web fonts that add about 4MiB to the size of the generated jar.

On my machine, by running `./gradlew pTML`, this means the following change in size of the publishing folder:
- `1.18.x`: `144.0` -> `16.7` MiB
- `1.16.x`: `142.4` -> `15.1` MiB

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -204,11 +204,11 @@ subprojects {
                 targetCompatibility = JavaVersion.VERSION_11
             }
 
-
             javadoc {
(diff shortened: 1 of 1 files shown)


shakuzen (approved):
Looks fine. I suppose it's worth mentioning the motivation for this is the new Maven Central limitations.

shakuzen:
I assume the PR target should be `1.16.x` instead of `main`

PR #7896, opened 2026-08-29
Proxy JMS session results using the method's declared return type

Gaurav1112:
**Problem**

`SessionInvocationHandler` wraps every `MessageProducer` result as `MessageProducer.class` and every `MessageConsumer` result as `MessageConsumer.class`, without consulting the intercepted method's declared return type.

`jakarta.jms.Session` declares:

```java
TopicSubscriber createDurableSubscriber(Topic topic, String name) throws JMSException;
TopicSubscriber createDurableSubscriber(Topic topic, String name, String messageSelector, boolean noLocal) throws JMSException;
```

`TopicSubscriber` is a `MessageConsumer` subtype, so the `instanceof MessageConsumer` branch fires, but the proxy handed back implements only `MessageConsumer`. Calling that method on an instrumented session fails:

```java
Session instrumented = JmsInstrumentation.instrumentSession(session, registry);
instrumented.createDurableSubscriber(topic, "sub");
// java.lang.ClassCastException: class jdk.proxy3.$Proxy74 cannot be cast to
// class jakarta.jms.TopicSubscriber
```

So durable topic subscriptions stop working as soon as JMS instrumentation is applied — adding observability to a working application turns subscription setup into a hard runtime failure.

**Solution**

Use `method.getReturnType()` as the proxy interface in both branches. Every relevant `Session` return type (`MessageProducer`, `MessageConsumer`, `TopicSubscriber`, `QueueReceiver`, `TopicPublisher`) is an interface, so this is safe. The producer path is behaviourally unchanged, since `createProducer` already declares `MessageProducer`.

**Tests**

Added `SessionInvocationHandlerTests` covering the durable-subscriber case plus the plain consumer and producer paths, so the existing behaviour is pinned.

The new test fails on current `main` with the `ClassCastException` above and passes with this change. I also reverted only the production hunk while keeping the test and confirmed it fails again as an assertion rather than a compile error, so the test genuinely exercises the fix.

Locally on JDK 21: `micrometer-jakarta9` is 87 tests / 0 failures, and `checkFormatMain` / `checkFormatTest` are clean.

**Notes**

There is no existing coverage of `createDurableSubscriber` anywhere in the repository, which is likely why this went unnoticed.

I noticed #7880 also touches this file (adding `@Nullable` to the `invoke` signature) — happy to rebase if that lands first.

### DIFF
--- a/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/SessionInvocationHandler.java
+++ b/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/SessionInvocationHandler.java
@@ -58,14 +58,14 @@ class SessionInvocationHandler implements InvocationHandler {
                 MessageProducerInvocationHandler producerHandler = new MessageProducerInvocationHandler(producer,
                         this.registry);
                 return Proxy.newProxyInstance(this.target.getClass().getClassLoader(),
-                        new Class[] { MessageProducer.class }, producerHandler);
+                        new Class[] { method.getReturnType() }, producerHandler);
             }
             if (result instanceof MessageConsumer) {
                 MessageConsumer consumer = (MessageConsumer) result;
                 MessageConsumerInvocationHandler consumerHandler = new MessageConsumerInvocationHandler(consumer,
                         this.registry);
                 return Proxy.newProxyInstance(this.target.getClass().getClassLoader(),
-                        new Class[] { MessageConsumer.class }, consumerHandler);
+                        new Class[] { method.getReturnType() }, consumerHandler);
             }
             return result;
         }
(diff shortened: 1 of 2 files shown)


Gaurav1112:
Gentle bump after two weeks — CI is awaiting first-time-contributor approval. Validated locally in a container: the full `./gradlew :micrometer-jakarta9:test` suite passes with the change, and the new tests fail without it (assertion, not compile error). Happy to adjust anything about the approach if it doesn't fit the project's direction.

shakuzen (commented):
Thanks for the PR and details about the bug. I think we overlooked this. Could you rebase the changes targeting `1.16.x`? We'll review this for inclusion in the next round of bug fixes. /cc @bclozel in case I'm missing anything

bclozel:
Looks good to me!

shakuzen (approved):

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you @Gaurav1112!

PR #7924, opened 2026-09-05
Allow configuring custom observation conventions for JMS

### LINKED ISSUE #5492: Configure a custom JmsProcessObservationConvention or JmsPublishObservationConvention
It should be possible to configure a custom JmsProcessObservationConvention. Currently a DefaultJmsProcessObservationConvention ist hardcoded in the MessageConsumerInvocationHandler.

**Rationale**
I want to write custom information from the JMS header to the observation (trace info).

**Additional context**
Its possible to define your own JmsProcessObservationConvention, but I found no possibilty to activate/configure it instead of the DefaultJmsProcessObservationConvention.

alxkm:
Closes gh-5492                                                                                                                                                           
                                                                                                                                                                           
  JmsInstrumentation always passed `null` as the custom convention, so the                                                                                                 
  observation name, contextual name and key values could not be customized                                                                                                 
  without a GlobalObservationConvention.                                                                                                                                   
                                                                                                                                                                           
  This adds an `instrumentSession` overload taking a                                                                                                                       
  JmsPublishObservationConvention and a JmsProcessObservationConvention,                                                                                                   
  following the same approach already used by InstrumentedTransport for the                                                                                                
  mail instrumentation. Passing `null` for a convention keeps the default one,                                                                                             
  so either observation can be customized on its own.                                                                                                                      
                                                                                                                                                                           
  The second commit fixes a race in the existing                                                                                                                           
  `shouldInstrumentMessageListenerWhenException` test, which the new tests                                                                                                 
  made surface reliably.

### DIFF
--- a/docs/modules/ROOT/pages/reference/jms.adoc
+++ b/docs/modules/ROOT/pages/reference/jms.adoc
@@ -54,6 +54,26 @@ MessageConsumer consumer = session.createConsumer(topic);
 consumer.setMessageListener(message -> consumeMessage(message));
 ----
 
+=== Customizing the observations
+
+By default, the observations are created with the `DefaultJmsPublishObservationConvention` and
+`DefaultJmsProcessObservationConvention` conventions. If you need to change the observation name,
+the contextual name or the `KeyValues` that are recorded, you can provide your own conventions:
+
+[source,java]
+----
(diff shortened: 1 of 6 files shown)


shakuzen (approved):
Looks good. Thank you for the very thorough PR.

PR #7944, opened 2026-09-10
Make active LTT opt-in in DefaultMeterObservationHandler

### LINKED ISSUE #7312: Make inclusion of `LongTaskTimer` by `DefaultMeterObservationHandler` opt-in instead of opt-out
`DefaultMeterObservationHandler` previously unconditionally made a `LongTaskTimer`. Later, in https://github.com/micrometer-metrics/micrometer/pull/4755, it was made possible to opt-out of the `LongTaskTimer` creation based on feedback from users who didn't need the `LongTaskTimer`. Given the feedback that some users don't need the `LongTaskTimer`, it is reasonable to question whether the majority of users benefit from it being included by default or not. If not, we should consider making it off by default, and users who benefit from it can opt-in to it. This would be a breaking change for users who utilize the LongTaskTimer created by default with the DefaultMeterObservationHandler. Hence, we plan to gather feedback on this.

shakuzen:
Deprecate existing constructors and introduce a builder that defaults to not including the active observation LongTaskTimer.

This is implemented in a way that is not a breaking change. It leaves the existing behavior of the constructors but marks them as deprecated. The non-deprecated builder has different default behavior.

I'm not sure this is the best way to evolve things but it avoids some problems. It is awkward to negate things with the enum `IgnoredMeters`. It's also potentially problematic to key on meter **types** because in theory there could be more than one meter of the same type created by the `DefaultMeterObservationHandler`. The varargs constructor shadows the single arg constructor making for some ambiguity if adding/evolving constructors.

The problems with this proposal are that everywhere instantiating `DefaultMeterObservationHandler` needs to be updated to get off deprecated API. The hope is this is limited impact because it should mostly be handled by frameworks. The other point to be aware of is that with no remaining non-deprecated constructors, this more clearly signals that this class is not designed for extension, and the JavaDoc has been updated to state as much. I'm not sure we ever did intend for it to be extended, and [a GitHub search](https://github.com/search?q=%22extends+DefaultMeterObservationHandler%22+language%3AJava&type=code) shows only two cases of that on GitHub.

Closes gh-7312

### DIFF
--- a/micrometer-test-aspectj-ctw/src/test/java/io/micrometer/test/ctw/MeasuredClassTest.java
+++ b/micrometer-test-aspectj-ctw/src/test/java/io/micrometer/test/ctw/MeasuredClassTest.java
@@ -46,10 +46,17 @@ class MeasuredClassTest {
         System.out.println("java.version: " + System.getProperty("java.version"));
     }
 
+    @SuppressWarnings("deprecation")
     @BeforeEach
     void setUp() {
-        observationRegistry.observationConfig()
-            .observationHandler(DefaultMeterObservationHandler.builder(registry).build());
+        // Intentionally uses the deprecated constructor rather than
+        // DefaultMeterObservationHandler.builder(MeterRegistry): on Java < 17 builds this
+        // module compiles against a published micrometer-core SNAPSHOT instead of the
+        // local project (see this module's build.gradle and gh-6064), so API added in the
+        // current branch is not available here. This test only needs some
+        // MeterObservationHandler to verify compile-time weaving; which constructor is
+        // used is irrelevant to what it covers.
+        observationRegistry.observationConfig().observationHandler(new DefaultMeterObservationHandler(registry));
         // Global registry must be used because aspect gets created for us
         Metrics.addRegistry(registry);
         Observations.setRegistry(observationRegistry);
--- a/micrometer-test-aspectj-ctw/src/test/java/io/micrometer/test/ctw/MeasuredClassTest.java
+++ b/micrometer-test-aspectj-ctw/src/test/java/io/micrometer/test/ctw/MeasuredClassTest.java
@@ -48,7 +48,8 @@ class MeasuredClassTest {
 
     @BeforeEach
     void setUp() {
-        observationRegistry.observationConfig().observationHandler(new DefaultMeterObservationHandler(registry));
+        observationRegistry.observationConfig()
+            .observationHandler(DefaultMeterObservationHandler.builder(registry).build());
         // Global registry must be used because aspect gets created for us
         Metrics.addRegistry(registry);
         Observations.setRegistry(observationRegistry);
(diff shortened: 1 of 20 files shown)


micrometer-test/src/main/java/io/micrometer/core/instrument/InstrumentationVerificationTests.java
```diff
@@ -49,12 +49,19 @@ protected MeterRegistry getRegistry() {
     /**
      * Helper method for creating a {@link TestObservationRegistry} to run the tests also
      * using the Observation API.
+     * <p>
+     * The registered handler uses the {@link DefaultMeterObservationHandler} defaults,
+     * which as of 1.18.0 do not include the active observation {@link LongTaskTimer}
```
shakuzen, line 54: I'm not sure I like referencing a specific version in the JavaDoc here, but it's being overly cautious about trying to not surprise users.

micrometer-test-aspectj-ctw/src/test/java/io/micrometer/test/ctw/MeasuredClassTest.java
```diff
@@ -46,8 +46,16 @@ static void setUpClass() {
...
     @BeforeEach
     void setUp() {
+        // Intentionally uses the deprecated constructor rather than
+        // DefaultMeterObservationHandler.builder(MeterRegistry): on Java < 17 builds this
+        // module compiles against a published micrometer-core SNAPSHOT instead of the
+        // local project (see this module's build.gradle and gh-6064), so API added in the
+        // current branch is not available here. This test only needs some
+        // MeterObservationHandler to verify compile-time weaving; which constructor is
+        // used is irrelevant to what it covers.
         observationRegistry.observationConfig().observationHandler(new DefaultMeterObservationHandler(registry));
```
shakuzen, line 59: This is nasty. We can't use the new API in the CTW module with the JDK 11 build because the version of the plugin we're using that does the CTW in the micrometer-core module requires JDK 17+, so we use published snapshots, which don't have the new API (the builder on `DefaultMeterObservationHandler`). We can revert this post-merge once a snapshot is published, but it shows how the current arrangement is not great from a reproducible build perspective and has these odd failures. I guess if we drop the JDK 11 build, we can avoid the underlying problem, as one long-term solution.

jonatan-ivanov: I think we should drop building `micrometer-test-aspectj-ctw` below Java 17.

jonatan-ivanov (approved):

micrometer-test-aspectj-ctw/src/test/java/io/micrometer/test/ctw/MeasuredClassTest.java
shakuzen: Good idea. I've updated the configuration.

PR #7953, opened 2026-09-15
Allow custom high cardinality meter selection

### LINKED ISSUE #7649: Returning the metric name with the highest cardinality of tags
**The feature request.**

Currently, HighCardinalityTagsDetector returns the first metric that exceeds the threshold, rather than the one with the highest tag cardinality. This occurs because the meter registry returns a list ordered by the metric's appearance, rather than by its cardinality.

The goal of this change is to ensure we evaluate and return the metric with the absolute highest number of tags.

**Rationale**

This improvement will allow us to more quickly identify the specific metric driving the highest costs, instead of settling for the first one that breaches the threshold.

**Additional context**

The focus of this request is the findFirstHighCardinalityMeterInfo method in HighCardinalityTagsDetector, which is currently implemented this way:

```
    public Optional<HighCardinalityMeterInfo> findFirstHighCardinalityMeterInfo() {
        Map<String, Long> meterNameFrequencies = new LinkedHashMap<>();
        for (Meter meter : this.registry.getMeters()) {
            meterNameFrequencies.compute(meter.getId().getName(), (k, v) -> v == null ? 1 : v + 1);
        }
        return meterNameFrequencies.entrySet()
            .stream()
            .filter((entry) -> entry.getValue() > this.threshold)
            .map((entry) -> new HighCardinalityMeterInfo(entry.getKey(), entry.getValue()))
            .findFirst();
    }
```

Below is a draft proposal. The main goal is to sort only the metrics that exceed the threshold and return the one with the highest cardinality. If multiple metrics share the same high cardinality, we return the first one we encounter to satisfy the interface.

  ```
@Override
  public Optional<HighCardinalityMeterInfo> findFirstHighCardinalityMeterInfo() {
    Map<String, Long> meterNameFrequencies = new LinkedHashMap<>();
    for (Meter meter : this.registry.getMeters()) {
      meterNameFrequencies.compute(meter.getId().getName(), (k, v) -> v == null ? 1 : v + 1);
    }

    SortedMap<Long, List<HighCardinalityMeterInfo>> sortedMeterNameAboveThresholdByFrequencies =
        meterNameFrequencies.entrySet().stream()
            .filter(entry -> isAboveThreshold(entry.getValue(), threshold))
            .map((entry) -> new HighCardinalityMeterInfo(entry.getKey(), entry.getValue()))
            .collect(
                java.util.stream.Collectors.groupingBy(
                    HighCardinalityMeterInfo::getCount,
                    java.util.TreeMap::new,
                    java.util.stream.Collectors.toList()));

    if (sortedMeterNameAboveThresholdByFrequencies.isEmpty()) {
      return Optional.empty();
    }

    return sortedMeterNameAboveThresholdByFrequencies.lastEntry().getValue().stream().findFirst();
  }

  private boolean isAboveThreshold(long count, long threshold) {
    return count > threshold;
  }
```

We will need to evaluate the introduced memory cost and the performance impact in heavy-load scenarios.

Thank you

dlwhdgus0810:
Closes #7649.

`HighCardinalityTagsDetector` always selects the first meter name above the threshold. Add a builder selector so callers can prioritize the highest count or apply another selection strategy, as discussed in the issue.

The default still selects the first candidate. Selectors receive a fresh sequential stream filtered by the threshold; returning an empty `Optional` suppresses the result. Both one-time checks and scheduled notifications use the selector. The reference guide includes a tested maximum-count example and explains its traversal cost.

Validation (JDK 25): the full core suite completed 1,090 tests with no failures and 16 skips, including all detector and allocation tests. All 4 documentation tests passed. Checkstyle and formatting checks passed for both changed modules. New coverage includes default selection, maximum count, threshold filtering, repeated checks, empty results, scheduling and null-selector rejection.


jonatan-ivanov:
Thank you for the PR!
I took a look and I was thinking if we can make this more flexible/general by returning all the info about meters with high cardinality, not only the first one. Something like this: https://github.com/micrometer-metrics/micrometer/pull/7969

Could you please let me know what do you think?

dlwhdgus0810:
> Thank you for the PR! I took a look and I was thinking if we can make this more flexible/general by returning all the info about meters with high cardinality, not only the first one. Something like this: #7969
> 
> Could you please let me know what do you think?

Thanks for reviewing this and putting together #7969!

Returning all high-cardinality meter info makes sense to me. It also covers the highest-count use case that motivated the selector approach, while allowing callers to sort, filter, or report multiple results without adding a selection strategy to the detector.

Keeping the existing first-match behavior as the default is helpful too. I’m happy with this direction, and happy to help with any additional tests!

PR #7955, opened 2026-09-15
Do not modify or retain the array passed to Tags.and(Tag...)

kdelay:
`Tags.of(Tag[])` and `Tags.and(Tag[])` pass the caller's array to `toTags`, which sorts and deduplicates it in place and then keeps the reference. So the factory reorders an array the caller still owns, and the returned `Tags` - documented as immutable - changes when the caller writes to that array afterwards. `KeyValues.and(KeyValue...)` behaves the same.

Before gh-4959 (1.13.0), `and(Tag...)` copied into a new array with `System.arraycopy`, so neither happened.

This copies the array at that entry point. The other `toTags`/`toKeyValues` call sites are given arrays these classes build themselves (`toArray`, `of(String...)`) and are unchanged.

Three tests are added per class; all six fail on 1.16.x without the change.


shakuzen:
Thanks for the PR. Is this something you ran into in practice? I'd expect the varargs API to be used as e.g. `Tags.of(Tag.of("a", "b"), Tag.of("c", "d"))` rather than passing a separately held array. I'm curious if there's such usage out there and why.

kdelay:
Not from a production incident, no. I found it diffing this class between 1.12 and 1.13.

There is such a caller in this repo: `benchmarks-core` holds `static final Tag[] unorderedTagsSet{2,4,10}` and passes them to `Tags.of(...)` on every invocation. Printing the array around the calls on 1.16.x:

```
before      : key1 key2 key3 key0
after call 1: key0 key1 key2 key3
```

From the second invocation on, those benchmarks measure an already-sorted input. `KeyValuesBenchmark` has the same shape.

Retention behaves the same: after `Tags.of(mine)`, assigning `mine[0]` changes the returned `Tags`.

shakuzen:
Thank you for the details. We'll have to take a closer look after the ongoing work for the upcoming milestone release.

PR #7962, opened 2026-09-15
Setup reproducible builds

jonatan-ivanov:


build.gradle
```diff
@@ -389,8 +398,18 @@ subprojects {
         if (project.name != 'micrometer-bom') {
             apply plugin: 'biz.aQute.bnd.builder'
 
+            tasks.withType(AbstractArchiveTask).configureEach {
+                preserveFileTimestamps = false
+                reproducibleFileOrder = true
```
shakuzen: I thought these were the default since Gradle 9: https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#reproducible_archives_by_default

shakuzen (approved):

build.gradle
jonatan-ivanov: Right, we don't need it here, only where we are on Gradle 8.

PR #7963, opened 2026-09-16
Fix MeterEquivalence.equals(null, null) to return true

### LINKED ISSUE #7934: MeterEquivalence.equals(null, null) returns false, inconsistent with Objects.equals
`MeterEquivalence.equals` has a `// TODO fix somehow or report bug?` comment. Tracing through it:

https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/util/MeterEquivalence.java#L29-L43

When both `m1` and `o` are `null`, neither of the first two checks triggers (both require exactly one side non-null), so it falls to `!(null instanceof Meter)` and returns `false` -  inconsistent with the usual `Objects.equals(null, null) == true` convention.

Confirmed by compiling the class in isolation and running it directly:

MeterEquivalence.equals(null, null) = false
Objects.equals(null, null) (for comparison) = true

Minor in practice - the only current caller (`AbstractMeter.equals()`) passes `this` as `m1`, which can never be `null`, so this path isn't reachable today. But `MeterEquivalence` is `public`, so a direct call with `null, null` would hit it. Given the existing TODO, thought it was worth filing.

Suggested fix:

```java
if (m1 == o)
    return true;
if (m1 == null || o == null)
    return false;
if (!(o instanceof Meter))
    return false;
Meter m2 = (Meter) o;
return m1.getId().equals(m2.getId());
```

Happy to open a PR with this plus a test if useful.

tkatta-stack:
## What
`MeterEquivalence.equals(null, null)` returned `false`, which is inconsistent with the standard Java equality convention (`Objects.equals(null, null)` returns `true`).

The old implementation checked whether either argument was non-null before reaching the `instanceof` check, so `(null, null)` fell through to `!(null instanceof Meter)`, i.e. `false`.

## Fix
Rewrote `equals` using the conventional null-safe pattern: identity check, then null check on either argument, then type check, then id comparison. This also removes the need for the `@SuppressWarnings("NullAway")` that was papering over the old control flow.

## Tests
Added `MeterEquivalenceTest` covering: both null, only first null, only second null, same instance, wrong type, same id (different instance), different id.

Fixes #7934


jonatan-ivanov:
This seems like a copy of https://github.com/micrometer-metrics/micrometer/pull/7952
Please check [`CONTRIBUTING.md`](https://github.com/micrometer-metrics/micrometer/blob/ca888dce7a5ddec85357e15638aae18c4a4c7e5f/CONTRIBUTING.md#include-a-signed-off-by-trailer), it tells you about what you certify when you sign the DCO, see:
- https://developercertificate.org/
- https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring

Please stop copying other people's work, this is your first and final warning.

jonatan-ivanov:
Duplicate of https://github.com/micrometer-metrics/micrometer/pull/7952

